PR #9, opened 2020-12-05
feat: setup Yarn workspaces

Kocal:
Hi! :wave: 

Since this repository is a multi-packages repository, I thought it would be interesting and useful to use Yarn workspaces on it (since we use Yarn).

This will prevent some code duplications (e.g.: in the CI) and make it more "DX-friendly" if you want to build, test, lint (... :eyes:) all packages in one time:
Capture d’écran de 2020-12-05 10-30-07

What do you think?
Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,8 @@
+{
+    "private": true,
+    "workspaces": ["src/*/Resources/assets"],
+    "scripts": {
+        "build": "yarn workspaces run build",
+        "test": "yarn workspaces run test"
+    }
+}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -85,23 +85,5 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@master
-            - name: Chartjs
-              run: |
-                  cd src/Chartjs/Resources/assets
-                  yarn
-                  yarn test
-            - name: Cropperjs
-              run: |
-                  cd src/Cropperjs/Resources/assets
-                  yarn
-                  yarn test
-            - name: Dropzone
-              run: |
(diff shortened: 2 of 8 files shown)


tgalopin:
Hi @Kocal!

We had a look at how workspaces could be useful for the feature of Symfony UX itself, but didn't consider it for the mono repo indeed. I don't use workspaces personally, but they would seem to make sense. 

Do you know of other mono repos in JS using them? My main concern would be to rely too much on them if at some point yarn disappear.

Kocal:
Here are some monorepos that I know. They either use [Lerna](https://github.com/lerna/lerna), Yarn workspaces, or both at the same time:
- https://github.com/facebook/jest
- https://github.com/babel/babel
- https://github.com/vuejs/vue-cli
- https://github.com/vuejs/vue-next

Also note that npm 7 implemented workspaces support (https://docs.npmjs.com/cli/v7/using-npm/workspaces) ([RFC](https://github.com/npm/rfcs/blob/latest/implemented/0026-workspaces.md)), so if something is going wrong with Yarn we could easily migrate to npm workspaces instead.

tgalopin:
Awesome, thanks! Let's do it then :) .

Could you rebase your PR on main to run the CI (I added the CI run on PR)?

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "private": true,
```
tgalopin, line 2: I used 4 spaces in every package.json (to follow the convention we use in PHP), I think we should stick with that. WDYT?

Kocal: Yeah let's go for it!
IMO the best things would to run Prettier on this file too. 

Stay tuned for a next PR... 👀

tgalopin:
Thank you @Kocal.

PR #13, opened 2020-12-05
Remove trailing whitespaces

derrabus:


Kocal:
IMO we should also run Prettier on markdown files

Issue #11, opened 2020-12-05
options are not considered when rendering chart


BigMichi1:
i tried to set some options for the chart in my controller like this

```php
        $chart->setOptions([
            'title' => ['display'=> true, 'text' => $this->translator->trans('headline', ['%year%' => $year])],
            'legend' => ['display' => true, 'position' => 'bottom'],
        ]);
```

this is then rendered as
```json5
{
  // omitting the data part that works
  "options": {
    "title": {
      "display": true,
      "text": "Statistik 2020"
    },
    "legend": {
      "display": true,
      "position": "bottom"
    }
  }
}
```

but the check here https://github.com/symfony/ux/blob/11bde53628b1d66d7b00ce3721cf9e2773e65990/src/Chartjs/Resources/assets/src/controller.js#L18
replaces the options with an empty object, because `payload.options.length` is undefined while `payload.options` contains the options

Kocal:
I think using `typeof payload.options === 'undefined' || Object.keys(payload.options).length === 0` would fix the issue.

tgalopin:
Ah, that's a bug! If you wish to open a PR, go for it! Otherwise I'll create it :)

tgalopin:
Your edit @Kocal was why I did it this way initially. I'm opening a PR :)

Kocal:
Nice :)

Gasjki:
Hey guys! I'm trying to do a workaround until everything is fixed. I extended the Stimulus controller as-in given example, but I don't know how to set the entrypoints. A little more clarity for this as well in documentation?

Thanks!

tgalopin:
Hi @Gasjki!

FYI I implemented a fix, we will release it tomorrow (at the same time as Stimulus 2 support :) ).

Regarding extension, I'm not sure what's your issue. Could you open a dedicated one describing your problem in more details?

Gasjki:
Hi @tgalopin ! My problem is not actually an issue and I think is not necessary to create a new issue here. My question was related to the extensibility. If we extend the JS controller (based on documentation example), how are we supposed to load it? That's not clear for me. 

Thanks!

tgalopin:
This has been fixed and released in 1.1.0, thanks for the report!

@Gasjki for support, please create a new issue or even better, join the Symfony Slack (symfony.com/slack): that's going to be much easier to help you :) !

PR #17, opened 2020-12-05
Fix Chartjs options support

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix https://github.com/symfony/ux/issues/11
| License       | MIT

Fix Chart.js options handling and add better tests to ensure no regression there.

### DIFF
--- a/src/Chartjs/Resources/assets/dist/controller.js
+++ b/src/Chartjs/Resources/assets/dist/controller.js
@@ -55,7 +55,7 @@ var _default = /*#__PURE__*/function (_Controller) {
     value: function connect() {
       var payload = JSON.parse(this.element.getAttribute('data-view'));
 
-      if (!payload.options.length) {
+      if (Array.isArray(payload.options) && 0 === payload.options.length) {
         payload.options = {};
       }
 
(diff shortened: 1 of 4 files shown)


src/Chartjs/Resources/assets/dist/controller.js
```diff
@@ -55,7 +55,7 @@ var _default = /*#__PURE__*/function (_Controller) {
     value: function connect() {
       var payload = JSON.parse(this.element.getAttribute('data-view'));
 
-      if (!payload.options.length) {
+      if (!Array.isArray(payload.options) && 0 === payload.options.length) {
```
Nyholm, line 58: Hm.. If it is not an array AND length is zero.. 

Doesn't that mean that `payload.options = {}`?

tgalopin: In JS, `[] !== {}`, which is the problem here, as `json_encode` creates `[]` when we would like `{}`

Nyholm: > In JS, [] !== {},

Yes,

> as json_encode creates [] when we would like {}

But, that is not what I read... 

If `payload.options = []` then `!Array.isArray(payload.options)` will be false and you will never set `payload.options = {}`

Kocal: What we can do is to always have a default value for `options`?

tgalopin: Right! I have an extra `!`

fabpot (approved):

fabpot:
Thank you @tgalopin.

PR #18, opened 2020-12-05
Setup ESLint, and lint/format files

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

Hi :wave: 

We are already using Prettier for the formatting and that's really nice, but using a linter is a must have, especially if it's an open-source project where many people can contribute.

I've configured ESLint (and Prettier) in the root `package.json`, because I saw that Jest were configured in workspaces's `package.json` files, but I can move it in external files if needed.

The ESLint configuration is pretty basic, we extends the recommended configuration and use the [Jest ESLint plugin](https://github.com/jest-community/eslint-plugin-jest#readme). 

This PR add 4 new Yarn commands:
- `lint`, run ESLint and fix linting issues
- `format`, run Prettier on `.js`, `.json` and `.md` files, and fix formatting issues
- `check-lint` (for the CI), check if ESLint was run
- `check-format` (for the CI), check if Prettier was run

Each workspaces have a new Yarn command `lint` that will run ESLint locally to the workspace, this way `@babel/eslint-parser` will find the Babel config file used by the workspace, and it won't fail if ESLint was ran in the repository root.

WDYT?
Thanks!

### DIFF
--- a/src/LazyImage/README.md
+++ b/src/LazyImage/README.md
@@ -5,8 +5,8 @@ image loading performance. It is part of [the Symfony UX initiative](https://sym
 
 It provides two key features:
 
-* a Stimulus controller to load lazily heavy images, with a placeholder
-* a [BlurHash implementation](https://blurha.sh/) to create data-uri thumbnails for images
+-   a Stimulus controller to load lazily heavy images, with a placeholder
+-   a [BlurHash implementation](https://blurha.sh/) to create data-uri thumbnails for images
 
 ## Installation
 
@@ -57,9 +57,9 @@ the BlurHash algorithm to create a light, blurred, data-uri thumbnail of the ima
 
 The `data_uri_thumbnail` function receives 3 arguments:
 
-* the server path to the image to generate the data-uri thumbnail for ;
-* the width of the BlurHash to generate
-* the height of the BlurHash to generate
+-   the server path to the image to generate the data-uri thumbnail for ;
+-   the width of the BlurHash to generate
+-   the height of the BlurHash to generate
 
 You should try to generate small BlurHash images as generating the image can be CPU-intensive.
 Instead, you can rely on the browser scaling abilities by generating a small image and using the
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,10 +20,9 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@master
-            - name: Prettier
-              run: |
-                  yarn global add prettier@^2.2.0
-                  ~/.yarn/bin/prettier --check src/*/Resources/assets/{src,test}/*.js --config .prettierrc.json
+            - run: yarn
+            - run: yarn check-lint
+            - run: yarn check-format
 
     tests-php-low-deps:
         runs-on: ubuntu-latest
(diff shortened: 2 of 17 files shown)


tgalopin:
Is that the same ESLint config as Webpack Encore? My most important concern is to be consistent regarding linting and formatting across the Symfony JS tools

Kocal:
That's not the same config, here I've just keep it very simple by extendind the recommended config by ESLint, where the Webpack Encore ESLint config is kind of big (see [.eslintrc.js](https://github.com/symfony/webpack-encore/blob/main/.eslintrc.js)).

If we want to keep the exact same config than Encore, maybe the best thing to do would to create a Symfony ESLint configuration that we could use like this:
```js
module.exports = {
  extends: ['@symfony/symfony'] // will resolve to @symfony/eslint-config-symfony package
}
```

But there are a lot of formatting rules that will surely conflict with Prettier.

tgalopin:
Is it possible to use ESLint only for actual linting (ie potential errors) and not coding style (which is already provided by prettier)?

Kocal:
Yes, by using [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) I think (I've used it in the PR).

So even if we have a Symfony ESLint config that contains coding styles rules, they will be disabled if we extends the Prettier config too:
```js
module.exports = {
  extends: ['@symfony/symfony', 'prettier']
}
```

src/LazyImage/README.md
```diff
@@ -5,8 +5,8 @@ image loading performance. It is part of [the Symfony UX initiative](https://sym
 
 It provides two key features:
 
-* a Stimulus controller to load lazily heavy images, with a placeholder
-* a [BlurHash implementation](https://blurha.sh/) to create data-uri thumbnails for images
+-   a Stimulus controller to load lazily heavy images, with a placeholder
```
stof, line 8: why extra spaces ?

Kocal: Because we are using a tab width of 4 spaces, see https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAqABAQ3QZRgSwFsBXAG2IGd1JYAnCU0uW9GCdUiTAEw8wC98pAJ7oAFnEwA3UUUwBzOBQA06AO74YYrOgAOpTGDhiG3ZgB0oGbAG0AQuVoAJTBW1F9cQghiYC0AF0ACjEYGF0KJAB6KIAjRzFMADo3KIBKVnYwWkl4dG4-TABaYlp8VjFiQlioTCEqADMIFjlFChBlEAhdfyh2pFBMWno1AAUhhH6QTFI1TGF2ztjaQwBrOBgcXUN8KHlkGFpiOE6vWLhuM24AGUw94gU4ADFmwj8CPeRp4jYOkFDCKQAOpiTRKbZGHCTTT4KSaYRfMAURYgXYUZgwUYreRvZANGbozoAKwoAA87CswOtNphvNddnA8QSTiASaScLt5EwAIrECDwJmkQkgba0dG0L5vWirbgQNRQP66MqwIH4bhaZAADgADJ0lRB0UCVrovkqlMwpIzOgBHPnwLHdKauIpQOAXC5-HK2-A5LEKXFIfFClnowj4A5HEOcnl2xmB5mdXyxVXqsTIAAsiZWQk5AGEIIQAyAlABWP6UOAAFUwsX6QeFUmOAEkoGZYDhsvgegBBVt4YRMQXogC+w6AA

tgalopin:
Thank you @Kocal.

PR #20, opened 2020-12-06
Add option to set swup main element via data attribute

ghost:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | 
| License       | MIT

This pull request adds the possibility to change the `mainElement` of the swup theme. For this purpose the new data attribute `data-main-element` is introduced.

### DIFF
--- a/src/Swup/Resources/assets/dist/controller.js
+++ b/src/Swup/Resources/assets/dist/controller.js
@@ -7,10 +7,24 @@ import SwupSlideTheme from '@swup/slide-theme';
 
 class default_1 extends Controller {
     connect() {
+        const dataContainers = this.containersValue;
+        const mainElement = this.mainElementValue || dataContainers[0] || '#swup';
+        const allElements = [mainElement].concat(dataContainers);
+        const containersList = allElements.filter((item, index) => {
+            return allElements.indexOf(item) === index;
+        });
         const options = {
-            containers: ['#swup'],
-            plugins: ['slide' === this.themeValue ? new SwupSlideTheme() : new SwupFadeTheme(), new SwupFormsPlugin()],
+            containers: containersList,
+            plugins: [
+                'slide' === this.themeValue
+                    ? new SwupSlideTheme({ mainElement: mainElement })
+                    : new SwupFadeTheme({ mainElement: mainElement }),
+                new SwupFormsPlugin(),
+            ],
         };
+        if (this.hasMainElementValue) {
+            options.mainElement = this.mainElementValue;
+        }
         if (this.hasAnimateHistoryBrowsingValue) {
             options.animateHistoryBrowsing = this.animateHistoryBrowsingValue;
         }
@@ -19,9 +33,6 @@ class default_1 extends Controller {
         }
         if (this.hasCacheValue) {
             options.cache = this.cacheValue;
-        }
-        if (this.hasContainersValue) {
-            options.containers = this.containersValue;
         }
         if (this.hasLinkSelectorValue) {
             options.linkSelector = this.linkSelectorValue;
@@ -45,6 +56,7 @@ default_1.values = {
     linkSelector: String,
     theme: String,
     debug: Boolean,
+    mainElement: String,
 };
 
 export { default_1 as default };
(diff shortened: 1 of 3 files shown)


src/Swup/Resources/assets/src/controller.js
```diff
@@ -18,12 +18,20 @@ import SwupSlideTheme from '@swup/slide-theme';
 
 export default class extends Controller {
     connect() {
+        let mainElement = '#swup';
+
+        if (this.element.getAttribute('data-main-element')) {
+            mainElement = this.element.getAttribute('data-main-element');
```
tgalopin, line 24: You can use a single line notation (and const):

```javascript
const mainElement = this.element.getAttribute('data-main-element') || '#swup';
```

tgalopin (commented):
On line 40 (`options.containers = this.element.getAttribute('data-containers').split(' ');`), the containers are overriden, which may create a weird behavior if you use both mainElement and containers.

WDYT we should do if both are present in the HTML?

optior:
That's a good question. If the attribute `data-containers` is meant to specify addtional css selectors for page reloading those values should be merged (`data-containers` and `data-main-element`). WDYT?

tgalopin:
Yes, I think that would be the best course of action:

1/ if neither a main element nor containers are provided, use `#swup` by default
2/ if only a data-main-element is provided, use it both as a container and in themes
3/ if only a data-containers is provided, use it in containers and consider the first container as the main element
4/ if both are provided, always add the main element in the containers list if it's not already, use the main element in themes and use the containers in Swup

I think that makes the most sense

Also, can you rebase on main as the controller changed a bit?

optior:
I rebased my branch on the current main and implemented the required code.

Do I have to build `src/Swup/Resources/assets/dist/controller.js` or is this file generated automatically?

tgalopin:
You should build it :)

stof:
> You should build it :)

@tgalopin it might be worth improving the CI to ensure this does not get forgotten, if possible.

tgalopin:
Hey @optior !

I had a local look at your PR, it's great, thanks!

The final thing that I think we can do it add tests for the various situations. In `test/controller.test.js`, there is a first initial test that doesn't check much things, but as we can access the options from inside a CheckController it should be possible to ensure the right `containers` option is passed to the Swup instance. Do you think you can create such tests for the different cases we talked about earlier?

optior:
I'll give it a try on the weekend, never done any js testing before.

tgalopin:
Awesome, thanks! Feel free to ping me on Symfony Devs Slack if you have questions!

nicolas-grekas:
Friendly ping @optior

optior:
I have no clue how to write those tests. Unfortunately I have currently no time to dig into it. It would be great if anybody else could write these tests.

optior:
Finally I got the tests working..but now I can't fix the linter errors.

@tgalopin Are the tests ok this way? Any idea how I can fix the linter errors?

src/Swup/Resources/assets/test/controller.test.js
```diff
@@ -56,16 +58,135 @@ describe('SwupController', () => {
...
+                        </div>
+                    </div>
+                </body>
+            </html>
+        `);
+        expect(getByTestId(container, 'body')).not.toHaveClass('connected');
+
+        startStimulus();
+        await waitFor(() => expect(getByTestId(container, 'body')).toHaveClass('connected'));
+        await waitFor(() => expect(swup.options.containers).toStrictEqual(["#swup"]));
```
tgalopin, line 94: swup isn't accessible here IIUC

optior: I solved this issue

weaverryan:
Hey @optior!

I'm sorry hasn't gotten more attention yet :/. Would you be willing to update it to the latest version? The big change is that we refactored away from `data-` attributes to use Stimulus values. I think, since you understand this PR, that should be a fairly simple change. But let me know: I'd like to get this merged one way or another.

Thanks!

ghost:
Hey @weaverryan,

I'll update this pr in the next couple days.

ghost:
Hey @weaverryan,

I just pushed the updated files.

weaverryan (approved):

weaverryan:
Thank you @optior! And nice additions to the tests too :)

PR #23, opened 2020-12-07
Improve composer.json files

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | -
| License       | -

Removes the `config` section, and replaces it with:
```json
    "provide": {
        "symfony/ux-component": "1.1"
    },
    "extra": {
        "branch-alias": {
            "dev-main": "1.1-dev"
        },
        "thanks": {
            "name": "symfony/ux",
            "url": "https://github.com/symfony/ux"
        }
    },
    "minimum-stability": "dev"
```

### DIFF
--- a/src/Chartjs/composer.json
+++ b/src/Chartjs/composer.json
@@ -37,10 +37,14 @@
         "symfony/twig-bundle": "^4.4.17|^5.0",
         "symfony/var-dumper": "^4.4.17|^5.0"
     },
-    "config": {
-        "preferred-install": {
-            "*": "dist"
+    "extra": {
+        "branch-alias": {
+            "dev-main": "1.1-dev"
         },
-        "sort-packages": true
-    }
+        "thanks": {
+            "name": "symfony/ux",
+            "url": "https://github.com/symfony/ux"
+        }
+    },
+    "minimum-stability": "dev"
 }
--- a/src/Cropperjs/composer.json
+++ b/src/Cropperjs/composer.json
@@ -41,10 +41,14 @@
         "symfony/twig-bridge": "^4.4.17|^5.0",
         "symfony/var-dumper": "^4.4.17|^5.0"
     },
-    "config": {
-        "preferred-install": {
-            "*": "dist"
+    "extra": {
+        "branch-alias": {
+            "dev-main": "1.1-dev"
         },
-        "sort-packages": true
-    }
+        "thanks": {
+            "name": "symfony/ux",
+            "url": "https://github.com/symfony/ux"
+        }
+    },
+    "minimum-stability": "dev"
 }
--- a/src/Dropzone/composer.json
+++ b/src/Dropzone/composer.json
@@ -39,10 +39,14 @@
         "symfony/twig-bundle": "^4.4.17|^5.0",
         "symfony/var-dumper": "^4.4.17|^5.0"
     },
-    "config": {
-        "preferred-install": {
-            "*": "dist"
+    "extra": {
+        "branch-alias": {
+            "dev-main": "1.1-dev"
         },
-        "sort-packages": true
-    }
+        "thanks": {
+            "name": "symfony/ux",
+            "url": "https://github.com/symfony/ux"
+        }
+    },
+    "minimum-stability": "dev"
 }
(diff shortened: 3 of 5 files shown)


src/Chartjs/composer.json
```diff
@@ -37,10 +37,17 @@
         "symfony/twig-bundle": "^4.4.17|^5.0",
         "symfony/var-dumper": "^4.4.17|^5.0"
     },
-    "config": {
-        "preferred-install": {
-            "*": "dist"
+    "provide": {
+        "symfony/ux-component": "1.1"
```
tgalopin, line 41: I suggest to create a dedicated PR for this: it's an interesting topic to have a dedicated PR for, in order to link it on a blog post/documentation/tweets/...

lyrixx: I'm not sure this is right anyway. provide is for "implementing an interface". this is not the case here

https://getcomposer.org/doc/04-schema.md#provide

tgalopin: The idea behind this provide is to easily find all the packages implementing Symfony UX features. Technically, it's not a PHP interface, but it's a contract that the package respect a certain structure. I'd be fine merging this, it would increase the discoverability of UX packages.

stof: I would say that this should be a keyword instead

tgalopin: Can packages with a given keyword be listed on packagist?

lyrixx: I mean, If it has been `symfony/ux-component-chart` (note the trailing chart) it would be fine.
But here, as soon as you'll have one package that provide `symfony/ux-component`, it'll be strange to install another one.

More over if I build a lib, like easy admin, I do not make sens to require `symfony/ux-component-implementation` neither.

I do recommend you to not use `provide`, but instead a new `type`

stof: @tgalopin yes. the search feature allows to filter by keywords, as they are mapped as tags in algolia.

using a keyword is also better than a custom type, as the composer package itself is not a special type when it contains a ux component (and it will often be a symfony bundle and you cannot have multiple types)

tgalopin: @nicolas-grekas let's use a type?

tgalopin: @stof I feel like the type makes more sense: it could be any php package (bundle or not), the key thing is that it ships JS in addition to PHP.

stof: @tgalopin the main reason of the type is to be able to customize the way composer will install the package (custom installers can register themselves based on the type). the fact that a ux component can be shipped by any PHP package is *precisely* the reason why using the type is bad IMO: it hijacks the type for itself with no need for that as it does not rely on customizing the composer installation, which then makes it incompatible with PHP packages that actually have a legitimate reason to use a custom type.

See https://packagist.org/?tags=ux for keyword-based filtering in the search

stof: using a type could make sense if the JS component was the only content of the package (but then it should not be on packagist.org but on npmjs.com). If it is *in addition* to PHP code, using a field allowing a single value restricts it too much.

tgalopin: @nicolas-grekas I suggest we create a dedicated PR for that anyway, so we should probably remove it from this PR

fabpot: I'm also :-1: for this change, let's move the discussion elsewhere so that we can merge the rest of the changes.

nicolas-grekas: To me, this declaration fits an interface: this declares that UX packages follow the conventions required to implement the "interface" that flex requires to make UX work.

I would go as far as making it a requirement for a package to declare this "provide" entry before flex considers wiring it. There will be no downside in doing so.

The "provide" entry will also create a page like [this one](https://packagist.org/providers/php-http/client-implementation) on packagist. That should ease discoverability.

Maybe we could ease a composer "type" entry, but they're much less discoverable on packagist. Same for tags.

nicolas-grekas: Entries removed, I'm going to submit them again in a separate PR to continue this discussion.

lyrixx (approved):

tgalopin (approved):

tgalopin:
Thank you @nicolas-grekas.

PR #28, opened 2020-12-07
Use "symfony/ux" tag

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | -
| License       | MIT

As advised by @stof in https://github.com/symfony/ux/pull/23#discussion_r537572307, here's using a tag instead of a "provide" entry.

This should allow:
1. to have flex install packages as ux-components if-and-only-if they have the tag,
2. to help discover ux-related packages by linking to https://packagist.org/?tags=symfony/ux

### DIFF
--- a/src/Chartjs/composer.json
+++ b/src/Chartjs/composer.json
@@ -3,6 +3,7 @@
     "type": "symfony-bundle",
     "description": "Chart.js integration for Symfony",
     "keywords": [
+        "symfony/ux",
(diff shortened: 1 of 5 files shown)


fabpot (approved):

tgalopin:
Thank you @nicolas-grekas.

PR #29, opened 2020-12-07
Fix data-theme attribut

cedriclombardot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | 
| License       | MIT

hasAttribute return a boolean not attribute value

### DIFF
--- a/src/Swup/Resources/assets/src/controller.js
+++ b/src/Swup/Resources/assets/src/controller.js
@@ -23,7 +23,7 @@ export default class extends Controller {
             cache: this.element.hasAttribute('data-cache'),
             animateHistoryBrowsing: this.element.hasAttribute('data-animate-history-browsing'),
             plugins: [
-                'slide' === this.element.hasAttribute('data-theme') ? new SwupSlideTheme() : new SwupFadeTheme(),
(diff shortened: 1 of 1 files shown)


src/Swup/Resources/assets/src/controller.js
```diff
@@ -23,7 +23,9 @@ export default class extends Controller {
             cache: this.element.hasAttribute('data-cache'),
             animateHistoryBrowsing: this.element.hasAttribute('data-animate-history-browsing'),
             plugins: [
-                'slide' === this.element.hasAttribute('data-theme') ? new SwupSlideTheme() : new SwupFadeTheme(),
+                this.element.hasAttribute('data-theme') && 'slide' === this.element.getAttribute('data-theme')
```
tgalopin, line 26: You should be able to just use `'slide' === this.element.getAttribute('data-theme')` as getAttribute will just return null if no attribute exists

cedriclombardot: @tgalopin fixed :)

tgalopin:
Thank you @cedriclombardot.

PR #31, opened 2020-12-08
feat(ci): check changes after yarn build

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Tickets       | Fix #... 
| License       | MIT

This is a first iteration for https://github.com/symfony/ux/pull/20#issuecomment-740803858.

We used a similar thing at work and it was enough for us, but maybe we should be more friendly/precise, like posting a comment on the PR with the list of files that have not been built.
WDYT @tgalopin?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -90,3 +90,17 @@ jobs:
             - uses: actions/checkout@master
             - run: yarn
             - run: yarn test
+
+    tests-build:
+        name: Check if JS assets are built
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


tgalopin:
Hi @Kocal !

Can you avoid creating Draft PR please? It notifies everyone watching the repo for unnecessary reasons :) .

Once your PR is ready we would be happy to review it though!

tgalopin:
After discussing with @weaverryan I think the best course of action in the short term is to have a release script that includes building the dist file. I'll add it to the script we use? Thanks for the PR anway!

Kocal:
Alright!

PR #39, opened 2020-12-13
Add Leaflet.js package

### LINKED ISSUE #38: Proposal: map component
a map component based on the lightweight https://leafletjs.com/ library would be a very good addition

BigMichi1:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #38 
| License       | MIT

This is a basic implementation for rendering a map with Leaflet, which allows further customization in a custom user controller.

This implementation adds a new Map with one Layer based on OpenStreetMap and displays the provided coordinates from PHP.

### DIFF
--- a/src/Leafletjs/Resources/assets/src/controller.js
+++ b/src/Leafletjs/Resources/assets/src/controller.js
@@ -1,0 +1,57 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import { Controller } from 'stimulus';
+
+export default class extends Controller {
+    static targets = ['placeholder'];
+
+    connect() {
+        import('leaflet').then((L) => {
+            let mapOptions = this._extractOptions('map-options');
+
+            this.map = L.map(this.placeholderTarget, mapOptions).setView(this._coordinates(), this.data.get('zoom'));
+
+            let layer = L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
+                attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+            });
+
+            layer.addTo(this.map);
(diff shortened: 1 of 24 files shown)


tgalopin (commented):
Thanks for your PR, that would be an awesome addition to Symfony UX! I did a quick review, only a small comment. My only concern for now is options and features.

UX Chart.js (which I guess you took inspiration from) has the advantage of relying on a quite simple library: almost everything is defined through an array of options that can be encoded as JSON, meaning 90% of the features of Chart.js are accessible through PHP.

I'd like for UX Leaflet.js to be usable as much as possible in PHP, but I think that's not yet the case. For instance, we should be able to define markers, or to change which provider to use (OSM/...). Some other ideas I didn't think of yet could be important as well.

To summarize: a great start, but I think we need to think deeply about what need to be put inside this :) !

tgalopin:
After thinking about it and following a proposal of @tarlepp I think there is an interesting idea relative to maps in UX: providing a way to choose a provider and display a map based on it (Leaflet, OSM, Google Maps, Mapbox, ...).

This would be very useful, especially as I think there are only 2 key usages we need to provide for this package to make sense:

* displaying a map with a given position and zoom ;
* displaying markers on this map, ideally with the ability to add a name/icon to use for the markers (if possible using the provider)

Then, additional options could be provided for specific implementations.

I feel having such package could be a great addition to UX!

src/Leafletjs/Resources/assets/src/controller.js
```diff
@@ -0,0 +1,57 @@
...
+export default class extends Controller {
+    static targets = ['placeholder'];
+
+    connect() {
+        import('leaflet').then((L) => {
+            let mapOptions = this._extractOptions('map-options');
+
+            this.map = L.map(this.placeholderTarget, mapOptions).setView(this._coordinates(), this.data.get('zoom'));
+
+            let layer = L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
```
wouterj, line 23: I think it's good to also make the tile server configurable, to allow other map styles (or other OSM tile servers if e.g. their usage policy doesn't comply with your usage).

Issue #33, opened 2020-12-09
ux-dropzone - Error connecting controller


dprzygodzki:
I'm doing something wrong. Once implemented ux-dropzone, I'm getting an error in the browser console:

> Error connecting controller
>
> DOMException: Failed to execute 'matches' on 'Element': '[data-@symfony/ux-dropzone/dropzone-target~="input"]' is not a valid > selector.

tgalopin:
Hi! Can you paste your usage code here?

tgalopin:
By seeing `[data-@symfony/ux-dropzone/dropzone-target~="input"]`, it feels like you're implementing your own HTML, is that the case? If so, can you tell me more about how you use the package?

Fredxd:
Hi @tgalopin,
I have same here.

## Front side

* I have move my controllers.js and my bootstrap.js file into shared folder in assets/.
* I have correctly called my boostrap.js in my entry.js file
* In my controllers.js i have add
```
"@symfony/ux-dropzone": {
            "dropzone": {
                "enabled": true,
                "webpackMode": "eager",
                "autoimport": {
                    "@symfony/ux-dropzone/src/style.css": true
                }
            }
        }
```

* In my webpack.config.js i have correctly change my path for enableStimulusBridge

## Back Side
* I have add new property imageFile like this (I use VichUploader: it's maybe that ?)
```
 /**
     * @Vich\UploadableField(mapping="adverts", fileNameProperty="imageName")
     */
    private ?File $imageFile;
```

* In my FormType i have
```
            ->add('imageFile', DropzoneType::class, [
                'label' => 'Image',
                'label_attr' => [
                    'class' => 'text-right',
                ],
            ])
```

I hope this can help you

Ncaln:
I am also having the same issue.

Could this be related to stimulus new syntax for target attributes?
> The new format is data-[identifier]-target="[name]" instead of data-target="[identifier].[name]"

tgalopin:
It seems to be ... but I'm not sure how to reproduce! If you have a reproducer for me that would be awesome :)

Fredxd:
> It seems to be ... but I'm not sure how to reproduce! If you have a reproducer for me that would be awesome :)

@tgalopin Hi,
You can see above what i did in order to have this issue ;)

Ncaln:
It does look like the form_theme.html.twig uses the old format. Is that the issue, or should it still function properly?

disjfa:
[...]

PR #43, opened 2020-12-31
Implement normalized controller names

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix https://github.com/symfony/ux/issues/33
| License       | MIT

Following https://github.com/symfony/stimulus-bridge/pull/12, implement controller names normalizing.

This will be a BC break for lazy-image and Swup users but it only requires a small change and is necessary for the development of additional UX packages based on Stimulus 2.

### DIFF
--- a/src/Chartjs/Tests/Twig/ChartExtensionTest.php
+++ b/src/Chartjs/Tests/Twig/ChartExtensionTest.php
@@ -59,7 +59,7 @@ class ChartExtensionTest extends TestCase
 
         $this->assertSame(
             '<canvas
-                data-controller="mycontroller @symfony/ux-chartjs/chart"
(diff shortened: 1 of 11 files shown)


lyrixx:
Should we document such BC break somewhere, to help people upgrading?

Not related: IMHO you should use .gitattribute to do not display compiled file in diff

tgalopin:
Yes, I planned to add it to the release, but it's true we don't have a CHANGELOG yet and it may be useful. I'll have a look next year ;) .

konshensx16:
Happy new year everyone and thank for the great package. 
Any idea when this PR will be merged ? :)

nicolas-grekas (approved):

PR #49, opened 2021-01-21
[Notify] Add Notify bundle

mtarld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | 
| License       | MIT

In Symfony 5.3, Notifier was shipped with a Mercure bridge (https://github.com/symfony/symfony/pull/39342)

Then, with a simple `$this->notifier->send(new Notification('My message', ['chat/mercure']));` (and a running Mercure server), it'll be easy to create "Server-Sent-Events".

Therefore, the Notify bundle idea is to listen to these events using JavaScript event sourcing and convert them as native HTML5 notifications (using the `{{ stream_notifications() }}` Twig function).

(The "Notify" name was the first that came to my mind a could/should be challenged)


src/Notify/Resources/assets/src/controller.js
```diff
@@ -0,0 +1,60 @@
...
+'use strict';
+
+import { Controller } from 'stimulus';
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ */
+export default class extends Controller {
+    connect() {
+        const hubUrl = this.element.getAttribute('data-mercure-hub-url');
```
dunglas, line 19: Shouldn't you use [values](https://stimulus.hotwire.dev/reference/values) instead?

src/Notify/Resources/assets/src/controller.js
```diff
@@ -0,0 +1,60 @@
...
+        const topic = this.element.getAttribute('data-mercure-topic');
+
+        if (!('Notification' in window) || !hubUrl || !topic) {
+            this._dispatchEvent('notify:connect');
```
dunglas, line 23: Is it a good idea to dispatch this event when it's not possible to connect like in this case?

dunglas (changes requested):

src/Notify/Resources/assets/src/controller.js
mtarld: Yes indeed! Thanks

src/Notify/Twig/NotifyExtension.php
```diff
@@ -0,0 +1,55 @@
...
+        ];
+    }
+
+    public function notify(?string $topic = null, ?string $hubUrl = null): string
+    {
+        $topic = $topic ?: 'https://symfony.com/notifier';
+        $hubUrl = $hubUrl ?: $this->defaultHubUrl;
+
+        $html = '
+            <div
```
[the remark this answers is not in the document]
mtarld: As discussed together, it does make sense but it has to be more meaningful. Therefore I renamed `{{ notify() }}` to `{{ stream_notifications() }}`

seb-jean:
Hi 😃 ,

With this bundle, will it be web push notifications (with allow and block) or simple notifications?

Push notifications :
Ca1pture

notification-message

Simple notification :
Capture

Thanks :)

tgalopin:
Push notifications (native)

tgalopin:
Hello @mtarld 

I think this component still makes sense, could we perhaps use latest Symfony changes (especially on ux-mercure) to finalize it? Would you have time to have a look at the failing tests/rebasing?

mtarld:
Hi @tgalopin! 
Yes indeed, I think it is worth it! A lot of improvements and changes happened, I'll update the bundle consequently :slightly_smiling_face:

mtarld:
Hi @tgalopin! I updated the bundle to be consistent with other bundles (and to reuse some good ideas taken from ux-turbo-mercure)

Still, some questions remain: 
- As the notify bundle is tightly coupled to mercure, should we name the bundle after it? Something like `MercureNotify` with a `stream_mercure_notification` twig function?
- Actually, `symfony/mercure-notifier` is a strong composer dependency of the bundle. But it could actually work without (as the bundle is just a "listener"). Maybe we can instead change that dependency to a composer suggestion?

src/Notify/README.md
```diff
@@ -0,0 +1,167 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class UseCustomNotifyMercureHubCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $container->getDefinition('notify.twig_extension')
+            ->replaceArgument(0, new Reference('mercure.hub.custom'));
```
yceruto: If this is a common need, I'd suggest creating a config for it, so one could change easily the Mercure hub linked to this service:
```yaml
notify:
    mercure_hub: mercure.hub.default # default service id
```

mtarld: That's a great idea indeed! :+1:

src/Notify/DependencyInjection/NotifyExtension.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\UX\Notify\Twig\NotifyExtension as TwigNotifyExtension;
+use Twig\Environment;
+
+/**
+ * @author Mathias Arlaud <mathias.arlaud@gmail.com>
+ *
+ * @internal
+ */
+final class NotifyExtension extends Extension
```
yceruto: Now that Configuration is added, you can use `ConfigurableExtension` for code simplification.

mtarld: Didn't know about it! Many thanks

src/Notify/DependencyInjection/NotifyExtension.php
```diff
@@ -0,0 +1,43 @@
...
+    public function loadInternal(array $config, ContainerBuilder $container)
+    {
+        if (class_exists(Environment::class)) {
+            $container->register('notify.twig_extension', TwigNotifyExtension::class)
+                ->setArguments([
+                    new Reference($config['mercure_hub']),
+                    new Reference('webpack_encore.twig_stimulus_extension'),
+                ])
+                ->addTag('twig.extension')
+                ->setPublic(false)
```
yceruto: you can omit this definition, it's false by default

src/Notify/DependencyInjection/NotifyExtension.php
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function loadInternal(array $config, ContainerBuilder $container)
+    {
+        if (class_exists(Environment::class)) {
+            $container->register('notify.twig_extension', TwigNotifyExtension::class)
+                ->setArguments([
+                    new Reference($config['mercure_hub']),
+                    new Reference('webpack_encore.twig_stimulus_extension'),
```
yceruto: should we ask for ` class_exists(StimulusTwigExtension::class)` too?

yceruto (approved):
LGTM

dunglas (approved):

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
+Symfony UX Notify
+===================
+
+Symfony UX Notify is a Symfony bundle providing light dropzones for
+file inputs in Symfony Forms. It is part of `the Symfony UX initiative`_.
+
```
kbond:
```suggestion

```

mtarld: Oops! Nice catch thanks!

src/Notify/Twig/NotifyExtension.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function __construct(HubInterface $hub, StimulusTwigExtension $stimulusTwigExtension)
+    {
+        $this->hub = $hub;
+        $this->stimulusTwigExtension = $stimulusTwigExtension;
+    }
+
+    public function getFunctions(): iterable
+    {
+        yield new TwigFunction('stream_notifications', [$this, 'renderStreamNotifications'], ['needs_environment' => true, 'is_safe' => ['html']]);
```
chalasr: let's leverage PHP 8.1+ first-class callables

```suggestion
        yield new TwigFunction('stream_notifications', $this->renderStreamNotifications(...), ['needs_environment' => true, 'is_safe' => ['html']]);
```

src/Notify/Twig/NotifyExtension.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * @param array|string $topics
+     */
+    public function renderStreamNotifications(Environment $environment, $topics = [], array $options = []): string
+    {
+        if (!\is_array($topics) && !\is_string($topics)) {
+            throw new \TypeError(sprintf('"%s()" expects parameter 1 to be an array of strings or a string, "%s" given.', __METHOD__, get_debug_type($topics)));
+        }
```
chalasr:
```suggestion
    private function renderStreamNotifications(Environment $environment, array|string $topics = [], array $options = []): string
    {
```

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+    # Don't forget to install the JavaScript dependencies as well and compile
+    $ yarn install --force
+    $ yarn encore dev
+
+Also make sure you have at least version 3.0 of
+`@symfony/stimulus-bridge`_ in your ``package.json`` file.
+
+Configuration
+-------------
```
weaverryan: I'd move this to the last section, as most people will not actually need to configure this. And if you DO have multiple hubs, you will be looking specifically for this config.

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+
+To use Symfony UX Notify you must have a `running Mercure server`_.
+
+Then, you can inject the ``NotifierInterface`` service and send messages on the ``chat/mercure`` channel::
+
+    // ...
+    use Symfony\Component\Notifier\Notification\Notification;
+    use Symfony\Component\Notifier\NotifierInterface;
+
+    class AnnounceFlashSalesCommand extends Command
```
weaverryan, line 52: We should include the `#[AsCommand]` above this, just to make this a complete example.

src/Notify/Twig/NotifyExtension.php
weaverryan: ~~Should we? I'd like to keep php 8.0 compat at least~~ Nope: we will require 8.1

src/Notify/composer.json
```diff
@@ -0,0 +1,54 @@
...
+    "autoload": {
+        "psr-4": {
+            "Symfony\\UX\\Notify\\": "."
+        },
+        "exclude-from-classmap": [
+            "/Tests/"
+        ]
+    },
+    "require": {
+        "php": ">=7.2.5",
```
weaverryan: I think, for new stuff, we should require php 8.0 at least. But I've asked the core team to weigh in

UPDATE: Require 8.1 as the minimum. And then, feel free to use any fancy 8.1 features you want ;)

src/Notify/composer.json
```diff
@@ -0,0 +1,54 @@
...
+        "symfony/config": "^4.4.17|^5.0|^6.0",
+        "symfony/dependency-injection": "^4.4.17|^5.0|^6.0",
+        "symfony/http-kernel": "^4.4.17|^5.0|^6.0",
+        "symfony/mercure-bundle": "^0.3.2",
+        "symfony/mercure-notifier": "^5.3|^6.0"
```
weaverryan: For Symfony packages, make 5.4 the minimum.

src/Notify/Twig/NotifyExtension.php
```diff
@@ -0,0 +1,62 @@
...
+final class NotifyExtension extends AbstractExtension
+{
+    private $hub;
+    private $stimulusTwigExtension;
+
+    public function __construct(HubInterface $hub, StimulusTwigExtension $stimulusTwigExtension)
+    {
+        $this->hub = $hub;
+        $this->stimulusTwigExtension = $stimulusTwigExtension;
+    }
```
weaverryan: Hmm, we should probably leverage a Twig extension "runtime class" - https://symfony.com/doc/current/templating/twig_extension.html#creating-lazy-loaded-twig-extensions - so that these dependencies (especially the hub) aren't loaded unless the user actually calls this Twig function.

src/Notify/Resources/assets/src/controller.ts
```diff
@@ -0,0 +1,85 @@
...
+            Notification.requestPermission().then((permission) => {
+                if ('granted' === permission) {
+                    new Notification(content);
+                }
+            });
+        }
+    }
+
+    _dispatchEvent(name: string, payload: any) {
+        this.element.dispatchEvent(new CustomEvent(name, { detail: payload }));
```
weaverryan: Also add `bubbles: true` as a sibling to `detail`. This is something we have done incorrectly in other packages so far.

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
+Symfony UX Notify
+===================
+
+Symfony UX Notify is a Symfony bundle providing light dropzones for
+file inputs in Symfony Forms. It is part of `the Symfony UX initiative`_.
+
+Symfony UX Notify is a Symfony bundle integrating realtime native
+notifications in Symfony applications using `Mercure`_.
+It is part of `the Symfony UX initiative`_.
```
weaverryan: I'd love a screenshot of a notification. I didn't really understand what you meant by "native notification" until I really dug into the code. Maybe also link `native notifications` to https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API

src/Notify/README.md
```diff
@@ -0,0 +1,15 @@
+# Symfony UX Notify
+
+Symfony UX Notify is a Symfony bundle integrating realtime native notifications
```
weaverryan: And also here, link `native notifications` to https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+
+        protected function execute(InputInterface $input, OutputInterface $output): int
+        {
+            $this->notifier->send(new Notification('Flash sales has been started!', ['chat/mercure']));
+
+            return 0;
+        }
+    }
+
+Finally, native notifications could be displayed using the ``stream_notifications`` Twig function:
```
weaverryan:
```rst
Finally, to "listen" and trigger the notifications in the user's browser, call the ``stream_notifications()`` Twig function anywhere on the page:
```

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. code-block:: twig
+
+    {{ stream_notifications(['/my/topic/1', '/my/topic/2']) }}
+    {{ stream_notifications() }}
+
+.. note::
+   *Note*: Calling ``stream_notifications`` without parameter will fallback to the
```
weaverryan:
```suggestion
   Calling ``stream_notifications()`` without parameter will fallback to the
```

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+            return 0;
+        }
+    }
+
+Finally, native notifications could be displayed using the ``stream_notifications`` Twig function:
+
+.. code-block:: twig
+
+    {{ stream_notifications(['/my/topic/1', '/my/topic/2']) }}
+    {{ stream_notifications() }}
```
weaverryan, line 72: I'm missing "the big picture" a little bit. Here is my understanding:

A) I send a `Notification` in PHP to the `chat/mercure` channel.
B) This communicates with https://github.com/symfony/mercure-notifier to send that notification to my Mercure hub
C) ??? How/where do the Mercure topics "connect" with this? If I listen to the `/my/topic/1` topic here... which notifications would I receive? Is the "topic" configured somewhere when sending the `Notification` to mercure?

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. note::
+   *Note*: Calling ``stream_notifications`` without parameter will fallback to the
+   following unique topic: ``https://symfony.com/notifier``.
+
+Extend the default behavior
```
weaverryan:
```suggestion
Extend the Stimulus Controller
```

might be better

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony UX Notify allows you to extend its default behavior using a
+custom Stimulus controller:
+
+.. code-block:: javascript
+
+    // mynotify_controller.js
```
weaverryan:
```suggestion
    // assets/controllers/mynotify_controller.js
```

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,140 @@
...
+
+    import { Controller } from '@hotwired/stimulus';
+
+    export default class extends Controller {
+        connect() {
+            this.element.addEventListener('notify:connect', this._onConnect);
```
weaverryan, line 105: I hate to have to worry about it, but with this example, users are going to have the `this` problem down in `_onConnect()`. So perhaps we should add:

```js
initialize() {
    // guarantees "this" refers to this object in _onConnect
    this._onConnect = this._onConnect.bind(this);
}
```

src/Notify/DependencyInjection/NotifyExtension.php
```diff
@@ -0,0 +1,43 @@
...
+ * @internal
+ */
+final class NotifyExtension extends ConfigurableExtension
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function loadInternal(array $config, ContainerBuilder $container)
+    {
+        if (class_exists(Environment::class) && class_exists(StimulusTwigExtension::class)) {
```
weaverryan: Why not make these "hard" dependencies in `composer.json`? My thinking: the main/only thing this bundle provides is the `stream_notifications()` Twig function. If you install this package and are missing either or these two dependencies (especially `StimulusTwigExtension`), it could confuse people to silently not register the new function. So, I think we make these both real dependencies.

(side note, you could argue that Encore shouldn't be needed... just to get the `StimulusTwigExtension` and that's valid. But that's a separate topic).

weaverryan (commented):
Very cool 😎

src/Notify/Resources/doc/index.rst
mtarld: When you send a notification through the mercure notifier, you'll have to specify channels in the `Notification`:
```php
$this->notifier->send(new Notification('Hello native notifications!', ['chat/mercureFoo']));
```

This channel is configured in the notifier configuration, there you can set a dsn:
```yaml
framework:
    notifier:
        chatter_transports:
           mercureFoo: '%env(MERCURE_FOO_DSN)%'
```

And it that very dsn, you'll be able to configure the topics to send the notification:
```
MERCURE_FOO_DSN=mercure://default?topics[]=fooOne&topics[]=fooTwo
MERCURE_BAR_DSN=mercure://default?topics[]=bar
```

Finally, you can subscribe the topics you want using the twig helper:
```twig
{{ stream_notifications(['fooOne']) }}
```

Is it now clearer for you? Maybe it worth to be written somewhere in the documentation?

src/Notify/DependencyInjection/NotifyExtension.php
mtarld: 100% agreed!

mtarld:
As we required PHP 8.1 for that bundle, I updated the test workflow to run the bundle's tests on PHP 8.1.
I think we may test other bundles over PHP 8.1 instead of 8.0 in another PR. 
WDYT?

mtarld:
Thanks for that wonderful review @weaverryan :slightly_smiling_face:!

src/Notify/Resources/doc/index.rst
weaverryan: Yes, you explained that VERY clearly. I understand that, technically, much of this should be documented in the notify + mercure bridge. But yes, it'd be great to include this in the README. I had originally thought that the `chat/mercure` channel used in your example was a special string that NEEDED to be exactly that string. Clearly, I was missing the point :)

weaverryan:
> 
As we required PHP 8.1 for that bundle, I updated the test workflow to run the bundle's tests on PHP 8.1.
I think we may test other bundles over PHP 8.1 instead of 8.0 in another PR.
WDYT?

The test matrix DOES need some work. But if a bundle advertises that it works on php8, we need to keep testing it on php 8.0 (though, we should probably also test on php 8.1). Btw, I think we need to improve the test matrix further to be dynamic. There is way too much repeated config, and that's only going to get worse as the number of components grows. If you're interested, I had a WIP version of this at https://github.com/weaverryan/ux/tree/typescript-central-ci - where each package created a dynamic build. I think it's an interesting approach.

For this PR, I'd love to see your excellent explanation put into the docs. But after that, it's ready to merge!

maxhelias (approved):

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,160 @@
...
+        notifier:
+            chatter_transports:
+               myMercureChatter: '%env(MERCURE_DSN)%'
+
+.. note::
+   It is possible to specify the topics to send the notification in the ``MERCURE_DSN``
+   environment variable by specifying the ``topics`` query parameter.
+   Otherwise, notifications will be sent to ``https://symfony.com/notifier`` topic.
+
+Then, you can inject the ``NotifierInterface`` service and send messages on the ``chat/mercure`` channel::
```
weaverryan, line 45:
```suggestion
Then, you can inject the ``NotifierInterface`` service and send messages on the ``chat/myMercureChatter `` channel::
```

src/Notify/Resources/doc/index.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Enjoy your server-sent native notifications!
+
+.. figure:: https://github.com/symfony/ux-notify/blob/2.x/Resources/native-notification-example.png?raw=true
+   :alt: Example of a native notification
+
+   Example of a native notification
+
+Extend the Stimulus Controller
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
weaverryan, line 87: This looks a tiny bit to short

weaverryan (approved):

weaverryan:
Thank you Mathias! Wonderful work!

weaverryan:
Docs tweak over at #325

PR #61, opened 2021-02-19
Introduce Symfony UX Turbo

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | -
| License       | MIT

This PR introduces a new package to Symfony UX: UX Turbo, integrating the [Hotwire Turbo](https://turbo.hotwire.dev) library with Symfony. It is based on an initial prototype made by Kevin Dunglas (@dunglas) as a [Mercure](https://mercure.rocks/) project.

Hotwire is an aggregation of several technologies developed by the Ruby on Rails ecosystem and the Bascamp company to build interactive applications with as less JavaScript code as possible. Stimulus, as the foundation of Symfony UX, is already well integrated into Symfony. This PR goes to the next step: integrating Turbo.

You can learn more about the Hotwire intiative on https://twitter.com/dhh/status/1341420143239450624. You can also read the [README in this PR](https://github.com/tgalopin/ux/blob/31e064fb2ab6801fd6bd62b9a1505c6125fb67da/src/Turbo/README.md) to better understand the usage in a Symfony application.

TLDR: Turbo is a generic way to build reactive applications (inline form edition, DOM changes without reload, real-time display of chat messages, etc.) without the need to develop any JavaScript (or at least as little as possible). This has two big advantages: it reduces dramatically the time necessary to build a reactive app and it eliminates in many places the need for a client-side state which could be a source of issues.

This PR is loosely based on the equivalent gem in Ruby that integrates Turbo in RoR. It has however a few key differences:

* it leverages Symfony Mercure by default for Streams instead of Websockets (as PHP isn't a long running process)
* it build upon other tools in Symfony like the `stimulus_controller` [helper introduced a few days ago in Webpack Encore](https://github.com/symfony/webpack-encore-bundle/pull/109)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -72,31 +72,26 @@ jobs:
             - name: Chartjs
               run: |
                   cd src/Chartjs
-                  composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
             - name: Cropperjs
               run: |
                   cd src/Cropperjs
-                  composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
             - name: Dropzone
               run: |
                   cd src/Dropzone
-                  composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
             - name: LazyImage
               run: |
                   cd src/LazyImage
-                  composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
             - name: Turbo
               run: |
                   cd src/Turbo
-                  composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -34,6 +34,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '7.2'
+                  extensions: gd
             - name: Chartjs
               run: |
                   cd src/Chartjs
@@ -52,6 +53,11 @@ jobs:
             - name: LazyImage
               run: |
                   cd src/LazyImage
+                  composer update --prefer-lowest --prefer-dist --no-interaction --no-ansi --no-progress
+                  php vendor/bin/simple-phpunit
+            - name: Turbo
+              run: |
+                  cd src/Turbo
                   composer update --prefer-lowest --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
 
@@ -62,6 +68,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '8.0'
+                  extensions: gd
             - name: Chartjs
               run: |
                   cd src/Chartjs
@@ -83,6 +90,12 @@ jobs:
             - name: LazyImage
               run: |
                   cd src/LazyImage
+                  composer config platform.php 7.4.99
+                  composer update --prefer-dist --no-interaction --no-ansi --no-progress
+                  php vendor/bin/simple-phpunit
+            - name: Turbo
+              run: |
+                  cd src/Turbo
                   composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
--- a/src/Turbo/DependencyInjection/Configuration.php
+++ b/src/Turbo/DependencyInjection/Configuration.php
@@ -39,6 +39,7 @@ class Configuration implements ConfigurationInterface
                             ->info('Key/value options passed to the adapter to create Turbo Streams')
                             ->normalizeKeys(false)
                             ->variablePrototype()
+                            ->end()
                         ->end()
                     ->end()
                 ->end()
--- a/src/Turbo/DependencyInjection/Configuration.php
+++ b/src/Turbo/DependencyInjection/Configuration.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ *
+ * @internal
+ */
+class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder('turbo');
+        $rootNode = $treeBuilder->getRootNode();
+
+        $rootNode
+            ->children()
+                ->arrayNode('streams')
+                    ->info('Enable this section to use Turbo Streams')
+                    ->children()
+                        ->scalarNode('adapter')
+                            ->defaultNull()
+                            ->info('The service to use to create Turbo Streams')
+                        ->end()
+                        ->arrayNode('options')
+                            ->info('Key/value options passed to the adapter to create Turbo Streams')
+                            ->normalizeKeys(false)
+                            ->variablePrototype()
+                        ->end()
+                    ->end()
+                ->end()
+            ->end()
+        ;
+
+        return $treeBuilder;
+    }
+}
--- a/src/Turbo/DependencyInjection/TurboExtension.php
+++ b/src/Turbo/DependencyInjection/TurboExtension.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo\DependencyInjection;
+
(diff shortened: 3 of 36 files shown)


src/Turbo/Twig/FrameTwigExtension.php
```diff
@@ -0,0 +1,56 @@
...
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ *
+ * @final
+ * @experimental
+ */
+class FrameTwigExtension extends AbstractExtension
```
nicolas-grekas, line 24: TurboFrameExtension?

src/Turbo/Twig/StreamTwigExtension.php
```diff
@@ -0,0 +1,80 @@
...
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ *
+ * @final
+ * @experimental
+ */
+class StreamTwigExtension extends AbstractExtension
```
nicolas-grekas, line 26: TurboStreamExtension?

src/Turbo/composer.json
```diff
@@ -0,0 +1,51 @@
...
+        "exclude-from-classmap": [
+            "/Tests/"
+        ]
+    },
+    "require": {
+        "php": ">=7.2.5",
+        "symfony/config": "^4.4.17|^5.0",
+        "symfony/dependency-injection": "^4.4.17|^5.0",
+        "symfony/http-kernel": "^4.4.17|^5.0",
+        "symfony/webpack-encore-bundle": "^v1.10.0"
```
nicolas-grekas, line 33: `^1.10`

src/Turbo/DependencyInjection/TurboExtension.php
```diff
@@ -0,0 +1,71 @@
...
+use Symfony\UX\Turbo\Streams\MercureStreamAdapter;
+use Symfony\UX\Turbo\Twig\FrameTwigExtension;
+use Symfony\UX\Turbo\Twig\StreamTwigExtension;
+use Symfony\WebpackEncoreBundle\Twig\StimulusTwigExtension;
+use Twig\Environment;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ *
+ * @internal
```
nicolas-grekas, line 27: I don't think we need all those `@internal` annotations, do we?
What about `@experimental` also? Some have it, not all.

src/Turbo/Twig/FrameTwigExtension.php
tgalopin: I followed the same naming strategy as Webpack Encore Bundle (https://github.com/symfony/webpack-encore-bundle/blob/main/src/Twig/StimulusTwigExtension.php). I quite like having the "TwigExtension" suffix to better understand the difference between DI and Twig extensions in IDEs. 

Adding Turbo as a prefix :)

src/Turbo/DependencyInjection/TurboExtension.php
tgalopin: I may be a bit conservative regarding these annotations indeed :) . Other UX packages have these annotations as well, not sure what our policy should be on them.

wouterj: I'm often adding `@internal` when the class should not be used in both framework and standalone usage. E.g. when someone is using this package outside the Symfony framework, they probably end up needing to initialize `MercureStreamAdapter`. So I wouldn't add `@internal` there.

As for `@experimental`, I'm often not adding it to `@internal` classes (as there is no BC promise in those anyway). Not sure if these are "Symfony policies" or only my personal policy regarding them :)

tgalopin:
@Kocal I'm not familiar with eslint, any idea why it fails on this package but not others? I may be missing required config but I don't see it

Kocal:
Hi @tgalopin, I think it's because of this https://github.com/tgalopin/ux/blob/ux-turbo/package.json#L34

I don't think this glob pattern will match `src/Turbo/Resources/assets/test/adapters/mercure-controller.test.js` since `mercure-controller.test.js` is under `adapters` folder.

Can you try using ` "src/*/Resources/assets/test/**/*.js"` instead?

tgalopin:
Aah I was looking in the package, I didn't think of this file. Let me try, thanks!

src/Turbo/Twig/FrameTwigExtension.php
nicolas-grekas: On the other end, none how the bridge's extension have Twig in their suffix:
https://github.com/symfony/symfony/tree/5.x/src/Symfony/Bridge/Twig/Extension
I don't think the suffix is useful personally...

nicolas-grekas:
Missing LICENSE file, see #64 :)

.github/workflows/test.yaml
```diff
@@ -86,6 +93,12 @@ jobs:
                   composer config platform.php 7.4.99
                   composer update --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
+            - name: Turbo
+              run: |
+                  cd src/Turbo
+                  composer config platform.php 7.4.99
```
stof, line 99: do we still need this or has the support of PHP 8 been solved in the meantime ?

src/Turbo/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,50 @@
...
+                    ->info('Enable this section to use Turbo Streams')
+                    ->children()
+                        ->scalarNode('adapter')
+                            ->defaultNull()
+                            ->info('The service to use to create Turbo Streams')
+                        ->end()
+                        ->arrayNode('options')
+                            ->info('Key/value options passed to the adapter to create Turbo Streams')
+                            ->normalizeKeys(false)
+                            ->variablePrototype()
```
stof, line 41: this is missing a `->end()` to be consistent with the indentation, to reflect which node is ended each time.

src/Turbo/README.md
```diff
@@ -0,0 +1,267 @@
...
+    {{ form(message_form) }}
+
+    <!--
+    We can also display another link to go back to the list: the original HTML from frame
+    "message_1" in the list will be displayed again.
+    -->
+    <a href="/messages">Back</a>
+
+<!-- End the frame -->
+{{ turbo_frame_end() }}
```
stof, line 103: shouldn't this be a tag rather than a couple of `_start` and `_end` functions ?

tgalopin: AFAIK, tags do things and functions display things, thus this notation. I do agree tags would probably look nicer though. Not fixed about this.

richardhj: Tags would help in IDE autocompletion (same as for 'apply' and 'endapply'), so it's less error-prone I guess.

nicolas-grekas: Do we have any examples of tags that output something?
forms use functions for the same reason.

wouterj: One that is very close to this use-case is NelmioSecurityBundle's CSP tags:

```
{% cspscript %}
<script>
    window.api_key = '{{ api_key }}';
</script>
{% endcspscript %}
```
(the modify the `<script>` tag to add the correct CSP digest)

tgalopin: That's a different use case though, as it modifies the content of the tag (a bit like autoescape would). It doesn't display anything by itself.

I think I would be in favor of keeping in functions. It matches the way the rest of Symfony works (esp. forms) and it doesn't have that big of an impact on DX IMO. IDE can usually autocomplete functions as well as tags.

tgalopin:
Fixed reviews

nicolas-grekas (commented):
We might need a pack to install mercure-bundle + ux-turbo at the same time, with a working recipe

src/Turbo/Twig/TurboStreamExtension.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $options = array_merge($this->options, ['id' => $id]);
+
+        $html = '<div ';
+        $html .= $this->stimulus->renderStimulusController($env, [
+            $this->adapter->getStimulusControllerName() => $this->adapter->createDataValues($options),
+            'symfony--ux-turbo--core' => [],
+        ]).' ';
+
+        foreach ($attrs as $name => $value) {
```
nicolas-grekas, line 67: maybe we don't need attrs at all here btw?

src/Turbo/Twig/TurboStreamExtension.php
```diff
@@ -0,0 +1,80 @@
...
+        return [
+            new TwigFunction('turbo_stream_from', [$this, 'streamFrom'], ['needs_environment' => true, 'is_safe' => ['html']]),
+        ];
+    }
+
+    public function streamFrom(Environment $env, string $id, array $attrs = []): string
+    {
+        $options = array_merge($this->options, ['id' => $id]);
+
+        $html = '<div ';
```
nicolas-grekas, line 61: I'm wondering: should this generate an html tag, or a list of attributes?

src/Turbo/README.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+```twig
+{# chat.html.twig #}
+{% extends 'base.html.twig' %}
+
+{% block body %}
+    {# This line configures Turbo to listen to the "chat" topic on Mercure #}
+    {{ turbo_stream_from('chat') }}
```
sroze, line 178: Though I appreciate you technically want it to live within a `block` in here, semantically it would make greater sense to register it within your Twig file like that:

```twig
{% listen_turbostream "chat" %}
```

sroze (commented):
That looks fun!

src/Turbo/README.md
stof: Well, these 2 functions are about rendering some HTML opening and closing tags. they are indeed not a kind of twig block

fbourigault: What happens if I forget some `turbo_frame_end()` function calls?

stof: the HTML tag is not closed, and the browser may auto-close it in a different place than where you expect it to be closed.

beberlei:
I think this should be named `symfony/ux-turbo-streams`, because it has nothing to do with Turbo support in general, which works without this PR by just installing the npm/yarn package.

tgalopin:
Closing in favor of the other PR

PR #62, opened 2021-02-20
Add missing LICENSE files

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | -
| License       | MIT

### DIFF
--- a/src/LICENSE
+++ b/src/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2020-2021 Fabien Potencier
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


nicolas-grekas (approved):

OskarStark (approved):
Thanks

fabpot (approved):

fabpot:
Thank you @chalasr.

src/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2020-2021 Fabien Potencier
```
stof, line 1: This one should be at the root, not in `src/`

chalasr: thanks, fixed in 57e02174f961aa537b58428ad7edecf12d879a5e

PR #63, opened 2021-03-01
using stimulus_controller() everywhere

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | None
| License       | MIT

I think we should embrace `stimulus_controller()`. The controller name normalization is especially important.

Also, it looks like we should also update many of the options (e.g. `data-hd-url`) to values (in a BC way, obviously). That would make things even nicer with the `stimulus_controller()` function.

Cheers!

### DIFF
--- a/src/LazyImage/README.md
+++ b/src/LazyImage/README.md
@@ -34,7 +34,7 @@ page has been rendered:
 ```twig
 <img
     src="{{ asset('image/small.png') }}"
-    data-controller="symfony--ux-lazy-image--lazy-image"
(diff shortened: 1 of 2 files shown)


tgalopin (approved):

tgalopin:
We need to fix the CI in main it seems, probably due to a change in Github actions default image

chalasr:
The CI is green on the main branch (#63).

chalasr (approved):

dunglas (approved):
Tests should be green after a rebase.

PR #64, opened 2021-03-02
Symfony UX Turbo: content negotiation and Doctrine integration

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | n/a
| License       | MIT

This PR introduces an integration of [Hotwire Turbo](https://turbo.hotwire.dev/) with Symfony. It has been developed and tested over the past few months by @tgalopin, @nicolas-grekas, @chalasr, @weaverryan and me. It is complementary with #61 (#61 focuses on the integration with Twig while this one focuses on higher level features, #61 will have to be rebased on top of this one).

The best way to discover the features provided by this new component is to read the provided README file. To summarize, it provides:

* an integration with Mercure to stream page changes to all connected browsers from the controller. It's also possible to plug other transports.
* an integration with the Form component allowing to send changes to apply to the existing DOM in response to a POST
* a new PHP 8 attribute (#[Broadcast]): when a Doctrine entity is marked with this attribute, all changes made to it will be broadcasted in real time to all connected clients, and the DOM will instantly update consequently (a Twig template defined the operations to apply to the existing DOM tree). An extension point is available to plug other persistence systems 
* a sample app, including Panther tests (used as functional tests for the bundle itself)
* a template for GitHub Actions workflow

TODO:

* [x] Merge #61 into this one (can be done in another PR)


src/Turbo/.gitattributes
```diff
@@ -0,0 +1,10 @@
+/.github export-ignore
+/.php_cs.dist export-ignore
+/assets/test export-ignore
+/tests/ export-ignore
+/phpstan.neon export-ignore
+/phpunit.xml.dist export-ignore
+/.gitignore export-ignore
+/.gitattributes export-ignore
+src/Resources/assets/package-lock.json
+src/Resources/assets/node_modules/
```
jdreesen, line 10: What do these lines do?

dunglas: This file isn't needed at all actually. I remove it.

src/Turbo/Broadcaster/TwigMercureBroadcaster.php
```diff
@@ -0,0 +1,142 @@
...
+        if (!isset($options['template'])) {
+            $dir = $entityClass;
+            if (null !== $this->entityNamespace && 0 === strpos($entityClass, $this->entityNamespace)) {
+                $dir = substr($entityClass, \strlen($this->entityNamespace));
+            }
+
+            $options['template'] = sprintf('broadcast/%s.stream.html.twig', str_replace('\\', '/', $dir));
+        }
+
+        return $options;
```
soyuka, line 140:
```suggestion
        if (isset($options['template'])) {
            return $options;
        }
        
        $dir = $entityClass;
        if ($this->entityNamespace && 0 === strpos($entityClass, $this->entityNamespace)) {
            $dir = substr($entityClass, \strlen($this->entityNamespace));
        }

        $options['template'] = sprintf('broadcast/%s.stream.html.twig', str_replace('\\', '/', $dir));

        return $options;
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Symfony UX Turbo requires PHP 8+ and Symfony 4.4+.
+
+Install this bundle using Composer and Symfony Flex:
+
+```sh
+composer require symfony/ux-turbo
+
+# Don't forget to install the JavaScript dependencies as well and compile
+yarn install --force
```
soyuka, line 25: why force?

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+        $client->request('GET', '/');
+
+        $client->clickLink('This block is scoped, the rest of the page will not change if you click here!');
+        $this->assertSelectorTextContains('body', 'This will replace the content of the Turbo Frame!');
+    }
+}
+```
+
+Run `bin/phpunit` to execute the test! Symfony Panther automatically started your application with a web server
+and tested it using Google Chrome (Firefox is also supported)!
```
soyuka, line 138:
```suggestion
Run `bin/phpunit` to execute the test! Symfony Panther automatically starts your application with a web server
and test it using Google Chrome (Firefox is also supported)!
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+You can even let Panther open Chrome, and see what happens: `PANTHER_NO_HEADLESS=1 bin/phpunit --debug`
```
soyuka, line 140:
```suggestion
You can even watch changes happening in Chrome by using: `PANTHER_NO_HEADLESS=1 bin/phpunit --debug`
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+[Read the Turbo Frames documentation](https://turbo.hotwire.dev/handbook/frames) to learn everything you can do using Turbo Frames.
+
+### Coming Alive with Turbo Streams
+
+Turbo Streams are a way for the server to send partial page updates to clients.
+There are two main way to receive the updates:
+
+-   in response to a user action, for instance when the user submits a form
+-   asynchronously, by sending updates to the clients using [Mercure](https://mercure.rocks), [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API) and similar protocols
```
soyuka, line 150:
```suggestion
-   asynchronously, by sending updates to clients using [Mercure](https://mercure.rocks), [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API) and similar protocols
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Supported actions are `append`, `prepend`, `replace`, `update` and `remove`.
+[Read the Turbo Streams documentation for more details](https://turbo.hotwire.dev/handbook/streams).
+
+#### Sending Async Changes using Mercure: a Chat
+
+Symfony UX Turbo also supports the [Mercure](https://mercure.rocks) protocol. With this feature you can broadcast HTML
+updates to all currently connected clients.
+
+To illustrate this, let's build a chat system in **0 lines of JavaScript**!
```
soyuka, line 215:
```suggestion
To illustrate this, let's build a chat system with **0 lines of JavaScript**!
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Start by installing [the Mercure support](https://symfony.com/doc/current/mercure.html) on your project.
```
soyuka, line 217:
```suggestion
Start by adding [the Symfony's Mercure component](https://symfony.com/doc/current/mercure.html) on your project.
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+The easiest way to have a working development (and production-ready) environment is to use [Symfony Docker](https://github.com/dunglas/symfony-docker/),
+which comes with a Mercure hub integrated in the web server.
+
+If you use Symfony Flex, the configuration has been generated for you, be sure to update the `MERCURE_SUBSCRIBE_URL` in
+the `.env` file to point a Mercure Hub (it's not necessary if you are using Symfony Docker).
```
soyuka, line 222:
```suggestion
the `.env` file to point to a Mercure Hub (it's not necessary if you are using Symfony Docker).
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+```
+
+### Broadcast Options
+
+The `Broadcast` attribute comes with a set of handy options:
+
+-   `template` (`string`): Twig template to render (see above)
+-   `topics` (`string[]`): Mercure topics to use, defaults to an array containing the Fully Qualified Class Name
+-   `private` (`bool`): marks Mercure updates as private
+-   `id` (`string`): `id` field of the SSE
```
soyuka, line 428:
```suggestion
-   `id` (`string`): `id` field of the Server-Sent Event
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+-   `type` (`string`): `type` field of the SSE
```
soyuka, line 429:
```suggestion
-   `type` (`string`): `type` field of the Server-Sent Event
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+-   `retry` (`int`): `retry` field of the SSE
```
soyuka, line 430:
```suggestion
-   `retry` (`int`): `retry` field of the Server-Sent Event
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+This bundle aims at following the same Backward Compatibility promise as the Symfony framework:
+[https://symfony.com/doc/current/contributing/code/bc.html](https://symfony.com/doc/current/contributing/code/bc.html)
+
+However, it is currently considered
+[**experimental**](https://symfony.com/doc/current/contributing/code/experimental.html),
+meaning it is not bound to Symfony's BC policy for the moment.
+
+## Credits
+
+Created by [Kévin Dunglas](https://dunglas.fr). Sponsored by [Les-Tilleuls.coop](https://les-tilleuls.coop).
```
soyuka, line 484: Maybe credit all the names you mentioned in the PR:

> It has been developed and tested over the past few months by @tgalopin, @nicolas-grekas, @chalasr, @weaverryan and me

src/Turbo/Tests/TurboStreamTest.php
```diff
@@ -0,0 +1,33 @@
...
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ */
+final class TurboStreamTest extends PantherTestCase
+{
+    public function testStream(): void
+    {
+        ($client = self::createPantherClient())->request('GET', '/');
+
+        $client->submitForm('Trigger Turbo Stream!');
+        $client->waitForElementToContain('body', 'This div replace the existing element with the DOM ID "form".');
```
soyuka, line 28:
```suggestion
        $client->waitForElementToContain('body', 'This div replaces the existing element with the DOM ID "form".');
```

src/Turbo/Tests/TurboStreamTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->assertSelectorTextContains('body', 'This div replace the existing element with the DOM ID "form".');
```
soyuka, line 29:
```suggestion
        $this->assertSelectorTextContains('body', 'This div replaces the existing element with the DOM ID "form".');
```

src/Turbo/Tests/app/Entity/Book.php
```diff
@@ -0,0 +1,40 @@
...
+class Book
+{
+    /**
+     * @ORM\Column(type="integer")
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     *
+     * @var int|null
+     */
+    public $id;
```
soyuka, line 32: ?$id if nullable?

src/Turbo/Tests/app/webpack.config.js
```diff
@@ -0,0 +1,74 @@
...
+    // uncomment to get integrity="..." attributes on your script & link tags
+    // requires WebpackEncoreBundle 1.4 or higher
+    //.enableIntegrityHashes(Encore.isProduction())
+
+    // uncomment if you're having problems with a jQuery plugin
+    //.autoProvidejQuery()
+
+    // uncomment if you use API Platform Admin (composer require api-admin)
+    //.enableReactPreset()
+    //.addEntry('admin', './assets/admin.js')
```
soyuka, line 71: commented code really needed?

src/Turbo/README.md
dunglas: It's like that in all readme fils of this repo.

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
+# Symfony UX Turbo
+
+Symfony UX Turbo is a Symfony bundle integrating the [Hotwire Turbo](https://turbo.hotwire.dev)
+library in Symfony applications. It is part of [the Symfony UX initiative](https://symfony.com/ux).
+
+Symfony UX Turbo allow having the same user experience as with [Single Page Apps](https://en.wikipedia.org/wiki/Single-page_application)
```
jdreesen, line 6:
```suggestion
Symfony UX Turbo allows having the same user experience as with [Single Page Apps](https://en.wikipedia.org/wiki/Single-page_application)
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+and tested it using Google Chrome (Firefox is also supported)!
+
+You can even let Panther open Chrome, and see what happens: `PANTHER_NO_HEADLESS=1 bin/phpunit --debug`
+
+[Read the Turbo Frames documentation](https://turbo.hotwire.dev/handbook/frames) to learn everything you can do using Turbo Frames.
+
+### Coming Alive with Turbo Streams
+
+Turbo Streams are a way for the server to send partial page updates to clients.
+There are two main way to receive the updates:
```
jdreesen, line 147:
```suggestion
There are two main ways to receive the updates:
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+```
+
+Let's create our chat:
+
+```php
+// src/Controller/ChatController.php
+namespace App\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;use Symfony\Component\Form\Extension\Core\Type\TextType;use Symfony\Component\HttpFoundation\Request;
```
jdreesen, line 240:
```suggestion
use Symfony\Component\Form\Extension\Core\Type\SubmitType;
use Symfony\Component\Form\Extension\Core\Type\TextType;
use Symfony\Component\HttpFoundation\Request;
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+class ChatController extends AbstractController
+{
+    public function chat(Request $request, PublisherInterface $mercure): Response
+    {
+        $form = $this->createFormBuilder()
+            ->add('message', TextType::class, ['attr' => ['autocomplete' => 'off']])
+            ->add('send', SubmitType::class)
+            ->getForm();
+
+        $emptyForm = clone $form; // Used display an empty form after a post
```
jdreesen, line 253:
```suggestion
        $emptyForm = clone $form; // Used to display an empty form after a POST request
```

src/Turbo/README.md
```diff
@@ -0,0 +1,487 @@
...
+                $this->renderView('chat/message.stream.html.twig', ['message' => $data['message']])
+            ));
+
+            // Force an empty form to be rendered below
+            // It will replace the content of the Turbo Frame after a post
+            $form = $emptyForm;
+        }
+
+        return $this->render('chat/index.html.twig',
+            ['form' => $form->createView()
```
jdreesen, line 273:
```suggestion
        return $this->render('chat/index.html.twig', [
            'form' => $form->createView(),
```

src/Turbo/README.md
dunglas: Actually, we need the bundle.

src/Turbo/Tests/app/Entity/Book.php
dunglas: It's not doable because we still support old PHP versions.

tgalopin:
IMO the most interesting things in https://github.com/symfony/ux/pull/61 are:

* the usage of stimulus_controller in the Twig extension
* the system of adapters that would allow external packages to register new adapters easily (both on PHP and JS side)

If we can merge them into this PR, I'd be happy to close 61.

Also, I do think we should provide 7.2 support for broadcasting if we put it in the same package. I can contribute to this PR to add 7.2 support if you wish.

dunglas:
> * the usage of stimulus_controller in the Twig extension

+1 on my side

> * the system of adapters that would allow external packages to register new adapters easily (both on PHP and JS side)

+1 for the PHP side, but I don't think it is worth it on the JS side. Turbo already provides the needed abstraction layer. This will add complexity (and probably increase the size of the bundle) for no benefits. We could maybe just rename the controller `turbo-stream-mercure` or something like that, and provide other controllers if we add other transports at some point.

> Also, I do think we should provide 7.2 support for broadcasting if we put it in the same package. I can contribute to this PR to add 7.2 support if you wish.

To be honest, I don't think it is worth it. It's a new experimental package that relies on a library (Turbo) that is still in beta. It will mostly be used in new projects, and existing projects should upgrade to the latest PHP version before adding such features. By the time Turbo and this package will be marked stable, PHP 8 will probably be the most used PHP version, even for existing projects.
Also, supporting PHP 7 means adding a dependency to Doctrine Annotation. It will make the code more complex, the maintenance more complex, and adding new features more complex. Also, supporting only newest PHP versions for new features is an incentive to upgrade (so it benefits to the whole community).

tgalopin:
> +1 for the PHP side, but I don't think it is worth it on the JS side. Turbo already provides the needed abstraction layer. This will add complexity (and probably increase the size of the bundle) for no benefits. We could maybe just rename the controller turbo-stream-mercure or something like that, and provide other controllers if we add other transports at some point.

Have a look at how it works in my PR, it's super simple and doesn't involve any downside you are mentioning:

* https://github.com/symfony/ux/pull/61/files#diff-2d885da522db9b6768ae1f9cf220c9576b9c8a633d1ef57167704528fc39e282 is the main controller that is used to start Turbo
* https://github.com/symfony/ux/pull/61/files#diff-45a6ab35411514fe9ab7246ba09284113d1a97ac975b9960f8940b28e8d431f5 is the JS adapter that creates the EventSource object when it's enabled, which is then used by core-controller (and because controllers can be disabled, it's not present in the final bundle if disabled)
* then the PHP adapter is the one driving which JS controller should be used in addition to core-controller, with options that can come from the Symfony config: https://github.com/symfony/ux/pull/61/files#diff-1150dfb1355ae6892b68c25083c00a6ac3bae8d77315431752260301304fdd63

I really do think it's a model that works well, because it would allow for external packages to provide a plug and play system very easily. I don't think this PR currently supports adding another transport from an external package without complex DI manipulations, which is my point.

> +1 on my side

It's used there for instance: https://github.com/symfony/ux/pull/61/files#diff-bd05dc034c7cb413d1a5d1f8b458b1a56363410f09aefcd7da00376217968b5d

src/Turbo/DependencyInjection/TurboExtension.php
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        if (class_exists(TwigBundle::class)) {
+            $loader->load('twig.php');
+            if (isset($config['mercure']['subscribe_url'])) {
+                $container->getDefinition('turbo.twig.extension.stream')->addArgument($config['mercure']['subscribe_url']);
+            }
+
+            if (class_exists(MercureBundle::class)) {
+                $loader->load('broadcaster.php');
```
tgalopin, line 53: Shouldn't we check for Doctrine presence as well? I'm not sure broadcasting make sense without Doctrine?

src/Turbo/Doctrine/BroadcastListener.php
```diff
@@ -0,0 +1,117 @@
...
+ * @todo backport MongoDB support
+ *
+ * @experimental
+ */
+final class BroadcastListener implements ResetInterface
+{
+    private $broadcaster;
+
+    /**
+     * @var \SplObjectStorage<object, object>
```
tgalopin, line 36: Are these annotations necessary?

src/Turbo/Twig/FrameExtension.php
```diff
@@ -0,0 +1,52 @@
...
+use Twig\TwigFunction;
+
+/**
+ * Twig helpers to generated the Turbo Frame elements.
+ *
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ *
+ * @experimental
+ */
+final class FrameExtension extends AbstractExtension
```
tgalopin, line 24: After experimenting with this myself, I tend to think we don't actually need this extension. Using the HTML tags would allow developers to better understand the usage and I don't think Frames need Twig helpers. What do you think?

src/Turbo/Twig/StreamExtension.php
```diff
@@ -0,0 +1,94 @@
...
+    private $mercureHub;
+
+    public function __construct(?string $mercureHub = null)
+    {
+        $this->mercureHub = $mercureHub;
+    }
+
+    public function getFunctions(): iterable
+    {
+        yield new TwigFunction('turbo_stream_start', [$this, 'turboStreamStart'], ['is_safe' => ['html']]);
```
tgalopin, line 43: In the same way as Frame, I don't think we need to provide helpers to write the turbo stream tags. They are quite easy to use and it would help understand the way it work IMO. It would also help people use documentation from Turbo itself.

We do want to keep the ability to stream from though, so this Twig extension is definitely important.

src/Turbo/Twig/StreamExtension.php
```diff
@@ -0,0 +1,94 @@
...
+        yield new TwigFunction('turbo_stream_end', [$this, 'turboStreamEnd'], ['is_safe' => ['html']]);
+        yield new TwigFunction('turbo_stream_from_start', [$this, 'turboStreamFromStart'], ['is_safe' => ['html']]);
```
tgalopin, line 45: In my PR, I chose not to wrap content inside this div, because I think it creates the false impression that only the content inside the div will be affected by the stream, which isn't the case.

That's why I declare a single `turbo_stream_from` function that creates an empty div: https://github.com/tgalopin/ux/blob/31e064fb2ab6801fd6bd62b9a1505c6125fb67da/src/Turbo/README.md#3-create-real-time-applications-with-turbo-streams

WDYT?

dunglas:
My implementation is way simpler (it doesn't involve new events, two controllers etc). Turbo already provides the `connectStreamSource()` and `disconnectStreamSource()` functions to plug transports, we don't need an abstraction on top of this abstraction. The Twig extension could select the controller to use (by delegating to a PHP adapter, as you have done PHP-side) without having to introduce more JS code. This will support external packages too.

> It's used there for instance: https://github.com/symfony/ux/pull/61/files#diff-bd05dc034c7cb413d1a5d1f8b458b1a56363410f09aefcd7da00376217968b5d

Thanks, I'll try to use this!

src/Turbo/Broadcast.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class Broadcast
+{
+    public const ACTION_CREATE = 'create';
+    public const ACTION_UPDATE = 'update';
+    public const ACTION_REMOVE = 'remove';
+
+    /**
+     * @var mixed[]
```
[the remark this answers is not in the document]
dunglas: It's just to be explicit and in sync with the new `mixed` type introduced in PHP 8.

tgalopin:
> My implementation is way simpler (it doesn't involve new events, two controllers etc). Turbo already provides the connectStreamSource() and disconnectStreamSource() functions to plug transports, we don't need an abstraction on top of this abstraction. The Twig extension could select the controller to use (by delegating to a PHP adapter, as you have done PHP-side) without having to introduce more JS code. This will support external packages too.

Could you show me how an external package could register a new adapter with this PR? I may be missing something.

src/Turbo/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+final class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $treeBuilder = new TreeBuilder('turbo');
+        $rootNode = $treeBuilder->getRootNode();
+        $rootNode
+            ->children()
+                ->arrayNode('mercure')
```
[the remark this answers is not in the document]
dunglas: You're right, I didn't change this part since my initial implementation. We must merge what you've done in #61.

src/Turbo/DependencyInjection/TurboExtension.php
dunglas: It does. I adapted this code from API Platform, which already supports a lot of persistence systems. We could add more later (and users can add more in third party packages).

src/Turbo/Doctrine/BroadcastListener.php
dunglas: Yes, they are necessary for PHPStan.

src/Turbo/Twig/FrameExtension.php
dunglas: I think we should provide a way to hide all Turbo(.js) related things. Even if it should be possible to also use the lower level HTML elements. IMHO it improves the DX to not have to know anything about Turbo to use the basic features (but maybe I'm wrong, I've no strong opinion about this.)

dunglas:
> Could you show me how an external package could register a new adapter with this PR? I may be missing something.

It's not implemented yet! But it could just ship its own controller calling `Turbo.connectStreamSource()`. The Mercure implementation is basically 5 lines of JS, an extra abstraction layer doesn't worth it.

src/Turbo/DependencyInjection/TurboExtension.php
tgalopin: Ah, nice, thanks :) !

src/Turbo/Twig/FrameExtension.php
tgalopin: I feel like writing `turbo_frame_start` or `<turbo-frame ...>` is basically similar in terms of not seing JS, but `<turbo-frame>` is easier to parse visually. I don't have strong opinion either. 

We could add the extension in the future if we want to though (much easier to add than remove)

tgalopin:
Aah that's why I missed it then :) . I'm fine doing it otherwise, I don't care about the way we achieve it, but I do think it's important external packages can ship transports easily.

src/Turbo/Twig/FrameExtension.php
dunglas: Having the tag is also more consistent with the Form extension. IMHO we should keep it, and we can remove it before marking the stable release if people don't use it (as it's entirely optional).

dunglas: After having discussed with Titouan and Ryan, I removed this extension to encourage using the HTML elements directly, as it's the new best practice we try to promote across the framework.

seb-jean:
Hi,
Will it support [Lazily Loading Frames](https://turbo.hotwire.dev/handbook/frames#lazily-loading-frames)  ?
If so, how should this be done?
Thanks :)

dunglas:
It does, see the readme! We could maybe improve that by using the fragment subsystem to generate URLs automatically, but it an be done in a follow up PR.

src/Turbo/Broadcaster/TwigMercureBroadcaster.php
```diff
@@ -0,0 +1,151 @@
...
+            $options = $this->expressionLanguage->evaluate($options[0], ['entity' => $entity, 'action' => $action]);
+        }
+
+        $entityClass = \get_class($entity);
+
+        if ($extraKeys = array_diff(array_keys($options), self::OPTIONS)) {
+            throw new \InvalidArgumentException(sprintf('Unknown broadcast options "%s" on class "%s". Valid options are: "%s"', implode('", "', $extraKeys), $entityClass, implode('", "', self::OPTIONS)));
+        }
+
+        $options['topics'] = (array) ($options['topics'] ?? sprintf(self::TOPIC_PATTERN, rawurlencode($entityClass), rawurlencode($this->propertyAccessor->getValue($entity, 'id'))));
```
chalasr, line 137: Should we get the identifier property from the entity class metadata?

soyuka: WDYM? I guess this could be improved in API Platform but not in symfony. What tells you this class is tight to doctrine?

chalasr: I mean that this makes the assumption that the entity class has a property named `id`, which might not be true. 

> What tells you this class is tight to doctrine?

Nothing, indeed. Should the `id` property be configurable then?

soyuka: :+1:

dunglas: It's a rabbit hole. In API Platform we started with this exact same convention, and later we added an option to make the ID configurable, then we added support for composed identifiers, complex identifiers (objects such as dates) etc...
To be honest, I regret having added that complexity to API Platform.

My opinion is that to use this feature, you **must** have a property named `id` or a getter named `getId()` (we could even introduce an interface for that). It's easy, and doable by the user even for the most complex case (even if you use another property as identifier, have complex identifiers etc, you can always add a `getId()` method.

By the way, this mandatory convention is documented in the readme.

WDYT?

chalasr: Works for me, we can revisit later on if needed.

src/Turbo/README.md
```diff
@@ -0,0 +1,627 @@
...
+```yaml
+# config/packages/mercure.yaml
+mercure:
+  hubs:
+    hub1:
+      url: https://hub1.example.net/.well-known/mercure
+      token: snip
+    hub2:
+      url: https://hub2.example.net/.well-known/mercure
+      token: snip
```
azjezz, line 500:
```suggestion
mercure:
  hubs:
    hub1:
      url: https://hub1.example.net/.well-known/mercure
      jwt: snip
    hub2:
      url: https://hub2.example.net/.well-known/mercure
      jwt: snip
```

src/Turbo/README.md
```diff
@@ -0,0 +1,627 @@
...
+        $emptyForm = clone $form; // Used to display an empty form after a POST request
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $data = $form->getData();
+
+            // 🔥 The magic happens here! 🔥
+            // The HTML update is pushed to the client using Mercure
+            // The Mercure topic (here "chat") MUST be the ID of the block that will display the received updates
+            $mercure(new Update(
```
azjezz, line 264:
```suggestion
            $mercure->publish(new Update(
```

.github/workflows/test-turbo.yml
```diff
@@ -0,0 +1,189 @@
+name: Symfony UX Turbo
```
tgalopin, line 1: Is it necessary to have a new CI file?

src/Turbo/README.md
```diff
@@ -0,0 +1,626 @@
...
+    }
+}
+```
+
+```php
+<?php
+
+namespace App\Turbo;
+
+use Symfony\UX\Turbo\Twig\TurboStreamListenRendererInterface;use Symfony\WebpackEncoreBundle\Twig\StimulusTwigExtension;use Twig\Environment;
```
tgalopin, line 585: Return lines issue

src/Turbo/composer.json
```diff
@@ -0,0 +1,82 @@
+{
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/azjezz/mercure-bundle"
```
tgalopin, line 5: We should not forget to remove this once https://github.com/symfony/mercure-bundle/pull/42 is merged

src/Turbo/README.md
```diff
@@ -0,0 +1,626 @@
...
+This bundle aims at following the same Backward Compatibility promise as the Symfony framework:
+[https://symfony.com/doc/current/contributing/code/bc.html](https://symfony.com/doc/current/contributing/code/bc.html)
+
+However, it is currently considered
+[**experimental**](https://symfony.com/doc/current/contributing/code/experimental.html),
+meaning it is not bound to Symfony's BC policy for the moment.
+
+## Credits
+
+Created by [Kévin Dunglas](https://dunglas.fr). Sponsored by [Les-Tilleuls.coop](https://les-tilleuls.coop).
```
nicolas-grekas, line 623: Can you please remove this line? Symfony doesn't put such credits anywhere, it's a community project :)

src/Turbo/Mercure/TurboStreamListenRenderer.php
```diff
@@ -0,0 +1,47 @@
...
+            $topic = sprintf(Broadcaster::TOPIC_PATTERN, rawurlencode(\get_class($topic)), rawurlencode($this->propertyAccessor->getValue($topic, 'id')));
+        } elseif (class_exists($topic)) {
+            // Generate a URI template to subscribe to updates for all objects of this class
+            $topic = sprintf(Broadcaster::TOPIC_PATTERN, rawurlencode($topic), '{id}');
+        }
+
+        return $this->stimulusTwigExtension->renderStimulusController(
+            $env,
+            'symfony/ux-turbo/turbo-stream-mercure',
+            ['topic' => $topic, 'hub' => $this->hub->getUrl()] // TODO: use the subscribe URL when it will be available
```
tgalopin, line 44: What's this TODO? I'm not sure to understand it

src/Turbo/README.md
```diff
@@ -0,0 +1,626 @@
...
+    public static function getDefaultIndexName(): string
+    {
+        // The transport name, used as key in the service locator
+        // Alternatively, configure it using the other supported methods: https://symfony.com/doc/current/service_container/service_subscribers_locators.html#indexing-the-collection-of-services
+        return 'my-transport';
+    }
+}
+```
+
+Because these classes implement the `BroadcasterInterface` and `TurboStreamListenRendererInterface` interfaces,
```
tgalopin, line 609: I think we should explain to use the `turbo.renderer.stream_listen` and `turbo.broadcaster` tags (and *then* say that if your services are autoconfigured, this is done automatically).

src/Turbo/Resources/config/doctrine_orm.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Symfony\UX\Turbo\Broadcaster\BroadcasterInterface;
+use Symfony\UX\Turbo\Doctrine\BroadcastListener;
+
+/*
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ */
+return static function (ContainerConfigurator $container): void {
+    $container
```
tgalopin, line 21: While I do like PHP config, I'm not sure why it's better than configuring it in the extension here, given the small size of concerned files. That would decrease the number of files to load.

WDYT?

tgalopin (commented):
I quite like the look of it

Here are a few comments

src/Turbo/Mercure/TurboStreamListenRenderer.php
azjezz: fixed in https://github.com/symfony/mercure/pull/39#event-4465467412, should be replaced by `$this->hub->getPublicUrl()` now.

azjezz, line 44:
```suggestion
            ['topic' => $topic, 'hub' => $this->hub->getPublicUrl()]
```

.github/workflows/test-turbo.yml
dunglas: It allows to not set up a useless Mercure service for all other bundles (https://github.com/symfony/ux/pull/64/files/e8b335bd5fb7d50b209fa2e832f22fd846c16bbe#diff-b65cafb402a3698791b8c52f1f17bd1887642e197a7d453c6c3379910f06bd3fR51), but I can merge both files if you prefer.

src/Turbo/README.md
dunglas: done

src/Turbo/Resources/config/doctrine_orm.php
dunglas: I've no strong opinion about this. WDYT @nicolas-grekas?

src/Turbo/Broadcaster/ImuxBroadcaster.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
```
nicolas-grekas, line 3: :ghost:

src/Turbo/Doctrine/BroadcastListener.php
```diff
@@ -0,0 +1,117 @@
...
+use Symfony\UX\Turbo\Broadcaster\BroadcasterInterface;
+
+/**
+ * Detects changes made from Doctrine entities and broadcasts updates to the Mercure hub.
+ *
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ *
+ * @see https://github.com/api-platform/core/blob/master/src/Bridge/Doctrine/EventListener/PublishMercureUpdatesListener.php Adapted from API Platform.
+ *
+ * @todo backport MongoDB support
```
nicolas-grekas, line 27: what does this mean? when will it be done?

src/Turbo/Resources/config/doctrine_orm.php
nicolas-grekas: I agree, we can remove the file. Or we might prefer doing the reverse as we do in Symfony:
have one config file and `removeDefinition` in the extension. This helps discoverability. (would apply to the other config files of course)

src/Turbo/Stream/AddTurboStreamFormatSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+ * Detects if it's a Turbo Stream request and set the format accordingly.
+ *
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ *
+ * @experimental
+ */
+final class AddTurboStreamFormatSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @see https://github.com/hotwired/turbo/issues/24 Explanation of why this hack is necessary
```
nicolas-grekas, line 29: Hasn't this been fixed?

src/Turbo/Tests/BroadcastTest.php
```diff
@@ -0,0 +1,46 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Kévin Dunglas <kevin@dunglas.fr>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
```
nicolas-grekas, line 12: please remove them all from the PR :)

src/Turbo/Mercure/Broadcaster.php
```diff
@@ -0,0 +1,160 @@
...
+    ) {
+        if (80000 > \PHP_VERSION_ID) {
+            throw new \LogicException('The broadcast feature requires PHP 8.0 or greater, you must either upgrade to PHP 8 or disable it.');
+        }
+
+        $this->name = $name;
+        $this->twig = $twig;
+        $this->propertyAccessor = $propertyAccessor ?? PropertyAccess::createPropertyAccessor();
+        $this->publisher = $publisher;
+        $this->expressionLanguage = $expressionLanguage ?? new ExpressionLanguage();
```
nicolas-grekas, line 86: will fail if EL is not installed

src/Turbo/Mercure/TurboStreamListenRenderer.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
```
nicolas-grekas, line 3: to be replaced by the copyright

src/Turbo/Doctrine/BroadcastListener.php
dunglas: The API Platform code I adapted here also supports Doctrine MongoDB ODM. We could add support for Mongo later.

src/Turbo/composer.json
```diff
@@ -0,0 +1,82 @@
...
+    },
+    "require-dev": {
+        "doctrine/doctrine-bundle": "^2.2",
+        "doctrine/orm": "^2.8",
+        "phpstan/phpstan": "^0.12",
+        "symfony/debug-bundle": "^5.2",
+        "symfony/expression-language": "^5.2",
+        "symfony/form": "^5.2",
+        "symfony/framework-bundle": "^5.2",
+        "symfony/mercure-bundle": "dev-feature/token-factory",
```
azjezz, line 72:
```suggestion
        "symfony/mercure-bundle": "^0.3",
```

( https://github.com/symfony/mercure-bundle/pull/42/commits/5261c424d626e142def2271fa3179e6d6eb9f43f )

src/Turbo/Stream/AddTurboStreamFormatSubscriber.php
beberlei: Yes, but this code also uses the new media type that was added, so the comment needs updating.

in addition this is probably only necessary, because Symfony Framework does not have `$request->setFormat(TurboStreamResponse::STREAM_FORMAT, TurboStreamResponse::STREAM_MEDIA_TYPE);` in core yet.

src/Turbo/Doctrine/BroadcastListener.php
```diff
@@ -0,0 +1,117 @@
...
+    {
+        $this->createdEntities = new \SplObjectStorage();
+        $this->updatedEntities = new \SplObjectStorage();
+        $this->removedEntities = new \SplObjectStorage();
+    }
+
+    private function storeEntitiesToPublish(object $entity, string $property): void
+    {
+        if ((new \ReflectionClass($entity))->getAttributes(Broadcast::class)) {
+            $this->{$property}->attach('removedEntities' === $property ? clone $entity : $entity);
```
beberlei, line 114: Clone can be pretty expensive for an entity and cause weird behavior. I see in `book.stream.html.twig` this is used to keep the entity id since its necessary to build the correct `target`. I wish there was a better solution to this.

src/Turbo/Mercure/Broadcaster.php
```diff
@@ -0,0 +1,160 @@
...
+        if (isset($options['template'])) {
+            return $options;
+        }
+
+        $dir = $entityClass;
+        if ($this->entityNamespace && 0 === strpos($entityClass, $this->entityNamespace)) {
+            $dir = substr($entityClass, \strlen($this->entityNamespace));
+        }
+
+        $options['template'] = sprintf('broadcast/%s.stream.html.twig', str_replace('\\', '/', $dir));
```
beberlei, line 156: Does this approach work with third party broadcasts?

beberlei:
Again i would name this `UX Turbo Streams` instead of just `UX Turbo`.

dunglas:
The name of the equivalent Rails library is `turbo-rails`. Also, I plan to add an improved support for lazy Turbo Frames.

tgalopin:
> Again i would name this UX Turbo Streams instead of just UX Turbo.

This package provides an integration with Turbo, both regarding the behavior itself (Streams but also media type, lazy turbo frames, ...) *and* by the fact itself that it's plug and play: when installing ux-turbo, Turbo Drive, Frames and Streams are enabled automatically because of the Flex/Stimulus system that autoload controllers from the vendor directory. It does bring Turbo itself, not only Turbo Streams.

Thus I prefer UX Turbo as a name.

src/Turbo/Mercure/Broadcaster.php
dunglas: Yes you can plug your own transport (it's documented at the end of the readme file).

src/Turbo/Stream/AddTurboStreamFormatSubscriber.php
dunglas: Yes indeed, but I think that we should wait for Turbo to be marked stable before adding this media type to `symfony/symfony` (the media type already changed).

src/Turbo/Resources/config/doctrine_orm.php
dunglas: @nicolas-grekas it's already done as described in your second solution. So I keep it as is?

src/Turbo/Doctrine/BroadcastListener.php
dunglas: As other parts of the bundle require to have a `$id` property or a `getId()` method, maybe should we use this convention here too and just store the id?

.github/workflows/test-turbo.yml
```diff
@@ -0,0 +1,189 @@
...
+              working-directory: src/Turbo
+              run: composer install --prefer-dist
+
+            - name: Install PHPUnit dependencies
+              working-directory: src/Turbo
+              run: vendor/bin/simple-phpunit --version
+
+            - name: PHPStan
+              working-directory: src/Turbo
+              run: vendor/bin/phpstan analyse --no-progress
```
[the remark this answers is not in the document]
azjezz: 👍 for psalm, also with https://github.com/psalm/psalm-plugin-symfony ( if the plugin is failing to detect type for something that is symfony related, let me know and i will try to get it fixed asap )

src/Turbo/Doctrine/BroadcastListener.php
dunglas: I just looked at this, unfortunately this will hurt the user experience. The current implementation allows accessing to the whole removed entity in the template. Storing only the ID will prevent to do this. I suggest keeping this as-is for now, until we find a better solution.

nicolas-grekas:
I submitted my review at https://github.com/dunglas/ux/pull/1, please have a look.

.github/workflows/test-turbo.yml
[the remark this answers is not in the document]
azjezz: Let's stick with phpstan for now, after merging i can migrate this ( and other package ) to psalm to be consistent with symfony framework ( people contributing would get used to one tool instead of two )

src/Turbo/Broadcaster/IdAccessor.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @return string[]
+     */
+    public function getEntityId(object $entity): ?array
+    {
+        $entityClass = \get_class($entity);
+
+        if ($this->doctrine && $em = $this->doctrine->getManagerForClass($entityClass)) {
+            return $em->getClassMetadata($entityClass)->getIdentifierValues($entity);
+        }
```
chalasr, line 38: nice :)

chalasr (approved):
Looks solid!

src/Turbo/Doctrine/BroadcastListener.php
chalasr: Fixed with the new `IdAccessor`.

nicolas-grekas:
Continued in #71

PR #70, opened 2021-03-23
Propose a way for user to use symfony cache with

cedriclombardot:
To not regenerate each time blurhash

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT

Just for helping devs implementing cache with symfony/cache

### DIFF
--- a/src/LazyImage/README.md
+++ b/src/LazyImage/README.md
@@ -139,3 +139,64 @@ php vendor/bin/phpunit
 cd Resources/assets
 yarn test
 ```
+
+## Want to implement a cache ?
+
+Use symfony/cache
+
+```
+composer require symfony/cache
+```
+
+And create your twig extension :
+
+```php
+<?php
+
+namespace App\Twig;
+
+use Symfony\Component\String\Slugger\AsciiSlugger;
+use Symfony\Contracts\Cache\CacheInterface;
+use Symfony\Contracts\Cache\ItemInterface;
+use Symfony\UX\LazyImage\BlurHash\BlurHashInterface;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+
+class BlurHashExtension extends AbstractExtension
+{
+    public function __construct(private BlurHashInterface $blurHash, private CacheInterface $cache)
+    {}
+
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('data_uri_thumbnail_cached', [$this, 'createDataUriThumbnailCached']),
+        ];
+    }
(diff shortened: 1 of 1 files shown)


src/LazyImage/README.md
```diff
@@ -139,3 +139,70 @@ php vendor/bin/phpunit
...
+        ];
+    }
+
+    public function createDataUriThumbnailCached(string $filename, int $width, int $height, int $encodingWidth = 75, int $encodingHeight = 75): string
+    {
+        $slugger = new AsciiSlugger();
+        $cacheKey = $slugger->slug(implode('-', func_get_args()));
+
+        $value = $this->cache->get($cacheKey, function (ItemInterface $item) use ($filename, $width, $height, $encodingWidth, $encodingHeight) {
+            if (!$item->isHit()) {
```
nicolas-grekas, line 184: there is no need for this check inside a cache callback

src/LazyImage/README.md
```diff
@@ -139,3 +139,70 @@ php vendor/bin/phpunit
...
+                $item->set(
```
nicolas-grekas, line 185: neither this is needed - just return the value, the pool will set the item :)

cedriclombardot: Oh thanks for the review, i missread the doc :)

src/LazyImage/README.md
```diff
@@ -139,3 +139,64 @@ php vendor/bin/phpunit
...
+use Symfony\Contracts\Cache\ItemInterface;
+use Symfony\UX\LazyImage\BlurHash\BlurHashInterface;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+
+class BlurHashExtension extends AbstractExtension
+{
+    public function __construct(private BlurHashInterface $blurHash, private CacheInterface $cache)
+    {}
```
Kocal, line 169: This is PHP 8 code and I don't think we want documentation/examples to be written with.
I think it would be better to write code compatible with the lowest PHP requirement which is [7.2.5](https://github.com/symfony/ux/blob/main/src/LazyImage/composer.json#L29).

WDYT?

weaverryan:
I like this PR... but can't we go further? It feels like the blurhash should *always* cache. Or, at the very worst, you just add one line of configuration (e.g. to point it at a cache pool, perhaps) and caching is all setup.

Kocal:
Hi, I'm sorry but I've just seen this PR this exists, and BlurHash caching has been implemented in #1755.
I'm closing the PR, but thanks for the contribution!

PR #78, opened 2021-04-04
Fix UX Turbo usage without Streams

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | -
| Tickets       | -
| License       | MIT

Currently, UX Turbo doesn't work without the Mercure bridge installed and configured as it requires JS to load it. This adds some JS in the UX Turbo package to automatically load Turbo when installed using an empty Stimulus controller.

### DIFF
--- a/src/Turbo/Resources/assets/src/turbo_controller.js
+++ b/src/Turbo/Resources/assets/src/turbo_controller.js
@@ -1,0 +1,21 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import { Controller } from 'stimulus';
+import * as Turbo from '@hotwired/turbo';
+
+// Expose Turbo to the rest of the app to allow for dynamic Turbo calls
+window.Turbo = Turbo;
+
+/**
+ * Empty Stimulus controller only used for Symfony Flex wiring.
+ *
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+export default class extends Controller {}
--- a/src/Turbo/Resources/assets/.babelrc
+++ b/src/Turbo/Resources/assets/.babelrc
@@ -1,0 +1,4 @@
+{
+    "presets": ["@babel/env"],
+    "plugins": ["@babel/plugin-proposal-class-properties"]
+}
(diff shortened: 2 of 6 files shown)


dunglas (approved):

jrushlow (approved):

src/Turbo/Resources/assets/src/turbo_controller.js
```diff
@@ -0,0 +1,21 @@
...
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import { Controller } from 'stimulus';
+import * as Turbo from '@hotwired/turbo';
```
weaverryan, line 11: Does importing it actually *enable* TurboDrive? Or does that require more work?

src/Turbo/Resources/assets/src/turbo_controller.js
```diff
@@ -0,0 +1,21 @@
...
+
+// Expose Turbo to the rest of the app to allow for dynamic Turbo calls
+window.Turbo = Turbo;
```
weaverryan, line 14: Is exposing it as a global variable the correct/normal way to do things? Where is it needed globally?

weaverryan (commented):
Just a few questions - an update to the docs might help illuminate your thinking about how this will be used :)

src/Turbo/Resources/assets/src/turbo_controller.js
tgalopin: Yes it does (tested locally :) ).

src/Turbo/Resources/assets/src/turbo_controller.js
tgalopin: I wasn't sure about this one, but it has two benefits I can see: it helps usage from outside Webpack (ie in a legacy JS static file for instance) and it ensures Webpack subtree split won't remove the dependency at compilation.

I'm not sure it's necessary, it's debatable.

nicolas-grekas: I'd be in favor or removing it then, if this plays well with subtree splitting from webpack. How can we check this?

dunglas: We should keep this as it's used by native adapters.

nicolas-grekas (approved):

PR #80, opened 2021-04-05
[WIP] Migrate to the new bundle directory structure

sadikoff:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | none
| License       | MIT

Hello everyone!

I'm happy to propose this refactoring according to the latest bundle [best practices](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure). I think it is a good time to do it because there are not so many packages and we can standardize this little piece of Symfony packages.

Cheers!


sadikoff:
I have one minor question about `assets/` directory it looks like it's not quite correct to rename it fully to `public/` or it IS? From my point of view this `assets/` directory in UX packages has nothing common with `public/` bundle directory or I'm missing something?

weaverryan:
> I have one minor question about assets/ directory it looks like it's not quite correct to rename it fully to public/ or it IS? From my point of view this assets/ directory in UX packages has nothing common with public/ bundle directory or I'm missing something?

That's correct - it shouldn't go into public. And actually, I think the `assets/` directory needs to NOT be moved from `Resources/assets` to `assets`. The reason is that part of Symfony Flex looks for `Resources/assets`. It *also* looks for `assets/` - https://github.com/symfony/flex/blob/e472606b4b3173564f0edbca8f5d32b52fc4f2c9/src/PackageJsonSynchronizer.php#L153-L162

The problem is that, whatever the path, once a package is installed, the user's package.json is updated to point to the directory - e.g. `"@symfony/ux-chartjs": "file:vendor/symfony/ux-chartjs/Resources/assets",`. And so, moving the directory in an existing project would be a BC break :)

sadikoff:
Yeah BC break is not good... However, because UX has an experimental status and to achieve more clean code built with the latest technologies we can allow this little BC here, also because that `package.json` file is owned by the user and can be easily edited. WDYT?

nicolas-grekas:
Im also reluctant to doing this change, because it diverges from what we do historically in other core Symfony bundles. Since this repo might end up being merged with Symfony/symfony, we should apply its conventions here.

sadikoff:
@nicolas-grekas so why not change these "historical" moments and force everything to the new way, as it was done for Symfony itself?

nicolas-grekas:
I'm mixed here. I also prefer the new structure, but Ryan's point is very valid, and I'm not sure the reasons to do this change are strong enough to outweigh the drawbacks and diverge from the current way in symfony/symfony...

sadikoff:
Honestly, I'd like to push this structure to core Symfony bundles too 👼🏻 , and probably here we can start 😄

PR #90, opened 2021-05-03
UX FormCollection

### LINKED ISSUE #3: Proposal: component to handle form collections
It would be nice if there was some available JS in the ux components to add basic stuff to handle form collections (add & remove) :)

stakovicz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #3 
| License       | MIT

Handle form collections, add & remove buttons


src/FormCollection/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
```
weaverryan, line 2: We *should* have this file, but it should match the contents from the other repos :) https://github.com/symfony/ux/blob/main/src/Dropzone/.gitattributes

src/FormCollection/DependencyInjection/FormCollectionExtension.php
```diff
@@ -0,0 +1,37 @@
...
+use Symfony\UX\FormCollection\Form\CollectionType;
+
+/**
+ * @internal
+ */
+class FormCollectionExtension extends Extension implements PrependExtensionInterface
+{
+    public function prepend(ContainerBuilder $container)
+    {
+    }
```
weaverryan, line 27: it looks like this is not needed? And also the `PrependExtensionInterface` is not needed?

src/FormCollection/Form/CollectionType.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return BaseCollectionType::class;
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'button_add' => [
+                'text' => 'Add',
+                'attr' => ['class' => 'btn btn-outline-primary'],
```
weaverryan, line 36: I'm not sure if we should default any `attr`

src/FormCollection/README.md
```diff
@@ -0,0 +1,141 @@
...
+// ...
+use Symfony\UX\FormCollection\Form\CollectionType;
+
+class BlogFormType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            // ...
+            ->add('comments', CollectionType::class, [
```
weaverryan, line 51: Maybe the class should be called `UXCollectionType` - that will avoid auto-complete confusion with the original.

src/FormCollection/README.md
```diff
@@ -0,0 +1,141 @@
...
+                    'attr' => ['class' => 'btn btn-outline-secondary']    
+                ],
+            ])
+            // ...
+        ;
+    }
+
+    // ...
+}
+```
```
weaverryan, line 72: You should also show how to render it (I know, it's just "rendering the field", but people will still wonder)

src/FormCollection/README.md
```diff
@@ -0,0 +1,141 @@
...
+You need to select the right theme from the one you are using :
+```yaml
+# config/packages/twig.yaml
+twig:
+  # For bootstrap for example
+  form_themes: ['@FormCollection/form_theme_div.html.twig']
+```
+You have 2 different themes :
+- `@FormCollection/form_theme_div.html.twig`
+- `@FormCollection/form_theme_table.html.twig`
```
weaverryan, line 32: This is the biggest question make for me with this very nice PR. If we provide a built-in form theme to "rendering everything for them", then we will also need a Bootstrap 4/5 theme... and a tailwind theme. And, if the user needs to customize how things look, then they need to override a fairly complex form theme that we've created here.

I'm not *definitely* against this. However, I'd like to see if we can document how this new feature could be used if there *were* no form themes included. For example,

> Now that your `BlogFormType` form set up and with a `comments` field that is a `CollectionType`, you can
> render it in your template:

```twig
{% macro commentFormRow(commentForm) %}
    <div
        class="col-4"
        data-symfony--ux-form-collection--collection-target="entry"
    >
        {{ form_errors(commentForm) }}
        {{ form_row(commentForm.content) }}
        {{ form_row(commentForm.otherField) }}

        <a data-action="symfony--ux-form-collection--collection#delete">
            Remove
        </a>
    </div>
{% endblock %}

<div
    class="row"
    {{ stimulus_controller('symfony/ux-form-collection/collection', {
        prototype: _self.commentFormRow(form.comments.vars.prototype),
    }) }}
>
    {% for commentForm in form.comments %}
        {{ _self.commentFormRow(commentForm) }}
    {% endfor %}

    <a data-action="symfony--ux-form-collection--collection#add">
        Add Another
    </a>
</div>
```

This would give me full control over everything, including the buttons. The Stimulus controller would just handle making it all work :). The worst pieces are the super-ugly `data-action` and `data-...-target` attributes. But this would look much nicer if we merged https://github.com/symfony/webpack-encore-bundle/pull/124

There may be some other missing pieces... but in theory, this should be enough to work. The "delete" button should not need a `data-index-entry` attribute, as we can look to its ancestors for the "closest" entry target to know which item is being removed.

So, this is my thinking :). No form theme, but we make it as dead-simple as possible for users to render things themselves.

weaverryan (commented):
I'm super excited that someone has taken this in - it will be a huge win! I have one major comment that discusses direction on this feature. Btw, don't hesitate to ping me in Symfony's slack if you'd like to discuss in real time :)

stakovicz:
Thanks @weaverryan, for taking the time to the review.
I'm intimidated by you because I know you through SymfonyCast ;-)

This is my first contribution, but you are all very kind to help me!

src/FormCollection/DependencyInjection/FormCollectionExtension.php
stakovicz: Yes, it's cleaned up!

src/FormCollection/.gitattributes
stakovicz: Yes, it's fixed

src/FormCollection/README.md
stakovicz: My idea was to make a component very easy to use.
It's true that template overloading can be complicated due to attributes related to Stimulus.
With these 2 templates (`@FormCollection/form_theme_div.html.twig`, `@FormCollection/form_theme_table.html.twig`) we cover a maximum of use cases.

For your proposal, there should be explicit errors when the controller is wrong or the action is not correctly entered.

I think you need to know Stimulus to be able to set up HTML according to the documentation.

weaverryan: To make a better judgement, I'll need some time to play with this in a real project :). Would you be willing/able to push a demo app to GitHub that uses this (with some quick setup instructions)?

weaverryan:
> Thanks @weaverryan, for taking the time to the review.
> I'm intimidated by you because I know you through SymfonyCast ;-)

Well, I'm happy to know you know too ;) 

> This is my first contribution, but you are all very kind to help me!

It's a very big feature for a first contribution - nice work!

src/FormCollection/README.md
stakovicz: Here you have my repo for my test :
https://github.com/stakovicz/ux-collection-test/
Enjoy !

tgalopin:
Hi @stakovicz !

Thanks very much for your PR, as Ryan I'm quite excited by this as it's a major requirement for Symfony Forms.

I think, as Ryan mentioned, the only point where I think we need to spend some time thinking is how templates are handled (ie. what HTML is generated). Generally speaking, Symfony as a framework has helpers that generate several HTML tags (like in forms) but usually these helpers become more complex to work with than the original HTML. This was the reasoning behind https://symfony.com/blog/new-in-symfony-5-2-form-field-helpers for instance: use inline attribute helpers instead of tags helpers.

Perhaps your solution is the best, I don't know yet as I didn't have a deep look into it, but it' worth spending some time: it will have a huge impact on developer experience.

I'll have a deeper look asap :) .

stakovicz (commented):
Hi @tgalopin!

Thank you for your look.

Indeed, I am divided.
On the one hand, the solution that I propose allows an almost instantaneous use.
On the other hand, your solution is very flexible and allows maximum customization.

You know the community better than I do.
You also have a lot more experience.
I trust you in the final choice and you will follow.

src/FormCollection/README.md
```diff
@@ -0,0 +1,153 @@
...
+                    'text' => 'Add',    
+                    // Add HTML classes to the add button
+                    'class' => 'btn btn-outline-primary'
+                ],
+                'button_delete' => [
+                    // Default text for the delete button
+                    'text' => 'Remove',    
+                    // Add HTML classes to the add button
+                    'class' => 'btn btn-outline-secondary'
+                ],
```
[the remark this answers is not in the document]
maxhelias: I'm not sure about me but have you tried with 2 ButtonType in the Form when the options "allow_add" & "allow_delete" are true?
This would allow to modify their appearance via the twig blocks

src/FormCollection/README.md
stakovicz: I've made an update.
Now you can use a predefined theme or not.
You have the choice.

I need help to do more tests.

src/FormCollection/Resources/views/form_theme_table.html.twig
```diff
@@ -0,0 +1,52 @@
+{%- block button_add -%}
+    {%- set attrDataAction = (attr['data-action']|default('') ~ ' ' ~ controllerName ~ '#add')|trim -%}
+    <tr>
+        <td colspan="2">
+            <button data-action="{{ attrDataAction }}"
+                    class="{{ button_add.class|default('') }}" type="button">{{ button_add.text|default('Add')|trans }}</button>
```
jmsche, line 6: Same comments as previous form theme

src/FormCollection/Resources/views/form_theme_table.html.twig
```diff
@@ -0,0 +1,52 @@
...
+        </td>
+    </tr>
+{%- endblock button_add -%}
+
+{%- block button_delete -%}
+    {%- set attrDataAction = (attr['data-action']|default('') ~ ' ' ~ controllerName ~ '#delete')|trim -%}
+    <button data-action="{{ attrDataAction }}"
+            class="{{ button_delete.class|default('') }}" type="button">
```
jmsche, line 14: Same comments as previous form theme

src/FormCollection/Resources/views/form_theme_table.html.twig
```diff
@@ -0,0 +1,52 @@
...
+    {% if prototype is defined and not prototype.rendered %}
+        {%- set prototype_attr = prototype.vars.attr|merge(attrDataTarget) -%}
+        {%- set attr = attr|merge({'data-prototype': form_row(prototype, {'row_attr': prototype_attr}) }) -%}
+    {% endif %}
+
+    <div data-controller="{{ dataController }}"
+         data-{{ controllerName }}-allow-add-value="{{ allow_add|json_encode }}"
+         data-{{ controllerName }}-allow-delete-value="{{ allow_delete|json_encode }}"
+         data-{{ controllerName }}-button-add-value="{{ block('button_add')|e }}"
+         data-{{ controllerName }}-button-delete-value="{{ block('button_delete')|e }}"
```
jmsche, line 34:
```suggestion
         data-{{ controllerName }}-button-add-value="{{ block('button_add')|e('html_attr') }}"
         data-{{ controllerName }}-button-delete-value="{{ block('button_delete')|e('html_attr') }}"
```

jmsche (commented):
Hi,

As my knowledge about JavaScript is pretty much not existent, I reviewed the PR but only for PHP, Twig & docs :)

My main goals with this review are (as explained in the review):

* Putting the default translations in the Form Type rather than in the templates
* Having 4 distinct options with allowed types, instead of two arrays with no types at all
* Docs: form theme instead of macros

Thanks for your PR! :)

src/FormCollection/Resources/views/form_theme_div.html.twig
```diff
@@ -0,0 +1,48 @@
+{%- block button_add -%}
+    {%- set attrDataAction = (attr['data-action']|default('') ~ ' ' ~ controllerName ~ '#add')|trim -%}
+    <button data-action="{{ attrDataAction }}"
+            class="{{ button_add.class|default('') }}"  type="button">{{ button_add.text|default('Add')|trans }}</button>
+{%- endblock button_add -%}
+
+{%- block button_delete -%}
+    {%- set attrDataAction = (attr['data-action']|default('') ~ ' ' ~ controllerName ~ '#delete')|trim -%}
+    <button data-action="{{ attrDataAction }}"
+            class="{{ button_delete.class|default('') }}" type="button">
```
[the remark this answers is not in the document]
stakovicz: Even if the `data-action` is not defined in FormType ?

src/FormCollection/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+[Check the Symfony doc](https://symfony.com/doc/4.4/form/form_themes.html) for the different ways to set themes in Symfony.
+
+## Use manual theming
+
+> Consider your `BlogFormType` form set up and with a comments field that is a `CollectionType`, you can
+render it in your template:
+
+```twig
+{% macro commentFormRow(commentForm) %}
```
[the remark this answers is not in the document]
stakovicz: It's @weaverryan's idea. It is easier and more flexible to implement.

src/FormCollection/README.md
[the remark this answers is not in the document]
stakovicz: @maxhelias allow_add is a boolean type from the original [FormType](https://symfony.com/doc/current/reference/forms/types/collection.html#allow-add).

src/FormCollection/Resources/views/form_theme_table.html.twig
stakovicz: OK with the changes of the default options.

src/FormCollection/Resources/views/form_theme_div.html.twig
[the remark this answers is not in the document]
stakovicz: I have done the test and it's required because the `attr`.`data-action` is not mandatory.

src/FormCollection/Form/UXCollectionType.php
```diff
@@ -0,0 +1,58 @@
...
+        $resolver->setDefaults([
+            'button_add' => [
+                'text' => '',
+                'class' => '',
+            ],
+            'button_delete' => [
+                'text' => '',
+                'class' => '',
+            ],
+        ]);
```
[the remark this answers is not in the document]
stakovicz: You are right.

I just made the modification [here](https://github.com/symfony/ux/pull/90/commits/7f5cdb1342f61b581b0b7ec87210674c9869f2b4).

src/FormCollection/Resources/views/form_theme_div.html.twig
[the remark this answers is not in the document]
jmsche: I'm sorry, I selected only line 10 IIRC :) It was only about the `class`, not `data-action` :)

src/FormCollection/README.md
```diff
@@ -0,0 +1,191 @@
...
+    <div
+        class="col-4"
+        data-symfony--ux-form-collection--collection-target="entry"
+    >
+        {{ form_errors(commentForm) }}
+        {{ form_row(commentForm.content) }}
+        {{ form_row(commentForm.otherField) }}
+
+        <button type="button" data-action="symfony--ux-form-collection--collection#delete">
+            Remove
```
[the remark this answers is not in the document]
jmsche:
```suggestion
            {{ button_delete_text|trans }}
```

src/FormCollection/README.md
```diff
@@ -0,0 +1,191 @@
...
+    class="row"
+    data-controller="symfony--ux-form-collection--collection"
+    data-prototype="{{ _self.commentFormRow(form.comments.vars.prototype)|e }}"
+>
+    {% for commentForm in form.comments %}
+        {{ _self.commentFormRow(commentForm) }}
+    {% endfor %}
+
+    <button type="button" data-action="symfony--ux-form-collection--collection#add">
+        Add Another
```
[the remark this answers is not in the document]
jmsche:
```suggestion
        {{ button_add_text|trans }}
```

src/FormCollection/README.md
jmsche: Hi @weaverryan,

Any reason to use a macro instead of a proper form theme here?

The macro forces you to use it everywhere it's needed (eg. prototype & entries as in your example) while a "real" form theme would apply automatically to both.

zmitic:
@stakovicz 
First I want to thank you for doing this. I have projects with deeply nested collections, this addition is more than welcome,

---

I have one question about the problem I encountered few times and suggested jquery.collection plugin suffered from it; I had to make my own implementation because of that.

TL;DR
Will it be possible to affect how the index is increased [in this line](https://github.com/symfony/ux/pull/90/files#diff-59a86209d87245dd7c3870ac0cd9ab54637d09ac5800199fea024a4c0dbbd67fR66)?

Or better: a way to set base index not as the number of existing, but for example seconds since Unix epoch?

<details>
  <summary markdown="span">Reason, click to expand</summary>

Imagine **editing** (create is not a problem) a form with a collection that has 3 Users.

So Symfony renders them with keys 0, 1, 2.

Now you delete all of them and add **new User**. Submitted array to the server will **only** have key **3**; so far it is good, Symfony knows you deleted first 3 and added new one.

Here comes the problem. Validation fails and the form is rendered again, with that one user (the new one). You add 2 new users but instead of creating keys 4 and 5, the suggested Symfony plugin will reset keys.

The problem is that backend will think that data is updated (because submitted keys are the same as when form was first created), and not know that old users were deleted and new ones were created. 

Keys are very important here; default mapper is sort-of doing some bypass by searching for objects (mostly entities) but I have more extreme cases.

I solved it by making the base index not to be the count of existing elements, but the seconds of the current time. Submitted data has big keys but no problem since that.
</details>

Any thoughts?

stakovicz:
@zmitic 

> First I want to thank you for doing this. I have projects with deeply nested collections, this addition is more than welcome,

Thanks a lot !
> ...
> Any thoughts?

OK I see the problem. I watch what I can do.

stakovicz:
@zmitic this commit allow you to set the default index  : 
https://github.com/symfony/ux/pull/90/commits/8b23e78ac07447a0109eef0fe5abbfe7991ea7ef

The value is set automatically if you use a predefined theme.

zmitic:
@stakovicz Perfect! I will find a way to override it, still didn't figure that for Stimulus.

Thank you.

stakovicz:
@weaverryan @tgalopin do you have any feedback ?

jorismak:
To be honest, I tried making - ahem, for my self - some re-usable javascript code to handle Collections.
But every time I try to use it, my use case for a Collection is slightly different, and I end up customizing a lot anyway.

The HTML changes per project (sometimes only widgets in a table, sometimes a 'normal' bootstrap horizontal form, then again something different, then the keys are numeric and have no meaning, then the keys are identifiers which need to be precise, etc...).

Now aboard the Stimulus hype-train, I have a simple 'collection_controller' that handles the add-button and delete-button code. With a delete-button with a `data-action=` attribute and a simple `data-remove-key=` attribute, I have enough that all my everytime-it-is-slightly-different templates can work.

I make a macro that handles a single child of the collection. That macro is then also usable to render a prototype string (simply calling the macro in an html-attribute will escape it properly) and for me it's easier to write a small template for the collection-at-hand then try to make a 'one size fits all template'.

So personally, I wouldn't bother with trying to come up with some template-style that fits all... because it never will in my experience.

Include a 'simple' piece of form theme that uses existing form_row() and form_label() so that it is as generic as possible, but make sure the stimulus controller doesn't depend on it hard. And include a little bit of 'demo' template for people whishing to do it their way.

This way Symfony has - ahem, finally - working out-of-the-box javascript for Collections :), but it's still easy enough for people who know how to customize a collection to use a partial-twig as form-theme and just render it there how they see fit.


In the end I sort of 'gave up'

ker0x (commented):
Thanks @stakovicz for this work. I have taken inspiration from what you did to build my own custom collection type with some improvements. 

I also made some feedback :wink: 

For those who are interested, here is a gist of my current implementation: https://gist.github.com/ker0x/8a1bad79a22f0ebae9ddf28c264dd13f

src/FormCollection/Form/UXCollectionType.php
```diff
@@ -0,0 +1,61 @@
...
+    public function getParent()
+    {
+        return CollectionType::class;
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'button_add_text' => 'Add',
+            'button_add_class' => '',
```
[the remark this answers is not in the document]
zmitic: Can you add [support for arrays](https://www.php.net/manual/en/function.is-countable.php) as well here: https://gist.github.com/ker0x/8a1bad79a22f0ebae9ddf28c264dd13f#file-enhancedcollectiontype-php-L42-L44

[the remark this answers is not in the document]
ker0x: Indeed it's better, gist updated! Thanks ;)

seb-jean:
I am interested :)

seb-jean:
Hi,

With example below, `data-ux-form-collection-prototype-value` is badly encoded because there is html_attr on value : https://github.com/symfony/webpack-encore-bundle/blob/main/src/Twig/StimulusTwigExtension.php#L69

```twig
{% macro commentFormRow(commentForm) %}
    <div
        class="col-4"
        data-symfony--ux-form-collection--collection-target="entry"
    >
        {{ form_errors(commentForm) }}
        {{ form_row(commentForm.content) }}
        {{ form_row(commentForm.otherField) }}

        <a data-action="symfony--ux-form-collection--collection#delete">
            Remove
        </a>
    </div>
{% endblock %}

<div
    class="row"
    {{ stimulus_controller('symfony/ux-form-collection/collection', {
        prototype: _self.commentFormRow(form.comments.vars.prototype),
    }) }}
>
    {% for commentForm in form.comments %}
        {{ _self.commentFormRow(commentForm) }}
    {% endfor %}

    <a data-action="symfony--ux-form-collection--collection#add">
        Add Another
    </a>
</div>
```

tgalopin:
Hello @stakovicz 

Sorry for the delayed answer, I didn't have much time to work on UX lately. I plan to spend more time on it in the coming weeks.

As Ryan explained, I think we need to find a way to tackle the templating, especially for custom needs. I'll have a look at your example repository and see if there are interesting ideas we could have.

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            // ...
+            ->add('comments', UXCollectionType::class, [
+                // ...
+                // Default text for the add button (used by predefined theme)
+                'button_add_text' => 'Add',
+                // Add HTML classes to the add button (used by predefined theme)
+                'button_add_class' => 'btn btn-outline-primary',
```
alexander-schranz: ~~Would more go with `button_add_attr` I think no other form types yet have option to configure `class` explicit instead they always had an option to render `_attr`.~~

As shown beneath I would go with `button_add_options` which contains all button options also the `label` and so on which is also rendered as a prototype.

src/FormCollection/Resources/views/form_theme_div.html.twig
```diff
@@ -0,0 +1,52 @@
+{%- block button_add -%}
+    {%- set attrDataAction = (attr['data-action']|default('') ~ ' ' ~ controllerName ~ '#add')|trim -%}
+    <button data-action="{{ attrDataAction }}"
```
alexander-schranz, line 3: Would be better if we can use the exist `button_widget` block:

https://github.com/symfony/symfony/blob/41697e78a158156952f8c445512f11bd58f90958/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L224

Maybe we can use in the Form Builder the `ButtonType` and then give it also a special `block_prefix` so we can go with `collection_button_add_widget` to overwrite it. 

So we could go with:

```php
'button_add_options => [
    'label' => 'app.button_add',
    'label_translation_parameters' => [],
    'attr' => ['class' => '...']
],
```

This make it possible to use any of the exist button options and would be easier for theming also for future themes.

src/FormCollection/Resources/views/form_theme_table.html.twig
```diff
@@ -0,0 +1,56 @@
...
+        {%- set attr = attr|merge({'data-prototype': form_row(prototype, {'row_attr': prototype_attr}) }) -%}
+    {% endif %}
+
+    {% set indexKeys = data|keys %}
+    {% set startIndex = indexKeys|length == 0 ? 0 : max(indexKeys) %}
+
+    <div data-controller="{{ dataController }}"
+         data-{{ controllerName }}-allow-add-value="{{ allow_add|json_encode }}"
+         data-{{ controllerName }}-allow-delete-value="{{ allow_delete|json_encode }}"
+         data-{{ controllerName }}-button-add-value="{{ block('button_add')|e('html_attr') }}"
```
alexander-schranz, line 35: Think we should go here with `form_widget(button_add_prototype)`.

We can create the button add prototype this way:

```php
// build form
$prototype = $builder->create('button_add', ButtonType, array_merge(['block_prefix' => 'collection_add_button'], $options['button_add_options']));
$builder->setAttribute('button_add_prototype', $prototype->getForm());

// build view
if ($form->getConfig()->hasAttribute('button_add_prototype')) {
    $prototype = $form->getConfig()->getAttribute('button_add_prototype');
    $view->vars['button_add_prototype'] = $prototype->setParent($form)->createView($view);
}
```

This is similar how the prototype of the collection itself is created:

https://github.com/symfony/symfony/blob/41697e78a158156952f8c445512f11bd58f90958/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L39-L40

https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L64-L67

This way we make sure the button using the form theme and we can use `button_widget` or `collection_add_button_widget` to overwrite it the `Theme` way of symfony forms.

src/FormCollection/Resources/views/form_theme_table.html.twig
```diff
@@ -0,0 +1,56 @@
...
+    {% if prototype is defined and not prototype.rendered %}
+        {%- set prototype_attr = prototype.vars.row_attr|merge(attrDataTarget) -%}
+        {%- set attr = attr|merge({'data-prototype': form_row(prototype, {'row_attr': prototype_attr}) }) -%}
+    {% endif %}
+
+    {% set indexKeys = data|keys %}
+    {% set startIndex = indexKeys|length == 0 ? 0 : max(indexKeys) %}
+
+    <div data-controller="{{ dataController }}"
+         data-{{ controllerName }}-allow-add-value="{{ allow_add|json_encode }}"
```
alexander-schranz, line 33: I find `-allow-add-value` and `button-add-value` confusing. I think I would just go with `button-add-prototype`. If it is set there is add if not there is no add else it could be confusing that I can set the prototype but its not used.

alexander-schranz (commented):
Would really like to see something like this in the symfony/ux component. Would just look that add and remove buttons are rendered using symfony form theme. I added a few comments about it.

/cc @weaverryan @tgalopin what do you think about them?

src/FormCollection/Resources/views/form_theme_table.html.twig
alexander-schranz: What I'm personally missing is a `-button-remove-target-selector`. So I can adjust were the ux javascript will render target the button.

mario-fehr:
@stakovicz any progress on this to finish it? Would love to use this.

weaverryan:
I like @alexander-schranz's ideas, and I think if we can implement those, I would support this. @stakovicz you've done a lot of great work on this, but we've been slow to review. Are you still able/interested to finish this?

Thx :)

stakovicz:
Yes, I'm in !
I think can do the last changes.

src/FormCollection/Resources/views/form_theme_div.html.twig
stakovicz: I made the changes based on your feedback.
Nexrt, the modifications on the buttons.

src/FormCollection/README.md
stakovicz: I made the changes based on your feedback.
Next, the modifications on the buttons.

src/FormCollection/composer.json
```diff
@@ -0,0 +1,52 @@
...
+        "symfony/config": "^4.4.17|^5.0",
+        "symfony/dependency-injection": "^4.4.17|^5.0",
+        "symfony/form": "^4.4.17|^5.0",
+        "symfony/http-kernel": "^4.4.17|^5.0"
+    },
+    "require-dev": {
+        "symfony/framework-bundle": "^4.4.17|^5.0",
+        "symfony/phpunit-bridge": "^5.2",
+        "symfony/twig-bundle": "^4.4.17|^5.0",
+        "symfony/var-dumper": "^4.4.17|^5.0"
```
weaverryan, line 40: These should all be updated to `^5.4|^6.0`

src/FormCollection/composer.json
```diff
@@ -0,0 +1,52 @@
...
+    "autoload": {
+        "psr-4": {
+            "Symfony\\UX\\FormCollection\\": ""
+        },
+        "exclude-from-classmap": [
+            "/Tests/"
+        ]
+    },
+    "require": {
+        "php": ">=7.2.5",
```
weaverryan, line 30: Change to `"php": ">=8.0",` and then we can update code accordingly (e.g. using property promotion, etc). We could even bump to `8.1` if necessary, but it looks like you got things working on 7.2, so 8.0 should, hopefully, be easy enough :).

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
+# UX Form Collection
```
weaverryan, line 1: We're now doing this as an `index.rst` file. So, it just needs to move, then convert to RST

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Symfony UX Form collection is a Symfony bundle providing light UX for collection
```
weaverryan, line 3:
```suggestion
Symfony UX Form collection is a Symfony bundle providing JavaScript-powered "add" and "remove"
buttons to your `CollectionType` fields.
```

It'd be GREAT to have a little screenshot :)

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+yarn install --force
+yarn encore dev
+```
+
+Also make sure you have at least version 2.0 of [@symfony/stimulus-bridge](https://github.com/symfony/stimulus-bridge)
+in your `package.json` file.
+
+## Use predefined theme
+
+You need to select the right theme from the one you are using:
```
weaverryan, line 25:
```suggestion
First, choose a form theme to use for styling your collection type:
```

src/FormCollection/Resources/assets/src/controller.ts
```diff
@@ -0,0 +1,156 @@
...
+    _textToNode(text: string) {
+        const template = document.createElement('template');
+        text = text.trim(); // Never return a text node of whitespace as the result
+        template.innerHTML = text;
+
+        return template.content.firstChild;
+    }
+
+    _dispatchEvent(name: string, payload: any) {
+        this.element.dispatchEvent(new CustomEvent(name, { detail: payload }));
```
weaverryan, line 154: Also make it bubble, like this: https://github.com/symfony/ux/blob/2.x/src/Typed/Resources/assets/src/controller.ts#L65

src/FormCollection/Resources/views/form_theme_div.html.twig
```diff
@@ -0,0 +1,51 @@
...
+            class="{{ button_add_options.class }}"  type="button">{{ button_add_options.label|trans }}</button>
+{%- endblock button_add -%}
+
+{%- block button_delete -%}
+    {%- set attrDataAction = (attr['data-action']|default('') ~ ' ' ~ controllerName ~ '#delete')|trim -%}
+    <button data-action="{{ attrDataAction }}"
+            class="{{ button_delete_options.class }}" type="button">{{ button_delete_options.label|trans }}</button>
+{%- endblock button_delete -%}
+
+{% block ux_collection_widget -%}
```
weaverryan, line 13: I'd love to NOT need this block. Instead, inside of `UxCollectionType`, in `finishView()`, we should (I *think*) be able to add all of the attributes we need. For example, adding a `data-controller` attribute should be easy enough... as well as all of the Stimulus "values".

Even the `attrDataTarget`, I believe, could be added inside `finishView`. There, you're passed a `FormView` object, you should be able to iterate over `$formView->children` to then add the `attrDataTarget` to each (and you would do the same with `$formView->vars['prototype']`.

I'm pushing for this because, if we accomplish this, then if the user has already customized their `ux_collection_widget` block in a form theme, then we are not overriding their work. Also, it would allow me to, for example, do something like this:

```twig
<div
    {{ stimulus_controller('symfony/ux-form-collection/collection') }}
    data-prototype="{{ form_row(form.files.vars.prototype)|e('html_attr') }}"
>
    Files:
    <button
        class="btn btn-sm btn-secondary"
        {{ stimulus_action('symfony/ux-form-collection/collection', 'add') }}
    >+</button>
    {% for fileForm in form.files %}
        {{ form_row(fileForm) }}
    {% endfor %}
</div>
```

Currently that doesn't work because I have *avoided* rendering the `ux_collection_widget`, and so the children (and prototype) are all missing their targets.

EDIT: Hmm, I see that you will still need this block so that you can use the `{{ block() }}` function to set some of the variables. But, Ideally we would JUST set some variables into `attr` then call the parent `collection_widget` block. So this block would only be setting variables, then calling that block.

src/FormCollection/Form/UXCollectionType.php
```diff
@@ -0,0 +1,77 @@
...
+            $value['class'] = $value['class'] ?? $defaultButtonAddOptions['class'];
+
+            return $value;
+        });
+        $resolver->setNormalizer('button_delete_options', function (Options $options, $value) use ($defaultButtonDeleteOptions) {
+            $value['label'] = $value['label'] ?? $defaultButtonDeleteOptions['label'];
+            $value['class'] = $value['class'] ?? $defaultButtonDeleteOptions['class'];
+
+            return $value;
+        });
```
weaverryan, line 61: I think we should set `'by_reference' => true` to override the parent `CollectionType`. Forgetting this is a big "WTF" moment. And so, since we're introducing this field new, let's use this "better" default setting.

src/FormCollection/Form/UXCollectionType.php
```diff
@@ -0,0 +1,77 @@
...
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $defaultButtonAddOptions = [
+            'label' => 'Add',
+            'class' => '',
+        ];
+        $defaultButtonDeleteOptions = [
+            'label' => 'Remove',
+            'class' => '',
+        ];
```
weaverryan, line 41: I'd also like a `label_html` option which is set to a boolean (defaults to false). If true, we would render the label "not escaped" (i.e. raw). Basically, we're trying to accomplish this: https://symfony.com/blog/new-in-symfony-5-1-form-improvements#allow-html-contents-in-form-labels

This would allow, for example, FontAwesome icons to be used with the buttons.

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+```yaml
+# config/packages/twig.yaml
+twig:
+    # For bootstrap for example
+    form_themes: ['@FormCollection/form_theme_div.html.twig']
```
weaverryan, line 31: I think we should only have one form theme (`form_theme_div.html.twig`). If we DID have a 2nd form theme, I would make it a bootstrap theme so that the user could instantly get something nice-looking inside of Bootstrap... but I'm not sure that's necessary because it's really easy to add `btn btn-primary` classes to the buttons from within the form.

Anyways, if the user needs to heavily customize their buttons, let's let them create their own form theme to override our's. Also, I think we should prevent the user from needing to do this step at all. We can automatically add this form theme for them: https://github.com/symfony/ux/blob/09c9effcc2582d4de0907c9e12c35afa313433ae/src/Dropzone/DependencyInjection/DropzoneExtension.php#L27-L37 (don't forget the `PrependExtensionInterface`.

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+There are 2 predefined themes available:
+
+-   `@FormCollection/form_theme_div.html.twig`
+-   `@FormCollection/form_theme_table.html.twig`
+
+[Check the Symfony doc](https://symfony.com/doc/4.4/form/form_themes.html) for the different ways to set themes in Symfony.
+
+## Use a custom form theme
```
weaverryan, line 41: Let's move this section down, below `Usage`. And we can call it: `Customizing how the Field Renders`.

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+```twig
+{# edit.html.twig #}
+
+{{ form_start(form) }}
+    {# ... #}
+    {{ form_row(comments) }}
+    {# ... #}
+{{ form_end(form) }}
+```
```
weaverryan, line 121: Somewhere, we need to mention that, in order for this to work, they likely need `orphanRemoval` and `cascade=['persist']` on their entities. This is what my field looks like:

```php
#[ORM\OneToMany(mappedBy: 'directory', targetEntity: File::class, orphanRemoval: true, cascade: ['persist'])]
private $files;
```

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+// ...
+use Symfony\UX\FormCollection\Form\UXCollectionType;
+
+class BlogFormType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            // ...
+            ->add('comments', UXCollectionType::class, [
```
weaverryan, line 92: To be as helpful as possible, I think we need to show the user a simple `CommentType` class. And then, use that here with `'entry_type' => `CommentType::class`.

Overall, this is how my example looks locally - I think these options are helpful to all show!

```php
'allow_add' => true,
'allow_delete' => true,
'entry_type' => FileType::class,
// ideally UXCollectionType can set this for us
'by_reference' => false,
'entry_options' => [
    // hides the "0", "1", "2" labels on the embedded forms
    'label' => false,
],
'button_add_options' => [
    'label' => '<span class="fa fa-plus"></span>',
    'label_html' => true,
    'class' => 'btn btn-secondary btn-sm'
],
'button_delete_options' => [
    'label' => '<span class="fa fa-times"></span>',
    'label_html' => true,
    'class' => 'btn btn-danger btn-sm'
],
```

src/FormCollection/Resources/views/form_theme_div.html.twig
weaverryan: Yea, this is interesting. To accomplish this, I think it would be very similar to how `prototype` is done, which is sort of a "fake field" that is added to the form... without being a real field.

You can see how the `prototype` is created in `buildForm()`... but not actually added to the field: https://github.com/symfony/symfony/blob/a05702965e9fc0de7ce77d1641e772e2fbfe9911/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L39-L40

Later, it's added as a view variable: https://github.com/symfony/symfony/blob/a05702965e9fc0de7ce77d1641e772e2fbfe9911/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L91-L101

It would be interesting to do the same thing with both buttons. We would pass `button_add_options` directly to the  `$builder->create()` call - e.g. how `prototype_options` are passed here: https://github.com/symfony/symfony/blob/a05702965e9fc0de7ce77d1641e772e2fbfe9911/src/Symfony/Component/Form/Extension/Core/Type/CollectionType.php#L33

src/FormCollection/README.md
```diff
@@ -0,0 +1,190 @@
...
+## Use a custom form theme
+
+Consider your `BlogFormType` form set up and with a comments field that is a `CollectionType`, you can
+render it in your template:
+
+```twig
+{% macro commentFormRow(commentForm) %}
+    <div
+        class="col-4"
+        data-symfony--ux-form-collection--collection-target="entry"
```
weaverryan, line 50: We should show `{{ stimulus_controller() }}` here, and `{{ stimulus_action() }}` and `{{ stimulus_target() }} in the other places.

weaverryan (commented):
Hey @stakovicz!

I just had an opportunity to put this to the test on a local project - I really enjoyed it! I like how you have built both an automatic feature (just render the "form row" and everything is added automatically) but also an "escape hatch" where you can render all of the part of the collection field manually, but still take advantage of the Stimulus controller.

I've left some feedback: I think we just need to smooth out some technical details and this will be ready. I'm really excited for it :).

Cheers!

src/FormCollection/Resources/views/form_theme_div.html.twig
```diff
@@ -0,0 +1,51 @@
...
+         data-{{ controllerName }}-allow-delete-value="{{ allow_delete|json_encode }}"
+         data-{{ controllerName }}-button-add-value="{{ block('button_add')|e('html_attr') }}"
+         data-{{ controllerName }}-button-delete-value="{{ block('button_delete')|e('html_attr') }}"
+         data-{{ controllerName }}-prototype-name-value="{{ prototype_name }}"
+         data-{{ controllerName }}-start-index-value="{{ startIndex }}"
+            {{ block('widget_container_attributes') }}
+    >
+        {%- if form is rootform -%}
+            {{ form_errors(form) }}
+        {%- endif -%}
```
weaverryan, line 40: What's this for?

src/FormCollection/README.md
stakovicz: I don't see any RST files in the other parts 🤨. They are all in .md

jmsche: Your branch may be a little out-of-date, see eg. https://github.com/symfony/ux/tree/2.x/src/Chartjs/Resources/doc

stakovicz: Thanks, I had not looked in this file 😬

src/FormCollection/Resources/assets/src/controller.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        this._dispatchEvent('form-collection:pre-connect', {
+            allowAdd: this.allowAddValue,
+            allowDelete: this.allowDeleteValue,
+        });
+
+        if (this.allowAddValue) {
+            // Add button Add
+            const buttonAdd = this._textToNode(this.buttonAddValue);
+            this.element.prepend(buttonAdd);
```
weaverryan, line 51: I think this should actually be added to the *bottom*. In the future, we may make it configurable, bottom seems like a better default to me.

weaverryan:
Friendly ping - let us know if and how we can help, or if anything is unclear :)

stakovicz:
Good evening Ryan,

Thanks for the reminder, everything is clear.
However, at the moment, I can't find time for this project.

If you wish, you can continue.

Best regards,

weaverryan:
Totally understandable :). If anyone else wants to continue this PR, please feel free to take this code, build on top of it, and submit a new PR. I *would* like it to be finished - it's a very nice idea.

alexander-schranz:
I'm currently trying based on this implement and the feedback trying to implemented a new variation of the UXCollectionType which does not requiring even `.html.twig` to solve also the comment about the overwrite the `collection_widget` block.

I'm nearly there to remove all `.html.twig` and set all required attributes and even the buttons inside the form types. I sadly has one thing I could not yet solve and that is I need to wrap the `entry_type` to add the remove button there, for this I'm doing the following in the UXCollectionType:

```php
$entryTypeNormalizer = function (OptionsResolver $options, $value) {
   return UXCollectionEntryType::class; // overwrite the given entry_point with wrapped component
};

$entryOptionsNormalizer = function (OptionsResolver $options, $value) {
    return [
        'entry_type' => $options['entry_type'], // TODO get the original entry_type here
        'entry_options' => $value, // forwards options
    ];
};

$resolver->addNormalizer('entry_type', $entryTypeNormalizer);
$resolver->addNormalizer('entry_options', $entryOptionsNormalizer);
```

The problem is I'm loosing the original given `'entry_type'` option this way. Which I need to forward to my UXCollectionEntryType. Sure we could go with `ux_entry_type` as option but that would in my case be a bad developer experience. So is there any way we get the original value here or somebody have another idea here?

PR #93, opened 2021-05-06
Moving Turbo to be a peer dependency

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | None
| License       | MIT

Like with Stimulus itself, I believe this should be a peer dependency. The only downside is that users need to install it directly, which I don't see as a huge downside.

This should guarantee that the the same "turbo" module is used from both the symfony/ux-turbo package and the end-users code (as well as from any other libraries that could potentially list turbo as a peer dependency). It should also fix this annoying little line I get in PhpStorm ;) 

Screen Shot 2021-05-06 at 11 31 18 AM

Cheers!

### DIFF
--- a/src/Turbo/README.md
+++ b/src/Turbo/README.md
@@ -21,9 +21,6 @@ Install this bundle using Composer and Symfony Flex:
 ```sh
 composer require symfony/ux-turbo
 
-# install turbo itself
-yarn add @hotwired/turbo --dev
-
 # Don't forget to install the JavaScript dependencies as well and compile
 yarn install --force
 yarn encore dev
--- a/src/Turbo/README.md
+++ b/src/Turbo/README.md
@@ -21,6 +21,9 @@ Install this bundle using Composer and Symfony Flex:
 ```sh
 composer require symfony/ux-turbo
 
+# install turbo itself
+yarn add @hotwired/turbo --dev
+
 # Don't forget to install the JavaScript dependencies as well and compile
 yarn install --force
 yarn encore dev
--- a/src/Turbo/Resources/assets/package.json
+++ b/src/Turbo/Resources/assets/package.json
@@ -19,10 +19,8 @@
         "test": "babel src -d dist && jest",
         "lint": "eslint src test"
     },
-    "dependencies": {
-        "@hotwired/turbo": "^7.0.0-beta.4"
-    },
     "peerDependencies": {
+        "@hotwired/turbo": "^7.0.0-beta.5",
         "stimulus": "^2.0.0"
     },
     "devDependencies": {
@@ -30,6 +28,7 @@
         "@babel/core": "^7.12.3",
         "@babel/plugin-proposal-class-properties": "^7.12.1",
         "@babel/preset-env": "^7.12.7",
+        "@hotwired/turbo": "^7.0.0-beta.5",
         "@symfony/stimulus-testing": "^1.1.0",
         "stimulus": "^2.0.0"
     },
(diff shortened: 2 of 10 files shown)


src/Turbo/README.md
```diff
@@ -21,6 +21,9 @@ Install this bundle using Composer and Symfony Flex:
 ```sh
 composer require symfony/ux-turbo
 
+# install turbo itself
+yarn add @hotwired/turbo --dev
```
nicolas-grekas, line 25: This looks like something for flex!

weaverryan: Oh... dang, you're right! That's an interesting idea!

weaverryan: For now, I *do* think we should merge this (as a peer dependency is more correct). But then we can add a new mechanism to Flex that would make this line redundant.

dunglas:
Isn't the whole point of Symfony UX to not have to install the packages directly? I mean, all other packages in this repo have the same "problem", including Swup which is very similar to Turbo: https://github.com/symfony/ux/blob/main/src/Swup/Resources/assets/package.json#L26

 For the record, the Mercure adapter already list Turbo as a peer dependencies, which should limit the problem: https://github.com/symfony/ux/blob/main/src/Turbo/Bridge/Mercure/Resources/assets/package.json#L23

dunglas:
I agree that we must either tell the user to not install the dep directly or move it to peer dependencies (this looks obvious for most JS developers, but probably not to PHP developers because NPM works differently than Composer).
However for consistency we should follow the same strategy for all UX packages. WDYT @tgalopin?

src/Turbo/Resources/assets/package.json
```diff
@@ -19,10 +19,8 @@
         "test": "babel src -d dist && jest",
         "lint": "eslint src test"
     },
-    "dependencies": {
-        "@hotwired/turbo": "^7.0.0-beta.4"
-    },
     "peerDependencies": {
+        "@hotwired/turbo": "^7.0.0-beta.4",
```
nicolas-grekas, line 23: you should also list this as a devDep now

src/Turbo/README.md
nicolas-grekas: can be removed now, see https://github.com/symfony/flex/pull/764

while at it, please check all package.json in this repo :)

weaverryan:
This is ready.

We *should* also move `chart.js`, `swup` (and plugins) and `cropperjs` to `peerDependencies`. However, I'd like to get https://github.com/symfony/flex/pull/764 merged & tagged first, to minimize any "experience" change people have (i.e. new projects will STILL not need to worry about manually adding those libs to their `package.json`). But since `ux-turbo` is unreleased, I'd like to get this correct before we do. Once the Flex PR is merged (and maybe after some time... to make sure most people are using the newest version) we could remove the addition to the README from this PR.

Thanks!

dunglas (approved):

nicolas-grekas:
flex v1.13.0 now released

nicolas-grekas:
> Once the Flex PR is merged (and maybe after some time...

no need to wait: just conflict with flex < 1.13

weaverryan:
All `package.json` dependencies have now been moved to `peerDependencies` (and `devDependencies` so tests can run). For the packages that now have `peerDependencies`, a `composer.json` conflict has been added to require `symfony/flex` `1.13` at least so the synchronization happens.

We should be ready to go!

nicolas-grekas (approved):

PR #100, opened 2021-06-16
Update Form example

pierrejoye:
Without specifying the action, Turbo redirect the full page to /.

name="product_new" is not necessary in this specific example, it however may help newcomer to understand they can have separate route for this action (even if not recommended by the Symfony Form docs).

Not sure if it is a bug in turbo or me doing wrong, adding this fixes it.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Tickets       | Fix #... 
| License       | MIT

### DIFF
--- a/src/Turbo/README.md
+++ b/src/Turbo/README.md
@@ -102,11 +102,9 @@ automatically:
 public function newProduct(Request $request): Response
 {
     return $this->handleForm(
-        $this->createForm(ProductFormType::class, null,
-        [
+        $this->createForm(ProductFormType::class, null, [
             'action' => $this->generateUrl('product_new'),
-        ]
-        ),
+        ]),
         $request,
         function (FormInterface $form) {
             // save...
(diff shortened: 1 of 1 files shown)


weaverryan:
Hey @pierrejoye - thanks for this… and moving it from the split ;).

I’m not familiar with this specific bad behavior with a blank action. Does the form submit to the wrong url for you? You mentioned it redirected - was that after a successful form submit, or did it redirect after an unsuccessful form submit to the wrong page?

Cheers!

src/Turbo/README.md
```diff
@@ -176,6 +180,17 @@ $builder
         ]
     ]);
 ```
+> **NOTE Turbo ^7.0.0-beta.5:**
+> If Turbo ^7.0.0-beta.5 or lower (may be still valid with future version), there is a bug with turbo-frame and form submit. F.e., a link to add a product will work the 1st time, the frame will have the new product form, on submit, the frame will be updated using the products listing. At this point the new link won't work anymore. This is cache issue in Turbo. To avoid it add a ramdom value as get parameter, microtime(true) will make it unique for one client:
```
weaverryan, line 184: We should verify, but I think this is fixed on beta 7.

pierrejoye: @weaverryan yes, I will test master or latest release later today. For the time, may I suggest to keep the note? Took some time to dig the issues reports and the turbo code :) I wish noone to have to do it ;)

pierrejoye:
> Hey @pierrejoye - thanks for this… and moving it from the split ;).
> 
> I’m not familiar with this specific bad behavior with a blank action. Does the form submit to the wrong url for you? You mentioned it redirected - was that after a successful form submit, or did it redirect after an unsuccessful form submit to the wrong page?
> 
> EDIT: oh, is your form in a turbo-frame? Indeed, in that case, with no action, it’s very easy to end up with that form on a url that doesn’t match the form submit. And then, the form submits to the wrong (current) url. We should specify the action more consistently to avoid this problem. The question is, as a general recommendation, whether we should show the action being passed when creating the form in PHP or via form_row() in Twig. I kind of prefer the latter, but I don’t have a strong reason.

Yes, it is in a turbo-frame, it refresh the full page at /, not the current one. Is it the expected behaviors? Reading the turbo doc, I was not expecting this as they mentioned they will follow the server redirect and update the frame.  I am still discovering and taking notes of the parts I had to dig in :)

src/Turbo/README.md
pierrejoye: I confirmed, beta.7 works as expected.

weaverryan:
Thanks @pierrejoye!

PR #106, opened 2021-06-21
Twig/LiveComponent Attributes

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | -
| License       | MIT

Based on discussions with @weaverryan, the consensus from the core team seems to be to require PHP8+ and use attributes for defining components instead of the interface. The updated [readme](https://github.com/symfony/ux/blob/a80d8df64a41860d2fec407dfecac36ce34809a1/src/TwigComponent/README.md) shows how this would now work.

Assuming this is the direction we want to go, I will update the `LiveComponent` package.

Todo:
- [x] Update `LiveComponent`
- [x] Fix duplicated `LiveProp` [issue](https://github.com/symfony/ux/pull/106#discussion_r659399618) on parent classes


src/TwigComponent/src/Attribute/TwigComponent.php
```diff
@@ -0,0 +1,54 @@
...
+
+    final public function getTemplate(): string
+    {
+        return $this->template ?? "components/{$this->name}.html.twig";
+    }
+
+    /**
+     * @internal
+     */
+    final public static function forClass(string $class): ?static
```
kbond, line 44: The `static` return seems to be what's causing the `coding-style-php` job to fail.

src/TwigComponent/src/Attribute/TwigComponent.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Symfony\UX\TwigComponent\Attribute;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @experimental
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class TwigComponent
```
kbond, line 20: My plan here is to have a `LiveComponent` attribute that extends `TwigComponent`.

src/TwigComponent/src/Attribute/TwigComponent.php
```diff
@@ -0,0 +1,54 @@
...
+
+    final public function getName(): string
+    {
+        return $this->name;
+    }
+
+    final public function getTemplate(): string
+    {
+        return $this->template ?? "components/{$this->name}.html.twig";
+    }
```
weaverryan, line 39: That's very nice

src/TwigComponent/src/Attribute/TwigComponent.php
weaverryan: which `static`? The first or the second?

src/TwigComponent/src/Attribute/TwigComponent.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @internal
+     */
+    final public static function forClass(string $class): ?static
+    {
+        $class = new \ReflectionClass($class);
+
+        if (!$attribute = $class->getAttributes(self::class, \ReflectionAttribute::IS_INSTANCEOF)[0] ?? null) {
+            return null;
+        }
```
weaverryan, line 50: Future-proofing for a `LiveComponent` attribute sub-class?

weaverryan (commented):
I like this approach!

src/TwigComponent/src/Attribute/TwigComponent.php
kbond: Sorry, the return type. It's new in PHP8 and coding-style-php runs on PHP7.4 (and an older version of php-cs-fixer).

wouterj:
I quickly discussed this on Slack already, but I think we should use a DI tag + `registerAttributeForAutoconfiguration()` for configuring components (and name + template). That seems to be the standard for all other Symfony attributes and it adds a nice extra: It means we have support for PHP <8.0 and Symfony <5.3 if a user just adds the tag to the component services.

It does require some rewriting of the renderer though, as it currently uses the attribute at runtime.

Is there anyone that can share their experiences with attributes on which approach is the best (based on performance or other thing that we didn't come up with)?

src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
```diff
@@ -27,23 +28,31 @@ public function process(ContainerBuilder $container): void
     {
         $serviceIdMap = [];
 
-        foreach (array_keys($container->findTaggedServiceIds('twig.component')) as $serviceId) {
-            $definition = $container->getDefinition($serviceId);
+        foreach ($container->getDefinitions() as $id => $definition) {
```
dbrekelmans, line 31: I think you already know, but just in case: if you use `registerAttributeForAutoconfiguration`, you can keep this implementation as it was (with `$container->findTaggedServiceIds('twig.component')`)

src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
```diff
@@ -27,23 +28,31 @@ public function process(ContainerBuilder $container): void
...
+            $class = $definition->getClass();
+
+            if (!class_exists($class, false) || !$attribute = TwigComponent::forClass($class)) {
+                continue;
+            }
+
+            $name = $attribute->getName();
```
dbrekelmans, line 38: In case you want to support doctrine annotations as well, you could add the following (if no name was defined in the tag)

```php
private function getNameFromDoctrineAnnotation(ContainerBuilder $container, Definition $definition): ?string
    {
        if (!$container->has(Reader::class)) {
            return null;
        }

        /** @var Reader $reader */
        $reader = $container->get(Reader::class);
        if ($reader === null) {
            return null;
        }

        $reflectionClass = new ReflectionClass($definition->getClass());

        /** @var Component|null $componentAnnotation */
        $componentAnnotation = $reader->getClassAnnotation($reflectionClass, Component::class);

        if ($componentAnnotation === null) {
            return null;
        }

        return $componentAnnotation->name;
    }
```

src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
dbrekelmans: BTW an easy way to implement this, is the following in `TwigComponentExtension.php`:

```php
// Unfortunately only works for PHP8 attributes, not for doctrine annotations
$container->registerAttributeForAutoconfiguration(TwigComponent::class, static function (ChildDefinition $definition, Component $attribute): void {
    $definition->addTag('twig.component', get_object_vars($attribute));
});
```

You would need to make the `TwigComponent` properties public I think and you could also remove the getters that way.

It's done the same way in `FrameworkExtension.php` (symfony/framework-bundle):

```php
$container->registerAttributeForAutoconfiguration(AsEventListener::class, static function (ChildDefinition $definition, AsEventListener $attribute): void {
    $definition->addTag('kernel.event_listener', get_object_vars($attribute));
});
```

dbrekelmans (commented):
You were just a bit ahead of me.. I hadn't seen this PR and was working on an almost identical implementation ;) Looks good so far!

kbond:
I've managed to use `registerAttributeForAutoconfiguration` in lieu of the compiler pass. Twig components can also be manually configured by adding a `twig.component` tag with a `key` parameter for the *name*.

src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
kbond: Thanks! I opted to set the tag parameters manually. `name` is reserved for the tag's name in XML.

src/TwigComponent/README.md
```diff
@@ -13,17 +13,13 @@ Every component consists of (1) a class:
 // src/Components/AlertComponent.php
 namespace App\Components;
 
-use Symfony\UX\TwigComponent\ComponentInterface;
+use Symfony\UX\TwigComponent\Attribute\TwigComponent;
 
-class AlertComponent implements ComponentInterface
+#[TwigComponent('alert')]
```
nicolas-grekas, line 18: this should follow the established convention and be named `AsTwigComponent`

wouterj: Actually, I don't really understand why it's coupled with twig. The nice thing about attributes vs interfaces it that it removes coupling and allows a generic way to configure metadata. What about `AsComponent`?

weaverryan:
> The failing test is related to components with the same name. I no longer ensure they are unique. Right now, the last one registered wins. Thoughts? Should I add a compiler pass to ensure they are unique? (I believe a compiler pass is the only way)

I vote for last one wins - like service id's

> Also, I now require Symfony 5.3+ so I can use registerAttributeForAutoconfiguration but could allow older version if desired. On older versions, components could only be registered manually.

If it's just a matter of doing a `method_exists()` before calling that method, then I'm +1 for doing that and at least allowing 5.2 and lower users to use this library with manual tagging.

src/TwigComponent/README.md
weaverryan: But in practice, the library that provides this component IS coupled to Twig (purposely). I'm also worried that `AsComponent` is a little generic (that component word is loaded).

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -20,19 +21,20 @@
...
     }
 
-    public function render(ComponentInterface $component): string
+    public function render(object $component): string
     {
-        // TODO: Template attribute/annotation/interface to customize
         // TODO: Self-Rendering components?
-        $templateName = sprintf('components/%s.html.twig', $component::getComponentName());
+        if (!$attribute = TwigComponent::forClass($component::class)) {
+            throw new \InvalidArgumentException(sprintf('"%s" is not a Twig Component, did you forget to add the TwigComponent attribute?', $component::class));
```
dbrekelmans, line 35: Doesn't this block users from specifying the component name and template directly in the tag (without using the attribute)?

I would go the route that the attribute is only used for auto-configuration and that the actual tag is used to find the name and template in all other places. That way you support both ways of configuring the component

kbond: I think this is similar to what @wouterj was thinking as well. I suppose we could have some kind of component name => template map injected into the `ComponentRenderer` but the only way to get the component name from an instance is to check the attribute...

kbond:
Do we want the `AsTwigComponent` to be completely optional as @dbrekelmans suggests above? Allow both the component name and template to be set manually via the `twig.component` tag? How should I map the component to a template for the `ComponentRenderer`? Build a component class => template map?

src/TwigComponent/src/ComponentRenderer.php
dbrekelmans: I would pass the template to the render function actually. It makes sense to me to say: "render this component using this template". Theoretically you could tag multiple services with the same component class passing a different template (although in practice I don't think this will be common)

kbond: That pushes the problem up to the twig extension (and, for live components, the event subscriber). Somehow, for `{{ component('my_component') }}` *my_component* needs to be mapped to a template. Or are you thinking the template should be passed to the component twig function?

kbond:
> I vote for last one wins - like service id's

I went with this.

> If it's just a matter of doing a method_exists() before calling that method, then I'm +1 for doing that and at least allowing 5.2 and lower users to use this library with manual tagging.

Manual tagging/4.4+ is now supported (I haven't updated the gh action to prove this). PHP8 and the `AsLiveComponent` is still required. Manual tagging is a little annoying as your `key` tag attribute must match your `AsLiveComponent` name property.

src/TwigComponent/src/ComponentRenderer.php
dbrekelmans: Hmm true. What about building a name => template map in the `TwigComponentPass` and setting the map as an argument to the `ComponentRuntime`? So it'd be something like this:

```php
public function __construct(ComponentFactory $componentFactory, ComponentRenderer $componentRenderer, array $templateMapping)
    {
        $this->componentFactory = $componentFactory;
        $this->componentRenderer = $componentRenderer;
        $this->templateMapping = $templateMapping;
    }

    {
        return $this->componentRenderer->render(
            $this->componentFactory->create($name, $props),
            $this->templateMapping[$name]
        );
    }

```

kbond: This would work in the `TwigComponent` package but would create a problem in the `LiveComponent` package. It needs to [pass an already created component instance to render](https://github.com/symfony/ux/blob/c118092650561e8e92cf87edf81448690b310da0/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php#L234).

dbrekelmans: The subscriber has access to the container so you could just get the ComponentRuntime instead of ComponentRenderer to render the template right?

kbond: Sorry, I'm not following. The `ComponentRuntime` example above wouldn't work for the subscriber, the component is already created.

dbrekelmans: Maybe we can talk on the symfony slack about it, might be a bit easier. I'm not available tonight, but I'll try to be online in the evenings later this week

weaverryan:
> Do we want the AsTwigComponent to be completely optional as @dbrekelmans suggests above?

Maybe? Right now we only have name and template options... but there is already the question of:

> How should I map the component to a template for the ComponentRenderer? Build a component class => template map?

And, what if we have more options in the future? I can envision that, if we decide to put it all on the tag, then we would, in a compiler pass, push that tag data into some sort of "component metadata service" where we would probably have a method like `getMetadata($class)` that would return an object that's very similar to the `AsLiveComponent`. But it would be tricky to make this service lazy (we would probably inject ALL of the tag data from ALL of the components into its constructor, so instantiating it would put that all into memory).

So I think we should NOT do this. We should follow the API Platform model (or the Doctrine entity model) where the metadata is pulled lazily at runtime when needed. The only stuff in the tag is the stuff that is useful to know without needing to instantiate the objects (e.g. with commands, it's nice to know the name of each command without instantiating it, and the same with Twig components).

If we *were* to make the PHP8 attribute optional, I think it makes sense to do so with an alternate configuration format - e.g. a PHP config format (like routing config or validation config).

weaverryan:
> Manual tagging is a little annoying as your key tag attribute must match your AsLiveComponent name property.

What happens if it doesn't? It seems like an odd situation, but I would imagine that the tag `name` would "win" over the one in the attribute.

kbond:
> So I think we should NOT do this

I agree.

> What happens if it doesn't?

For twig components, I don't believe it will be a problem as we never, at runtime, get the component name from the class. For live components, as of right now, the url is generated with the name the component class gives. The subscriber would look for the component based on the DI tag so there could be a mismatch.

dbrekelmans:
Fair points @weaverryan, I agree.

kbond:
> I could enforce this in the compiler pass.

Alternatively, and I think I prefer this, I can remove the requirement to set a tag `key` and configure the service locator in the compiler pass. This would remove any possibility of a mismatch when manually tagging.

kbond:
I think the TwigComponent refactor is complete and ready for review. I added a GH action job to run on php8 with low deps to prove it's compatible with 4.4.

> Alternatively, and I think I prefer this, I can remove the requirement to set a tag key and configure the service locator in the compiler pass. This would remove any possibility of a mismatch when manually tagging.

I went with this. No possibility for a mismatch. The "name" is now only pulled from the attribute.

dbrekelmans (approved):

src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
```diff
@@ -25,25 +28,30 @@ final class TwigComponentPass implements CompilerPassInterface
...
+        $componentMap = [];
 
-        foreach (array_keys($container->findTaggedServiceIds('twig.component')) as $serviceId) {
-            $definition = $container->getDefinition($serviceId);
+        foreach (array_keys($container->findTaggedServiceIds('twig.component')) as $id) {
+            $componentDefinition = $container->findDefinition($id);
 
-            // make all component services non-shared
-            $definition->setShared(false);
+            $attributes = (new \ReflectionClass($componentDefinition->getClass()))
```
weaverryan, line 36: Are we able to use `$container->getReflectionClass()`?

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -31,13 +31,18 @@ final class TwigComponentExtension extends Extension
...
+                static function (ChildDefinition $definition) {
+                    $definition->addTag('twig.component');
+                }
+            );
+        }
 
         $container->register(ComponentFactory::class)
             ->setArguments([
-                new ServiceLocatorArgument(new TaggedIteratorArgument('twig.component', null, 'getComponentName')),
+                new ServiceLocatorArgument(),
```
weaverryan, line 45: Use an abstract argument here? Since it will entirely be filled in via the compiler pass?

weaverryan (commented):
Minor comments - but this looks really solid! Let's roll this into LiveComponents 👍

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
kbond: Cool, didn't know about this but `AbstractArgument` is a 5.1+ feature.

kbond:
I have switched LiveComponents to PHP8 attributes (only) so this is ready for another round of reviews (ping @dbrekelmans, @weaverryan).

Not sure why the coding-style-js job is failing.. Looks to be caused by the live component readme.

src/LiveComponent/src/Attribute/LivePropContext.php
```diff
@@ -0,0 +1,41 @@
...
+ * @experimental
+ *
+ * @internal
+ */
+final class LivePropContext
+{
+    private LiveProp $liveProp;
+    private \ReflectionProperty $reflectionProperty;
+
+    public function __construct(LiveProp $liveProp, \ReflectionProperty $reflectionProperty)
```
ker0x, line 26: As the package now requires PHP 8, wouldn't it be interesting to use [constructor property promotion](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion) here and in other classes where it can be implemented?

kbond: Absolutely, I didn't use these PHP 8 features as it breaks the php-cs-fixer job for the mono repo. I'm not sure how we want to handle this.

src/LiveComponent/composer.json
```diff
@@ -40,7 +40,8 @@
         "doctrine/doctrine-bundle": "^2.0",
         "doctrine/orm": "^2.7",
         "zenstruck/foundry": "^1.10",
-        "zenstruck/browser": "^0.5.0"
+        "zenstruck/browser": "^0.5.0",
+        "doctrine/annotations": "^1.0"
```
weaverryan, line 44: Was this intentional?

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -0,0 +1,123 @@
...
+    {
+        $class = $classOrObject instanceof \ReflectionClass ? $classOrObject : new \ReflectionClass($classOrObject);
+
+        foreach ($class->getProperties() as $property) {
+            yield $property;
+        }
+
+        if ($parent = $class->getParentClass()) {
+            yield from self::propertiesFor($parent);
+        }
```
weaverryan, line 121: To get all of the properties on a class, we need to manually do the recursion up the tree? Does that handle traits? What if a property is overridden in a child component? Is it reported twice?

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param string|object $classOrObject
+     *
+     * @return \ReflectionProperty[]
+     */
+    private static function propertiesFor($classOrObject): \Traversable
```
weaverryan, line 111: I've done a quick scan (perhaps too quick), and I don't see any instance where we pass a string `$class` to any of the methods on this annotation. Could we change everything to `object $object`?

src/LiveComponent/src/Attribute/LivePropContext.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->liveProp = $liveProp;
+        $this->reflectionProperty = $reflectionProperty;
+    }
+
+    public function liveProp(): LiveProp
+    {
+        return $this->liveProp;
+    }
+
+    public function reflectionProperty(): \ReflectionProperty
```
weaverryan, line 37: This is a nit pick, but I noticed that we only ever need the `->getName()` off of this - we could just pass in that string.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -48,7 +65,6 @@ public function load(array $configs, ContainerBuilder $container): void
             ->setArguments([
```
weaverryan, line 65: On the previous line, I just noticed we're registering some services with the class as their id. I'd like to avoid that - it makes them autowireable.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -48,7 +65,6 @@ public function load(array $configs, ContainerBuilder $container): void
...
                 new TaggedIteratorArgument('twig.component.property_hydrator'),
                 new Reference('property_accessor'),
-                new Reference('annotation_reader'),
                 '%kernel.secret%',
             ])
         ;
```
weaverryan, line 70: Below, for `LiveComponentSubscriber`, it's minor, but we could, I think, pass an abstract argument to arg 0, since it's filled in via the compiler pass.

weaverryan (commented):
Minor notes - this is looking great!!!

src/LiveComponent/composer.json
kbond: Unfortunately, yes. The newest version of doctrine (bundle or orm, not quite sure) doesn't require annotations. The lowest version we currently support doesn't allow attributes so I was stuck...

I didn't look too deep into the issue but this was an easy fix. I'm open to ideas!

src/LiveComponent/README.md
```diff
@@ -127,16 +119,18 @@ class RandomNumberComponent implements ComponentInterface
 ```
 
 To transform this into a "live" component (i.e. one that
-can be re-rendered live on the frontend), change your
-component's interface to `LiveComponentInterface`:
+can be re-rendered live on the frontend), replace
+component's `AsTwigComponent` attribute to `AsLiveComponent`:
```
dbrekelmans, line 123: I think it's either `change ... to` or `replace ... with` :)

src/LiveComponent/README.md
```diff
@@ -183,11 +177,13 @@ namespace App\Components;
 // ...
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
 
-class RandomNumberComponent implements LiveComponentInterface
+#[AsLiveComponent('random_number')]
+class RandomNumberComponent
 {
-    /** @LiveProp */
+    #[LiveProp]
```
dbrekelmans, line 183: Out of scope for this PR, but a general question: Why is it called `LiveProp` instead of `LiveProperty`?

src/LiveComponent/README.md
```diff
@@ -702,9 +694,8 @@ class PostFormComponent extends AbstractController implements LiveComponentInter
      * the form renders fields with names like `name="post[title]"`.
      * We set fieldName="" so that this live prop doesn't collide
```
dbrekelmans, line 695:
```suggestion
     * We set fieldName: '' so that this live prop doesn't collide
```

src/LiveComponent/README.md
```diff
@@ -1020,35 +996,27 @@ First use the `ValidatableComponentTrait` and add any constraints you need:
...
-
-    public static function getComponentName() : string
-    {
-        return 'edit_user';
-    }
 }
 ```
 
-Be sure to add the `@Assert\IsValid` to any property where you want
+Be sure to add the `Assert\IsValid` to any property where you want
```
dbrekelmans, line 1019: Is this also available as a PHP8 attribute? Maybe clearify it by writing `... add the #[Assert\IsValid] attribute to any ...`

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -0,0 +1,123 @@
...
+    {
+        $class = $classOrObject instanceof \ReflectionClass ? $classOrObject : new \ReflectionClass($classOrObject);
```
dbrekelmans, line 113:
```suggestion
        $class = $classOrObject instanceof \ReflectionClass ?: new \ReflectionClass($classOrObject);
```

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -47,11 +45,12 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
     private const JSON_FORMAT = 'live-component-json';
     private const JSON_CONTENT_TYPE = 'application/vnd.live-component+json';
 
-    /** @var ContainerInterface */
-    private $container;
+    private array $componentServiceMap;
```
dbrekelmans, line 48: Can we specify with `@var` what type we expect in this array?

src/LiveComponent/tests/Fixture/Kernel.php
```diff
@@ -61,9 +62,16 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
...
         $c->register('logger', NullLogger::class);
 
-        $c->register(Component1::class)->setAutoconfigured(true)->setAutowired(true);
-        $c->register(Component2::class)->setAutoconfigured(true)->setAutowired(true);
-        $c->register(Component3::class)->setAutoconfigured(true)->setAutowired(true);
+        $componentA = $c->register(Component1::class)->setAutoconfigured(true)->setAutowired(true);
+        $componentB = $c->register(Component2::class)->setAutoconfigured(true)->setAutowired(true);
+        $componentC = $c->register(Component3::class)->setAutoconfigured(true)->setAutowired(true);
+
+        if (self::VERSION_ID < 50300) {
```
dbrekelmans, line 69: Shouldn't this be lower than 8? Since we autoconfigure with attributes now.

src/TwigComponent/tests/Fixture/Kernel.php
```diff
@@ -50,13 +50,21 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
...
-        $c->register(ComponentC::class)->setAutoconfigured(true)->setAutowired(true);
+        $service = $c->register(ServiceA::class)->setAutoconfigured(true)->setAutowired(true);
 
-        if ('multiple_component_b' === $this->environment) {
-            $c->register('different_component_b', ComponentB::class)->setAutoconfigured(true)->setAutowired(true);
+        $componentA = $c->register(ComponentA::class)->setAutoconfigured(true)->setAutowired(true);
+        $componentB = $c->register('component_b', ComponentB::class)->setAutoconfigured(true)->setAutowired(true);
+        $componentC = $c->register(ComponentC::class)->setAutoconfigured(true)->setAutowired(true);
+
+        if (self::VERSION_ID < 50300) {
```
dbrekelmans, line 59: Shouldn't this be lower than PHP 8?

dbrekelmans (commented):
Nice work @kbond!

src/LiveComponent/src/Attribute/AsLiveComponent.php
weaverryan: We had a report on Slack (with the existing code) that this algorithm (which I think you just copied from a different part of the code for this PR) double-counts LiveProps in a parent class. In the `live-demo`, they created a `ChildNotificationComponent` and made it extend `NotificationComponent` - https://github.com/weaverryan/live-demo/blob/main/src/Twig/Components/NotificationComponent.php and got:

> An exception has been thrown during the rendering of a template ("The field name "expanded" cannot be used by multiple LiveProp properties in a component. Currently, both "expanded" and "expanded" are trying to use it in "App\Twig\Components\ChildNotificationComponent". Try adding LiveProp(fieldName="somethingElse") for the "expanded" property to avoid this.").

So indeed, I think we may have some double counting :).

kbond: Ah ok, so what do you think? First one wins?

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
kbond: Ah good point. I will change in both bundles.

src/LiveComponent/tests/Fixture/Kernel.php
kbond: I'm checking the Symfony version here. Auto configuration with attributes is only available in Symfony 5.3+.

dbrekelmans: Ahh sorry. Cool 👍

src/LiveComponent/composer.json
weaverryan: That's fine then - this is just for running some functional/integration test that contain a Doctrine entity (and we use annotations for those to function on all supported DoctrineBundle versions)?

src/LiveComponent/src/Attribute/AsLiveComponent.php
weaverryan: Hmm, probably if you literally have 2 properties that both have `LiveProp` it *should* have an error. But in the user's case, I think (though I have no tested) that they don't actually have duplicated properties - it's just that our system is double-counting the same ONE property (e.g. perhaps counting it once when it scans the actual class [even though it lives in the parent] and a 2nd time when it scans the parent).

But also: especially if we force an *object* to be passed in (not just a `$class`), is there really no built-in way to get all the properties of an object (including parent classes and traits)?

src/LiveComponent/README.md
weaverryan: Because, conceptually, these have some similarities to React props. Though, that is less true now (we had originally thought that we might have `LiveProp` and `LiveState` (the difference being that the 2nd is modifiable and the first is not). But, for simplicity (iirc), we changed just `LiveProp` with a `writable` option. 

Anyways, the "prop" comes from React props. I think it also sounds cooler ;)

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
kbond: I think we would have the same problem as discussed [here](https://github.com/symfony/ux/pull/106#discussion_r659069501). How should I handle older versions?

src/LiveComponent/src/Attribute/LivePropContext.php
kbond: We do the actual `\ReflectionProperty` [here](https://github.com/symfony/ux/pull/106/files#diff-b6ca1bfe380a148203d16367e6f7ae91a700b7d7b38b0d34114d4995c9eeafefR227).

src/LiveComponent/src/Attribute/AsLiveComponent.php
kbond: This would make `$class` `true` if it's an instance of `\ReflectionClass`.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
kbond: Done.

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
weaverryan: Yea, I think so. Its presence will remind us that we can remove the conditional code someday :).

src/LiveComponent/src/Attribute/LivePropContext.php
weaverryan: Ah, ok I missed it 👍

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
weaverryan: I say let’s use it (conditionally) if it’s not too much trouble.

src/LiveComponent/src/Attribute/AsLiveComponent.php
dbrekelmans: Ah my bad

src/LiveComponent/src/Attribute/AsLiveComponent.php
kbond: Confirmed the issue and added a fix in 4d3c8e87cb1a1df72d78200d3a0ce7cb809e4c37. I took a _first one found_ wins approach. Let me know if you'd like me to add something more robust.

> But also: especially if we force an object to be passed in (not just a $class), is there really no built-in way to get all the properties of an object (including parent classes and traits)?

Not for private/protected properties I don't believe (we allow LiveProp's to be private/protected).

weaverryan (approved):
👍

weaverryan:
Thank you Kevin for this huge work!!

Issue #102, opened 2021-06-18
[LiveComponent] Meta issue for todos, changes & bugs


weaverryan:
Hi!

Earlier, the TwigComponent and LiveComponent packages were introduced. Since the community is just getting its first look at these, and because there are known TODOs and ideas, I though it would be best to start with a single meta issue to track things :).

Some of these reference the "Demo" - found here: https://github.com/weaverryan/live-demo

### Features

~~Feature A) Add support for `@Security` annotation/attribute. Also for `@Cache` annotation.~~ #116

~~Feature B) Action arguments - e.g. `data-action-name="delete(id=5)"`. The frontend syntax is already supported. We need to add a way to "allow" certain action arguments to be "passable" from the frontend.~~ #218

Feature C) CSS Transition support for removed/added elements

~~Feature D) Ability to add cache headers in addition to the `@Cache` annotation~~ #116

~~Feature E) Allow loading behavior to be specific to a model or action -e.g. `data-loading="action(save)|show"` to "show" this element only when the "save" action is loading.~~ #470

~~Feature F) Ability to control the "route" for a component, instead of using the built-in `/_components/{componentName}/{actionName}`. And the same for specific actions of a component.~~ WONTFIX unless requested with a good reason.

Feature G) Audit and add more JavaScript events that are dispatched.

Feature H) Testing tools: tools to help you create a component with some data, call actions on it, and assert final things on the rendered HTML.

Feature I) Allow controlling the debounce time on a component and model level.

~~Feature J) A system to allow for computed properties (e.g. a method that is called only once, then its value is returned every time after)~~

~~Feature K) A `make:component` command~~

Feature L) Look again at "child" components and how they communicate back and forth

### ~~Changes~~

~~Change A)  Investigate making the initial render a sub-request. This would make initial render more consistent with a re-render, but we need to check the performance penalty.~~ Will not do for now - can revisit later.

~~Change B) `getComponentName()` -> `getName()`~~

~~Change C) Investigate removing `ComponentInterface` and `LiveComponentInterface` in favor of annotations/attributes.~~ #106

~~Change D) Consider forcing the template name to be configured (e.g. a `template=""` option on the annotation/attribute from above~~ #106

~~Change E) Normalize naming of hooks (Before / Pre/Post)~~ #281

~~Change F) Change annotation's constructor - https://github.com/symfony/ux/pull/101#discussion_r654411346~~

[...]

PR #111, opened 2021-06-28
[LiveComponent] Adding rollup to make an esm build

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Addresses `Bug A` from #102
| License       | MIT

Hi!

This changes how LiveComponent is bundle. There are a few important pieces:

1) I decided to continue support IE 11. The difference in the unminified file size (including polyfills) is about 5kg (56kb vs 61kb).

2) I'm building ESM only right now. I have considered a UMD build... but I don't know if there is a real use-case. In practice, this library will always be used in a build system... but I cannot find any other libraries that only output ESM.

3) This sets `babelHelpers` to `runtime`, which means that instead of embedding the babel helpers (like `_typeof` https://github.com/symfony/ux/blob/e239983ef55805224df7eef4fe7eff089c70b7cd/src/LiveComponent/assets/dist/live_controller.js#L3) in every file, they all import from `@babel/runtime`. 

4) We're testing the `src/` files, not the `dist/` files.

I don't have a lot of experience with compiling files for the frontend - and there seems to be a lot of variability on how to do it. If you see anything weird, let me know :).

Cheers!


src/LiveComponent/assets/rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+            name: 'LiveController',
+            file: 'dist/live-controller.esm.js',
+            format: 'esm',
+            sourcemap: false,
+        },
+        external: [
+            /@babel\/runtime/,
+            /core-js\//,
+            'stimulus',
+        ],
```
Kocal, line 17: In the _Rollup world_ we usually automate `external` by using the following code:
```js
import pkg from './package.json';

export default [
  {
    // ...
    external: Object.keys(pkg.dependencies)
  }
]
```

WDYT?

src/LiveComponent/assets/package.json
```diff
@@ -1,16 +1,17 @@
 {
     "name": "@symfony/live-stimulus",
     "description": "Live Component: bring server-side re-rendering & model binding to any element.",
-    "main": "dist/live_controller.js",
+    "main": "dist/live-controller.esm.js",
```
Kocal, line 4: Do you think it would be interesting to also configure `module` field?

weaverryan: Even though they're identical, yea, I think that's a good idea.

src/LiveComponent/assets/rollup.config.js
weaverryan: I want to do whatever is standard :). But, the `@babel/runtime` needed to be a regex so that it matches all the individual, nested paths that are actually imported - like `@babel/runtime/...`. I got that from the note in this section - https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers

Kocal: Oh alright, I wasn't aware of that

PR #113, opened 2021-06-29
[LiveComponent] Smooth out parent-child component relationship

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Tickets       | #102 addresses **Bug C** and **Bug F**
| License       | MIT

This PR is to address a few outstanding issues related to nested or "parent/child" components:

* [X] #102 Bug C: When a model is updated in a child component, that parent component does not receive that component update (e.g. imagine a form that renders a component... and one field is rendered inside another component - when the child component's model is updated, it should also update that same model in the parent component [the form]).
* [X] Change `data-model` to take priority over `name` so that child components can use `data-model` and allow parent components to still use the `name` attribute to match up with their model.
* [X] Change the new `live:update-model` event to also (when possible) pass the relevant element so that the listener can look at both the `name` and `data-model` attribute to find a match. Or find an alternative solution.
* [X] Document how parent/child components work (e.g. that when a parent re-renders, the child does not re-render)
* [X] #102 Bug F:  Add a way to "force" a child component to re-render
* [x] Document a known edge case: if a child component re-renders, and the shared model value (shared with a parent component) CHANGES in the AJAX response (e.g. `content=foo` is sent to the server but `content=FOO` is returned, and so `FOO` is now the new value of the `content` model), a parent will not be aware of this change.
* [x] Add way to "map" a child model onto a parent model so that you can "map" a private `LiveProp` from a child onto a parent so that it's not lost if the parent renders over the child.
* [x] Standardize / cleanup more tests with `mockRerender()`.

### DIFF
--- a/src/LiveComponent/README.md
+++ b/src/LiveComponent/README.md
@@ -361,6 +361,9 @@ code works identically to the previous example:
 </div>
 ```
 
+If an element has _both_ `data-model` and `name` attributes, the
+`data-model` attribute takes precedence.
+
 ## Loading States
 
 Often, you'll want to show (or hide) an element while a component is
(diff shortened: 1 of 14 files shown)


weaverryan:
This is ready to go - the `README.md` is updated. Mostly, this makes parent-child relationships much smarter to capture the "95%" use-case without any extra config.

kbond (approved):
Reviewed the readme: looks great. The _full example_ is a nice touch!

PR #115, opened 2021-07-01
[bug] remove duplicated doctrine/annotations entry

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Looks like 2 different PR's added this. (https://github.com/symfony/ux/pull/111 and https://github.com/symfony/ux/pull/106)


jmsche (approved):

weaverryan:
Thanks :). I was expecting a conflict would arise that would cause us to fix this. But right now (and we should fix this at some point), or composer.json isn't alphabetic, so it's easier to have this happen. Anyways, good catch!

PR #116, opened 2021-07-01
[LiveComponent] Add a required "default" action for live components

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | #102 (feature A/D)
| License       | MIT

In order to use the `@Security/@Cache` annotations for your components, we need the default action to be on the component itself. Currently, for the default action, the component is wrapped inside an internal callable `DefaultComponentController` object. This won't inherit these annotations from the component class. The best solution @weaverryan and I could think of is to enforce a "default action" method exists on your component itself. 

By default, `__invoke()` is used but this can be customized via the `AsLiveComponent::$defaultAction` property. A `DefaultActionTrait` is provided that just adds an empty `__invoke()` method.

### DIFF
--- a/src/LiveComponent/README.md
+++ b/src/LiveComponent/README.md
@@ -16,10 +16,13 @@ A real-time product search component might look like this:
 namespace App\Components;
 
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
 
 #[AsLiveComponent('product_search')]
 class ProductSearchComponent
 {
+    use DefaultActionTrait;
+
     public string $query = '';
 
     private ProductRepository $productRepository;
@@ -130,18 +133,21 @@ class RandomNumberComponent
 
 To transform this into a "live" component (i.e. one that
 can be re-rendered live on the frontend), replace the
-component's `AsTwigComponent` attribute with `AsLiveComponent`:
+component's `AsTwigComponent` attribute with `AsLiveComponent`
+and add the `DefaultActionTrait`:
 
 ```diff
 // src/Components/RandomNumberComponent.php
 
 -use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 +use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
++use Symfony\UX\LiveComponent\DefaultActionTrait;
 
 -#[AsTwigComponent('random_number')]
 -#[AsLiveComponent('random_number')]
 class RandomNumberComponent
 {
++    use DefaultActionTrait;
 }
 ```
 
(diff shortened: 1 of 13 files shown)


src/LiveComponent/src/DefaultActionTrait.php
```diff
@@ -15,24 +15,11 @@
...
-    }
-
     public function __invoke(): void
     {
-    }
-
-    public function getComponent(): object
-    {
-        return $this->component;
+        // noop - this is the default action
```
weaverryan, line 23: Maybe some docs above it?

```
The "default" action for a component.

This is executed when your component is being re-rendered, but no custom action
is being called. You probably don't want to do any work here because this method
is *not* executed when a custom action is triggered. 
```

weaverryan (commented):
Very nice - even if NEEDING to force the user to have an `__invoke()` method isn't our favorite (though `DefaultActionTrait` will help with that)

src/LiveComponent/composer.json
```diff
@@ -53,5 +53,11 @@
             "url": "https://github.com/symfony/ux"
         }
     },
+    "repositories": [
+        {
+            "type": "path",
+            "url": "../TwigComponent"
+        }
+    ],
```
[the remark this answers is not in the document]
kbond: Oops! I didn't, it's a relic from my last PR.

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -21,6 +21,32 @@
 #[\Attribute(\Attribute::TARGET_CLASS)]
 final class AsLiveComponent extends AsTwigComponent
 {
+    private string $defaultAction;
+
+    public function __construct(string $name, ?string $template = null, string $defaultAction = '__invoke')
```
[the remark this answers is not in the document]
kbond: Yeah, still new to attributes but I thought (since most Symfony users are converting annotations to attributes) the common way to configure would be with named parameters.

src/LiveComponent/src/DefaultActionTrait.php
kbond: Haha thanks, as I was writing the doc I thought: "don't put too much effort here, Ryan will suggest something better".

src/LiveComponent/src/Attribute/AsLiveComponent.php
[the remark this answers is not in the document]
kbond: For ones with defaults, I guess just keep tacking onto the end to keep BC. Hopefully there will be no more required arguments....

src/LiveComponent/README.md
```diff
@@ -430,10 +436,18 @@ changes until loading has taken longer than a certain amount of time:
 
 ## Actions
 
-You can also trigger actions on your component. Let's pretend we
-want to add a "Reset Min/Max" button to our "random number"
-component that, when clicked, sets the min/max numbers back
-to a default value.
+Live component's require a single "default action" that is
```
weaverryan, line 439:
```suggestion
Live components require a single "default action" that is
```

src/LiveComponent/README.md
```diff
@@ -524,6 +538,26 @@ class RandomNumberComponent
...
+
+#[AsLiveComponent('product_search', defaultAction: 'defaultAction()')]
+class ProductSearchComponent
+{
+    public function defaultAction(): void
+    {
+    }
+}
+```
+
```
weaverryan, line 560: Boy, I just can't think of a use-case for this. I'm inclined to not document it at all!

weaverryan (commented):
just 2 minor comments about the docs - thanks!

src/LiveComponent/README.md
kbond: I know, me either. I almost documented injecting services but again, couldn't think of a use case.

I'm in favor of axing this section. What do you think?

weaverryan: 💯 let's kill it, but obviously, keep the option.

Could you also add one bullet to the CHANGELOG for LiveComponent?

weaverryan (approved):

weaverryan:
Thanks Kevin!

PR #117, opened 2021-07-05
Update hotwire.dev links

t-richard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

Not sure why but hotwire.dev moved to hotwired.dev (see https://twitter.com/dhh/status/1411720913641373704)

Here I changed all links to the new domain in the main README files. There may be more elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ to build the chart in PHP. The JavaScript is handled for you automatically.
 
 ## Components of UX
 
-Symfony UX leverages [Stimulus](https://stimulus.hotwire.dev/) for JavaScript
+Symfony UX leverages [Stimulus](https://stimulus.hotwired.dev/) for JavaScript
 and the [Stimulus Bridge](https://github.com/symfony/stimulus-bridge) for
 integrating it into [Webpack Encore](https://github.com/symfony/webpack-encore).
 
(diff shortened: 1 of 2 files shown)


javiereguiluz:
@t-richard Stimulus was mostly created by a developer that abandoned his company because of some discussions (I won't give more details or links to not contribute to the discussion) ... so the old employee redirected the original URL to another page and that's why we must updated URLs as soon as possible. Thanks!

t-richard:
@javiereguiluz that's a pity to take revenge like that but anyway, there is no other choice for us than to adapt.

It may be worth noting there are also links to the old site in some blog posts on symfony.com.

Here are two I noticed but again there may be more :
* https://symfony.com/blog/new-in-symfony-the-ux-initiative-a-new-javascript-ecosystem-for-symfony#symfony-ux-building-highly-interactive-applications-by-leveraging-javascript-giants 
* https://symfony.com/blog/webpack-encore-1-0-and-stimulus-bridge-2-0-released#symfony-stimulus-bridge-2-0

javiereguiluz:
Good catch! I've fixed those blog posts too. Thanks!

weaverryan:
Thank you Thibault!

PR #122, opened 2021-07-06
[Turbo] Update hotwired/turbo to beta 8

ker0x:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | none
| License       | MIT

This PR update `hotwired/turbo` to it's latest beta

### DIFF
--- a/src/Turbo/Bridge/Mercure/Resources/assets/package.json
+++ b/src/Turbo/Bridge/Mercure/Resources/assets/package.json
@@ -20,7 +20,7 @@
         "lint": "eslint src test"
     },
     "peerDependencies": {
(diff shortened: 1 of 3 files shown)


weaverryan (approved):
It's not strictly needed, as users would still download the latest.

On the nice side, this will cause beta8 to be listed in the user's `package.json` file.

On the negative, it means we only guarantee beta 8 support. But actually, I think that's good - it's beta, so we need to continue to guarantee support for only the latest version.

chalasr (approved):

weaverryan:
Thanks @ker0x!

PR #126, opened 2021-07-08
Update chart.js dependency

### LINKED ISSUE #86: [Chart.js] v3 support
Currently Symfony UX offers version 2.9.4 of Chart.js.
In december and the beginning of 2021 it was good.

But since april version 3 has been released. And we are already at 3.2 as I write this.

I also think it would be a good idea to specify on the repository readme (for each integration) which version is supported (or did I not see it?).
It would have saved me a few hours of trying to figure out why certain options weren't working.

[Migration guide](https://www.chartjs.org/docs/latest/getting-started/v3-migration.html)

Can you tell us what the procedure would be for the update?

gbere:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes/no
| Tickets       | Fix #86
| License       | MIT

### DIFF
--- a/src/Chartjs/Resources/assets/dist/controller.js
+++ b/src/Chartjs/Resources/assets/dist/controller.js
@@ -17,7 +17,9 @@ exports["default"] = void 0;
 
 var _stimulus = require("stimulus");
 
-var _chart = require("chart.js");
+var _auto = _interopRequireDefault(require("chart.js/auto"));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { "default": obj }; }
 
 function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
 
@@ -35,7 +37,7 @@ function _possibleConstructorReturn(self, call) { if (call && (_typeof(call) ===
 
 function _assertThisInitialized(self) { if (self === void 0) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return self; }
 
-function _isNativeReflectConstruct() { if (typeof Reflect === "undefined" || !Reflect.construct) return false; if (Reflect.construct.sham) return false; if (typeof Proxy === "function") return true; try { Date.prototype.toString.call(Reflect.construct(Date, [], function () {})); return true; } catch (e) { return false; } }
+function _isNativeReflectConstruct() { if (typeof Reflect === "undefined" || !Reflect.construct) return false; if (Reflect.construct.sham) return false; if (typeof Proxy === "function") return true; try { Boolean.prototype.valueOf.call(Reflect.construct(Boolean, [], function () {})); return true; } catch (e) { return false; } }
 
 function _getPrototypeOf(o) { _getPrototypeOf = Object.setPrototypeOf ? Object.getPrototypeOf : function _getPrototypeOf(o) { return o.__proto__ || Object.getPrototypeOf(o); }; return _getPrototypeOf(o); }
 
@@ -63,7 +65,7 @@ var _default = /*#__PURE__*/function (_Controller) {
         options: payload.options
       });
 
-      var chart = new _chart.Chart(this.element.getContext('2d'), payload);
+      var chart = new _auto["default"](this.element.getContext('2d'), payload);
 
       this._dispatchEvent('chartjs:connect', {
         chart: chart
--- a/src/Chartjs/Resources/assets/package.json
+++ b/src/Chartjs/Resources/assets/package.json
@@ -19,7 +19,7 @@
         "lint": "eslint src test"
     },
     "peerDependencies": {
-        "chart.js": "^2.9.4",
+        "chart.js": "^3.4.1",
         "stimulus": "^2.0.0"
     },
     "devDependencies": {
@@ -28,7 +28,7 @@
         "@babel/plugin-proposal-class-properties": "^7.12.1",
         "@babel/preset-env": "^7.12.7",
         "@symfony/stimulus-testing": "^1.1.0",
-        "chart.js": "^2.9.4",
+        "chart.js": "^3.4.1",
         "jest-canvas-mock": "^2.3.0",
         "stimulus": "^2.0.0"
     },
--- a/src/Chartjs/Resources/assets/src/controller.js
+++ b/src/Chartjs/Resources/assets/src/controller.js
@@ -10,7 +10,7 @@
 'use strict';
 
 import { Controller } from 'stimulus';
-import { Chart } from 'chart.js';
+import Chart from 'chart.js/auto';
 
 export default class extends Controller {
     connect() {


src/Chartjs/Resources/assets/dist/controller.js
```diff
@@ -63,7 +65,7 @@ var _default = /*#__PURE__*/function (_Controller) {
         options: payload.options
       });
 
-      var chart = new _chart.Chart(this.element.getContext('2d'), payload);
+      var chart = new _auto["default"](this.element.getContext('2d'), payload);
```
gbere, line 68: From tests:
```js
var chart = new _auto["default"](this.element.getContext('2d'), payload);
```
ReferenceError: ResizeObserver is not defined.

But from one application woks well

weaverryan:
Yo @gbere!

Thanks for this! I think this is an important one to get done. It's also the first time that we've hit this situation (though, I think it'll be fairly common): where we are bumping the dependency of a JavaScript library a full version. 

In a perfect world, we would have 2 controllers (for awhile at least) - a Chartjs v2 version and a Chartjs v3 version - and automatically (at build time) "toggle" between them based on which version of chartjs is installed. This is easy in theory (it's just an if statement), but Webpack doesn't like conditional imports like this :). You CAN do something like this (taken from React):

```js
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react.production.min.js');
} else {
  module.exports = require('./cjs/react.development.js');
}
```

But I don't believe you can add a more complex conditional that checks what version of Chart.js is installed (at least, we can't do this without requiring the user to add some extra fanciness to their webpack config, which we can't do). 

So, another solution would be to make 2 controllers (for v2 chartjs and v3) and then make the user toggle between them. That would be done by adding a 2nd controller here - https://github.com/symfony/ux/blob/351538011587680660c1fe0b09d4038200abad6b/src/Chartjs/Resources/assets/package.json#L8-L13 - probably called `chartv3`. Then, to avoid the user having BOTH activated when you install the package, you would set the original controller to `enabled: false`.

The downside of this would be that, if you use `chartv3`, then that is literally the new name of the controller (so the documentation would have to be tweaked to show that you need to use one or the other based on your situation). 

A *third* solution, which isn't possible right now, would be to allow users to have an extra key in their `controller.json` file that would point to which specific file (e.g. `dist/controller-v3.js`) in the package that should be imported. Then we could keep the old one (but deprecate it), and make sure that this new key (e.g. `file: 'dist/controller-v3.js'`) is included in new installs.

OR, hopefully I'm missing some cool idea :)

gbere:
> Yo @gbere!
> 
> Thanks for this! I think this is an important one to get done. It's also the first time that we've hit this situation (though, I think it'll be fairly common): where we are bumping the dependency of a JavaScript library a full version.
> 
> In a perfect world, we would have 2 controllers (for awhile at least) - a Chartjs v2 version and a Chartjs v3 version - and automatically (at build time) "toggle" between them based on which version of chartjs is installed. This is easy in theory (it's just an if statement), but Webpack doesn't like conditional imports like this :). You CAN do something like this (taken from React):
> 
> ```js
> if (process.env.NODE_ENV === 'production') {
>   module.exports = require('./cjs/react.production.min.js');
> } else {
>   module.exports = require('./cjs/react.development.js');
> }
> ```
> 
> But I don't believe you can add a more complex conditional that checks what version of Chart.js is installed (at least, we can't do this without requiring the user to add some extra fanciness to their webpack config, which we can't do).
> 
> So, another solution would be to make 2 controllers (for v2 chartjs and v3) and then make the user toggle between them. That would be done by adding a 2nd controller here -
> 
> https://github.com/symfony/ux/blob/351538011587680660c1fe0b09d4038200abad6b/src/Chartjs/Resources/assets/package.json#L8-L13
> 
> - probably called `chartv3`. Then, to avoid the user having BOTH activated when you install the package, you would set the original controller to `enabled: false`.
> The downside of this would be that, if you use `chartv3`, then that is literally the new name of the controller (so the documentation would have to be tweaked to show that you need to use one or the other based on your situation).
> 
> A _third_ solution, which isn't possible right now, would be to allow users to have an extra key in their `controller.json` file that would point to which specific file (e.g. `dist/controller-v3.js`) in the package that should be imported. Then we could keep the old one (but deprecate it), and make sure that this new key (e.g. `file: 'dist/controller-v3.js'`) is included in new installs.
> 
> OR, hopefully I'm missing some cool idea :)

I think the best way to switch between major versions of chartjs dependency is increase the major version of symfony / ux-chartjs. But seeing that all ux- packages share version, I understand that they have another usage.

Any suggestions how to pass the tests?

weaverryan:
@gbere Yea, this is really tricky.

After thinking about this more, I think we should attempt to combine solutions 1 & 3.

This would require work in https://github.com/symfony/stimulus-bridge. The general idea is this:

A) That library contains a webpack "loader" that's responsible for parsing the `controllers.json` file and then generating a "JavaScript source string" that contains `import` statements for all the files. Most of that work is done in this file: https://github.com/symfony/stimulus-bridge/blob/ed4434c4a80a603ada2fb648c8d2aa8924df7565/src/webpack/create-controllers-module.js#L14 - this line specifically reads the "main" key from the package's `package.json` file to complete the filename that should be imported - https://github.com/symfony/stimulus-bridge/blob/ed4434c4a80a603ada2fb648c8d2aa8924df7565/src/webpack/create-controllers-module.js#L59

B) I think we need add a "hook" in this process where an end-user library could have a module that exports a function that the above code calls. The purpose of this function would be to dynamically resolve this file should ultimately be imported.

For example, let's think about chart.js. It currently has a `package.json` file that specifies exactly which one file to load when the user's `controllers.json` file points to it: https://github.com/symfony/ux/blob/main/src/Chartjs/Resources/assets/package.json#L9

Instead of having a `main` key, we could optionally allow packages to have a `resolveMain` key, which points to a file in the package (e.g. `dist/resolve_main.js`). This would export a function that would return the file to import - e.g.:

```js
export default () => {
    return 'dist/controller.js';
}
```

Of course, if your use-case is this simple, then you would continue to specify the `main` key and not create this function. BUT, you could then include conditional code in this callback:

```js
export default () => {
    let chartVersion = null;
    try {
        chartMajorVersion = Math.floor(require(`chart.js/package.json`).version);
    } catch (e) {
        throw new Error('Please make sure that chartjs is installed. Run yarn ...');
    }

    switch (chartMajorVersion) {
        case 2:
            return 'dist/controller.js';
        case 3:
            return 'dist/controller-v3.js';
        default:
            throw new Error(`chart.js version ${chartMajorVersion} is not supported`);
    }
}
```

This would allow the library - ux-chatjs - to intelligently toggle which of the 2 controllers to import without extra config needed from the user (since it detects which version is available).

Huluti:
Hi! For information chart.js is now at version 3.6.1. Can't wait to enjoy it! Thank you for your work :)

weaverryan:
I've rebased (after all of our build changes) over on #185

Issue #120, opened 2021-07-05
Proposal: Put back the Annotations and PHP7 compatibility


gesof:
I noticed that PHP7 support was dropped a few days ago (and annotations support) in favor of PHP8 attributes. In the long run it is a good idea but it will take us a while for transitioning to PHP8 and therefore testing the new UX components will not benefit of community feedback as much as it would if PHP7 would still be an option.

There is a huge trend of pushing the business logic toward Javascript and client-side programming which is somehow abnormal in my opinion. One of the reasons is the lack of alternatives and the UX component is so promising to fill in the gap. But if it takes 1-2 years to become a standard I'm afraid we'll all loose a big opportunity to keep the web programming sane.

nicolas-grekas:
This choice was made in order to spend our time on building the actual thing rather than making it work with phpdoc. There are challenges to add support for php7 that are not worth this time.

Said another way, if you want to see this happen, the best and maybe only way would be to send a PR.

Up to giving it a try?

gesof:
Well, the support for PHP7 was there already last week. Maintaining it may require additional work but the benefits reflect into the present by immediate adoption.

weaverryan:
> Well, the support for PHP7 was there already last week. Maintaining it may require additional work but the benefits reflect into the present by immediate adoption.

The support *was* there, until we replaced the "wiring guts" of the components. So, in reality, what happened is that we had a system that supported PHP 7, then we totally replaced it with a new system (and decided not to add PHP 8 support then). So, it *is* kind of a thing that needs to be re-added.

However, I would be very happy if someone added it :). Probably we would need to extract the logic that reads things from the PHP 8 attributes (search the codebase for `AsTwigComponent` and `AsLiveComponent`) into a separate service. Then, we could make two "versions" of that service: one that reads attributes and one that reads from annotations. A config option would swap between the 2 services.

Note: if if we did the above, users would still (unless we did more work) need to manually tag their components with `twig.component`, but that's fairly minor :).

gesof:
Will Symfony 6 have support for annotations, like routing and mapping ?

wouterj:
I think existing readers will not be removed, but none of the newer attributes are implemented as annotations.

I think the gist here is: it is doable, but we don't want to invest time in this as we don't think we have a worthy usecase. However, this is open source, so we are happy to work together **with** someone that has a use-case and time to implement it.

gesof:
Using PHP8 attributes like this just feel wrong for two reasons:
[...]

PR #127, opened 2021-07-23
Configure Twig/Live Component via service config (remove `As*Component` requirement)

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Related to discussion in #120
| License       | MIT

As discussed in #120, twig/live components can now be completely configured via service tag attributes. `AsTwigComponent`/`AsLiveComponent` are just used for auto-configuration.


src/LiveComponent/src/Twig/LiveComponentRuntime.php
```diff
@@ -25,19 +25,21 @@
...
         $this->hydrator = $hydrator;
+        $this->factory = $factory;
         $this->urlGenerator = $urlGenerator;
         $this->csrfTokenManager = $csrfTokenManager;
     }
 
     public function renderLiveAttributes(Environment $env, object $component): string
     {
-        $url = $this->urlGenerator->generate('live_component', ['component' => AsLiveComponent::forClass($component::class)->getName()]);
+        $url = $this->urlGenerator->generate('live_component', ['component' => $this->nameFor($component)]);
```
kbond, line 42: The one thing that is unfortunate is, right now, _live components_ can only be registered _once per class_. If you have `MarkdownComponent` registered twice as `markdown1` and `markdown2`. When using these twig functions, only the _first one found_'s config will be used. This is because these functions need to look up the _component name_ from the class (as only the component's object is passed).

One solution could be to allow these twig functions to have a second, optional `$name` argument. If passed, it is used. I'm open to other ideas!

kbond: As an example of where this could be a problem: say you used a 3rd party bundle that registered a `MarkdownComponent` with name/key `markdown` and template `path/to/some/template.html.twig`. You need an _additional implementation_ where you customize the template so you register it manually (same class but with different service name) with name/key `app_markdown` and template `components/app_markdown.html.twig`. When using these twig functions, the bundle's config will always be loaded for this component.

If you completely override the service, _I believe_ you'd be ok - this would only be a problem if you need both implementations...

So maybe an obscure issue that no one will ever come across?

weaverryan: I understand the issue. I can't think of any other workaround than allowing a `name` option to be specified. It's pretty simple: if we pass a component object to the Twig function... and that object could potentially have 2 different configs, there's just no way for us to choose the correct one without additional info.

If adding a `$name` argument isn't too much trouble, it would be nice to be pre-planning for this edge-case.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -42,12 +41,19 @@ final class LiveComponentExtension extends Extension
...
         if (method_exists($container, 'registerAttributeForAutoconfiguration')) {
-            $container->registerAttributeForAutoconfiguration(AsLiveComponent::class, function (ChildDefinition $definition) {
-                $definition
-                    ->addTag('twig.component')
-                    ->addTag('controller.service_arguments')
-                ;
-            });
+            $container->registerAttributeForAutoconfiguration(
+                AsLiveComponent::class,
+                function (ChildDefinition $definition, AsLiveComponent $attribute) {
```
weaverryan, line 46: That 2nd argument is pretty handy - this looks great

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -76,18 +75,28 @@ public function onKernelRequest(RequestEvent $event): void
...
+            $config = $this->container->get(ComponentFactory::class)->configFor($componentName);
+        } catch (\InvalidArgumentException $e) {
+            throw new NotFoundHttpException(sprintf('Component "%s" not found.', $componentName), $e);
         }
 
-        [$componentServiceId, $componentClass] = $this->componentServiceMap[$componentName];
+        $request->attributes->set('_component_template', $config['template']);
+
+        $componentServiceId = $config['service_id'];
+        $componentClass = $config['class'];
```
weaverryan, line 87: nit pick: we don't need either of these until we're inside of the `if` below

weaverryan (commented):
Given that this is the agreed-upon approach we want for the low-level wiring details, this looks good to me - other than a couple very minor things :)

src/LiveComponent/src/Twig/LiveComponentRuntime.php
kbond: I added this. I opted to not document as I believe this is a very obscure edge case but let me know if you'd like documentation now.

My solution is a bit ugly - a lot of logic added to `ComponentFactory::configFor()` that isn't really necessary for the twig component itself. A possible improvement could be to just have `ComponentFactory::getConfig()` that returns the entire config array and let places that require it add their own error checking. Not sure if it's worth it though.

weaverryan (approved):

weaverryan:
Thanks Kevin!

PR #143, opened 2021-09-28
[LiveComponent]: Adding docs for two problematic cases

### LINKED ISSUE #139: Live Component - Empty form data
Hi,

I am trying to make a Todo list with the live components.

I am using the form way described in the Live Component documentation.
The AJAX query runs when I click on the action button but it sends an empty `text` field.

- Here is my Task entity: https://github.com/sandoche/Symfony-ux-live-component-todo-list/blob/main/src/Entity/Task.php
- The create task component: https://github.com/sandoche/Symfony-ux-live-component-todo-list/blob/main/src/Components/CreateTaskForm.php
- The related view: https://github.com/sandoche/Symfony-ux-live-component-todo-list/blob/main/templates/components/create_task_form.html.twig
The full repo: https://github.com/sandoche/Symfony-ux-live-component-todo-list/

As you can see below the task['test'] is empty even though I filled the form:
image

Also it's not well documented I had to use `dehydrateWith` maybe I haven't done correctly.
I am not very sure if it's a bug or if I did something wrong.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #139
| License       | MIT


src/LiveComponent/README.md
```diff
@@ -793,6 +793,45 @@ This is possible thanks to a few interesting pieces:
...
+yet, that's not possible.
+
+The solution is to pass `null` into your component instead of a
+non-persisted entity object. If you need to, you can re-create
+your `new Post()` inside of your component:
+
+```diff
+{{ component('post_form', {
+-    post: post,
++    post: post.id ? post : null,
```
Kocal, line 830:
```suggestion
+    post: post ? post.id : null,
```

no?

PR #155, opened 2021-10-29
[TwigComponent] - Dynamic template

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix https://github.com/symfony/ux-twig-component/pull/1
| License       | MIT

Hi,

I add this PR which has already been added here (https://github.com/symfony/ux-twig-component/pull/1) and I modified it a bit.

The idea is to have a dynamic template:

```php
<?php

namespace App\Components;

#[AsTwigComponent('alert')]
class AlertComponent
{
    public string $message;
    
    public function getTemplate() : ?string
    {
        return 'components/alert-'.$this->type.'.html.twig';
    }
}

```

### DIFF
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -30,6 +30,7 @@ final class ComponentRenderer
     public function render(object $component, string $template): string
     {
         // TODO: Self-Rendering components?
-        return $this->twig->render($template, ['this' => $component]);
+        $templateName = $component->getTemplate() ?? $template;     
+        return $this->twig->render($templateName, ['this' => $component]);
     }
 }
--- a/src/TwigComponent/src/Attribute/AsTwigComponent.php
+++ b/src/TwigComponent/src/Attribute/AsTwigComponent.php
@@ -27,4 +27,9 @@ class AsTwigComponent
         $this->name = $name;
         $this->template = $template;
     }
+
+    public function getTemplate(): ?string
+    {
+        return null;
+    }
 }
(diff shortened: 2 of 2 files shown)


weaverryan:
Hey @seb-jean!

Thanks for bringing this over here! Does this work yet? I would expect that this would be configured on the attribute itself - e.g. `#[AsTwigComponent('alert'), template='templates/foo.html.twig']`

Cheers!

seb-jean:
@weaverryan Yes PR works well.

seb-jean:
In fact, I don't see how to get:
`#[AsTwigComponent('alert'), template='components/alert-{type}.html.twig']`

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -30,6 +30,7 @@ public function __construct(Environment $twig)
     public function render(object $component, string $template): string
     {
         // TODO: Self-Rendering components?
-        return $this->twig->render($template, ['this' => $component]);
+        $templateName = $component->getTemplate() ?? $template;
```
weaverryan, line 33: I don't understand how this works (I could be totally missing something). The `$component` here is your component instance. But your component does not use or extend `AsTwigComponent`: that's an attribute on your component.

seb-jean: I didn't need to put `function getTemplate()` in `AsTwigComponent`. It works without.

seb-jean:
Yes that might solve the problem :).

kbond:
@seb-jean, are you up to modify this PR to a self-rendering implementation?

seb-jean:
@kbond , no, I'm sorry, I don't see what to do next.

kbond:
No worries, I'll create a PR when I have a chance.

kbond:
@seb-jean what about https://github.com/symfony/ux/pull/166#issuecomment-970893084 as a solution?

seb-jean:
@kbond, the solution works very well.
Thank you so much :pray:.

Issue #133, opened 2021-08-24
[LiveComponent] How to use a DTO instead of an entity


benr77:
I've just been playing around with the Live Component functionality. It's basically totally awesome, but I got completely stuck when trying to use a form data class instead of an entity.

I see in #102 there is a note about `Document using a DTO is a LiveProp / the dehydration system.`. This is exactly what I'm stuck on trying to use a DTO rather than an entity, and getting an error message saying something like "dehydration not available".

Can anyone point me in the right direction? If I can get it working I will do a PR to update the readme with the new info. Thanks

weaverryan:
I think the key is that you will need a custom dehydrator for your DTO. Well, it depends, there are two paths:

1) When creating your component, you pass in some raw data  - e.g. `{ name: 'ryan', isNamedRyan: true }`. Then, inside your component, you use this data to create the DTO inside of `instantiateForm()` and pass it directly to your form. In this situation, your DTO does not need to be set onto a property and dehydrated/hydrated.

2) If this doesn't work (because you want to be able to pass the already-instantiated DTO into your component), then you'll need a custom dehydrator for it. You can use these 2 options inside `LiveProp` to set a custom hydrator/dehydrator to a method on your component - https://github.com/symfony/ux/blob/4e7cd8cc976968b1379f36da0543db63fed4933d/src/LiveComponent/src/Attribute/LiveProp.php#L25-L27 - I think these options are not-yet-documented.

kbond:
With #162, I believe a custom hydrator is no longer required.

benr77:
Confirmed. This is now working with the latest commits from #162

PR #162, opened 2021-11-08
[LiveComponent] use "full" serializer for NormalizerBridgePropertyHydrator

### LINKED ISSUE #152: symfony-ux-live-component: hydrating via the NormalizerBridgePropertyHydrator
While trying out symfony-ux-live-component (which is awesome btw; thank you for creating this!), I stumbled upon an issue.
The project I am experimenting with, uses a lot of value objects, and I have a decent amount of normalizers in place to normalize these value objects. Since a class `NormalizerBridgePropertyHydrator` exists, I thought using my value objects with the live components might work.

When I tried that, I got these kinds of error messages:

```
An exception has been thrown during the rendering of a template ("Cannot dehydrate property "playerIdentifier" of "App\Components\PlayerSwitchComponent". The value "App\Domain\ValueObject\Player\PlayerIdentifier" does not have a dehydrator.").
```

With the help of XDebug, I noticed that the `NormalizerBridgePropertyHydrator` was trying to use Symfony's `DateTimeNormalizer` to normalize/denormalize my value objects, which obviously failed.

Now I have managed to fix this by adding these lines to the `services.yml` file of my project:
```
    Symfony\UX\LiveComponent\PropertyHydratorInterface:
        class: Symfony\UX\LiveComponent\Hydrator\NormalizerBridgePropertyHydrator
        arguments:
            - '@Symfony\Component\Serializer\Normalizer\NormalizerInterface'
```

I'm not completely sure this is how this hydrator is intended to be used. But it seems to work for me.

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #152, #133
| License       | MIT

This injects the entire serializer into `NormalizerBridgePropertyHydrator`. Previously, it was just the `DateTimeNormalizer`. We originally decided it would be best to "pick and choose" which normalizers we would enable by default. The reasoning was we wanted to prevent the `DoctrineEntityPropertyHydrator` being called when hydrating/dehydrating every live prop. This hydrator requires checking with doctrine if the property's type-hint is a doctrine entity. I believe we were concerned about the performance overhead. Maybe this is negligible? Maybe we can adjust this hydrator to be more performant?

### DIFF
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -62,12 +62,12 @@ final class LiveComponentExtension extends Extension
 
         $container->register('ux.live_component.doctrine_entity_property_hydrator', DoctrineEntityPropertyHydrator::class)
             ->setArguments([[new Reference('doctrine')]])
-            ->addTag('twig.component.property_hydrator', ['priority' => -200])
+            ->addTag('twig.component.property_hydrator', ['priority' => -100])
         ;
 
-        $container->register('ux.live_component.datetime_property_hydrator', NormalizerBridgePropertyHydrator::class)
-            ->setArguments([new Reference('serializer.normalizer.datetime')])
-            ->addTag('twig.component.property_hydrator', ['priority' => -100])
+        $container->register('ux.live_component.serializer_property_hydrator', NormalizerBridgePropertyHydrator::class)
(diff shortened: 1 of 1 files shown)


weaverryan (approved):
makes sense - if there are performance issues, we can find and worry about them later :)

weaverryan:
Thank you Kevin!

PR #165, opened 2021-11-16
[LiveComponent] fix silent serializer/doctrine dependency issue

### LINKED ISSUE #164: [LiveComponent] Remove silent dependency on Doctrine
To reproduce: in a project where Doctrine is not installed, run:

```
composer require symfony/ux-live-component:@dev
```

Result:

> The service "ux.live_component.doctrine_entity_property_hydrator" has a dependency
> on a non-existent service "doctrine". Did you mean this: "ux.live_component.doctrine_entity_property_hydrator"?                                 

That's true! This service should be optional and not registered if Doctrine is not present.

**UPDATE**

We have the same problem if `serializer` is not installed - with the service `ux.live_component.serializer_property_hydrator`.

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #164
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -22,8 +22,6 @@ use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\LiveComponent\ComponentValidator;
 use Symfony\UX\LiveComponent\ComponentValidatorInterface;
 use Symfony\UX\LiveComponent\EventListener\LiveComponentSubscriber;
-use Symfony\UX\LiveComponent\Hydrator\DoctrineEntityPropertyHydrator;
-use Symfony\UX\LiveComponent\Hydrator\NormalizerBridgePropertyHydrator;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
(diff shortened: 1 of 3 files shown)


src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -60,15 +61,19 @@ function (ChildDefinition $definition, AsLiveComponent $attribute) {
...
+                ->setArguments([[new Reference('doctrine')]])
+                ->addTag('twig.component.property_hydrator', ['priority' => -100])
+            ;
+        }
 
-        $container->register('ux.live_component.serializer_property_hydrator', NormalizerBridgePropertyHydrator::class)
-            ->setArguments([new Reference('serializer')])
-            ->addTag('twig.component.property_hydrator', ['priority' => -200])
-        ;
+        if (class_exists(Serializer::class)) {
```
kbond, line 71: This will still be an issue if the serializer is installed but not enabled. Is there a way to make this check without a compiler pass?

weaverryan: Not that I'm aware of. I think we should just "bite" and move both of these to a compiler pass.

weaverryan (approved):

PR #166, opened 2021-11-16
[TwigComponent] add Self-Rendering Components

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Alternative to #155
| License       | MIT

Gives the ability to assume full control of the component rendering.


src/TwigComponent/src/SelfRenderingInterface.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Symfony\UX\TwigComponent;
+
+use Twig\Environment;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+interface SelfRenderingInterface
+{
+    public function render(Environment $twig): string;
```
kbond, line 12: maybe `renderComponent` is better to avoid possible conflicts?

src/TwigComponent/src/SelfRenderingInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Symfony\UX\TwigComponent;
+
+use Twig\Environment;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+interface SelfRenderingInterface
```
yceruto, line 10: `RenderableInterface` ?

src/TwigComponent/README.md
```diff
@@ -338,6 +338,44 @@ If you're using [Live Components](https://github.com/symfony/ux-live-component),
...
+
+    public function mount(bool $isSuccess = true)
+    {
+        $this->type = $isSuccess ? 'success' : 'danger';
+    }
+
+    public function render(\Twig\Environment $twig) : string
+    {
+        $template = 'components/alert-'.$this->type.'.html.twig';
+        $template = $this->getLoader()->exists($template) ? $template : 'components/alert-default.html.twig';
```
yceruto, line 367: Maybe this line is not completely necessary for this example and we can assume both templates `'components/alert-'.$this->type.'.html.twig'` already exist in our side.

kbond:
I'm sort of second guessing this feature. Does it give us something we can't do already. For instance, my readme example could be achieved with the following:

```twig
{# templates/components/alert.html.twig #}

{{ include(['components/alert-'~this.type~'.html.twig', 'components/alert-default.html.twig'], {this: this, something: 'else'}) }}
```

src/TwigComponent/README.md
kbond: I was trying to demonstrate a complex example where we'd need this feature but I think this isn't it.

kbond:
I'm withdrawing this PR for now.

PR #171, opened 2021-11-25
[Turbo] Update @hotwired/turbo - v7.1.0

seb-jean:
@hotwired/turbo 7.1.0 is released:
https://github.com/hotwired/turbo/releases/tag/v7.1.0


weaverryan:
I think we should keep it at 7.0 because we don't "need" the user to be using 7.1

tgalopin:
I agree with Ryan, 7.0 is the minimal version, we always recommend using the latest minor if possible.

PR #173, opened 2021-11-26
Changing branch alias to 2.0

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yrs
| New feature?  | no
| Tickets       | none
| License       | MIT

Since we're going to skip 1.4 and go straight to 2.0, it is the next release.

Buuuut, a better solution is to create 1.x and 2.x branches. Then I believe we can remove this completely - e.g. Flex https://github.com/symfony/flex/commit/375e01daedd481501c29f3dea443cf885858382f#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34

### DIFF
--- a/src/Chartjs/composer.json
+++ b/src/Chartjs/composer.json
@@ -42,7 +42,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-main": "1.4-dev"
+            "dev-main": "2.0-dev"
         },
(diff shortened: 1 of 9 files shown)


weaverryan:
Renaming branches to 1.x and 2.x instead

PR #183, opened 2021-12-02
Removing branch-alias

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | None
| License       | MIT

We now have named branches, so no need to have branch-alias, I beileve.

### DIFF
--- a/src/Chartjs/composer.json
+++ b/src/Chartjs/composer.json
@@ -41,9 +41,6 @@
         "symfony/flex": "<1.13"
     },
     "extra": {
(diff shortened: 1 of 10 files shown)


tgalopin:
@nicolas-grekas I trust your POV on this: I think the same as Ryan but could we have side effects perhaps?

weaverryan:
I got the "ok" from Nicolas on Slack about this.

PR #185, opened 2021-12-02
Adding support for Chartjs 3

### LINKED ISSUE #86: [Chart.js] v3 support

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fixes #86
| License       | MIT

This takes the work from #126 and rebases so we can finish on the 2.x branch.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -4,6 +4,39 @@ import glob from 'glob';
 import path from 'path';
 import pkgUp from 'pkg-up';
 
+/**
+ * Guarantees that any files imported from a peer dependency are treated as an external.
+ *
+ * For example, if we import `chart.js/auto`, that would not normally
+ * match the "chart.js" we pass to the "externals" config. This plugin
+ * catches that case and adds it as an external.
+ *
+ * Inspired by https://github.com/oat-sa/rollup-plugin-wildcard-external
+ */
+const wildcardExternalsPlugin = (peerDependencies) => ({
+    name: 'wildcard-externals',
+    resolveId(source, importer) {
+        if (importer) {
+            let matchesExternal = false;
+            peerDependencies.forEach((peerDependency) => {
+                if (source.includes(`/${peerDependency}/`)) {
+                    matchesExternal = true;
+                }
+            });
+
+            if (matchesExternal) {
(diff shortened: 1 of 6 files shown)


tgalopin:
There seems to be a missing browser feature required by Chartjs 3 (ResizeObserver), perhaps would it be necessary to add a new polyfill.

weaverryan:
> There seems to be a missing browser feature required by Chartjs 3 (ResizeObserver), perhaps would it be necessary to add a new polyfill.

You're right - it WAS missing in the test environment. I polyfilled it there. But I don't think we need to polyfill it for the users, if that's what you meant (let me know if that IS what you were thinking and let's chat).

I have just one more tweak on this one - it's in the "externals" of the built file (if you build the file right now, the dist file will "contain" chartjs instead of just having the `import line`). I know the fix - I just need to make a nice addition to the build system for it.

weaverryan:
This is now ready!

tgalopin:
Could you squash your commits? I can't do it automatically due to two authors :)

tgalopin (approved):

Huluti:
ChartJS is now at v3.6.2. Is it something to change in this PR since it's based on v3.4.1?

weaverryan:
The 3.4.1 only means that it's the minimum version. In reality, if you installed this, you would have `^3.4.1` in your package.json, but yarn/npm would install the latest version 3 (so 3.6.2). It's really not a big deal one way or another, but I'd say just keep it.

I've got it rebased to only 2 commits - that's how (I think) we should keep it.

tgalopin:
Thanks @weaverryan.

PR #187, opened 2021-12-03
Replacing deprecated initCustomEvent() with modern method

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | none
| License       | MIT

The old method is deprecated - https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/initCustomEvent - we were already using the new way in live component.

Cheers!

### DIFF
--- a/src/Chartjs/Resources/assets/dist/controller.js
+++ b/src/Chartjs/Resources/assets/dist/controller.js
@@ -18,10 +18,8 @@ class default_1 extends Controller {
         const chart = new Chart(canvasContext, payload);
         this._dispatchEvent('chartjs:connect', { chart });
     }
-    _dispatchEvent(name, payload = null, canBubble = false, cancelable = false) {
-        const userEvent = document.createEvent('CustomEvent');
-        userEvent.initCustomEvent(name, canBubble, cancelable, payload);
-        this.element.dispatchEvent(userEvent);
+    _dispatchEvent(name, payload) {
+        this.element.dispatchEvent(new CustomEvent(name, { detail: payload }));
     }
 }
 default_1.values = {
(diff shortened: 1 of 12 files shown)


src/Chartjs/Resources/assets/dist/controller.js
```diff
@@ -1,16 +1,12 @@
...
     connect() {
         if (!(this.element instanceof HTMLCanvasElement)) {
             throw new Error('Invalid element');
         }
-        const viewData = this.element.getAttribute('data-view');
-        if (!viewData) {
-            throw new Error('Missing data-view attribute.');
-        }
-        const payload = JSON.parse(viewData);
+        const payload = this.viewValue;
```
weaverryan: Built file has a few extra changes because I think we didn't rebuild after the chartjs -> values PR :)

src/Chartjs/Resources/assets/src/controller.ts
```diff
@@ -38,9 +38,10 @@ export default class extends Controller {
         this._dispatchEvent('chartjs:connect', { chart });
     }
 
-    _dispatchEvent(name: string, payload: any = null, canBubble = false, cancelable = false) {
-        const userEvent = document.createEvent('CustomEvent');
-        userEvent.initCustomEvent(name, canBubble, cancelable, payload);
+    _dispatchEvent(name: string, payload: any) {
+        const userEvent = new CustomEvent(name, {
```
tgalopin: I guess we don't need this structure anymore do we? Perhaps could we replace all `this._dispatchEvent(...)` calls by `this.element.dispatchEvent(new CustomEvent(...));` directly?

tgalopin:
Thanks @weaverryan.

PR #189, opened 2021-12-03
[TwigComponent] add `PreMount` hook

### LINKED ISSUE #142: [Twig Component] Feature: add possibility to validate all incoming data to component
I want to validate all data passed from twig into my php component. Right now I can do such thing only by creating a setter for each property. Eventually use `mount` method and explicitly define there each property as an argument. 

What I need is more generic method to validate such data. I was thinking about 2 solutions.
1. Built-in method like `mount` named `configureProperties`
2. A callback to some method inside my component.

As I prefer the second solution I wrote some code to show you the case using the first one (was easier to me).

```php
//ComponentFactory.php

public function create(string $name, array $data = []): object
    {
        $component = $this->getComponent($name);

        $this->mount($component, $data);
        $this->configureProperties($component, $data); // <- add this line

        foreach ($data as $property => $value) {
            if (!$this->propertyAccessor->isWritable($component, $property)) {
                throw new \LogicException(sprintf('Unable to write "%s" to component "%s". Make sure this is a writable property or create a mount() with a $%s argument.', $property, \get_class($component), $property));
            }

            $this->propertyAccessor->setValue($component, $property, $value);
        }

        return $component;
    }

// add these methods

private function configureProperties(object $component, array &$data): void
    {
        try {
            $method = (new \ReflectionClass($component))->getMethod('configureProperties');
        } catch (\ReflectionException $e) {
            // no hydrate method
            return;
        }
        $data = array_merge($data, $this->getComponentPropertiesDefaultValues($component));
        $optionsResolver = new OptionsResolver();
        $component->configureProperties($optionsResolver);
        $data = $optionsResolver->resolve($data);
    }

    private function getComponentPropertiesDefaultValues(object $component): array
    {
        $reflection = new \ReflectionClass($component);
        $propertiesReflections = array_filter(
            $reflection->getProperties(\ReflectionProperty::IS_PUBLIC),
            static fn($property): bool => !$property->isStatic() && $property->getDefaultValue() !== null,
        );

        $properties = [];

        array_map(static function ($property) use (&$properties) {
            $properties[$property->name] = $property->getDefaultValue();
        }, $propertiesReflections);

        return $properties;
    }
```

```php
// Components/CustomComponent.php

final class CustomComponent
{
    public string $name;

    public string $size = 'm';

    public string $align;

    public function configureProperties(OptionsResolver $optionsResolver)
    {
        $optionsResolver
            ->setDefined(['size'])
            ->setDefaults(['align' => 'center'])
            ->setAllowedValues('size', ['xs', 's', 'm', 'l', 'xl',])
            ->setAllowedValues('align', ['left', 'center', 'right'])
            ->setRequired(['name']);

    }
```

```twig
#components/custom_component.html.twig
<div class="size-{{ this.size }} align-{{ this.align }}">
    <h1>{{ this.name }}</h1>
    <p>Size: {{ this.size }}</p>
    <p>Align: {{ this.align }}</p>
</div>
```

```twig
{{ component('custom_component', {name: 'some name'}) }}
```

`configureProperties` method in `ComponentFactory` is responsible for setting default values defined in component's class properties into `$data` variable. Then it tries to resolve provided options using my custom components rules and allows TwigComponent to set validated data.

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #142
| License       | MIT

### DIFF
--- a/src/TwigComponent/README.md
+++ b/src/TwigComponent/README.md
@@ -217,6 +217,40 @@ If an option name matches an argument name in `mount()`, the
 option is passed as that argument and the component system
 will _not_ try to set it directly on a property.
 
+### PreMount Hook
+
+If you need to modify/validate data before it's _mounted_ on the
+component use a `PreMount` hook:
+
+```php
+// src/Components/AlertComponent.php
+
+use Symfony\UX\TwigComponent\Attribute\PreMount;
+// ...
+
+#[AsTwigComponent('alert')]
+class AlertComponent
+{
+    public string $message;
+    public string $type = 'success';
+
+    #[PreMount]
+    public function preMount(array $data): array
+    {
+        // validate data
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults(['type' => 'success']);
+        $resolver->setAllowedValues('type', ['success', 'danger']);
+        $resolver->setRequired('message');
+        $resolver->setAllowedTypes('message', 'string');
+
+        return $resolver->resolve($data)
+    }
+
+    // ...
+}
+```
+
 ## Fetching Services
 
 Let's create a more complex example: a "featured products" component.
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -100,20 +100,6 @@ final class AsLiveComponent extends AsTwigComponent
     }
 
     /**
-     * @param string|object $classOrObject
-     *
-     * @return \ReflectionMethod[]
-     */
-    private static function attributeMethodsFor(string $attribute, object $component): \Traversable
-    {
-        foreach ((new \ReflectionClass($component))->getMethods(\ReflectionMethod::IS_PUBLIC) as $method) {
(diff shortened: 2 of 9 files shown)


src/TwigComponent/README.md
```diff
@@ -217,6 +217,39 @@ If an option name matches an argument name in `mount()`, the
...
+use Symfony\UX\TwigComponent\Attribute\PreMount;
+// ...
+
+#[AsTwigComponent('alert')]
+class AlertComponent
+{
+    public string $message;
+    public string $type = 'success';
+
+    public function preMount(array $data): array
```
weaverryan, line 238: This needs the `PreMount` attribute, correct?

weaverryan (commented):
Very cool - just noticed a README thing!

src/TwigComponent/README.md
kbond: Fixed!

weaverryan (approved):

weaverryan:
Thanks Kevin!

PR #199, opened 2021-12-10
Fix Turbo Mercure bridge dependency

### LINKED ISSUE #198: Wrong dependency on `symfony/ux-turbo-mercure`
When following upgrade guide I stumbled upon dependency conflict within `ux-turbo-mercure`:

```
Problem 1
    - Root composer.json requires symfony/ux-turbo-mercure ^2.0 -> satisfiable by symfony/ux-turbo-mercure[v2.0.0].
    - symfony/ux-turbo-mercure v2.0.0 requires symfony/ux-turbo ^1.3 -> found symfony/ux-turbo[v1.3.0] but it conflicts with your root composer.json require (^2.0).
```

Looks like `ux-turbo-mercure` needs dep bump on `ux-turbo` to `^2.0` that was missed.

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix https://github.com/symfony/ux/issues/198
| License       | MIT

I think the package should be able to handle both 1.x and 2.x without change. @weaverryan I guess this will require a dedicated tag right?

### DIFF
--- a/src/Turbo/Bridge/Mercure/composer.json
+++ b/src/Turbo/Bridge/Mercure/composer.json
@@ -29,7 +29,7 @@
         "php": ">=7.2.5",
         "symfony/mercure-bundle": "^0.3",
         "symfony/twig-bundle": "^5.2|^6.0",
-        "symfony/ux-turbo": "^1.3"
+        "symfony/ux-turbo": "^2.0"
     },
     "extra": {
         "thanks": {


weaverryan:
I think should only allow 2.0. Otherwise you will have one lib trying to use stimulus and another using @hotwired/stimulus is you have a mix of 1.0 and 2.0

@kbond this is the issue you were talking about (we saw this also with twig/live components) - our test suite needs to be enhanced to catch these :)

weaverryan (approved):

weaverryan:
Thanks Tito!

PR #200, opened 2021-12-10
Adjust live-component CI to use twig-component release

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

This changes the live-component deps in CI to use "real" releases and not a symlink to the directory in this repo. Doing this will better catch BC breaks.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -66,7 +66,6 @@ jobs:
             - name: LiveComponent
               run: |
                   cd src/LiveComponent
-                  php ../../.github/build-packages.php
                   composer update --prefer-lowest --prefer-dist --no-interaction --no-ansi --no-progress
                   php vendor/bin/simple-phpunit
 


.github/workflows/test.yaml
```diff
@@ -66,7 +66,6 @@ jobs:
             - name: LiveComponent
               run: |
                   cd src/LiveComponent
-                  php ../../.github/build-packages.php
```
kbond, line 69: We definitely want this removed for _low-deps_ but what about high?

weaverryan: Hmm. Removing this wouldn't work for high deps if we added a new feature to Twig & Live components at the same time, which use each other. In that case, the Live component would install the stable Twig Component, which wouldn't include the still-unreleased new Twig component feature.

The same could technically be true for low deps... but it's likely much more less common and this would work fine. The edge-case I'm talking about is (same as above) if we added a new feature to Twig + Live components that rely on each other. Then, assuming the current stable release is 2.0, we change LiveComponent to rely on TwigComponent 2.1. That'll fail, because 2.1 isn't released, etc, etc.

This is why the complex solution in symfony/symfony exists. WDYT?

kbond: Ok, so definitely run this script for high deps.

Not sure how to handle your second point. This does seem valid - some of the things we discussed recently (embedded components) would do this.

weaverryan: Yea, the "complex solution" in symfony/symfony looks... complex to me 🙃. But i think it's also the only true solution. Consider #248: you remembered to bump the composer.json file there. But if you had forgotten, the tests would still pass. The "real" solution is to, when testing live-component, to download its real dependencies... except if the version specified in composer.json is the next, unreleased version (e.g. like 2.1, which you've correctly specified). In that case, I think symfony/symfony basically "patches" in the twig-component code from the PR as version "2.1".

kbond:
Closing for now, we can re-evalute when the library's are no longer experimental. Preserving BC will be much more important then.

Issue #195, opened 2021-12-09
[LiveComponent] Add an data-attribute to modify the value of a prop


BeyerJC:
Hey everyone !

I think it would be nice if we had an additional data-attribute to set the value of a prop from elements like <a>. In my specific case, i tried to create a pagination and want to update the prop "page".

Im thinking of something like the data-model attribute --> data-value for example.

The result could be something like this

` <a class="page-link" data-action="live#update" data-model="page" data-value="{{ page }}">{{ page }}</a>`

PR #202, opened 2021-12-13
[LiveComponent] feat: consider data-value attribute

BeyerJC:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       |  #195 
| License       | MIT

Allow to set the value using data-value. This can be handy if the model is updated from an \<a\> or \<button\> to set it to a fixed value for example.


WhiteRabbitDE:
I'm not sure about the failing js tests. I dont think it is caused by my change but i think they worked before so maybe someone with more js experience can have a look at it please ?

PR #203, opened 2021-12-13
[LiveComponent] consider data-value attribute

BeyerJC:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       |  #195 
| License       | MIT

Allow to set the value using data-value. This can be handy if the model is updated from an \<a\> or \<button\> to set it to a fixed value for example.

### DIFF
--- a/src/LiveComponent/README.md
+++ b/src/LiveComponent/README.md
@@ -370,6 +370,27 @@ code works identically to the previous example:
 If an element has _both_ `data-model` and `name` attributes, the
 `data-model` attribute takes precedence.
 
+### Using data-value="" for non input elements
+
+If you want to set the value of a model with an element that is not an input element and does not have a `value` attribute you can use `data-value`
+
+```twig
+<div {{ init_live_component(this)>
+
+    <a
+        data-action="live#update"
+        data-model="min"
+        data-value="10">
+        Set min to 10
+    </a>
+
+    // ...
+</div>
+```
+
+If an element has _both_ `data-value` and `value` attributes, the
+`data-value` attribute takes precedence.
+
 ## Loading States
 
 Often, you'll want to show (or hide) an element while a component is
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1057,11 +1057,11 @@ class default_1 extends Controller {
         window.removeEventListener('beforeunload', this.markAsWindowUnloaded);
     }
     update(event) {
-        const value = event.target.value;
+        const value = this._getValueFromElement(event.target);
         this._updateModelFromElement(event.target, value, true);
     }
     updateDefer(event) {
-        const value = event.target.value;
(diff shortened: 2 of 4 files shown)


weaverryan:
Hey @WhiteRabbitDE!

My apologies for the slow response on this! I like this. Could you (A) add a test and (B) update the documentation? The test would be inside `model.test.ts`.

Thanks!

BeyerJC:
Hallo @weaverryan ,

i just tried to add a test but i have little experience with Jest and how it is used here. Im not sure how to add my case in a proper way. Can you give me a small hint please?

weaverryan:
Absolutely :). 

Here are a few ideas to get you started:

A) This is the template used in the test - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/test/controller/model.test.ts#L19-L32 - it is a little template with just a `name` data. I would add something like `<a data-action="live#update" data-model="name" data-value="Kevin">Change name to Kevin</a>`. Adding this link shouldn't mess up in other tests that use this.

B) Duplicate this test and update its description to make sense - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/test/controller/model.test.ts#L39-L62

C) Inside that test, instead of typing into the input (that's the `userEvent.type` part), you will click the "Change name to Kevin" link. Here's an example of clicking things: https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/test/controller/action.test.ts#L58. Then, for the "assert" part - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/test/controller/model.test.ts#L54-L55 - you will be looking for "Kevin" instead of Ryan Weaver.

D) The last bit you'll need to change is the fetchMock - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/test/controller/model.test.ts#L54-L55 - change both `Ryan Weaver` to `Kevin`. This asserts that the Ajax request was made using this correct data and it basically controls the template re-rendering (which is what allows us to assert that the input box (after re-render) has changed to `Kevin`.

To run the test, move into `src/LiveComponent/assets/` and run `yarn` then `yarn jest`. 

Let me know if this helps!

BeyerJC:
Hey @weaverryan ,
thanks for that super helpful and in detail explaination 🥇  The fetchMock part was what i was missing but now things are much clearer !

I hope i can use that new knowledge in future PRs :-)

src/LiveComponent/README.md
```diff
@@ -370,6 +370,27 @@ code works identically to the previous example:
...
+        Set min to 10
+    </a>
+
+    // ...
+</div>
+```
+
+If an element has _both_ `data-value` and `value` attributes, the
+`data-value` attribute takes precedence.
+
```
weaverryan, line 393: Perfectly explained!

weaverryan (approved):
Beautiful ❤️

weaverryan:
Thank you Jan!

PR #214, opened 2022-01-11
Make turbo stream format subscriber opt-in and not enabled automatically

alexander-schranz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes 
| Tickets       | Fix #... 
| License       | MIT

The subscriber should be opt-in. As it currently manipulates the requestFormat which is not expected by all applications (e.g. Sulu or any other application, working with $request->getRequestFormat() for getting its template). A controller handling the turbo can activate the listener with:

```yaml
my_route:
    defaults:
       turbo: true
```

This indicates this routes  handles turbo route and so request format is set to turbo, for all other routes it behaves the symfony way and the request format is not manipulated in this case.

### DIFF
--- a/src/Turbo/Stream/AddTurboStreamFormatSubscriber.php
+++ b/src/Turbo/Stream/AddTurboStreamFormatSubscriber.php
@@ -28,6 +28,11 @@ final class AddTurboStreamFormatSubscriber implements EventSubscriberInterface
     public function onKernelRequest(RequestEvent $event): void
     {
         $request = $event->getRequest();
+
+        if (!$request->attributes->get('turbo', false)) {
+            return;
+        }
+
         if (!($accept = $request->headers->get('Accept')) || 0 !== strpos($accept, TurboStreamResponse::STREAM_MEDIA_TYPE)) {
             return;
         }


dunglas:
In my opinion, https://github.com/symfony/symfony/pull/44980 is a better fix. It should allow removing this listener entirely.

alexander-schranz:
@dunglas when then:

```
$request->setFormat(TurboStreamResponse::STREAM_FORMAT, TurboStreamResponse::STREAM_MEDIA_TYPE);
$request->setRequestFormat(TurboStreamResponse::STREAM_FORMAT);
```

Can be removed from this listener I'm totally in for that change :)

alexander-schranz:
Tested your changes with sulu (without symfony/ux):

```php
$request->getPreferredFormat(); // turbo_stream
$request->getRequestFormat(); // html
```

This would match my expected behaviour 👍  and would so not longer crash sulu's website controller [here](https://github.com/sulu/sulu/blob/ee6e61de48f20d1514e4f2bf9fc49401a5882000/src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php#L55-L57) for turbo requests.

dunglas:
Maybe should we merge symfony/symfony#44980 as a bug fix in Symfony then, to not have to wait for 6.1 to be released.

alexander-schranz:
@dunglas that would be great :)

chalasr:
Well, symfony/symfony#44980 really is a feature... I'm for merging both PRs. We can remove the listener once this package supports 6.1+ only, as usual.

dunglas:
Actually it's a bug fix for Turbo. The question is: do we merge fixes needed for Symfony UX in the maintenance branches of the monorepo. In my opinion, we should at least in this case as this patch is very very small and cannot break anything.

dunglas:
@alexander-schranz could you try this patch instead, please? https://github.com/symfony/ux/pull/217

nicolas-grekas:
Closing in favor of #217
Thanks for pushing this forward.

PR #215, opened 2022-01-11
[Turbo] Add Vary header for stream responses in docs

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | n/a
| License       | MIT

### DIFF
--- a/src/Turbo/README.md
+++ b/src/Turbo/README.md
@@ -292,6 +292,7 @@ Let's discover how to use Turbo Streams to enhance your [Symfony forms](https://
 namespace App\Controller;
 
 // ...
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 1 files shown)


src/Turbo/README.md
```diff
@@ -300,6 +301,10 @@ use App\Entity\Task;
 
 class TaskController extends AbstractController
 {
+    // The same URL will return a response with a different Content-Type HTTP header
+    // depending on the Accept request header.
+    // To prevent cache pollution issues, we must set the Vary response header.
+    #[Cache(vary: ['Accept'])]
```
nicolas-grekas, line 307: but this is not needed actually, since the call to `getPreferredFormat` is nested in `$form->isSubmitted() && $form->isValid()`, which means POST, which means no-cache.

dunglas: More or less but:

1. Turbo Streams can be used with GET requests too, this adds an example of how to handle content negotiation properly
2. It's a best practice anyway to set Vary regardless of the verb
3. Even if it's not commonly implemented by cache servers, responses to POST requests can be cached (by the spec) and reused for subsequent GET requests for the same resource

nicolas-grekas: Yes, but this example is about a form. The example you added in #217 has `return $response->setVary('Accept');`, which is enough a hint to me.
Here, yes, POST can be cached, but ONLY if Cache-Control is sent. I think that if one does that (set the cc header), then it's their job to also add the Vary, and we don't have to tell everybody to add it.

PR #217, opened 2022-01-11
[Turbo] Simplify stream negotiation

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | n/a
| License       | MIT

This patch replaces the ad-hoc content negotiation system for Turbo Streams used by the bundle. HttpFoundation's native content negotiation features are now used instead. This greatly simplifies the code, which is security-sensitive, and improves DX (content negotiation is now consistent with how it is handled in the rest of the framework).

Replaces https://github.com/symfony/ux/pull/214. Needs https://github.com/symfony/ux/pull/215 and https://github.com/symfony/symfony/pull/44980.


src/Turbo/TurboBundle.php
```diff
@@ -23,6 +25,15 @@
  */
 final class TurboBundle extends Bundle
 {
+    public const STREAM_FORMAT = 'turbo_stream';
+    public const STREAM_MEDIA_TYPE = 'text/vnd.turbo-stream.html';
+
+    public function boot()
+    {
+        // TODO: remove this hack when HttpFoundation 6.1 will be required https://github.com/symfony/symfony/pull/44980
+        (new Request())->setFormat(self::STREAM_FORMAT, self::STREAM_MEDIA_TYPE);
```
dunglas, line 32: This could be nice to allow bundles to register their own media types without having to rely on this hack. I don't know why `Request::setFormat()` isn't marked `static`.

nicolas-grekas: Actually, I'd be in favor of closing #44980 and removing the todo above.
The current API might not be ideal but it works.

src/Turbo/CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # CHANGELOG
 
+## 3.0
```
nicolas-grekas: We should to this in 2.1 because:
- it would make no sense to bump all ux packages to 3.0 
- turbo is experimental, which means that we can do it.

src/Turbo/TurboBundle.php
```diff
@@ -23,6 +24,15 @@
  */
 final class TurboBundle extends Bundle
 {
+    public const STREAM_FORMAT = 'turbo_stream';
+    public const STREAM_MEDIA_TYPE = 'text/vnd.turbo-stream.html';
+
+    public function boot(): void
+    {
+        // TODO: remove this hack when HttpFoundation 6.1 will be required https://github.com/symfony/symfony/pull/44980
```
nicolas-grekas: should be removed to me

nicolas-grekas (changes requested):
I really like that. Here are some comments.

nicolas-grekas (approved):

chalasr (approved):

chalasr:
Thanks Kévin.

src/Turbo/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+        {
+            // ...
+            if (TurboBundle::STREAM_FORMAT === $request->getPreferredFormat()) {
+                $request->setRequestFormat(TurboBundle::STREAM_FORMAT);
+                $response = $this->render('task/success.stream.html.twig', ['task' => $task]);
+            } else {
+                $response = $this->render('task/success.html.twig', ['task' => $task]);
+            }
+
+            return $response->setVary('Accept');
```
alexander-schranz, line 22: 👍

src/Turbo/TurboBundle.php
```diff
@@ -23,6 +24,14 @@
  */
 final class TurboBundle extends Bundle
 {
+    public const STREAM_FORMAT = 'turbo_stream';
+    public const STREAM_MEDIA_TYPE = 'text/vnd.turbo-stream.html';
+
+    public function boot(): void
```
alexander-schranz, line 30: Didn't know about this method. Nice to know something like this exists.

src/Turbo/TurboBundle.php
alexander-schranz: @dunglas maybe we can add a `@static` as phpdoc and make it really static then in symfony 7.

alexander-schranz (commented):
Great work. Thank you 👍

alexander-schranz:
@weaverryan anything blocking here for a release?

florentdestremau:
Indeed, but you still need to specify the request format ? I feel like you still need to "think" about changing something in your code when you switch from a turbo-frame to a turbo-stream. Or did I misread ?

nicolas-grekas:
The call to `$request->setRequestFormat(TurboBundle::STREAM_FORMAT);` will be propagated to the response so that the right content-type is returned.

florentdestremau:
I'm not sure I understand the improvement here :thinking: . If you can either return a turbo-frame (e.g. Form Invalid → 400, or Form display → 200) or a turbo-stream (e.g. Form Submitted Successfully → 302 + 200 or 200) then you can't always set the request format, so you don't really "gain" code length vs extending the TurboStreamRequest object in your `$this->render()` from what I see. I can imagine it helps when you are returning in 100% of the cases a turbo-stream to avoid repeating?
What I would like is for Symfony to auto-detect when a response is a turbo-stream to update the headers. Is this what the PR is doing ?

florentdestremau:
Other point: I'm thankful for you @nicolas-grekas pointing this PR out, because it causes a BC-break in our code: the object TurboStreamResponse is removed in a minor version so I feel like it should have been deprecated instead, no ? For now we downgraded to 2.0.1 until we take the time to update our codebase with 2.1

alexander-schranz:
> The object TurboStreamResponse is removed in a minor version so I feel like it should have been deprecated instead, no ? For now we downgraded to 2.0.1 until we take the time to update our codebase with 2.1

The Symfony/UX component is still marked as [`experimental`](https://github.com/symfony/ux/blob/e4ab138168b481c051cb63b80b6f20d19c5d03c3/src/Turbo/TurboBundle.php#L23) and so can contain bc breaks in minor release. When using a experimental components you should go with e.g.: `2.0.*` in your composer.json to avoid accidentally update to a version you are maybe not supporting.

nicolas-grekas:
At least from an internal pov, this removes a listener, which is a net win. About userland, that's similar yes, although more straightforward now I believe.

florentdestremau:
@alexander-schranz  Yeah I kinda suspected that, but I can't find an explicit message mentioning this, would be interesting to make sur it is :wink:

PR #218, opened 2022-01-12
[LiveComponent] Send live action arguments to backend

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | #102 Feature B
| License       | MIT

Include to the request to backend the named arguments from live action.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1070,7 +1070,7 @@ class default_1 extends Controller {
         directives.forEach((directive) => {
             const _executeAction = () => {
                 this._clearWaitingDebouncedRenders();
-                this._makeRequest(directive.action);
+                this._makeRequest(directive.action, directive.named);
             };
             let handled = false;
             directive.modifiers.forEach((modifier) => {
@@ -1173,11 +1173,14 @@ class default_1 extends Controller {
             }, this.debounceValue || DEFAULT_DEBOUNCE);
         }
     }
-    _makeRequest(action) {
+    _makeRequest(action, args) {
         const splitUrl = this.urlValue.split('?');
         let [url] = splitUrl;
         const [, queryString] = splitUrl;
         const params = new URLSearchParams(queryString || '');
+        if (typeof args === 'object' && Object.keys(args).length > 0) {
+            params.set('args', new URLSearchParams(args).toString());
+        }
         const fetchOptions = {};
         fetchOptions.headers = {
             'Accept': 'application/vnd.live-component+json',
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -138,7 +138,7 @@ export default class extends Controller {
                 // taking precedence
                 this._clearWaitingDebouncedRenders();
 
-                this._makeRequest(directive.action);
+                this._makeRequest(directive.action, directive.named);
             }
 
             let handled = false;
@@ -294,11 +294,15 @@ export default class extends Controller {
         }
     }
 
-    _makeRequest(action: string|null) {
+    _makeRequest(action: string|null, args: Record<string,unknown>) {
         const splitUrl = this.urlValue.split('?');
         let [url] = splitUrl
         const [, queryString] = splitUrl;
         const params = new URLSearchParams(queryString || '');
+
+        if (typeof args === 'object' && Object.keys(args).length > 0) {
+            params.set('args', new URLSearchParams(args).toString());
+        }
 
         const fetchOptions: RequestInit = {};
         fetchOptions.headers = {
--- a/src/LiveComponent/assets/test/controller/action.test.ts
+++ b/src/LiveComponent/assets/test/controller/action.test.ts
@@ -35,6 +35,8 @@ describe('LiveController Action Tests', () => {
                 data-action="live#action"
                 data-action-name="save"
             >Save</button>
+
+            <button data-action="live#action" data-action-name="sendNamedArgs(a=1, b=2, c=3)">Send named args</button>
         </div>
     `;
 
@@ -63,5 +65,16 @@ describe('LiveController Action Tests', () => {
         fetchMock.done();
 
         expect(postMock.lastOptions().body.get('comments')).toEqual('hi WEAVER');
+    });
+
+    it('Sends action named args', async () => {
+        const data = { comments: 'hi' };
+        const { element } = await startStimulus(template(data));
+
+        fetchMock.postOnce('http://localhost/_components/my_component/sendNamedArgs?values=a%3D1%26b%3D2%26c%3D3', {
+            html: template({ comments: 'hi' }),
+        });
+
+        getByText(element, 'Send named args').click();
     });
 });
--- a/src/LiveComponent/src/Attribute/LiveArg.php
+++ b/src/LiveComponent/src/Attribute/LiveArg.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Attribute;
+
+/**
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ *
+ * @experimental
+ */
+#[\Attribute(\Attribute::TARGET_PARAMETER)]
+final class LiveArg
+{
+    public function __construct(public ?string $name = null)
+    {
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    public static function liveArgs(object $component, string $action): array
+    {
+        $method = new \ReflectionMethod($component, $action);
+        $liveArgs = [];
+
+        foreach ($method->getParameters() as $parameter) {
+            foreach ($parameter->getAttributes(self::class) as $liveArg) {
+                /** @var LiveArg $attr */
+                $attr = $liveArg->newInstance();
+                $parameterName = $parameter->getName();
+
+                $liveArgs[$parameterName] = $attr->name ?? $parameterName;
+            }
+        }
+
+        return $liveArgs;
+    }
+}
--- a/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
@@ -28,6 +28,7 @@ use Symfony\Component\Security\Csrf\CsrfToken;
 use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 use Symfony\Contracts\Service\ServiceSubscriberInterface;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveArg;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\TwigComponent\ComponentFactory;
 use Symfony\UX\TwigComponent\ComponentRenderer;
@@ -137,11 +138,21 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
 
         $this->container->get(LiveComponentHydrator::class)->hydrate($component, $data);
 
+        $request->attributes->set('_component', $component);
+
+        if (!\is_string($queryString = $request->query->get('args'))) {
+            return;
+        }
+
         // extra variables to be made available to the controller
         // (for "actions" only)
-        parse_str($request->query->get('values'), $values);
-        $request->attributes->add($values);
-        $request->attributes->set('_component', $component);
+        parse_str($queryString, $args);
+
+        foreach (LiveArg::liveArgs($component, $action) as $parameter => $arg) {
+            if (isset($args[$arg])) {
+                $request->attributes->set($parameter, $args[$arg]);
+            }
+        }
     }
 
     public function onKernelView(ViewEvent $event): void
--- a/src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
+++ b/src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
@@ -17,6 +17,7 @@ use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\LiveComponent\Tests\ContainerBC;
 use Symfony\UX\LiveComponent\Tests\Fixture\Component\Component1;
 use Symfony\UX\LiveComponent\Tests\Fixture\Component\Component2;
+use Symfony\UX\LiveComponent\Tests\Fixture\Component\Component6;
 use Symfony\UX\LiveComponent\Tests\Fixture\Entity\Entity1;
 use Symfony\UX\TwigComponent\ComponentFactory;
 use Zenstruck\Browser\Response\HtmlResponse;
@@ -213,6 +214,47 @@ final class LiveComponentSubscriberTest extends KernelTestCase
             ])
             ->assertStatus(204)
             ->assertHeaderEquals('Location', '/')
+        ;
+    }
+
+    public function testInjectsLiveArgs(): void
+    {
+        /** @var LiveComponentHydrator $hydrator */
+        $hydrator = self::getContainer()->get('ux.live_component.component_hydrator');
+
+        /** @var ComponentFactory $factory */
+        $factory = self::getContainer()->get('ux.twig_component.component_factory');
+
+        /** @var Component6 $component */
+        $component = $factory->create('component6');
+
+        $dehydrated = $hydrator->dehydrate($component);
+        $token = null;
+
+        $dehydratedWithArgs = array_merge($dehydrated, [
+            'args' => http_build_query(['arg1' => 'hello', 'arg2' => 666, 'custom' => '33.3']),
+        ]);
+
+        $this->browser()
+            ->throwExceptions()
+            ->get('/_components/component6?'.http_build_query($dehydrated))
+            ->assertSuccessful()
+            ->assertHeaderContains('Content-Type', 'html')
+            ->assertContains('Arg1: not provided')
+            ->assertContains('Arg2: not provided')
+            ->assertContains('Arg3: not provided')
+            ->use(function (HtmlResponse $response) use (&$token) {
+                // get a valid token to use for actions
+                $token = $response->crawler()->filter('div')->first()->attr('data-live-csrf-value');
+            })
+            ->post('/_components/component6/inject?'.http_build_query($dehydratedWithArgs), [
+                'headers' => ['X-CSRF-TOKEN' => $token],
+            ])
+            ->assertSuccessful()
+            ->assertHeaderContains('Content-Type', 'html')
+            ->assertContains('Arg1: hello')
+            ->assertContains('Arg2: 666')
+            ->assertContains('Arg3: 33.3')
         ;
     }
 }
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -5,6 +5,8 @@
 -   The Live Component AJAX endpoints now return HTML in all situations
     instead of JSON.
 
+-   Send live action arguments to backend
+
 ## 2.0.0
 
 -   Support for `stimulus` version 2 was removed and support for `@hotwired/stimulus`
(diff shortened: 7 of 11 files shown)


norkunas:
Added some tests, but I trully don't understand how they works, so just adopted from existing test :D

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -296,11 +296,26 @@ export default class extends Controller {
...
+        let valueParams = null;
+
+        if (Array.isArray(values)) {
+            valueParams = new URLSearchParams();
+
+            values.forEach((v, i) => valueParams.set(i, v));
+        } else if (typeof values === 'object' && Object.keys(values).length > 0) {
+            valueParams = new URLSearchParams();
+
+            Object.keys(values).forEach(k => valueParams.set(k, values[k]));
```
Kocal: If `values` if following this structure `{key: value}`, then you can use:
```suggestion
            valueParams = new URLSearchParams(values);
```

image

norkunas: cool, forgot about that :)

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -296,11 +296,24 @@ export default class extends Controller {
         }
     }
 
-    _makeRequest(action: string|null) {
+    _makeRequest(action: string|null, values: Array<string>|object) {
```
Kocal: Prefer use [`Record<K,V>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) instead of `object`:
```suggestion
    _makeRequest(action: string|null, values: Array<string>|Record<string,unknown>) {
```

norkunas: Sure. I'm not familiar with TS :)

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -1162,11 +1162,22 @@ class default_1 extends Controller {
...
+    _makeRequest(action, values) {
         const splitUrl = this.urlValue.split('?');
         let [url] = splitUrl;
         const [, queryString] = splitUrl;
         const params = new URLSearchParams(queryString || '');
+        let valueParams = null;
+        if (Array.isArray(values)) {
+            valueParams = new URLSearchParams();
+            values.forEach((v, i) => valueParams.set(i, v));
+        }
```
weaverryan: I believe we should ONLY support named arguments. Due to the way Symfony's routing+arguments system works (i.e. that it matches by name), sending `?values=0=a&1=b&2=c` won't be meaningful: Symfony won't match this to your action arguments. And I think that's ok: let's do named arguments only.

src/LiveComponent/assets/test/controller/action.test.ts
```diff
@@ -64,4 +68,46 @@ describe('LiveController Action Tests', () => {
...
+
+        await waitFor(() => expect(element).toHaveTextContent('Comment Changed!'));
+        expect(getByLabelText(element, 'Comments:')).toHaveValue(' a b c');
+
+        fetchMock.done();
+
+        expect(postMock.lastOptions().body.get('comments')).toEqual('hia b c');
+    });
+
+    it('Sends action named args', async () => {
```
weaverryan, line 70: We can simplify this test a bit - it won't be as complex as the original test: I was doing some extra sanity checks on that test (like making sure that the element re-renders after an action) that we don't need to do here. I'll comment the changes below

src/LiveComponent/assets/test/controller/action.test.ts
```diff
@@ -64,4 +68,46 @@ describe('LiveController Action Tests', () => {
...
+        const data = { comments: 'hi' };
+        const { element } = await startStimulus(template(data));
+
+        const postMock = fetchMock.postOnce('http://localhost/_components/my_component/updateNamed?values=a%3D1%26b%3D2%26c%3D3', {
+            html: template({ comments: ' a=1 b=2 c=3', isChanged: true }),
+            data: { isChanged: true }
+        });
```
weaverryan, line 76: This `fetchMock` is the key to the test. But only because it "asserts" that clicking the action button sent the correct POST request with the correct URL (the most important part being the `?value=` query parameter). We don't really care about how the component re-renders.

So, small simplification:

```
html: template({ comments: 'hi', isChanged: true }),
```

should be fine

src/LiveComponent/assets/test/controller/action.test.ts
```diff
@@ -64,4 +68,46 @@ describe('LiveController Action Tests', () => {
...
+
+        await userEvent.type(getByLabelText(element, 'Comments:'), 'a b c');
```
weaverryan: can remove this

src/LiveComponent/assets/test/controller/action.test.ts
```diff
@@ -64,4 +68,46 @@ describe('LiveController Action Tests', () => {
...
+
+        getByText(element, 'Update named').click();
+
+        await waitFor(() => expect(element).toHaveTextContent('Comment Changed!'));
+        expect(getByLabelText(element, 'Comments:')).toHaveValue(' a=1 b=2 c=3');
+
+        fetchMock.done();
+
+        expect(postMock.lastOptions().body.get('comments')).toEqual('hia b c');
```
weaverryan: can be removed

src/LiveComponent/assets/test/controller/action.test.ts
```diff
@@ -64,4 +68,46 @@ describe('LiveController Action Tests', () => {
...
+            html: template({ comments: ' a=1 b=2 c=3', isChanged: true }),
+            data: { isChanged: true }
+        });
+
+        await userEvent.type(getByLabelText(element, 'Comments:'), 'a b c');
+
+        getByText(element, 'Update named').click();
+
+        await waitFor(() => expect(element).toHaveTextContent('Comment Changed!'));
+        expect(getByLabelText(element, 'Comments:')).toHaveValue(' a=1 b=2 c=3');
```
weaverryan: can be removed

weaverryan (commented):
Thanks for taking this!

I left a few comments (and thanks for adding the test). There's one important piece missing, and it's on the server-side. We can't allow ANY values to be passed to the server, because then people could do something like `?values=_controller=App\SomeController::action`. That's because `values` is added to `$request->attributes`.

So, this needs to be an "opt in" situation. I mentioned in the issue:

> We need to add a way to "allow" certain action arguments to be "passable" from the frontend

Probably this would be via an attribute above the action, like:

```php
#[ActionArgument('someValue')]
public function updateUser(string $someValue)
{
}
```

Then, before adding the `?values` to `$request->attributes` (this happens in `LiveComponentSubscriber`), we would check the controller for these attributes, and only add the ones allowed.

I also thought about doing this "automatically"... by reading the arguments on the controller and allowing any that are actually there (e.g. in the example above, we would see there is a `$someValue` argument and automatically allow `someValue` to be passed as a value... but I'm not confident from a security standpoint).

norkunas:
> 

What if we would not directly append to the `request->attributes`, but set an attribute `values` and then add a custom argument value resolver to inject to live actions? Then this would not be a security risk (currently it is, because values can be passed if faking request manually via post man or some http client). And as for the non named args, I added it so if people really need, they can manually retrieve then from the request attributes.

And: currently symfony argument value resolver injects the values from attributes, so `#[ActionArgument('someValue')]` wouldn't do anything

weaverryan:
> What if we would not directly append to the `request->attributes`, but set an attribute `values` and then add a custom argument value resolver to inject to live actions?

That's a fantastic idea! I love it! I'd use some attribute named something like `_live_values` instead of `values` (just to not collide with anything and because it's kind of an internal value). Also, we should set the priority on the arg resolver low. What we need to avoid (a security edge-case) is someone having a `$iAmConfiguredViaAServiceBind` argument to an action that is configured via service config. We don't want a bad user to be able to override that by passing this as a value. I think (?) that if we set our arg resolver to a low priority, then it will only be called for arguments that have not been resolved in any other way.

> And as for the non named args, I added it so if people really need, they can manually retrieve then from the request attributes.

Unless you are using this feature (and if so, give some explanation as to how it's useful to you), we should remove it. 

And one last thing, can you update the documentation a bit for the new feature?

Thanks a lot for this - it's excellent work and you've got excellent ideas!

norkunas:
> > And as for the non named args, I added it so if people really need, they can manually retrieve then from the request attributes.
> 
> Unless you are using this feature (and if so, give some explanation as to how it's useful to you), we should remove it.

Trully - I don't, I only need named args currently :) Ok, will remove, then if someone will have a usecase for this then could be proposed again later :)

> And one last thing, can you update the documentation a bit for the new feature?

Yes, sure, will do tomorrow :) 

> I'd use some attribute named something like _live_values instead of values (just to not collide with anything and because it's kind of an internal value). Also, we should set the priority on the arg resolver low. What we need to avoid (a security edge-case) is someone having a $iAmConfiguredViaAServiceBind argument to an action that is configured via service config. We don't want a bad user to be able to override that by passing this as a value. I think (?) that if we set our arg resolver to a low priority, then it will only be called for arguments that have not been resolved in any other way.

Yes, wanted just to propose so didn't think much about the attribute name :) But this will be a BC break, but as till now it was not fully supported, I think it shouldn't impact anyone?

weaverryan:
> Ok, will remove, then if someone will have a usecase for this then could be proposed again later :)

Very true - thank you!

> Yes, wanted just to propose so didn't think much about the attribute name :) But this will be a BC break, but as till now it was not fully supported, I think it shouldn't impact anyone?

BC break is ok, because we're still experimental. And it's unlikely anyone is using this right now anyways.

BUUT, after talking with @kbond, instead of doing your argument resolver, we like a different path (let me explain). But, this full solution doesn't need to be done in THIS PR.

At the very least, we should stop merging all of the values into `$request->attributes` and instead put them in some new key, as we've been discussing. I was now thinking about `_live_args`. I'm not sure why I was calling this `values` until now. We may also want to update the frontend query param to be `?args=` instead of `?values=` unless you can think of a reason to keep it as `values`.

And, we could stop here for this PR if you want to. It would mean the user would need to, for now, use `$request->attributes->get('_live_args')` in their live action. Annoying, but temporary.

The rest of the solution would involve a new `LiveArg` attribute:

```php
#[LiveAction]
public function updateUser(#[LiveArg] string $someValue, #[LiveArg]  string $otherValue, IAmANormalService $service)
{
}
```

The logic would be:

We read all of the `LiveArg`. If any of these are present in `$request->query->get('values')`, then we add that avlue to `$request->attributes`. 

The reason behind doing this is so that the values can be available to things like FrameworkExtraBundle's param converters. In other words, so that the values work even MORE identically to route `{params}`.

Here is a more advanced example that shows off an argument to `LiveArg`:

In HTML:

```
data-action-name="updatePost(id=1)
```

```php
#[LiveAction]
public function updatePost(#[LiveValue('id')] Post $post)
{
}
```

In this case, `id` is what is sent to the server. And thanks to the `'id'` arg to `LiveValue`, we recognize that and place it into `$request->attributes`. But then, FrameworkExtraBundle can work its normal magic to convert this into the `Post` entity.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -141,11 +142,19 @@ public function onKernelController(ControllerEvent $event): void
 
         $this->container->get(LiveComponentHydrator::class)->hydrate($component, $data);
 
-        // extra variables to be made available to the controller
-        // (for "actions" only)
-        parse_str($request->query->get('values'), $values);
-        $request->attributes->add($values);
         $request->attributes->set('_component', $component);
+
+        if (\is_string($queryString = $request->query->get('args'))) {
```
norkunas: This was needed to resolve deprecation: `10x: parse_str(): Passing null to parameter #1 ($string) of type string is deprecated`. Also it would fail with other types, so added sanity check

kbond: Good call, let's just return early if not a string.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -141,11 +142,19 @@ public function onKernelController(ControllerEvent $event): void
...
+            // extra variables to be made available to the controller
+            // (for "actions" only)
+            parse_str($queryString, $args);
+
+            foreach ($this->readLiveActionArgs($component, $action) as $parameter => $arg) {
```
kbond: Instead of including this method in this class, let's move to an internal static method on `LiveArg`.

I'm not sure if this is the best practice but it's what I've done for the other attributes (ie [`AsLiveComponent`](https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Attribute/AsLiveComponent.php#L38)). I've found it helps keep the code a bit neater and concise.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
kbond: Were we going to call `args` `_live_args`?

kbond (changes requested):
Thanks for working on this @norkunas! Just a few minor things.

src/LiveComponent/src/Attribute/LiveArg.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function __construct(?string $name = null)
+    {
+        $this->name = $name;
+    }
+
+    public function getName(): ?string
+    {
+        return $this->name;
+    }
```
kbond:
```suggestion
    public function __construct(public ?string $name = null)
    {
    }
```

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -250,4 +259,25 @@ private function isLiveComponentJsonRequest(Request $request): bool
...
+        $method = new \ReflectionMethod($component, $action);
+        $liveArgs = [];
+
+        foreach ($method->getParameters() as $parameter) {
+            foreach ($parameter->getAttributes(LiveArg::class) as $liveArg) {
+                /** @var LiveArg $attr */
+                $attr = $liveArg->newInstance();
+                $parameterName = $parameter->getName();
+
+                $liveArgs[$parameterName] = $attr->getName() ?? $parameterName;
```
kbond:
```suggestion
                $liveArgs[$parameterName] = $attr->name ?? $parameterName;
```

norkunas:
> Were we going to call args _live_args?

I am not sure it's needed anymore as we don't store all args

norkunas:
Would be nice if cs-fixer in CI would report in files where code style was not followed, because now when running cs fixer it fixes too much unrelated fixes.

edit: well now green without any changes in code :man_shrugging:

kbond:
> I am not sure it's needed anymore as we don't store all args

Ah ok so we aren't going to have the catch all way to grab the raw args from the request? The only way into get them is to inject (or parse the raw args yourself).

I don't have a strong opinion on this, I was just curious.

> Would be nice if cs-fixer in CI would report in files where code style was not followed, because now when running cs fixer it fixes too much unrelated fixes.

Would be nice to get fabbot on this repo. I realized after I suggested property promotion for the attribute that the cs might fail (think it's a syntax error). ~Guess that's been fixed.~ guess not... https://github.com/symfony/ux/runs/4817869695?check_suite_focus=true#step:4:7

src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
```diff
@@ -240,4 +241,43 @@ public function testCanRedirectFromComponentAction(): void
...
+        /** @var ComponentFactory $factory */
+        $factory = self::getContainer()->get('ux.twig_component.component_factory');
+
+        /** @var Component6 $component */
+        $component = $factory->create('component6');
+
+        $dehydrated = $hydrator->dehydrate($component);
+        $token = null;
+
+        $dehydrated['args'] = http_build_query(['arg1' => 'hello', 'arg2' => 666, 'custom' => '33.3']);
```
weaverryan: Do we need `http_build_query` in both spots?

src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
```diff
@@ -240,4 +241,43 @@ public function testCanRedirectFromComponentAction(): void
...
+
+        $this->browser()
+            ->throwExceptions()
+            ->get('/_components/component6?'.http_build_query($dehydrated))
```
weaverryan, line 240: It's minor. But to be more precise, this request should NOT include the `args=` part of the query parameter. Only the next usage below should include that query param.

src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
```diff
@@ -240,4 +241,43 @@ public function testCanRedirectFromComponentAction(): void
...
+            ->post('/_components/component6/inject?'.http_build_query($dehydrated), [
+                'headers' => ['X-CSRF-TOKEN' => $token],
+            ])
+            ->assertSuccessful()
+            ->assertHeaderContains('Content-Type', 'html')
+            ->assertContains('Arg1: hello')
+            ->assertContains('Arg2: 666')
+            ->assertContains('Arg3: 33.3')
+        ;
+    }
```
weaverryan, line 259: That's a very nice test!

src/LiveComponent/README.md
```diff
@@ -541,6 +541,56 @@ class RandomNumberComponent
 }
 ```
 
+##### Passing arguments to live action
+
+You can also provide custom arguments to your action. For example Symfony Form CollectionType needs some JavaScript
+to make form dynamic, but it can be avoided.
```
weaverryan: Before going into the `CollectionType` example, let's first show a very basic example. For example:

```html
<button data-action="live#action" data-action-name="addItem(id={{ item.id }})">Add Item</button>
```

Then:

```php
    #[LiveAction]
    public function addItem(#[LiveArg] int $id)
    {
        // ...
    }
```

We should probably also, after this, show the example of using `#[LiveArg('id')] Item $item` to mention how this would work with the ParamConverter.

THEN we can show the `CollectionType` example. Or, better-yet (it's up to you if you want to do this, it's a bit more work), you could contribute a working `CollectionType` to https://github.com/weaverryan/live-demo and we could mention it and link to it from here.

weaverryan (commented):
Lookin' awesome!

src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
kbond: This browser thing is pretty neat 😉

kbond:
@weaverryan what do you think we should do for the failing cs? Easiest solution for this pr is to not use a promoted property for the attribute. I'm thinking switch that out, we can always change later once the cs fixer can handle.

src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
norkunas: Yes, because `parse_str` expects string

norkunas:
I think this is ready :)

kbond (changes requested):
Can you add a changelog entry for 2.1?

.github/workflows/test.yaml
```diff
@@ -10,10 +10,11 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '7.4'
+                  tools: cs2pr
             - name: php-cs-fixer
               run: |
                   wget https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v2.18.2/php-cs-fixer.phar -q
-                  php php-cs-fixer.phar fix --dry-run --diff
+                  php php-cs-fixer.phar fix --dry-run --diff --format=checkstyle | cs2pr
```
kbond: This is nice!

norkunas:
> Can you add a changelog entry for 2.1?

Done

.github/workflows/test.yaml
kbond: @norkunas, you can remove this change - it was done in #240. This will fix the conflicts. We can also use PHP 8 syntax now!

norkunas: Done, ~~but promoted properties are available since PHP8.1 and LiveComponent declares 8.0 support~~ :) mixed with readonly props :D

kbond: Fyi, Php 8.0 does support constructor property promotion: https://www.php.net/releases/8.0/en.php.

norkunas:
Will it be merged now? :stuck_out_tongue_closed_eyes:

kbond (approved):

weaverryan:
Woohoo! Thank you for this GREAT feature Tomas!

PR #220, opened 2022-01-14
[Twig][Live] add component attribute system/helper

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

A common pattern with other component library's (ie blade components/vue) is to pass html attributes to your component that are set on the component's root node. This proposal adds a way to do this with twig components. Enable on your component by adding the `HasAttributesTrait`. Attributes are any data passed to `component()` that cannot be mounted on the component itself. This extra data is added to a `ComponentAttributes` object that lives as a public property on your component (available as `attributes` in your component's template).

This should all work out-of-the box with live components.

Todo:
- [x] Tests
- [x] Documentation
- [x] #240
- [x] #241 
- [x] Merge `init_live_component()` into attributes when available
- [x] Document `PreRender` event.
- [x] Adjust `ComponentAttributes` api (defaults instead of merge)
- [ ] `attributes` always available in normal twig component templates (not sure about this - it would effectively make attributes _native_ but it a lame way)
- [x] replace `init_live_component()` with `attributes` in docs
- [x] Finalize docs (versionadded tags)

## Usage

To use, add the `HasAttributesTrait` to your component:

```php
#[AsTwigComponent('my_component')]
class MyComponent
{
    use HasAttributesTrait;
}
```

Then, in your component's template:

```twig
{# templates/components/my_component.html.twig #}

<div{{ attributes }}>
  My Component!
</div>
```

When rendering the component, you can pass an array of html attributes to add:

```twig
{{ component('my_component', { class: 'foo', style: 'color:red' }) }}

{# renders as: #}
<div class="foo" style="color:red">
  My Component!
</div>
```

### Defaults & Merging

In your component template, you can set defaults that are merged with
passed attributes. The passed attributes override the default with
the exception of *class*. For class, the defaults are prepended:

```twig
{# templates/components/my_component.html.twig #}

<button{{ attributes.defaults({ class: 'bar', type: 'button' }) }}>
  My Component!
</button>

{# render component #}
{{ component('my_component', { style: 'color:red' }) }}
{{ component('my_component', { class: 'foo', type: 'submit' }) }}

{# renders as: #}
<div class="bar" style="color:red">
  My Component!
</div>
<div class="bar foo" type="submit">
  My Component!
</div>
```

### Only

```twig
{# templates/components/my_component.html.twig #}

<div{{ attributes.only('class')) }}>
  My Component!
</div>

{# render component #}
{{ component('my_component', { class: 'foo', style: 'color:red' }) }}

{# renders as: #}
<div class="foo">
  My Component!
</div>
```

### Without

```twig
{# templates/components/my_component.html.twig #}

<div{{ attributes.without('class')) }}>
  My Component!
</div>

{# render component #}
{{ component('my_component', { class: 'foo', style: 'color:red' }) }}

{# renders as: #}
<div style="color:red">
  My Component!
</div>
```

### Live Components

We removed the `init_live_component()` twig function (**BC BREAK**) and replaced with the `attributes` variable which is always available (even if your component doesn't use `HasAttributesTrait`).

To upgrade, replace all calls to `init_live_component()` with `attributes`:

```diff
- <div {{ init_live_component() }}>
+ <div {{ attributes }}>
```

### DIFF
--- a/src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
+++ b/src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\EventListener;
+
+use Psr\Container\ContainerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
+use Symfony\UX\LiveComponent\Twig\LiveComponentRuntime;
+use Symfony\UX\TwigComponent\ComponentAttributes;
+use Symfony\UX\TwigComponent\EventListener\PreRenderEvent;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class AddLiveAttributesSubscriber implements EventSubscriberInterface, ServiceSubscriberInterface
+{
+    public function __construct(private ContainerInterface $container)
+    {
+    }
+
+    public function onPreRender(PreRenderEvent $event): void
+    {
+        if (!$event->getMetadata()->get('live', false)) {
+            // not a live component, skip
+            return;
+        }
+
+        /** @var ComponentAttributes $attributes */
+        $attributes = $this->container->get(LiveComponentRuntime::class)
+            ->getLiveAttributes($event->getComponent(), $event->getMetadata())
+        ;
+
+        $variables = $event->getVariables();
+
+        if (isset($variables['attributes']) && $variables['attributes'] instanceof ComponentAttributes) {
+            // merge with existing attributes if available
+            $attributes = $attributes->defaults($variables['attributes']->all());
+        }
+
+        $variables['attributes'] = $attributes;
+
+        $event->setVariables($variables);
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [PreRenderEvent::class => 'onPreRender'];
+    }
+
+    public static function getSubscribedServices(): array
+    {
+        return [
+            LiveComponentRuntime::class,
+        ];
+    }
+}
--- a/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
@@ -31,6 +31,7 @@ use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\LiveComponent\Attribute\LiveArg;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\TwigComponent\ComponentFactory;
+use Symfony\UX\TwigComponent\ComponentMetadata;
 use Symfony\UX\TwigComponent\ComponentRenderer;
 
 /**
@@ -73,24 +74,28 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
         $componentName = (string) $request->get('component');
 
         try {
-            $config = $this->container->get(ComponentFactory::class)->configFor($componentName);
+            /** @var ComponentMetadata $metadata */
+            $metadata = $this->container->get(ComponentFactory::class)->metadataFor($componentName);
         } catch (\InvalidArgumentException $e) {
             throw new NotFoundHttpException(sprintf('Component "%s" not found.', $componentName), $e);
         }
 
-        $request->attributes->set('_component_template', $config['template']);
+        if (!$metadata->get('live', false)) {
+            throw new NotFoundHttpException(sprintf('"%s" (%s) is not a Live Component.', $metadata->getClass(), $componentName));
+        }
+
+        $request->attributes->set('_component_metadata', $metadata);
 
         if ('get' === $action) {
-            $defaultAction = trim($config['default_action'] ?? '__invoke', '()');
-            $componentClass = $config['class'];
+            $defaultAction = trim($metadata->get('default_action', '__invoke'), '()');
 
-            if (!method_exists($componentClass, $defaultAction)) {
+            if (!method_exists($metadata->getClass(), $defaultAction)) {
                 // todo should this check be in a compiler pass to ensure fails at compile time?
-                throw new \LogicException(sprintf('Live component "%s" requires the default action method "%s".%s', $componentClass, $defaultAction, '__invoke' === $defaultAction ? ' Either add this method or use the DefaultActionTrait' : ''));
+                throw new \LogicException(sprintf('Live component "%s" (%s) requires the default action method "%s".%s', $metadata->getClass(), $componentName, $defaultAction, '__invoke' === $defaultAction ? ' Either add this method or use the DefaultActionTrait' : ''));
             }
 
             // set default controller for "default" action
-            $request->attributes->set('_controller', sprintf('%s::%s', $config['service_id'], $defaultAction));
+            $request->attributes->set('_controller', sprintf('%s::%s', $metadata->getServiceId(), $defaultAction));
             $request->attributes->set('_component_default_action', true);
 
             return;
@@ -106,7 +111,7 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
             throw new BadRequestHttpException('Invalid CSRF token.');
         }
 
-        $request->attributes->set('_controller', sprintf('%s::%s', $config['service_id'], $action));
+        $request->attributes->set('_controller', sprintf('%s::%s', $metadata->getServiceId(), $action));
     }
 
     public function onKernelController(ControllerEvent $event): void
@@ -232,7 +237,7 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
 
         $html = $this->container->get(ComponentRenderer::class)->render(
             $component,
-            $request->attributes->get('_component_template')
+            $request->attributes->get('_component_metadata')
         );
 
         return new Response($html);
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -43,10 +43,15 @@ A real-time product search component might look like this::
     The ability to reference local variables in the template (e.g. ``query``) was added in TwigComponents 2.1.
     Previously, all data needed to be referenced through ``this`` (e.g. ``this.query``).
 
+.. versionadded:: 2.1
+
+    The ability to initialize live component with the ``attributes`` twig variable was added in LiveComponents
+    2.1. Previously, the ``init_live_component()`` function was required (this function was removed in 2.1).
+
 .. code-block:: twig
 
     {# templates/components/product_search.html.twig #}
-    <div {{ init_live_component(this) }}>
+    <div {{ attributes }}>
         <input
             type="search"
             name="query"
@@ -159,13 +164,13 @@ re-rendered live on the frontend), replace the component's
       }
 
 Then, in the template, make sure there is *one* HTML element around your
-entire component and use the ``{{ init_live_component() }}`` function to
-initialize the Stimulus controller:
+entire component and use the ``{{ attributes }}`` variable to initialize
+the Stimulus controller:
 
 .. code-block:: diff
 
     - <div>
-    + <div {{ init_live_component(this) }}>
+    + <div {{ attributes }}>
           <strong>{{ this.randomNumber }}</strong>
       </div>
 
@@ -176,7 +181,7 @@ and give the user a new random number:
 
 .. code-block:: twig
 
-    <div {{ init_live_component(this) }}>
+    <div {{ attributes }}>
         <strong>{{ this.randomNumber }}</strong>
 
         <button
@@ -239,6 +244,44 @@ exceptions being properties that hold services (these don't need to be
 stateful because they will be autowired each time before the component
 is rendered) and `properties used for computed properties`_.
 
+Component Attributes
+--------------------
+
+.. versionadded:: 2.1
+
+    The ``HasAttributes`` trait was added in TwigComponents 2.1.
+
+`Component attributes`_ allows you to render your components with extra
+props that are are converted to html attributes and made available in
+your component's template as an ``attributes`` variable. When used on
+live components, these props are persisted between renders. You can enable
+this feature by having your live component use the ``HasAttributesTrait``:
+
+.. code-block:: diff
+
+      // ...
+      use Symfony\UX\LiveComponent\Attribute\LiveProp;
+    + use Symfony\UX\TwigComponent\HasAttributesTrait;
+
+      #[AsLiveComponent('random_number')]
+      class RandomNumberComponent
+      {
+    +     use HasAttributesTrait;
+
+          #[LiveProp]
+          public int $min = 0;
+
+Now, when rendering your component, you can pass html attributes
+as props and these will be added to ``attributes``:
+
+.. code-block:: twig
+
+    {{ component('random_number', { min: 5, max: 500, class: 'widget', style: 'color: black;' }) }}
+
+    {# renders as: #}
+    <div class="widget" style="color: black;" <!-- other live attributes -->>
+        <!-- ... -->
+
 data-action=“live#update”: Re-rendering on LiveProp Change
 ----------------------------------------------------------
 
@@ -251,7 +294,7 @@ Let's add two inputs to our template:
 .. code-block:: twig
 
     {# templates/components/random_number.html.twig #}
-    <div {{ init_live_component(this) }}>
+    <div {{ attributes }}>
         <input
             type="number"
             value="{{ min }}"
@@ -368,7 +411,7 @@ property. The following code works identically to the previous example:
 
 .. code-block:: diff
 
-      <div {{ init_live_component(this)>
+      <div {{ attributes }}>
           <input
               type="number"
               value="{{ min }}"
@@ -791,7 +834,7 @@ as ``this.form`` thanks to the trait:
 
     {# templates/components/post_form.html.twig #}
     <div
-        {{ init_live_component(this) }}
+        {{ attributes }}
         {#
             Automatically catch all "change" events from the fields
             below and re-render the component.
@@ -815,8 +858,7 @@ as ``this.form`` thanks to the trait:
     </div>
 
 Mostly, this is a pretty boring template! It includes the normal
-``init_live_component(this)`` and then you render the form however you
-want.
+``attributes`` and then you render the form however you want.
 
 But the result is incredible! As you finish changing each field, the
 component automatically re-renders - including showing any validation
@@ -1024,7 +1066,7 @@ section above) is to add:
 .. code-block:: diff
 
       <div
-          {{ init_live_component(this) }}
+          {{ attributes }}
     +     data-action="change->live#update"
       >
 
@@ -1056,7 +1098,7 @@ rendered the ``content`` through a Markdown filter from the
 
 .. code-block:: twig
 
-    <div {{init_live_component(this)}}>
+    <div {{ attributes }}>
         <input
             type="text"
             value="{{ post.title }}"
@@ -1221,7 +1263,7 @@ You can also use “polling” to continually refresh a component. On the
 .. code-block:: diff
 
       <div
-          {{ init_live_component(this) }}
+          {{ attributes }}
     +     data-poll
       >
 
@@ -1233,7 +1275,7 @@ delay for 500ms:
 .. code-block:: twig
 
     <div
-        {{ init_live_component(this) }}
+        {{ attributes }}
         data-poll="delay(500)|$render"
     >
 
@@ -1242,7 +1284,7 @@ You can also trigger a specific “action” instead of a normal re-render:
 .. code-block:: twig
 
     <div
-        {{ init_live_component(this) }}
+        {{ attributes }}
 
         data-poll="save"
         {#
@@ -1437,7 +1479,7 @@ In the ``EditPostComponent`` template, you render the
 .. code-block:: twig
 
     {# templates/components/edit_post.html.twig #}
-    <div {{ init_live_component(this) }}>
+    <div {{ attributes }}>
         <input
             type="text"
             name="post[title]"
@@ -1459,7 +1501,7 @@ In the ``EditPostComponent`` template, you render the
 
 .. code-block:: twig
 
-    <div {{ init_live_component(this) }} class="mb-3">
+    <div {{ attributes }} class="mb-3">
         <textarea
             name="{{ name }}"
             data-model="value"
@@ -1496,3 +1538,4 @@ bound to Symfony's BC policy for the moment.
 .. _`experimental`: https://symfony.com/doc/current/contributing/code/experimental.html
 .. _`dependent form fields`: https://symfony.com/doc/current/form/dynamic_form_modification.html#dynamic-generation-for-submitted-forms
 .. _`Symfony UX configured in your app`: https://symfony.com/doc/current/frontend/ux.html
+.. _`Component attributes`: https://symfony.com/bundles/ux-twig-component/current/index.html#component-attributes
--- a/src/LiveComponent/src/Twig/LiveComponentRuntime.php
+++ b/src/LiveComponent/src/Twig/LiveComponentRuntime.php
@@ -14,7 +14,9 @@ namespace Symfony\UX\LiveComponent\Twig;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
+use Symfony\UX\TwigComponent\ComponentAttributes;
 use Symfony\UX\TwigComponent\ComponentFactory;
+use Symfony\UX\TwigComponent\ComponentMetadata;
 use Twig\Environment;
 
 /**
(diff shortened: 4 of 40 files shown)


kbond:
629d8343e615a89371a82b1f0a8d581e989284c9 demonstrates a bit of a hack to remove the requirement for `{{ attributes|raw }}`. Maybe there's a better way to do this?

_BTW, I'm totally fine to just live with `|raw`._

seb-jean:
I love this feature @kbond

kbond:
To make this feature _live_, I was thinking to mark the property as a `LiveProp` (with a hydrator) so it _just works_ when making an existing component _live_.

weaverryan:
I like this - and the idea is proven by other libraries.

> 629d834 demonstrates a bit of a hack to remove the requirement for {{ attributes|raw }}. Maybe there's a better way to do this?

See https://twig.symfony.com/doc/3.x/api.html#escaper-extension - look for `addSafeClass`. I *think* that's the correct way to handle this.

> To make this feature live, I was thinking to mark the property as a LiveProp (with a hydrator) so it just works when making an existing component live.

I can't think of why this would be a problem. So 👍 

> This proposal adds a way to do this (albeit a bit verbose) with twig components

2 ideas for this, both which are beyond the scope of this PR:

A) Allow using "…args" for the “props”

```twig
{{ component('my_component', attributes: { class: 'foo', style: 'color:red' }, otherProp: 'bar' ) }}
```

It removes an extra `{}` basically.

2. Allowing using an `<component attributes=""` style for the components.

src/TwigComponent/src/AttributeBag.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @experimental
+ */
+final class AttributeBag
```
weaverryan: `ComponentAttributes` might be more descriptive? It's mostly internal anyways...

weaverryan:
Also, it would be great if, for live components, we could also do:

```html
<div{{ attributes }}>
```

Instead of:

```html
<div {{ init_live_component(this) }}>
```

kbond:
> See https://twig.symfony.com/doc/3.x/api.html#escaper-extension - look for addSafeClass. I think that's the correct way to handle this.

This is what I used. The hack I was referring to is the weird way of adding it in the extension. I don't think there's a great way for a third party bundle to add these.

src/TwigComponent/src/HasAttributesTrait.php
```diff
@@ -0,0 +1,58 @@
...
+        return $data;
+    }
+
+    #[PostMount(priority: -1000)]
+    public function mountAttributes(array $data): array
+    {
+        $attributes = array_merge($data['_attributes'], $data);
+        unset($attributes['_attributes']);
+
+        $this->attributes = new ComponentAttributes($attributes);
```
kbond: Should we do some error checking here to ensure these values are all scalar? If they are not scalar, should I throw an exception here or return the array including this value and let the "Unable to write prop to component..." exception be thrown?

src/TwigComponent/src/HasAttributesTrait.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+trait HasAttributesTrait
+{
+    #[LiveProp(hydrateWith: 'hydrateAttributes', dehydrateWith: 'dehydrateAttributes')]
+    public ComponentAttributes $attributes;
+
+    #[PreMount]
+    public function preMountAttributes(array $data): array
+    {
+        $data['_attributes'] = $data['attributes'] ?? [];
```
weaverryan: I'm missing the point of this. Is this to try to avoid `attributes` being available during the normal "mounting" process?

src/TwigComponent/src/HasAttributesTrait.php
weaverryan: My gut says to start MORE strict. It feels weird to add the check here, but I think we should: loop over them, and check for scalar values. If not scalar, throw an exception.

But the error message is key, because there are two possibilities:

A) I am trying to set an attribute, but I accidentally passed something weird
B) I am trying to pass a real piece of data, but I typo'ed the data or the property name, so there's a mismatch.

Probably (B) is more common. But if we throw the exception here, we can address both in the same message.

src/TwigComponent/src/HasAttributesTrait.php
kbond: Yes, this also gives the option to use the attributes sub-key to set attributes if you have a conflict with a property on your component (ie `$style`).

weaverryan: That makes sense. The downside is a small, potential wtf, if you're wanting, for some reason, to work with the `attributes` key on some pre/post mount hook or even by adding an `$attributes` property or argument to your `mount()` method. 

To play devil's advocate:

> this also gives the option to use the attributes sub-key to set attributes if you have a conflict with a property on your component

That could *technically* be done just via line 41 if it were `array_merge($data['attributes'], $data)`, right? The "renaming" to `_attributes` is technically accomplishing a separate goal.

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -264,6 +264,40 @@ component use a ``PreMount`` hook::
...
+        }
+        // ...
+    }
+
+.. note::
+
+    If your component has multiple ``PostMount`` hooks, and you'd like to control
+    the order in which they're called, use the ``priority`` attribute parameter:
+    ``PostMount(priority: 10)`` (higher called earlier).
+>>>>>>> 0cd3131 (add `PostMount` component hook)
```
weaverryan: Sorry again about causing the conflict ;)

src/TwigComponent/src/HasAttributesTrait.php
kbond: I can't quite tell, are you thinking to remove the PreMount hook? Or to create the object in the PreMount hook and again in the PostMount hook (merging with the one created earlier)?

src/TwigComponent/src/Resources/doc/index.rst
kbond: haha, good catch!

src/TwigComponent/src/HasAttributesTrait.php
kbond: I refactored this to remove the PreMount hook - [much simpler now](https://github.com/symfony/ux/pull/220/files#diff-f384809c2d40307baac83a415e98f245d60906abc92c3f605700dc61d7c86f1d). Let me know if this is easier to understand.

src/TwigComponent/src/HasAttributesTrait.php
kbond: Added in 3123f351c4e9fcd0a4b2c54b0a9ad2a0053f70bd.

kbond:
The `LiveComponent` failing test will be fixed once #241 is merged/rebased. The cs fixer will pass once #240 is merged/rebased.

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -264,6 +264,39 @@ component use a ``PreMount`` hook::
...
+        }
+        // ...
+    }
+
+.. note::
+
+    If your component has multiple ``PostMount`` hooks, and you'd like to control
+    the order in which they're called, use the ``priority`` attribute parameter:
+    ``PostMount(priority: 10)`` (higher called earlier).
+
```
weaverryan, line 297: Only missing part now is docs about the actual attribute feature :)

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -377,6 +410,135 @@ you can store its result on a private property:
...
+    use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+    use Symfony\UX\TwigComponent\HasAttributesTrait;
+
+    #[AsTwigComponent('my_component')]
+    class MyComponent
+    {
+        use HasAttributesTrait;
+    }
+
+.. code-block:: twig
```
weaverryan, line 438: Above this, I'd say:

> Then render the attributes on the root element:

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @param array<string, string> $attributes
+     */
+    public function __construct(array $attributes)
+    {
+        $this->attributes = $attributes;
+    }
+
+    public function __toString(): string
```
weaverryan, line 29: were you still thinking about adding validation for non-scalar values? (Not sure if it should live here, just using this as a spot for the comment)

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -377,6 +410,135 @@ you can store its result on a private property:
           }
       }
 
+Component Attributes
+--------------------
```
weaverryan, line 412: We probably need something in the Live component docs. As it stands now, I think you'd need `{{ init_live_component(this) }} {{ attributes }}` on the root element, right?

src/TwigComponent/src/ComponentAttributes.php
kbond: There is validation but in the `HasAttributesTrait`. Not ideal but improves the error message if misused. I didn't think we should validate twice but what do you think?

src/TwigComponent/src/Resources/doc/index.rst
kbond: Correct, as discussed in slack, I'll add the system to append the live stuff to these attributes in this PR.

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -382,6 +415,117 @@ you can store its result on a private property:
...
+
+A common need for components is to configure/render attributes for the
+root node. You can enable this feature by having your component use
+the ``HasAttributesTrait``.  Attributes are any data passed to ``component()``
+that cannot be mounted on the component itself. This extra data is added
+to a ``ComponentAttributes`` object that lives as a public property on your
+component (available as ``attributes`` in your component's template).
+
+To use, add the HasAttributesTrait to your component:
```
```suggestion
To use, add the `HasAttributesTrait` to your component:
```

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -382,6 +415,117 @@ you can store its result on a private property:
...
+    <div class="foo" style="color:red">
+      My Component!
+    </div>
+
+Defaults & Merging
+~~~~~~~~~~~~~~~~~~
+
+In your component template, you can set defaults that are merged with
+passed attributes. The passed attributes override the default with
+the exception of *class*. For class, the defaults are prepended:
```
```suggestion
the exception of *class*. For `class`, the defaults are prepended:
```

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -382,6 +415,117 @@ you can store its result on a private property:
...
+    {{ component('my_component', { style: 'color:red' }) }}
+    {{ component('my_component', { class: 'foo', type: 'submit' }) }}
+
+    {# renders as: #}
+    <div class="bar" style="color:red">
+      My Component!
+    </div>
+    <div class="bar foo" type="submit">
+      My Component!
+    </div>
```
weaverryan: I would group the "renders as" right below each example. So, show the first example, then immediately show how it renders below. Then, blank line, then show the 2nd example and how it renders.

Also, if we changed the text from `My Component!` to something short like `Save`, then probably we could put the component template onto 1 line, and then also show the "renders as:" on a single line.

src/TwigComponent/src/Resources/doc/index.rst
kbond: This is now available in this PR. One neat thing is the live component doesn't need the trait for this to work. If it has the trait, the live attributes are merged. `<div {{ attributes }}>` always works.

Question is, should I make the attributes always available in non-live component templates (this is possible too)? It wouldn't contain any data in this case but we'd basically be able to say: "If you want to be able to pass attributes when mounting your component, add the trait to your component".

Sooo close to attributes being "native"!

weaverryan: Wow, that's cool!

> Question is, should I make the attributes always available in non-live component templates (this is possible too)?

I say yes! It would make converting to a live component easier, because you'd already have the `{{ attributes }}`.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -0,0 +1,41 @@
...
+use Symfony\UX\TwigComponent\EventListener\PreRenderEvent;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class AddLiveAttributesSubscriber implements EventSubscriberInterface
+{
+    public function __construct(private LiveComponentRuntime $runtime)
+    {
+        // todo make lazy?
```
weaverryan: ... quite possibly, yes. If I'm only rendering non-live components on a page, this would trigger a lot of instantiation - I followed the instantiation chain a little bit.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function onPreRender(PreRenderEvent $event): void
+    {
+        if (!isset($event->config['live'])) {
```
weaverryan:
```suggestion
        if (!i$event->config['live'] ?? false) {
```

Would allow for `live: false`, not that this is a use-case.

src/LiveComponent/src/Twig/LiveComponentRuntime.php
```diff
@@ -24,51 +25,51 @@
...
+        $params = ['component' => $this->factory->configFor($name ?? $component)] + $data;
 
         return $this->urlGenerator->generate('live_component', $params);
     }
 
-    private function nameFor(object $component, string $name = null): string
+    public function getLiveAttributes(object $component, array $config): ComponentAttributes
     {
-        return $this->factory->configFor($component, $name)['name'];
+        if (!isset($config['live'])) {
```
weaverryan:
```suggestion
        if (!$config['live'] ?? false) {
```

src/TwigComponent/src/EventListener/PreRenderEvent.php
```diff
@@ -0,0 +1,28 @@
...
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class PreRenderEvent extends Event
+{
+    public function __construct(
+        public object $component,
+        public string $template,
+        public array $context,
+        public array $config
+    ) {
```
weaverryan, line 33: I'd prefer private with getters on this. It still feels like it gives us more BC flexibility in the future.

And we should document this stuff - especially those last 2 array arguments, which look very generic (i.e. "what are those?")

src/LiveComponent/src/Twig/LiveComponentRuntime.php
```diff
@@ -24,51 +25,51 @@
...
-
-        $ret = sprintf(
-            'data-controller="live" data-live-url-value="%s" data-live-data-value="%s"',
-            twig_escape_filter($env, $url, 'html_attr'),
-            twig_escape_filter($env, json_encode($data, \JSON_THROW_ON_ERROR), 'html_attr'),
-        );
-
-        if (!$this->csrfTokenManager) {
-            return $ret;
+        if (!isset($context['_component_config'])) {
```
weaverryan: Correct me if I'm wrong, but if we removed `renderLiveAttributes()`, then we would NOT need to include `_component_config` here, right? That would be a nice cleanup - it looks ugly and the `AddLiveAttributesSubscriber` does not need it.

src/TwigComponent/src/EventListener/PreRenderEvent.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class PreRenderEvent extends Event
+{
+    public function __construct(
+        public object $component,
+        public string $template,
+        public array $context,
```
weaverryan: Maybe `$variables` instead? These are literally the Twig variables I believe

src/LiveComponent/src/Twig/LiveComponentRuntime.php
```diff
@@ -24,51 +25,51 @@
...
 
-    private function nameFor(object $component, string $name = null): string
+    public function getLiveAttributes(object $component, array $config): ComponentAttributes
     {
-        return $this->factory->configFor($component, $name)['name'];
+        if (!isset($config['live'])) {
+            throw new \LogicException(sprintf('"%s" is not a Live Component.', $config['class']));
+        }
+
+        $url = $this->urlGenerator->generate('live_component', ['component' => $config['name']]);
```
weaverryan: It seems like this and `getComponentUrl()` should share some code, right? Is that method still kind of a TODO to fix?

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -77,7 +77,11 @@ public function onKernelRequest(RequestEvent $event): void
             throw new NotFoundHttpException(sprintf('Component "%s" not found.', $componentName), $e);
         }
 
-        $request->attributes->set('_component_template', $config['template']);
+        if (!isset($config['live'])) {
```
weaverryan:
```suggestion
        if (!$config['live'] ?? false) {
```

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -46,7 +46,7 @@ A real-time product search component might look like this::
 .. code-block:: twig
 
     {# templates/components/product_search.html.twig #}
-    <div {{ init_live_component(this) }}>
+    <div {{ init_live_component() }}>
```
weaverryan: I think we go all the way and change this to use `{{ attributes }}`

src/TwigComponent/src/ComponentAttributes.php
weaverryan: It's good there - I forgot about it

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -36,38 +36,34 @@ public function __construct(ServiceLocator $components, PropertyAccessorInterfac
     }
 
     /**
-     * @param string|object $component Component name as string or component object
+     * @param string|object $name Component name, component object or component FQCN
      */
-    public function configFor($component, string $name = null): array
+    public function configFor(string|object $name): array
```
weaverryan: What about returning some new class - e.g. `ComponentConfig`?

This would make typing on the `PreRenderEvent` nicer. But it's also motivated by an urge to (as a sanity check) make sure that if a user needs to "do crazy things manually", their code looks sane.

I'm looking and thinking about what it would look like if someone wanted to render a component manually... and maybe also pass their own custom custom. For example:

```php
$component = $componentFactory->create('alert', ['message' => 'foo']);
$config = new ComponentConfig('components/alert.html.twig', AlertComponent::class, isLive: true);

echo $componentRenderer->render($component, $config);
```

src/TwigComponent/src/HasAttributesTrait.php
```diff
@@ -0,0 +1,61 @@
...
+    #[LiveProp(hydrateWith: 'hydrateAttributes', dehydrateWith: 'dehydrateAttributes')]
+    public ComponentAttributes $attributes;
+
+    public function setAttributes(array|ComponentAttributes $attributes): void
+    {
+        $this->attributes = \is_array($attributes) ? new ComponentAttributes($attributes) : $attributes;
+    }
+
+    #[PostMount(priority: -1000)]
+    #[PostHydrate]
```
weaverryan: Why also on PostHydrate? This method may deserve some phpdoc explanation

weaverryan (commented):
I really like this version!

src/TwigComponent/src/EventListener/PreRenderEvent.php
kbond: I made `template` and `context` (which is the vars sent to the template) public because they have the option to me changed. So these would need setters. I have been trying not to create _anemic_ getters/setters for these type of properties but can if you'd like. Getters/setters _would_ give the ability to better document "what they do".

For config, private with just a getter makes sense as this value is not meant to be changed.

src/LiveComponent/src/Twig/LiveComponentRuntime.php
kbond: Yes, if we decide to remove `init_live_component()` completely (I'm in favor of this), I think we should include this logic in the subscriber itself.

src/LiveComponent/src/Twig/LiveComponentRuntime.php
kbond: Yes, that twig function needs a big overhaul but wanted to save for another PR.

src/LiveComponent/src/Resources/doc/index.rst
kbond: Are you in favor of dropping `init_live_component()` entirely?

src/TwigComponent/src/ComponentFactory.php
kbond: Yes, this is on my radar!

src/TwigComponent/src/HasAttributesTrait.php
kbond: Heh, this took me while to figure out... When dehydrating a component with empty attributes, there is no attribute key in the dehydrated array (it's filtered out). In this case, when re-hydrating, the `hydrateAttributes()` is never called so the `$attributes` property is never instantiated. The `PostHydrate` ensures this method is always called after hydrating.

weaverryan: > When dehydrating a component with empty attributes, there is no attribute key in the dehydrated array (it's filtered out).

What filters it out exactly? This still feels like maybe we're covering up something that doesn't work quite right...

src/TwigComponent/src/EventListener/PreRenderEvent.php
kbond: To further on my explanation, manipulating, say `$context`, would be more verbose:

```php
$context = $event->getContext();
$context['foo'] = 'bar';
$event->setContext($context);

// opposed to
$event->context['foo'] = 'bar';
```

Additionally, in PHP 8.1, I'd make `$config` readonly.

This all being said, if you think we should follow established patterns, I'm totally fine with that. The BC flexibility _is_ an important factor.

src/LiveComponent/src/Resources/doc/index.rst
kbond: Also, should I update all the sample components to include `HasAttributesTrait`?

src/TwigComponent/src/EventListener/PreRenderEvent.php
kbond: I chose `$context` because it matches the `$twig->render()` variable name. Also, I noticed in Symfony, `TemplatedEmail`, for instance, uses `$context`. I don't have strong opinion on this though.

src/TwigComponent/src/ComponentFactory.php
kbond: Are you thinking just a simple object?

```php
class ComponentConfig
{
    public function __construct(private array $config) {}

    public function get(string $key, mixed $default = null): mixed {}
}
```

Maybe we can call it `ComponentMetadata` and use as a mounting point for other things we've talked about.

src/TwigComponent/src/HasAttributesTrait.php
kbond: Couldn't recreate the problem so I removed.

src/LiveComponent/src/Twig/LiveComponentRuntime.php
weaverryan: Let's remove it completely please :)

src/TwigComponent/src/ComponentFactory.php
weaverryan: Yup, simple object, exactly :). I would prefer to have as many named args in the constructor as we can, before having the final "everything else" `$config`, but no string opinion. The specific args just make the class easier to understand (e.g. `$template`).

`ComponentMetadata` sounds excellent.

src/TwigComponent/src/EventListener/PreRenderEvent.php
weaverryan: I wonder why `$context` is used for those places? In Twig, I can't think of ANY other purpose for those things than `$variables`. It feels more clear to me, so small preference towards variables.

src/LiveComponent/src/Resources/doc/index.rst
weaverryan: Yes to dropping entirely!

> Also, should I update all the sample components to include HasAttributesTrait?

Sure, why not :)

src/TwigComponent/src/EventListener/PreRenderEvent.php
weaverryan: Bah, no strong opinion and, obviously, things are changing with newer php versions. I think I'd like to stay "boring" for the time being. For context, (or variables, whatever we call it), we could also have a `addVariable()` method. This is all kind of academic, since this event won't be seen by many.

weaverryan (approved):

weaverryan:
Thank you Kevin!

seb-jean:
Hi,

I wanted to use a `null` value but I got the following error:

`An exception has been thrown during the rendering of a template ("Unable to use "value" (null) as an attribute. Attributes must be scalar. If you meant to mount this value on your component, make sure this is a writable property.").`

```twig
{{ component('input', {type: 'text', name: 'query', value: form.address.vars.data }) }}
```

`form.address.vars.data` can be `null` at times.

I saw that it was this error that is affected: https://github.com/symfony/ux-twig-component/blob/2.x/src/HasAttributesTrait.php#L49

kbond:
Ah, good catch, so you'd expect the html to be:

```html
<input type="text" name="query" value="" />
```

seb-jean:
@kbond Exactly

seb-jean:
Something else to improve (maybe):
Sometimes, we don't need to provide a value like for the autofocus property.
How could we do to have just `autofocus` instead of `autofocus=""` ?
https://developer.mozilla.org/fr/docs/Web/HTML/Global_attributes/autofocus

kbond:
Another good catch, @weaverryan and I were discussing and we think null or empty string should mean _render the attribute name only_:

```twig
{{ component('input', {type: 'text', name: 'query', value: null, autofocus: null }) }}

<input type="text" name="query" value autofocus />
```

WDYT?

seb-jean:
Yes, `null` is better I think.
I don't see any other solutions anyway.

kbond:
You're thinking?

```twig
{{ component('input', {type: 'text', name: 'query', value: '', autofocus: null }) }}

<input type="text" name="query" value="" autofocus />
```

There are some limitations to doing this when working with live components. But we are discussing some solutions.

seb-jean:
Yeah

PR #221, opened 2022-01-14
[LiveComponent] Fix expanded choices and checkboxes in forms

Lustmored:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | 
| License       | MIT

When using live component with forms that contains expanded choices list (rendered as a list of radio buttons) extracting form values returns array of nulls (one per choice), while frontend expects single value.

Therefore after calling first `update` and then trying to call update on expanded field controller throws (`The property used in data-model="${propertyPath}" was never initialized. Did you forget to add exposed={"${lastPart}"} to its LiveProp?`).

This simple change fixes the issue. I have also added simple test case for a few basic form types.


src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -166,7 +166,13 @@ private function extractFormValues(FormView $formView): array
...
         foreach ($formView->children as $child) {
             $name = $child->vars['name'];
-            if (\count($child->children) > 0) {
+            if (
+                (
+                    !\array_key_exists('expanded', $child->vars)
+                    || !$child->vars['expanded']
+                )
+                && \count($child->children) > 0
+            ) {
```
weaverryan: This should work, is a bit simpler:

```
if (!($child->vars['expanded'] ?? false) && \count($child->children) > 0)
```

src/LiveComponent/tests/Unit/Form/ComponentWithFormTest.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+class ComponentWithFormTest extends TypeTestCase
+{
+    public function testFormValues(): void
+    {
+        $component = new Component6($this->factory->createBuilder());
+
+        $values = $component->getFormValues();
+        $this->assertSame(
+            $values,
```
weaverryan:
```suggestion
            $component->getFormValues(),
```

Then no `$values` variable needed above.

src/LiveComponent/tests/Unit/Form/ComponentWithFormTest.php
```diff
@@ -0,0 +1,41 @@
...
+                'text' => '',
+                'textarea' => '',
+                'range' => '',
+                'choice' => '',
+                'choice_expanded' => '',
+                'checkbox' => null,
+                'file' => '',
+                'hidden' => '',
+            ]
+        );
```
weaverryan, line 50: Swap the args to `assertSame` - the expected vs actual are in the wrong spot

src/LiveComponent/tests/Fixture/Component/Component6.php
```diff
@@ -0,0 +1,70 @@
...
+                    'bar' => 2,
+                ],
+            ])
+            ->add('choice_expanded', ChoiceType::class, [
+                'choices' => [
+                    'foo' => 1,
+                    'bar' => 2,
+                ],
+                'expanded' => true,
+            ])
```
weaverryan: What about a `ChoiceType` with `expanded` true AND `multiple` true (i.e. checkboxes)? Is that correctly handled? Or does the check for `$child->vars['expanded']` mess up that case? I'm genuinely just not sure - I'd love it if you could verify by testing that in a real project.

weaverryan (commented):
A few comments - but this is great! Especially the test - that fixes a bit missing hole.

src/LiveComponent/tests/Fixture/Component/Component6.php
Lustmored: Thanks for quick review! I've followed your suggestions in other places and allowed myself to resolve them.

You are probably right about `multiple` option - I have missed that. But when experimenting with component with form I've found that `CheckboxType` was also misbehaving (haven't got enough time yet to investigate it or write a test case for bug report), therefore my intuition tells me that `expanded` and `multiple` (that expands into multiple checkboxes) isn't working right now anyways.

My intent is to spend my OSS Friday next week to investigate problems with checkboxes and either fix them or provide meaningful bug report. I can do that as part of this PR or the next one if you prefer to have at least fix for `expanded` merged earlier :)

weaverryan: Thank you! 

> My intent is to spend my OSS Friday next week to investigate problems with checkboxes and either fix them or provide meaningful bug report.

This sounds wonderful. Tbh, I believe I wrote this original code that has the bug(s) in it. I knew I needed to handle the checkbox case, but I'm not certain that I dove deeply (or thought deeply enough) about all of these `expanded` and `multiple` situations. Someone needs to dive deep and look at the whole problem. If you can do that, it would be *wonderful*.

> I can do that as part of this PR or the next one if you prefer to have at least fix for expanded merged earlier :)

Let's get everything fixed up on this PR :) 

Thanks!

src/LiveComponent/tests/Fixture/Component/Component6.php
```diff
@@ -0,0 +1,70 @@
...
+use Symfony\Component\Form\FormInterface;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\ComponentWithFormTrait;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+/**
+ * @author Jakub Caban <kuba.iluvatar@gmail.com>
+ */
+#[AsLiveComponent('component6')]
+class Component6 extends AbstractController
```
norkunas: Conflicts conflicts :D
https://github.com/symfony/ux/pull/218/files#diff-26b9f4a13fdb743a0f3fb04b6b3606c1c6e3ba47c4d92e58fd0a25aa5c8df7f7

kbond: First merge wins!

weaverryan: Haha, I noticed that too when reviewing. Maybe we should be a bit more creative with our test class name 😆

Lustmored: I think I will just rename it to `FormComponent1` or something along those lines. Not really creative, but at least it will separate concerns a bit :smile:

src/LiveComponent/tests/Fixture/Component/Component6.php
Lustmored: I have updated PR with fixes for single checkboxes (their `null` value was getting lost after any update) and also `expanded` and `multiple` choice fields.

As it is right now all basic data types (except `FileType`, which is another story) that I've tested are working as intended.

To check out how checkbox inputs are working I used https://github.com/weaverryan/live-demo - as it is right now writing anything in registration form in e-mail first and then clicking checkbox throws in JS. With this PR it synchronizes data between frontend and backend.

To see `expanded`+`multiple` in action you can use the same registration form from my fork: https://github.com/Lustmored/live-demo

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -126,11 +126,7 @@ public function getFormName(): string
 
     public function getFormValues(): array
     {
-        if (null === $this->formValues) {
-            $this->formValues = $this->extractFormValues($this->getForm());
-        }
-
-        return $this->formValues;
+        return $this->formValues = $this->extractFormValues($this->getForm(), $this->formValues ?? []);
```
weaverryan: This is an important change - it's repeated in a different form (as you know) in #234.

Can you explain the motivation behind this change for YOUR Pr. Locally, if I revert this change, all the tests still pass. So whatever motivated this isn't covered by the tests, at the very least.

I also tried, inspired by your comment about lost data on #234, this code (so, effectively the same as #234), and the tests also pass over here.

```php
return $this->formValues = $this->extractFormValues($this->getForm(), []);
```

src/LiveComponent/assets/src/set_deep_data.ts
```diff
@@ -1,5 +1,5 @@
 // post.user.username
-export function setDeepData(data, propertyPath, value) {
+export function setDeepData(data, propertyPath, value, modelValue = null) {
```
weaverryan: It looks like you now ALWAYS pass `modelValue`, so does it need to be optional?

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -110,13 +110,21 @@ export default class extends Controller {
...
         this._updateModelFromElement(event.target, value, true);
     }
 
     updateDefer(event: any) {
-        const value = this._getValueFromElement(event.target);
+        let value = this._getValueFromElement(event.target);
+
+        if (event.target.type === 'checkbox' && !event.target.checked) {
+            value = null;
+        }
```
weaverryan: Should we move this `if` logic into `_updateModelFromElement()`?

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -243,8 +251,9 @@ export default class extends Controller {
      * @param {boolean} shouldRender Whether a re-render should be triggered
      * @param {string|null} extraModelName Another model name that this might go by in a parent component.
      * @param {Object} options Options include: {bool} dispatch
+     * @param {any} modelValue Original HTML element value (for handling collection checkbox fields)
      */
-    $updateModel(model: string, value: any, shouldRender = true, extraModelName: string | null = null, options: any = {}) {
+    $updateModel(model: string, value: any, shouldRender = true, extraModelName: string | null = null, options: any = {}, modelValue: any = null) {
```
weaverryan: Your logic is all sound - I was able to play with your version of the demo and it all works perfectly (I'd like to incorporate the newsletter into the field after this is merged).

But, I think we need to reorganize the code a little bit. Let's look and think about JUST this method. It's called `$updateModel`. And so, it should be that simple: we pass it the model name and value, and it takes care of the rest. And so, if you think about calling this method programmatically, if we were to update the `newsletter` field, it should look like this:

```
liveController.$updateModel('registration_form[newsletter][]', ['mailing']);
```

Notice I pass in the model name - `registration_form[newsletter][]` - and then the FINAL value - `['mailing']`.

So what I'm saying is this:

A) Remove the `modelValue` argument from this method.
B) Instead, in `_updateModelFromElement`, we need to handle the logic of calculating the correct value from there. For example, if we determine that I just unchecked `notification`, we would grab (all in `_updateModelFromElement()`) the current value for the `registration_form[newsletter][]` model and *remove* `notification`. Basically, your logic from `set_deep_data.ts` would move to `_updateModelFromElement()`. This will give us the final value, and we can call `$updateModel` like normal.
C) All the changes in `set_deep_data.ts` can be removed, because they will be done in `_updateModelFromElement()`.

Let me know if that makes sense - we can also chat on Symfony's Slack.

weaverryan: Btw, we'll also need a test for the checkboxes - basically you'll want to replicate what you did in your live demo in our test suite here. 

The proper place for the test is `model.test.ts`. You should create my own "template" inside the test itself like is done in `standardizes user[firstName] style models into post.name` with the two checkboxes. You can then "check" one of those boxes, and assert (via that `mockRerender()` function) that the correct query params are sent. You might even create a couple of tests, like starting with both checked, and unchecking one. Let me know if you need some guidance - those tests can be weird at first.

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -161,14 +157,17 @@ private function submitForm(bool $validateAll = true): void
...
-                $values[$name] = $this->extractFormValues($child);
+            if (!($child->vars['expanded'] ?? false) && \count($child->children) > 0) {
+                $values[$name] = $this->extractFormValues($child, $values[$name] ?? []);
+
+                continue;
+            }
 
+            if (null !== ($values[$name] ?? null)) {
                 continue;
             }
```
weaverryan: What is this if statement for? If I comment it out, the tests still pass. So if it IS needed for something, we need to expand the test suite the cover it.

weaverryan (commented):
This is very nice work on a super complex problem which, thanks to this PR, I can finally wrap my head around :). A few explanations, tests and refactorings are needed - but this is well on its way.

Thanks!

src/LiveComponent/src/ComponentWithFormTrait.php
Lustmored: Of course. The scenario is:

- initial form (with checkbox field `checkbox` and any other field) is rendered in initial state -> `$formValues` contains `'checkbox' => null` and that is passed to controller,
- when other field is updated controller is sending new data, but this data (as it should) does not contain `checkbox` value. Thus `$formValue` does not contain `checkbox` key at all and this is sent back to controller,
- now controller doesn't have information about state of `checkbox` value and whenever update is called on the checkbox it throws.

This change (along with changes within `extractFormValues`) fills back `$formValue` with initial form values that are missing (in fact probably only checkboxes, but maybe there are other similar scenarios - forms are difficult :) ).

I will work on those functional tests probably next week and will try to cover them. I am trying to work on OSS once a week at work and I didn't wrap my head around those interactive tests yet.

src/LiveComponent/assets/src/set_deep_data.ts
Lustmored: It does not. I have started with signature and worked up from that (working on unfamiliar JavaScript code is always demanding 😄) and missed that fact 👍

src/LiveComponent/src/ComponentWithFormTrait.php
Lustmored: It is here to make sure we don't overwrite changed fields with initial form data. I will surely write tests for that 👍

src/LiveComponent/assets/src/live_controller.ts
Lustmored: It all makes perfect sense. I will probably come back with refactored version (hopefully with tests) next week and I'll make sure to join Sf Slack in case I need someone to explain to me what happens in those tests 👍

src/LiveComponent/src/ComponentWithFormTrait.php
Lustmored: On a second thought it might not work well with checkboxes that are initially checked. Noting for myself to take that situation into consideration within tests.

Lustmored: In fact it wasn't needed there at all :)

src/LiveComponent/assets/src/live_controller.ts
Lustmored: I have done refactoring part. As you know my tests (at least JS parts) are not functioning as expected yet, but hopefully with some help I'll get there.

During refactor I found it useful to extract traversing deepData out of `setDeepData` so that I could easily access data without code duplication. In addition I had to make a small change in `normalizeModelName` as trailing `[]` should be stripped during normalization unconditionally IMO - in HTML forms `[]` at the end of the name is only used to mark form elements that should be sent as an array.

Other changes are I think in line with what you suggested.

src/LiveComponent/src/ComponentWithFormTrait.php
Lustmored: I have added a simple functional test for backend that I think covers all of PHP files changes.

weaverryan:
@Lustmored check out the test modifications I made here - it was tricky: https://gist.github.com/weaverryan/5d35bf94a43c888970cb7ca2dade2d9f/revisions

Notes:

* Ignore my `console.log()` in there ;) 
* The initial `data` needed to be more initialized for me, else I got a deep JS err: https://gist.github.com/weaverryan/5d35bf94a43c888970cb7ca2dade2d9f/revisions#diff-35b353a0a2e3abf41b7f7bb8b9e0db7e2dd4b9d9bee88226176aac7f9e87ddf7R236
* These waits - https://gist.github.com/weaverryan/5d35bf94a43c888970cb7ca2dade2d9f/revisions#diff-35b353a0a2e3abf41b7f7bb8b9e0db7e2dd4b9d9bee88226176aac7f9e87ddf7L243 - didn't make sense and weren't doing anything. When you check, that's instant. You were trying to wait for the re-render to finish, but you can't do it this way. That's why I added the "Checkbox 2 is checked" data to the template. We would only see that after a re-render. So that's the way to check to see if the render has finished

Cheers!

Lustmored:
> @Lustmored check out the test modifications I made here - it was tricky: https://gist.github.com/weaverryan/5d35bf94a43c888970cb7ca2dade2d9f/revisions
> 
> Notes:
> 
>     * Ignore my `console.log()` in there ;)
> 
>     * The initial `data` needed to be more initialized for me, else I got a deep JS err: https://gist.github.com/weaverryan/5d35bf94a43c888970cb7ca2dade2d9f/revisions#diff-35b353a0a2e3abf41b7f7bb8b9e0db7e2dd4b9d9bee88226176aac7f9e87ddf7R236
> 
>     * These waits - https://gist.github.com/weaverryan/5d35bf94a43c888970cb7ca2dade2d9f/revisions#diff-35b353a0a2e3abf41b7f7bb8b9e0db7e2dd4b9d9bee88226176aac7f9e87ddf7L243 - didn't make sense and weren't doing anything. When you check, that's instant. You were trying to wait for the re-render to finish, but you can't do it this way. That's why I added the "Checkbox 2 is checked" data to the template. We would only see that after a re-render. So that's the way to check to see if the render has finished
> 
> 
> Cheers!

Thanks a lot! That looks well - I see you had the same problem as I did with mocks :) I think I know everything I need now to finish those tests and will probably do so by this Friday :+1: 

If you could check out if my changes in set_deep_data.ts sound reasonable and acceptable - that'd help me push things a bit faster. I have another approach in my head, but this one looked easiest from implementation perspective :)

norkunas:
Tried this PR to check if it would fix when using CheckboxType (expanded: false, multiple: true), but it fixed only js error that the prop is not exposed, but still it sends for me always the first selected choice instead of multiple selected values

Lustmored:
@norkunas Can you please share an example?

Also - don't you mean `ChoiceType`? AFAIK `CheckboxType` does not have `multiple` option. In that case I must admit I have not yet worked with `expanded: false, multiple: true` as I haven't used that in a while and probably forgot about it. Will be great addition to the tests :+1:

norkunas:
yeah sorry, choice indeed :)

norkunas:
Example - CountryType (expanded: false, multiple: true) :)

Lustmored:
> yeah sorry, choice indeed :)

Then thank you for the report - I must've missed that case. I will work on it next time I sit to this PR :+1:

norkunas:
Hey @Lustmored,
I've checked why it doesn't return proper values and it looks like that it's by design: change/input events returns only the first selected value like it's not a multiselect.
We need to use https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedOptions to get all selected values :)

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -206,7 +202,31 @@ export default class extends Controller {
...
+                    const index = currentValue.indexOf(this._getValueFromElement(element));
+                    if (index > -1) {
+                        currentValue.splice(index, 1);
+                    }
+                } else {
+                    currentValue.push(value);
+                }
+            }
+            value = currentValue;
+        }
```
weaverryan, line 222: If I understand things correctly, this handles a case where, for example, we have an `<input type="checkbox" name="newsletters[]" value="email">`. And all we know in this method is that, for example, that the `email` checkbox was changed. And so, the purpose of this block is to try to figure out what the true current value of `newsletters` is and then add/remove `email` (based on if it was checked or unchecked). Is that correct?

If so, I'd love to isolate this logic to an external module/file - like in `set_deep_data.ts`. It would help because we could give the method a name (the logic is very unclear right now: it's just complex, so it's hard to see what it's doing) and we could even unit test it.

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -206,7 +202,31 @@ export default class extends Controller {
             throw new Error(`The update() method could not be called for "${clonedElement.outerHTML}": the element must either have a "data-model" or "name" attribute set to the model name.`);
         }
 
-        this.$updateModel(model, value, shouldRender, element.hasAttribute('name') ? element.getAttribute('name') : null);
+        if (element instanceof HTMLInputElement && element.type === 'checkbox' && !element.checked) {
```
weaverryan: I was going to say the same thing as @norkunas: I think the only other case we need to check is for the `<select multiple` (ChoiceType multiple=true, expanded=false).

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -164,9 +160,10 @@ private function submitForm(bool $validateAll = true): void
     private function extractFormValues(FormView $formView): array
     {
         $values = [];
+
         foreach ($formView->children as $child) {
             $name = $child->vars['name'];
-            if (\count($child->children) > 0) {
+            if (!($child->vars['expanded'] ?? false) && \count($child->children) > 0) {
```
weaverryan: this deserves a small `// comment` above to help understand, as it's complex. I'm actually not entirely sure what this should be doing - I think it is (correctly) fixing ChoiceType with expanded => true, but it's very complex.

src/LiveComponent/assets/src/set_deep_data.ts
```diff
@@ -64,9 +75,11 @@ export function doesDeepPropertyExist(data, propertyPath) {
  */
 export function normalizeModelName(model) {
     return model
+        .replace(/\[]$/, '')
```
weaverryan, line 81: Can you add a test to the bottom of `set_deep_data.test.ts` to test & explain what this does?

weaverryan (commented):
Thanks for you work on this. It's a super complex topic, but we're getting closer!

src/LiveComponent/src/ComponentWithFormTrait.php
weaverryan: Actually, I stole your logic over in #264: https://github.com/symfony/ux/pull/264/files#diff-9d445f9954f64a41536bb9470d1abcea4986100b295a038022b55ba192a85cf7R187

It DOES fix a critical bug - nice work! Please keep this here... and also, I didn't include a test for that functionality in my PR, but I believe you have one here.

Lustmored: Right, there is already a test in my PR for that :+1:

src/LiveComponent/assets/src/set_deep_data.ts
Lustmored: I will paste here comment as I'm adding it:

> Names ending in "[]" represent arrays in HTML.
> To get normalized name we need to ignore this part.
> For example: "user[mailing][]" becomes "user.mailing" (and has array typed value)

I have also added simple test case

src/LiveComponent/assets/src/live_controller.ts
Lustmored: I just did as you asked - great call! Resulting function is much simpler and cleaner in my opinion and it was extremely easy to unit test it. I have already incorporate "select multiple" case within it, so it should start working soon.

Still some test to write and I'll manually check how it behaves too, but for the first time I can see the finish line :smile:

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
...
+            if (index === -1) {
+                currentValue.push(value);
+            }
+        } else {
+            // Remove value from an array
+            if (index > -1) {
+                currentValue.splice(index, 1);
+            }
+        }
+    } else if (element instanceof HTMLSelectElement) {
```
norkunas: Should there be a check for multiple attribute?

Lustmored: I don't think so - `getArrayValues` is only called in a situation where return value has to be an array (or null), therefore only logic for extracting values as an array is required.

norkunas: Oh, okay, didn't look through refactored `_updateModelFromElement` method :)

Lustmored:
I think I have finished all refactoring and have added enough tests to cover all cases. The `<select multiple>` became trivial once checkbox logic was in place and has a few tests of it's own right now.

I have tested this on locally modified live twig demo (it is a bit behind with changes here btw.) - my real world project usage is right now a bit too hacky (to make it work without this PR).

I think I'm ready for the next wave of reviews :grinning:

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -2,10 +2,11 @@ import { Controller } from '@hotwired/stimulus';
 import morphdom from 'morphdom';
 import { parseDirectives, Directive } from './directives_parser';
 import { combineSpacedArray } from './string_utils';
-import { setDeepData, doesDeepPropertyExist, normalizeModelName } from './set_deep_data';
+import {setDeepData, doesDeepPropertyExist, normalizeModelName, parseDeepData} from './set_deep_data';
```
weaverryan: minor, watch spacing here

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -2,10 +2,11 @@ import { Controller } from '@hotwired/stimulus';
...
 import { haveRenderedValuesChanged } from './have_rendered_values_changed';
 import { normalizeAttributesForComparison } from './normalize_attributes_for_comparison';
 import { cloneHTMLElement } from './clone_html_element';
+import {getArrayValue} from "./get_array_value";
```
weaverryan: and here

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
...
+    element: HTMLElement,
+    value: string|null,
+    currentValue: any
+): Array<string> {
+    // Enforce returned value is an array
+    if (!(currentValue instanceof Array)) {
+        currentValue = [];
+    }
+
+    if (element instanceof HTMLInputElement && element.type === 'checkbox') {
```
weaverryan: No need for the instanceof: the `element` argument is correctly typed to `HTMLElement`

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Resolve multiple value data from changed HTML element.
+ *
+ * @param element       Current HTML element
+ * @param value         Resolved value of a single HTML element (.value or [data-value])
+ * @param currentValue  Current data value
+ */
+export function getArrayValue(
```
weaverryan: How bout:

```suggestion
export function updateArrayDataFromChangedElement(
```

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Resolve multiple value data from changed HTML element.
```
weaverryan: > Adds or removes a key from an array element based on an array element.

> Given an "array" element (e.g. <input type="checkbox" name="preferences[]" value="email">)
> and the current data for "preferences" (e.g. ["text", "phone"]), this function will add or
> remove the value (e.g. email) from that array (based on if the element (un)checked) and
> return the final, updated array (e.g. ["text", "phone", "email"]).

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param element       Current HTML element
+ * @param value         Resolved value of a single HTML element (.value or [data-value])
```
weaverryan:
```suggestion
 * @param value         The value that should be set or removed from currentValues
```

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
...
+        } else {
+            // Remove value from an array
+            if (index > -1) {
+                currentValue.splice(index, 1);
+            }
+        }
+    } else if (element instanceof HTMLSelectElement) {
+        // Select elements with `multiple` option require mapping chosen options to their values
+        currentValue = Array.from(element.selectedOptions).map(el => el.value);
+    }
```
weaverryan: Is there an "else" case? Is there a case where we hit this function but the element is NOT a checkbox or an `HTMLSelectElement`? 

If there "shouldn't" be any other case, then we should throw an exception. That way, if there IS a case, it can be reported to us.

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
...
+            // Add value to an array if it's not in it already
+            if (index === -1) {
+                currentValue.push(value);
+            }
+        } else {
+            // Remove value from an array
+            if (index > -1) {
+                currentValue.splice(index, 1);
+            }
+        }
```
weaverryan: Minor: return early whenever you can. So:

```js
if (element.type === 'checkbox') {
    const index = currentValue.indexOf(value);

    if (element.checked) {
        // Add value to an array if it's not in it already
        if (index === -1) {
            currentValue.push(value);
        }

        return currentValue;
    }

    // Remove value from an array
    if (index > -1) {
        currentValue.splice(index, 1);
    }

    return currentValue;
}
```

Then the next `else if` just becomes an `if`.

src/LiveComponent/assets/src/get_array_value.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Resolve multiple value data from changed HTML element.
+ *
+ * @param element       Current HTML element
+ * @param value         Resolved value of a single HTML element (.value or [data-value])
+ * @param currentValue  Current data value
+ */
+export function getArrayValue(
+    element: HTMLElement,
+    value: string|null,
+    currentValue: any
```
weaverryan:
```suggestion
    currentValues: any
```

src/LiveComponent/assets/src/set_deep_data.ts
```diff
@@ -14,6 +13,18 @@ export function setDeepData(data, propertyPath, value) {
...
+        currentLevelData,
+        finalData,
+        finalKey,
+        parts
+    }
+}
+
+// post.user.username
+export function setDeepData(data, propertyPath, value) {
+    const {currentLevelData, finalData, finalKey, parts} = parseDeepData(data, propertyPath)
```
weaverryan:
```suggestion
    const { currentLevelData, finalData, finalKey, parts } = parseDeepData(data, propertyPath)
```

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -203,7 +200,25 @@ export default class extends Controller {
...
+        // we need to handle addition and removal of values from it to send
+        // back only required data
+        if (/\[]$/.test(model)) {
+            // Get current value from data
+            const {currentLevelData, finalKey} = parseDeepData(this.dataValue, normalizeModelName(model))
+            const currentValue = currentLevelData[finalKey];
+
+            value = getArrayValue(element, value, currentValue);
+        } else if (
+            element instanceof HTMLInputElement
```
weaverryan, line 216: This line isn't needed

src/LiveComponent/assets/src/live_controller.ts
weaverryan: This is all reading VERY nicely now. It's a complex situation, but I can understand it - nice work!

src/LiveComponent/assets/test/controller/model.test.ts
```diff
@@ -200,6 +200,240 @@ describe('LiveController data-model Tests', () => {
...
+
+        mockRerender({ form: {check1: '1', check2: '1'}}, checkboxTemplate);
+
+        await userEvent.click(check2Element);
+        await waitFor(() => expect(element).toHaveTextContent('Checkbox 2 is checked'));
+
+        expect(controller.dataValue).toEqual({form: {check1: '1', check2: '1'}});
+
+        // assert all calls were done the correct number of times
+        fetchMock.done();
```
weaverryan: This isn't needed anymore. I recently added code above in `afterEach` to do this for us.

src/LiveComponent/assets/test/controller/model.test.ts
```diff
@@ -200,6 +200,240 @@ describe('LiveController data-model Tests', () => {
...
+
+        await userEvent.deselectOptions(selectElement, 'bar');
+
+        await waitFor(() => expect(element).toHaveTextContent('Select: foo bar Option 2 is unselected'));
+
+        expect(controller.dataValue).toEqual({form: {select: []}});
+
+        // assert all calls were done the correct number of times
+        fetchMock.done();
+    });
```
weaverryan, line 417: These tests are GORGEOUS 😍

src/LiveComponent/tests/Fixtures/Form/FormType1.php
```diff
@@ -0,0 +1,83 @@
...
+use Symfony\Component\Form\Extension\Core\Type\RangeType;
+use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Jakub Caban <kuba.iluvatar@gmail.com>
+ */
+class FormType1 extends AbstractType
```
weaverryan: How about we call this `FormWithManyDifferentFieldsType` (and update the component also).

src/LiveComponent/tests/Fixtures/Form/FormType1.php
```diff
@@ -0,0 +1,83 @@
...
+                ],
+                'multiple' => true,
+                'data' => [2],
+            ])
+            ->add('checkbox', CheckboxType::class)
+            ->add('checkbox_checked', CheckboxType::class, [
+                'data' => true,
+            ])
+            ->add('file', FileType::class)
+            ->add('hidden', HiddenType::class)
```
weaverryan: A lot of good field types included here 👍

src/LiveComponent/tests/Fixtures/templates/components/form_component1.html.twig
```diff
@@ -0,0 +1,5 @@
+<div
+        {{ attributes }}
+>
```
weaverryan: Minor, but:

```
<div {{ attributes }}>
```

src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
```diff
@@ -143,4 +144,105 @@ public function testFormRemembersValidationFromInitialForm(): void
...
+            'textarea' => '',
+            'range' => '',
+            'choice' => '',
+            'choice_expanded' => '',
+            'choice_multiple' => ['2'],
+            'select_multiple' => ['2'],
+            'checkbox_checked' => '1',
+            'file' => '',
+            'hidden' => '',
+        ];
```
weaverryan, line 182: This test might be a bit easier to follow if:

A) In the form type, we do NOT use the `data` option on some fields to initialize them.
B) Instead, in the form component, add a `public $initialData` property (it does not need to be live)
C) Inside `instantiateForm()`, use `$this->initialData` as the 2nd arg when creating the form.

If you did this, then in this test, you could do `$component = $factory->create('form_component1', [ ... initial form data ])`. This would be more clear because it would be more obvious why we're asserting that, for example, `choice_multiple` should be equal to `['2']`.

src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
```diff
@@ -143,4 +144,105 @@ public function testFormRemembersValidationFromInitialForm(): void
...
+        $fullBareData = array_merge(
+            $bareForm,
+            [
+                'checkbox' => null,
+            ]
+        );
```
weaverryan: This extra variable is confusing. I believe it is because we want to ASSERT that the `checkbox` field equals null. But when we SEND the data to the server, we want this field to be omitted? Is that right?

If so, are you sure that's correct? With the new way of sending the data as JSON, if I have a form with an unchecked checkbox, the initial data sent to the frontend will store this value as `null` (and also, if I check and uncheck the box, the model will be set to `null`). And we DO now send `null` values to the Ajax calls.

weaverryan (commented):
This is looking GREAT. I added quite a few comments because it's a complex feature and I want to get it perfect. However, I am quite confident now that you've got this fully-working - the JS test suite is particularly awesome. Thank you!

src/LiveComponent/assets/src/get_array_value.ts
Lustmored: This check is to ensure it is `<input>` and not other kind of element (like `<select>` in other path that has no `type` parameter). I can remove it if you prefer other way :)

Lustmored:
Thanks @weaverryan for the review - I will try to work on it during weekend, as it seems like mostly tidying up, so shouldn't be difficult 👍

src/LiveComponent/assets/src/live_controller.ts
Lustmored: Just as in other comment: This check is to ensure it is <input> and not other kind of element (like <select> in other path that has no type parameter). I can remove it if you prefer other way :)

src/LiveComponent/assets/src/get_array_value.ts
Lustmored: This function is called whenever array data is expected (defined by `[]` at the end of model name), so user can write broken code that will trigger it with incorrect element passed. Therefore I'll go with a `throw` in the final step :+1:

src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
Lustmored: Fun fact is - in pure HTML world unchecked checkboxes would not be sent to the server at all and before parsing data as JSON live component was following those trails and it looked like perfect solution.

BUT since data is now sent as JSON and as Symfony Form component is handling nulls as unchecked state (as my updated test shows) I believe it will be simpler to just send all nulls.

This also simplifies this test case a lot by removing variable that probably only I knew why existed :)

Lustmored:
@weaverryan I think I have done most of your change request. I have resolved some of them to ease tracking what was done and what's left. In a few I left comments where explanation was needed or I had doubts about change suggestions :+1:

src/LiveComponent/assets/src/get_array_value.ts
weaverryan: Please leave it - I misread this line as `HTMLElement`!

weaverryan (approved):
WONDERFUL work. You rock @Lustmored  - this was a huge, complex problem and you were a joy to work with. Let's get this merged, and if we missed any edge cases, we'll find out :).

Lustmored:
Thank you @weaverryan :+1: working with you and on this project is also a pleasure, so I surely will be back - I love how enabling components+live-components are architecturally in my projects already. See you around soon :smile:

PR #232, opened 2022-01-20
[TwigComponent] make public component properties available directly

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

This makes all public properties available directly in your component template. Assuming your component has a public property `foo`:

```twig
{# instead of #}
{{ this.foo }}

{# you can use #}
{{ foo }}
```

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -45,7 +45,7 @@ A real-time product search component might look like this::
         <input
             type="search"
             name="query"
-            value="{{ this.query }}"
+            value="{{ query }}"
             data-action="live#update"
         >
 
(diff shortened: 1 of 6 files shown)


src/TwigComponent/README.md
```diff
@@ -128,11 +128,12 @@ class AlertComponent
...
 In the template, the `AlertComponent` instance is available via
-the `this` variable. Use it to render the two new properties:
+the `this` variable and public properties are available directly.
+Use them to render the two new properties:
 
 ```twig
-<div class="alert alert-{{ this.type }}">
-    {{ this.message }}
+<div class="alert alert-{{ type }}">
+    {{ message }}
```
weaverryan: Maybe after this, just to emphasize:

```twig
{# Same as above, but using "this", which is the component object #}
{{ this.message }}
```

weaverryan (commented):
This seems very practical. There are a few more examples in the TwigComponent docs and many examples in LiveComponents. I think we should lean in and prefer the short version 90%+ of the time.

kbond:
Rebased to pickup the new doc format and I think I updated all the applicable `this.` calls in the docs.

weaverryan:
Thanks Kevin!

PR #234, opened 2022-01-21
Always re-extract formValues in LiveComponent

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

When creating a form builder with LiveComponent I've found out that proper values are not populated when form changes.
For example:
  1. Form has many questions;
  1. Question have a `type` field (text/choice/date/datetime, etc) and then dynamically added `options` field based on it's `type`;
  1. Date/datetime has min/max options;
  1. I change type to datetime:
     1. Component renders min/max option inputs with name like `form_builder[questions][0][options][minDateTime][date][year]`;
     1. It gets two levels deeper with `[date][year]` because that how Symfony renders datetime by default;
     1. Trying to change value of an option - get UI LiveController error that it's not exposed;
     1. And that's because formValues were already extracted without that deep nested data.

Not sure about the reason why it was done this way previously. My guess would be for perf? Maybe this could live as some trait option by which `getFormValues` would decide to always re-extract or take from cache?

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -126,9 +126,7 @@ trait ComponentWithFormTrait
 
     public function getFormValues(): array
     {
-        if (null === $this->formValues) {
-            $this->formValues = $this->extractFormValues($this->getForm());
-        }
+        $this->formValues = $this->extractFormValues($this->getForm());
 
         return $this->formValues;
     }


Lustmored:
I was just attacking same problem, but with different angle in #221. I am afraid your solution could lead to loss of filled form values on each render (frontend would still show them, but they could be wrong in the backend). Do you have a test case where we can see the problem described here? And could you check if it still happens on #221 branch?

norkunas:
I don't see any loss of data in my app :)

norkunas:
I can live currently without this as I have a workaround, so if you think your solution is better than I can close

Lustmored:
I am not a person in power here, just trying to understand your problem better to see if it collides with my work in other PR :+1:

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -126,9 +126,7 @@ public function getFormName(): string
 
     public function getFormValues(): array
     {
-        if (null === $this->formValues) {
-            $this->formValues = $this->extractFormValues($this->getForm());
-        }
+        $this->formValues = $this->extractFormValues($this->getForm());
```
weaverryan, line 129: If you were able to put together a failing test, that would be *awesome*. As you know, this change is also in #221 (with a slight difference). But, at the moment, even in that PR, there are no tests covering the problem (i.e. if I revert that change in #221, its tests still pass). So, a test to show off the problem would be 💯

norkunas: @Lustmored PR covers more cases than mine, so probably It'd be better to leave to put that test in it's PR, so closing this one :)

PR #239, opened 2022-01-21
[LiveComponent][doc] Document how to deal with locales

### LINKED ISSUE #231: [LiveComponent] Component loses locale with every render
I just noticed that live components are not sending locale information in any way, thus enforcing backend to render with default locale every time (or maybe I'm missing something?). I can work around that in components themselves, but it'd be great to utilize all that Symfony offers for that.

I think best approach would be to have another live component route - one with locale (like `localized_live_component` with route `/_components/{_locale}/{component}/{action}` or something) so that Symfony would handle all the hard work.

I am unsure if this approach seems legit and how to wire it up with frontend in a reasonable way, thus just an issue for now and PR might follow if we agree upon actual solution :+1:

Lustmored:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #231
| License       | MIT

Locale was lost after any live component action. It turns out that simply by adding `{_locale}` to the route fixes the issue even when no locales are configured (router still seems to inject `en`).

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -99,6 +99,8 @@ Oh, and just one more step! Import a routing file from the bundle:
     # config/routes.yaml
     live_component:
         resource: '@LiveComponentBundle/Resources/config/routing/live_component.xml'
+        # uncomment to add localization to your components
+        #prefix: '/{_locale}'
 
 That's it! We're ready!
 


kbond:
Can this be achieved by adding the locale prefix when importing the route? https://symfony.com/doc/current/routing.html#route-groups-and-prefixes

I'd prefer suggesting this. Maybe a note in the docs/recipe?

Lustmored:
@kbond it looks like you are right and your approach seems much cleaner. I will follow up on that :+1:

Lustmored:
@kbond I don't think there is any recipe for live component (at least I couldn't find any), therefore I believe putting this in installation step instructions is enough.

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -99,6 +99,7 @@ Oh, and just one more step! Import a routing file from the bundle:
     # config/routes.yaml
     live_component:
         resource: '@LiveComponentBundle/Resources/config/routing/live_component.xml'
+        prefix: '/{_locale}' # comment out if your app is not localised
```
kbond:
```suggestion
        # uncomment to add localization to your components
        #prefix: '/{_locale}'
```

I'd prefer to comment by default.

kbond (changes requested):
> I don't think there is any recipe for live component

You're right, there isn't yet.

weaverryan (approved):

kbond (approved):

weaverryan:
Thank you Jakub!

PR #240, opened 2022-01-21
Upgrade php-cs-fixer used in gh action and run on php 8

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Currently, php-cs-fixer doesn't like the php 8 syntax (ie unions/promoted properties) we'd like to use in twig/live components.


weaverryan:
Nice idea - thanks Kevin!

PR #241, opened 2022-01-21
[LiveComponent] add ability to have array props

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Discovered a bug when working on #220. You currently can't have `LiveProp`'s that are "arrays". This is because the [_expose_ system](https://symfony.com/bundles/ux-live-component/current/index.html#modifying-embedded-properties-with-the-exposed-option) uses arrays and we weren't strict enough in determining when this system was being used.

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -28,7 +28,7 @@ use Symfony\UX\LiveComponent\Exception\UnsupportedHydrationException;
 final class LiveComponentHydrator
 {
     private const CHECKSUM_KEY = '_checksum';
-    private const EXPOSED_PROP_KEY = 'id';
+    private const EXPOSED_PROP_KEY = '_id';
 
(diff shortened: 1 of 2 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -28,7 +28,7 @@
 final class LiveComponentHydrator
 {
     private const CHECKSUM_KEY = '_checksum';
-    private const EXPOSED_PROP_KEY = 'id';
+    private const EXPOSED_PROP_KEY = '_id';
```
kbond, line 31: This doesn't mess up the frontend, does it?

weaverryan: I don't believe so

weaverryan (approved):

weaverryan:
Thanks Kevin!

PR #242, opened 2022-01-24
Minor clarifications to TwigComponent doc examples

allejo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | n/a 
| License       | MIT

The examples I've touched sounded a bit misleading to me at first comparing to the Live Component package. Having something available as `{{ message }}` isn't possible since an error is thrown since it's not a property nor does it exist in the `mount()` method. I feel like examples should just make the `this` variable very clear.

### DIFF
--- a/src/TwigComponent/src/Resources/doc/index.rst
+++ b/src/TwigComponent/src/Resources/doc/index.rst
@@ -27,8 +27,8 @@ And (2) a corresponding template:
 .. code-block:: twig
 
     {# templates/components/alert.html.twig #}
-    <div class="alert alert-{{ type }}">
-        {{ message }}
+    <div class="alert alert-{{ this.type }}">
+        {{ this.message }}
     </div>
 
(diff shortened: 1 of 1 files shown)


weaverryan:
Hi there!

Ah, this is tricky. The reason the documentation doesn't work is that you CAN now access local variables, but that feature has not been released yet :/. I think what we SHOULD do is clarify the version this stuff is available in. So what we're missing is:

```rst
.. versionadded:: 2.1
    The ability to reference local variables in the template (e.g. ``message``) was added in TwigComponents 2.1.
    Previously, all data needed to be referenced through ``this`` (e.g. ``this.message``).
```

That would hopefully give a hint of a feature doesn't work.

allejo:
Ah that would explain things! Thanks for the clarification! I'll close this PR in favor of #244 then :+1:

PR #245, opened 2022-01-24
[Live] Always send HTML over the wire instead of JSON

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (one part of it is)
| New feature?  | yes
| Tickets       | None
| License       | MIT

Hi!

Currently, the LiveComponent controller expects to get JSON back from the server, containing `{ data: ..., html: '...' }`. But, Stimulus is so cool, that this isn't necessary! The LiveComponent Ajax requests can simply return HTML in all cases. The rendered element will have an updated `data-live-data-value` attribute with the latest data. And so, the Stimulus component instantly reads that updated data and incorporates it: there is no need to send back the "component data" on a separate key in the JSON.

This makes the live component system more boring in a good way: it's just a way to return HTML (and then we have fancy JavaScript that adds all the interactivity). It's also a bit faster now, as previously we were dehydrating the component data twice on re-render.

I can't think of any downsides to this approach - everything seems to work just fine.

Cheers!

### DIFF
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -6,12 +6,6 @@ import { buildFormData, buildSearchParams } from './http_data_helper';
 import { setDeepData, doesDeepPropertyExist, normalizeModelName } from './set_deep_data';
 import { haveRenderedValuesChanged } from './have_rendered_values_changed';
 
-interface LiveResponseData {
-    redirect_url?: string,
-    html?: string,
-    data?: any,
-}
-
 interface ElementLoadingDirectives {
     element: HTMLElement,
     directives: Directive[]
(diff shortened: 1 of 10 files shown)


src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -195,20 +189,8 @@ export default class extends Controller {
...
-
-        if (!value) {
-            const clonedElement = (element.cloneNode());
-            // helps typescript know this is an HTMLElement
-            if (!(clonedElement instanceof HTMLElement)) {
-                throw new Error('cloneNode() produced incorrect type');
-            }
-
-            throw new Error(`The update() method could not be called for "${clonedElement.outerHTML}": the element must either have a "data-value" or "value" attribute set.`);
-        }
```
weaverryan, line 209: Unrelated bug fix from a feature I merged recently #203. That PR added this check, but it IS ok to have an empty value - it just means, for example, that a textarea is empty! For some reason, the tests didn't catch this before, but after this change, one test DID start failing.

kbond (approved):
Wow, this simplifies things a lot!

PR #249, opened 2022-01-27
[LiveComponent] fix `component_url` twig function

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes (kind of)
| Tickets       | n/a
| License       | MIT

Previously, it didn't work as intended. It required a component instance. It was supposed to work like `component()` (from twig-components), except generate a url to the live component (instead of rendering).


weaverryan (approved):

weaverryan:
Thank you Kevin!

PR #250, opened 2022-01-28
[Live] Fixing bug where inputs would not re-render if the value changed

### LINKED ISSUE #144: [LiveComponent] How to clear input form fields?
Hi,

I am working on a demo todo app using LiveComponent.
The project is here: https://github.com/sandoche/Symfony-ux-live-component-todo-list/

I manage to submit the form to create the tasks but I don't manage to clear the form input field.
Is there a way to do it directly from the LiveComponent or do I need to write javascript?

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #144 and Bug E on #102.
| License       | MIT

~~NOTE to self: I'm considering reversing this PR as it has some side effects where we lose input values on re-render. That can be solved with `data-live-ignore`, but it may be better to reverse this, and "fix" the original problem by telling the user to add a `data-live-update` attribute (or something like that) where we opt INTO re-rendering. There is basically a situation where we don't know if the user will want a form element to re-render (and update the value) or not. And so, the user needs to choose. The question is, which way should the "default" be.~~

For example, if an input rendered initially empty, then you typed
into it, and, on re-render, the value was set BACK to the original,
the input would *not* update on re-render, and it would be stuck
with the old value.

This was because the old input node and new input node were seen
as identical... and so morphdom didn't bother to update that node.
However, in reality, the value property of the "old" input had
been changed and would no longer match the value "attribute" of the
incoming input. Effectively, the 2 inputs had different values, but
this difference was not caught by the system.

### DIFF
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -5,6 +5,8 @@ import { combineSpacedArray } from './string_utils';
 import { buildFormData, buildSearchParams } from './http_data_helper';
 import { setDeepData, doesDeepPropertyExist, normalizeModelName } from './set_deep_data';
 import { haveRenderedValuesChanged } from './have_rendered_values_changed';
+import { normalizeAttributesForComparison } from './normalize_attributes_for_comparison';
+import { cloneHTMLElement } from './clone_html_element';
 
 interface ElementLoadingDirectives {
     element: HTMLElement,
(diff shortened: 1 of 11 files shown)


src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -5,6 +5,8 @@ import { combineSpacedArray } from './string_utils';
 import { buildFormData, buildSearchParams } from './http_data_helper';
 import { setDeepData, doesDeepPropertyExist, normalizeModelName } from './set_deep_data';
 import { haveRenderedValuesChanged } from './have_rendered_values_changed';
+import { normalizeAttributesForComparison } from './normalize_attributes_for_comparison';
+import { cloneHTMLElement } from "./clone_html_element";
```
norkunas:
```suggestion
import { cloneHTMLElement } from './clone_html_element';
```
for consistency? :)

norkunas:
Ran into a case today where I've defined a modal (with managed form) in a LiveComponent and after changing value in form everything was re-rendered and modal was closed, so `data-live-ignore` should solve this for me? :)

weaverryan:
> Ran into a case today where I've defined a modal (with managed form) in a LiveComponent and after changing value in form everything was re-rendered and modal was closed, so data-live-ignore should solve this for me? :)

Ideally, yes! That's exactly the case that we should test to make sure works :)

norkunas:
And I think this also fixed the #102 Bug H?

PR #252, opened 2022-01-28
[Twig][Live] Require Symfony 5.4+

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Both @weaverryan and I agree that bumping the min requirement of these library's would be beneficial. The primary reason being, the usability pre-5.3 is sub-par. Component's cannot be autoconfigured with the `AsTwigComponent`/`AsLiveComponent` attributes - they must be tagged manually. Requiring 5.4 makes documentation simpler, cleans up the test suite, and fixes some other compatibility issues. We figured, as we require PHP 8.0 already, requiring Symfony 5.4+ would be palatable.


kbond:
Pinging some community members who are clearly using these library's: @Lustmored, @seb-jean, @norkunas, @allejo, @mdoutreluingne, @sabat24, @BeyerJC, @benr77. Any thoughts/objections?

norkunas:
None from me, using 6.0+PHP8.1 for the project

BeyerJC:
I think its a good idea. My projects are all >=5.4 , so lets go :-)

Lustmored:
Same here - 6.0 and PHP8.1 only. In fact I was just thinking about that when making changes in test Kernel :smiley:

sabat24:
Keeping it up to date with last active Symphony's LTS version is perfectly fine for me.

benr77:
Agreed - 5.4 as the minimum is fine by me and I also think it's sensible to make this change sooner rather than later.

mdoutreluingne:
It's totally understandable and acceptable to demand Symfony 5.4+, as it's an LTS release. Anyway, active projects have every interest in moving to Symfony 5.4 or more for security reasons and more... So let's go 👍

kbond:
@symfony/mergers any objections?

nicolas-grekas (approved):
Works for me!

Tobion (approved):

Nyholm (approved):
Thank you

allejo:
No concerns here! Using PHP8.1 & Symfony 5.4 here 🥳

PR #254, opened 2022-01-28
[Live] Re-extracting form data after submit

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Tickets       | Addresses part of #221
| License       | MIT

This fixes 3 bugs at once:

1) This recalculates `formValues` after submit, in case the submitted data *changes* the form's data or underlying structure (e.g. `CollectionType`)

2) If a field is suddenly missing entirely from a form (hint: think removing an embedded CollectionType form), then previously `validatedFields` did not "forget" that field. This caused edge-case bug when removing an embedded form that was validated, then adding a new form a moment later (which would be empty, but now validated).

3) When passing a `FormView` into your component that was already validated, on the first re-render, the validation would be lost and only modified fields would be validated.

Ping @Lustmored and @norkunas. This mostly addresses @norkunas's problems with how `getFormValues()` works, but this was also causing trouble for @Lustmored, potentially for different reasons. So, this may not solve the problems addressed in #221, but it's certainly related.

Cheers!

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -17,6 +17,8 @@ use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
 use Symfony\UX\LiveComponent\Attribute\BeforeReRender;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\Util\LiveFormUtility;
+use Symfony\UX\TwigComponent\Attribute\PostMount;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
@@ -63,17 +65,29 @@ trait ComponentWithFormTrait
      */
     abstract protected function instantiateForm(): FormInterface;
 
-    /**
-     * Override in your class if you need extra mounted values.
-     *
-     * Call $this->setForm($form) manually in that situation
-     * if you're passing in an initial form.
-     */
-    public function mount(?FormView $form = null)
+    #[PostMount]
+    public function postMount(array $data): array
     {
-        if ($form) {
-            $this->setForm($form);
+        // allow the FormView object to be passed into the component() as "form"
+        if (\array_key_exists('form', $data)) {
+            $this->formView = $data['form'];
+            unset($data['form']);
+
+            if ($this->formView) {
+                // if a FormView is passed in and it contains any errors, then
+                // we mark that this entire component has been validated so that
+                // all validation errors continue showing on re-render
+                if (LiveFormUtility::doesFormContainAnyErrors($this->formView)) {
+                    $this->isValidated = true;
+                    $this->validatedFields = [];
+                }
+            }
         }
+
+        // set the formValues from the initial form view's data
+        $this->initializeFormValues();
+
+        return $data;
     }
 
     /**
@@ -87,7 +101,7 @@ trait ComponentWithFormTrait
     public function submitFormOnRender(): void
     {
         if (!$this->getFormInstance()->isSubmitted()) {
-            $this->submitForm(false);
+            $this->submitForm($this->isValidated);
         }
     }
 
@@ -103,18 +117,6 @@ trait ComponentWithFormTrait
         return $this->formView;
     }
 
-    /**
-     * Call this from mount() if your component receives a FormView.
-     *
-     * If your are not passing a FormView into your component, you
-     * don't need to call this directly: the form will be set for
-     * you from your instantiateForm() method.
-     */
-    public function setForm(FormView $form): void
-    {
-        $this->formView = $form;
-    }
-
     public function getFormName(): string
     {
         if (!$this->formName) {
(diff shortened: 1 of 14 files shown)


norkunas:
So if I'll have hundred live actions I'll need to write same number submit/reset statements? :)

weaverryan:
> So if I'll have hundred live actions I'll need to write same number submit/reset statements? :)

Yup! The same as if you has a hundred controller actions 😉 

BUT, I'm re-thinking this approach - I think it's flawed, as trying to "submit the form" then "change the data the form was created from" then... re-recreate the form with the new mix of data... just doesn't work, as the final "new" form wouldn't be "submitted", so it wouldn't have correct validation errors. If we re-submit the 2nd form, then we overwrite the new values we just changed in our underlying object. I think it's a dead approach.

I think the easier flow is to, in your actions, manipulate `$formValues` directly. This, I believe, is actually what you were doing all along @norkunas  - I was trying to do something fancier. It doesn't look as neat as manipulating the underlying form data DTO/entity, but it's much simpler. The flow in an action would be what you have currently @norkunas (I'm just arriving to the solution later). I still need to try it, and see if my fix for `getFormValues()` still works with your approach. Will check into it next week.

norkunas:
>  just doesn't work, as the final "new" form wouldn't be "submitted", so it wouldn't have correct validation errors.

Yeah, at least I currently get them 😉

Lustmored:
From a quick glimpse I can tell that this does not collide nor solve the problem addressed in my PR, which focuses on handling different kind of checkbox field types within forms :+1:

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -124,18 +118,18 @@ public function getFormName(): string
...
     }
 
     private function submitForm(bool $validateAll = true): void
     {
-        $this->getFormInstance()->submit($this->formValues);
+        $form = $this->getFormInstance();
+        $form->submit($this->formValues);
+        // re-extract the "view" values in case the submitted data
+        // changed the underlying data or structure of the form
+        $this->formValues = $this->extractFormValues($this->getForm());
```
weaverryan: This is the key. It's super simple, it just took me a long time to understand what's going on. The flow is this:

A) A custom action is called
B) In the custom action, the user changes `$this->formValues`. For example, to add a new array to initialize a new collection type.
C) We re-extract the `formValues` from the form.

Normally, if the user sends `formValues` and we then use those to submit the form, the `formValues` that should be sent back to the frontend would be the same: that is STILL the "formValues" data. However, if your form *changes* as a result of submitted data (e.g. `CollectionType` which adds more fields if you add a new item to a collection type array), then we need to re-extract the `formValues` so that they match the new reality.

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -124,18 +118,18 @@ public function getFormName(): string
...
-    public function getFormValues(): array
+    private function initializeFormValues(): void
     {
-        if (null === $this->formValues) {
-            $this->formValues = $this->extractFormValues($this->getForm());
-        }
-
-        return $this->formValues;
+        $this->formValues = $this->extractFormValues($this->getForm());
     }
```
weaverryan, line 132: Everything above this was just cleaning up the flow, but not significant

weaverryan:
Other than a small TODO in the documentation, this is ready to go. This is much simpler now & mirrors what @norkunas was doing in his project, except now with the fix to `formValues`. It mentally took me awhile to get the clarity that this small solution was needed (I previously had a much bigger solution in this PR).

> So if I'll have hundred live actions I'll need to write same number submit/reset statements? :)

So now the answer to that is "no": your actions modify `formValues` and nothing else.

src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
```diff
@@ -0,0 +1,82 @@
...
+                $dehydrated['validatedFields'] = ['blog_post_form.content'];
+                $token = $response->crawler()->filter('div')->first()->attr('data-live-csrf-value');
+            })
+            // post to action, which will change the "name" field in the form
+            ->post('/_components/form_with_collection_type/addComment?'.http_build_query($dehydrated), [
+                'headers' => ['X-CSRF-TOKEN' => $token],
+            ])
+            ->assertStatus(422)
+            ->dump()
+            // look for original embedded form
```
kbond:
```suggestion
            ->assertStatus(422)
            // look for original embedded form
```

weaverryan: Oh, but the dump is such a cool feature! ;)

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -66,14 +66,20 @@ abstract protected function instantiateForm(): FormInterface;
     /**
      * Override in your class if you need extra mounted values.
      *
-     * Call $this->setForm($form) manually in that situation
-     * if you're passing in an initial form.
+     * If you override, call $this->initializeFormValues() manually
+     * and also $this->formView = $form if you are passing in a Form.
+     *
+     * TODO: Refactor to PostMount when available.
```
kbond: `PostMount` is now available.

weaverryan: yeeeea it is 😎

weaverryan:
This is now ready. The form process is now much smoother, working with `CollectionType` is a wonderful experience.

norkunas (approved):

kbond (approved):

src/LiveComponent/src/Util/LiveFormUtility.php
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            // key was found at all levels
+            return true;
+        }));
+    }
+
+    public static function doesFormContainAnyErrors(FormView $formView): bool
+    {
+        if ($formView->vars['errors'] ?? null && \count($formView->vars['errors']) > 0) {
```
norkunas: Maybe?:
```suggestion
        if (($formView->vars['errors'] ?? null) && \count($formView->vars['errors']) > 0) {
```
As EA hints about:
image

weaverryan: Not sure - but your suggestion is safer 👍

PR #255, opened 2022-01-31
[Twig][Live] Native attributes

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

This removes the need for the `HasAttributesTrait` that was introduced in #220. Attributes are now "a thing all component have".

`ComponentFactory::create()` returns a new `MountedComponent` object that holds the "mounted" component, attributes and component metadata. `LiveComponentHydrator::dehydrate()` now accepts an instance of `MountedComponent` and `LiveComponentHydrator::hydrate()` now returns an instance of `MountedComponent`.

If the component is live, any attributes are dehydrated/hydrated as "readonly" data (they can't be modified by the frontend).

**TODO**
- [x] Update docs referring to `HasAttributesTrait`


src/TwigComponent/src/EventListener/PreRenderEvent.php
```diff
@@ -77,6 +77,11 @@ public function setVariables(array $variables): self
 
     public function getMetadata(): ComponentMetadata
     {
-        return $this->metadata;
+        return $this->mounted->getMetadata();
+    }
+
+    public function getMountedComponent(): MountedComponent
```
kbond, line 86: As discussed with @weaverryan on slack, we'd like to keep the `MountedComponent` object "internal". This getter is required for `AddLiveAttributesSubscriber` but maybe we should mark it as `@internal`?

kbond:
@seb-jean, you've been experimenting with attribute system added in #220, what do you think of this PR?

seb-jean:
I think this PR is a good idea.

src/TwigComponent/src/MountedComponent.php
```diff
@@ -0,0 +1,52 @@
...
+ * @experimental
+ *
+ * @internal
+ */
+final class MountedComponent
+{
+    public function __construct(
+        private object $component,
+        private ComponentAttributes $attributes,
+        private ComponentMetadata $metadata
```
weaverryan: It's minor, but for some reason, it's bothering me that `ComponentMetadata` is inside this object. I think it's because the `$component` and `$attributes` are both "run-time" values (i.e. the user can pass in different values) but `ComponentMetadata` is static config.

And, looking at where we use it, I think it may not be needed. If we:

* Add `string $name` to `MountedComponent`
* In `ComponentRenderer`, inject `ComponentFactory` and use it to get the metadata - e.g. `$this->factory->metadataFor($mountedComponent->getName())`. THEN, ONLY pass the `->getTemplate()` into `PreRenderEvent`... because that's all we need! We currently use `->getName()` via `AddLiveAttributesSubscriber` calling `LiveComponentRuntime::getLiveAttributes`, but that can be replaced with `$mountedComponent->getName()`/

kbond: Sounds good, I'll play with this idea.

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -65,11 +65,18 @@ public function create(string $name, array $data = []): object
...
-        foreach ($data as $property => $value) {
-            throw new \LogicException(sprintf('Unable to write "%s" to component "%s". Make sure this is a writable property or create a mount() with a $%s argument.', $property, \get_class($component), $property));
+        // create attributes from "attributes" key if exists
+        $attributes = $data['attributes'] ?? [];
+        unset($data['attributes']);
+
+        // ensure remaining data is scalar
+        foreach ($data as $key => $value) {
+            if (!is_scalar($value)) {
+                throw new \LogicException(sprintf('Unable to use "%s" (%s) as an attribute. Attributes must be scalar. If you meant to mount this value on "%s", make sure $%s is a writable property.', $key, get_debug_type($value), $component::class, $key));
```
weaverryan: Double-check this, but:

```suggestion
                throw new \LogicException(sprintf('Unable to use "%s" (%s) as an attribute. Attributes must be scalar. If you meant to mount this value on "%s", make sure $%s is a writable property or create a mount() method with a "$%s" argument.', $key, get_debug_type($value), $component::class, $key, $key));
```

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -47,7 +47,7 @@ public function metadataFor(string $name): ComponentMetadata
     /**
      * Creates the component and "mounts" it with the passed data.
      */
-    public function create(string $name, array $data = []): object
+    public function create(string $name, array $data = []): MountedComponent
```
weaverryan, line 50: This is one of the spots where adding this feature makes things feel bigger and less "pure". However, this has grown on me. I like that we have a method where you pass in the string component name + array of data, and you get back the (A) the component object and (B) the extra stuff as attributes. I think that feels good.

src/TwigComponent/src/Twig/ComponentRuntime.php
```diff
@@ -32,9 +32,6 @@ public function __construct(ComponentFactory $componentFactory, ComponentRendere
 
     public function render(string $name, array $props = []): string
     {
-        return $this->componentRenderer->render(
-            $this->componentFactory->create($name, $props),
-            $this->componentFactory->metadataFor($name)
-        );
+        return $this->componentRenderer->render($this->componentFactory->create($name, $props));
```
weaverryan, line 35: Hey! That's really pretty!

weaverryan (commented):
I think this is good. It introduces `MountedComponent`, but that feels ok to me. 👍 for mer personally

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -25,23 +25,22 @@ final class ComponentRenderer
...
 
             $this->safeClassesRegistered = true;
         }
 
-        $event = new PreRenderEvent(
-            $component,
-            $metadata,
-            array_merge(['this' => $component], get_object_vars($component))
-        );
+        $event = new PreRenderEvent($mounted, $this->factory->metadataFor($mounted->getName()));
```
kbond, line 43: Requiring the service here is a bit unfortunate.

weaverryan:
Thanks Kevin!

PR #247, opened 2022-01-25
Fix Typescript alerts

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | -
| Tickets       | -
| License       | MIT

Fix Typescript alerts in controllers.

PR #226, opened 2022-01-17
Fix error message

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | N/A
| License       | MIT

PR #256, opened 2022-02-01
Rebuild js dist files

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

Tried to update my project deps to use 2-x.dev of LiveComponent, but now that renderer returns HTML instead of JSON, it fails, because the LiveController still uses JSON.

This rebuilds dist files from #226, #245, #247 changes.


kbond:
Thank you @norkunas.

PR #259, opened 2022-02-01
[Twig] add `ExposeInTemplate` attribute

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

#232 made all of a component's public properties available directly (without the `this.`) in it's template. This adds an `ExposeInTemplate` attribute to make non-public properties available in this way.

Example:

```php
use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;

#[AsTwigComponent('with_exposed_variables')]
final class WithExposedVariables
{
    #[ExposeInTemplate]
    private string $prop1; // available as `{{ prop1 }}`

    #[ExposeInTemplate('customProp')]
    private string $prop2; // available as `{{ customProp }}`

    #[ExposeInTemplate(name: 'ico', getter: 'fetchIcon')]
    private string $icon = 'ico-warning'; // available as `{{ ico }}` in the template using `fetchIcon()` as the getter

    /**
     * Required to access $prop1
     */
    public function getProp1(): string
    {
        return $this->prop1;
    }

    /**
     * Required to access $prop1
     */
    public function getProp2(): string
    {
        return $this->prop2;
    }

    /**
     * Required to access $this->icon
     */
    public function fetchIcon(): string
    {
        return $this->icon;
    }
}
```

**TODO:**
- [x] use to expose `form` in `ComponentWithFormTrait` (requires https://github.com/symfony/ux/pull/254)

### DIFF
--- a/src/TwigComponent/src/Resources/doc/index.rst
+++ b/src/TwigComponent/src/Resources/doc/index.rst
@@ -219,6 +219,60 @@ If an option name matches an argument name in ``mount()``, the option is
 passed as that argument and the component system will *not* try to set
 it directly on a property.
 
+ExposeInTemplate Attribute
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.1
+
+    The ``ExposeInTemplate`` attribute was added in TwigComponents 2.1.
+
+All public component properties are available directly in your component
+template. You can use the ``ExposeInTemplate`` attribute to expose
+private/protected properties directly in a component template (``someProp``
+vs ``this.someProp``). These properties must be *accessible* (have a getter).
+
+    // ...
+    use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+
+    #[AsTwigComponent('alert')]
+    class AlertComponent
+    {
+        #[ExposeInTemplate]
+        private string $message; // available as `{{ message }}` in the template
+
+        #[ExposeInTemplate('alert_type')]
+        private string $type = 'success'; // available as `{{ alert_type }}` in the template
+
+        #[ExposeInTemplate(name: 'ico', getter: 'fetchIcon')]
+        private string $icon = 'ico-warning'; // available as `{{ ico }}` in the template using `fetchIcon()` as the getter
+
+        /**
+         * Required to access $this->message
+         */
+        public function getMessage(): string
+        {
+            return $this->message;
+        }
+
+        /**
+         * Required to access $this->type
+         */
+        public function getType(): string
+        {
+            return $this->type;
+        }
+
+        /**
+         * Required to access $this->icon
+         */
+        public function fetchIcon(): string
+        {
+            return $this->icon;
+        }
+
+        // ...
+    }
+
 PreMount Hook
 ~~~~~~~~~~~~~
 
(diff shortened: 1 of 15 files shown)


src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -219,6 +219,50 @@ If an option name matches an argument name in ``mount()``, the option is
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.1
+
+    The ``ExposeInTemplate`` attribute was added in TwigComponents 2.1.
+
+All public component properties are available directly in your component
+template. You can use the ``ExposeInTemplate`` attribute to expose
+private/protected properties and public methods as variables directly in a
+component template (`someProp` vs `this.someProp`). For properties, they
```
```suggestion
component template (``someProp`` vs ``this.someProp``). For properties, they
```

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -219,6 +219,50 @@ If an option name matches an argument name in ``mount()``, the option is
...
+        #[ExposeInTemplate('alert_type')]
+        private string $type = 'success'; // available as `{{ alert_type }}` in the template
+
+        #[ExposeInTemplate]
+        public function getIcon(): string
+        {
+            // return value will be available as `{{ icon }}` in the template (note the get prefix is trimmed)
+        }
+
+        public function getMessage(): string
```
weaverryan, line 252: I liked how you had a comment above this in the PR readme, and I think it should be here too:

```
// Required to access $this->message

File: src/TwigComponent/src/Resources/doc/index.rst [hunk 3]
```diff
@@ -219,6 +219,50 @@ If an option name matches an argument name in ``mount()``, the option is
...
+            return $this->message;
+        }
+
+        public function getType(): string
+        {
+            return $this->type;
+        }
+
+        // ...
+    }
```
weaverryan, line 274: Maybe a note about the methods:

```rst
.. note::

    When ``#[ExposeInTemplate]`` is used above a method, that method will be called one
    time before the template is rendered.
```

Also... this is kind of ~calculated~ computed properties, isn't it? They're not "lazy" (they WILL be called), but I think it's as close as we're ever going to get (assuming some new feature isn't added to Twig)?

weaverryan (commented):
❤️

src/TwigComponent/src/Resources/doc/index.rst
kbond: Yeah, I don't know that I'd recommend using this for that (but this is probably because I really don't mind the `this.` syntax).

weaverryan: The problem is that there IS no other computed prop syntax. If you call a method, it's called every time you use it. With this, the method IS called automatically (even if you never use it), but it's only called once, hence it is "computed".

kbond: Yep, fair enough. I'm thinking of the scenario where you have a method that's a bit "heavy" and is only called under certain circumstances. Wouldn't want to expose this. In this case, we'd want to recommend [this](https://symfony.com/bundles/ux-twig-component/current/index.html#computed-properties).

kbond:
What do you think about making this a property-only attribute? Then for methods, we'll recommend either using `this.` or `computed.` (from #266)? I think this could help with documentation.

For making `form` available on `ComponentWithFormTrait`, I'm thinking we could use this syntax:

```
trait ComponentWithFormTrait
{
    #[ExposeInTemplate(name: 'form', getter: 'getForm')]
    private ?FormView $formView = null;

    // ...
}
```

We wouldn't have to rename `$formView`. The `getter` attribute argument wouldn't make sense for methods.

weaverryan:
> What do you think about making this a property-only attribute? Then for methods, we'll recommend either using this. or computed. (from https://github.com/symfony/ux/pull/266)? I think this could help with documentation

Hmm, ok. Let's do this. It doesn't prevent us from adding this later by popular demand, so I don't see any risk. 👍

kbond:
I believe this to be ready.

weaverryan:
Thank you Kevin - looks awesome!

PR #263, opened 2022-02-02
[Live] migrate Hydration system to standard Symfony normalizers

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

This removes our custom hydration system and uses `symfony/serializer` directly. Hydrators are now just (de)normalizer's registered with Symfony as normal.

The serializer context (`['live-component' => true]`) can be used to restrict (de)normalizers to just live component properties.


src/LiveComponent/src/Normalizer/LiveComponentNormalizer.php
```diff
@@ -0,0 +1,257 @@
...
+                }
+            } else {
+                $data[$frontendName] = $dehydratedValue;
+            }
+        }
+
+        $data[self::CHECKSUM_KEY] = $this->computeChecksum($data, $readonlyProperties);
+
+        return $data;
+    }
```
weaverryan: What's the "profit" of moving this logic here vs keeping it inside of `LiveComponentHydrator`. Right now, it looks like we are simply setting a special context flag... then calling the normalizer so that the normalizer can call this class exactly. I like that the individual properties of the component go through the (de)normalizer. But I'm not sure about moving this main logic here. WDYT?

kbond: My thinking was to eventually remove `LiveComponentHydrator`. Keeping it for now helped with the refactor. When I went to remove it, I saw how ugly it was to use the symfony serializer directly (and we need to do this in multiple places) so kept it for now.

I want to see how #264 pans out before I continue this. I'll be able to unserialize the json directly.

weaverryan: Ah, you're thinking to call the serializer directly form `LiveComponentSubscriber`? Grab the raw JSON string from there and pass it straight to the serializer system? That is, indeed, interesting

kbond: Yes, same with `LiveComponentRuntime`. I want to see how that looks.

kbond: With attributes/mounted component I discovered I couldn't easily have the `LiveComponentNormalizer`, I've kept the logic in `LiveComponentHydrator`.

weaverryan:
Seems very resonable and makes custom code unnecessary. Let's give it a try - thanks Kevin!

PR #264, opened 2022-02-02
Send data back and forth as JSON

### LINKED ISSUE #251: [LiveComponent] `_willDataFitInUrl` does not calculate correctly
Based on my findings:
 - It doesn't properly serialize to string value when `this.dataValue` contains nested objects, because in my case it `Object.values(this.dataValue).join(',')` returns: `[object Object],form_builder,[object Object],false,form_builder.slug,form_builder.title,NJQRgieAbBmGi0vFxpGlGJVYTrJpA0se7rz4rDTl4N8=`
 - It doesn't take into account `params` and only uses `this.dataValue` to determine it

So in my case when editing a big form with a LiveComponent it always fails to render because it does GET requests with too long query string

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       |  Fixes #251 (a bit unrelated, but it fixes it)
| License       | MIT

tl;dr the "data" of your component is now sent via JSON for the Ajax calls. Ajax calls are still made via `GET` (unless you're triggering an action or your data is too long) with a new `?data={... json}`.

**Longer Explanation**

Until now, I've been trying to send data back to the server as `?query` params or as POST data (which is just query params stored in the body of the request).

However, using query parameters simply isn't robust enough. Most notably, it makes handling `null` values impossible. If you have a LiveProp - e.g. `firstName` - on the server set to `null`, how should that be sent back to the server? As `?firstName=`? In that case, how do you distinguish from an empty string? One option is to NOT send `firstName` at all on the Ajax request. But then, what if you have a `LiveProp` that is an array with `['firstName' => null]`? If we don't send `firstName`, the array will dehydrate to an empty array.

In the end, the goal is to effectively "freeze" your component's data, send it to the frontend, and have the frontend send it back. Using query parameters simply isn't robust enough to do this without, but JSON is.

There is no real downside to this: it just felt more pure to send data back and forth as query parameters and return HTML. We still return HTML, but sending JSON will make the behavior solid.


.php-cs-fixer.dist.php
```diff
@@ -24,6 +24,7 @@
             ->append([__FILE__])
             ->notPath('#/Fixtures/#')
             ->notPath('#/app/var/#')
+            ->notPath('#/var/cache/#')
```
kbond, line 27: THANK YOU!

kbond (approved):

norkunas:
Also just want to point out that in this way it will be impossible to support #102 Bug I :) Unless with some workarounds which would allow to convert files to base64 so symfony serializer would transform it

weaverryan:
> https://github.com/symfony/ux/issues/102 Bug I :)
> Bug I) Check file uploads with actions

You're right that uploads will definitely NOT work after this. But I think it's still fixable - data uri is one approach, but we can also look into submitting a "multipart" request body (like a normal file upload) where one part is the file upload and another part is the JSON data.

PR #266, opened 2022-02-03
[Twig] add computed properties system

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

This adds a `computed` component template variable to make "computed methods" easier. In your component's template, to call a method on your component and cache the return (to use again later), instead of calling `this.methodName`, call `computed.methodName`:

```twig
{# templates/components/featured_products.html.twig #}

<div>
    <h3>Featured Products</h3>

    {% for product in computed.products %}
        ...
    {% endfor %}

    ...
    {% for product in computed.products %} {# use cache, does not result in a second query #}
        ...
    {% endfor %}
</div>
```

### DIFF
--- a/src/TwigComponent/src/ComputedPropertiesProxy.php
+++ b/src/TwigComponent/src/ComputedPropertiesProxy.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @experimental
+ */
+final class ComputedPropertiesProxy
+{
+    private array $cache = [];
+
+    public function __construct(private object $component)
+    {
+    }
+
+    public function __call(string $name, array $arguments): mixed
+    {
+        if ($arguments) {
+            throw new \InvalidArgumentException('Passing arguments to computed methods is not supported.');
+        }
+
+        if (isset($this->component->$name)) {
+            // try property
+            return $this->component->$name;
+        }
+
+        if ($this->component instanceof \ArrayAccess && isset($this->component[$name])) {
+            return $this->component[$name];
+        }
+
+        $method = $this->normalizeMethod($name);
+
+        if (isset($this->cache[$method])) {
+            return $this->cache[$method];
+        }
+
+        if ((new \ReflectionMethod($this->component, $method))->getNumberOfRequiredParameters()) {
+            throw new \LogicException('Cannot use computed methods for methods with required parameters.');
+        }
+
+        return $this->cache[$method] = $this->component->$method();
+    }
+
+    private function normalizeMethod(string $name): string
+    {
+        if (method_exists($this->component, $name)) {
+            return $name;
+        }
+
+        foreach (['get', 'is', 'has'] as $prefix) {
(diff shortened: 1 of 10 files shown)


kbond:
Should we make the public properties available via computed (`computed.prop1`)? Right now only public methods are available.

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -360,8 +360,12 @@ need to populate, you can render it with:
     means that you can safely render the same component multiple times with
     different data because each component will be an independent instance.
 
-Computed Properties
-~~~~~~~~~~~~~~~~~~~
+Computed Methods
```
kbond: Not sure about this change. Should we keep as _Computed Properties_? Seems to be the more common term.

weaverryan:
> Should we make the public properties available via computed (computed.prop1)? Right now only public methods are available.

Ideally yes... just so that `this` and `component` work identically. But more so, could we re-use the Twig logic for calling the method / fetching the property? For example, we know that Twig checks several things when we say `this.foo` to get that property / call that method. Can we reuse that inside the proxy class?

src/TwigComponent/src/Resources/doc/index.rst
weaverryan: I had been wondering about this too. Yes, I think we keep it as `Computed Properties`. And you've already documented it in the way that I would have: you document using it *like* you're referencing a property: `this.products`. A method is called behind the scenes to calculate this, but it looks like a property.

kbond:
> Can we reuse that inside the proxy class?

I believe so, yes. I only check get/no-get method prefix now (so `computed.getMethod` and `computed.method` both work and refer to the same cache). I'll take a look at twig's internals and try to match.

kbond:
Ok, I think this is ready. I ensure `computed.x` works with public properties, components implementing `\ArrayAccess`, _getter's_, _isser's_ and _hasser's_.

src/TwigComponent/src/ComputedPropertiesProxy.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        if (method_exists($this->component, $name)) {
+            return $name;
+        }
+
+        foreach (['get', 'is', 'has'] as $prefix) {
+            if (method_exists($this->component, $method = sprintf('%s%s', $prefix, ucfirst($name)))) {
+                return $method;
+            }
+        }
```
weaverryan, line 65: So I'm guessing there wasn't a nice way to "steal" and re-use the Twig native functionality?

weaverryan (commented):
This looks great! i just wanted to triple-check about reusing the Twig logic

src/TwigComponent/src/ComputedPropertiesProxy.php
kbond: There was not...

kbond: I looked at using [`twig_get_attribute`](https://github.com/twigphp/Twig/blob/237f1789a7db3a0cbb8e28a75832a45b7f9fc304/src/Extension/CoreExtension.php#L1433) but it requires injecting the twig environment and [`Source`](https://github.com/twigphp/Twig/blob/3.x/src/Source.php).

kbond:
Rebased and conflicts resolved.

src/TwigComponent/src/ComputedPropertiesProxy.php
weaverryan: Wow, `Source`, that's a new class to me :)

weaverryan:
Thank you Kevin!

PR #269, opened 2022-02-04
Allow SVG elements in LiveComponent

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

When the component contains svg elements then it throws error on rerender:
```
Uncaught (in promise) Error: Could not clone element
    at cloneHTMLElement (live_controller.js:977:15)
    at onBeforeElUpdated (live_controller.js:1331:46)
    at morphEl (live_controller.js:492:21)
    at morphChildren (live_controller.js:605:33)
    at morphEl (live_controller.js:507:15)
    at morphChildren (live_controller.js:605:33)
    at morphEl (live_controller.js:507:15)
    at morphChildren (live_controller.js:605:33)
    at morphEl (live_controller.js:507:15)
    at morphdom (live_controller.js:722:13)
```

### DIFF
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -8,7 +8,7 @@ import { normalizeAttributesForComparison } from './normalize_attributes_for_com
 import { cloneHTMLElement } from './clone_html_element';
 
 interface ElementLoadingDirectives {
-    element: HTMLElement,
+    element: HTMLElement|SVGElement,
     directives: Directive[]
 }
 
@@ -190,11 +190,15 @@ export default class extends Controller {
         this._makeRequest(null);
     }
 
-    _getValueFromElement(element: HTMLElement) {
+    _getValueFromElement(element: HTMLElement|SVGElement) {
         return element.dataset.value || (element as any).value;
     }
 
-    _updateModelFromElement(element: HTMLElement, value: string, shouldRender: boolean) {
+    _updateModelFromElement(element: Element, value: string, shouldRender: boolean) {
+        if (!(element instanceof HTMLElement)) {
+            throw new Error('Could not update model for non HTMLElement');
+        }
+
         const model = element.dataset.model || element.getAttribute('name');
 
         if (!model) {
@@ -419,7 +423,7 @@ export default class extends Controller {
     /**
      * @private
      */
-    _handleLoadingDirective(element: HTMLElement, isLoading: boolean, directive: Directive) {
+    _handleLoadingDirective(element: HTMLElement|SVGElement, isLoading: boolean, directive: Directive) {
         const finalAction = parseLoadingAction(directive.action, isLoading);
 
         let loadingDirective: (() => void);
@@ -490,7 +494,7 @@ export default class extends Controller {
         const loadingDirectives: ElementLoadingDirectives[] = [];
 
         this.element.querySelectorAll('[data-loading]').forEach((element => {
-            if (!(element instanceof HTMLElement)) {
+            if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) {
                 throw new Error('Invalid Element Type');
             }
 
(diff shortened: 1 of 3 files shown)


src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -506,19 +506,19 @@ export default class extends Controller {
...
         element.style.display = 'none';
     }
 
-    _addClass(element: HTMLElement, classes: string[]) {
+    _addClass(element: HTMLElement|SVGElement, classes: string[]) {
         element.classList.add(...combineSpacedArray(classes));
     }
 
-    _removeClass(element: HTMLElement, classes: string[]) {
+    _removeClass(element: HTMLElement|SVGElement, classes: string[]) {
```
weaverryan, line 525: Below, inside `onBeforeElUpdated()`, morphom documents that both `fromEl` and `toEl` will be `HTMLElement`. Apparently that is a lie (I just tested, if you have an svg element, it is, of course, an `SVGElement`). 

So, right inside the `if (fromEl.isEqualNode(toEl)) {` check, on the first line (before we call `cloneHTMLElement`), we should add:

```js
if (!(fromEl instanceof HTMLElement) || !(toEl instanceof HTMLElement)) {
    return false;
} 
```

In other words, if we are not dealing with HTMLElements, then assume that the elements are unchanged and do not re-render them (at this point the `isEqualNode()` has already done most checks for equality).

src/LiveComponent/assets/src/clone_html_element.ts
```diff
@@ -1,6 +1,6 @@
-export function cloneHTMLElement(element: HTMLElement): HTMLElement {
+export function cloneHTMLElement(element: HTMLElement|SVGElement): HTMLElement|SVGElement {
     const newElement = element.cloneNode(true);
-    if (!(newElement instanceof HTMLElement)) {
+    if (!(newElement instanceof HTMLElement) && !(newElement instanceof SVGElement)) {
         throw new Error('Could not clone element');
     }
```
weaverryan: NOT cloning in the case of an SVG could cause unexpected problems (in some cases we clone because we then modify the cloned version, thinking that this is safe because it is not the REAL version on the page).

I think that we should use `element: HTMLElement` as the first argument and `HTMLElement` as the return type. Then, we'll rely on Typescript to tell us if we're being untruthful and are passing something potentially different into this method (though our Typescript still needs a lot of cleanup).

If we do this, there is only one place we will need to update to avoid Typescript being mad. It is in `live_controller` `_updateModelFromElement()`. This function should, I suppose, allow `element: Element` as the first arg. Then, on the first line, check if `!(element instanceof HTMLElement)`, then throw an exception.

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -419,7 +419,7 @@ export default class extends Controller {
     /**
      * @private
      */
-    _handleLoadingDirective(element: HTMLElement, isLoading: boolean, directive: Directive) {
+    _handleLoadingDirective(element: HTMLElement|SVGElement, isLoading: boolean, directive: Directive) {
```
weaverryan, line 426: Excellent!  I think `_getValueFromElement()` might also need this update? There are several other spots that use only `HTMLElement` that you correctly did NOT update because we know we are working with `HTMLElement` in those situations 👍

src/LiveComponent/assets/src/normalize_attributes_for_comparison.ts
```diff
@@ -5,14 +5,14 @@
...
-export function normalizeAttributesForComparison(element: HTMLElement): void {
+export function normalizeAttributesForComparison(element: HTMLElement|SVGElement): void {
     if (element.value) {
         element.setAttribute('value', element.value);
     } else if (element.hasAttribute('value')) {
         element.setAttribute('value', '');
     }
 
-    Array.from(element.children).forEach((child: HTMLElement) => {
+    Array.from(element.children).forEach((child: HTMLElement|SVGElement) => {
```
weaverryan: I think you can revert these after you follow my instructions about what to do in `onBeforeElUpdated()`.

weaverryan (commented):
Thank you for this! Kind of a pain that we need to deal with the union type. As best I can tell, DOM nodes will be either `HTMLElement` or `SVGElement`: I think there aren't more types we're missing.

norkunas:
Updated

weaverryan (approved):

weaverryan:
Thank you @norkunas!

PR #271, opened 2022-02-07
[LiveComponent] Skip clearing non clearable form errors

### LINKED ISSUE #265: [Live Component] Cannot handle forms containing SubmitType
When using a live component with a form, if I include a form field that uses `SubmitType` then I get a 500 error.

```
App\Component\InitiateTransferFormComponent::clearErrorsForNonValidatedFields(): Argument #1 ($form) 
must be of type Symfony\Component\Form\Form, Symfony\Component\Form\SubmitButton given, called in 
/var/www/app/vendor/symfony/ux-live-component/src/ComponentWithFormTrait.php on line 214
```

Is this because `SubmitType` does not extend in the same way as regular form fields? 

Perhaps `ComponentWithFormTrait` just needs to test for `SubmitType` and deal with accordingly?

For now I can fix the problem by just hardcoding the submit button HTML in the template instead of defining it in the form type class.

Thanks

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #265
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -11,7 +11,7 @@
 
 namespace Symfony\UX\LiveComponent;
 
-use Symfony\Component\Form\Form;
+use Symfony\Component\Form\ClearableErrorsInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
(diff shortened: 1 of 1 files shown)


weaverryan (approved):
That's very nice - thank you @norkunas!

PR #272, opened 2022-02-07
[LiveComponent] docs: `array_splice` instead of `unset`

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | N/A
| License       | MIT

When using unset, then the array keys are not re-indexed, which leaves gaps in the structure, while `array_splice` re-indexes

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -1137,7 +1137,7 @@ Now, create a Twig component to render the form::
         #[LiveAction]
         public function removeComment(#[LiveArg] int $index)
         {
-            unset($this->formValues['comments'][$index]);
+            array_splice($this->formValues['comments'], $index, 1);
         }
     }
 


src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1137,7 +1137,7 @@ Now, create a Twig component to render the form::
         #[LiveAction]
         public function removeComment(#[LiveArg] int $index)
         {
-            unset($this->formValues['comments'][$index]);
+            array_splice($this->formValues['comments'], $index, 1);
```
weaverryan, line 1140: Can you explain what practical problem the gaps leave? I wrote the docs from an app I built locally - I didn't notice any problems.

norkunas: For example when you add 3 entries, and then remove middle one, then the form `getData` returns:

```
data[0]
data[2]
```

And in backend I expect list, not an associative array :) I know that I could call `array_values`, but it's better to manage it properly, in same way how items are incrementally created, no? :)

weaverryan: That makes sense. But does this have any practical, real-world bad consequence? 

I'm only pushing back because `array_splice` is harder to read/understand than a simple `unset()`. If we were to fix this, I might actually prefer `array_values()`.

norkunas: But then it's a 2 function calls over a single :)

> But does this have any practical, real-world bad consequence

I don't think so. Ok then we can leave it as is :)

PR #276, opened 2022-02-15
[Twig][Live] add ability to configure exposing public props and `attributes` variable name

### LINKED ISSUE #275: Component attributes - way to pass html attributes to component more directly
You lately added a concept of [component attributes](https://symfony.com/bundles/ux-twig-component/current/index.html#component-attributes). The concept per se is a good direction to help managing attributes in TWIG files. However I've got a problem with the way how you pass that HTML attributes to TWIG component. As I understand the main idea behind this concept is to render everything passed to component as HTML attribute, unless it has own property declared in PHP class. Except some auto-magic involved I discovered some issues with such approach when I tried to implement this approach to my project:

1. ~~I use the `preMount` method to validate attributes passed from TWIG into PHP component (more information why I need such validation can be found in issue [#142](https://github.com/symfony/ux/issues/142)):~~
```php
#[PreMount]
    public function configureProperties(array $data): array
    {
        $resolver->setRequired(["text", "url", "action", "icon", "icon_2", "class", "text_transform", "color"]);
        $resolver->setAllowedValues('icon', ["ico_add_card_white", "button_search_loupe", "ico_plus", "arrow_grey_right", "arrow_mob_right_orange", "arrow_mob_left_orange", "arrow_mob_right_white", "arrow_mob_left_white", "arrow_mob_right_green", "arrow_mob_left_green"]);
        $resolver->setAllowedValues('icon_2', ["ico_add_card_white", "button_search_loupe", "ico_plus", "arrow_grey_right", "arrow_mob_right_orange", "arrow_mob_left_orange", "arrow_mob_right_white", "arrow_mob_left_white", "arrow_mob_right_green", "arrow_mob_left_green"]);
        $resolver->setAllowedValues('text_transform', ["capitalize", "lowercase", "uppercase"]);
        $resolver->setAllowedValues('color', ["default", "green"]);
    }
```

~~Right now I can't pass any HTML attributes, because there is no direct way to do it and my validation will block any unmapped keys.~~

2. ~~In my PHP component I use `action` and `class` properties. I won't be able to pass the same HTML attributes into TWIG component, because my properties will have priority and values with these keys will be assigned to them. So in general there is no way to pass for example `action` property into form if that name is already taken by PHP component. I can of course change my names, but in general such approach may lead to unexpected behaviour.~~ 

~~For me it would be better to provide more direct way to pass such HTML attributes other then using unmapped keys. Then components can be more strict and much easier validatable against disallowed properties. I don't know if it would be possible to introduce third argument to pass other options and do not mix them with variables. Something like:~~

```twig
{{ component('alert', {properties: values}, {htmlAttributes: attributes}) }}
```

~~If it's not possible maybe special property named: `_htmlAttributes`~~

```twig
{{ component('alert', {key_1: value_1, _htmlAttributes: {class: 'red', type: 'button'}}) }}
```

3. Right now I use attributes [Twig Extension](https://api.druphelp.com/api/drupal/core%21lib%21Drupal%21Core%21Template%21TwigExtension.php/class/TwigExtension/9) created by Drupal team. 

```php
final class AttributesExtension extends AbstractExtension implements GlobalsInterface
{
    /**
     * @return mixed[]
     */
    public function getGlobals(): array
    {
        return [
            'attributes' => new Attributes(),
        ];
    }
}
```

However your implementation conflicts with my globals with following error

```
Neither the property "setAttribute" nor one of the methods "setAttribute()", "getsetAttribute()"/"issetAttribute()"/"hassetAttribute()" or "__call()" exist and have public access in class "Symfony\UX\TwigComponent\ComponentAttributes
```

I wonder if is possible to avoid such conflicts by prefixing by you things added to Twig? It's not a problem for me to change name from `attributes` to something else. However I think more about any other conflicts that may arise in future. 

**EDIT** I saw in code that you already use special property like `attributes` to pass them directly, so the first two points make no sense any more.

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #275
| License       | MIT

To avoid possible collisions with other variables/libraries, it was requested to make the `attributes` variable configurable. Additionally, I added an option to disable exposing public props:

```php
#[AsTwigComponent('my_component', attributesVar: '_attributes', exposePublicProps: false)]
class MyComponent {}
```

In the component template, no public properties are available directly (unless explicitly marked with `ExposeInTemplate` and attributes are available as `_attributes`.

### DIFF
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -21,13 +21,25 @@ use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 #[\Attribute(\Attribute::TARGET_CLASS)]
 final class AsLiveComponent extends AsTwigComponent
 {
-    public ?string $defaultAction;
+    public function __construct(
+        string $name,
+        ?string $template = null,
+        private ?string $defaultAction = null,
+        bool $exposePublicProps = true,
+        string $attributesVar = 'attributes'
+    ) {
+        parent::__construct($name, $template, $exposePublicProps, $attributesVar);
+    }
 
-    public function __construct(string $name, ?string $template = null, ?string $defaultAction = null)
+    /**
+     * @internal
+     */
+    public function serviceConfig(): array
     {
-        parent::__construct($name, $template);
-
-        $this->defaultAction = $defaultAction;
+        return array_merge(parent::serviceConfig(), [
+            'default_action' => $this->defaultAction,
+            'live' => true,
+        ]);
     }
 
     /**
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -39,12 +39,7 @@ final class LiveComponentExtension extends Extension
             AsLiveComponent::class,
             function (ChildDefinition $definition, AsLiveComponent $attribute) {
                 $definition
-                    ->addTag('twig.component', array_filter([
-                        'key' => $attribute->name,
-                        'template' => $attribute->template,
-                        'default_action' => $attribute->defaultAction,
-                        'live' => true,
-                    ]))
+                    ->addTag('twig.component', array_filter($attribute->serviceConfig()))
                     ->addTag('controller.service_arguments')
                 ;
             }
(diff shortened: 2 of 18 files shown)


weaverryan (approved):
Makes sense - nice implementation 👍  - just need to fix the conflicts... from the other PR I merged of your's just now ;)

PR #279, opened 2022-02-18
Update index.rst

94noni:
Some improvement while reading form https://symfony.com/bundles/ux-twig-component/current/index.html

I am unsure of the code render via `::`
For the component pattern I took as exemple the `{# templates/components/featured_products.html.twig #}` defined in the middle of the page doc

ℹ️ the button `edit this page` points to the readonly repo, perhaps it should be updated

PR #280, opened 2022-02-18
Update index.rst

94noni:
IMHO it is better to add those note after the related twig code, as they add info on this note the php code above
feel free to close if it is done on purpose

PS: same as https://github.com/symfony/ux/pull/279, the link of the doc points to the readonly repo


kbond (approved):

kbond:
Thanks @94noni.

PR #295, opened 2022-03-14
[Live] fix bug when trying to hydrate nullable Doctrine Entity

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Fixes a regression in 2.1 where a nullable entity is attempted to be loaded from Doctrine. In 2.0, `null` was dropped from the component data loading the entity wasn't attempted.


kbond:
I wonder if this change should be in the hydration system itself:
1. If raw prop === the property's default, skip
2. If raw prop === null and the property allows null, skip

weaverryan:
> I wonder if this change should be in the hydration system itself:

Let's keep the behavior simple and predictable (i.e. always run a value through the hydration system) :)

weaverryan (approved):

weaverryan:
Thank you Kevin!

PR #297, opened 2022-03-16
Use correct method to set the Request Format

janklan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes, docs
| New feature?  | no
| Tickets       | n/a
| License       | MIT

The `Request::setFormat()` signature is `public function setFormat(?string $format, $mimeTypes)` in Symfony 5.4 and `public function setFormat(?string $format, string|array $mimeTypes)` in Symfony 6. Both versions require two arguments to be passed. The example appears to be calling an incorrect method. Shouldn't it call `Request::setRequestFormat()` instead?


weaverryan:
Thank you Jan!

PR #301, opened 2022-03-19
[Live] remove anemic class

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Small refactor to remove a useless internal class.


weaverryan (commented):
I'm fine with this. Though, I could counter that the anemic class is, sort of, a "typed array": making it easier to understand the 2 things that are being returned.

kbond:
Fair critique.

Elsewhere, I've been fiddling with patterns of parsing attributes that require the reflector as context like here. Using a tuple was one solution.

wouterj:
I would also like to add that objects provide an easier and smoother BC path when deprecating one of the returned items or changing their types. I don't know whether that is something that is needed here, but maybe it's something to consider as well :)

kbond:
Good point. In this case, it's all internal functionality. This is just me experimenting tbh :)

weaverryan:
It's subjective... but sorry... I'm going to close :). But coding practices change - if the "wind" blows more in this direction, we could do more of this in the future.

PR #309, opened 2022-04-05
Make adding a custom template more specific

htmlshaman1:
The docs are not wrong because 
 #[AsTwigComponent('alert',  'my/custom/template.html.twig')]

will indeed load the custom template.

But if you are using multiple parameters:
#[AsTwigComponent('alert', exposePublicProps:false,  'my/custom/template.html.twig')]
Just adding a value for the template won't work. You need specify it's a template.

The custom template works with the template parameter:
#[AsTwigComponent('alert', exposePublicProps:false,  template:'my/custom/template.html.twig')]

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Tickets       | Fix #... 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/Resources/doc/index.rst
+++ b/src/TwigComponent/src/Resources/doc/index.rst
@@ -184,7 +184,7 @@ as the second argument to the ``AsTwigComponent`` attribute:
       // ...
 
     - #[AsTwigComponent('alert')]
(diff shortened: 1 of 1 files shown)


kbond (approved):
Good idea, makes the examples consistent.

weaverryan:
I like this too - thanks @htmlshaman1!

PR #317, opened 2022-05-01
[Twig] Embedded components

### LINKED ISSUE #300: [Twig] Add tag component
Hi,

It would be great to have the tag `component`

```twig
{% component %}
    Content...
{% endcomponent %}
```

Here is an example that @weaverryan and @kbond had made:
https://github.com/weaverryan/twig-components-poc/blob/master/templates/main/index.html.twig#L7

Thanks ;)

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #300
| License       | MIT

This adds a `{% component name %}` tag to allow rendering components and overriding their _slots_ (by using twig's native block system).

Example:

```twig
{# templates/components/alert.html.twig #}
<div{{ attributes.defaults({class: 'alert alert-'~type}) }}>
    {% block default %}<strong>{{ type }}</strong> {{ message }}{% endblock %}
</div>

{# "embed" in another template: #}
{% component 'alert' with {message: 'embedded message', class: 'extra-class', role: 'alert'} %}
    {% block default %} {# override "default slot (block)" #}
        {{ parent() }} {# call parent "slot" #}
        Custom stuff...
        {{ this.message }} {# access component properties/methods as normal #}
        {{ computed.someMethod }} {# all component features available #}
    {% endblock %}
{% endcomponent %}

{# can still use the function as normal #}
{{ component('alert', {message: 'some message'}) }}
```

More Advanced Example:

```twig
{# templates/components/dialog.html.twig #}
<div{{ attributes }}>
    <h1>{% block title %}Default Title{% endblock %}</h1>
    <p>{% block body %}Default Body{% endblock %}</p>
    <div>{% block footer %}Default Footer{% endblock %}</div>
</div>

{# "embed" in another template: #}
{% component dialog %} {# quotes can optionally be left off the component name #}
    {% block title %}Custom Title{% endblock %}
    {% block body %}Custom Body{% endblock %}
    {% block footer %}Custom Footer{% endblock %}
{% endcomponent %}
```

Some notes:
1. Most code is copied and modified from twig's native `embed` tag
2. I really wanted to have a default slot to avoid the need for overriding the default block (see first example above) but couldn't figure out how to do this or if it's even possible. Maybe someone with a deeper knowledge of twig internals could help?
3. The `PreRender` event is still dispatched like normal but because of the nature of how this feature works, modifying the template via a listener is not supported.
4. Using with LiveComponent's is not supported and throws an exception if attempted. Perhaps some trickery could be added in the future to achieve but I think this feature adds enough value to stand on its own.
5. This with #283 could add more value I think - it would remove the need for simple components to be _backed by_ a component service.

**TODO:**
- [x] Update changelog
- [x] Tests
- [x] Docs
- [x] Clarify nomenclature, _nested_ components vs _embedded_ components


kbond:
Some context on what I would like the default slot feature to look like (using the first example above):

```twig
{# "embed" in another template: #}
{% component 'alert' with {message: 'embedded message', class: 'extra-class', role: 'alert'} %}
    {# anything not wrapped in a block is considered the default slot #}
    {{ parent() }} {# call parent "slot" #}
    Custom stuff...
    {{ this.message }} {# access component properties/methods as normal #}
    {{ computed.someMethod }} {# all component features available #}

    {% block title %}
        {# can still override "named slots" #}
        Custom Title
    {% endblock %}
{% endcomponent %}
```

I certainly don't think _not having_ this feature is a showstopper - just a nice DX improvement.

gharlan:
> Some context on what I would like the default slot feature to look like (using the first example above):

I'm not sure about the mixture of default slot and other blocks. But if only the default block is needed, I would also prefer the notation without `{% block %}`.

> I certainly don't think not having this feature is a showstopper - just a nice DX improvement.

100% agree!

WebMamba:
Thanks for your work kbond ! I tried it out, and it's working great ! I was thinking about the default slots with parent(). What do you think about doing it in the other way. We can use a children function in the component:

```php
{# templates/components/alert.html.twig #}
<div{{ attributes.defaults({class: 'alert alert-'~type}) }}>
		{{ children() }}
    {% block default %}<strong>{{ type }}</strong> {{ message }}{% endblock %}
</div>`

{# "embed" in another template: #}
{% component 'alert' with {message: 'embedded message', class: 'extra-class', type: 'alert'} %}
    {# anything not wrapped in a block is considered the children slot #}
    <p>My custom stuff...</p>
{% endcomponent %}
```

This children function is only accessible inside a component. I think it can be much easier to read, and closer to front-end framework like react. 🤔

gharlan:
> Some context on what I would like the default slot feature to look like (using the first example above):

fyi: found an issue regarding similar functionality for original embed tag: https://github.com/twigphp/Twig/issues/3572

weaverryan:
I love this! The code already looks clean. Let's finish this.

> I really wanted to have a default slot to avoid the need for overriding the default block (see first example above) but couldn't figure out how to do this or if it's even possible. Maybe someone with a deeper knowledge of twig internals could help?

I agree & also agree that NOT having this is NOT a show stopper. If we merge this, perhaps some brave curious person will come along and add this. My "hunch" is that it the implementation might involve "hacking in" a `{% block default %}` and then `{% endblock %}` into the token stream inside `ComponentTokenParser`, similar to how a fake `{% extends %}` is already added there. The tricky part would be to figure out IF there is already any blocks defined. My thinking is that (in part for simplicity of implementation) you would either (A) define zero blocks, and thus all content is the `default` block else (B) you define ALL blocks, including `default`. If we have this assumption, then we just need to determine if the content contains ANY blocks. And if not, hack in the `{% block default %}`. However, with more work, someone could probably make this even smarter.

> Using with LiveComponent's is not supported and throws an exception if attempted. Perhaps some trickery could be added in the future to achieve but I think this feature adds enough value to stand on its own.

Agree 👍. In theory, we could stash the raw Twig contents into a system cache, and re-render using that. A bit crazy, but I actually don't see any issue with it. But, later.

Thanks!

kbond:
I believe this to be ready for a final review.

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -637,14 +637,81 @@ the twig template and twig variables before components are rendered::
...
+
+Embedded Components
+-------------------
+
+.. versionadded:: 2.2
+
+    Embedded components were added in TwigComponents 2.2.
+
+You can write your component's Twig template with blocks that can be overridden
+when rendering using the ``{% component %}``. These blocks can be thought of as
```
```suggestion
when rendering using the ``{% component %}`` syntax. These blocks can be thought of as
```

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -637,14 +637,81 @@ the twig template and twig variables before components are rendered::
...
+            </tbody>
+        </table>
+        {% if block('footer') is defined %}
+            <div class="data-table-footer">
+                {{ block('footer') }}
+            </div>
+        {% endif %}
+    </div>
+
+When rendering, you can override the ``th``, ``th`` blocks and add a ``footer`` block.
```
```suggestion
When rendering, you can override the ``th`` and ``td`` blocks, and also add a ``footer`` block.
```

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -637,14 +637,81 @@ the twig template and twig variables before components are rendered::
...
+The ``with`` data is what's mounted on the component object.
+
+.. code-block:: twig
+
+    {# templates/some_page.html.twig #}
+
+    {% component table with {headers: ['key', 'value'], data: [[1, 2], [3, 4]]} %}
+        {% block th %}custom th ({{ parent() }}){% endblock %}
+        {% block td %}custom td ({{ parent() }}){% endblock %}
```
weaverryan: I like the example overall, but these aren't terribly realistic. What if we put the block around the actual `<td>` elements themselves, then show adding a class:

```
{% block td %}<td class="col-class">{{ parent() }}</td>{% endblock %}
```

src/TwigComponent/src/Resources/doc/index.rst
```diff
@@ -637,14 +637,81 @@ the twig template and twig variables before components are rendered::
...
+                        {% for cell in row %}
+                            <td>
+                                {% block td %}{{ cell }}{% endblock %}
+                            </td>
+                        {% endfor %}
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+        {% if block('footer') is defined %}
```
weaverryan: I think we should make this a tiny bit less fancy. Just:

```twig
```

Then below, when we render the component, we include the entire `<div class="data-table-footer">`.

My think is that, the main & simplest use case (I think) for this is where you have empty blocks that you want to fill in (e.g. filling in the content of an `alert` component). And currently, we only show more advanced use-cases (calling `parent()` & using a block with `is defined`).

weaverryan (approved):
Looks great! Minor comments on the docs

src/TwigComponent/src/Resources/doc/index.rst
kbond: Hmm, that won't quite work because `parent()` would include the `<td>`

weaverryan:
Thanks Kevin!

PR #319, opened 2022-05-06
[TwigComponent] Allow to pass stringable object as non mapped component attribute

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | N/A
| License       | MIT

Sometimes we have some value objects that are meant to be provided as stimulus controller values and when using `{{ component('..', { 'data-item-id-param': objectUuid }) }}` it throws an exception that attribute must be scalar.

Now it will allow to accept these objects.

cc @kbond

### DIFF
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -71,6 +71,10 @@ final class ComponentFactory
 
         // ensure remaining data is scalar
         foreach ($data as $key => $value) {
+            if ($value instanceof \Stringable) {
+                continue;
+            }
+
             if (!\is_scalar($value) && null !== $value) {
                 throw new \LogicException(sprintf('Unable to use "%s" (%s) as an attribute. Attributes must be scalar or null. If you meant to mount this value on "%s", make sure "$%1$s" is a writable property or create a mount() method with a "$%1$s" argument.', $key, get_debug_type($value), $component::class));
             }
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.2
+
+-   Allow to pass stringable object as non mapped component attribute
+
 ## 2.1
 
 -   Make public component properties available directly in the template (`{{ prop }}` vs `{{ this.prop }}`).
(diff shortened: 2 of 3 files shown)


src/TwigComponent/src/ComponentFactory.php
```diff
@@ -71,6 +71,10 @@ public function create(string $name, array $data = []): MountedComponent
 
         // ensure remaining data is scalar
         foreach ($data as $key => $value) {
+            if (\is_object($value) && method_exists($value, '__toString')) {
```
kbond:
```suggestion
            if (\is_object($value) && $value instanceof \Stringable) {
```

kbond (changes requested):
One small thing but looks good to me.

src/TwigComponent/src/ComponentFactory.php
norkunas: updated.
also from php docs:
> instanceof does not throw any error if the variable being tested is not an object, it simply returns false. Constants, however, were not allowed prior to PHP 7.3.0.

so removed the is_object :)

kbond: Perfect! I wondered about that too.

kbond (approved):

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -71,6 +71,10 @@ public function create(string $name, array $data = []): MountedComponent
 
         // ensure remaining data is scalar
         foreach ($data as $key => $value) {
+            if ($value instanceof \Stringable) {
```
weaverryan, line 74: Funny timing - we JUST did this with `stimulus_controller()` also. There, we went further and checked for `__toString()`: https://github.com/symfony/webpack-encore-bundle/pull/171/files#diff-aac7ca1aa215ab99c76587c963238c082fb42b7a4315344002b7df3046306b8fR64

kbond: I don't believe we need to check for the method because this lib is PHP 8+.

weaverryan: Good catch!

weaverryan:
Thanks Tomas!

PR #324, opened 2022-05-22
[LiveComponent] Add a generic LiveCollectionType

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | - 
| License       | MIT

Hello, this is just a quick draft, extracted from a project, to see if is it interesting enough to include here.

**TODO**

- [x] add to changelog
- [x] add docs
- [x] add tests
- [ ] add a demo

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -1220,6 +1220,106 @@ When the user clicks ``removeComment()``, a similar process happens.
     attribute above the ``Post.comments`` property. These help new
     items save and deletes any items whose embedded forms are removed.
 
+Using LiveCollectionType
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.2
+
+    The ``LiveCollectionType`` and the ``LiveCollectionTrait`` was added in LiveComponent 2.2.
+
+
+The ``LiveCollectionType`` uses the same method described above, but in
+a generic way, so it needs even less code. This form type adds an 'Add'
+and a 'Delete' button for each row by default, which work out of the box
+thanks to the ``LiveCollectionTrait``.
+
+Let's take the same example as before, a "Blog Post" form with an embedded "Comment" forms
+via the ``LiveCollectionType``::
+
+    namespace App\Form;
+
+    use Symfony\Component\Form\AbstractType;
+    use Symfony\Component\Form\FormBuilderInterface;
+    use Symfony\Component\OptionsResolver\OptionsResolver;
+    use Symfony\UX\LiveComponent\Form\Type\LiveCollectionType;
+    use App\Entity\BlogPost;
+
+    class BlogPostFormType extends AbstractType
+    {
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder
+                ->add('title', TextType::class)
+                // ...
+                ->add('comments', LiveCollectionType::class, [
+                    'entry_type' => CommentFormType::class,
+                    'allow_add' => true,
+                    'allow_delete' => true,
+                    'by_reference' => false,
+                ])
+            ;
+        }
+
+        public function configureOptions(OptionsResolver $resolver)
+        {
+            $resolver->setDefaults(['data_class' => BlogPost::class]);
+        }
+    }
+
+Now, create a Twig component to render the form::
+
+    namespace App\Twig;
+
+    use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+    use Symfony\Component\Form\FormInterface;
+    use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+    use Symfony\UX\LiveComponent\DefaultActionTrait;
+    use Symfony\UX\LiveComponent\LiveCollectionTrait;
+    use App\Entity\BlogPost;
+    use App\Form\BlogPostFormType;
+
+    #[AsLiveComponent('blog_post_collection_type')]
+    class BlogPostCollectionTypeComponent extends AbstractController
+    {
+        use LiveCollectionTrait;
+        use DefaultActionTrait;
+
+        #[LiveProp]
+        public BlogPost $post;
+
+        protected function instantiateForm(): FormInterface
+        {
+            return $this->createForm(BlogPostFormType::class, $this->post);
+        }
+    }
+
+There is no need for a custom template just render the form as usual:
+
+.. code-block:: twig
+
+    <div {{ attributes }} data-action="change->live#update">
+        {{ form(form) }}
+    </div>
+
+The ``add`` and ``delete`` buttons rendered as separate ``ButtonType`` form
+types and can be customized like a normal form type via the ``live_collection_button_add``
+and ``live_collection_button_delete`` respectively:
+
+.. code-block:: twig
+
+    {% block live_collection_button_add_widget %}
+        {% set attr = attr|merge({'class': attr.class|default('btn btn-ghost')}) %}
+        {% set translation_domin = false %}
+        {% set label_html = true %}
+        {%- set label -%}
+            <svg xmlns="http://www.w3.org/2000/svg" class="h-6 w-6" fill="none" viewBox="0 0 24 24" stroke="currentColor" stroke-width="2">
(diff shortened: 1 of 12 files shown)


src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1220,6 +1220,91 @@ When the user clicks ``removeComment()``, a similar process happens.
     attribute above the ``Post.comments`` property. These help new
     items save and deletes any items whose embedded forms are removed.
 
+Using LiveCollectionType
```
1ed, line 1223: Just copied the previous section and modified only the code parts..

src/LiveComponent/src/Resources/views/form_theme.html.twig
```diff
@@ -0,0 +1,35 @@
...
+        {{ form_row(button_delete_prototype, { attr: button_delete_prototype.vars.attr|merge({
+            'data-action': 'live#action',
+            'data-action-name': 'removeCollectionItem(name=' ~ form.parent.vars.full_name ~ ', index=' ~ form.vars.name ~ ')'
+        }) }) }}
+    {% endif %}
+{% endblock live_collection_entry_row %}
+
+{% block live_collection_button_add_widget %}
+    {{ block('button_widget') }}
+{% endblock live_collection_button_add_widget %}
```
1ed: This should be removed by default. Just added here to highlight we have block for the buttons, so we can customize as we like, e.g.:

```twig
{% block live_collection_button_add_widget %}
    {% set attr = attr|merge({'class': attr.class|default('btn btn-ghost')}) %}
    {% set translation_domin = false %}
    {% set label_html = true %}
    {%- set label -%}
        <svg xmlns="http://www.w3.org/2000/svg" class="h-6 w-6" fill="none" viewBox="0 0 24 24" stroke="currentColor" stroke-width="2">
            <path stroke-linecap="round" stroke-linejoin="round" d="M12 6v6m0 0v6m0-6h6m-6 0H6" />
        </svg>
        {{ 'form.collection.button.add.label'|trans({}, 'forms') }}
    {%- endset -%}
    {{ block('button_widget') }}
{% endblock live_collection_button_add_widget %}
```

src/LiveComponent/src/Resources/views/form_theme.html.twig
```diff
@@ -0,0 +1,35 @@
...
+        {%- if form is rootform -%}
+            {{ form_errors(form) }}
+        {%- endif -%}
+        {{- block('form_rows') -}}
+        {{- form_rest(form) -}}
+
+        {% if button_add_prototype is defined and not button_add_prototype.rendered %}
+            {{ form_row(button_add_prototype, { attr: button_add_prototype.vars.attr|merge({
+                'data-action': 'live#action',
+                'data-action-name': 'addCollectionItem(name=' ~ form.vars.full_name ~ ', index=' ~ (form.children is empty ? 0 : max(form.children|keys)) ~ ')'
```
norkunas: The `index` here that you provide is not used in `addCollectionItem` :)

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1220,6 +1220,91 @@ When the user clicks ``removeComment()``, a similar process happens.
...
+        }
+    }
+
+The template for this component has two jobs: (1) render the form
+like normal and (2) include links that trigger the ``addComment()``
+and ``removeComment()`` actions:
+
+.. code-block:: twig
+
+    <div {{ attributes }} data-action="change->live#update">
```
kbond, line 1300: Should we show adding the form theme?

src/LiveComponent/src/Resources/views/form_theme.html.twig
```diff
@@ -0,0 +1,35 @@
+{%- block live_collection_widget -%}
```
kbond, line 1: I think this should be a different name. Maybe `live_collection_form_theme.html.twig`?

kbond (commented):
This is great, thanks for working on this @1ed. Perfect use for live components!

src/LiveComponent/src/Resources/doc/index.rst
weaverryan: Because the new form theme template is adding a new block for a NEW form type (i.e. it does not modify any existing types), why not add it automatically? https://github.com/symfony/ux/blob/2.x/src/Dropzone/DependencyInjection/DropzoneExtension.php#L36

src/LiveComponent/src/Resources/views/form_theme.html.twig
```diff
@@ -0,0 +1,35 @@
...
+        {%- endif -%}
+        {{- block('form_rows') -}}
+        {{- form_rest(form) -}}
+
+        {% if button_add_prototype is defined and not button_add_prototype.rendered %}
+            {{ form_row(button_add_prototype, { attr: button_add_prototype.vars.attr|merge({
+                'data-action': 'live#action',
+                'data-action-name': 'addCollectionItem(name=' ~ form.vars.full_name ~ ')'
+            }) }) }}
+        {% endif %}
```
weaverryan: It's funny, #90 adds this at the top of the list, and you have it at the bottom. Nothing will work for everyone, but the bottom I think DOES seem better.

src/LiveComponent/src/Resources/views/form_theme.html.twig
```diff
@@ -0,0 +1,35 @@
+{%- block live_collection_widget -%}
+    {% set attr = attr|merge({'class': attr.class|default('space-y-4')}) %}
+    <div {{ block('widget_container_attributes') }}>
+        {%- if form is rootform -%}
+            {{ form_errors(form) }}
+        {%- endif -%}
+        {{- block('form_rows') -}}
+        {{- form_rest(form) -}}
```
weaverryan: It seems like you're repeating the `form_widget_compound` block. Could we `use` the core `form_div_layout.html.twig` template and then render that block directly?

src/LiveComponent/src/Resources/views/form_theme.html.twig
weaverryan: A very nice thing to document then :)

weaverryan (commented):
I think this is a really cool idea! Let's go for it!

src/LiveComponent/src/Resources/views/form_theme.html.twig
norkunas: Best would be probably both places :) saying this from experience when I've implemented this :)

weaverryan: Let's start with one... but it occurs to me that it may be be useful to configure this easily (e.g. `top`, `bottom` or `both`)

src/LiveComponent/src/Resources/views/form_theme.html.twig
1ed: I think there should be only one theme per component (and per variant like e.g. bootstrap or tailwind) and not per type, and maybe later we would like to add more blocks for an another type, so I would keep the name. WDYT?

src/LiveComponent/src/Resources/views/form_theme.html.twig
1ed: I done that because I wanted the `Add` button inside the `div` of the widget. Updated to use the `form_widget` block for now.

src/LiveComponent/src/Resources/views/form_theme.html.twig
kbond: Good points, makes sense to me.

seb-jean:
What is the difference with this PR?
https://github.com/symfony/ux/pull/90

kbond:
This requires your entire form be a live component, #90 does not. Basically, if your form already is a live component, just this would be required to add this functionality.

weaverryan (approved):
This is really wonderful work ❤️ . We may need more docs to mention that you should customize the `live_collection_widget` and `live_collection_entry_row` blocks in order to position the buttons just how you want them. I'm excited to try this in a real project and see how it goes :).

weaverryan:
Thank you Gábor!

1ed:
Thank you Ryan!

PR #327, opened 2022-05-29
[Turbo] Add expression language to topics

rskuipers:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | 
| License       | MIT

## Why
With the current implementation of the Broadcast attribute there's no way to scope newly added entities. A newly added entity will be broadcasted to anywhere that uses `turbo_stream_listen("App\\Entity\\Book")`. But there are use-cases where you want to limit this.

For one, imagine having a template `author/show.html.twig` where you want to show all the books belonging to a specific author. A newly added Book entity would show up on all the author pages because there's no way to supply a scope to turbo_stream_listen.

Secondly, there's currently no way to limit access based on the logged in user. For simplicity's sake let's imagine every author belongs to a publisher. The publisher can login and add/edit/remove authors that belong to them. Now let's say we have a template called `author/list.html.twig` on which we use `turbo_stream_listen("App\\Entity\\Author")`. In the current implementation, this would also show all newly created Authors created by other Publishers.

~~So adding a scope is both useful for showing specific listings as well as listings that should only be shown to a specific user.~~

Adding expression language support to topics makes it possible to scope these topics based on a relationship or some other context.

## ~~How~~
~~We can achieve the above by changing two points in the code.~~

~~https://github.com/rskuipers/ux/blob/feature/add-scope-option-to-broadcast/src/Turbo/Tests/app/Entity/Book.php#L23~~

~~Expanding the Broadcast attribute with a `scope` option that supports Expression Language. This will be executed when an entity is created/changed typically in order to retrieve an ID of a parent entity that will function as a scope.~~

~~Note: this expression has null checking because the relationship is optional. Also in Expression Language 6.1 you can use the [null-safe operator](https://symfony.com/doc/current/components/expression_language/syntax.html#null-safe-operator): `entity.author?.id`~~

~~https://github.com/rskuipers/ux/blob/feature/add-scope-option-to-broadcast/src/Turbo/Twig/TwigExtension.php#L44~~

~~Adding a third parameter `$scope = null` here so that you can provide a value used to add a scope to the topic you're listening to.~~

~~https://github.com/rskuipers/ux/blob/feature/add-scope-option-to-broadcast/src/Turbo/Twig/ScopedTurboStreamListenRendererInterface.php~~

~~Adding a separate interface with a method supporting a third parameter `?string $scope = null` here so that this can be integrated into the topic to listen for a scoped topic.~~

### DIFF
--- a/src/Turbo/Bridge/Mercure/Broadcaster.php
+++ b/src/Turbo/Bridge/Mercure/Broadcaster.php
@@ -11,9 +11,9 @@
 
 namespace Symfony\UX\Turbo\Bridge\Mercure;
 
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\Mercure\HubInterface;
 use Symfony\Component\Mercure\Update;
-use Symfony\UX\Turbo\Attribute\Broadcast;
 use Symfony\UX\Turbo\Broadcaster\BroadcasterInterface;
 
 /**
@@ -44,10 +44,17 @@ final class Broadcaster implements BroadcasterInterface
     private $name;
     private $hub;
 
+    /** @var ExpressionLanguage|null */
+    private $expressionLanguage;
+
     public function __construct(string $name, HubInterface $hub)
     {
         $this->name = $name;
         $this->hub = $hub;
+
+        if (class_exists(ExpressionLanguage::class)) {
+            $this->expressionLanguage = new ExpressionLanguage();
+        }
     }
 
     /**
@@ -69,7 +76,31 @@ final class Broadcaster implements BroadcasterInterface
             throw new \InvalidArgumentException(sprintf('Cannot broadcast entity of class "%s": either option "topics" or "id" is missing, or the PropertyAccess component is not installed. Try running "composer require property-access".', $entityClass));
         }
 
-        $options['topics'] = (array) ($options['topics'] ?? sprintf(self::TOPIC_PATTERN, rawurlencode($entityClass), rawurlencode(implode('-', (array) $options['id']))));
+        $topics = [];
+
+        foreach ((array) ($options['topics'] ?? []) as $topic) {
+            if (!\is_string($topic)) {
+                $topics[] = $topic;
+                continue;
+            }
+
+            if (0 !== strpos($topic, '@=')) {
+                $topics[] = $topic;
+                continue;
+            }
+
+            if (null === $this->expressionLanguage) {
+                throw new \LogicException('The "@=" expression syntax cannot be used without the Expression Language component. Try running "composer require symfony/expression-language".');
+            }
+
+            $topics[] = $this->expressionLanguage->evaluate(substr($topic, 2), ['entity' => $entity]);
+        }
+
+        $options['topics'] = $topics;
+
+        if (0 === \count($options['topics'])) {
+            $options['topics'] = (array) sprintf(self::TOPIC_PATTERN, rawurlencode($entityClass), rawurlencode(implode('-', (array) $options['id'])));
+        }
 
         $update = new Update(
             $options['topics'],
--- a/src/Turbo/composer.json
+++ b/src/Turbo/composer.json
@@ -55,7 +55,8 @@
         "symfony/security-core": "^5.2|^6.0",
         "symfony/stopwatch": "^5.2|^6.0",
         "symfony/twig-bundle": "^5.2|^6.0",
-        "symfony/web-profiler-bundle": "^5.2|^6.0"
+        "symfony/web-profiler-bundle": "^5.2|^6.0",
+        "symfony/expression-language": "^5.4|^6.0"
     },
     "conflict": {
         "symfony/flex": "<1.13"
--- a/src/Turbo/Bridge/Mercure/composer.json
+++ b/src/Turbo/Bridge/Mercure/composer.json
@@ -31,6 +31,9 @@
         "symfony/twig-bundle": "^5.2|^6.0",
         "symfony/ux-turbo": "^2.0"
     },
+    "suggest": {
+        "symfony/expression-language": "To use expression language in topics"
+    },
     "extra": {
         "thanks": {
             "name": "symfony/ux-turbo",
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.2
+
+-   The topics defined in the Broadcast attribute now support expression language when prefixed with `@=`.
+
 ## 2.1
 
 -   `TurboStreamResponse` and `AddTurboStreamFormatSubscriber` have been removed, use native content negotiation instead:
--- a/src/Turbo/CONTRIBUTING.md
+++ b/src/Turbo/CONTRIBUTING.md
@@ -26,6 +26,9 @@ Start the test app:
 -   `http://localhost:8000`: basic features
 -   `http://localhost:8000/chat`: chat using Turbo Streams
 -   `http://localhost:8000/books`: broadcast
+-   `http://localhost:8000/authors`: broadcast
+-   `http://localhost:8000/artists`: broadcast
+-   `http://localhost:8000/songs`: broadcast
 
 ## Run tests
 
--- a/src/Turbo/Resources/doc/index.rst
+++ b/src/Turbo/Resources/doc/index.rst
@@ -301,7 +301,7 @@ Forms
 ^^^^^
 
 .. versionadded:: 2.1
-  
+
     Prior to 2.1, ``TurboStreamResponse::STREAM_FORMAT`` was used instead of ``TurboBundle::STREAM_FORMAT``.
     Also, one had to return a new ``TurboStreamResponse()`` object as the third argument to ``$this->render()``.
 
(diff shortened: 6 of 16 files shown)


rskuipers:
This would obviously still need documentation, but I'm just awaiting confirmation before I start writing that.

src/Turbo/Bridge/Mercure/Broadcaster.php
```diff
@@ -71,6 +72,15 @@ public function broadcast(object $entity, string $action, array $options): void
 
         $options['topics'] = (array) ($options['topics'] ?? sprintf(self::TOPIC_PATTERN, rawurlencode($entityClass), rawurlencode(implode('-', (array) $options['id']))));
 
+        if (isset($options['scope'])) {
+            $expressionLanguage = new ExpressionLanguage();
+            $scope = $expressionLanguage->evaluate($options['scope'], [
+                'entity' => $entity,
+            ]);
+
+            $options['topics'][] = sprintf(self::TOPIC_PATTERN, rawurlencode($entityClass.'_'.$scope), rawurlencode(implode('-', (array) $options['id'])));
```
weaverryan: Question about this. If I'm passing in a "scope", do I want the update to be published BOTH to the "scoped topic" AND also the normal topic from above? Or ONLY the scoped topic? It looks like you purposely published to both topics - I'd like to know your thinking on that.

src/Turbo/Bridge/Mercure/TurboStreamListenRenderer.php
```diff
@@ -44,10 +58,10 @@ public function renderTurboStreamListen(Environment $env, $topic): string
                 throw new \LogicException(sprintf('Cannot listen to entity of class "%s" as the PropertyAccess component is not installed. Try running "composer require symfony/property-access".', $class));
             }
 
-            $topic = sprintf(Broadcaster::TOPIC_PATTERN, rawurlencode($class), rawurlencode(implode('-', $id)));
+            $topic = sprintf(Broadcaster::TOPIC_PATTERN, rawurlencode($class.(\is_string($scope) ? '_'.$scope : '')), rawurlencode(implode('-', $id)));
```
weaverryan:
```suggestion
            $topic = sprintf(Broadcaster::TOPIC_PATTERN, rawurlencode($class.(null === $scope ? '' : '_'.$scope)), rawurlencode(implode('-', $id)));
```

weaverryan: It might be worth adding a public static function to `Broadcaster` to help generate this string wherever we need to.

src/Turbo/Twig/TwigExtension.php
```diff
@@ -39,15 +39,26 @@ public function getFunctions(): iterable
...
         $transport = $transport ?? $this->default;
 
         if (!$this->turboStreamListenRenderers->has($transport)) {
             throw new \InvalidArgumentException(sprintf('The Turbo stream transport "%s" doesn\'t exist.', $transport));
         }
 
-        return $this->turboStreamListenRenderers->get($transport)->renderTurboStreamListen($env, $topic);
+        $turboStreamListenRenderer = $this->turboStreamListenRenderers->get($transport);
+
+        if (isset($scope)) {
```
weaverryan:
```suggestion
        if (null !== $scope) {
```

weaverryan (commented):
A few questions - but I think the approach is solid.

@dunglas - what do you think about this idea? tl;dr

1) Optional `scope` option added using expression language - e.g. `[Broadcast(scope: 'entity.author.id')]`
2) This causes the topic to change to `https://symfony.com/ux-turbo/App%5CEntity%5CBook_15/7` where `15` is the `entity.author.id` value and `7` is the main `Book` object's value that was just added/updated.

src/Turbo/Bridge/Mercure/Broadcaster.php
```diff
@@ -71,6 +72,15 @@ public function broadcast(object $entity, string $action, array $options): void
 
         $options['topics'] = (array) ($options['topics'] ?? sprintf(self::TOPIC_PATTERN, rawurlencode($entityClass), rawurlencode(implode('-', (array) $options['id']))));
 
+        if (isset($options['scope'])) {
+            $expressionLanguage = new ExpressionLanguage();
```
dunglas: Couldn't we store the instance in a property to prevent re-creating a next instance at every call to `broadcast()`?

dunglas:
Nice idea!

We have a similar feature in API Platform: https://github.com/api-platform/core/blob/main/src/Doctrine/EventListener/PublishMercureUpdatesListener.php#L225-L244

Basically, the idea is to let the user generate the whole topic using Expression Language. This has the benefit of giving full control to the user and to leverage all the capabilities of the underlying Transport.

What do you think about following this approach?

src/Turbo/Bridge/Mercure/Broadcaster.php
rskuipers: Let's take the Book and Author example again, just like the tests.
Imagine you want to show a list of every newly created book. By publishing to both topics we retain this possibility. You can subscribe to the scoped topic for books for a specific author, and the non-scoped topic for any book.

rskuipers:
@dunglas That's brilliant! Looks like it will fit here as well, definitely adds even more flexibility and it solves the problem with the current implementation where it's limited to 1 scope.

If @weaverryan agrees then I'll go ahead and reimplement it using your suggestion.

src/Turbo/Bridge/Mercure/TurboStreamListenRenderer.php
rskuipers: Resolving this since it's no longer relevant with the new approach.

src/Turbo/Bridge/Mercure/Broadcaster.php
rskuipers: This discussion no longer relevant with the new approach

rskuipers:
I have updated this PR to the method suggested by @dunglas.

dunglas (approved):

src/Turbo/Bridge/Mercure/Broadcaster.php
```diff
@@ -69,7 +72,27 @@ public function broadcast(object $entity, string $action, array $options): void
...
+                $topics[] = $topic;
+                continue;
+            }
+
+            if (0 !== strpos($topic, '@=')) {
+                $topics[] = $topic;
+                continue;
+            }
+
+            $topics[] = $this->expressionLanguage->evaluate(substr($topic, 2), ['object' => $entity]);
```
rskuipers: I used `object` here to follow the API platform example, but maybe `entity` is more consistent for symfony/ux...

src/Turbo/Tests/app/templates/authors.html.twig
```diff
@@ -0,0 +1,16 @@
+{% extends 'base.html.twig' %}
+
+{% block body %}
+    <h1>Authors</h1>
+
+    <div id="authors" {{ turbo_stream_listen('App\\Entity\\Author') }}></div>
+
+    <turbo-frame id="api">
```
weaverryan: Is it important that there is a `turbo-frame` here? I can see you copied it from something existing, but I'd like to only have it if it's necessary.

src/Turbo/Tests/app/templates/books.html.twig
```diff
@@ -3,12 +3,13 @@
 {% block body %}
     <h1>Books</h1>
 
-    <div id="books" {{ turbo_stream_listen('App\\Entity\\Book') }}></div>
+    <div id="books" {{ turbo_stream_listen('books') }}></div>
```
weaverryan: What was the significance of this change?

src/Turbo/composer.json
```diff
@@ -37,7 +37,8 @@
     },
     "require": {
         "php": ">=7.2.5",
-        "symfony/webpack-encore-bundle": "^1.11"
+        "symfony/webpack-encore-bundle": "^1.11",
+        "symfony/expression-language": "^5.4|^6.0"
```
weaverryan: I'd like to make this an optional requirement. Then, if it's not installed, we throw a clear exception:

> To use the expression language, please run "composer require symfony/expression-language".

weaverryan (commented):
Minor comments - but this is just about ready for merge - thanks!

src/Turbo/Tests/app/templates/authors.html.twig
rskuipers: Hm, not AFAIK. It's copied from [books](https://github.com/symfony/ux/blob/2.x/src/Turbo/Tests/app/templates/books.html.twig#L8), so it can probably be removed there as well.

src/Turbo/Tests/app/templates/books.html.twig
rskuipers: It's due to this: https://github.com/symfony/ux/pull/327/files#diff-de5c61126b72f0bca139694b277c0a2ac1cd5ff7b1873a3d695783a4e374e95fR23

I could also fallback on `App\\Entity\\Book` in that else in the expression but this is more in line with `book_by_author_1`.
If there's a preference I can change it.

src/Turbo/composer.json
rskuipers: I was wondering this too, I saw that the API Platform example also had expression language as optional. I'll implement it the same way.

src/Turbo/Bridge/Mercure/Broadcaster.php
rskuipers: @weaverryan do you have an opinion on this?

src/Turbo/Tests/app/templates/books.html.twig
dunglas: You should keep the current test as-is and create a new one (with a new entity if necessary) to test the new feature.

rskuipers: @dunglas Ah yes, of course, otherwise we don't know if BC was broken. 👍

weaverryan:
Friendly ping @rskuipers - I'm hoping to tag all of UX this week, it'd be great if this got in there :)

rskuipers:
I'll make some time for it! No problem :)

rskuipers:
@weaverryan I've reverted the existing books test and created separate entities to test the expression language feature as advised by @dunglas. I've also made expression language optional and changed `object` to `entity` to be more in-line with the rest of the code. I've also added a small bit of documentation. Not sure if you want to dedicate an entire chapter to it or if this is enough.

weaverryan:
Thank you Rick!

seb-jean:
Thanks @rskuipers for your feature! But how can we use a service in topics?

For instance:
```php
#[Broadcast(topics: ['@=service("App\\Service\\Generator").generate("songs_by_artist_" ~ entity.getId())'])]
```

PR #329, opened 2022-06-02
Introduce UX React component

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | -
| License       | MIT

This PR introduces a new component called UX React. This UX component creates a bridge between Symfony and React by allowing Twig developers to render React components, passing data smoothly between the two.

After installing this component, it is possible to use the `react_component` Twig function to render a React component in a specific HTML container:

```js
// assets/react/controllers/MyComponent.jsx
import React from 'react';

export default function (props) {
    return (
        <div>Hello {props.fullName}</div>
    );
}
```

```twig
{# templates/home.html.twig #}

<div {{ react_component('MyComponent', { 'fullName': app.user.fullName }) }}></div>
```

This behavior works by allowing UX React users to register React components in a global key-value store that is then used by a generic Stimulus controller (`Resources/assets/src/render_controller.tsx`) to do that the actual rendering. I propose to introduce a `assets/react/controllers` directory to automatically register all components inside this directory, to ease usage. You can read the README for more details.

This component supports Symfony 4.4+ (super simple bundle, no reason not to be IMO) and React 16+ (released in 2017).

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -971,11 +971,14 @@ function haveRenderedValuesChanged(originalDataJson, currentDataJson, newDataJso
 }
 
 function normalizeAttributesForComparison(element) {
-    if (element.value) {
-        element.setAttribute('value', element.value);
-    }
-    else if (element.hasAttribute('value')) {
-        element.setAttribute('value', '');
+    const isFileInput = element instanceof HTMLInputElement && element.type === 'file';
+    if (!isFileInput) {
+        if (element.value) {
+            element.setAttribute('value', element.value);
+        }
+        else if (element.hasAttribute('value')) {
+            element.setAttribute('value', '');
+        }
     }
     Array.from(element.children).forEach((child) => {
         normalizeAttributesForComparison(child);
--- a/src/React/DependencyInjection/ReactExtension.php
+++ b/src/React/DependencyInjection/ReactExtension.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\React\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\UX\React\Twig\ReactComponentExtension;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ *
+ * @internal
+ */
+class ReactExtension extends Extension
+{
+    public function load(array $configs, ContainerBuilder $container)
+    {
+        $container
+            ->setDefinition('twig.extension.react', new Definition(ReactComponentExtension::class))
+            ->setArgument(0, new Reference('webpack_encore.twig_stimulus_extension'))
+            ->addTag('twig.extension')
+            ->setPublic(false)
+        ;
+    }
+}
--- a/src/React/README.md
+++ b/src/React/README.md
@@ -1,0 +1,14 @@
+# Symfony UX React
+
+Symfony UX React integrates [React](https://reactjs.org/) into Symfony applications.
+It provides tools to render React components from Twig.
+
+**This repository is a READ-ONLY sub-tree split**. See
+https://github.com/symfony/ux to create issues or submit pull requests.
+
+## Resources
+
+-   [Documentation](https://symfony.com/bundles/ux-react/current/index.html)
+-   [Report issues](https://github.com/symfony/ux/issues) and
+    [send Pull Requests](https://github.com/symfony/ux/pulls)
+    in the [main Symfony UX repository](https://github.com/symfony/ux)
--- a/src/React/Resources/assets/package.json
+++ b/src/React/Resources/assets/package.json
@@ -1,0 +1,30 @@
+{
+    "name": "@symfony/ux-react",
+    "description": "Integration of React in Symfony",
+    "license": "MIT",
+    "version": "1.0.0",
+    "main": "dist/register_controller.js",
+    "symfony": {
+        "controllers": {
+            "react": {
+                "main": "dist/render_controller.js",
+                "webpackMode": "eager",
+                "fetch": "eager",
+                "enabled": true
+            }
+        }
+    },
+    "peerDependencies": {
+        "@hotwired/stimulus": "^3.0.0",
+        "react": "^18.0",
+        "react-dom": "^18.0"
+    },
+    "devDependencies": {
+        "@hotwired/stimulus": "^3.0.0",
+        "@types/react": "^18.0",
+        "@types/react-dom": "^18.0",
+        "@types/webpack-env": "^1.16",
+        "react": "^18.0",
+        "react-dom": "^18.0"
+    }
+}
--- a/src/React/Resources/assets/src/register_controller.ts
+++ b/src/React/Resources/assets/src/register_controller.ts
@@ -1,0 +1,30 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+export function registerReactControllerComponents(context: __WebpackModuleApi.RequireContext) {
+    const reactControllers: { [key: string]: object } = {};
+
+    const importAllReactComponents = (r: __WebpackModuleApi.RequireContext) => {
+        r.keys().forEach((key) => (reactControllers[key] = r(key).default));
+    };
+
+    importAllReactComponents(context);
+
+    // Expose a global React loader to allow rendering from the Stimulus controller
+    (window as any).resolveReactComponent = (name: string): object => {
+        const component = reactControllers[`./${name}.jsx`] || reactControllers[`./${name}.tsx`];
+        if (typeof component === 'undefined') {
+            throw new Error('React controller "' + name + '" does not exist');
+        }
+
+        return component;
+    };
+}
--- a/src/React/Resources/assets/src/render_controller.ts
+++ b/src/React/Resources/assets/src/render_controller.ts
@@ -1,0 +1,53 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import React, { ReactElement } from 'react';
+import { createRoot } from 'react-dom/client';
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+    readonly componentValue: string;
+    readonly propsValue: object;
+
+    static values = {
+        component: String,
+        props: Object,
+    };
+
+    connect() {
+        this._dispatchEvent('react:connect', { component: this.componentValue, props: this.propsValue });
+
+        const component = window.resolveReactComponent(this.componentValue);
+        this._renderReactElement(React.createElement(component, this.propsValue, null));
+
+        this._dispatchEvent('react:mount', {
+            componentName: this.componentValue,
+            component: component,
+            props: this.propsValue,
+        });
+    }
+
+    disconnect() {
+        (this.element as any).root.unmount();
+        this._dispatchEvent('react:unmount', { component: this.componentValue, props: this.propsValue });
+    }
+
+    _renderReactElement(reactElement: ReactElement) {
+        const root = createRoot(this.element);
+        root.render(reactElement);
+
+        (this.element as any).root = root;
+    }
+
+    _dispatchEvent(name: string, payload: any) {
+        this.element.dispatchEvent(new CustomEvent(name, { detail: payload, bubbles: true }));
+    }
+}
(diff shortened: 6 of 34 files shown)


src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -971,11 +971,14 @@ function haveRenderedValuesChanged(originalDataJson, currentDataJson, newDataJso
...
-    else if (element.hasAttribute('value')) {
-        element.setAttribute('value', '');
+    const isFileInput = element instanceof HTMLInputElement && element.type === 'file';
+    if (!isFileInput) {
+        if (element.value) {
+            element.setAttribute('value', element.value);
+        }
+        else if (element.hasAttribute('value')) {
+            element.setAttribute('value', '');
+        }
```
weaverryan, line 981: Thanks - this was my bad from a PR made earlier. I thought we had a check that building the files in CI didn't cause any changes, but we may have lost that at some point.

src/React/README.md
```diff
@@ -0,0 +1,59 @@
+# Symfony UX React
+
+Symfony UX React integrates [React](https://reactjs.org/) into Symfony applications. It provides tools 
+to render React components from Twig.
+
+**This repository is a READ-ONLY sub-tree split**. See
+https://github.com/symfony/ux to create issues or submit pull requests.
+
+## Usage
```
weaverryan: It looks like you've got the TODO already, but most of this should be moved into the `index.rst` file.

src/React/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash
+composer require symfony/ux-react
```
weaverryan:
```
$ composer require symfony/ux-react

# Don't forget to install the JavaScript dependencies as well and compile
$ yarn install --force
$ yarn encore dev
```

(though, unrelated, I'm thinking we should start leaning towards showing `npm` commands first, as the user will definitely have this installed).

src/React/README.md
```diff
@@ -0,0 +1,59 @@
...
+// Registers React controller components to allow loading them from Twig
+//
+// React controller components are components that are meant to be rendered
+// from Twig. These component then rely on other components that won't be called
+// directly from Twig.
+//
+// By putting only controller components in `react/controllers`, you ensure that
+// internal components won't be automatically included in your JS built file if
+// they are not necessary.
+registerReactControllerComponents(require.context('./react/controllers', true, /\.(j|t)sx?$/));
```
weaverryan: We MAY soon get the ability to do simple "patches" in Flex. If so, this could be added to a recipe.

src/React/Resources/assets/package.json
```diff
@@ -0,0 +1,31 @@
...
+            }
+        }
+    },
+    "peerDependencies": {
+        "@hotwired/stimulus": "^3.0.0",
+        "react": "^16.0 || ^17.0 || ^18.0",
+        "react-dom": "^16.0 || ^17.0 || ^18.0"
+    },
+    "dependencies": {
+        "@types/webpack-env": "^1.16"
```
weaverryan: Is this needed? I assume it's for some Typescript types? Is it needed in `dependencies` or is it a `devDependency`?

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import React, {ReactElement} from 'react';
```
weaverryan:
```suggestion
import React, { ReactElement } from 'react';
```

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    static values = {
+        component: String,
+        props: Object,
+    };
+
+    connect() {
+        this._dispatchEvent('react:connect', { component: this.componentValue, props: this.propsValue });
+
+        // Use a timeout to avoid mounting and demounting right after
```
weaverryan: why would it mount and then demount right after without the timeout?

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,72 @@
...
+        if ((this.element as any).timeout) {
```
weaverryan: I believe we can use this?

```suggestion
        if ((this.element as HTMLElement).timeout) {
```

And/or could the `timeout` be stored as a property on the component? I don't fully understand the problem, so these suggestions may be nonsense.

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,72 @@
...
+            clearTimeout((this.element as any).timeout);
+        }
+
+        (this.element as any).timeout = setTimeout(() => {
+            const component = window.resolveReactComponent(this.componentValue);
+            this._renderReactElement(React.createElement(component, this.propsValue, null));
+
+            this._dispatchEvent('react:mount', { component: this.componentValue, props: this.propsValue });
```
weaverryan: Pass the object too?

```suggestion
            this._dispatchEvent('react:mount', { componentName: this.componentValue, component: component, props: this.propsValue });
```

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,72 @@
...
+        reactDom.render(reactElement, this.element);
+
+        // Register a way to unmount this element
+        (this.element as any).unmount = () => {
+            reactDom.unmountComponentAtNode(this.element);
+        };
+    }
+
+    _dispatchEvent(name: string, payload: any) {
+        this.element.dispatchEvent(new CustomEvent(name, { detail: payload }));
```
weaverryan: Make this bubble: https://github.com/symfony/ux/blob/040e75d55b4c8c267ffef1b61cc2f8670974ab96/src/Typed/Resources/assets/src/controller.ts#L65

src/React/Resources/assets/test/fixtures/MyJsxComponent.jsx
```diff
@@ -0,0 +1,3 @@
+export default function () {
+    return <div>Hello</div>;
```
weaverryan, line 2: Hey! o/

src/React/Twig/ReactComponentExtension.php
```diff
@@ -0,0 +1,50 @@
...
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('react_component', [$this, 'renderReactComponent'], ['needs_environment' => true, 'is_safe' => ['html_attr']]),
+        ];
+    }
+
+    public function renderReactComponent(Environment $env, string $componentName, array $props = []): string
+    {
+        $props['key'] = md5($componentName.random_bytes(8));
```
weaverryan: I assume you're trying to solve the problem of a component being rendered inside of a list? If so, isn't this as bad as *not* specifying the key (except that you have suppressed the warning)? Or is the thinking that... this doesn't really apply to us, because we can't have a parent React component - e.g. `CommentList` that renders (and re-renders) a collection of children e.g. `Comment`?

If so, then it makes sense. But let's only set this if `key` does *not* exist already... just to be safe.

src/React/composer.json
```diff
@@ -0,0 +1,48 @@
...
+        "symfony/config": "^4.4.17|^5.0|^6.0",
+        "symfony/http-kernel": "^4.4.17|^5.0|^6.0",
+        "symfony/dependency-injection": "^4.4.17|^5.0|^6.0",
+        "symfony/webpack-encore-bundle": "^1.11"
+    },
+    "require-dev": {
+        "symfony/framework-bundle": "^4.4.17|^5.0|^6.0",
+        "symfony/phpunit-bridge": "^5.2|^6.0",
+        "symfony/twig-bundle": "^4.4.17|^5.0|^6.0",
+        "symfony/var-dumper": "^4.4.17|^5.0|^6.0"
```
weaverryan: Recently, for new components, we have been much more aggressive on versions. For example, Notify requires 8.1 and php 5.4: https://github.com/symfony/ux/blob/2.x/src/Notify/composer.json

Obviously, you've made this work with older versions of things. Still, I think we should require at least 8.0 and definitely 5.4 or higher.

weaverryan (commented):
WONDERFUL!

src/React/Resources/assets/package.json
tgalopin: Yes, it is used by registerReactControllerComponents and I think Typescript users of the package would need to have it installed for it to work (or perhaps is Typescript not checking the vendors? would make sense)

src/React/Resources/assets/src/render_controller.ts
tgalopin: https://stackoverflow.com/questions/66618415/why-is-my-stimulus-js-controller-firing-twice

Turbo with cache will trigger two connects in a row (cache then real) which means rendering the whole component again in a short time if the page renders quickly. I used this technique to circumvent the issue, maybe we don't want to care :) .

src/React/composer.json
tgalopin: I solved this by relying on Webpack Encore Bundle deps

src/React/Resources/assets/src/register_controller.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    const importAllReactComponents = (r: __WebpackModuleApi.RequireContext) => {
+        r.keys().forEach((key) => (reactControllers[key] = r(key).default));
+    };
+
+    importAllReactComponents(context);
+
+    // Expose a global React loader to allow rendering from the Stimulus controller
+    (window as any).resolveReactComponent = (name: string): object => {
+        const component = reactControllers['./' + name + '.jsx'] || reactControllers['./' + name + '.tsx'];
```
dunglas: Nitpicking (I'm not sure what our CS are regarding this).

```suggestion
        const component = reactControllers[`./${name}.jsx`] || reactControllers[`./${name}.tsx`];
```

src/React/Resources/assets/src/render_controller.ts
dunglas: Maybe should we at least provide a way to disable this (or maybe could we add this later if needed)? I fear that it may be the cause of some edge cases.

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,78 @@
...
+
+            // Register a way to unmount this element
+            (this.element as any).unmount = () => {
+                root.unmount();
+            };
+
+            return;
+        }
+
+        // Up to React 17 rendering
```
dunglas: Should we really support old versions of React in the new code we introduce? The React team only supports the current major version: https://github.com/reactjs/reactjs.org/issues/1745#issuecomment-466767389

 I'm not against supporting old versions, but I feel that it isn't worth it.

src/React/DependencyInjection/ReactExtension.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Symfony\UX\React\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\UX\React\Twig\ReactComponentExtension;
+
+/**
+ * @author Titouan Galopin <galopintitouan@gm ail.com>
```
dunglas:
```suggestion
 * @author Titouan Galopin <galopintitouan@gmail.com>
```

src/React/Resources/assets/src/render_controller.ts
tgalopin: That may be true indeed

tgalopin:
Updated after reviews. I dropped support for React 17 and less (not worth it indeed).

The PR is ready I think!

weaverryan (approved):

weaverryan:
Thank you Titouan!

src/React/Resources/assets/src/render_controller.ts
```diff
@@ -0,0 +1,53 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import React, { ReactElement } from 'react';
+import { createRoot } from 'react-dom/client';
```
weaverryan, line 13: Hmm, actually this is proving problematic. I merged, thinking the `yarn build` failure was something unrelated, but it's not. `yarn build` fails here with:

> Error: 'createRoot' is not exported by node_modules/react-dom/client.js, imported by src/React/Resources/assets/src/render_controller.ts

That is not correct... as `createRoot` IS an export of that file. I'm not sure if we may need `commonjs` for rollup (https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module... also ref similar issue: https://github.com/JedWatson/react-select/issues/3956#issuecomment-612104302

weaverryan: Looks like I fixed it over here: https://github.com/symfony/ux/pull/330

tgalopin: Great! I was going to have a look :)

kingjulien1:
is there a way to pass children to a react component? for example like this:

```js
// assets/react/controllers/MyComponent.jsx
import React from 'react';

export default function (props) {
    return (
        <div>
              <h1>heading</h1>
              {props.children}
        </div>
    );
}
```
```twig
{# templates/home.html.twig #}

<div {{ react_component('MyComponent', { 'fullName': app.user.fullName }) }}>
    <p>content to be displayed inside the react component}</p>
</div>
```

or, if not, is there a workaround to achieve this functionality?

weaverryan:
Could you open a separate issue? We COULD do this - in the Stimulus controller, we could read the `innerHTML` of the element and make it available as `props.children` via `dangerouslyInnerHtml` (we would make this opt-in - so we would need some sort of new option to enable this).

seb-jean:
How can Symfony forms be used with this component?

weaverryan:
What exactly would you want to accomplish?

seb-jean:
The idea is to click on a button, it opens a modal with the form in react.
For example a password change form.

PR #330, opened 2022-06-06
[Autocompleter] New Ajax-powered, autocomplete component

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | none
| License       | MIT

Hi! A new component! One I've wanted for a LONG time. In short, it creates things that look like this, where the autocomplete search is backed by Ajax.

ux-autocomplete-animation

Big thanks to EasyAdmin, whose AssociationField heavily inspired this.

* [ ] We could possibly avoid the requirement of creating a separate class for Ajax autocomplete by using signed URLs. However, this would only work for "simpler" fields - e.g. if you had a custom `query_builder` (or perhaps other options that we might not want to include in the URL, like `security`), then you would still be required to have the extra class. It's something to explore.

Cheers!

Recipe: https://github.com/symfony/recipes/pull/1094


src/Autocomplete/src/Doctrine/EntitySearchUtil.php
```diff
@@ -0,0 +1,122 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Autocomplete\Doctrine;
+
+use Doctrine\ORM\QueryBuilder;
+use Symfony\Component\Uid\Ulid;
+use Symfony\Component\Uid\Uuid;
+
+class EntitySearchUtil
```
weaverryan, line 23: Most of the code in this directory - including this file - are taken directly from EasyAdmin. Hence the already-mature-looking code.

seb-jean:
G R E A T 💯 
Would using https://github.com/afcapel/stimulus-autocomplete be better?
In particular to reuse the controller :
https://github.com/afcapel/stimulus-autocomplete/blob/main/src/autocomplete.js

weaverryan:
> Would using https://github.com/afcapel/stimulus-autocomplete be better?

I looked at this. IT';sn

weaverryan:
Apparently, I just tripped over my keyboard :)

> Would using https://github.com/afcapel/stimulus-autocomplete be better?

Tom Select (and this package) is focused on transforming `<select>` elements vs creating a generic "autocomplete" (e.g. a search bar where you have HTML results before). If you have that use-case, you can use that library, or live components where it's super easy to create this type of thing ("search bar where results load as you type").

package.json
```diff
@@ -17,7 +17,8 @@
         "@babel/preset-env": "^7.15.8",
         "@babel/preset-react": "^7.15.8",
         "@babel/preset-typescript": "^7.15.8",
-        "@rollup/plugin-node-resolve": "^13.0.0",
+        "@rollup/plugin-commonjs": "^22.0.0",
+        "@rollup/plugin-node-resolve": "^13.0.6",
```
tgalopin, line 21: Is this bump needed? The [changelog](https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md) doesn't seem to be related, but I may be wrong.

src/Autocomplete/composer.json
```diff
@@ -0,0 +1,54 @@
...
+            "Symfony\\UX\\Autocomplete\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Symfony\\UX\\Autocomplete\\Tests\\": "tests/"
+        }
+    },
+    "require": {
+        "php": ">=8.0"
```
tgalopin: Don't you need at least DI and Config?

src/Autocomplete/tests/Fixtures/Autocompleter/CustomProductAutocompleter.php
```diff
@@ -0,0 +1,55 @@
...
+use Doctrine\ORM\QueryBuilder;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\Security\Core\Security;
+use Symfony\UX\Autocomplete\EntityAutocompleterInterface;
+use Symfony\UX\Autocomplete\Tests\Fixtures\Entity\Product;
+
+class CustomProductAutocompleter implements EntityAutocompleterInterface
+{
+    public function __construct(
+        private RequestStack $requestStack
```
tgalopin, line 15: Single line

src/Autocomplete/tests/Fixtures/Form/CategoryAutocompleteType.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Security\Core\Security;
+use Symfony\UX\Autocomplete\Form\AsEntityAutocompleteField;
+use Symfony\UX\Autocomplete\Form\ParentEntityAutocompleteType;
+
+#[AsEntityAutocompleteField]
+class CategoryAutocompleteType extends AbstractType
+{
+    public function __construct(
+        private RequestStack $requestStack
```
tgalopin: Single line

tgalopin (approved):
This is great, thanks Ryan!

package.json
weaverryan: Yea, it was - when I brought in commonjs, it gave me an error that said that 13.0.6 was required. A bit specific... but that was the reason :)

src/Autocomplete/src/Form/ParentEntityAutocompleteType.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $form = current(iterator_to_array($forms, false));
+        $form->setData($data);
+    }
+
+    public function mapFormsToData($forms, &$data)
+    {
+        $form = current(iterator_to_array($forms, false));
+        $data = $form->getData();
+    }
```
bendavies, line 88: what does this do?

PR #343, opened 2022-06-15
Include required changes to bootstrap.js in ux-live-component docs

jannes-io:
| Q             | A
| ------------- | ---
| License       | MIT

When installing ux-live-components, it did not automatically modify bootstrap.js, so installation instructions were incomplete in my experience (near-fresh Symfony 6.2 installation with stimulus working). I compared with the demo, and they had this crucial piece of JavaScript in `assets/bootstrap.js` that was required to actually make the components "live". There's no mention of this in the documentation at all. Even if some automatic process didn't happen on my machine, it's still worth mentioning in the installation instructions since the file technically belongs to the user.

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -90,8 +90,26 @@ Now install the library with:
     $ yarn install --force
     $ yarn watch
 
-Also make sure you have at least version 3.2 of
+Make sure you have at least version 3.2 of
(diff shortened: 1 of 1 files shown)


kbond:
I believe the automatic process with work in the next release that includes https://github.com/symfony/ux/pull/336.

weaverryan:
That's correct - we will hopefully tag a new release this week, which will include #336, and then no modification of `bootstrap.js` will be needed anymore :).

Thanks for the close attention!

jannes-io:
Oh that’s even better. Nice work!

PR #345, opened 2022-06-16
[Typed] Using -- instead of / in name

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | none
| License       | MIT

The "/" is not normalized to -- when you define it explicitly. So we need to use -- here. This was an oversight when I set the name. @symfony/stimulus-bridge could also be changed to normalize this. But let's fix it here now.


weaverryan:
Actually, it's fixed in stimulus-bridge, so let's keep it here.

PR #352, opened 2022-06-17
Improve .gitattributes files

jmsche:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | N/A
| License       | MIT

Hi,

This PR improves the .gitattributes files so composer will only download necessary files in our projects :)

### DIFF
--- a/src/Autocomplete/.gitattributes
+++ b/src/Autocomplete/.gitattributes
@@ -1,1 +1,7 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.symfony.bundle.yaml export-ignore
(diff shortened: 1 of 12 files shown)


weaverryan (approved):
Look good to me. This is one of the many things that need to be "enforced" for consistency, either with a generator script that can be executed to "write" all of these files (which could be run in CI, then fails if the tree is dirty) or at least via a CI check. I have a long list of things that need checks or scripts like this, so we can keep consistency with low effort.

weaverryan:
Thanks @jmsche!

PR #367, opened 2022-06-23
[Autocomplete] Not destroyed on disconnect()

### LINKED ISSUE #366: [Autocomplete] Not destroyed on disconnect()
When using the Autocomplete package, if the Autocomplete Form Type is replaced in the DOM, when the Controller is disconnected the TomSelect instance is not destroyed.
A new one is mounted on the new DOM instance that result in 2 "search box".

In the next exemple, the `<select>` element is totally replaced by a new one, replace only options is not take in consideration per TomSelect.
https://user-images.githubusercontent.com/10453508/175276261-2c3656d8-114f-4092-a375-0822b04af7a0.mp4

I don't know if just add a `disconnect()` function is sufficient:
```js
    disconnect() {
        this.tomSelect.destroy();
    }
```

mpiot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #366
| License       | MIT

Call the `destroy()` method from TomSelect on `disconnect` event of the Controller. (https://tom-select.js.org/docs/api/)

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -39,6 +39,10 @@ export default class extends Controller {
         this.tomSelect = this.#createAutocomplete();
     }
 
+    disconnect() {
+        this.tomSelect.destroy();
+    }
+
     #getCommonConfig(): Partial<TomSettings> {
         const plugins: any = {}
 


weaverryan:
Good catch on this - thank you Mathieu!

tgalopin:
I built the dist file for this in https://github.com/symfony/ux/pull/368

PR #368, opened 2022-06-24
Allow to use react_component without props

### LINKED ISSUE #364: [Symfony UX React] Component without props
Hey Symfony team! Thanks for all the amazing work you did so far! 🙏🏻 

`"symfony/ux-react": "^2.2"`

I've noticed an issue when I tried to render react component without props using the `react_component` twig function.

When I try to render react component like this:
```twig
<div {{ react_component('MyComponent') }}></div>
```
I get the following console error:
image

As you can see from the image, it renders props as an array instead an object

It works this way (without props data attribute)
```html
<div data-controller="symfony--ux-react--react" data-symfony--ux-react--react-component-value="MyComponent"></div>
```

I think that props should be optional.

Thanks!

tgalopin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix https://github.com/symfony/ux/issues/364
| License       | MIT

Allow to use react_component without props:

```twig
<div {{ react_component('MyComponent') }}></div>
```

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -1,7 +1,7 @@
 import { Controller } from '@hotwired/stimulus';
 import TomSelect from 'tom-select';
 
-/*! *****************************************************************************
+/******************************************************************************
 Copyright (c) Microsoft Corporation.
 
 Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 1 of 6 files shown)


weaverryan:
Thanks Tito!

PR #375, opened 2022-07-04
[Typed] Use stimulus values api to set default values

nexxome:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | None
| License       | MIT

It isn't possible to set some boolean options to "false", because they were always overridden by the default value, for example:

`showCursor: this.showCursorValue || true`

So in this commit the stimulus Values API is used for default values https://stimulus.hotwired.dev/reference/values#default-values


weaverryan:
Thank you David! We likely need more of this in other components too - the "defaults" api for values was new in 3.0, and many of the components predate that and so haven't taken advantage of this yet :).

PR #376, opened 2022-07-04
Fix syntax, missing closing `}}`

maartendekeizer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes in docs
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Tried to copy paste the example, but did't work because of missing closing `}}`.
I also suggests to including the closing `</select>` tag.


weaverryan:
Good catch - thanks @maartendekeizer! I don't feel too strongly either way about the closing select tag :)

PR #390, opened 2022-07-11
feat(composer): allow symfony 5.4 app with php min version

### LINKED ISSUE #389: [Notify] Allow symfony 5.4 app with min php version.
Symfony 5.4 allow php 7.2.5 but ux-notify require php 8.1.

it would be interesting to allow this component, I didn't see any blocking things

thewalkingcoder:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | ye
| Tickets       | Fix #389
| License       | MIT

Allow php 7.2.5 and change compatibility method signature

### DIFF
--- a/src/Notify/Twig/NotifyRuntime.php
+++ b/src/Notify/Twig/NotifyRuntime.php
@@ -24,10 +24,15 @@ use Twig\Extension\RuntimeExtensionInterface;
  */
 final class NotifyRuntime implements RuntimeExtensionInterface
 {
+    private $hub;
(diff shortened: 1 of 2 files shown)


weaverryan:
Hi! 

This is definitely reasonable, but we've chosen - for new packages - to be a bit more aggressive on PHP versions. You're right that Symfony 5.4 only requires 7.2.5, but we've chosen to require php 8.1 for the new UX packages.

Sorry, but cheers!

PR #397, opened 2022-07-18
Add ux collection type component

alexander-schranz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | Fix #... 
| License       | MIT

This PR is based on the awesome work of @stakovicz in: https://github.com/symfony/ux/pull/90/

### TODO

 - [x] Move logic for data attribute from twig files to form types
 - [x] Use Symfony Button FormType for Add and Remove
 - [x] Implement Toolbar FormType for extendability over Form Extensions (Copy & Paste, ..)
 - [x] Add Toolbar and ButtonType over symfony Form for easier theming and extending
 - [x] Make JS component compatible with new form type and simplify the logic
 - [x] Fix possibility of overwrite `data-controller` for customization
 - [x] Keep nested collections in mind
 - [ ] Find a way to use `entry_type` instead of `ux_entry_type`


src/FormCollection/Form/UXCollectionEntryToolbarType.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\Form\Extension\Core\Type\ButtonType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @final
+ * @experimental
+ */
+class UXCollectionEntryToolbarType extends AbstractType
+{
```
alexander-schranz, line 15: The `UXCollectionEntryToolbarType` and `UXCollectionToolbarType` allows to add additional buttons later via a form extension. Example can be a `Copy & Paste` Functionality.

dunglas:
Does this approach support nested dynamic collections (here is a test case: https://github.com/symfony/ux/tree/e901026aa1d2e2cd2b27a5796962bc8fc4b5dcca/src/Collection/Tests/app/Form)? 

For the record, I proposed an alternative patch in https://github.com/symfony/ux/pull/398 which doesn't require changing the existing forms to use this feature (see https://github.com/symfony/ux/pull/398#issuecomment-1188619481 for the rationale).

alexander-schranz:
@dunglas yes you can have and I'm using nested collections this way. But like documented in the `CollectionType` you should use the different `prototype_name` to avoid conflicts between the 2 prototypes.

Kocal:

Since this PR didn't have any activity for years, I'm closing it.
Feel free to re-open it if necessary.

Thanks!

PR #400, opened 2022-07-19
Add Collection component as form type extension

alexander-schranz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | Fix #... 
| License       | MIT

This is based on discussion in #397, #398 and #90.

It targets to combine all discussion points from this pull requests:

 - [x] Plug & Play Functionality like in #398
 - [x] Avoid using of a `[data-prototype]` selector (issue in #397)
 - [x] Avoid wrapping or usage of custom form type (issue in #398)
 - [x] Add support for nested collection with same prototype_name (regex by @dunglas form #398)
 - [x] Render Add and Remove Button using Form Theme Buttons (#397, #90)
   - [x] Allow option for it to control (#397, #90)

## TODO

 - [ ] Tests
 - [ ] Docs
 - [ ] Discuss open TODOs
 
## Requirements for a Collection Form Type

After having a chat with @weaverryan today (2022-08-04) I want to add some comments from the chat here so they are not lost in Slack, about the requirements:

- [ ] A) Clear way to be able to customize the “entry” (this is a Symfony thing, but still good to show it in examples), add button & remove button, including their positioning. This doesn’t necessarily need to be “super easy” or “automatic”. As long as it is “reasonable” and well-documented, then I’m happy. Currently achieved in the theming. I need to make it updated to the latest version.

- [ ] B) Adding more buttons to a “toolbar” like “Cut & Copy” or “Paste” is interesting… though it seems pretty custom. This should either (1) be included in UX at some point or (2) at the very least, this should be reasonably possible for a user to add by extending UX. In other words, UX should at least allow for this possibility.

- [ ] C) For having an “add” button on each “row”, I’ve never thought about this, but that would be a great thing to support.

- [ ] D) Sortable: I think this is requested with a decent frequency. This would be interesting to support in the future.

A is the  most important thing and B, C, D are interesting feature which should keep in mind to be able to support in userland or in our own code in the future.


codedmonkey:
Hi there, I don't want to disturb your work on this but I'm very interested to see this come to life. I've been trying to build it myself since before the UX initiative was announced although it definitely isn't finished. You can check out my code [here](https://github.com/onlinq/form-collection) but you're unlikely to get any new insights from it haha. Thanks :)

src/Collection/Form/CollectionTypeExtension.php
```diff
@@ -0,0 +1,119 @@
...
+        $prototype = $form->getConfig()->getAttribute('prototype');
+
+        if (!$prototype) {
+            return;
+        }
+
+        if ($options['allow_delete']) {
+            // add delete button to rendered elements from the Collection ResizeListener
+            foreach ($form as $child) {
+                $child->add('deleteButton', $options['delete_type'], $options['delete_options']);
```
ihmels, line 51: I get an `AlreadySubmittedException` when the form is submitted and shown again (e. g. due to validation errors).

> You cannot add children to a submitted form.

I think the solution is to add the buttons in an `FormEvents::PRE_SET_DATA` event listener in `buildForm()`:


```php
$builder->addEventListener(FormEvents::PRE_SET_DATA, function (FormEvent $event) use ($options) {
    $form = $event->getForm();

    foreach ($form as $child) {
        $child->add('deleteButton', $options['delete_type'], $options['delete_options']);
    }

    $form->add('addButton', $options['add_type'], $options['add_options']);
});
```

ihmels: The add button needs to be re-added to the form again on submit because it is removed in the `ResizeFormListener`.

```php
$builder->addEventListener(FormEvents::SUBMIT, function (FormEvent $event) use ($options) {
    $form = $event->getForm();

    if ($form->has('addButton')) {
        return;
    }

    $form->add('addButton', $options['add_type'], $options['add_options']);
});
```

connorhu:
Some controller method ideas:
- deleteEntry (private)
- deleteAll (public)
- get entries (public?)
- deleteAtIndex (public)
- entryAtIndex (public)

src/Collection/Form/CollectionTypeExtension.php
```diff
@@ -0,0 +1,119 @@
...
+
+        if (!$prototype) {
+            return;
+        }
+
+        // TODO add button only if `delete_type` is defined and set `delete_type` default to null?
+        if ($options['allow_delete']) {
+            // add delete button to prototype
+            // TODO add toolbar here to allow extension add other buttons
+            $prototype->add('deleteButton', $options['delete_type'], $options['delete_options']);
```
connorhu, line 35: when i use this code with a simple form type this throws: "You cannot add children to a simple form. Maybe you should set the option "compound" to true?"

```
$builder->add('fields', CollectionType::class, [
            'entry_type' => ChoiceType::class,
```

I'm not sure that it's a good idea to force a complex form just for frontend functionality. It seems more logical to me to put the delete button only in the template (like the maker bundle does with the submit button https://github.com/symfony/maker-bundle/blob/main/src/Resources/skeleton/crud/templates/_form.tpl.php )

alexander-schranz: Using not the button of the form theme make theming a lot harder. And I really recommend when possible use the button which is rendered by the form theme. So it works out of the box with the different form themes and don't require special form themes again for every css framework.

Kocal:

Since this PR didn't have any activity for years, I'm closing it.
Feel free to re-open it if necessary.

Thanks!

PR #401, opened 2022-07-26
[Autocomplete] Allow plugins to be disabled through `tom_select_options.plugins.<plugin> = false`

### LINKED ISSUE #379: [Autocomplete] Don't want a clear button
A clear button is automaticly added when the autocomplete is not disabled and not multiple:
```
        if (!this.formElement.disabled && !isMultiple) {
            plugins.clear_button = { title: '' };
        }
```

How can we always disable the clear_button plugin?

Already tried, but that will result in a error
```
{{ stimulus_controller('symfony/ux-autocomplete/autocomplete', {tom_select_options: {maxItems: 1, plugins: {clear_button: null}}})}}
```

maartendekeizer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #379
| License       | MIT

When plugins option is defined Symfony UX will no longer add some plugins by default. This make it possible to render a autocomplete without clear button extensions for example.

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -330,10 +330,10 @@ export default class extends Controller {
             ...config1,
             ...config2,
             // Plugins from both configs should be merged together.
-            plugins: {
+            plugins: this.#normalizePlugins({
                 ...this.#normalizePluginsToHash(config1.plugins || {}),
                 ...this.#normalizePluginsToHash(config2.plugins || {}),
-            },
+            }),
         };
     }
 
@@ -357,6 +357,16 @@ export default class extends Controller {
 
         return plugins;
     };
+
+    #normalizePlugins(plugins: TPluginHash): TPluginHash {
+        return Object.entries(plugins).reduce((acc, [pluginName, pluginOptions]) => {
+            if (pluginOptions !== false) {
+                acc[pluginName] = pluginOptions;
+            }
+
+            return acc;
+        }, {} as TPluginHash);
+    }
 
     /**
      * Returns the element, but only if it's a select element.
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,33 @@
 # CHANGELOG
 
+## 2.28.0
+
+-   Default plugins like `clear_button` or `remove_button` can now be removed when setting their value to `false` in the `tom_select_options.plugins` option, for example:
+```php
+<?php
+#[AsEntityAutocompleteField]
+class IngredientAutocompleteType extends AbstractType
+{
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefaults([
+            'class' => Ingredient::class,
+            'tom_select_options' => [
+                'plugins' => [
+                    'clear_button' => false, // Disable the clear button
+                    'remove_button' => false, // Disable the remove button
+                ],
+            ],
+        ]);
+    }
+
+    public function getParent(): string
+    {
+        return BaseEntityAutocompleteType::class;
+    }
+}
+```
+
 ## 2.25.0
 
 -   Escape `querySelector` dynamic selector with `CSS.escape()` #2663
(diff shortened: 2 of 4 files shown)


src/Autocomplete/assets/src/controller.ts
```diff
@@ -46,18 +46,21 @@ export default class extends Controller {
...
-        }
-
-        if (isMultiple) {
-            plugins.remove_button = { title: '' };
-        }
-
-        if (this.urlValue) {
-            plugins.virtual_scroll = {};
+        // automaticly add plugins when NOT explicit configured
+        if (!this.tomSelectOptionsValue || !this.tomSelectOptionsValue.plugins !== undefined) {
```
weaverryan: Is this line right? That extra `!` in front of `this.tomSelectOptionsValue.plugins` looks funny to me.

I'd love to have a test for this. It would live in https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/test/controller.test.ts. In the existing `pre-connect` listener - https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/test/controller.test.ts#L23 - we could add:

```js
this.element.tomSelectOptions = event.detail.options;
```

Then, we could add a test that passes a `data-autocomplete-tom-select-options-value="{ plugins: { virtual_scroll: {} } }"` and verify with something like `getByTestId(container, 'main-element').tomSelectOptions.plugins.toBe({ virtual_scroll: {} })`.

weaverryan (commented):
I like this approach - just one comment

maartendekeizer:
@weaverryan Bit off topic but related to solving your feedback: Where can I find information about how to run tests cases etc. I think I am missing a good _contribute to_ symfony/ux page. The part in the readme is very short and is also missing a part about building...

weaverryan:
@maartendekeizer that's fair - that could certainly use some help :).

To run tests for a specific component, from the root of the project

```
yarn install
cd src/Autocomplete/assets
yarn jest

# or to run a specific test
yarn jest -t 'connect without options'
```

Let me know if that gives you any trouble - the tests can be tricky, but the existing ones in autocomplete should be fairly similar to what you're trying to do.

weaverryan:
Friendly ping @maartendekeizer :)

daFish:
Hey @maartendekeizer is there anything I can help with? 😄

daFish:
@weaverryan How should we proceed with this PR?

simonsolutions:
+1 for a solution for this problem. I'm facing it too and would like to have the PR procceded.

simonsolutions:
I would vote for merging it to the next release please. We really need the fix.

simonsolutions:
As the PR seems mergable without conflict, when can we expect this fix to be released?

Kocal:
Working on it

src/Autocomplete/assets/src/controller.ts
Kocal: Indeed the `!` is weird, but asserting on `this.tomSelectOptionsValue.plugins !== undefined` to not adding plugins is a BC. 

The current behavior is to merge plugins from user configuration and the one we define here. If we want to permit the user to remove **all** plugins, then `plugins: false` should be passed IMO

Kocal: Or better, add specific options like `no_plugin_clear_button` and `no_plugin_remove_button`...

Kocal: Or much better, keep it simple and do not register the plugin if `false` is passed, like `plugins: ['clear_button' => false]`

Kocal:
I've pushed a new commit that allows default plugins to be disableable when passing `false`, e.g.:
```php
<?php
#[AsEntityAutocompleteField]
class IngredientAutocompleteType extends AbstractType
{
    {
        $resolver->setDefaults([
            'class' => Ingredient::class,
            'tom_select_options' => [
                'plugins' => [
                    'clear_button' => false, // Disable the clear button
                    'remove_button' => false, // Disable the remove button
                ],
            ],
        ]);
    }

    {
    }
}
```

I added some tests too

Kocal (approved):

src/Autocomplete/assets/src/controller.ts
Kocal: Done in https://github.com/symfony/ux/pull/401#issuecomment-3110296129

kbond (approved):

Kocal:
Thank you @maartendekeizer.

CODEheures:
Thank you all. I don't find this in current documentation, but it work.

PR #404, opened 2022-07-29
[Live][POC] Defining interface for files uploader

Lustmored:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

This is an attempt to define interface(s) that would allow fully working file upload system and allow high level of customization. The idea comes from discussion in #395 and is by no means finished. We need to start somewhere :)

I was trying to achieve interface with parameters that can be easily defined with attributes and then utilized by library, so that developers won't need to use uploaders directly in most common cases. The problem with `File` serialization might be one of the more difficult topics to explore.

Regarding initial requirements from the discussion by @tgalopin 

>    the ability to store files in different places (locally, S3, CDN, ...) ;

Definitely doable by implementing service (temporary and final storage can differ).

>    the ability to validate files easily (covered by Validator in this PR, which is great) ;

As easy as possible - by defining set of validator constraints (that can easily come from attributes) and utilizing Validator exception

>    the ability to manipulate files before they are stored (resize images, encode them to JPG, decrease the quality, ... to save space) ;

This one needs some discussion as I have no clear vision here. We could go with another interface or focus on a single library (Glide or directly Intervention maybe?).

>    the ability to invalidate HTTP cache (ie. I want each file to have a unique name so that cache is automatically invalidated on upload, and I want to be able to store this name in my database) ;

I think temporary files names should be unique, but implementation may choose whatever solution suits them.

TODOs / problems
===============
- [ ] No clear vision for file manipulators (probably just image files)
- [ ] There might be an issue with using `File` class and custom storage. For example - with Flysystem extended `File` object needs reference to the `FilesystemOperator`, which makes it not serializable. We might need intermediate object and method to (de-)hydrate it.
- [ ] Is that all we need? :wink:

### DIFF
--- a/src/LiveComponent/src/UploaderInterface.php
+++ b/src/LiveComponent/src/UploaderInterface.php
@@ -1,0 +1,128 @@
+<?php
+declare(strict_types=1);
+
+namespace Symfony\UX\LiveComponent;
+
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\File\Exception\FileException;
+use Symfony\Component\HttpFoundation\File\File;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Exception\ValidationFailedException;
+
+/**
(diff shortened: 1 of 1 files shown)


Lustmored:
Closing in favor of #834

Issue #407, opened 2022-08-06
[Autocomplete] and [LiveComponent] don't play well together


janklan:
I noticed when the UX Autocompleter is part of the form and the form is rendered as a live component following https://symfony.com/bundles/ux-live-component/current/index.html#forms, the UX Autocompleter disappears once the component refreshes.

I made a minimal PoC site to demo the problem: https://github.com/janklan/symfony-ux-playground/tree/autocompleter-and-form-structure

This is what is happening:

![2022-08-06 14 42 41](https://user-images.githubusercontent.com/5463371/183235153-2442e148-04f2-486d-a502-6216aca4b520.gif)

maartendekeizer:
Hit the same bug today. It will also happens when you change another field, in your example the Name field for example.
It looks like the new autocomplete is found by Stimilus and connected to the correct element

![image](https://user-images.githubusercontent.com/1061334/183308080-8df5cbe7-4130-4404-99b7-cbee869ffd38.png)

janklan:
> Hit the same bug today. It will also happens when you change another field

Yes, any time the  live component re-renders, the autocomplete function is lost, no matter the cause for the re-render.

> It looks like the new autocomplete is found by Stimilus and connected to the correct element

Stimulus knows about everything, [but is told to do nothing](https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/src/controller.ts#L21-L25).

[I fixed it](https://github.com/symfony/ux/pull/409) and I have a gif to prove it!

![2022-08-08 07 37 05](https://user-images.githubusercontent.com/5463371/183313053-3a8da717-fed2-4bd3-b184-dbc89cddbe07.gif)

Note `author` is the original, and `author2` in an umapped field that uses my patched controller.

weaverryan:
Re-opening - this is kind of a duplicate of #354, but we can close when live works well with autocomplete

PR #409, opened 2022-08-07
Unset TomSelect after destroying the Stimulus Controller

janklan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | [407](https://github.com/symfony/ux/issues/407)
| License       | MIT

Sometimes Stimulus re-uses an instance of a controller after it's destroyed and re-created - for example when a Live Twig Component re-renders. Without unsetting the variable, [connect() method thinks TomSelect is already in place, and does nothing](https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/src/controller.ts#L21-L25).

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -41,6 +41,8 @@ export default class extends Controller {
 
     disconnect() {
         this.tomSelect.destroy();
+        // Fixes https://github.com/symfony/ux/issues/407
+        this.tomSelect = undefined;
     }
 
     #getCommonConfig(): Partial<TomSettings> {


src/Autocomplete/assets/src/controller.ts
```diff
@@ -41,6 +41,8 @@ export default class extends Controller {
 
     disconnect() {
         this.tomSelect.destroy();
+        // Fixes https://github.com/symfony/ux/issues/407
+        this.tomSelect = undefined;
```
weaverryan, line 45: Yup, good catch - that was an oversight.

weaverryan:
Thanks @janklan!

bendavies:
@weaverryan this failed the build since `yarn build` wasn't done?

PR #411, opened 2022-08-08
[LiveComponent] Add missing ComponentWithFormTrait in docs

janklan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

The Live Form is missing a trait that's needed for it to work.

PR #415, opened 2022-08-09
[LiveComponent] Add missing default action in docs

maartendekeizer:
Add missing default action in docs about FormCollection example. Without you got an error after copy pasting the example; resulting in bad DX.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Other option is to use the DefaultActionTrait but I think it is better to hint developers about the fact that you need to set `#[Security]`-attributes.

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -779,6 +779,14 @@ make it easy to deal with forms::
     class PostFormComponent extends AbstractController
     {
         use ComponentWithFormTrait;
+        
+        /**
+         * Default action. Should not return anything.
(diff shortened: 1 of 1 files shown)


weaverryan:
Thanks for this - definitely want to get this fixed.

But, we *should* use the `DefaultActionTrait`. If you want to add the `Security` attribute it should be done above your class - that way it'll apply to all actions (the default re-render action, but also any custom actions... though you could add additional Security attributes to those methods if you need even strong security for one action).

Could you refactor to use the trait?

weaverryan:
Whoops - nevermind - I just saw that #411 just fixed this also. But thanks for the PR!

PR #416, opened 2022-08-09
[Live] Support (de)hydrating nullable backed Enum's

### LINKED ISSUE #414: [LiveComponent] Incorrect handling of nullable enum types
I created a component with 2 writeable live props, both nullable and with a BackedEnumType.
When I changed a value a sync is happen as expected, but this will result in an error when one of the fields are empty

With request payload `{"data":1,"newPriority":null,"newStatus":"new","_checksum":"***"}` the following error occured
`The data is neither an integer nor a string, you should pass an integer or a string that can be parsed as an enumeration case of type App\Enum\Priority.`

And with request payload `{"data":1,"newPriority":"high","newStatus":"","_checksum":"***"}` the following error occured
`"" is not a valid backing value for enum "App\Enum\Priority"`

It think this is caused by the fact that the hydrator always try to denormalize and that empty strings are not converted to null values.
https://github.com/symfony/ux/blob/ebe425c03d03643ab76080ee2860e019443b4d6e/src/LiveComponent/src/LiveComponentHydrator.php#L170-L175

A suggestion for the fix
```
            if ($method = $liveProp->hydrateMethod()) {
                // TODO: Error checking
                $value = $component->$method($dehydratedValue);
            } elseif ($property->getType() instanceof \ReflectionNamedType && !$property->getType()->isBuiltin()) {
                if ($value === '' && $property->getType()->allowsNull() === true) {
                    // value is given as empty string, the property type is not a string (isBuiltin-method)
                    // so we can assume that value is meant to be empty
                    $value = null;
                }
                if ($value !== null || $property->getType()->allowsNull() === false) {
                    // if value is null but the property does not allow null run denormalization too
                    $value = $this->normalizer->denormalize($value, $property->getType()->getName(), 'json', [self::LIVE_CONTEXT => true]);
                }
            }
```

Component

```
<?php

declare(strict_types=1);

namespace App\Twig\Component;

use App\Enum\Priority;
use App\Enum\Status;


#[AsLiveComponent('prio', template: 'component/status.html.twig')]
class StatusComponent extends AbstractController
{
    #[LiveProp(writable: true)]
    public ?Priority $newPriority = null;

    #[LiveProp(writable: true)]
    public ?Status $newStatus = null;

    public function __invoke(): void
    {
    }
}
```

Template
```
    <div {{ attributes }}>
        Status: <input type="text" data-model="newStatus" value="{{ this.newStatus.value|default('') }}">
        Prio: <input type="text" data-model="newPriority" value="{{ this.newPriority.value|default('') }}">
        <br>
        {{ this.newStatus.value|default('') }} / {{ this.newPriority.value|default('') }}
    </div>
```

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #414
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -166,12 +166,15 @@ final class LiveComponentHydrator
             }
 
             $value = $dehydratedValue;
+            $type = $property->getType() instanceof \ReflectionNamedType ? $property->getType() : null;
 
             if ($method = $liveProp->hydrateMethod()) {
                 // TODO: Error checking
                 $value = $component->$method($dehydratedValue);
-            } elseif ($property->getType() instanceof \ReflectionNamedType && !$property->getType()->isBuiltin()) {
-                $value = $this->normalizer->denormalize($value, $property->getType()->getName(), 'json', [self::LIVE_CONTEXT => true]);
+            } elseif (!$value && $type && $type->allowsNull() && is_a($type->getName(), \BackedEnum::class, true) && !\in_array($value, array_map(fn (\BackedEnum $e) => $e->value, $type->getName()::cases()))) {
+                $value = null;
+            } elseif (null !== $value && $type && !$type->isBuiltin()) {
+                $value = $this->normalizer->denormalize($value, $type->getName(), 'json', [self::LIVE_CONTEXT => true]);
             }
 
             foreach ($liveProp->exposed() as $exposedProperty) {
--- a/src/LiveComponent/tests/Fixtures/Component/WithEnums.php
+++ b/src/LiveComponent/tests/Fixtures/Component/WithEnums.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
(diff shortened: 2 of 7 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -166,12 +166,15 @@ public function hydrate(object $component, array $data, string $componentName):
...
 
             $value = $dehydratedValue;
+            $type = $property->getType() instanceof \ReflectionNamedType ? $property->getType() : null;
 
             if ($method = $liveProp->hydrateMethod()) {
                 // TODO: Error checking
                 $value = $component->$method($dehydratedValue);
-            } elseif ($property->getType() instanceof \ReflectionNamedType && !$property->getType()->isBuiltin()) {
-                $value = $this->normalizer->denormalize($value, $property->getType()->getName(), 'json', [self::LIVE_CONTEXT => true]);
+            } elseif (!$value && $type && $type->allowsNull() && is_a($type->getName(), \BackedEnum::class, true) && !\in_array($value, array_map(fn (\BackedEnum $e) => $e->value, $type->getName()::cases()))) {
```
weaverryan, line 174: This feels a bit specific or heavy-handed, as we discussed. But I think we should do it... if nothing else to see if this fixes things or helps us discover more use-cases so we can get this just right.

weaverryan:
Thanks Kevin!

PR #421, opened 2022-08-11
[LiveComponent] Tweak live collection rendering

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes?
| Tickets       | -
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/Form/Type/LiveCollectionType.php
+++ b/src/LiveComponent/src/Form/Type/LiveCollectionType.php
@@ -43,8 +43,14 @@ final class LiveCollectionType extends AbstractType
     {
         if ($form->getConfig()->hasAttribute('button_add_prototype')) {
             $prototype = $form->getConfig()->getAttribute('button_add_prototype');
-            $view->vars['button_add_prototype'] = $prototype->setParent($form)->createView($view);
-            array_splice($view->vars['button_add_prototype']->vars['block_prefixes'], 1, 0, 'live_collection_button_add');
+            $view->vars['button_add'] = $prototype->setParent($form)->createView($view);
+
+            $attr = $view->vars['button_add']->vars['attr'];
+            $attr['data-action'] ??= 'live#action';
+            $attr['data-action-name'] ??= sprintf('addCollectionItem(name=%s)', $view->vars['full_name']);
+            $view->vars['button_add']->vars['attr'] = $attr;
+
+            array_splice($view->vars['button_add']->vars['block_prefixes'], 1, 0, 'live_collection_button_add');
         }
     }
 
@@ -75,8 +81,14 @@ final class LiveCollectionType extends AbstractType
             }
 
             foreach ($view as $k => $entryView) {
-                $entryView->vars['button_delete_prototype'] = $prototypes[$k]->createView($entryView);
-                array_splice($entryView->vars['button_delete_prototype']->vars['block_prefixes'], 1, 0, 'live_collection_button_delete');
+                $entryView->vars['button_delete'] = $prototypes[$k]->createView($entryView);
+
+                $attr = $entryView->vars['button_delete']->vars['attr'];
+                $attr['data-action'] ??= 'live#action';
+                $attr['data-action-name'] ??= sprintf('removeCollectionItem(name=%s, index=%s)', $view->vars['full_name'], $k);
+                $entryView->vars['button_delete']->vars['attr'] = $attr;
+
+                array_splice($entryView->vars['button_delete']->vars['block_prefixes'], 1, 0, 'live_collection_button_delete');
             }
         }
     }
@@ -92,6 +104,9 @@ final class LiveCollectionType extends AbstractType
             'button_add_options' => [],
             'button_delete_type' => ButtonType::class,
             'button_delete_options' => [],
+            'allow_add' => true,
+            'allow_delete' => true,
+            'by_reference' => false,
         ]);
     }
 
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -1265,7 +1265,7 @@ There is no need for a custom template just render the form as usual:
 
 The ``add`` and ``delete`` buttons are rendered as separate ``ButtonType`` form
 types and can be customized like a normal form type via the ``live_collection_button_add``
-and ``live_collection_button_delete`` respectively:
+and ``live_collection_button_delete`` block prefix respectively:
 
 .. code-block:: twig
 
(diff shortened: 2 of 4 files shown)


weaverryan (commented):
I like this a lot! Would you mind adding a small section to the docs with an example of overriding `live_collection_widget` (e.g. to render the add button first then the widget) and `live_collection_entry_row`. (e.g. to add a `<div>` around each row)?

1ed:
> I like this a lot! Would you mind adding a small section to the docs with an example of overriding `live_collection_widget` (e.g. to render the add button first then the widget) and `live_collection_entry_row`. (e.g. to add a `<div>` around each row)?

Thanks! I will.

src/LiveComponent/src/Resources/views/form_theme.html.twig
```diff
@@ -1,19 +1,13 @@
 {%- block live_collection_widget -%}
     {{ block('form_widget') }}
-    {%- if button_add_prototype is defined and not button_add_prototype.rendered -%}
-        {{ form_row(button_add_prototype, { attr: button_add_prototype.vars.attr|merge({
-            'data-action': 'live#action',
-            'data-action-name': 'addCollectionItem(name=' ~ form.vars.full_name ~ ')'
-        }) }) }}
+    {%- if skip_add_button|default(false) is same as false and button_add_prototype is defined and not button_add_prototype.rendered -%}
+        {{ form_row(button_add_prototype) }}
```
1ed: What do you think about changing the name of the exposed variable to `button_add` or maybe something more specific to the collection `live_collection_item_add` or just `item_add`?

`skip_add_button`, `button_add`, `button_delete`?

kbond: `add_button`, `delete_button`?

1ed: Actually I kind of like `button_add` and `button_delete` better, but not sure about it :smiley:

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+
+    <table class="table table-borderless form-no-mb">
+        <thead>
+            <tr>
+                <td>Item</td>
+                <td>Priority</td>
+            </tr>
+        </thead>
+        <tbody>
+            {{ form_row(form.todoItems, { skip_add_button: true }) }}
```
1ed, line 1391: Not sure about this. The same can be done by rendering the button into a variable before this line.

src/LiveComponent/src/Form/Type/LiveCollectionType.php
```diff
@@ -92,6 +104,9 @@ public function configureOptions(OptionsResolver $resolver): void
             'button_add_options' => [],
             'button_delete_type' => ButtonType::class,
             'button_delete_options' => [],
+            'allow_add' => true,
+            'allow_delete' => true,
+            'by_reference' => false,
```
1ed, line 109: Is it a good idea to change the defaults? I think it is more common this way.

src/LiveComponent/src/Resources/views/form_theme.html.twig
kbond: I'm fine with that.

src/LiveComponent/src/Form/Type/LiveCollectionType.php
weaverryan: I think that's fine. If you don't want these, then there isn't a point to using `LiveCollectionType`.

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+            'attr' => [
+                'class' => 'btn btn-outline-danger',
+            ],
+        ]
+    ])
+    // ...
+
+If you want more control over how each row is rendered you can override the blocks
+related to the ``LiveCollectionType``. This works the same way as for `the traditional
+collection type`_, but you should use ``live_collection`` and ``live_collection_entry`` as
```
weaverryan: Did you mean `live_collection_widget` here?

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+prefixes instead.
+
+Let's continue our previous example and customize the rendering of the blog post comments form.
```
weaverryan:
```suggestion
For example, et's continue our previous example and customize the rendering of the blog post comments form.
```

I'm adding this because, if I'm understand things correctly, the example below is intended to show how you could override the blocks that were mentioned above.

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+By default the add comment button is placed after the comments, let's move it before them.
+
+.. code-block:: twig
+
+    {%- block _blog_post_form_comments_widget -%}
```
weaverryan: Should this `live_collection_widget`? I mean, I know what you have is correct for this situation, but I think showing `live_collection_widget` would be better (it would be very rare someone would need to override `LiveCollectionType` twice in the same form in a different way).

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+            {{ block('form_row') }}
+            {%- if button_delete is defined and not button_delete.rendered -%}
+                {{ form_row(button_delete) }}
+            {%- endif -%}
+        </div>
+    {%- endblock -%}
+
+.. note::
+
+    The ``LiveCollectionType`` under the hood adds a ``button_add`` and
```
weaverryan:
```suggestion
    Under the hood, ``LiveCollectionType`` adds ``button_add`` and ``button_delete`` fields...
```

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+        {%- if skip_add_button|default(false) is same as false and button_add is defined and not button_add.rendered -%}
+            {{ form_widget(button_add, { label: '+ Add Item', class: 'btn btn-outline-primary' }) }}
+        {%- endif -%}
+    {%- endblock -%}
+
+    {%- block live_collection_entry_row -%}
+        <tr>
+            {% for child in form|filter(child => not child.rendered) %}
+                <td>{{- form_row(child, { label: false }) -}}</td>
+            {% endfor %}
```
weaverryan, line 1370: Is this a clever way to loop over all of the fields in the sub-form and render them? If so, what's the purpose of the `|filter(child => not child.rendered`?

src/LiveComponent/src/Resources/doc/index.rst
weaverryan: `skip_add_button` feels less hacky - so I like it 👍

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -1282,6 +1282,116 @@ and ``live_collection_button_delete`` respectively:
...
+            <td>
+                {{ form_row(button_delete, { label: 'X', attr: { class: 'btn btn-outline-danger' } }) }}
+            </td>
+        </tr>
+    {%- endblock -%}
+
+If you don't want to customize each row but want to put the add button to a different place,
+later in the template, the ``skip_add_button`` parameter can be used.
```
weaverryan: To render the add button later in the template, you can skip rendering it initially with ``skip_add_button``,
then render it manually after:

weaverryan (commented):
This is great! I've left some questions about the docs - my form theming is a bit fuzzy :)

src/LiveComponent/src/Resources/doc/index.rst
1ed: Yes, this just renders the children in a `<td>`. It's safer to use in this way, because if a field was rendered already, rendering it again will throw an exception. But in this case it's not really necessary, so I leave it out.

src/LiveComponent/src/Resources/doc/index.rst
1ed: No, I just meant to mention it's just a prefix, it can be `live_collection_widget`, `live_collection_row`, `live_collection_errors`, ... and linked the docs for the collection field (https://symfony.com/doc/current/form/form_themes.html#fragment-naming-for-collections). Maybe better to write `live_collection_*` instead.

src/LiveComponent/src/Resources/doc/index.rst
1ed: I meant this could be an example for customizing just for a specific form type, and the next with the bootstrap table could be a generic one. But it's fine by me to keep it simple and just use `live_collection_widget` here too.

weaverryan:
Thank you Gábor!

1ed:
Thank you!

PR #422, opened 2022-08-11
[Autocomplete] Only add the clear_button plugin to TomSelect if the form element is not required

### LINKED ISSUE #402: [Autocomplete] required choice types end up not being required.
Hi there,

There seems to be an issue with how tom-select is configured, for required, non-multiple choice types.

Given a form type configuration like

```php
$builder->add('isAttending', ChoiceType::class, [
    'choices' => [
        'Yes' => 'yes',
        'No' => 'no',
        'Maybe' => 'maybe',
    ],
]);
```
You will be left with a select box like:
image

All well and good, an option is mandatory.

However, convert to `autocomplete` and you're left with:
image

see the issue? It's clearable because the `clear plugin` [is added](https://github.com/symfony/ux/blob/8e84d49a92b70f99001b8817cd3129a6215175f6/src/Autocomplete/assets/src/controller.ts#L52), so you can be left with no value.

image

this shouldn't be possible, as it isn't with the simple select box above.

We can try fix fix:
```php
$builder->add('isAttending', ChoiceType::class, [
    'choices' => [
        'Yes' => 'yes',
        'No' => 'no',
        'Maybe' => 'maybe',
    ],
    'autocomplete' => true,
    'required' => true,
]);
```

this has no effect for 2 reasons:
1. the symfony/ux-autocomplete controller still adds the clear plugin (while the [easyadmin tom-select controller](https://github.com/EasyCorp/EasyAdminBundle/blob/579fcb45a3f98e66b2fcf40e01d54cd8ffe05104/assets/js/autocomplete.js#L36-L38) takes `requried` into considertion, from which symfony/ux-autocomplete was inspired by)
2. the easy admin version wouldn't work either, because [symfony removes the required attribute](https://github.com/symfony/symfony/blob/b1b77f366d42a58c1cd0879c605b997d05279d01/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L59-L61) for select boxes when it's impossible not to submit a value.

Number 2 can be "fixed"/overridden like this:
```php
$builder->add('isAttending', ChoiceType::class, [
    'choices' => [
        'Yes' => 'yes',
        'No' => 'no',
        'Maybe' => 'maybe',
    ],
    'attr' => [
        'required' => true,
    ],
    'autocomplete' => true,
]);
```

but that would mean we still have to fix the stimulus controller here to **not** add the clear button if there is a `required` attribute.

my "fix" above (adding `'attr' => ['required' => true]`) is probably too onerous to put on the end user. Maybe the autocomplete form type should do this automatically?

Thanks

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fixes #402
| License       | MIT

Only add the `clear_button` plugin to TomSelect if the form element is not required

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -120,7 +120,7 @@ export default class extends Controller {
 
         // multiple values excepted if this is NOT A select (i.e. input) or a multiple select
         const isMultiple = !this.selectElement || this.selectElement.multiple;
-        if (!this.formElement.disabled && !isMultiple) {
+        if (!this.formElement.required && !this.formElement.disabled && !isMultiple) {
             plugins.clear_button = { title: '' };
         }
 
--- a/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
+++ b/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
@@ -34,10 +34,21 @@ class AutocompleteFormRenderingTest extends KernelTestCase
             ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type?extra_options=')
             ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-min-characters-value', '2')
             ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-max-results-value', '25')
+            ->assertElementAttributeContains('#product_category', 'required', 'required')
+
+            ->assertElementAttributeContains('#product_ingredients', 'required', 'required')
 
             ->assertElementAttributeContains('#product_portionSize', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
+            ->assertElementAttributeContains('#product_portionSize', 'required', 'required')
+
+            ->assertElementAttributeContains('#product_colour', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
+            // not implemented in zenstruck/browser
+            // ->elementAttributeNotExists('#product_colour', 'required')
+            ->assertNotSeeElement('#product_colour[required]')
+
             ->assertElementAttributeContains('#product_tags', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
             ->assertElementAttributeContains('#product_tags', 'data-symfony--ux-autocomplete--autocomplete-tom-select-options-value', 'createOnBlur')
+            ->assertElementAttributeContains('#product_tags', 'required', 'required')
         ;
     }
 
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -219,7 +219,7 @@ class default_1 extends Controller {
 _default_1_instances = new WeakSet(), _default_1_getCommonConfig = function _default_1_getCommonConfig() {
     const plugins = {};
     const isMultiple = !this.selectElement || this.selectElement.multiple;
-    if (!this.formElement.disabled && !isMultiple) {
+    if (!this.formElement.required && !this.formElement.disabled && !isMultiple) {
         plugins.clear_button = { title: '' };
     }
     if (isMultiple) {
--- a/src/Autocomplete/tests/Fixtures/Form/ProductType.php
+++ b/src/Autocomplete/tests/Fixtures/Form/ProductType.php
@@ -46,6 +46,16 @@ class ProductType extends AbstractType
                 'autocomplete' => true,
                 'mapped' => false,
             ])
+            ->add('colour', ChoiceType::class, [
+                'choices' => [
+                    'red' => 'red',
+                    'blue' => 'blue',
+                    'green' => 'green',
+                ],
+                'autocomplete' => true,
+                'mapped' => false,
+                'required' => false,
+            ])
             ->add('tags', TextType::class, [
                 'mapped' => false,
                 'autocomplete' => true,


src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -47,6 +47,8 @@ public function finishView(FormView $view, FormInterface $form, array $options)
 
         $attr = $view->vars['attr'] ?? [];
 
+        $attr['required'] = $view->vars['required'];
```
bendavies: do we only want to do this for `ChoiceType` @weaverryan ?

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
```diff
@@ -31,11 +31,20 @@ public function testFieldsRenderWithStimulusController()
             ->get('/test-form')
             ->assertElementAttributeContains('#product_category_autocomplete', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
             ->assertElementAttributeContains('#product_category_autocomplete', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type')
+            ->assertElementAttributeContains('#product_category_autocomplete', 'required', 'required')
 
             ->assertElementAttributeContains('#product_portionSize', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
+            ->assertElementAttributeContains('#product_portionSize', 'required', 'required')
+
+            //not implemented in zenstruck/browser
+            //->elementAttributeNotExists('#product_colour', 'required')
```
bendavies, line 40: @kbond can i get at this mink assertion somehow?

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
```diff
@@ -31,10 +31,17 @@ public function testFieldsRenderWithStimulusController()
             ->get('/test-form')
             ->assertElementAttributeContains('#product_category_autocomplete', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
             ->assertElementAttributeContains('#product_category_autocomplete', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type')
+            ->assertElementAttributeContains('#product_category_autocomplete', 'required', 'required')
 
             ->assertElementAttributeContains('#product_portionSize', 'data-controller', 'symfony--ux-autocomplete--autocomplete')
+            ->assertElementAttributeContains('#product_portionSize', 'required', 'required')
+
+            // not implemented in zenstruck/browser
+            // ->elementAttributeNotExists('#product_colour', 'required')
```
bendavies, line 40: can i get at this mink assertion somehow?

bendavies, line 46: @kbond can I get at this assertion somehow?

src/Autocomplete/assets/src/controller.ts
```diff
@@ -50,7 +50,7 @@ export default class extends Controller {
 
         // multiple values excepted if this is NOT A select (i.e. input) or a multiple select
         const isMultiple = !this.selectElement || this.selectElement.multiple;
-        if (!this.formElement.disabled && !isMultiple) {
+        if (!this.formElement.required && !this.formElement.disabled && !isMultiple) {
```
bendavies, line 123: @weaverryan how do i update the dist js?

bendavies: `yarn build?` it seems this hasn't been for for a few PR's touching this file.
who should be doing this?

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
kbond: Try `->assertNotSeeElement('#product_colour[required]')`.

bendavies: works, thanks!

src/Autocomplete/assets/src/controller.ts
weaverryan: Yup, `yarn build`. Sometimes we forget to run it for a PR - we have a CI job that should fail if you forget, but at the moment, it's failing for a different reason #427

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
weaverryan: I think so yes. However, the `required` attribute is removed from the `select` for a reason in https://github.com/symfony/symfony/blob/b1b77f366d42a58c1cd0879c605b997d05279d01/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L59-L61 ... though I don't know what that reason is :). I'm wondering if we should avoid re-adding that attribute, and instead pass in a Stimulus variable.

src/Autocomplete/assets/src/controller.ts
bendavies: i'm on vacation for 2 weeks now. if you want to pick this up in my absense, feel free.

weaverryan: I was too - I'll wait for you to get back - I think this PR just needs minor tweaks

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
weaverryan: Ping on this! I think this is the only thing left - I'm thinking we should choose a variable other than `required`.

bendavies: @weaverryan my concern with this is that this will not work with bare (non symfony) forms, i.e. https://symfony.com/bundles/ux-autocomplete/current/index.html#manually-using-the-stimulus-controller

thoughts?

bendavies: @weaverryan see pending question above :)

weaverryan: I did some digging. The original behavior in https://github.com/symfony/symfony/blob/b1b77f366d42a58c1cd0879c605b997d05279d01/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L59-L61 comes from https://github.com/symfony/symfony/issues/8942 - it's an HTML5 spec thing. So I'd like to not add `required` (or, in the case of people manually creating `select` elements - force them to add `required`) in invalid situations.

Playing with the demo for this plugin - https://tom-select.js.org/plugins/clear-button/#clear-button - it feels like the invalid situation is when you have a `single` select and NO placeholder (defined as the first `option` having a `value=""`). If you DO have a placeholder, the "x" correctly resets things back to that value. But if you don't, that's when things get weird and it changes your select to an empty value... even though none originally existed.

I think this is what we should trigger on:

```
const isMultiple = !this.selectElement || this.selectElement.multiple;
const isSelectWithPlaceholder = this.selectElement && this.selectElement.options[0] && this.selectElement.options[0].value === '' ;
if (!this.formElement.disabled && !isMultiple && !isSelectWithPlaceholder) {
```

bendavies: @weaverryan just pushed this JS changed, but noticed it does not take `this.formElement.required` into account at all now.
Was this your intention?
long time since i looked at this so i'm a bit hazy...

bendavies: i think we still need to take `required` into account? added this. please take a look.

bendavies: > do we only want to do this for ChoiceType @weaverryan ?

this is done.

as discussed on slack, i reverted all the code related to placeholders.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -54,6 +54,10 @@ public function finishView(FormView $view, FormInterface $form, array $options):
 
         $attr = $view->vars['attr'] ?? [];
 
+        if ($form->getConfig()->getType()->getInnerType() instanceof ChoiceType) {
```
bendavies: hmm this does not work if a form type is `AsEntityAutocompleteField`

bendavies: `$form->getConfig()->getType()->getInnerType()->getParent() === BaseEntityAutocompleteType::class` works, but i'll just go with a negation

bendavies: actually i think this is not needed?
TextType can be marked as required too.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
bendavies: > do we only want to do this for ChoiceType @weaverryan ?

i don't think we need to do this? `TextType` can be required or not, too?

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -54,6 +54,8 @@ public function finishView(FormView $view, FormInterface $form, array $options):
 
         $attr = $view->vars['attr'] ?? [];
 
+        $attr['required'] = $view->vars['required'];
```
weaverryan: Do we still need this? What problem is this solving?

bendavies: tested without - i don't think we do.

bendavies: i see what it's for. it's to make my functional test work:
```php
->assertElementAttributeContains('#product_portionSize', 'required', 'required')
```

this won't work without my change as symfony does not always render the required form view variable for select boxes when it's impossible not to submit a value.

https://github.com/symfony/symfony/blob/b1b77f366d42a58c1cd0879c605b997d05279d01/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L59-L61

https://github.com/symfony/symfony/blob/b1b77f366d42a58c1cd0879c605b997d05279d01/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L459

the javascript will still work though.
up to you...?

weaverryan: My thinking is:

A) Symfony very purposely avoids the `required` attribute in illegal situations. It's super important to *not* change that
B) And so, if Symfony moves `required` for one situation, it just means that the clear button will show up in that one edge-case situation... and if I'm reading the code correctly, it already shows up in that situation today.

Bartheyrman22:
The validation is working correctly if you provide the following options:
```php
$builder
    ->add('testElement', ChoiceType::class, [
        'autocomplete' => true,
        'choices' => $choices,
        'required' => false,
        'attr' => [
            'required' => true,
        ],
    ])
```
Set  `required` to false and add a required-attribute instead.

Kocal:

Thanks!

Status: Needs Work

PR #426, opened 2022-08-12
[Vue] Introduce Vue UX component

t-richard:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       |  N/A 
| License       | MIT

This PR introduces a UX Vue component which allows to render Vue.js components inside Twig templates, with the ability to pass down props to Vue components from Twig. It is implemented in the same way as the UX React component for coherence.

1. Call `registerVueControllerComponents`  in `app.js` to register Vue components

```js
import { registerVueControllerComponents } from "@symfony/ux-vue";

registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/));
```

2. Create Vue "controller components" (eg. `Hello.vue`) inside the folder specified above (here `assets/vue/controllers`)
```vue
// assets/vue/controllers/Hello.vue
<template>
    <h1>Hello {{ name }}</h1>
</template>
```
3. All those components can now be used inside Twig using the helper function provided by the bundle, the second parameter are props that will be provided to the component
```twig
<div {{ vue_component('Hello', { name: 'Thibault' }) }}></div>
```

The implementation was **heavily** inspired by the work from @tgalopin in #329 (that's why I left him as author everywhere).

I'll gladly write docs and ux.symfony.com samples if accepted and once we agree on the API.

NB: Only Vue.js v3 is supported

### DIFF
--- a/src/Vue/README.md
+++ b/src/Vue/README.md
@@ -1,0 +1,14 @@
+# Symfony UX Vue.js
+
+Symfony UX Vue integrates [Vue.js](https://vuejs.org/) into Symfony applications.
+It provides tools to render Vue.js v3 components from Twig.
+
+**This repository is a READ-ONLY sub-tree split**. See
+https://github.com/symfony/ux to create issues or submit pull requests.
+
+## Resources
+
+-   [Documentation](https://symfony.com/bundles/ux-vue/current/index.html)
+-   [Report issues](https://github.com/symfony/ux/issues) and
+    [send Pull Requests](https://github.com/symfony/ux/pulls)
+    in the [main Symfony UX repository](https://github.com/symfony/ux)
--- a/src/Vue/Resources/assets/src/register_controller.ts
+++ b/src/Vue/Resources/assets/src/register_controller.ts
@@ -1,0 +1,30 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+export function registerVueControllerComponents(context: __WebpackModuleApi.RequireContext) {
+    const vueControllers: { [key: string]: object } = {};
+
+    const importAllVueComponents = (r: __WebpackModuleApi.RequireContext) => {
+        r.keys().forEach((key) => (vueControllers[key] = r(key).default));
+    };
+
+    importAllVueComponents(context);
+
+    // Expose a global Vue loader to allow rendering from the Stimulus controller
+    (window as any).resolveVueComponent = (name: string): object => {
+        const component = vueControllers[`./${name}.vue`];
+        if (typeof component === 'undefined') {
+            throw new Error(`Vue controller "${name}" does not exist`);
+        }
+
+        return component;
+    };
+}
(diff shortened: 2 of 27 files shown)


t-richard:
CI fail is unrelated, I fixed it in #427

src/Vue/README.md
```diff
@@ -0,0 +1,14 @@
+# Symfony UX Vue.js
+
+Symfony UX Vue integrates [Vue.js](https://vuejs.org/) into Symfony applications.
+It provides tools to render Vue.js components from Twig.
```
weaverryan: We can't do it until after the PR is merged (and the subtree split is setup), but we will also need to add a demo to ux.symfony.com and linking to it from here would be greta.

src/Vue/Resources/assets/src/register_controller.ts
```diff
@@ -0,0 +1,30 @@
...
+        r.keys().forEach((key) => (vueControllers[key] = r(key).default));
+    };
+
+    importAllVueComponents(context);
+
+    // Expose a global Vue loader to allow rendering from the Stimulus controller
+    (window as any).resolveVueComponent = (name: string): object => {
+        const component = vueControllers[`./${name}.vue`];
+        if (typeof component === 'undefined') {
+            throw new Error('Vue controller "' + name + '" does not exist');
```
weaverryan: Minor, but we can use "ticks" for this:

```js
throw new Error(`Vue controller "${name}" does not exist`);
```

weaverryan (commented):
This is wonderful work! Very minor comments

weaverryan (approved):

src/Vue/README.md
94noni: Passing by and reminded me one of my PR https://github.com/symfony/symfony-docs/pull/16403
 Could it be great to show case example of vuejs mounting with object serialized given to?
More a real ex as just passing a « name »

94noni: Is it worth it to write that it is v3+ ?

src/Vue/composer.json
```diff
@@ -0,0 +1,48 @@
+{
+    "name": "symfony/ux-vue",
+    "type": "symfony-bundle",
+    "description": "Integration of Vue.js in Symfony",
+    "keywords": [
+        "symfony-ux"
+    ],
+    "homepage": "https://symfony.com",
```
94noni, line 8: Can it be the ux.symfony.com ?
(I do not know the standards)

t-richard: This is the same for all UX components at the moment. Changing this would require to update all components and not sure it's wanted

src/Vue/README.md
t-richard: I'm already planning to add a ux.symfony.com example (maybe in a following PR ?).

I'll link it here when ready :+1:

t-richard: Will definitely mention it in the documentation, I'll see how I can phrase it to include it here too

tgalopin (approved):
It's a great PR, thanks so much!

src/Vue/README.md
94noni: Great 👍🏻

t-richard:
I've added docs and fixed the review comments.

I'll do the ux.symfony.com page in a separate PR.

CI is still failing due to the Chart.js error but built files are up to date for the Vue component.

tgalopin:
Thanks @t-richard for working on this feature, this is much appreciated.

PR #444, opened 2022-08-27
Add Vue app to the vue:mount event

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | N/A
| License       | MIT

Add the Vue app to the `vue:mount` event, so that users can have access to it when hooking into the event in order to use [Vue Plugins](https://vuejs.org/guide/reusability/plugins.html#introduction) with components. Without access to the app instance, it's not possible to use plugins.

### DIFF
--- a/src/Vue/Resources/assets/src/render_controller.ts
+++ b/src/Vue/Resources/assets/src/render_controller.ts
@@ -35,6 +35,13 @@ export default class extends Controller {
         if (this.element.__vue_app__ !== undefined) {
             this.element.__vue_app__.unmount();
         }
+        
+        this._dispatchEvent('vue:before-mount', {
+            componentName: this.componentValue,
+            component: component,
+            props: this.props,
+            app: this.app,
+        });
 
         this.app.mount(this.element);
 


weaverryan (approved):
Thanks @pierredup!

It's outside of the scope of this PR, but if you were able to add a small section to the docs (https://github.com/symfony/ux/blob/2.x/src/Vue/Resources/doc/index.rst) about the events, that'd be much appreciated. It's currently missing.

t-richard:
I thought about doing it in the intial PR but it requires special attention.

My fear was that people would use things like Vue router and that could get messy. 

Maybe a simple notice in the docs would be enough ?

weaverryan:
A simple notice so that users are aware of the events is all that's needed - would you be able to do that @t-richard?

pierredup:
Actually I think this patch won't work, because plugins needs to be added before mounting the app. So there would need to be another event between `createApp` and `mount`. Something like `vue:app`, or `vue:before-mount`?

> My fear was that people would use things like Vue router and that could get messy.

Vue router will work fine, as long as the hash history is used. For HTML5 history mode, the router might become a problem.

weaverryan:
WDYT @t-richard? This makes sense to me

weaverryan:
Thanks Pierre!

PR #447, opened 2022-09-01
[Autocomplete] Fix extending Tom Select when using custom Ajax class

ralfkuehnel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #...
| License       | MIT

I have tried to follow the [documentation](https://symfony.com/bundles/ux-autocomplete/current/index.html#extending-tom-select) to extend Tom Select using a custom Stimulus controller. It worked when updating the field configuration but did not when using a custom Ajax class. In this case the core Autocomplete Stimulus controller was missing in the rendered data-controller-Tag. Changing the autocomplete form theme to use the attributes from the autocomplete field seems to fix this!


weaverryan:
Thank you Ralf! Very nice test addition also :)

PR #454, opened 2022-09-08
Fix failing CI

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

- Latest version of php-cs-fixer adds a fixer that separates php docs. Do we want to keep the defaults or customize the config?
- js-dist-current [is failing](https://github.com/symfony/ux/runs/8251702511?check_suite_focus=true#step:6:1268). Not sure what's required to fix.


weaverryan (approved):
I'm fine with this change.

About js-dist, we're tracking that over on https://github.com/symfony/ux/pull/427 - but I need to follow-up

Issue #108, opened 2021-06-21
[TwigComponent] Improvements suggestion


ker0x:
First I want to thanks and congratulates @weaverryan and all people involved in making those 2 new components 👏!

I just start a fresh new project, using the Twig Component. and it's already saving me some extra work, this is really appreciate.

Here is a list of some suggestions that could improve this package:

#### Default path

Add a Bundle configuration parameter `default_path`, similar to the one for Twig, to define where components are stored.

```yaml
default_path: '%kernel.project_dir%/templates/components' # default value
```

#### Additional paths

Be able to defined multiple Component paths. This can be useful in applications where we want to split each parts:

```yaml
component_paths: 
 - '%kernel.project_dir%/templates/website/components'
 - '%kernel.project_dir%/templates/admin/components'
 - '%kernel.project_dir%/templates/common/components'
```

When rendering, it will parse each path to find the component template then fallback to `default_path` if none is found.

#### Prefix

It's a common practice to prefix "partial" twig template with `_` (e.g. `_form_delete.html.twig`). Being able to define a "global" prefix without having to specified it in the `getComponentName` method will be great. The prefix will be concatenate with the component name.

```yaml
component_prefix: '_' # optional, none by default
```

#### Alternate templates

Sometimes, depending on the context, it can be useful to display the component with a different template.

```php
public function getComponentTemplate(): string
{
    return match(this->context) {
        $this->foo() => 'bar',
        $this->bar() => 'foo',
    };
}
```

#### Maker command

I'm pretty sure this one is already in the pipe, but having a `make:component` command to quickly build a component will save some extra work!

WDYT ?

weaverryan:
Hey @ker0x!

Thanks for testing things and giving feedback. That is 💯!

I think what you're asking for is entirely sensible - it's just a matter of getting the details right.

We're using the main Twig service, so we need to use paths that are known to twig - e.g. the default path (`templates/`) or any Twig namespaces that you might have configured. So, in general, instead of `default_path: '%kernel.project_dir%/templates/components' ` it would be something more like default_path: 'components/' .

> Be able to defined multiple Component paths. This can be useful in applications where we want to split each parts

To *try* to play devil's advocate (i.e. find an existing solution that works), in #106, we will likely add a "template" option to the PHP 8 attribute - something like this:

```php
#[AsTwigComponent('alert', 'notifications/alert.html.twig')]
```

[...]

PR #460, opened 2022-09-12
[Twig] add a component name/template `:` "namespace" notation

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | #108
| License       | MIT

Proposes `:` as a namespace separator for component names. The default template replaces the `:` with `/` to create sub-directories. A component with the name `form:input`'s default template would be `templates/components/form/input.html.twig`.

Don't believe this would be a BC break. If someone is using the `:` notation already, they likely would have customized the template.


kbond:
Adjusted this PR to just document/test using `/` as the namespace separator instead of introducing a custom notation.

veewee:
> I've adjusted this PR to just document/test using / in the component's name as a namespace separator. After discussing with @weaverryan, we'd like avoid a custom notation for now.

That was the first thing I tried and it makes a lot of sense to use that as a separator - yet it doesn't work for live components.
Live components have their own URLs to refresh the content on the page. This URL doesn't allow the name of the live template to contain a `/`. This means that if you use / as the namespace separator, you won't be able to apply the same standard to live components. (Unless we make it possible in the router)

That is why I went with `:`.  Not saying it's the best, but the reasoning behind that separator could be:

* the old way twig did includes.
* in line with how you'dd namespace in html (xml) `<atom:button />`

kbond:
> yet it doesn't work for live components.

Ah good point!

> in line with how you'dd namespace in html (xml) <atom:button />

Yeah, we'd like to add an html parser in the future. Slashes wouldn't work with this.

kbond:
Ok, switched back to `:` for the reasons stated by @veewee.

weaverryan (approved):

weaverryan:
Thanks Kevin!

PR #466, opened 2022-09-19
[WIP] [Live] Make Ajax calls happen in serial, with "queued" changes

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Tickets       | None
| License       | MIT

Hi!

Consider the following situation:

A) Model update is made: Ajax call starts
B) An action is triggered BEFORE the Ajax call from (A) finishes.

Previously, we would start a 2nd Ajax call for (B) before the Ajax call from (A) finished... meaning two Ajax calls were happening at the same time. There are a few problems with this: (i) Ajax call (B) is missing any potential data changes from Ajax call (A) and (i) complexity of multiple things loading at once, and potentially finishing in a different order.

This PR simplifies things, which matches Livewire's behavior anyways. Now, the action from (B) will WAIT until the Ajax call from (A) finishes and THEN start. In fact, while an Ajax call is being made, all changes (potentially multiple model updates or actions) will be "queued" and then all set at once on the next request.

TODO:

* [X] Update the backend: for POST requests, the "data" was moved under a `data` key in JSON. Previously the entire body was the "data".
* [X] Update the backend: for POST/action requests, action "arguments" were moved from query parameters to an `args` key in the JSON.
* [x] Frontend: add tests for the `batch` action Ajax calls
* [x] Update the backend: a new fake `/batch` action needs to be added that can handle multiple actions at once
* [ ] A new `updatedModels` is sent on the ajax requests. If the signature fails, use this to give a better error message about what readonly properties were just modified. (in fact, this is the only purpose of sending this new field to the backend at this time).
* [X] ~~Pass a `includeUpdatedModels` value to the Stimulus controller ONLY when in `dev` mode.~~ For consistency, we will always pass the `updatedModels` in our Ajax requests, though this is intended to be "internal" and we won't use it other than to throw better errors.
* [X] ~~(Optional) If the backend has an unexpected exception (i.e. 5xx or maybe some 4xx where we mark that this is a problem we should show the developer), then render the entire HTML exception page (in dev mode only) so the user can see the error.~~ Done in #467 

Cheers!

### DIFF
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -445,7 +445,6 @@ export default class extends Controller implements LiveController {
 
                     url += `/${encodeURIComponent(actions[0].name)}`;
                 } else {
-                    // TODO: support on the server
                     url += '/_batch';
                     requestData.actions = actions;
                 }
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -47,28 +47,18 @@ export default class extends Controller implements LiveController {
     readonly debounceValue!: number;
     readonly hasDebounceValue: boolean;
 
+    backendRequest: BackendRequest|null;
     valueStore!: ValueStore;
 
-    /**
-     * The current "timeout" that's waiting before a model update
-     * triggers a re-render.
-     */
-    renderDebounceTimeout: number | null = null;
+    /** Actions that are waiting to be executed */
+    pendingActions: Array<{ name: string, args: Record<string, string> }> = [];
+    /** Has anything requested a re-render? */
+    isRerenderRequested = false;
 
     /**
-     * The current "timeout" that's waiting before an action should
-     * be taken.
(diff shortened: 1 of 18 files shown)


src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -391,90 +373,109 @@ export default class extends Controller implements LiveController {
...
-        if (!action) {
-            const dataJson = this.valueStore.asJson();
-            if (this._willDataFitInUrl(dataJson, params)) {
-                params.set('data', dataJson);
-                fetchOptions.method = 'GET';
-                dataAdded = true;
+                    url += `/${encodeURIComponent(actions[0].name)}`;
+                } else {
+                    // TODO: support on the server
+                    url += '/batch';
```
kbond:
```suggestion
                    url += '/_batch';
```

To avoid conflicts with a custom action.

weaverryan: yea I was thinking about that too 👍

PR #467, opened 2022-09-19
[Live] Rendering any errors in a simple modal

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | #102 Bugs G
| License       | MIT

Beautiful error handling! Inspired by Livewire:

Screen Shot 2022-09-19 at 4 10 35 PM

You can even `dd()` from your code and that will render in the modal:

Screen Shot 2022-09-19 at 5 10 29 PM

If an Ajax call returns anything OTHER than a rendered component (i.e. there was SOME sort of error), it's rendered in a modal. This is mostly for development errors. The modal will show on production also (and would show, for example, the 500 production page), but that should not happen in normal situations.

Cheers!


kbond (approved):
Looks great!

PR #468, opened 2022-09-20
Restricting chart.js to 3.8.2

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fixes tests
| License       | MIT

In 3.9, they broke typescript with chart.js/auto - see https://github.com/chartjs/Chart.js/issues/10599

It will be fixed in 4.0... that's the best we can do I think :).

I DID leave the peer dependencies alone, as I believe a user using 3.9 will still work.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,6 +33,8 @@ jobs:
             - run: yarn check-format
 
     js-dist-current:
+        # always skip check for now - building is too inefficient and large for CI
+        if: false
         runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


t-richard (approved):
Seems legit 👍

There's nothing more we can do until V4 I guess

weaverryan:
Unfortunately, the js-dist job needs to be skipped due to heap size problems. I'm not sure why - but out rollup build looks super inefficient (runs fine locally though)

PR #476, opened 2022-09-22
refactor(Autocomplete Doc): adding tip to modify the results number returned by a query on automatic endpoint

DarrHeLL:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | N/A 
| License       | MIT

Adding tip to to the doc to explain how can it be possible to modify the max results number when using autocomplete UX.

### DIFF
--- a/src/Autocomplete/src/Resources/doc/index.rst
+++ b/src/Autocomplete/src/Resources/doc/index.rst
@@ -258,7 +258,37 @@ to the options above, you can also pass:
             $qb->andWhere('entity.name LIKE :filter OR entity.description LIKE :filter')
                 ->setParameter('filter', '%'.$query.'%');
         }
+..tip::
 
+    By default, the query results number is limited to ten results.
+    In the case where you use the ``filter_query`` attribute, you can define the expected number results
+    by calling the setMaxResults() method of the query builder::
+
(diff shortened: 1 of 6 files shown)


src/Autocomplete/src/Resources/doc/index.rst
```diff
@@ -258,7 +258,37 @@ to the options above, you can also pass:
...
+    by calling the setMaxResults() method of the query builder::
+
+        'filter_query' => function(QueryBuilder $qb, string $query, EntityRepository $repository) {
+            if (!$query) {
+                return;
+            }
+
+            $qb->andWhere('entity.name LIKE :filter OR entity.description LIKE :filter')
+                ->setParameter('filter', '%'.$query.'%')
+                ->setMaxResults(100);
```
weaverryan: I like the idea of including this, but it's a huge block to communicate a relatively simple idea. WDYT about making this easier with a `max_results` option? If set, we call `setMaxResults()` for you.

DarrHeLL: To be honest I don't do that because i've no Idea how to implement this option. 

I will try

PR #478, opened 2022-09-23
[Autocomplete] implement max_results option

DarrHeLL:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       |N/A
| License       | MIT

Added max_results option in formType options to control the results number returned by a query on automatic endpoint.

### DIFF
--- a/src/Autocomplete/src/Resources/doc/index.rst
+++ b/src/Autocomplete/src/Resources/doc/index.rst
@@ -258,6 +258,8 @@ to the options above, you can also pass:
             $qb->andWhere('entity.name LIKE :filter OR entity.description LIKE :filter')
                 ->setParameter('filter', '%'.$query.'%');
         }
+``max_results`` (default: 10)
+    Allow you to control the max number of results returned by the automatic autocomplete endpoint.
 
 Using with a TextType Field
 ---------------------------
--- a/src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
+++ b/src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
@@ -67,6 +67,10 @@ final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
             $values['tom-select-options'] = json_encode($options['tom_select_options']);
         }
 
+        if ($options['max_results']) {
+            $values['max-results'] = $options['max_results'];
+        }
+
         $values['no-results-found-text'] = $this->trans($options['no_results_found_text']);
         $values['no-more-results-text'] = $this->trans($options['no_more_results_text']);
 
@@ -87,6 +91,7 @@ final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
             'allow_options_create' => false,
             'no_results_found_text' => 'No results found',
             'no_more_results_text' => 'No more results',
+            'max_results' => 10,
         ]);
 
         // if autocomplete_url is passed, then HTML options are already supported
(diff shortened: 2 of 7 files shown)


src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -39,11 +39,6 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
             $query
         );
 
-        // if no max is set, set one
-        if (!$queryBuilder->getMaxResults()) {
-            $queryBuilder->setMaxResults(10);
-        }
```
weaverryan: This should probably stay - for the (edge) case when someone is using this class but have their own custom `EntityAutocompleterInterface` implementation (one whose `createFilteredQueryBuilder` may not have the max results set).

This will make it so that setting to `null` does NOT avoid limiting results. But, I think that's ok - I'm not sure you should ever return ALL results - that's a recipe for hydrating too many entities.

src/Autocomplete/src/Resources/doc/index.rst
```diff
@@ -258,7 +258,9 @@ to the options above, you can also pass:
             $qb->andWhere('entity.name LIKE :filter OR entity.description LIKE :filter')
                 ->setParameter('filter', '%'.$query.'%');
         }
-
+``max_results`` (default: 10)
+    Allow you to control the reults number returned by a query on automatic endpoint.
+    Set it to ``null`` if you don't want to limit results.
```
weaverryan: blank line after this line

src/Autocomplete/src/Resources/doc/index.rst
```diff
@@ -258,7 +258,9 @@ to the options above, you can also pass:
             $qb->andWhere('entity.name LIKE :filter OR entity.description LIKE :filter')
                 ->setParameter('filter', '%'.$query.'%');
         }
-
+``max_results`` (default: 10)
+    Allow you to control the reults number returned by a query on automatic endpoint.
```
weaverryan:
```suggestion
    Allow you to control the max number of results returned by the automatic autocomplete endpoint.
```

weaverryan (commented):
Thanks for this! We could use a test - I'm thinking that `FieldAutocompleterTest` would be a good spot. We could add a new test method that creates, for example, 10 categories. Then in `CategoryAutocompleteType` we can pass `max_results => 5`. Then in the test, we do some search and assert that we only see 5 results, not the full 10.

DarrHeLL:
OK I will correct your suggestions and I tel you when it's done

src/Autocomplete/src/AutocompleteResultsExecutor.php
DarrHeLL: Done in [f46e34d](https://github.com/symfony/ux/pull/478/commits/f46e34de7ddcb485e0f6168c535a146514a90e88)

src/Autocomplete/src/Resources/doc/index.rst
DarrHeLL: Done in 40c7632

src/Autocomplete/src/Resources/doc/index.rst
DarrHeLL: Done in 40c7632

DarrHeLL:
Unit test added in commit [f34e8b0](https://github.com/symfony/ux/pull/478/commits/f34e8b0af6b0426ca92c465d7ada18f13f1daa0e)

DarrHeLL:
Outch don't know why CS is crying

weaverryan:
This is fantastic work! Thank you @DarrHeLL! I'll check out the CI - it's just a code styling issue, and it doesn't look like the CI outputs in a very helpful way.

PR #481, opened 2022-09-24
[Autocomplete]: implement group_by option on Entity Autocompleter results

### LINKED ISSUE #477: [Autocomplete] group_by option from EntityType does not work on Ajax auto-complete field
### Description 

EntityType's [group_by](https://symfony.com/doc/current/reference/forms/types/entity.html#group-by) option does not work on auto-complete field. The ajax return does not allow us to make a groupment we lose the `optgroup` information

Do you think that with a custom [tom-select extension](https://symfony.com/bundles/ux-autocomplete/current/index.html#extending-tom-select) and a custom AJAX autocomplete endpoint, could have the expected behavior?

In this case we would lose all the magic of the bundle ;(

jvancoillie:
| Q             | A
| ------------- | ---
| Bug fix?      | ?
| New feature?  | yes 
| Tickets       | Fix #477 
| License       | MIT

Add possibility to group results (with `group_by` option) on ajax autocomplete result

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -2,11 +2,15 @@
 
 ## 2.8.0
 
+
 -   The autocomplete will now update if any of the `option` elements inside of
     it change, including the empty / placeholder element. Additionally, if the
     `select` or `input` element's `disabled` attribute changes, the autocomplete
     instance will update accordingly. This makes Autocomplete work perfectly inside
     of a LiveComponent.
+
+-   Added support for using [OptionGroups](https://tom-select.js.org/examples/optgroups/).
+
 
 ## 2.7.0
 
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -1,7 +1,7 @@
 import { Controller } from '@hotwired/stimulus';
 import TomSelect from 'tom-select';
 import { TPluginHash } from 'tom-select/dist/types/contrib/microplugin';
-import { RecursivePartial, TomSettings, TomTemplates } from 'tom-select/dist/types/types';
+import { RecursivePartial, TomSettings, TomTemplates, TomLoadCallback } from 'tom-select/dist/types/types';
 
 export interface AutocompletePreConnectOptions {
     options: any;
@@ -147,20 +147,21 @@ export default class extends Controller {
             // VERY IMPORTANT: use 'function (query, callback) { ... }' instead of the
             // '(query, callback) => { ... }' syntax because, otherwise,
             // the 'this.XXX' calls inside this method fail
-            load: function (query: string, callback: (results?: any) => void) {
+            load: function (query: string, callback: TomLoadCallback) {
                 const url = this.getUrl(query);
                 fetch(url)
                     .then((response) => response.json())
                     // important: next_url must be set before invoking callback()
                     .then((json) => {
                         this.setNextUrl(query, json.next_page);
-                        callback(json.results);
+                        callback(json.results.options || json.results, json.results.optgroups || []);
                     })
-                    .catch(() => callback());
+                    .catch(() => callback([], []));
             },
             shouldLoad: function (query: string) {
                 return query.length >= minCharacterLength;
             },
+            optgroupField: 'group_by',
             // avoid extra filtering after results are returned
             score: function (search: string) {
                 return function (item: any) {
--- a/src/Autocomplete/doc/index.rst
+++ b/src/Autocomplete/doc/index.rst
@@ -513,6 +513,20 @@ a :ref:`custom autocompleter <custom-autocompleter>`:
             ]
         }
 
+    for using `Tom Select Option Group`_ the format is as follows
+
+    .. code-block:: json
+
+        {
+            "results": {
+                "options": [
+                    { "value": "1", "text": "Pizza", "group_by": ["food"] },
+                    { "value": "2", "text": "Banana", "group_by": ["food"] }
+                ],
+                "optgroups": [{ "value": "food", "label": "food" }]
+            }
+        }
+
     Once you have this, generate the URL to your controller and
     pass it to the ``url`` value of the ``stimulus_controller()`` Twig
     function, or to the ``autocomplete_url`` option of your form field.
@@ -557,3 +571,4 @@ the Symfony framework: https://symfony.com/doc/current/contributing/code/bc.html
 .. _`Tom Select Options`: https://tom-select.js.org/docs/#general-configuration
 .. _`controller.ts`: https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/src/controller.ts
 .. _`Tom Select Render Templates`: https://tom-select.js.org/docs/#render-templates
+.. _`Tom Select Option Group`: https://tom-select.js.org/examples/optgroups/
--- a/src/Autocomplete/src/AutocompleteResults.php
+++ b/src/Autocomplete/src/AutocompleteResults.php
@@ -19,6 +19,7 @@ final class AutocompleteResults
     public function __construct(
         public array $results,
         public bool $hasNextPage,
+        public array $optgroups = [],
     ) {
     }
 }
--- a/src/Autocomplete/src/AutocompleteResultsExecutor.php
+++ b/src/Autocomplete/src/AutocompleteResultsExecutor.php
@@ -12,6 +12,11 @@
 namespace Symfony\UX\Autocomplete;
 
 use Doctrine\ORM\Tools\Pagination\Paginator;
+use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
+use Symfony\Component\PropertyAccess\PropertyAccessor;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\PropertyAccess\PropertyPath;
+use Symfony\Component\PropertyAccess\PropertyPathInterface;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Security\Core\Security;
 use Symfony\UX\Autocomplete\Doctrine\DoctrineRegistryWrapper;
@@ -21,10 +26,22 @@ use Symfony\UX\Autocomplete\Doctrine\DoctrineRegistryWrapper;
  */
 final class AutocompleteResultsExecutor
 {
+    private PropertyAccessorInterface $propertyAccessor;
+    private ?Security $security;
+
     public function __construct(
         private DoctrineRegistryWrapper $managerRegistry,
-        private ?Security $security = null
+        $propertyAccessor,
+        /* Security $security = null */
     ) {
+        if ($propertyAccessor instanceof Security) {
+            trigger_deprecation('symfony/ux-autocomplete', '2.8.0', 'Passing a "%s" instance as the second argument of "%s()" is deprecated, pass a "%s" instance instead.', Security::class, __METHOD__, PropertyAccessorInterface::class);
+            $this->security = $propertyAccessor;
+            $this->propertyAccessor = new PropertyAccessor();
+        } else {
+            $this->propertyAccessor = $propertyAccessor;
+            $this->security = \func_num_args() >= 3 ? func_get_arg(2) : null;
+        }
     }
 
     public function fetchResults(EntityAutocompleterInterface $autocompleter, string $query, int $page): AutocompleteResults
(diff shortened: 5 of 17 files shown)


src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -46,14 +50,57 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
...
+        $options = [];
+        $optgroupLabels = [];
+
+        if (null !== $groupBy = $autocompleter->getFormOption('group_by')) {
+            if (\is_string($groupBy)) {
+                $groupBy = new PropertyPath($groupBy);
+            }
+
+            if ($groupBy instanceof PropertyPathInterface) {
+                $accessor = PropertyAccess::createPropertyAccessor();
```
weaverryan: Can we just inject the `property_path` service into this class instead of creating the new object?

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -46,14 +50,57 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
...
+                $groupBy = function ($choice) use ($accessor, $groupBy) {
+                    try {
+                        return $accessor->getValue($choice, $groupBy);
+                    } catch (UnexpectedTypeException) {
+                        return null;
+                    }
+                };
+            }
+
+            if (\is_callable($groupBy)) {
```
weaverryan: And what happens if `$groupBy` is not a callable? Should that trigger an error?

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -46,14 +50,57 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
...
+                        $optgroupLabels = array_merge($optgroupLabels, $groupLabels);
+                    }
+
+                    $options[] = $option;
+                }
+
+                return [
+                    'options' => $options,
+                    'optgroups' => array_map(fn (string $label) => ['value' => $label, 'label' => $label], array_unique($optgroupLabels)),
+                ];
```
weaverryan: The additions to this file as the riskiest part of this PR form a perspective if missing some edge cases and introducing bugs. Did you copy/adapt your code here form somewhere in core that handles the `group_by` option?

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -46,4 +46,9 @@ public function getValue(object $entity): mixed;
      * Note: if SecurityBundle is not installed, this will not be called.
      */
     public function isGranted(Security $security): bool;
+
+    /**
+     * Return Form option by name.
+     */
+    public function getFormOption(string $name): mixed;
```
weaverryan: I think the new method should be `getGroupBy()`. This interface is meant to be usable in custom classes that have nothing to do with a for.

src/Autocomplete/assets/src/controller.ts
```diff
@@ -134,14 +134,15 @@ export default class extends Controller {
...
             // '(query, callback) => { ... }' syntax because, otherwise,
             // the 'this.XXX' calls inside this method fail
-            load: function (query: string, callback: (results?: any) => void) {
+            load: function (query: string, callback: (options?: any, optgroups?: any) => void) {
                 const url = this.getUrl(query);
                 fetch(url)
                     .then(response => response.json())
                     // important: next_url must be set before invoking callback()
-                    .then(json => { this.setNextUrl(query, json.next_page); callback(json.results) })
+                    .then(json => { this.setNextUrl(query, json.next_page); callback(json.results.options || json.results, json.results.optgroups || []) })
```
weaverryan: Instead of `json.results.options || json.results` can we just now use `json.results.options`? Or do we need to look for `json.results` still for some situation?

Also, can you add an entry to the CHANGELOG about the Ajax return format changing?

weaverryan (commented):
Hi!

Really sorry for the slow review - this someone slipped past me! I've left some comments, but this is excellent work and I would love to incorporate it when you have some time to make these changes.

Thanks!

jvancoillie:
Hi @weaverryan 👋 

Thank you for this feedback, I will make the changes following your comments.

src/Autocomplete/src/AutocompleteResultsExecutor.php
jvancoillie: I assumed that if ever it is not callable then it falls back to the simple case without group_by but, indeed, we can trigger an error, which do you prefer?

src/Autocomplete/src/AutocompleteResultsExecutor.php
jvancoillie: Yes, I was inspired by what is done in the form component
https://github.com/symfony/form/blob/6.1/ChoiceList/Factory/PropertyAccessDecorator.php#L168
https://github.com/symfony/form/blob/6.1/ChoiceList/Factory/DefaultChoiceListFactory.php#L95

jvancoillie:
Oups!  Sorry for this, I rebased from 2.x :(

jvancoillie:
sorry for the small setback, I had some difficulties on the rebase of the branch 2.x :(

I had to do a force push because of my merge error.

jvancoillie:
Here are the changes after rebase and the pagination supports. 
There have been some changes, a review would be welcome. thanks

jvancoillie:
Hi @weaverryan,

what should we do with this pr ? I have however tried to be reactive because we are waiting for this feature.

norkunas:
Maybe if `group_by` is not configured, it should return the results in the old format? Because for me it would now be a BC break

weaverryan:
> Maybe if group_by is not configured, it should return the results in the old format? Because for me it would now be a BC break

I think that's an excellent idea. It would make the PR a lot safer to merge 👍

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -46,4 +46,9 @@ public function getValue(object $entity): mixed;
      * Note: if SecurityBundle is not installed, this will not be called.
      */
     public function isGranted(Security $security): bool;
+
+    /**
+     * Return group_by option.
+     */
+    public function getGroupBy(): mixed;
```
weaverryan, line 53: Instead of physically adding this method, which would break BC, add some documentation to the top for it: https://github.com/SymfonyCasts/reset-password-bundle/blob/6601f15fce7a4934bc9570f76feaf1b93536b1a7/src/ResetPasswordHelperInterface.php#L19 - that will allow people to be aware of it, and then we will add the actual method whenever we release a new major version.

weaverryan: Also is the return value really mixed? Or is it `?string`?

weaverryan (commented):
Sorry again for being late on this. Let us know if it is possible to use the "old format" if `group_by` is not set. 

Cheers!

src/Autocomplete/src/EntityAutocompleterInterface.php
jvancoillie: Yes i think, It can be string, callable or null

jvancoillie:
Hello, I made some changes to keep the old format if the group_by option is not set.

Would you have some time to give some feedback ?

src/Autocomplete/src/AutocompleteResults.php
```diff
@@ -14,11 +14,25 @@
...
         public bool $hasNextPage,
     ) {
     }
+
+    /**
+     * @return array{options: array{text: string, value: mixed, group_by?: string}, optgroups: array{label: string, value: string}}
+     */
+    public function getResults(): array
+    {
+        if (!empty($this->optgroups)) {
```
norkunas:
```suggestion
        if ($this->optgroups) {
```
Based on Symfony CS :)

src/Autocomplete/CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
 # CHANGELOG
 
+## 2.8.0
+
+-   Added support for using `Option Group`: the JSON format with option groupe takes two entries
```
weaverryan:
```suggestion
-   Added support for using [OptionGroups](https://tom-select.js.org/examples/optgroups/).
```

And then I don't think we need to show the JSON format below. 99% of people won't care about this - it's just handled for them :)

src/Autocomplete/assets/src/controller.ts
```diff
@@ -142,20 +142,21 @@ export default class extends Controller {
             // VERY IMPORTANT: use 'function (query, callback) { ... }' instead of the
             // '(query, callback) => { ... }' syntax because, otherwise,
             // the 'this.XXX' calls inside this method fail
-            load: function (query: string, callback: (results?: any) => void) {
+            load: function (query: string, callback: (options?: any, optgroups?: any) => void) {
```
weaverryan: After a little digging, it looks like this callback as a type:

```suggestion
            load: function (query: string, callback: TomLoadCallback) {
```

The `TomLoadCallback` can be imported from `'tom-select/dist/types/types'` like 3 other types already. Then, for the `.catch()` on line 154, it would become:

```
.catch(() => callback([], []));
```

to satisfy this type.

src/Autocomplete/doc/index.rst
```diff
@@ -501,10 +501,12 @@ a :ref:`custom autocompleter <custom-autocompleter>`:
...
-            "results": [
-                { "value": "1", "text": "Pizza" },
-                { "value": "2", "text":"Banana"}
-            ]
+            "results": {
+                "options": [
+                    { "value": "1", "text": "Pizza" },
+                    { "value": "2", "text":"Banana"}
+                ]
+            }
```
weaverryan, line 527: This change is not valid anymore, right?

And we need docs for the new option :)

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -51,14 +56,56 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
 
         $nbPages = (int) ceil($paginator->count() / $queryBuilder->getMaxResults());
 
-        $results = [];
-        foreach ($paginator as $entity) {
-            $results[] = [
-                'value' => $autocompleter->getValue($entity),
-                'text' => $autocompleter->getLabel($entity),
-            ];
```
weaverryan, line 59: Minor, but I think we should "exit early" first if there is no group by. Basically:

```php
if (null === $groupBy = $autocompleter->getGroupBy()) {
    // use the old logic
    $results = [];
    foreach ($paginator as $entity) {
        $results[] = [
            'value' => $autocompleter->getValue($entity),
            'text' => $autocompleter->getLabel($entity),
        ];
    }

    return $results;
}
```

Then the code that handles `$groupBy` can continue without needing to be in an else statement (no `else` will be needed at all).

src/Autocomplete/src/EntityAutocompleterInterface.php
weaverryan: Ah, so it is: https://github.com/symfony/symfony/blob/7ed43d1a749d95ae3f3d8ffa9b72858cac3e39fd/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php#L370

That is QUITE the long list of things :). I think we could probably just use `null|string|callable`. Then, if someone passes a `PropertyPath` or `GroupBy` in a form, inside of `WrappedEntityTypeAutocompleter`, we could "normalize" those into a `string|callable`.

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -51,14 +56,56 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
...
+                    try {
+                        return $accessor->getValue($choice, $groupBy);
+                    } catch (UnexpectedTypeException) {
+                        return null;
+                    }
+                };
+            }
+        }
+
+        if (\is_callable($groupBy)) {
```
weaverryan: If it's not a callable at this point, that would be an error, right? I'm imagining that we check for `null` and `string` above (and for string, you're converting into a callable). So we could say:

```php
if (!\is_callable($groupBy)) {
    // then throw an exception
}
```

src/Autocomplete/src/AutocompleteResults.php
```diff
@@ -14,11 +14,25 @@
...
     /**
-     * @param list<array{label: string, value: mixed}> $results
+     * @param list<array{text: string, value: mixed, group_by?: string}> $options
+     * @param list<array{label: string, value: string}>                  $optgroups
      */
     public function __construct(
-        public array $results,
+        public array $options,
+        public array $optgroups,
         public bool $hasNextPage,
```
weaverryan, line 21: We can't break BC here, even though it's unlikely to cause problems. So:

A) Rename `$options` back to `$results`
B) Move `$optgroups` to the final argument and make it optional

Then, let's move the `getResults()` logic into the controller. That'll keep this class dead-simple - just a carrier of data.

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -23,7 +27,8 @@ final class AutocompleteResultsExecutor
 {
     public function __construct(
         private DoctrineRegistryWrapper $managerRegistry,
-        private ?Security $security = null
+        private PropertyAccessorInterface $propertyAccessor,
```
weaverryan: To protect BC and trigger a deprecation, we can make the top of the class look something like this:

```php
final class AutocompleteResultsExecutor
{
    private PropertyAccessorInterface $propertyAccessor;
    private ?Security $security;

    public function __construct(
        private DoctrineRegistryWrapper $managerRegistry,
        $propertyAccessor,
        /* Security $security = null */
    ) {
        if ($propertyAccessor instanceof Security) {
            trigger_deprecation('symfony/ux-autocomplete', '2.8.0', 'Passing a "%s" instance as the second argument of "%s()" is deprecated, pass a "%s" instance instead.', Security::class, __METHOD__, PropertyAccessorInterface::class);
            $this->security = $propertyAccessor;
            $this->propertyAccessor = new PropertyAccessor();
        } else {
            $this->propertyAccessor = $propertyAccessor;
            $this->security = func_num_args() >= 3 ? func_get_arg(2) : null;
        }
    }
```

weaverryan (commented):
This is VERY close now - let's push it across the finish line!

src/Autocomplete/src/EntityAutocompleterInterface.php
jvancoillie: I'm struggling to see how to normalize this, how do you see the implementation of this normalization

weaverryan:
Thank you @jvancoillie for this wonderful feature and your excellent work. Apologies for the slow merge on my end - but I'm super excited to happen (I just tried it locally and it was super fun)

PR #482, opened 2022-09-25
[Vue] Add support for lazy-loading with Async Components

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | Fix #... 
| License       | MIT

I was a bit surprised when trying the Symfony UX' Vue integration, to see all Vue components/controllers to be loaded to render at least only one controllers.
This PR aims to reduce the bundle size and only load the required code and Vue controllers to render on the page.

To enable lazy-loading, add the 4th parameter `"lazy"` to `require.context()`:  
```diff
- registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/));
+ registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/, "lazy"));
```
_Note: the sync-loading still works_ 

# _Real life_ example

Given 6 really simple Vue controllers which all look the same:
image

If I render only two of them on a single page:
image

## Before

Capture d’écran 2022-09-25 à 11 03 16

You can see the `app.js` is about 4.8 kB, it contains all of the 6 Vue controllers.

## After

Capture d’écran 2022-09-25 à 11 03 38

You can see the `app.js` is now about 3.4 kB, and two more files have been lazy-loaded (our two Vue controllers rendered on the page).

On a real project, with a lot of Vue controllers, the difference would be very significative. I will try to post some results when migrating our main project at work.

--- 

Note: ~I still have to add tests and update the documentation.~ done

### DIFF
--- a/src/Vue/Resources/assets/dist/register_controller.js
+++ b/src/Vue/Resources/assets/dist/register_controller.js
@@ -1,15 +1,39 @@
+import { defineAsyncComponent } from 'vue';
+
 function registerVueControllerComponents(context) {
-    const vueControllers = {};
-    const importAllVueComponents = (r) => {
-        r.keys().forEach((key) => (vueControllers[key] = r(key).default));
-    };
-    importAllVueComponents(context);
-    window.resolveVueComponent = (name) => {
-        const component = vueControllers[`./${name}.vue`];
-        if (typeof component === 'undefined') {
-            throw new Error(`Vue controller "${name}" does not exist`);
+    const vueControllers = context.keys().reduce((acc, key) => {
+        acc[key] = undefined;
+        return acc;
+    }, {});
+    function loadComponent(name) {
+        const componentPath = `./${name}.vue`;
+        if (componentPath in vueControllers && typeof vueControllers[componentPath] === 'undefined') {
+            const module = context(componentPath);
+            if (module.default) {
+                vueControllers[componentPath] = module.default;
+            }
+            else if (module instanceof Promise) {
+                vueControllers[componentPath] = defineAsyncComponent(() => new Promise((resolve, reject) => {
+                    module
+                        .then((resolvedModule) => {
+                        if (resolvedModule.default) {
+                            resolve(resolvedModule.default);
+                        }
+                        else {
+                            reject(new Error(`Cannot find default export in async Vue controller "${name}".`));
+                        }
+                    })
+                        .catch(reject);
+                }));
+            }
+            else {
+                throw new Error(`Vue controller "${name}" does not exist.`);
+            }
         }
-        return component;
+        return vueControllers[componentPath];
+    }
+    window.resolveVueComponent = (name) => {
+        return loadComponent(name);
     };
 }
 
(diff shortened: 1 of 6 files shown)


src/Vue/Resources/assets/dist/register_controller.js
```diff
@@ -1,15 +1,39 @@
...
+                vueControllers[componentPath] = defineAsyncComponent(() => new Promise((resolve, reject) => {
+                    module
+                        .then((resolvedModule) => {
+                        if (resolvedModule.default) {
+                            resolve(resolvedModule.default);
+                        }
+                        else {
+                            reject(new Error(`Cannot find default export in async Vue controller "${name}".`));
+                        }
+                    })
```
Geekimo, line 25: There may be an indentation issue there.

Kocal: Maybe, but this is what Prettier outputs.

weaverryan: this is the `dist` file - so we don't care what it looks like :)

src/Vue/Resources/doc/index.rst
```diff
@@ -47,6 +47,10 @@ You also need to add the following lines at the end to your ``assets/app.js`` fi
     // they are not necessary.
     registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/));
 
+    // If you prefer to lazy-load your Vue.js controller components, in order to reduce to keep the JavaScript bundle the smallest as possible,
+    // and improve performances, you can use the following line instead:
+    registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/, 'lazy'));
```
weaverryan:
```suggestion
    //registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/, 'lazy'));
```

weaverryan (commented):
This is awesome! I've never used `defineAsycComponent` and `require.context` is always complex :). But you've got the tests, I love it.

Do you have anything else you need to do?

Kocal:
Thanks for the review @weaverryan!

On this PR, I think nothing is missing, however I thought about a usecase for a 2nd PR: I think some people would like to have both lazy-loaded and not lazy-loaded controllers/components at the same time.
E.g.: common components like a `NavBar` or `Aside` that would be **always** shown on every pages, and specific components like a date picker that would be sometimes shown in some context.

Currently, that's not possible to have both `registerVueControllerComponents` calls like this:
```js
registerVueControllerComponents(require.context('./vue/controllers', true, /\.vue$/));
registerVueControllerComponents(require.context('./vue/controllers-lazy', true, /\.vue$/, 'lazy'));
```

because `vueControllers` and `loadComponent`, which are used by the global function `resolveVueComponent()`, will be erased.

weaverryan:
@Kocal I think you're right - but I think that will be solved by #461

weaverryan (approved):

weaverryan:
Thanks Hugo!

PR #492, opened 2022-10-07
[Autocomplete] add option to specify minimum characters for autocomplete

### LINKED ISSUE #432: [Autocomplete] Add a way to require a minimum char number to before send ajax request

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes 
| Tickets       | Fix #432 
| License       | MIT

This change adds a new option `min_characters` for specifying the value of entered characters before the autocomplete is triggerd. This avoids calls with an empty query.

Todos:
- [x] add tests

@weaverryan I also added a `CHANGELOG.md` because this file wasn't present.

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -8,6 +8,7 @@ export default class extends Controller {
         optionsAsHtml: Boolean,
         noResultsFoundText: String,
         noMoreResultsText: String,
+        minCharacters: Number,
         tomSelectOptions: Object,
     }
 
@@ -15,6 +16,7 @@ export default class extends Controller {
     readonly optionsAsHtmlValue: boolean;
     readonly noMoreResultsTextValue: string;
     readonly noResultsFoundTextValue: string;
+    readonly minCharactersValue: number;
     readonly tomSelectOptionsValue: object;
     tomSelect: TomSelect;
 
@@ -30,7 +32,7 @@ export default class extends Controller {
 
     connect() {
         if (this.urlValue) {
-            this.tomSelect = this.#createAutocompleteWithRemoteData(this.urlValue);
+            this.tomSelect = this.#createAutocompleteWithRemoteData(this.urlValue, this.minCharactersValue);
 
             return;
         }
@@ -124,7 +126,7 @@ export default class extends Controller {
         return this.#createTomSelect(config);
     }
 
-    #createAutocompleteWithRemoteData(autocompleteEndpointUrl: string): TomSelect {
+    #createAutocompleteWithRemoteData(autocompleteEndpointUrl: string, minCharacterLength: number): TomSelect {
         const config: Partial<TomSettings> = this.#mergeObjects(this.#getCommonConfig(), {
             firstUrl: (query: string) => {
                 const separator = autocompleteEndpointUrl.includes('?') ? '&' : '?';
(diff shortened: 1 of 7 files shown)


src/Autocomplete/assets/src/controller.ts
```diff
@@ -30,7 +32,7 @@ export default class extends Controller {
 
     connect() {
         if (this.urlValue) {
-            this.tomSelect = this.#createAutocompleteWithRemoteData(this.urlValue);
+            this.tomSelect = this.#createAutocompleteWithRemoteData(this.urlValue, this.minCharactersTextValue);
```
weaverryan:
```suggestion
            this.tomSelect = this.#createAutocompleteWithRemoteData(this.urlValue, this.minCharactersValue);
```

Right?

src/Autocomplete/assets/src/controller.ts
```diff
@@ -8,13 +8,15 @@ export default class extends Controller {
         optionsAsHtml: Boolean,
         noResultsFoundText: String,
         noMoreResultsText: String,
+        minCharacters: String,
```
weaverryan: Should be a `Number` type, right?

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -91,6 +95,7 @@ public function configureOptions(OptionsResolver $resolver)
             'allow_options_create' => false,
             'no_results_found_text' => 'No results found',
             'no_more_results_text' => 'No more results',
+            'min_characters' => 3,
             'max_results' => 10,
         ]);
```
weaverryan, line 100: To test this, you could:

A) Pass the `min_character` option to one of the fields here https://github.com/symfony/ux/blob/2.x/src/Autocomplete/tests/Fixtures/Form/ProductType.php

B) Assert that the proper value attribute is rendered here: https://github.com/symfony/ux/blob/bb10dcff9dc9dc06fb023cf046bfcf9cc14beb5c/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php#L28-L40

src/Autocomplete/assets/test/controller.test.ts
```diff
@@ -81,6 +81,7 @@ describe('AutocompleteController', () => {
                 data-testid="main-element"
                 data-controller="check autocomplete"
                 data-autocomplete-url-value="/path/to/autocomplete"
+                data-autocomplete-min-characters-value="3"
```
weaverryan: To test this, I would remove this attribute from this test. Then create a new test - e.g. `it limits updates when min-characters` is passed. Inside, you would duplicate this test, but a bit simpler:

A) Set the min-characters to 3
B) "type" only 2 characters into the box
C) `expect(container.querySelectorAll('.option[data-selectable]')).toHaveLength(0);` - make sure 0 options are still shown.

No need to mess with `fetchMock` at all, though if an Ajax request WERE made by mistake, `fetchMock` will cause the test to fail since we were not expecting any Ajax calls. In other words, don't do anything with `fetchMock`, but `fetchMock` will "have your back" in case an accidental Ajax request were made.

weaverryan (commented):
Some notes - but looking good!

daFish:
@weaverryan Should all be fine now and all checks are happy.

weaverryan:
Thank you Marcus!

PR #494, opened 2022-10-07
[Autocomplete] Fix loading autocomplete choices with ID value objects

### LINKED ISSUE #493: [Autocomplete] Fails to load provided options if `multiple=true` on submit
The problem is in these lines:
https://github.com/symfony/ux/blob/7a2eda6af227a65cc1912e25b5ab98c368dc78ce/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php#L56-L58

~~Doctrine entity repository `findBy` method doesn't support finding entities by array so it always returns empty array.~~
Okay it supports, but only when using plain ID's, but doesn't support when using ID value objects like Symfony uid.

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | Fix #493
| License       | MIT

### DIFF
--- a/src/Autocomplete/composer.json
+++ b/src/Autocomplete/composer.json
@@ -42,6 +42,7 @@
         "symfony/security-bundle": "^5.4|^6.0",
         "symfony/security-csrf": "^5.4|^6.0",
         "symfony/twig-bundle": "^5.4|^6.0",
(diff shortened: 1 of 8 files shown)


weaverryan:
It adds some complexity, but it makes sense. Excellent work here - thanks Tomas!

PR #502, opened 2022-10-13
[Autocomplete] Add pagination

### LINKED ISSUE #399: [Autocomplete] Support for Virtual Scrolling/Pagination
Hi,

Support for [Virtual Scroll](https://tom-select.js.org/plugins/virtual_scroll/) is included in the javascript:
https://github.com/symfony/ux/blob/8e84d49a92b70f99001b8817cd3129a6215175f6/src/Autocomplete/assets/src/controller.ts#L131

But there is no support in the default php implmentation.
Basically, the backend needs to support pagintion fo the data, returning a `next_page` item in the response, if there are more results.

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #399
| License       | MIT

Adds pagination as discussed in #399.

- `AutocompleteResultsExecutorTest::testItExecutesTheResults` is broken now because mocked `AbstractQuery` doesn't have `setMaxResults` and `setFirstResult` methods and also because we use now Doctrine Paginator it's basically no point to unit test this.
- > (we could do this by cloning the current Request object, calling $request->query->set('page', $nextPage), then calling $request->getRequestUri()

  this was not possible so because request uri is already prepared at that point so calling `$request->query->set('page')` was not changing the uri for this reason I've used `UrlGenerator` with `_route`, `_route_params` and `$request->query->all()` and merged `

- Added some tests in `WiringTest` but not sure if this is the right place.
- Wanted to do the functional test which tries to scroll the options dropdown, but I don't know if this possible, because didn't find a way :)

### DIFF
--- a/src/Autocomplete/src/Controller/EntityAutocompleteController.php
+++ b/src/Autocomplete/src/Controller/EntityAutocompleteController.php
@@ -15,6 +15,7 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\Autocomplete\AutocompleteResultsExecutor;
 use Symfony\UX\Autocomplete\AutocompleterRegistry;
 
@@ -27,7 +28,8 @@ final class EntityAutocompleteController
 {
     public function __construct(
         private AutocompleterRegistry $autocompleteFieldRegistry,
-        private AutocompleteResultsExecutor $autocompleteResultsExecutor
+        private AutocompleteResultsExecutor $autocompleteResultsExecutor,
+        private UrlGeneratorInterface $urlGenerator,
     ) {
     }
 
@@ -38,10 +40,20 @@ final class EntityAutocompleteController
             throw new NotFoundHttpException(sprintf('No autocompleter found for "%s". Available autocompleters are: (%s)', $alias, implode(', ', $this->autocompleteFieldRegistry->getAutocompleterNames())));
         }
 
-        $results = $this->autocompleteResultsExecutor->fetchResults($autocompleter, $request->query->get('query', ''));
+        $page = $request->query->getInt('page', 1);
+        $nextPage = null;
+
+        $data = $this->autocompleteResultsExecutor->fetchResults($autocompleter, $request->query->get('query', ''), $page);
+
+        if ($data->hasNextPage) {
+            $parameters = array_merge($request->attributes->all('_route_params'), $request->query->all(), ['page' => $page + 1]);
+
+            $nextPage = $this->urlGenerator->generate($request->attributes->get('_route'), $parameters, UrlGeneratorInterface::ABSOLUTE_URL);
+        }
 
         return new JsonResponse([
-            'results' => $results,
+            'results' => $data->results,
+            'next_page' => $nextPage,
         ]);
     }
 }
--- a/src/Autocomplete/tests/Unit/AutocompleteResultsExecutorTest.php
+++ b/src/Autocomplete/tests/Unit/AutocompleteResultsExecutorTest.php
@@ -11,63 +11,15 @@
 
 namespace Symfony\UX\Autocomplete\Tests\Unit;
 
-use Doctrine\ORM\AbstractQuery;
-use Doctrine\ORM\EntityRepository;
-use Doctrine\ORM\QueryBuilder;
 use PHPUnit\Framework\TestCase;
-use Symfony\Bundle\SecurityBundle\Security\Security;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Security\Core\Security;
 use Symfony\UX\Autocomplete\AutocompleteResultsExecutor;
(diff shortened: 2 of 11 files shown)


norkunas:
Hm, CI doesn't test autocomplete at all :scream:

weaverryan:
> Hm, CI doesn't test autocomplete at all 😱

🤦  I have an old branch locally that dynamically creates CI jobs by scanning all of the component directories, instead of relying on remember to add the hardcoded config to the CI. That needs to get finished

src/Autocomplete/tests/Unit/AutocompleteResultsExecutorTest.php
```diff
@@ -65,7 +66,7 @@ public function testItExecutesTheResults()
         $this->assertEquals([
             ['value' => 1, 'text' => 'Result 1'],
             ['value' => 2, 'text' => 'Result 2'],
-        ], $executor->fetchResults($autocompleter, 'foo'));
+        ], $executor->fetchResults($autocompleter, 'foo', 1));
```
weaverryan: Ok, we can remove this test.

weaverryan:
> Wanted to do the functional test which tries to scroll the options dropdown, but I don't know if this possible, because didn't find a way :)

Hmm, yea, this would test the JavaScript itself to make sure that it, on scroll down, makes the correct request to the next page. That *would* be nice to have. The test would live here - https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/test/controller.test.ts - what I would attempt to do is:

A) Type in the box, wait for the results to appear
B) Then use [type](https://testing-library.com/docs/ecosystem-user-event/#typeelement-text-options) with `{arrowdown}` several times to "move down to the bottom of the list". Then, we wait to see if the second Ajax request is made.

Let me know if that works or if I can help with that part.

src/Autocomplete/tests/Unit/AutocompleteResultsExecutorTest.php
```diff
@@ -15,8 +15,9 @@
 use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\QueryBuilder;
 use PHPUnit\Framework\TestCase;
-use Symfony\Bundle\SecurityBundle\Security\Security;
+//use Symfony\Bundle\SecurityBundle\Security;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Security\Core\Security;
```
weaverryan, line 16: What's going on here? Playing with making things work on multiple Symfony versions?

src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -38,10 +40,20 @@ public function __invoke(string $alias, Request $request): Response
...
-        $results = $this->autocompleteResultsExecutor->fetchResults($autocompleter, $request->query->get('query', ''));
+        $page = $request->query->getInt('page', 1);
+        $nextPage = null;
+
+        $data = $this->autocompleteResultsExecutor->fetchResults($autocompleter, $request->query->get('query', ''), $page);
+
+        if ($data->hasNextPage) {
+            $parameters = array_merge($request->attributes->all('_route_params'), $request->query->all(), ['page' => $page + 1]);
+
+            $nextPage = $this->urlGenerator->generate($request->attributes->get('_route'), $parameters, UrlGeneratorInterface::ABSOLUTE_URL);
```
weaverryan, line 51: clever, clever

weaverryan (commented):
This looks fantastic! Awesome work!

src/Autocomplete/tests/Unit/AutocompleteResultsExecutorTest.php
norkunas: yes, the namespace has changed on newer versions..

norkunas:
> > Wanted to do the functional test which tries to scroll the options dropdown, but I don't know if this possible, because didn't find a way :)
> 
> Hmm, yea, this would test the JavaScript itself to make sure that it, on scroll down, makes the correct request to the next page. That _would_ be nice to have. The test would live here - https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/test/controller.test.ts - what I would attempt to do is:
> 
> A) Type in the box, wait for the results to appear B) Then use [type](https://testing-library.com/docs/ecosystem-user-event/#typeelement-text-options) with `{arrowdown}` several times to "move down to the bottom of the list". Then, we wait to see if the second Ajax request is made.
> 
> Let me know if that works or if I can help with that part.

Okay, so probably I'll need help with this. I've added the code that I've tried for the test but it does not work as expected:
 - on focusing input for some reason the "loading more results" is immediately added to the options dropdown;
 - triggering arrow down (tried to write trigger statement even more than 10 times) does not trigger the ajax request to load next page.

weaverryan:
Got the test passing. You did the hard work, but I gave you some bad advice! Apparently TomSelect entirely uses the `scroll` event and "scroll position" to determine when the next page should be loaded. So a little extra work was needed to fake / short-circuit some of that. But then, the test fell into place nicely with all the pieces you had :).

norkunas:
> Got the test passing. You did the hard work, but I gave you some bad advice! Apparently TomSelect entirely uses the `scroll` event and "scroll position" to determine when the next page should be loaded. So a little extra work was needed to fake / short-circuit some of that. But then, the test fell into place nicely with all the pieces you had :).

Nice, thanks :)

weaverryan:
Thank you Tomas!

PR #504, opened 2022-10-13
Update Vue docs

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | N/A
| License       | MIT

* Fix typo for event (`vue:mount` instead of `vue:before-mount`)
* Add note about history mode when using Vue router

### DIFF
--- a/src/Vue/Resources/doc/index.rst
+++ b/src/Vue/Resources/doc/index.rst
@@ -105,6 +105,7 @@ The event ``vue:before-mount`` is called before a component is mounted on the pa
 
         // Example with Vue Router
         const router = VueRouter.createRouter({
+            history: VueRouter.createWebHashHistory(),
             routes: [
                 /* ... */
             ],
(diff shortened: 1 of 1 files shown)


src/Vue/Resources/doc/index.rst
```diff
@@ -113,7 +114,11 @@ The event ``vue:before-mount`` is called before a component is mounted on the pa
         app.use(router);
     });
 
-The event ``vue:before-mount`` is called when a component has been mounted on the page:
+.. note::
+   When using Vue Router, it is recommended to use "hash" (or "memory") history mode,
+   to prevent your Vue routes from being served through Symfony controllers.
```
weaverryan, line 119: true history mode WOULD work, however, if you had some sort of a "catch all" route that executed the same controller, template and Vue component, right? For example, you are building a Vue app just for a "survey" section, for example. So you have a route:

```php
#Route('/survey/{path<.+>}')
public function survey($path = ''): Response
{
    // render the template
}
```

Assuming all Vue links point to URLs that start with `/survey`, this would a proper way to allow for the usage of the real history mode I think?

I like your note - and it made me think "Can we help instruct users also on how to use real history?"

pierredup: @weaverryan

> Can we help instruct users also on how to use real history?

Good idea, thanks. I added an extra section on using web history mode with a catch all route

weaverryan (approved):
This is awesome!

Kocal:
Thanks @pierredup ❤️

Issue #354, opened 2022-06-17
[LiveComponent] Using other UX components in Live form


akyoscommunication:
Hi,

I'm testing Live Components to make multiple steps forms with live validation. That's working fine! The tricky part is to use some js libs like datepicker, select-2, etc.. but with Stimulus controllers that works fine too. The missing thing is that other ux component doesn't work at all together. What could we do to make autocomplete, dropzone or even turbo work inside Live components ?

weaverryan:
Hey @akyoscommunication!

This is something I've thought about, but haven't played with yet. In theory, they should work fine, because Stimulus "always works". However, in practice, it's possible that the re-rendering of live components will basically reset/reinitialize certain fields that have another component attached to it (e.g. autocomplete). 

What bad behavior have you seen with mixing components inside of Live components? I would expect some to work fine - like Turbo - but others might need some help (e.g. autocomplete). One option is to use `data-live-ignore` on some of these elements - https://symfony.com/bundles/ux-live-component/current/index.html#skipping-updating-certain-elements - but that should really be a "last resort": we should make the components "play nicely" inside of live components as much as possible with no extra work.

Cheers!

akyoscommunication:
Hey,

Yes Stimulus controllers still works, but on each blur of a field or on each live action, Live component re-renders: some ux components like dropzone are resetted but still works (value is lost but we can continue put files in it), some others like autocomplete doesn't work anymore (JS code doesn't re-init automatically). We also tried Turbo to add some sub-component (push a "notification" component in a list) but we have some checksum problems..

Exemple with autocomplete (no css) : https://watch.screencastify.com/v/bwvcZrVCsRSMUjYiXwoR

Maybe the dropzone problem will disappear when  issue #289 will be merge. For autocomplete (and incoming Datepicker, Map, etc..), some idea:
- In Stimulus controller, detect on connect if the element is in a Livecomponent
- If so, store needed values (file input, select value, ...) in a variable and add a listener / a mutation observer on element.
- When mutation is triggered, destroy instance if still running, re-build it, and re-set the value

I made something like this with Datepicker and CKEDitor: https://watch.screencastify.com/v/731WxuOxx2DJdamkeAei , but there's still some bugs when component rerenders, it disappears before re-init js.

Thanks

akyoscommunication:
Miss click, sorry, that's not closed

weaverryan:
Thanks for all the extra info!

[...]

PR #519, opened 2022-11-02
[WIP][Live] New data-rerender mode for handling custom JavaScript

### LINKED ISSUE #270: [LiveComponent] data-live-ignore and nested elements
Currently when using `data-live-ignore` on an element, it will be ignored from re-render and also it's all child elements also will be ignored.

I have an use case with a modal, where I have a live action which triggers re-render and I would like to keep the modal open, but that the form in the modal would be re-rendered.

### LINKED ISSUE #490: [LiveComponent] Better control over re-rendering behavior (relevant for js widgets)
Hi!

Inspired by: https://github.com/symfony/ux/issues/489#issuecomment-1268246019

tl;dr: Suppose you have:

```twig
<div {{ stimulus_controller('my-date-picker') }}>
    <input type="text"> <!-- the date field that will be transformed
</div>
```

A) You add a JS widget inside your live component - e.g. a date picker - via a stimulus controller
B) That widget modifies the markup of part of your widget
C) On re-render, the modified markup is "reset" back to its original state, effectively killing your JS widget functionality

One solution to this is to force live components to ALWAYS fully re-render the entire `div` element attached to the Stimulus controller.. This forces the existing Stimulus controller (which loads the date picker) to be destroyed and totally reinitializing. So now, step (C) is different:

C) On re-render, the entire date picker element is re-rendered from scratch, causing the Stimulus controller to reinitialize and the date picker functionality to be re-added.

## Current Solution

Currently, this is possible by doing something silly like this:

```twig
<div data-live-id="{{ random() }}" {{ stimulus_controller('my-date-picker') }}>
```

The always-changing `data-live-id` tells live components to always re-render this.

## Proposed Solution

@tdumalin proposed something like:

```twig
<div data-live-rerender="always" {{ stimulus_controller('my-date-picker') }}>
```

With possible values of:

A) `always` - ALWAYS re-render this element
B) `never` - would replace `data-live-ignore`
C) `onChange` - @tdumalin what did you have in mind for this?

### LINKED ISSUE #515: [Live+Autocomplete] Wrong behavior of Autocomplete with dynamic forms
I have a Live component form with a Autocomplete input.
Autocomplete input is added or not depending on other form inputs.
When Autocomplete field is removed, tomselect field are not removed, because of data-live-ignore.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Tickets       | Fix #515, fix #490, fix #270, Bug H on #102, hopefully can include fix for #354 
| License       | MIT

Hi!

Just the docs to start. This aims to address the biggest pain-point with live components currently: when you have add custom JavaScript to your component HTML (e.g. a date picker). If that JS modifies the markup at all (e.g. adding classes or elements), when the component re-renders, those changes are "removed" and the custom JS stops working.

This adds a very flexible model for re-rendering. However, I think the `data-rerender="if(startDate)"` will handle 95% of use-cases because it will:

A) Avoid re-rendering the component in most cases (e.g. similar to `data-live-ignore` currently
B) But if some specific `LiveProp` changes (e.g. `startDate` for a date picker), then it will do a `complete` re-render of that element, where the old element is completely replaced by a new one. That's important because it will re-trigger the `connect()` method on any Stimulus controllers on (or inside) this element, which will allow them to completely re-initialize the JavaScript.

Please let me know if you think this will fit your use-case. And if, not, please tell me why!

Cheers!

### DIFF
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -500,8 +500,8 @@ can control via a ``data-rerender`` attribute.
     new one. This works will with Stimulus because the ``connect()``
     method will be called again.
 
-``data-rerender="if(modelName)"``
-    Do a ``complete`` re-render but *only* if the ``modelName`` LiveProp
+``data-rerender="if(propName)"``
+    Do a ``complete`` re-render but *only* if the ``propName`` LiveProp
     changed since the last re-render. This is useful to completely
     re-render an element (e.g. a date picker) but *only* if the value
     of the date picker changed (to avoid the date picker JavaScript from
@@ -510,19 +510,20 @@ can control via a ``data-rerender`` attribute.
 
 You can also specify "targets" for the re-render:
 
-``data-rerender="children|always"``
+``data-rerender="children"``
     Always re-render this element, but *only* the children. Avoid making
     any changes to the parent element that holds this attribute. This is
     useful to add to modal wrappers so that, if the modal is open, re-rendering
-    won't override the classes that caused it to display as open.
+    won't override the classes that caused it to display as open. Note: ``children``
+    are always re-rendered using the normal, "diff" mode.
 
-``data-rerender="self|if(someModel)"``
+``data-rerender="self|if(someProp)"``
     Only re-render the parent element that holds this attribute and only
-    if the ``someModel`` LiveProp changed. This could be useful if you're
+    if the ``someProp`` LiveProp changed. This could be useful if you're
     rendering a chart
 
 You can even specify multiple directives separated by a space:
-``data-rerender="children|if(someModel) self|always"``.
+``data-rerender="children|if(someProp) self|always"``.
 
 .. note::
 
--- a/src/LiveComponent/src/Resources/doc/index.rst
+++ b/src/LiveComponent/src/Resources/doc/index.rst
@@ -475,6 +475,99 @@ The following hooks are available (along with the arguments that are passed):
 * ``loading.state:finished`` args ``(element: HTMLElement)``
 * ``model:set`` args ``(model: string, value: any, component: Component)``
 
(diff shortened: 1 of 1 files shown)


src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -475,6 +475,100 @@ The following hooks are available (along with the arguments that are passed):
...
+.. note::
+
+    Want to get geeky & technical? The full, format of
+    ``data-rerender`` is: ``{target}|{mode}|{when}``:
+
+        - ``target`` includes ``self``, ``children``, ``all`` (default)
+        - ``mode`` includes ``diff`` & ``complete`` (default). ``diff`` is the
+          default mode used by live components where the new element is
+          "morphed" onto the existing element by applying a diff of changes. However,
+          as soon as you specify ``data-render``, ``complete`` becomes the default
```
bendavies, line 537:
```suggestion
          as soon as you specify ``data-rerender``, ``complete`` becomes the default
```

src/LiveComponent/src/Resources/doc/index.rst
```diff
@@ -475,6 +475,100 @@ The following hooks are available (along with the arguments that are passed):
...
+JavaScript: Date Picker Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Suppose you have a ``date-picker`` custom Stimulus controller that
+some fancy date picker, like ``flatpickr``. To avoid removing the
+date picker markup on re-render, we want to
+
+.. code-block:: twig
+
+    <div data-render"">
```
bendavies, line 562:
```suggestion
    <div data-rerender="">
```

jcrombez:
Very nice idea, i hope it gets implemented <3

weaverryan:
Replaced by #728

PR #532, opened 2022-11-06
[Live] coerce falsey strings to scalar types

### LINKED ISSUE #526: [Live] scalar type coercion when hydrating
If a component property's type-hint is `int` and you try to hydrate with `''`, you get an error. I propose we automatically cast simple, scalar types:

- `int`: `''` or `'  '` => `0`
- `float`: `''` or `'  '` => `0.0`
- `bool`:  `''` or `'  '` => `false`
- `?int`: `''` or `'  '` => `null`
- `?float`: `''` or `'  '` => `null`
- `?bool`:  `''` or `'  '` => `null`

How to handle?
- `int`: `'apple'`?

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #526
| License       | MIT

I chose to allow _weird_ values to be cast to the proper type instead of throwing an exception (ie `0 === (int) 'apple'`).

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -148,6 +148,8 @@ final class LiveComponentHydrator
             if ($method = $liveProp->hydrateMethod()) {
                 // TODO: Error checking
                 $value = $component->$method($value);
+            } elseif (\is_string($value) && $type && \in_array($type->getName(), ['int', 'float', 'bool'], true)) {
+                $value = self::coerceScalarValue($value, $type);
             } elseif (!$value && $type && $type->allowsNull() && is_a($type->getName(), \BackedEnum::class, true) && !\in_array($value, array_map(fn (\BackedEnum $e) => $e->value, $type->getName()::cases()))) {
                 $value = null;
             } elseif (null !== $value && $type && !$type->isBuiltin()) {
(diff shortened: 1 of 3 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -148,6 +148,8 @@ public function hydrate(object $component, array $data, string $componentName):
             if ($method = $liveProp->hydrateMethod()) {
                 // TODO: Error checking
                 $value = $component->$method($value);
+            } elseif (\is_string($value) && $type && \in_array($type->getName(), ['int', 'float', 'bool'], true)) {
+                $value = self::coerceScalarValue($value, $type);
             } elseif (!$value && $type && $type->allowsNull() && is_a($type->getName(), \BackedEnum::class, true) && !\in_array($value, array_map(fn (\BackedEnum $e) => $e->value, $type->getName()::cases()))) {
                 $value = null;
             } elseif (null !== $value && $type && !$type->isBuiltin()) {
```
weaverryan, line 155: This area is increasingly feeling "ad-hoc", but it's well-covered in tests. And if there is some broader solution for handling these edge cases, it'll become obvious eventually 👍

weaverryan:
Thanks Kevin!

PR #536, opened 2022-11-07
[Live] allow dehydrating unpersisted entities

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | n/a
| License       | MIT

If unpersisted, fallback to standard object normalizer.

### DIFF
--- a/src/LiveComponent/tests/Fixtures/Entity/Entity1.php
+++ b/src/LiveComponent/tests/Fixtures/Entity/Entity1.php
@@ -23,5 +23,10 @@ class Entity1
      * @ORM\GeneratedValue
      * @ORM\Column(type="integer")
      */
-    public $id;
+    public ?int $id;
+
+    /**
+     * @ORM\Column(nullable=true)
+     */
+    public ?string $title = null;
 }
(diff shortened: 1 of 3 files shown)


src/LiveComponent/tests/Fixtures/Entity/Entity1.php
```diff
@@ -23,5 +23,10 @@ class Entity1
      * @ORM\GeneratedValue
      * @ORM\Column(type="integer")
      */
-    public $id;
+    public ?int $id;
```
kbond, line 26: Identifier needs to be in this exact format unfortunately. 

`public $id;` or `public ?int $id = null;` cause an exception during denormalization:

> Symfony\Component\Serializer\Exception\NotNormalizableValueException: The type of the "id" attribute for class "Symfony\UX\LiveComponent\Tests\Fixtures\Entity\Entity1" must be one of "int" ("null" given).

I believe this is because the Doctrine property type extractor concludes this field cannot be null...

welcoMattic (approved):
:+1:

ker0x:
Hello @weaverryan, @kbond! If this PR is ready, can it be merged ?

Thanks for your work 😉

kbond:
@weaverryan, was this taken care of in your recent refactor of the hydration system?

weaverryan:
Ah, yes. This will be possible in 2.8. Out-of-the-box, an unpersisted entity will be dehydrated to `[]` and rehydrated to via `new YourEntity()` with no constructor args. If that's not good enough, you can add `LiveProp(useSerializerForHydration: true)` and then your entity will go through the serializer for (de)hydration.

PR #538, opened 2022-11-09
[Autocomplete] max_results option is not passed to parent entity

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | none 
| License       | MIT

There seems to be an oversight from #478 when it comes to passing the option from types that use `#[AsEntityAutocompleteField]`.

This fix removes the unsetting of `$options['max_results']` from the `AutocompleteEntityTypeSubscriber` to allow passing the information. I also extended the tests to cover this case.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


weaverryan:
Great catch and excellent coverage in the test. Thank you @daFish!

PR #546, opened 2022-11-11
Fix redirection with nginx and http2

### LINKED ISSUE #543: [Live] With nginx and http2, the "Content-Type" header is stripped in a 204 response
Pre-requisites: 
* nginx with http2
* redirection in a LiveAction
* symfony/ux-live-component v2.5.0

Steps:

* Any redirection in a controller with a LiveAction
* Symfony returns a 204 response with a specific header  Content-Type - [\Symfony\UX\LiveComponent\EventListener](https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php#L283)
* Nginx drops the header - https://trac.nginx.org/nginx/ticket/2109
* Symfony displays an error because the header is missing - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/dist/live_controller.js#L1520

Technically it is a nginx bug but a 204 response means "no content" so it is confusing to have a content type header when there is none?
Moreover nginx does not seem in a hurry to fix it (2 years and counting) so I think it should be fixed in this package. Maybe simply an other header or another way to detect errors?

xaviermarchegay:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix https://github.com/symfony/ux/issues/543
| License       | MIT

> Let’s drop the header during the redirect (which is when we have an empty body) and, instead, add a 2nd header - something custom like X-Live-Redirect. Then, in the javascript, only throw the error of the content-type is wrong/missing AND that special header is missing.
> 
> I’d welcome a PR of you’re open to it :)

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1517,7 +1517,8 @@ class Component {
             const backendResponse = new BackendResponse(response);
             thisPromiseResolve(backendResponse);
             const html = await backendResponse.getBody();
-            if (backendResponse.response.headers.get('Content-Type') !== 'application/vnd.live-component+html') {
+            const headers = backendResponse.response.headers;
+            if (headers.get('Content-Type') !== 'application/vnd.live-component+html' && !headers.get('X-Live-Redirect')) {
                 this.renderError(html);
                 return response;
             }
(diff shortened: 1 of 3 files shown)


weaverryan (approved):
That was EXACTLY what I had in mind

weaverryan:
Thanks Xavier!

PR #552, opened 2022-11-14
Add `mount_component` twig function

### LINKED ISSUE #542: [TwigComponent] Component refs
Currently when you use `component` twig function, it immediatelly renders the component.
I'd like to have a possibility to get the component ref and modify it if needed before rendering.

For example rendering a table with custom components:
```php
class TableComponent {
    public string $align;
    public function mount(string|array $children) { $this->children = is_array($children) ? implode('', $children) : $children; }
}

class TableCellComponent {
    public string $align;
    public string $children;
}
```

When using it I need to pass the `align` prop down to the TableCell components:
```twig
{{ component('Table', {
  align: 'center',
  children: [
    component('TableCell', { align: 'center', children: 'Cell 1' }),
    component('TableCell', { align: 'center', children: 'Cell 2' }),
    component('TableCell', { align: 'center', children: 'Cell 3' }),
  ]
}}
```
What I'd like to have is:
```twig
Table.html.twig:
<table>
  {% for cellComponentRef in children %}
    {% do cellComponentRef.align = align %} {# set it from the table align prop #}
    {{ cellComponentRef.render() }}
  {% endfor %}
</table>
```

Is it feasible? Or there are better ways to manage child components from the parent? Or use some sort of contexts like in react?

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #542
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/MountedComponent.php
+++ b/src/TwigComponent/src/MountedComponent.php
@@ -62,6 +62,11 @@ final class MountedComponent
         return $this->inputProps;
     }
 
+    public function withProps(array $props): void
+    {
+        $this->inputProps = array_merge($this->inputProps ?? [], $props);
+    }
+
     public function addExtraMetadata(string $key, mixed $metadata): void
     {
         $this->extraMetadata[$key] = $metadata;
--- a/src/TwigComponent/src/RenderableComponent.php
+++ b/src/TwigComponent/src/RenderableComponent.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+/**
+ * @internal
+ */
+final class RenderableComponent
+{
+    public function __construct(
+        private ComponentRenderer $componentRenderer,
+        private MountedComponent $component,
+    ) {
+    }
+
+    public function withProps(array $props): self
+    {
+//        $r = new \ReflectionObject($this->component);
+//        $inputProps = $r->getProperty('inputProps');
+//        $inputProps->setValue($this->component, array_merge($inputProps->getValue($this->component), $props));
+
+        $this->component->withProps($props);
+        return $this;
+    }
+
+    public function __toString(): string
+    {
+        return $this->componentRenderer->render($this->component);
+    }
+}
--- a/src/TwigComponent/src/Twig/ComponentExtension.php
+++ b/src/TwigComponent/src/Twig/ComponentExtension.php
@@ -15,9 +15,12 @@ use Psr\Container\ContainerInterface;
 use Symfony\Contracts\Service\ServiceSubscriberInterface;
 use Symfony\UX\TwigComponent\ComponentFactory;
 use Symfony\UX\TwigComponent\ComponentRenderer;
+use Symfony\UX\TwigComponent\MountedComponent;
+use Symfony\UX\TwigComponent\RenderableComponent;
 use Twig\Error\RuntimeError;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
+use Twig\TwigTest;
 
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
@@ -41,7 +44,14 @@ final class ComponentExtension extends AbstractExtension implements ServiceSubsc
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('component', [$this, 'render'], ['is_safe' => ['all']]),
+            new TwigFunction('component', [$this, 'create'], ['is_safe' => ['all']]),
+        ];
+    }
+
+    public function getTests(): array
+    {
+        return [
+            new TwigTest('component', [$this, 'isComponent']),
         ];
     }
 
@@ -52,10 +62,13 @@ final class ComponentExtension extends AbstractExtension implements ServiceSubsc
         ];
     }
 
-    public function render(string $name, array $props = []): string
+    public function create(string $name, array $props = []): RenderableComponent
     {
         try {
-            return $this->container->get(ComponentRenderer::class)->createAndRender($name, $props);
+            return new RenderableComponent(
+                $this->container->get(ComponentRenderer::class),
+                $this->container->get(ComponentFactory::class)->create($name, $props),
+            );
         } catch (\Throwable $e) {
             $this->throwRuntimeError($name, $e);
         }
@@ -76,5 +89,10 @@ final class ComponentExtension extends AbstractExtension implements ServiceSubsc
             $e = new \Exception($e->getMessage(), $e->getCode(), $e->getPrevious());
         }
         throw new RuntimeError(sprintf('Error rendering "%s" component: %s', $name, $e->getMessage()), previous: $e);
+    }
+
+    public function isComponent(mixed $value): bool
+    {
+        return $value instanceof RenderableComponent;
     }
 }
--- a/src/TwigComponent/tests/Fixtures/Component/TableCellComponent.php
+++ b/src/TwigComponent/tests/Fixtures/Component/TableCellComponent.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+
+#[AsTwigComponent('table_cell')]
+final class TableCellComponent
+{
+    public string $align = 'left';
+    public string $content;
+}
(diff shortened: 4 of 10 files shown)


norkunas:
As the library is still experimental, should I add `render` method directly to the `ComponentRendererInterface` ?

norkunas:
~~Also - I still would like to have `__toString()` implemented in the MountedComponent, unless you have a suggestion how to detect in Twig if I have a MountedComponent or a rendered component..~~

Added a `var is component` twig test.

norkunas:
Can I get feedback here or should I not expect this? :disappointed:

weaverryan:
Yup - sorry about that. Going back to #542, could we solve it a different way? I *really* want to challenge that something like this is a solution we LIKE. So, you had:

```
{{ component('Table', {
  align: 'center',
  children: [
    component('TableCell', { align: 'center', children: 'Cell 1' }),
    component('TableCell', { align: 'center', children: 'Cell 2' }),
    component('TableCell', { align: 'center', children: 'Cell 3' }),
  ]
}}
```

Another solution is (with some pseudo code) would be:

```
// some_template.html.twig
{{ component('Table', {
  align: 'center',
  children: [
    { children: 'Cell 1' },
    { children: 'Cell 2' },
    { children: 'Cell 3' },
  ]
}}
```

Then in `Table.html.twig`:

```
<table>
  {% for cell in children %}
    {{ component('TableCell', cell|merge({align: align})) }}
  {% endfor %}
</table>
```

WDYT? I don't see any profit to creating `TableCell` component objects in the original template (e.g. `some_template.html.twig`) vs passing the raw data (e.g. `{ children: 'Cell 1' }` for each one). With the new Twig component syntax, we could probably make the `Table.html.twig` template a bit nicer too:

```
<table>
  {% for cell in children %}
    <twig:TableCell ...cell :align="align" />
  {% endfor %}
</table>
```

(though we'd probably need to play with the ideal syntax a bit).

norkunas:
> Yup - sorry about that. Going back to #542, could we solve it a different way? I _really_ want to challenge that something like this is a solution we LIKE. So, you had:
> 
> ```
> {{ component('Table', {
>   align: 'center',
>   children: [
>     component('TableCell', { align: 'center', children: 'Cell 1' }),
>     component('TableCell', { align: 'center', children: 'Cell 2' }),
>     component('TableCell', { align: 'center', children: 'Cell 3' }),
>   ]
> }}
> ```
> 
> Another solution is (with some pseudo code) would be:
> 
> ```
> // some_template.html.twig
> {{ component('Table', {
>   align: 'center',
>   children: [
>     { children: 'Cell 1' },
>     { children: 'Cell 2' },
>     { children: 'Cell 3' },
>   ]
> }}
> ```
> 
> Then in `Table.html.twig`:
> 
> ```
> <table>
>   {% for cell in children %}
>     {{ component('TableCell', cell|merge({align: align})) }}
>   {% endfor %}
> </table>
> ```

What If I am not in control of the component? :) Also this was just a brief example.
We have `mui.com` alike design system made of Twig components and in many places we have to adjust the `inner` rendering based on the provided parent component props.

weaverryan:
Hmm, ok. So let's keep trying then :). The solution in this PR might be the best, but I'd really like to make sure first. A few other ideas (using the new syntax, for readability + a non-existing `parent.props` thing):

```
<twig:Table align="center">
    <TableCell children="Cell 1" align="{{ parent.props.align }}">
</twig>
```

I'm not really sure that this is what you want either. What would your *dream* syntax be? Is it what you have in https://github.com/symfony/ux/issues/542#issue-1443460520 ?

And if so, perhaps we think about this a slightly different way. This may be crazy or silly (and ignoring BC), what if `{{ component() }}` returned some sort of `RenderableComponent` with a `__toString()` method? Inside (other than the needed services), it would contain JUST the info passed to `component()`: the component name + input props + embedded blocks somehow. So, at this point, we have NOT yet "mounted" the component: we're just storing the "mount" info for later.

The "normal" usage would stay the same: when you actually echo this object - e.g. `{{ component('Alert') }}`, it would then create the component, mount it, and render it like now.

Using this for your problem would be very similar to what you proposed in #542:

```
{{ component('Table', {
  align: 'center',
  children: [
    component('TableCell', { children: 'Cell 1' }),
    component('TableCell', { children: 'Cell 2' }),
    component('TableCell', { children: 'Cell 3' }),
  ]
}}
```

```
Table.html.twig:
<table>
  {% for cellComponentRef in children %}
    {# a convenience method to modify the align "input prop" before actually rendering #}
    {{ cellComponentRef.withProps({ align: align }) }}
  {% endfor %}
</table>
```

Allowing the user to modify the "input prop" vs just modifying the component object directly (e.g. `cellComponentRef.align = align`) is important for live parent->child components. That's because when a parent component re-renders, we check to see if any "input props" passed to the child have changed (e.g. `align`) and then, depending on your config, we might re-render the child.

norkunas:
about the `parent.props` - I'm not sure.. because that's an additional work anyway instead of just letting to `Table` component override cells alignment.

`withProps` would work for me :) the point is that in a higher component level we should be able to decide how we want to render based on the provided data.

and as an additional solution - for example I don't want to be strict that only TableCell components are accepted - they are just part of the components, but if in another place it just wouldn't fit, so I could pass also a simple `<td>` without any component so then the Table component would ignore it and that's all - so I need somehow to be able to check if it's a component and what component to be able to decide what to do

weaverryan:
> withProps would work for me :) the point is that in a higher component level we should be able to decide how we want to render based on the provided data.

I think we have a proposed solution then - just need to implement it.

> and as an additional solution - for example I don't want to be strict that only TableCell components are accepted - they are just part of the components, but if in another place it just wouldn't fit, so I could pass also a simple <td> without any component so then the Table component would ignore it and that's all - so I need somehow to be able to check if it's a component and what component to be able to decide what to do

What would the code look like to render the `Table` component but pass in `<td>` instead? This?

```
{{ component('Table', {
  align: 'center',
  children: [
    '<td>Foo</td>,
  ]
}}
```

norkunas:
> I think we have a proposed solution then - just need to implement it.

Ok, will try to update PR tomorrow :)

> What would the code look like to render the `Table` component but pass in `<td>` instead? This?
> 
> ```
> {{ component('Table', {
>   align: 'center',
>   children: [
>     '<td>Foo</td>,
>   ]
> }}
> ```

Yes, that's the current solution :) Table accepts string/array of strings, so would need to update to also support RenderableComponent

norkunas:
I'm having some troubles with this because of `src/TwigComponent/src/ComponentRenderer.php:L77 preRender` method.
It just uses directly properties from the `MountedComponent` instead of taking it's `inputProps` :disappointed: 

To make things less complex I've tried to also introduce `__toString` directly on the `MountedComponent` but then the circular reference gets into play because of ComponentFactory<->ComponentRenderer

seb-jean:
I like this solution because the inner render can be adjusted from one template to another:
```twig
<twig:Table align="center">
    <TableCell children="Cell 1" align="{{ parent.props.align }}">
    <TableCell children="Cell 2" align="{{ parent.props.align }}">
    <TableCell children="Cell 3" align="{{ parent.props.align }}">
</twig>
```

For my case, I make a menu that looks like [Headless UI](https://headlessui.com/react/menu). Here is my code:
```twig
<twig:Menu element="div" class="relative inline-block text-left">
  <div>
    <twig:MenuButton type="options" class="inline-flex w-full justify-center rounded-md bg-black bg-opacity-20 px-4 py-2 text-sm font-medium text-white hover:bg-opacity-30 focus:outline-none focus-visible:ring-2 focus-visible:ring-white focus-visible:ring-opacity-75" id="menu-button" aria-controls="...">
      Options
      <twig:Heroicons class="ml-2 -mr-1 h-5 w-5 text-violet-200 hover:text-violet-100" aria-hidden="true" />
    </twig:MenuButton>
  </div>
  <twig:MenuItems class="absolute right-0 mt-2 w-56 origin-top-right divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none" id="menu-items" aria-labelledby="...">
    <div class="px-1 py-1">
      <twig:MenuItem>
        <button class="text-gray-900 group flex w-full items-center rounded-md px-2 py-2 text-sm">
          <twig:Heroicons class="mr-2 h-5 w-5" type="edit" aria-hidden="true" />
          Edit
        </button>
      </twig:MenuItem>
      <twig:MenuItem>
        <button class="text-gray-900 group flex w-full items-center rounded-md px-2 py-2 text-sm">
          <twig:Heroicons class="mr-2 h-5 w-5" type="duplicate" aria-hidden="true" />
          Duplicate
        </button>
      </twig:MenuItem>
    </div>
    <div class="px-1 py-1">
      <twig:MenuItem>
        <button class="text-gray-900 group flex w-full items-center rounded-md px-2 py-2 text-sm">
          <twig:Heroicons class="mr-2 h-5 w-5" type="archive" aria-hidden="true" />
          Archive
        </button>
      </twig:MenuItem>
      <twig:MenuItem>
        <button class="text-gray-900 group flex w-full items-center rounded-md px-2 py-2 text-sm">
          <twig:Heroicons class="mr-2 h-5 w-5" type="move" aria-hidden="true" />
          Move
        </button>
      </twig:MenuItem>
    </div>
    <div class="px-1 py-1">
      <twig:MenuItem>
        <button class="text-gray-900 group flex w-full items-center rounded-md px-2 py-2 text-sm">
          <twig:Heroicons class="mr-2 h-5 w-5" type="delete" aria-hidden="true" />
          Delete
        </button>
      </twig:MenuItem>
    </div>
  </twig:MenuItems>
</twig:Menu>
```

I would like to dynamically retrieve the CSS id of `<twig:MenuButton />` and `<twig:MenuItems />` to put them respectively in an HTML attribute `aria-labelledby` and `aria-controls`

Issue #556, opened 2022-11-17
[LiveComponent] SVG is not updated on rerender


norkunas:
I've tried to render a SVG in a live component. And after re-render if component renders a different SVG, it is not updated.

PR #557, opened 2022-11-17
[LiveComponent] Fix svg re-render

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | #556
| License       | MIT

### DIFF
--- a/src/LiveComponent/assets/src/morphdom.ts
+++ b/src/LiveComponent/assets/src/morphdom.ts
@@ -40,7 +40,10 @@ export function executeMorphdom(
                 return true;
             }
 
-            if (!(fromEl instanceof HTMLElement) || !(toEl instanceof HTMLElement)) {
+            if (
+                !(fromEl instanceof HTMLElement || fromEl instanceof SVGElement) ||
+                !(toEl instanceof HTMLElement || toEl instanceof SVGElement)
+            ) {
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


weaverryan:
Thanks for the reproducer test - super useful. I can't think, immediately, what the problem might be - maybe something in morphdom? Unless someone beats me to it (which would be great!) I'll check it out when I have some time.

norkunas:
Thought about morphdom too, but they have svg tests,so not sure how this could happen currently..

norkunas:
Found the root cause and added a fix.

src/LiveComponent/assets/src/morphdom.ts
```diff
@@ -40,7 +40,10 @@ export function executeMorphdom(
                 return true;
             }
 
-            if (!(fromEl instanceof HTMLElement) || !(toEl instanceof HTMLElement)) {
+            if (
+                !(fromEl instanceof HTMLElement || fromEl instanceof SVGElement) ||
+                !(toEl instanceof HTMLElement || toEl instanceof SVGElement)
+            ) {
```
weaverryan, line 46: Nice find. SVG's not being part of `HTMLElement` is such a pain in the butt :). Thanks for the test case to prevent it from slipping through in the future.

weaverryan:
Thanks Tomas!

PR #559, opened 2022-11-22
Removing "experimental" from all components (except for Live)

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | None
| License       | MIT

Hi!

As promised last week at SymfonyCon, it's time to remove the experimental status from the UX components! 🔥

This means we will protect backward compatibility as we do in the core. The great news is that, for the vast majority of the components, we haven't made breaking changes during the last 1+ year anyways!

The one exception is LiveComponents which still needs more work before stabilizing.

Cheers!

### DIFF
--- a/src/Autocomplete/README.md
+++ b/src/Autocomplete/README.md
@@ -2,9 +2,6 @@
 
 Javascript-powered auto-completion functionality for your Symfony forms!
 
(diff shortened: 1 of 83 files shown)


chalasr (approved):
🚀

PR #561, opened 2022-11-25
[Autocomplete] Don't add WHERE IN criteria without params

### LINKED ISSUE #549: [Autocomplete] 500 when using multiple=true on a not required field
Since https://github.com/symfony/ux/pull/494, Autocomplete fields doesnt work for not required inputs.
The error : `SELECT o FROM App\Entity\City o WHERE o.id IN ()`

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #549
| License       | MIT

### DIFF
--- a/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
+++ b/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
@@ -73,11 +73,15 @@ final class AutocompleteEntityTypeSubscriber implements EventSubscriberInterface
                     ++$idx;
                 }
 
-                $options['choices'] = $repository->createQueryBuilder('o')
-                    ->where(sprintf("o.$idField IN (%s)", implode(', ', array_keys($params))))
-                    ->setParameters(new ArrayCollection($params))
-                    ->getQuery()
-                    ->getResult();
+                $queryBuilder = $repository->createQueryBuilder('o');
+
+                if ($params) {
+                    $queryBuilder
+                        ->where(sprintf("o.$idField IN (%s)", implode(', ', array_keys($params))))
+                        ->setParameters(new ArrayCollection($params));
+                }
+
+                $options['choices'] = $queryBuilder->getQuery()->getResult();
             } else {
                 $options['choices'] = $repository->createQueryBuilder('o')
                     ->where("o.$idField = :id")
--- a/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
+++ b/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
@@ -93,4 +93,25 @@ class AutocompleteFormRenderingTest extends KernelTestCase
             ->assertContains('Sugar')
         ;
     }
+
+    public function testMultipleDoesNotFailWithoutSelectedChoices()
+    {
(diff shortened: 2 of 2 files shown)


src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
```diff
@@ -73,11 +73,15 @@ public function preSubmit(FormEvent $event)
...
-                    ->where(sprintf("o.$idField IN (%s)", implode(', ', array_keys($params))))
-                    ->setParameters(new ArrayCollection($params))
-                    ->getQuery()
-                    ->getResult();
+                $queryBuilder = $repository->createQueryBuilder('o');
+
+                if ($params) {
+                    $queryBuilder
+                        ->where(sprintf("o.$idField IN (%s)", implode(', ', array_keys($params))))
+                        ->setParameters(new ArrayCollection($params));
```
Kocal, line 81: My two cents, using `->setParameters()` will override the previous bound parameters (if any) to the `QueryBuilder`.

IMO it would be preferable to use `->setParameter()` for each `$params`, to be future-proof.

norkunas: There can't be any previous parameters, because the query is constructed 4 lines before :)

daFish:
@norkunas Great fix. Would you mind adding a testcase?

src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
Kocal: For the moment yes, but we can't predict the future.

norkunas: Imho they could be added to $params,no? :)

src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
```diff
@@ -73,11 +73,15 @@ public function preSubmit(FormEvent $event)
...
                 }
 
-                $options['choices'] = $repository->createQueryBuilder('o')
-                    ->where(sprintf("o.$idField IN (%s)", implode(', ', array_keys($params))))
-                    ->setParameters(new ArrayCollection($params))
-                    ->getQuery()
-                    ->getResult();
+                $queryBuilder = $repository->createQueryBuilder('o');
+
+                if ($params) {
```
1ed, line 78: Why don't just check if `$data['autocomplete']` is `[]` and skip this else part?
Btw this can be simplified a lot like `$repository->createQueryBuilder('o')->->where(sprintf('o.%s IN (:ids)', $idField)))->setParameter('ids', $data['autocomplete'])->getQuery()->getResult()`, doctrine will handle the rest, doesn't it?

norkunas: No it won't handle the types properly,that's what I've fixed before.

norkunas:
> @norkunas Great fix. Would you mind adding a testcase?

testcase added.

weaverryan:
Awesome, thank you Tomas!

PR #562, opened 2022-11-26
[Live] Allow to edit component route

### LINKED ISSUE #514: [Live] Allow to edit component route
Following our discussion with @weaverryan

I have a use case where I need to specify my component's route.
I have a website with two firewalls (front + back). Back firewall is registered on all `/back*` routes on `security.yml`.
I need to get the current logged in back user on my Live component, but because all components are registered on `_component/` routes, I can't.

simondaigre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #514
| License       | MIT

Added a new `route` parameter to `AsLiveComponent`, which allows to choose another route for Ajax calls.

### DIFF
--- a/src/LiveComponent/src/Controller/BatchActionController.php
+++ b/src/LiveComponent/src/Controller/BatchActionController.php
@@ -37,7 +37,7 @@ final class BatchActionController
                 '_controller' => [$serviceId, $name],
                 '_component_action_args' => $action['args'] ?? [],
                 '_mounted_component' => $_mounted_component,
-                '_route' => 'ux_live_component',
+                '_live_component' => $serviceId,
             ]);
 
             $response = $this->kernel->handle($subRequest, HttpKernelInterface::SUB_REQUEST, false);
--- a/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
@@ -313,7 +313,7 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
 
     private function isLiveComponentRequest(Request $request): bool
     {
-        return 'ux_live_component' === $request->attributes->get('_route');
+        return str_starts_with($request->attributes->get('_route'), 'ux_live_component');
     }
 
     private function hydrateComponent(object $component, string $componentName, Request $request): MountedComponent
--- a/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
@@ -76,8 +76,8 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
         }
 
         // the default "action" is get, which does nothing
-        $action = $request->attributes->get('action', 'get');
-        $componentName = (string) $request->attributes->get('component');
+        $action = $request->attributes->get('_live_action', 'get');
+        $componentName = (string) $request->attributes->get('_live_component');
 
         $request->attributes->set('_component_name', $componentName);
 
@@ -313,7 +313,7 @@ class LiveComponentSubscriber implements EventSubscriberInterface, ServiceSubscr
 
     private function isLiveComponentRequest(Request $request): bool
     {
-        return str_starts_with($request->attributes->get('_route'), 'ux_live_component');
+        return $request->attributes->has('_live_component');
     }
 
     private function hydrateComponent(object $component, string $componentName, Request $request): MountedComponent
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## 2.7.0
 
-- Added a new `route` parameter to `AsLiveComponent`, which allows to choose another route for Ajax calls.
+-   Added a new `route` parameter to `AsLiveComponent`, which allows to choose
+    another route for Ajax calls.
 
 ## 2.6.0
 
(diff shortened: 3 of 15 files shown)


src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -313,7 +313,7 @@ private function createResponse(MountedComponent $mounted): Response
 
     private function isLiveComponentRequest(Request $request): bool
     {
-        return 'ux_live_component' === $request->attributes->get('_route');
+        return str_starts_with($request->attributes->get('_route'), 'ux_live_component');
```
kbond: This assumes a user will start their custom route name with `ux_live_component`?

simondaigre: Yes or we can loop thru all Live components and find if a route match. WDYT ?

kbond: Hmm I think we can do that when compiling the container to create a list of live component routes. Then we just see if the current route is in this compiled list.

@weaverryan, any other ideas?

kbond: Could it be as simple as having the JS send a custom header and not caring about the route name?

simondaigre: Both solutions are great. @weaverryan any ideas ?

weaverryan: Hmmm. As the AJAX endpoints return HTML, it should be legal to make a request to the URL with the normal `Content-Type: text/html` header / just put the URL in your browser and have it work. So I think we need to do the normal complex container compilation step :). 

Or, 3rd option. The current route looks like this:

```
$routes->add('ux_live_component', '/{component}/{action}')
```

We could change `{component}` to `{_live_component}` and `{action}` to `{_live_action}`. Then we would look for  a`_live_component` attribute to determine if this is a "live component request". 

Also, thinking about the workflow, I guess you would create your custom route... in YAML somewhere with no controller? That's fine - it feels a bit odd, but I can't think of a better way.

simondaigre: @weaverryan I like your 3rd option but I think there will be BC with it ?

> Also, thinking about the workflow, I guess you would create your custom route... in YAML somewhere with no controller? That's fine - it feels a bit odd, but I can't think of a better way.

I need to define the path to match my firewall, so yes YAML will be the best way IMHO.

kbond: > in YAML somewhere with no controller?

That's effectively what we're doing in this package.

> I like your 3rd option but I think there will be BC with it ?

I don't think so, the routing file should be imported and not defined manually by the user. Everywhere we generate the url is internal. I guess it could be a BC break if someone is doing something unexpected. This package is still experimental so we're good regardless.

simondaigre: I just figure another problem with Ryan's option: we can't rename `{component}` to `{_live_component}` and `{action}` to `{_live_action}`, because `{component}` and `{action}` are also related to TwigComponent ...

@kbond would you like to continue this PR ?

kbond: > I just figure another problem with Ryan's option...

Can you explain this a bit more - I'm not following.

> would you like to continue this PR ?

Time's a bit tight for me right now but I'll see what I can do.

weaverryan: What do you mean by this?

> `{component}` and `{action}` are also related to TwigComponent

simondaigre: Nevermind, I forgot to clear my project cache after my changes.
PR updated with Ryan's proposal 3.

kbond:
This looks good to me! @simondaigre, can you fixup the CI and add some docs?

simondaigre:
@kbond would you mind taking a look ? I can't figure out why they fail...

src/LiveComponent/src/Controller/BatchActionController.php
```diff
@@ -37,7 +37,7 @@ public function __invoke(Request $request, MountedComponent $_mounted_component,
                 '_controller' => [$serviceId, $name],
                 '_component_action_args' => $action['args'] ?? [],
                 '_mounted_component' => $_mounted_component,
-                '_route' => 'ux_live_component',
+                '_live_component' => $serviceId,
```
kbond, line 40: Fixed, this was the culprit (since the `_route` doesn't trigger the subscriber anymore).

kbond:
FYI @simondaigre, for future PR's, it'll be easier to use a new branch (rather than making changes on 2.x directly).

simondaigre:
Understood. Thanks for your help !

kbond:
CI is green.

- We should add a test for this feature. @simondaigre, would you like to add this or want me to add one?
- Can you add documentation?

simondaigre:
You can add tests if you want. I'll do the docs tomorrow.

kbond:
> You can add tests if you want.

Will do, the test suite is a bit complex 😬

simondaigre:
Doc updated. I have no clue to how to preview this. @weaverryan feel free to edit it if you want.

kbond:
@weaverryan, will this be an issue for the batch actions? I'm thinking no but want to double-check with you.

kbond:
Tests added, @weaverryan, can you give a final review?

weaverryan:
Big thanks for both of you for this awesome work!

PR #564, opened 2022-11-28
[Autocomplete] Add support for latest tom select-version

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (sort of)
| New feature?  | no 
| Tickets       | does not apply
| License       | MIT

This change is an attempt to enable the update of Tom Select to the recent version 2.2.2. There are 2 more or less specific issues orchidjs/tom-select/issues/500 and orchidjs/tom-select/issues/501.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+-   Add support for recent versions of Tom Select
+
 ## 2.6.0
 
 -   [BC BREAK]: The path to `routes.php` changed and you should update your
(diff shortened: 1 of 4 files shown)


weaverryan:
Thanks for this! The change just affects compiling the TypeScript, correct? It doesn't affect end-users?

daFish:
Exactly. I encountered the problem after an automated dependency update didn't work because I couldn't compile to Javascript anymore.

weaverryan:
Ok, thanks! I see this is still a draft - is there more to add? Or are you waiting on those 2 linked issues to see if the problem might be resolved/altered there?

daFish:
I wasn't sure if there is something else to add.

weaverryan:
Me either! But I also didn't see what the original build failure was. But this seems reasonable.

weaverryan:
Thanks Marcus!

PR #313, opened 2022-04-18
[Twig] throw error if twig-bundle is not installed

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Ran into an issue with `ContainerBuilder::willBeAvailable()` and `ux-live-component`'s test suite when requiring `symfony/twig-bundle` directly. Not perfect but throwing an exception on container build if `twig-bundle` is not available seemed like the easiest solution.

PR #569, opened 2022-11-28
[LiveComponent][TwigComponent] Define requirement to the twig bundle for UX Component and UX Live Components

alexander-schranz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

Define requirement to the twig bundle for UX Component and UX Live Components. Currently installing ux component and live component fails. Both have a direct requirement to the `TwigBundle` as both require a service named `twig`.

Without defining this requirement the installation of the package will fail when not already have twig-bundle installed: https://github.com/schranz-php-recipes/symfony-recipes-php/actions/runs/3567748595/jobs/5995813798

Related PR: #313

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -29,7 +29,9 @@
         "php": ">=8.0",
         "symfony/property-access": "^5.4|^6.0",
         "symfony/serializer": "^5.4|^6.0",
-        "symfony/ux-twig-component": "^2.5"
+        "symfony/twig-bundle": "^5.4|^6.0",
+        "symfony/ux-twig-component": "^2.5",
+        "twig/twig": "^2.0|^3.0"
     },
     "require-dev": {
         "doctrine/annotations": "^1.0",
@@ -40,7 +42,6 @@
         "symfony/framework-bundle": "^5.4|^6.0",
         "symfony/phpunit-bridge": "^6.0",
         "symfony/security-csrf": "^5.4|^6.0",
-        "symfony/twig-bundle": "^5.4|^6.0",
         "symfony/validator": "^5.4|^6.0",
         "zenstruck/browser": "^1.2.0",
         "zenstruck/foundry": "^1.10"
--- a/src/TwigComponent/composer.json
+++ b/src/TwigComponent/composer.json
@@ -30,12 +30,12 @@
         "symfony/dependency-injection": "^5.4|^6.0",
         "symfony/event-dispatcher": "^5.4|^6.0",
         "symfony/property-access": "^5.4|^6.0",
+        "symfony/twig-bundle": "^5.4|^6.0",
         "twig/twig": "^2.0|^3.0"
     },
     "require-dev": {
         "symfony/framework-bundle": "^5.4|^6.0",
-        "symfony/phpunit-bridge": "^6.0",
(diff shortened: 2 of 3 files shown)


src/LiveComponent/composer.json
```diff
@@ -29,7 +29,9 @@
         "php": ">=8.0",
         "symfony/property-access": "^5.4|^6.0",
         "symfony/serializer": "^5.4|^6.0",
-        "symfony/ux-twig-component": "^2.5"
+        "symfony/twig-bundle": "^5.4|^6.0",
```
1ed, line 32: I think `TwigBundle` is not required here (nothing imported from `TwigBundle` in `src` only from `twig`).

alexander-schranz: Like the component it requires the specific service `twig` and so has a requirement to twig bundle itself. Because that is the bundle providing that service definition.

1ed: Ahh do you think about this one? https://github.com/symfony/ux/blob/4bd27efca1851bdaf1574c922dd8fa437229aa36/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php#L131

If, we add the bundle this can be removed https://github.com/symfony/ux/blob/4bd27efca1851bdaf1574c922dd8fa437229aa36/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php#L51-L55 

And there are other missing dependencies too, like dependency-injection, router, http-kernel, ... don't we should add those too?

alexander-schranz: Yes, that is what currently failing the installation inside a empty skeleton. I personally would if they are required, but not sure, wait what @weaverryan think about that.

kbond:
Thanks for looking into this!

I can't remember exactly how it is related but the failing CI is what I ran into when iterating on #313: the form component not being available. Does manually enabling the form component in the test kernel help?

alexander-schranz:
@kbond I did found out it its an issue in the Symfony core here a little inconsistent is between `TwigExtension.php` and `ExtensionPass.php`: https://github.com/symfony/symfony/issues/48373#issuecomment-1329736501 I'm not sure if we should add the `$container::willBeAvailable` to the compilerpass or even remove it from the TwigExtension.

kbond:
So the issue is there's no way to enable the twig form extension if `symfony/form` isn't in `composer.json`'s `require`?

I think that was the nuance that #313 solved: adding twig-bundle _and_ form to require-dev allowed it to be enabled.

Agree that this is a problem - I remember it took me some time to figure out what was going on.

alexander-schranz:
> I think that was the nuance that https://github.com/symfony/ux/pull/313 solved: adding twig-bundle and form to require-dev allowed it to be enabled.

Yes that is the case as then also the `form.extension` is not installed and so that is not failing in that case, but it does sadly so not define all require dependencies and fails if twig-bundle is not already installed. I stumble it over it in the recipes CI testing as the recipe update of symfony/ux components failed as it did not define this requirement.

I think I will target a pull request in symfony/symfony so that issue is fixed in the core that it can not appear again.

Kocal:

Thanks!

Status: Needs Work

PR #579, opened 2022-12-01
Update @hotwired/stimulus - v3.2.1

seb-jean:
@hotwired/stimulus 3.2.1 is released:
https://github.com/hotwired/stimulus/releases

### DIFF
--- a/src/Autocomplete/assets/package.json
+++ b/src/Autocomplete/assets/package.json
@@ -20,11 +20,11 @@
         }
     },
     "peerDependencies": {
-        "@hotwired/stimulus": "^3.0.0",
+        "@hotwired/stimulus": "^3.2.1",
         "tom-select": "^2.2.2"
     },
     "devDependencies": {
-        "@hotwired/stimulus": "^3.0.0",
+        "@hotwired/stimulus": "^3.2.1",
         "fetch-mock-jest": "^1.5.1",
         "tom-select": "^2.2.2"
     }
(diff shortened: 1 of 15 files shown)


weaverryan:
Thanks for this! But there's no real advantage to doing this. In the current state, `^3.0.0` still allows projects to use 3.2.1. But by changing this, anyone using an older version (e.g. 3.2.0) will get "peer dependency" warnings, even though UX works fine on 3.2.0. So, these constraints are meant to show the *full* range of which versions we support.

Related: we do need to fix https://github.com/symfony/flex/issues/938

stof:
Btw, due to the way the npm resolution works, bumping the min version of a peer dependency is a BC break that requires a major version (well, any kind of reducing the range actually, but it is rare to reduce it on the higher bound)

weaverryan:
> due to the way the npm resolution works, bumping the min version of a peer dependency is a BC break that requires a major version

Can you link to why that is the case? I don't understand it, but I would love to. thx :)

stof:
A peerDependency defines a constraint that must be satisfied by the package installed by the parent in the package tree. But it does not impact the choice of which version of the package is selected when resolving the dependencies of the parent.
That's exactly the same reason why changing the supported engine versions is a BC break in the npm ecosystem (the `engine` constraint works in a similar way than peer dependencies)

weaverryan:
In other words, if we suddenly bumped to 3.2.1 and were actually USING some new feature in 3.2.1, the user might update and... boom! Things are broken because our code breaks due to their "too old" Stimulus. They would get a warning, but we still "broke things" on a minor update. Got it - thx Stof!

PR #589, opened 2022-12-02
[Twig] Add a strategy for adding a Stimulus controller to a Twig component

### LINKED ISSUE #527: [Live] Adding a custom Stimulus controller to a live component
Sometimes you'll want to attach a stimulus controller to the root element of your live controller. Currently, you might try this:

```
<div
    {{ attributes.defaults({ 'data-controller: 'foo'})
>
```

But, I don't think that works, as I think we're always setting `data-controller` to `live`: https://github.com/symfony/ux/blob/2a4997497b3d1d17e7e731a23f6fa33067d14bc0/src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php#L102

Also, that's a pretty ugly syntax, especially if you need to pass in some values 😱 .

What about one of the following:

```
<div
    {{ attributes.stimulusController('foo', {}) }]

    {{ stimulus_controller('foo', {})|attributes }}

    Is this one even possible? Would `|merge` instead of "+" work?
    {{ stimulus_controller('foo', {}) + attributes }}
>
```

`stimulus_controller` returns a [StimulusControllersDto](https://github.com/symfony/webpack-encore-bundle/blob/main/src/Dto/StimulusControllersDto.php) which has a `toArray()` method.

weaverryan:
…onent root element

| Q             | A
| ------------- | ---
| Bug fix?      | yes-ish
| New feature?  | yes
| Tickets       | Fix #527
| License       | MIT

Hi! 

This allows adding a custom Stimulus controller to the root element of a component. The syntax is:

```
<div {{ attributes.add(stimulus_controller('my-controller', { someValue: 'foo' })) }}>
```

This is a compromise of having a nice syntax vs not duplicating code. The `attributes.add()` method is designed to work specifically with the `stimulus_*` functions, so it's name is kinda funny. But it feels way better than something like `attributes->stimulusController(stimulus_controller())`. Just doing `attributes.stimulusController('my-controller')` isn't possible at the moment, because it would require us to create a `StimulusControllersDto` and that, unfortunately, requires a Twig Environment for escaping purposes.

Cheers!

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -11,9 +11,10 @@
 
 namespace Symfony\UX\TwigComponent;
 
+use Symfony\WebpackEncoreBundle\Dto\AbstractStimulusDto;
+
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
-
  *
  * @immutable
  */
@@ -93,5 +94,22 @@ final class ComponentAttributes
         }
 
         return $clone;
+    }
+
+    public function add(AbstractStimulusDto $stimulusDto): self
+    {
+        $controllersAttributes = $stimulusDto->toArray();
+        $attributes = $this->attributes;
+
+        $attributes['data-controller'] = implode(' ', array_merge(
+            explode(' ', $attributes['data-controller']),
+            explode(' ', $controllersAttributes['data-controller'] ?? [])
+        ));
+        unset($controllersAttributes['data-controller']);
+
+        $clone = new self($attributes);
+
+        // add the remaining attributes for values/classes
+        return $clone->defaults($controllersAttributes);
     }
 }
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -475,6 +475,15 @@ The following hooks are available (along with the arguments that are passed):
 * ``loading.state:finished`` args ``(element: HTMLElement)``
 * ``model:set`` args ``(model: string, value: any, component: Component)``
 
+Adding a Stimulus Controller to your Component Root Element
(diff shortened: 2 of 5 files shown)


althaus:
Hey.... I've just stumbled upon this issue as the feature is already in the docs but not released yet.

Could you pack a new version with this change?

Cheers
Matthias

weaverryan:
Oooh, we need to add the versionadded annotation above the docs for this to make it clear! We’ll have as new release sometime soon… but not immediately.

weaverryan:
Added in sha: 8e5cab32f7150a0d6e7f37124aeecd5db8bdeeb8

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -94,4 +95,21 @@ public function without(string ...$keys): self
 
         return $clone;
     }
+
+    public function add(AbstractStimulusDto $stimulusDto): self
```
Kocal, line 99: Isn't a bit weird to have a method `add` that takes a `StimulusDto`? 
Shouldn't this method be named `withStimulus` or something like that?

weaverryan: Probably :). But it looked a bit terrible when used:

```
{{ attributes.withStimulus(stimulus_controller(...)) }}
```

The 2x Stimulus. But, we could potentially "open" this `add()` method later to accept other types of arguments, if we think of something.

althaus:
> Oooh, we need to add the versionadded annotation above the docs for this to make it clear! We’ll have as new release sometime soon… but not immediately.

Thanks for fixing the doc. I'm totally fine with using a non-release for trying things out. I just felt that integrating custom/legacy JS is quite common use case.

src/TwigComponent/src/ComponentAttributes.php
Kocal: Yeah, an array of key/values for example

PR #590, opened 2022-12-02
[Live] Adding 2 errors related to an invalid component set up

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | None
| License       | MIT

1) A better error if you have an uninitiailzed property, though there are still other cases of uninitialized properties that aren't handled yet.

2) A better error if there is a problem hydrating an object through the serializer.

Btw, I'm still not convinced we should use the serializer automatically to (de)hydrate properties. But anyways, even if we decide to NOT do this by default, it would still be something the user could easily enable... and  so having a nice message is important.

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -13,7 +13,9 @@ namespace Symfony\UX\LiveComponent;
 
 use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
 use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
+use Symfony\Component\PropertyAccess\Exception\UninitializedPropertyException;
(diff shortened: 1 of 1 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -153,7 +159,24 @@ public function hydrate(object $component, array $data, string $componentName):
...
+                        $exception->getMessage()
+                    );
+
+                    // unless the data is gigantic, include it in the error to help
+                    if (strlen($json) < 1000) {
+                        $message .= sprintf(' The data sent from the frontend was: %s', $json);
+                    }
+
+                    throw new \LogicException($message, 0, $exception);
+                }
```
weaverryan, line 179: An example of this:

Screen Shot 2022-12-02 at 3 42 06 PM

I wish I could add line breaks, but at least all the info is there.

PR #602, opened 2022-12-08
Making all events bubble - affects Chartjs, Cropperjs, Dropzone, LazyImage, Swup

### LINKED ISSUE #30: Allow events to bubble
Currently, the `*:connect` events are dispatched as non-bubbling, making them incompatible with event delegation. I think it would make sense to let them bubble instead.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #30
| License       | MIT

Long-overdue fix for #30. This is a small behavior change, but I consider it a bug fix. The only way that you could be affected by this change in practice is if you have one of the UX components listed *inside* of itself - e.g. a chart inside a chart, so now the existing listener on the outer chart is fired twice. But I don't think that makes any practical sense for any of these components.

Cheers!

### DIFF
--- a/src/Autocomplete/assets/dist/controller.d.ts
+++ b/src/Autocomplete/assets/dist/controller.d.ts
@@ -28,5 +28,6 @@ export default class extends Controller {
     disconnect(): void;
     get selectElement(): HTMLSelectElement | null;
     get formElement(): HTMLInputElement | HTMLSelectElement;
+    private dispatchEvent;
     get preload(): string | boolean;
 }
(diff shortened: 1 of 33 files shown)


norkunas:
Just interested: why not using `dispatchEvent` method that exists in stimulus controllers and implementing a custom method for this? :)

weaverryan:
Lol, good question. Two reasons: 1) Until about a week ago I didn't know this existed! And, related, (2), this `dispatch()` method was added in Turbo 3, so many of the packages predate that. But, indeed, it looks like we can start using it now by passing `false` for the prefix option. Good call

weaverryan:
This is now ready for a quick look-over - all of the new ways to dispatch the events *should* result in the same event name & target as before, unless I've done something silly or copy-pasta'ed something poorly.

PR #605, opened 2022-12-11
new BootstrapTable component

### LINKED ISSUE #230: New component: Bootstrap Table?
I'm using bootstrap table for my projects to Ajax load table data and display it with formatting and additional functions (print, export, search, show/hide columns, any many more):
https://bootstrap-table.com

Is it possible to maybe add this as new Symfony UX component? I think it would fit great in the eco system, because symfony itself could provide everything needed and javascript can add some additional possibilities.

Just an idea. Let me know what you think about it. And if interested I can sure assist in implementing it.

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #230
| License       | MIT

This PR add a new Bootstrap Table component.

To generate that:

Capture d’écran 2022-12-11 à 23 47 09

you only need this few line of code:

```php

class HomeController extends AbstractController
{
    #[Route("/", name: "app_index")]
    public function index(TableBuilderInterface $tableBuilder, ChartBuilderInterface $chartBuilder): Response
    {
        $table = $tableBuilder
            ->addColumns([
                ['title' => 'ID', 'field' => 'id'],
                ['title' => 'Pseudo', 'field' => 'pseudo'],
                ['title' => 'Age', 'field' => 'age'],
                ['title' => 'Gender', 'field' => 'gender']
            ])
            ->addData([
                ['id' => 1, 'pseudo' => 'Bob', 'age' => 23, 'gender' => 'M'],
                ['id' => 2, 'pseudo' => 'Kitty', 'age' => 36, 'gender' => 'F'],
                ['id' => 3, 'pseudo' => 'Spongy', 'age' => 34, 'gender' => 'M'],
                ['id' => 4, 'pseudo' => 'Slurp', 'age' => 20, 'gender' => 'F'],
                ['id' => 5, 'pseudo' => 'Zoom', 'age' => 29, 'gender' => 'M'],
            ])
            ->addOptions([
                'pagination'=> 'true',
                'search' => 'true'
            ])
            ->createTable()
        ;

        return $this->render('index.html.twig', [
            'table' => $table,
        ]);
    }
}
```

and in your template:

```twig

{% block body %}
    <div style="width: 100%; padding: 100px">
        {{ render_table(table) }}
    </div>
{% endblock %}

```

After few update the Builder now support pagination:

```php

...
        $table = $tableBuilder
            ->addColumns([
                ['title' => 'ID', 'field' => 'id'],
                ['title' => 'Pseudo', 'field' => 'pseudo'],
                ['title' => 'Age', 'field' => 'age'],
                ['title' => 'Gender', 'field' => 'gender']
            ])
            ->addData([
                ['id' => 1, 'pseudo' => 'Bob', 'age' => 23, 'gender' => 'M'],
                ['id' => 2, 'pseudo' => 'Kitty', 'age' => 36, 'gender' => 'F'],
                ['id' => 3, 'pseudo' => 'Spongy', 'age' => 34, 'gender' => 'M'],
                ['id' => 4, 'pseudo' => 'Slurp', 'age' => 20, 'gender' => 'F'],
                ['id' => 5, 'pseudo' => 'Zoom', 'age' => 29, 'gender' => 'M'],
            ])
            ->addOptions([
                'pagination'=> 'true',
                'search' => 'true'
            ])
          ->enablePagination()
          ->itemPerPage(5)
          ->sort('pseudo', 'ASC')
          ->search('23')
         ->createTable()
        ;
   
```

### DIFF
--- a/src/BootstrapTable/assets/src/controller.js
+++ b/src/BootstrapTable/assets/src/controller.js
@@ -1,0 +1,5 @@
+'use strict';
+
+import 'bootstrap/dist/css/bootstrap.min.css'
+import 'bootstrap';
+import 'bootstrap-table';
--- a/src/BootstrapTable/assets/src/controller.js
+++ b/src/BootstrapTable/assets/src/controller.js
@@ -1,5 +1,0 @@
-'use strict';
-
-import 'bootstrap/dist/css/bootstrap.min.css'
-import 'bootstrap';
-import 'bootstrap-table';
--- a/src/BootstrapTable/src/Builder/TableBuilder.php
+++ b/src/BootstrapTable/src/Builder/TableBuilder.php
@@ -41,7 +41,6 @@ class TableBuilder implements TableBuilderInterface
         return $this;
     }
 
-
     public function addOptions(array $options): TableBuilderInterface
     {
         $this->options = $options;
--- a/src/BootstrapTable/src/Builder/TableBuilder.php
+++ b/src/BootstrapTable/src/Builder/TableBuilder.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\BootstrapTable\Builder;
 
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\UX\BootstrapTable\Model\Table;
 
 /**
@@ -18,14 +19,18 @@ use Symfony\UX\BootstrapTable\Model\Table;
  */
 class TableBuilder implements TableBuilderInterface
 {
-    public $data = [];
-    public $columns = [];
-    public $options = [];
+    public const SORT_ASC = 'ASC';
+    public const SORT_DESC = 'DESC';
 
-    public function createTable(): Table
-    {
-        return new Table($this->data, $this->columns, $this->options);
-    }
+    private $data = [];
+    private $columns = [];
+    private $options = [];
+    private $sort;
+    private $search = '';
+    private $pagination = false;
+    private $itemsPerPage = 5;
+    private $currentPage = 0;
+    private $request;
 
     public function addData(array $data): TableBuilderInterface
     {
@@ -46,5 +51,118 @@ class TableBuilder implements TableBuilderInterface
         $this->options = $options;
 
         return $this;
+    }
+
+    public function setSort(string $field, string $direction): TableBuilderInterface
+    {
+        if (self::SORT_ASC !== $direction && self::SORT_DESC !== $direction) {
+            throw new \RuntimeException(sprintf('the direction should be equal to %s or %s', self::SORT_ASC, self::SORT_DESC));
+        }
+
+        $this->sort = [$field, $direction];
+
+        return $this;
+    }
+
+    public function setSearch(string $search): TableBuilderInterface
+    {
+        $this->search = $search;
+
+        return $this;
+    }
+
+    public function enablePagination(): TableBuilderInterface
+    {
+        $this->pagination = true;
+
+        return $this;
+    }
+
+    public function setItemPerPage(int $itemsPerPage): TableBuilderInterface
+    {
+        $this->itemsPerPage = $itemsPerPage;
+
+        return $this;
+    }
+
+    public function setCurrentPage(int $currentPage): TableBuilderInterface
+    {
+        $this->currentPage = $currentPage;
+
+        return $this;
+    }
+
+    public function handleRequest(Request $request): TableBuilderInterface
+    {
+        $this->request = $request;
+
+        return $this;
+    }
+
+    public function createTable(): Table
+    {
+        $this->init();
+
+        $data = $this->data;
+
+        if ('' !== $this->search) {
+            foreach ($data as $key => $row) {
+                if (!\in_array($this->search, $row, false)) {
+                    unset($data[$key]);
+                }
+            }
+            $data = array_values($data);
+        }
+
+        if (null !== $this->sort) {
+            usort($data, function ($a, $b) {
+                return $this->sortItems($a, $b);
+            });
+        }
+
+        if ($this->pagination) {
+            $data = \array_slice($data, $this->currentPage * $this->itemsPerPage, $this->itemsPerPage);
+        }
+
+        return new Table($data, $this->columns, $this->options);
+    }
+
+    private function init(): void
+    {
+        /** @var Request $request */
+        $request = $this->request;
+
+        if (null === $request) {
+            return;
+        }
+
+        $page = $request->get('page');
+
+        if (null !== $page) {
+            $this->currentPage = (int) $page;
+        }
+
+        $search = $request->get('search');
+
+        if (null !== $search) {
+            $this->setSearch($search);
+        }
+
+        $sort = $request->get('sort');
+
+        if (null !== $sort) {
+            $this->sort = explode(',', $sort, 2);
+        }
+    }
+
+    private function sortItems($a, $b): bool
+    {
+        [$field, $direction] = $this->sort;
+
+        if (self::SORT_DESC === $direction) {
+            return $a[$field] < $b[$field];
+        }
+
+        return $a[$field] > $b[$field];
     }
 }
--- a/src/BootstrapTable/src/Builder/TableBuilder.php
+++ b/src/BootstrapTable/src/Builder/TableBuilder.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\BootstrapTable\Builder;
+
+use Symfony\UX\BootstrapTable\Model\Table;
+
+/**
+ * @author Mathéo Daninos <matheo.daninos@gmail.com>
+ */
+class TableBuilder implements TableBuilderInterface
+{
+    public $data = [];
+    public $columns = [];
+    public $tableAttributes = [];
+    public $columnsAttributes = [];
+
+    public function createTable(): Table
+    {
+        return new Table($this->data, $this->columns, $this->tableAttributes, $this->columnsAttributes);
+    }
+
+    public function addData(array $data): TableBuilderInterface
+    {
+        $this->data = $data;
+
+        return $this;
+    }
+
+    public function addColumns(array $columns): TableBuilderInterface
+    {
+        $this->columns = $columns;
+
+        return $this;
+    }
+
+    public function addTableAttributes(array $tableAttributes): TableBuilderInterface
+    {
+        $this->tableAttributes = $tableAttributes;
+
+        return $this;
+    }
+
+    public function addColumnsAttributes(array $columnsAttributes): TableBuilderInterface
+    {
+        $this->columnsAttributes = $columnsAttributes;
+
+        return $this;
+    }
+}
--- a/src/BootstrapTable/src/Builder/TableBuilder.php
+++ b/src/BootstrapTable/src/Builder/TableBuilder.php
@@ -20,12 +20,11 @@ class TableBuilder implements TableBuilderInterface
 {
     public $data = [];
     public $columns = [];
-    public $tableAttributes = [];
-    public $columnsAttributes = [];
+    public $options = [];
 
     public function createTable(): Table
     {
-        return new Table($this->data, $this->columns, $this->tableAttributes, $this->columnsAttributes);
+        return new Table($this->data, $this->columns, $this->options);
     }
 
     public function addData(array $data): TableBuilderInterface
@@ -42,16 +41,10 @@ class TableBuilder implements TableBuilderInterface
         return $this;
     }
 
-    public function addTableAttributes(array $tableAttributes): TableBuilderInterface
-    {
-        $this->tableAttributes = $tableAttributes;
 
-        return $this;
-    }
-
-    public function addColumnsAttributes(array $columnsAttributes): TableBuilderInterface
+    public function addOptions(array $options): TableBuilderInterface
     {
-        $this->columnsAttributes = $columnsAttributes;
+        $this->options = $options;
 
         return $this;
     }
--- a/src/BootstrapTable/src/DependencyInjection/BootstrapTableExtension.php
+++ b/src/BootstrapTable/src/DependencyInjection/BootstrapTableExtension.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\BootstrapTable\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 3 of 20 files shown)


src/BootstrapTable/assets/src/controller.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+import 'bootstrap/dist/css/bootstrap.min.css'
+import 'bootstrap';
```
Kocal: I don't think the end-user wants to import the whole Bootstrap CSS/JS code to render a table.

weaverryan: This was also the first part of the PR that jumped out at me. A few things:

1) My guess is that we don't need to import bootstrap from inside here. If `bootstrap-table` uses `bootstrap`, IT will import it internally.

2) We also shouldn't import `bootstrap.min.css`. That'll be up to the user to make sure they're importing this, or importing the parts they want :).

3) We probably don't want to rely on the "data attributes" way of initializing the functionality. We should initialize on `this.element` directly.

WebMamba: I agree with both of you on the import of the css and js. I put that here more for the demo. But I am not sure about your third point @weaverryan. Do you mean instead of using data attributes we should generate everything by javascript? And if yes why?

simondaigre:
A Table component is the feature I need the most for my Symfony projects. But I dont think using Bootstrap Table is a good approach here, while we have a powerful Live component for live reloading, and search/pagination features with Autocomplete component.

Also Bootstrap Table requires Jquery, which is not great. 

I dont know if @weaverryan has plans about a Table component, but it would be great to find a better approach for this feature IMHO.

weaverryan:
Hey @simondaigre!

I very much appreciate having this exact conversation about what people want and need. The idea behind this would be to allow for a UX table component that is NOT built on live components - it's something you can just use. But let's talk about some options:

A) If we use this approach, one missing piece, I think, is pagination, searching and ordering. But I think that could be accomplished pretty easily - e.g. the Ajax calls are made right back to the current URL and, in that controller, you have a method that accepts the `Request` object and automatically reads off the sort, search & ordering data. Then there'd be some way to have the user hook that all up to their query.

B) The dependency on jQuery is a bummer, I agree with that. The classic DataTables - https://datatables.net/ - has that same downside. I wish there was some nice, out-of-the-box solution that didn't include jQuery. 

I think, if we solved (A), then this would be useful for many people. But I'd love for people to 👍  or 👎 that opinion of mine. What I love about this PR from @WebMamba is that it's getting this conversation going :). 

If we DO keep going with this PR, then you're right that, for a certain percentage of people, needing to introduce jQuery will be a deal breaker. Can we solve this instead with Live Components? Absolutely! And you could build that by hand today, and it would make a great example for ux.symfony.com. What's not clear to me yet what tools LiveComponents can offer to make creating a "DataTable" like setup really easy. Maybe... there is a core "UX Table" package that creates a `TableBuilder` system that hooks up with Doctrine ORM (or optionally other things)? And then that could be used potentially to help build a BootstrapTable or you could use it in your LiveComponent to build and manage a table?

Anyways, I'd love to know what thoughts you have - or anyone else :) 

Btw, here is a well-known Laravel package https://github.com/rappasoft/laravel-livewire-tables - it has an interesting API for describing the tables.

src/BootstrapTable/assets/src/controller.js
weaverryan: Yes! Exactly! I don't think that relying on the `data-` attributes will actually work if that HTML is loaded later via Ajax onto the page. That's why, with all the other UX components, we leverage Stimulus for this. So, on the table, you would output something like `data-controller="symfony--ux-bootstrap-table--controller"` to connect to this controller.  You would also pass in the data via a "value" (just like chart.js does). Then, in this controller, in the `connect()` method, you would initialize BootstrapTable programmatically - e.g.

```js
$(this.element).bootstrapTable(this.tableDataValue).
```

simondaigre:
Live component could be useful if we choose to render the table on server side.

With this approach, we can use Twig to customize the table or define special columns (a Badge column, etc..) without writing JS. 
Pagination / sort and search could be done on server side also. 

And last, we will not be "tied" with Bootstrap, for devs who wants to use another CSS library. I saw Bootstrap Table is compatible with many libs, but bootstrap seems always included in this PR.

Dannebicque:
Hello,

I am very interested in a "UX component" for table. 

I tried several solutions like DataTable, ... and then I decided not to depend on jQuery anymore (more by challenge than real need). I tried to develop such a component which would be based on Symfony to generate my table somewhat automatically on the principle of a FormBuilder, exactly like the idea of ​​this PR. I searched, and I found this repository https://github.com/acantepie/umbrella-corebundle. 

This bundle offers a back-office, and above all has a table management system in this idea. The downside of this solution is that it uses DataTable (which is a great library, but depends on jQuery), so not fully my answer.

Based on this idea, I started from the principle and redid the JS part without jQuery.
Recently I read Ryan's tweet, and decided to try to extract this part of my application and make my very first symfony bundle :) => https://github.com/Dannebicque/TableBundle (Be indulgent this is my very first bundle)

Like in the first bundle the idea is to create a table like a form and simply display in twig like a form.

Example :

```php
public function index(Request $request): Response
    {
        $this->denyAccessUnlessGranted('MINIMAL_ROLE_ASS', $this->getDepartement());

        $table = $this->createTable(DateTableType::class, [
            'departement' => $this->getDepartement(),
        ]);
        $table->handleRequest($request);

        if ($table->isCallback()) {
            return $table->getCallbackResponse();
        }

        return $this->render('administration/date/index.html.twig',
            [
                'table' => $table,
            ]);
    }
```

The DateTableType

```php
<?php
/*
 * Copyright (c) 2022. | David Annebicque | IUT de Troyes  - All Rights Reserved
 * @file /Users/davidannebicque/Sites/intranetV3/src/Table/DateTableType.php
 * @author davidannebicque
 * @project intranetV3
 * @lastUpdate 18/11/2022 08:54
 */

namespace App\Table;

use DavidAnnebicque\TableBundle\Widget\Type\ExportDropdownType;
use DavidAnnebicque\TableBundle\Widget\Type\RowDeleteLinkType;
use DavidAnnebicque\TableBundle\Widget\Type\RowDuplicateLinkType;
use DavidAnnebicque\TableBundle\Widget\Type\RowEditLinkType;
use DavidAnnebicque\TableBundle\Widget\Type\RowShowLinkType;
use DavidAnnebicque\TableBundle\Widget\WidgetBuilder;
...

class DateTableType extends TableType
{
    private ?Departement $departement = null;

    public function __construct(private readonly CsrfTokenManagerInterface $csrfToken)
    {
    }

    public function buildTable(TableBuilder $builder, array $options): void
    {
        $this->departement = $options['departement'];

        $builder->addFilter('search', SearchType::class);
        $builder->addFilter('from', DatePickerType::class, [
            'input_prefix_text' => 'Du',
        ]);
        $builder->addFilter('to', DatePickerType::class, [
            'input_prefix_text' => 'Au',
        ]);

        $builder->addWidget('export', ExportDropdownType::class, [
            'route' => 'administration_date_export',
        ]);

        $builder->addColumn('libelle', PropertyColumnType::class,
            ['label' => 'table.libelle', 'translation_domain' => 'messages']);
        $builder->addColumn('lieu', PropertyColumnType::class,
            ['label' => 'table.lieu', 'translation_domain' => 'messages']);
        $builder->addColumn('typeDestinataire', BadgeColumnType::class,
            ['label' => 'table.typeDestinataire', 'translation_domain' => 'messages']);
        $builder->addColumn('semestres', SemestreColumnType::class,
            ['label' => 'table.semestres', 'translation_domain' => 'messages']);
        $builder->addColumn('dateDebut', DateColumnType::class, [
            'order' => 'DESC',
            'format' => 'd/m/Y',
            'label' => 'table.dateDebut',
            'translation_domain' => 'messages',
        ]);
        $builder->addColumn('heureDebut', DateColumnType::class, [
            'order' => 'DESC',
            'format' => 'H:i',
            'label' => 'table.heureDebut',
            'translation_domain' => 'messages',
        ]);

        $builder->addColumn('links', WidgetColumnType::class, [
            'build' => function (WidgetBuilder $builder, Date $s) {
                $builder->add('duplicate', RowDuplicateLinkType::class, [
                    'route' => 'administration_date_duplicate',
                    'route_params' => ['id' => $s->getId()],
                    'xhr' => false,
                ]);
                $builder->add('show', RowShowLinkType::class, [
                    'route' => 'administration_date_show',
                    'route_params' => [
                        'id' => $s->getId(),
                    ],
                    'xhr' => false,
                ]);
                $builder->add('edit', RowEditLinkType::class, [
                    'route' => 'administration_date_edit',
                    'route_params' => [
                        'id' => $s->getId(),
                    ],
                    'xhr' => false,
                ]);
                $builder->add('delete', RowDeleteLinkType::class, [
                    'route' => 'administration_date_delete',
                    'route_params' => ['id' => $s->getId()],
                    'attr' => [
                        'data-csrf' => $this->csrfToken->getToken('delete'.$s->getId()),
                    ],
                ]);
            },
        ]);
        $builder->setLoadUrl('administration_date_index');

        $builder->useAdapter(EntityAdapter::class, [
            'class' => Date::class,
            'fetch_join_collection' => false,
            'query' => function (QueryBuilder $qb, array $formData) {
                $qb->innerJoin('e.semestres', 'c')// récupération de la jointure dans la table dédiée
                ->innerJoin(Semestre::class, 's', 'WITH', 'c.id = s.id')
                    ->innerJoin(Annee::class, 'a', 'WITH', 's.annee = a.id')
                    ->innerJoin(Diplome::class, 'd', 'WITH', 'a.diplome = d.id')
                    ->where('d.departement = :departement')
                    ->setParameter('departement', $this->departement->getId());

                if (isset($formData['search'])) {
                    $qb->andWhere('LOWER(e.libelle) LIKE :search');
                    $qb->orWhere('LOWER(e.lieu) LIKE :search');
                    $qb->setParameter('search', '%'.$formData['search'].'%');
                }

                if (isset($formData['from'])) {
                    $qb->andWhere('e.updated >= :from');
                    $qb->setParameter('from', $formData['from']);
                }

                if (isset($formData['to'])) {
                    $qb->andWhere('e.updated <= :to');
                    $qb->setParameter('to', $formData['to']);
                }
            },
        ]);
    }

    {
        parent::configureOptions($resolver);

        $resolver->setDefaults([
            'orderable' => true,
            'departement' => null,
        ]);
    }
}

```

And the view

```twig
 {{ table_render(table) }}
```

And the result (sorry it's a french application): 

Capture d’écran 2022-12-15 à 17 35 37

Now it works fine, with badges, linked columns, filters,... But reading this post and thinking from the Tweet, I think Stimulus/LiveTwig... can probably improve the UX of this solution a lot.

If I can help or share on this subject, it will be with great pleasure.

David

src/BootstrapTable/assets/src/controller.js
WebMamba: Ok fine! I commit that this week-end

WebMamba:
Thanks all! For your review on this. This PR is here to procure an easy and fast way to render a boostrap table. It's been asked by the community multiple times. The idea here is to have something really easy to set up. More for prototypes, or small projects I guess. We already know this PR does not respond to all, and this is why I call this component BootstrapTableComponent and not just DataTableComponent. We can have different approaches to living side by side.
But I am pretty happy with the debate raised here because we are now sure there is a need for DataTable without jQuery based on LiveComponent. I will be happy to contribute to it! Here is a small POC I did about it just before proposing this PR (https://github.com/WebMamba/ux_datatable)

WebMamba:
@weaverryan I made a big update on this PR to use the js way 😁

wannevancamp:
This PR and POC look promising. I'm very interested in a Table builder UX component. 
In the Laravel ecosystem there is a powerful 'Table builder' component that the UX component can possibly be inspired by
https://filamentphp.com/docs/2.x/tables

WebMamba:
Hey all! I just made an update to support pagination, sorting, and searching directly from the builder. By working on it, I realized that is completely possible to do what @weaverryan said here: 

>  there is a core "UX Table" package that creates a TableBuilder system that hooks up with Doctrine ORM (or optionally other things)? And then that could be used potentially to help build a BootstrapTable or you could use it in your LiveComponent to build and manage a table?

We can build a table Builder, and then from this builder procure different approaches for the front (LiveComponent, BootstrapTable, or any front-end library we want).

So if we choose this approach we need to: 

A) create a builder that returns a generic table that we are able to use with any front-end library
B) think about how we organize everything to ease the integration of new adapters for new js libraries

src/BootstrapTable/src/Builder/TableBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+
+    private $data = [];
+    private $columns = [];
+    private $options = [];
+    private $sort;
+    private $search = '';
+    private $pagination = false;
+    private $itemsPerPage = 5;
+    private $currentPage = 0;
+    private $request;
```
skigun, line 33: Can be typehinted 

```php
    /** @var Request $request */
    private $request;
```

src/BootstrapTable/src/Builder/TableBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+        if ($this->pagination) {
+            $data = \array_slice($data, $this->currentPage * $this->itemsPerPage, $this->itemsPerPage);
+        }
+
+        return new Table($data, $this->columns, $this->options);
+    }
+
+    private function init(): void
+    {
+        /** @var Request $request */
```
skigun, line 132: Can be removed

src/BootstrapTable/src/Builder/TableBuilder.php
```diff
@@ -0,0 +1,168 @@
...
+                }
+            }
+            $data = array_values($data);
+        }
+
+        if (null !== $this->sort) {
+            usort($data, function ($a, $b) {
+                return $this->sortItems($a, $b);
+            });
+        }
```
skigun, line 121: What about removing the `sortItems` method and keep something like
```php 
       if (null !== $this->sort) {
            [$field, $direction] = $this->sort;
            self::SORT_ASC === $direction ? asort($data) : arsort($data);
        }
```

src/BootstrapTable/src/DependencyInjection/BootstrapTableExtension.php
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @author Mathéo Daninos <mathéo.daninos@gmail.com>
+ */
+class BootstrapTableExtension extends Extension
+{
+    public function load(array $configs, ContainerBuilder $container)
+    {
+        $container
+            ->setDefinition('bootstrapTable.builder', new Definition(TableBuilder::class))
```
skigun, line 32: snakeCase for alias/id definition

src/BootstrapTable/src/DependencyInjection/BootstrapTableExtension.php
```diff
@@ -0,0 +1,50 @@
...
+            ->setPublic(false)
+        ;
+
+        $container
+            ->setAlias(TableBuilderInterface::class, 'bootstrapTable.builder')
```
skigun, line 37: snakeCase for alias/id definition

src/BootstrapTable/src/Builder/TableBuilder.php
WebMamba: The sort is more complex since you sort base on a specific field

PR #610, opened 2022-12-14
[ChartJs] chart.js v4 support

### LINKED ISSUE #607: [ChartJs] Add support for chart.js:^4.0
Looks like it won't have a big effect on the ux-component? But probably best to leave that to the experts ;)

- https://www.chartjs.org/docs/latest/migration/v4-migration.html
- https://github.com/chartjs/Chart.js/releases

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #607
| License       | MIT

**Current status:**
- Separate controllers per version (`3` and `4`)
  - We could also add specific support for the `3.9` version, see #427
- Added separate test controllers per version
- Added support for running `low`/`high` dependency tests via GH workflow

**Todo**
- [x] Fix importing `chart.js` in the `chart_v4_controller.ts` and fix the test-suite
- [ ] Test specific low/high dependency dir/paths 
  - on `low-deps` (`chart.js:^3.4.1`); only test `chart_v3_controller.test.ts` and on `high-deps` (`chart.js:^4.0`); only test `chart_v4_controller.test.ts`. Otherwise v3 deps will error on v4 controller and vice versa.
- [ ] Files in `dist` directory should also be updated? Or is that part of the Webpack loader magic :)
- [ ] Cleanup/reuse (DRY) code in the test controllers?

### DIFF
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -7,11 +7,19 @@
     "types": "dist/controller.d.ts",
     "symfony": {
         "controllers": {
-            "chart": {
-                "main": "dist/controller.js",
+            "chart_v3": {
+                "main": "dist/chart_v3_controller.js",
+                "name": "symfony--ux-chartjs--chart",
                 "webpackMode": "eager",
                 "fetch": "eager",
                 "enabled": true
+            },
+            "chart_v4": {
+                "main": "dist/chart_v4_controller.js",
+                "name": "symfony--ux-chartjs--chart",
+                "webpackMode": "eager",
+                "fetch": "eager",
+                "enabled": false
             }
         }
     },
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -30,7 +30,7 @@
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
         "@types/chart.js": "^2.9.34",
-        "chart.js": "^4.0",
+        "chart.js": "^3.4.1 <3.9 || ^4.0",
         "jest-canvas-mock": "^2.3.0",
         "resize-observer-polyfill": "^1.5.1"
     }
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -17,7 +17,7 @@
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1 || ^4.0"
+        "chart.js": "^3.4.1 <3.9 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -22,7 +22,7 @@
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
         "@types/chart.js": "^2.9.34",
-        "chart.js": "^3.4.1 <3.9",
+        "chart.js": "^4.0",
         "jest-canvas-mock": "^2.3.0",
         "resize-observer-polyfill": "^1.5.1"
     }
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -17,7 +17,7 @@
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1"
+        "chart.js": "^3.4.1 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
--- a/src/Chartjs/assets/src/chart_v4_controller.ts
+++ b/src/Chartjs/assets/src/chart_v4_controller.ts
@@ -10,7 +10,7 @@
 'use strict';
 
 import AbstractChartController from './abstract_controller';
-import Chart from 'chart.js/auto';
+import Chart from 'chart.js/auto/auto.cjs';
 import { ChartConfiguration, ChartItem } from 'chart.js';
 
 export default class extends AbstractChartController {
--- a/src/Chartjs/assets/src/chart_v4_controller.ts
+++ b/src/Chartjs/assets/src/chart_v4_controller.ts
@@ -1,0 +1,20 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import AbstractChartController from './abstract_controller';
+import Chart from 'chart.js/auto';
+import { ChartConfiguration, ChartItem } from 'chart.js';
+
+export default class extends AbstractChartController {
+    createChart(canvasContext: ChartItem, payload: ChartConfiguration): any {
+        return new Chart(canvasContext, payload);
+    }
+}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -235,13 +235,39 @@ jobs:
               working-directory: src/Notify
               run: php vendor/bin/simple-phpunit
 
-    tests-js:
+    tests-js-low-deps:
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: actions/checkout@master
+
+            -   name: Get yarn cache directory path
+                id: yarn-cache-dir-path
+                run: echo "dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
+
+            -   uses: actions/cache@v2
+                id: yarn-cache
+                with:
+                    path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+                    key: ${{ runner.os }}-low-yarn-${{ hashFiles('**/package.json') }}
+                    restore-keys: |
+                        ${{ runner.os }}-low-yarn-
+
+            -   name: Force Lowest Dependencies
+                run: node ./src/Chartjs/assets/scripts/force-lowest-dependencies.js
+
+            -   run: yarn
+
+            -   run: yarn test
+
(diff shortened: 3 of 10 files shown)


simonsolutions:
Does this line also has to be changed?

https://github.com/symfony/ux/blob/23639956f56ace782495f69f9b483ea565222673/src/Chartjs/assets/package.json#L23

evertharmeling:
Good catch! Guess that could even be set to the latest version only (`"chart.js": "^4.0"`) right?

simonsolutions:
> Good catch! Guess that could even be set to the latest version only (`"chart.js": "^4.0"`) right?

I guess so too, in development the latest version of chart.js should be used, because that is the only where the maintainer of the chart.js repository are giving support for.

evertharmeling:
😕 the fails don't seem directly related to the changes...?

evertharmeling:
Oh wait it looks like it is related, the way `'chart.js/auto'` is exposed (used in https://github.com/symfony/ux/blob/2.x/src/Chartjs/assets/src/controller.ts#L13) changed.

Error:
- https://github.com/symfony/ux/actions/runs/3696243151/jobs/6259682636

Change
- v3.9: https://github.com/chartjs/Chart.js/blob/3.9/auto/auto.js
- v4: https://github.com/chartjs/Chart.js/blob/master/auto/auto.js

However not sure how to fix supporting both versions... (my js/ts knowledge lacks 😅)

weaverryan:
Oh yes, I forgot about that! So, indeed, the way the chart is imported changed in 3.9 and we couldn’t figure out how to get it working :/. https://github.com/symfony/ux/pull/427. So, you were showing the difference between 3.9 and 4.0, but even 3.8 to 3.9 changed.

Anyways, we can’t drop support for Chart.js 3 without triggering a new major version of UX. The correct solution probably involves having 2 controllers in the bundle: one for 3.8 and lower and one for 3.9 and above (or maybe one for 3.8 and lower and another for 4.0 and above iv 3.9 and 4.0 are actually different as you mentioned). In theory, I think that would be fairly easy to do.

Then, in the `package.json` file for the UX package, we would need to expose TWO controllers: one would be enabled and the other disabled by default. One trick, however, is that we need the "name" of each controller to be the same. So, for the new controller (the one that supports Chart.js 4), we would use the `name` key (example https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/package.json#L12) to set the name to the "normal" name - which, iirc, would be `symfony--ux-chartjs--chart`.

So, if possible, I think this is what we should do :). We MIGHT, once this is done, be able to make some tweak in https://github.com/symfony/stimulus-bridge so that, instead of enabling 1 controller and disabling another controller in `controllers.json`, we choose which controller to use dynamically based on the version of Chart.js the user has installed. But, let's get the work done here first and then investigate that, which involves some Webpack loader magic :).

evertharmeling:
No worries, thanks for the elaborate explanation. Let me put on my superman suit and try to jump in this twilight zone! 😬 🦸🏼‍♂️

evertharmeling:
Something like this... 😟😰 I'm kinda lost at the import part as the (chart.js) [docs:3.8.2](https://www.chartjs.org/docs/3.8.2/getting-started/integration.html) / [docs:v3.9.1](https://www.chartjs.org/docs/3.9.1/getting-started/integration.html) / [docs:4.0](https://www.chartjs.org/docs/latest/getting-started/integration.html) just refer to `import Chart from 'chart.js/auto';` which should work for all versions...

- To simplify for now I reverted the exclusion of 3.9, as if you want to use the latest version you could just head over to 4.0
- I added a `AbstractChartController` to keep it a bit DRY

src/Chartjs/assets/package.json
```diff
@@ -7,22 +7,30 @@
...
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1"
+        "chart.js": "^3.4.1 <3.9 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
         "@types/chart.js": "^2.9.34",
-        "chart.js": "^3.4.1 <3.9",
+        "chart.js": "^4.0",
```
weaverryan: Make this look like `peerDependencies`. We don't do it at the moment, but we might/should add something to our CI to test "low" dependency versions as well as high (we do this in webpack-encore).

src/Chartjs/assets/src/chart_v4_controller.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import AbstractChartController from './abstract_controller';
+import Chart from 'chart.js/auto';
+import { ChartConfiguration, ChartItem } from 'chart.js';
+
+export default class extends AbstractChartController {
+    createChart(canvasContext: ChartItem, payload: ChartConfiguration): any {
+        return new Chart(canvasContext, payload);
+    }
+}
```
weaverryan, line 20: So, at the moment, both the v3 and v4 controllers are identical - is that correct? Or am I missing a detail?

weaverryan (commented):
This is excellent work... though I feel like we're still working so that we can ultimately discover the problem / solution for what's different between v3 and v4 :p. 

Can you update the tests - they're still importing the old `controller.ts` file, so we can't see the actual failure. Btw, are you able to test this in a local project? Is the v4 version working?

simonsolutions:
I upgraded chart.js to the 4.x version with Symfony UX component, and so far no issues and problems with chart rendering.

src/Chartjs/assets/src/chart_v4_controller.ts
evertharmeling: Yes, that's right they do :) Because (as commented above) I was a bit stuck on the importing `chart.js` part, as the documentation states for all versions (`v3.8.2`, `v3.9.1` and `v4.0`) the import is the same (just `import Chart from 'chart.js/auto';`),  but the [tests](https://github.com/symfony/ux/actions/runs/3696243151/jobs/6259682636) in this PR gave an error on the import... 😕 so it was still a WIP :)

evertharmeling:
> I upgraded chart.js to the 4.x version with Symfony UX component, and so far no issues and problems with chart rendering.

Yeah, got it working too, however the [tests](https://github.com/symfony/ux/actions/runs/3696243151/jobs/6259682636) in this PR showed an error.

src/Chartjs/assets/src/chart_v4_controller.ts
evertharmeling: So the difference in 'importing' should be added specifically to the `v3` and `v4` controller files.

src/Chartjs/assets/src/chart_v4_controller.ts
```diff
@@ -0,0 +1,20 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import AbstractChartController from './abstract_controller';
+import Chart from 'chart.js/auto';
```
weaverryan: I suspect that this is caused by something "odd" with our setup (I believe it's because Chart.js 4 is ESM-only) but to fix the test:

```suggestion
import Chart from 'chart.js/auto/auto.cjs';
```

But I'm not sure that's the proper solution... and it seems like, from a user's perspective, `chart.js/auto` DOES work, and we shouldn't need 2 controllers to solve this problem. Again, I've got the feeling that we're doing something silly...

evertharmeling: Well that indeed seemed to fix the problem, and the tests in the PR are still failing because of reason mentioned in the todo-list.

Yeah I got that hunch too, seems like we are only satisfying the test-suite for some reason. But that part goes beyond my current js (test-suite) knowledge unfortunately...

evertharmeling: I looked at something like [resolve](https://www.npmjs.com/package/resolve), to dynamically check for the module existence so we could still have a single controller. But couldn't get that to work 😞

weaverryan: Btw, I think that:

A) Our code already works with V4 and we just need to allow it in `package.json`. I think you and/or other users have already verified this.
B) The actual test failing problem is a problem with our tests - specifically, I think jest is behaving poorly because it doesn't it doesn't play well with modules.

Possible solution is to move from jest to https://vitest.dev/

94noni (approved):
Nice one 👍🏻
Moreover as it is the lib showcased in the readme :)

simonsolutions:
Does the version in package.json has to be incremented?
https://github.com/symfony/ux/blob/267a5f7af18449883fce24cfab232d5efe11ae14/src/Chartjs/assets/package.json#L5

boedah:
@weaverryan I can also confirm v4 of ChartJS works.
However, each time we update/install composer packages, the `"chart.js": "^4.2.1"` in `package.json` gets reset to `^3.4.1` (which has to be rolled-back manually each time).

Is there something we can do to get a release for this change?
(besides having to rewrite the tests using vitest, as u mentioned :)

weaverryan:
> Is there something we can do to get a release for this change?
> (besides having to rewrite the tests using vitest, as u mentioned :)

That's all I know of :p. Though, I'm not sure it's a full rewrite - just the lower plumbing needs to be changed if my research was all correct. The other option is to let some tests fail for chart.js / silence them, which also sucks. Simply put: we can't find a way (other than my proposed idea) to get the tests to pass with chart.js v4, even though we all know that v4 works fine.

pculka:
Any ETA on supporting 4.0+ btw? Will it hit the 6.4LTS release?

boedah:
@weaverryan I just saw the fantastic #1202 - but have no idea what is left to be done here.
Maybe just release a new major (?) ux-chartjs which requires Chart.js 4+?

evertharmeling:
Fixed by https://github.com/symfony/ux/pull/1389

PR #616, opened 2022-12-18
[Translator] Add Symfony UX Translator package

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | Fix #... 
| License       | MIT

Hi everyone, this PR add a new UX Translator package, which allow to re-use the same logic than the Symfony Translator, but in JavaScript.
The Intl ICU syntax is also supported.

# Why?

This package is not a revolution, there is already https://github.com/willdurand/BazingaJsTranslationBundle which allows to re-use translations (defined in your Symfony app) in JavaScript.

## Loading issues

However, there is one big difference between https://github.com/willdurand/BazingaJsTranslationBundle and this new UX package: the way translations are passed to JavaScript.

To me, Bazinga contains performance issues. If you choose to dump your translations into a separated `.js` or `.json`, or if you choose to load translations through a separate HTTP request, in each case you will need to dump and load the **whole** catalogue (for a given domain), which contains messages that you will not use into the webpage.  

With the Symfony UX Translator package's logic, you explicitly define which translations you wants to load, making HTTP responses **less heavier** and/or JavaScript execution **faster**.

## Types definition

Another issue is that Bazinga does not provide types definitions in the same repository.
Instead, you must install [`@types/bazinga-translator`](https://www.npmjs.com/package/@types/bazinga-translator) package separately, which can lead to some "desynchronization" between the upstream code and types.

# How 

Everything is documented in https://github.com/Kocal/symfony-ux/blob/feat/translator/src/Translator/doc/index.rst, but to be quick:
- a CacheWarmer dumps your app's translations as constants into a JavaScript file (types definitions are generated too), that you can imports later in your JavaScript 
- a function `trans()` can be imported from `@symfony/ux-translator`, which:
  - works the same as `TranslatorInterface#trans()` from Symfony (PHP)
  - fully supports ICU, thanks to the code ported from https://formatjs.io/docs/icu-messageformat-parser
  - support default locale (configured, or from the `<html>`'s `lang` attribute, and locales fallbacks 
  - support TypeScript types definitions, telling you and validate which parameters are usable or not, their types, which domain or locale is usable, ... 
- the bundle's size will stay as lightweight as possible, thanks to the Tree Shaking

### DIFF
--- a/src/Translator/LICENSE
+++ b/src/Translator/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2023-present Fabien Potencier
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is furnished
+to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/src/Translator/assets/dist/translator_controller.js
+++ b/src/Translator/assets/dist/translator_controller.js
@@ -1,0 +1,253 @@
+import { IntlMessageFormat } from 'intl-messageformat';
+
+function formatIntl(id, parameters = {}, locale) {
+    if (id === '') {
+        return '';
+    }
+    const intlMessage = new IntlMessageFormat(id, [locale.replace('_', '-')], undefined, { ignoreTag: true });
+    parameters = Object.assign({}, parameters);
+    Object.entries(parameters).forEach(([key, value]) => {
+        if (key.includes('%') || key.includes('{')) {
+            delete parameters[key];
+            parameters[key.replace(/[%{} ]/g, '').trim()] = value;
+        }
+    });
+    return intlMessage.format(parameters);
+}
+
+function strtr(string, replacePairs) {
+    const regex = Object.entries(replacePairs).map(([from]) => {
+        return from.replace(/([-[\]{}()*+?.\\^$|#,])/g, '\\$1');
+    });
+    if (regex.length === 0) {
+        return string;
+    }
+    return string.replace(new RegExp(regex.join('|'), 'g'), (matched) => replacePairs[matched].toString());
+}
+
+function format(id, parameters = {}, locale) {
+    if (null === id || '' === id) {
+        return '';
+    }
+    if (typeof parameters['%count%'] === 'undefined' || Number.isNaN(parameters['%count%'])) {
+        return strtr(id, parameters);
+    }
+    const number = Number(parameters['%count%']);
+    let parts = [];
+    if (/^\|+$/.test(id)) {
+        parts = id.split('|');
+    }
+    else {
+        const matches = id.match(/(?:\|\||[^|])+/g);
+        if (matches !== null) {
+            parts = matches;
+        }
+    }
+    const intervalRegex = /^(?<interval>({\s*(-?\d+(\.\d+)?[\s*,\s*\-?\d+(.\d+)?]*)\s*})|(?<left_delimiter>[[\]])\s*(?<left>-Inf|-?\d+(\.\d+)?)\s*,\s*(?<right>\+?Inf|-?\d+(\.\d+)?)\s*(?<right_delimiter>[[\]]))\s*(?<message>.*?)$/s;
+    const standardRules = [];
+    for (let part of parts) {
+        part = part.trim().replace(/\|\|/g, '|');
+        let matches = part.match(intervalRegex);
+        if (matches !== null) {
+            if (matches[2]) {
+                for (const n of matches[3].split(',')) {
+                    if (number === Number(n)) {
+                        return strtr(matches.groups['message'], parameters);
+                    }
+                }
+            }
+            else {
+                const leftNumber = '-Inf' === matches.groups['left'] ? Number.NEGATIVE_INFINITY : Number(matches.groups['left']);
+                const rightNumber = ['Inf', '+Inf'].includes(matches.groups['right']) ? Number.POSITIVE_INFINITY : Number(matches.groups['right']);
+                if (('[' === matches.groups['left_delimiter'] ? number >= leftNumber : number > leftNumber)
+                    && (']' === matches.groups['right_delimiter'] ? number <= rightNumber : number < rightNumber)) {
+                    return strtr(matches.groups['message'], parameters);
+                }
+            }
+        }
+        else {
+            matches = part.match(/^\w+:\s*(.*?)$/);
+            if (matches !== null) {
+                standardRules.push(matches[1]);
+            }
+            else {
+                standardRules.push(part);
+            }
+        }
+    }
+    const position = getPluralizationRule(number, locale);
+    if (typeof standardRules[position] === 'undefined') {
+        if (1 === parts.length && typeof standardRules[0] !== 'undefined') {
+            return strtr(standardRules[0], parameters);
+        }
+        throw new Error(`Unable to choose a translation for "${id}" with locale "${locale}" for value "${number}". Double check that this translation has the correct plural options (e.g. "There is one apple|There are %count% apples").`);
+    }
+    return strtr(standardRules[position], parameters);
+}
+function getPluralizationRule(number, locale) {
+    number = Math.abs(number);
+    let _locale = locale;
+    if (locale === 'pt_BR' || locale === 'en_US_POSIX') {
+        return 0;
+    }
+    _locale = _locale.length > 3 ? _locale.substring(0, _locale.indexOf('_')) : _locale;
+    switch (_locale) {
+        case 'af':
+        case 'bn':
+        case 'bg':
+        case 'ca':
+        case 'da':
+        case 'de':
+        case 'el':
+        case 'en':
+        case 'en_US_POSIX':
+        case 'eo':
+        case 'es':
+        case 'et':
+        case 'eu':
+        case 'fa':
+        case 'fi':
+        case 'fo':
+        case 'fur':
+        case 'fy':
+        case 'gl':
+        case 'gu':
+        case 'ha':
+        case 'he':
+        case 'hu':
+        case 'is':
+        case 'it':
+        case 'ku':
+        case 'lb':
+        case 'ml':
+        case 'mn':
+        case 'mr':
+        case 'nah':
+        case 'nb':
+        case 'ne':
+        case 'nl':
+        case 'nn':
+        case 'no':
+        case 'oc':
+        case 'om':
+        case 'or':
+        case 'pa':
+        case 'pap':
+        case 'ps':
+        case 'pt':
+        case 'so':
+        case 'sq':
+        case 'sv':
+        case 'sw':
+        case 'ta':
+        case 'te':
+        case 'tk':
+        case 'ur':
+        case 'zu':
+            return (1 == number) ? 0 : 1;
+        case 'am':
+        case 'bh':
+        case 'fil':
+        case 'fr':
+        case 'gun':
+        case 'hi':
+        case 'hy':
+        case 'ln':
+        case 'mg':
+        case 'nso':
+        case 'pt_BR':
+        case 'ti':
+        case 'wa':
+            return (number < 2) ? 0 : 1;
+        case 'be':
+        case 'bs':
+        case 'hr':
+        case 'ru':
+        case 'sh':
+        case 'sr':
+        case 'uk':
+            return ((1 == number % 10) && (11 != number % 100)) ? 0 : (((number % 10 >= 2) && (number % 10 <= 4) && ((number % 100 < 10) || (number % 100 >= 20))) ? 1 : 2);
+        case 'cs':
+        case 'sk':
+            return (1 == number) ? 0 : (((number >= 2) && (number <= 4)) ? 1 : 2);
+        case 'ga':
+            return (1 == number) ? 0 : ((2 == number) ? 1 : 2);
+        case 'lt':
+            return ((1 == number % 10) && (11 != number % 100)) ? 0 : (((number % 10 >= 2) && ((number % 100 < 10) || (number % 100 >= 20))) ? 1 : 2);
+        case 'sl':
+            return (1 == number % 100) ? 0 : ((2 == number % 100) ? 1 : (((3 == number % 100) || (4 == number % 100)) ? 2 : 3));
+        case 'mk':
+            return (1 == number % 10) ? 0 : 1;
+        case 'mt':
+            return (1 == number) ? 0 : (((0 == number) || ((number % 100 > 1) && (number % 100 < 11))) ? 1 : (((number % 100 > 10) && (number % 100 < 20)) ? 2 : 3));
+        case 'lv':
+            return (0 == number) ? 0 : (((1 == number % 10) && (11 != number % 100)) ? 1 : 2);
+        case 'pl':
+            return (1 == number) ? 0 : (((number % 10 >= 2) && (number % 10 <= 4) && ((number % 100 < 12) || (number % 100 > 14))) ? 1 : 2);
+        case 'cy':
+            return (1 == number) ? 0 : ((2 == number) ? 1 : (((8 == number) || (11 == number)) ? 2 : 3));
+        case 'ro':
+            return (1 == number) ? 0 : (((0 == number) || ((number % 100 > 0) && (number % 100 < 20))) ? 1 : 2);
+        case 'ar':
+            return (0 == number) ? 0 : ((1 == number) ? 1 : ((2 == number) ? 2 : (((number % 100 >= 3) && (number % 100 <= 10)) ? 3 : (((number % 100 >= 11) && (number % 100 <= 99)) ? 4 : 5))));
+        default:
+            return 0;
+    }
+}
+
+let _locale = null;
+let _localeFallbacks = {};
+function setLocale(locale) {
+    _locale = locale;
+}
+function getLocale() {
+    return (_locale ||
+        document.documentElement.getAttribute('data-symfony-ux-translator-locale') ||
+        document.documentElement.lang ||
+        'en');
+}
+function setLocaleFallbacks(localeFallbacks) {
+    _localeFallbacks = localeFallbacks;
+}
+function getLocaleFallbacks() {
+    return _localeFallbacks;
+}
+function trans(message, parameters = {}, domain = 'messages', locale = null) {
+    if (typeof domain === 'undefined') {
+        domain = 'messages';
+    }
+    if (typeof locale === 'undefined' || null === locale) {
+        locale = getLocale();
+    }
+    if (typeof message.translations === 'undefined') {
+        return message.id;
+    }
+    const localesFallbacks = getLocaleFallbacks();
+    const translationsIntl = message.translations[`${domain}+intl-icu`];
+    if (typeof translationsIntl !== 'undefined') {
+        while (typeof translationsIntl[locale] === 'undefined') {
+            locale = localesFallbacks[locale];
+            if (!locale) {
+                break;
+            }
+        }
+        if (locale) {
+            return formatIntl(translationsIntl[locale], parameters, locale);
+        }
+    }
+    const translations = message.translations[domain];
+    if (typeof translations !== 'undefined') {
+        while (typeof translations[locale] === 'undefined') {
+            locale = localesFallbacks[locale];
+            if (!locale) {
+                break;
+            }
+        }
+        if (locale) {
+            return format(translations[locale], parameters, locale);
+        }
+    }
+    return message.id;
+}
+
+export { getLocale, getLocaleFallbacks, setLocale, setLocaleFallbacks, trans };
--- a/src/Translator/assets/jest.config.js
+++ b/src/Translator/assets/jest.config.js
@@ -1,0 +1,1 @@
+module.exports = require('../../../jest.config.js');
--- a/src/Translator/assets/package.json
+++ b/src/Translator/assets/package.json
@@ -1,0 +1,20 @@
+{
+    "name": "@symfony/ux-translator",
+    "description": "Symfony Translator for JavaScript",
+    "license": "MIT",
+    "version": "1.0.0",
+    "main": "dist/translator_controller.js",
+    "types": "dist/translator_controller.d.ts",
+    "peerDependencies": {
+        "intl-messageformat": "^10.2.5"
+    },
+    "peerDependenciesMeta": {
+        "intl-messageformat": {
+            "optional": false
+        }
+    },
+    "devDependencies": {
+        "intl-messageformat": "^10.2.5",
+        "ts-jest": "^27.1.5"
+    }
+}
--- a/src/Translator/assets/src/formatters/formatter.ts
+++ b/src/Translator/assets/src/formatters/formatter.ts
@@ -1,0 +1,238 @@
+import {strtr} from '../utils';
+
+/**
+ * This code is adapted from the Symfony Translator Trait (6.2)
+ * @see https://github.com/symfony/symfony/blob/015d5015e353ee5448bf7c350de0db4a03f8e13a/src/Symfony/Contracts/Translation/TranslatorTrait.php
+ */
+
+/**
+ * Translates the given message.
+ *
+ * When a number is provided as a parameter named "%count%", the message is parsed for plural
+ * forms and a translation is chosen according to this number using the following rules:
+ *
+ * Given a message with different plural translations separated by a
+ * pipe (|), this method returns the correct portion of the message based
+ * on the given number, locale and the pluralization rules in the message
+ * itself.
+ *
+ * The message supports two different types of pluralization rules:
+ *
+ * interval: {0} There are no apples|{1} There is one apple|]1,Inf] There are %count% apples
+ * indexed:  There is one apple|There are %count% apples
+ *
+ * The indexed solution can also contain labels (e.g. one: There is one apple).
+ * This is purely for making the translations more clear - it does not
+ * affect the functionality.
+ *
+ * The two methods can also be mixed:
+ *     {0} There are no apples|one: There is one apple|more: There are %count% apples
+ *
+ * An interval can represent a finite set of numbers:
+ *  {1,2,3,4}
+ *
+ * An interval can represent numbers between two numbers:
+ *  [1, +Inf]
+ *  ]-1,2[
+ *
+ * The left delimiter can be [ (inclusive) or ] (exclusive).
+ * The right delimiter can be [ (exclusive) or ] (inclusive).
+ * Beside numbers, you can use -Inf and +Inf for the infinite.
+ *
+ * @see https://en.wikipedia.org/wiki/ISO_31-11
+ *
+ * @private
+ *
+ * @param id         The message id
+ * @param parameters An array of parameters for the message
+ * @param locale     The locale
+ */
+export function format(id: string, parameters: Record<string, string | number> = {}, locale: string): string {
+    if (null === id || '' === id) {
+        return '';
+    }
+
+    if (typeof parameters['%count%'] === 'undefined' || Number.isNaN(parameters['%count%'])) {
+        return strtr(id, parameters);
+    }
+
+    const number = Number(parameters['%count%']);
+
+    let parts: Array<string> = [];
+    if (/^\|+$/.test(id)) {
+        parts = id.split('|');
+    } else {
+        const matches = id.match(/(?:\|\||[^|])+/g);
+        if (matches !== null) {
+            parts = matches;
+        }
+    }
+
+    const intervalRegex = /^(?<interval>({\s*(-?\d+(\.\d+)?[\s*,\s*\-?\d+(.\d+)?]*)\s*})|(?<left_delimiter>[[\]])\s*(?<left>-Inf|-?\d+(\.\d+)?)\s*,\s*(?<right>\+?Inf|-?\d+(\.\d+)?)\s*(?<right_delimiter>[[\]]))\s*(?<message>.*?)$/s;
+
+    const standardRules: Array<string> = [];
+    for (let part of parts) {
+        part = part.trim().replace(/\|\|/g, '|');
+
+        let matches = part.match(intervalRegex);
+        if (matches !== null) {
+            if (matches[2]) {
+                for (const n of matches[3].split(',')) {
+                    if (number === Number(n)) {
+                        return strtr(matches.groups!['message'], parameters);
+                    }
+                }
+            } else {
+                const leftNumber = '-Inf' === matches.groups!['left'] ? Number.NEGATIVE_INFINITY : Number(matches.groups!['left']);
+                const rightNumber = ['Inf', '+Inf'].includes(matches.groups!['right']) ? Number.POSITIVE_INFINITY : Number(matches.groups!['right']);
+
+                if (('[' === matches.groups!['left_delimiter'] ? number >= leftNumber : number > leftNumber)
+                    && (']' === matches.groups!['right_delimiter'] ? number <= rightNumber : number < rightNumber)
+                ) {
+                    return strtr(matches.groups!['message'], parameters);
+                }
+            }
+        } else {
+            matches = part.match(/^\w+:\s*(.*?)$/);
+            if (matches !== null) {
+                standardRules.push(matches[1]);
+            } else {
+                standardRules.push(part);
+            }
+        }
+    }
+
+    const position = getPluralizationRule(number, locale);
+    if (typeof standardRules[position] === 'undefined') {
+        // when there's exactly one rule given, and that rule is a standard
+        // rule, use this rule
+        if (1 === parts.length && typeof standardRules[0] !== 'undefined') {
+            return strtr(standardRules[0], parameters);
+        }
+
+        throw new Error(`Unable to choose a translation for "${id}" with locale "${locale}" for value "${number}". Double check that this translation has the correct plural options (e.g. "There is one apple|There are %count% apples").`)
+    }
+
+    return strtr(standardRules[position], parameters);
+}
+
+/**
+ * Returns the plural position to use for the given locale and number.
+ *
+ * The plural rules are derived from code of the Zend Framework (2010-09-25),
+ * which is subject to the new BSD license (http://framework.zend.com/license/new-bsd).
+ * Copyright (c) 2005-2010 Zend Technologies USA Inc. (http://www.zend.com)
+ */
+function getPluralizationRule(number: number, locale: string): number {
+    number = Math.abs(number);
+    let _locale = locale;
+
+    if (locale === 'pt_BR' || locale === 'en_US_POSIX') {
+        return 0;
+    }
+
+    _locale = _locale.length > 3 ? _locale.substring(0, _locale.indexOf('_')) : _locale;
+
+    switch (_locale) {
+        case 'af':
+        case 'bn':
+        case 'bg':
+        case 'ca':
+        case 'da':
+        case 'de':
+        case 'el':
+        case 'en':
+        case 'en_US_POSIX':
+        case 'eo':
+        case 'es':
+        case 'et':
+        case 'eu':
+        case 'fa':
+        case 'fi':
+        case 'fo':
+        case 'fur':
+        case 'fy':
+        case 'gl':
+        case 'gu':
+        case 'ha':
+        case 'he':
+        case 'hu':
+        case 'is':
+        case 'it':
+        case 'ku':
+        case 'lb':
+        case 'ml':
+        case 'mn':
+        case 'mr':
+        case 'nah':
+        case 'nb':
+        case 'ne':
+        case 'nl':
+        case 'nn':
+        case 'no':
+        case 'oc':
+        case 'om':
+        case 'or':
+        case 'pa':
+        case 'pap':
+        case 'ps':
+        case 'pt':
+        case 'so':
+        case 'sq':
+        case 'sv':
+        case 'sw':
+        case 'ta':
+        case 'te':
+        case 'tk':
+        case 'ur':
+        case 'zu':
+            return (1 == number) ? 0 : 1;
+        case 'am':
+        case 'bh':
+        case 'fil':
+        case 'fr':
+        case 'gun':
+        case 'hi':
+        case 'hy':
+        case 'ln':
+        case 'mg':
+        case 'nso':
+        case 'pt_BR':
+        case 'ti':
+        case 'wa':
+            return (number < 2) ? 0 : 1;
+        case 'be':
+        case 'bs':
+        case 'hr':
+        case 'ru':
+        case 'sh':
+        case 'sr':
+        case 'uk':
+            return ((1 == number % 10) && (11 != number % 100)) ? 0 : (((number % 10 >= 2) && (number % 10 <= 4) && ((number % 100 < 10) || (number % 100 >= 20))) ? 1 : 2);
+        case 'cs':
+        case 'sk':
+            return (1 == number) ? 0 : (((number >= 2) && (number <= 4)) ? 1 : 2);
+        case 'ga':
+            return (1 == number) ? 0 : ((2 == number) ? 1 : 2);
+        case 'lt':
+            return ((1 == number % 10) && (11 != number % 100)) ? 0 : (((number % 10 >= 2) && ((number % 100 < 10) || (number % 100 >= 20))) ? 1 : 2);
+        case 'sl':
+            return (1 == number % 100) ? 0 : ((2 == number % 100) ? 1 : (((3 == number % 100) || (4 == number % 100)) ? 2 : 3));
+        case 'mk':
+            return (1 == number % 10) ? 0 : 1;
+        case 'mt':
+            return (1 == number) ? 0 : (((0 == number) || ((number % 100 > 1) && (number % 100 < 11))) ? 1 : (((number % 100 > 10) && (number % 100 < 20)) ? 2 : 3));
+        case 'lv':
+            return (0 == number) ? 0 : (((1 == number % 10) && (11 != number % 100)) ? 1 : 2);
+        case 'pl':
+            return (1 == number) ? 0 : (((number % 10 >= 2) && (number % 10 <= 4) && ((number % 100 < 12) || (number % 100 > 14))) ? 1 : 2);
+        case 'cy':
+            return (1 == number) ? 0 : ((2 == number) ? 1 : (((8 == number) || (11 == number)) ? 2 : 3));
+        case 'ro':
+            return (1 == number) ? 0 : (((0 == number) || ((number % 100 > 0) && (number % 100 < 20))) ? 1 : 2);
+        case 'ar':
+            return (0 == number) ? 0 : ((1 == number) ? 1 : ((2 == number) ? 2 : (((number % 100 >= 3) && (number % 100 <= 10)) ? 3 : (((number % 100 >= 11) && (number % 100 <= 99)) ? 4 : 5))));
+        default:
+            return 0
+    }
+}
--- a/src/Translator/assets/src/translator.ts
+++ b/src/Translator/assets/src/translator.ts
@@ -1,0 +1,167 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+export type DomainType = string;
+export type LocaleType = string;
+
+export type TranslationsType = Record<DomainType, { parameters: ParametersType }>;
+export type NoParametersType = Record<string, never>;
+export type ParametersType = Record<string, string | number> | NoParametersType;
+
+export type RemoveIntlIcuSuffix<T> = T extends `${infer U}+intl-icu` ? U : T;
+export type DomainsOf<M> = M extends Message<infer Translations, LocaleType> ? keyof Translations : never;
+export type LocaleOf<M> = M extends Message<TranslationsType, infer Locale> ? Locale : never;
+export type ParametersOf<M, D extends DomainType> = M extends Message<infer Translations, LocaleType>
+    ? Translations[D] extends { parameters: infer Parameters }
+        ? Parameters
+        : never
+    : never;
+
+export interface Message<Translations extends TranslationsType, Locale extends LocaleType> {
+    id: string;
+    translations: {
+        [domain in DomainType]: {
+            [locale in Locale]: string;
+        };
+    };
+}
+
+import { formatIntl } from './formatters/intl-formatter';
+import { format } from './formatters/formatter';
+
+let _locale: LocaleType | null = null;
+let _localeFallbacks: Record<LocaleType, LocaleType> = {};
+
+export function setLocale(locale: LocaleType | null) {
+    _locale = locale;
+}
+
+export function getLocale(): LocaleType {
+    return (
+        _locale ||
+        document.documentElement.getAttribute('data-symfony-ux-translator-locale') || // <html data-symfony-ux-translator-locale="en">
+        document.documentElement.lang || // <html lang="en">
+        'en'
+    );
+}
+
+export function setLocaleFallbacks(localeFallbacks: Record<LocaleType, LocaleType>): void {
+    _localeFallbacks = localeFallbacks;
+}
+
+export function getLocaleFallbacks(): Record<LocaleType, LocaleType> {
+    return _localeFallbacks;
+}
+
+/**
+ * Translates the given message, in ICU format (see https://formatjs.io/docs/intl-messageformat) or Symfony format (see below).
+ *
+ * When a number is provided as a parameter named "%count%", the message is parsed for plural
+ * forms and a translation is chosen according to this number using the following rules:
+ *
+ * Given a message with different plural translations separated by a
+ * pipe (|), this method returns the correct portion of the message based
+ * on the given number, locale and the pluralization rules in the message
+ * itself.
+ *
+ * The message supports two different types of pluralization rules:
+ *
+ * interval: {0} There are no apples|{1} There is one apple|]1,Inf] There are %count% apples
+ * indexed:  There is one apple|There are %count% apples
+ *
+ * The indexed solution can also contain labels (e.g. one: There is one apple).
+ * This is purely for making the translations more clear - it does not
+ * affect the functionality.
+ *
+ * The two methods can also be mixed:
+ *     {0} There are no apples|one: There is one apple|more: There are %count% apples
+ *
+ * An interval can represent a finite set of numbers:
+ *  {1,2,3,4}
+ *
+ * An interval can represent numbers between two numbers:
+ *  [1, +Inf]
+ *  ]-1,2[
+ *
+ * The left delimiter can be [ (inclusive) or ] (exclusive).
+ * The right delimiter can be [ (exclusive) or ] (inclusive).
+ * Beside numbers, you can use -Inf and +Inf for the infinite.
+ *
+ * @see https://en.wikipedia.org/wiki/ISO_31-11
+ *
+ * @param message    The message
+ * @param parameters An array of parameters for the message
+ * @param domain     The domain for the message or null to use the default
+ * @param locale     The locale or null to use the default
+ */
+export function trans<
+    M extends Message<TranslationsType, LocaleType>,
+    D extends DomainsOf<M>,
+    P extends ParametersOf<M, D>
+>(
+    ...args: P extends NoParametersType
+        ? [message: M, parameters?: P, domain?: RemoveIntlIcuSuffix<D>, locale?: LocaleOf<M>]
+        : [message: M, parameters: P, domain?: RemoveIntlIcuSuffix<D>, locale?: LocaleOf<M>]
+): string;
+export function trans<
+    M extends Message<TranslationsType, LocaleType>,
+    D extends DomainsOf<M>,
+    P extends ParametersOf<M, D>
+>(
+    message: M,
+    parameters: P = {} as P,
+    domain: RemoveIntlIcuSuffix<DomainsOf<M>> | undefined = 'messages' as RemoveIntlIcuSuffix<DomainsOf<M>>,
+    locale: LocaleOf<M> | null = null
+): string {
+    if (typeof domain === 'undefined') {
+        domain = 'messages' as RemoveIntlIcuSuffix<DomainsOf<M>>;
+    }
+
+    if (typeof locale === 'undefined' || null === locale) {
+        locale = getLocale() as LocaleOf<M>;
+    }
+
+    if (typeof message.translations === 'undefined') {
+        return message.id;
+    }
+
+    const localesFallbacks = getLocaleFallbacks();
+
+    const translationsIntl = message.translations[`${domain}+intl-icu`];
+    if (typeof translationsIntl !== 'undefined') {
+        while (typeof translationsIntl[locale] === 'undefined') {
+            locale = localesFallbacks[locale] as LocaleOf<M>;
+            if (!locale) {
+                break;
+            }
+        }
+
+        if (locale) {
+            return formatIntl(translationsIntl[locale], parameters, locale);
+        }
+    }
+
+    const translations = message.translations[domain];
+    if (typeof translations !== 'undefined') {
+        while (typeof translations[locale] === 'undefined') {
+            locale = localesFallbacks[locale] as LocaleOf<M>;
+            if (!locale) {
+                break;
+            }
+        }
+
+        if (locale) {
+            return format(translations[locale], parameters, locale);
+        }
+    }
+
+    return message.id;
+}
--- a/src/Translator/assets/src/translator_controller.ts
+++ b/src/Translator/assets/src/translator_controller.ts
@@ -1,0 +1,12 @@
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+export * from './translator';
--- a/src/Translator/assets/src/utils.ts
+++ b/src/Translator/assets/src/utils.ts
@@ -1,0 +1,19 @@
+/**
+ * PHP strtr's equivalent, inspired and adapted from https://stackoverflow.com/a/37949642.
+ *
+ * @private
+ *
+ * @param string The string to replace in
+ * @param replacePairs The pairs of characters to replace
+ */
+export function strtr(string: string, replacePairs: Record<string, string | number>): string {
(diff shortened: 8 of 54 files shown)


src/Translator/assets/test/formatters/formatter.test.ts
```diff
@@ -0,0 +1,149 @@
+/*
```
Kocal, line 1: For this file, I've reused the same tests data from https://github.com/symfony/symfony/blob/015d5015e353ee5448bf7c350de0db4a03f8e13a/src/Symfony/Contracts/Translation/Test/TranslatorTest.php

src/Translator/assets/test/formatters/intl-formatter.test.ts
```diff
@@ -0,0 +1,60 @@
+/*
```
Kocal, line 1: For this file, I've reused some of the same test data from https://github.com/symfony/symfony/blob/015d5015e353ee5448bf7c350de0db4a03f8e13a/src/Symfony/Component/Translation/Tests/Formatter/IntlFormatterTest.php

src/Translator/assets/test/formatters/intl-formatter.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    test('percents and brackets are trimmed', function() {
+        expect(formatIntl('Hello {name}', { name: 'Fab'}, 'messages+intl-icu', 'en')).toEqual('Hello Fab');
+        expect(formatIntl('Hello {name}', { '%name%': 'Fab'}, 'messages+intl-icu', 'en')).toEqual('Hello Fab');
+        expect(formatIntl('Hello {name}', { '{{ name }}': 'Fab'}, 'messages+intl-icu', 'en')).toEqual('Hello Fab');
+    });
+
+    test('format with HTML inside', function() {
+        expect(formatIntl('Hello <b>{name}</b>', { name: 'Fab'}, 'messages+intl-icu', 'en')).toEqual('Hello <b>Fab</b>');
+        expect(formatIntl('Hello {name}', { name: '<b>Fab</b>'}, 'messages+intl-icu', 'en')).toEqual('Hello <b>Fab</b>');
+    })
```
Kocal: We want to ensure that the issue fixed in https://github.com/willdurand/BazingaJsTranslationBundle/pull/325 is not present here.

src/Translator/assets/src/formatters/formatter.ts
```diff
@@ -0,0 +1,238 @@
...
+    number = Math.abs(number);
+    let _locale = locale;
+
+    if (locale === 'pt_BR' || locale === 'en_US_POSIX') {
+        return 0;
+    }
+
+    _locale = _locale.length > 3 ? _locale.substring(0, _locale.indexOf('_')) : _locale;
+
+    switch (_locale) {
```
Kern046, line 136: Is there no way to extract this switch in a specific function to avoid duplicating it both here and in `translator_controller` ? Easier to maintain and unit-test.

Kocal: The `formatter.ts` is the only file which contains this logic, the `translator_controller` you are reffering to is the one from `dist/` folder, which basically contains all the code from `src/` folder.

src/Translator/assets/dist/translator_controller.js
```diff
@@ -0,0 +1,229 @@
...
+    }
+    const intlMessage = new IntlMessageFormat(id, [locale.replace('_', '-')], undefined, { ignoreTag: true });
+    Object.entries(parameters).forEach(([key, value]) => {
+        if (key.includes('%') || key.includes('{')) {
+            delete parameters[key];
+            parameters[key.replace(/[%{} ]/g, '').trim()] = value;
+        }
+    });
+    return intlMessage.format(parameters);
+}
```
Mikescops, line 16: This function has two drawbacks that you may want to solve:

1. it is updating the parent object "parameters" without making a proper copy
2. it's unclear from your regex what is the expected behavior for cleaning `{` and `}` because here you're only looking for open brackets to trigger the clean
3. (also the if is not really needed if you use a regex)

I'm suggesting below a different approach (you may use the test scenarios for the unit tests too):

```ts
function sample(parameters = {}) {
  Object.entries(parameters).forEach(([key, value]) => {
    if (key.includes('%') || key.includes('{')) {
      delete parameters[key];
      parameters[key.replace(/[%{} ]/g, '').trim()] = value;
    }
  });
  return parameters
}

function rework(parameters = {}) {
  return Object.entries(parameters).reduce((acc, [key, value]) => {
    return {
      ...acc,
      [key.replace(/[%{} ]/g, '').trim()]: value
    }
  }, {});
}

const parameters = {
  'example1': 'content1',
  'example2%': 'content2',
  '{example3}': 'content3',
  '%{ example4 }%': 'content4',
  'example5 }': 'content5',
  'example6% }': 'content6'
}

console.log('params', parameters);

console.log('rework', rework(parameters));
console.log('params', parameters);

console.log('sample', sample(parameters));
console.log('params', parameters);

```

I made a quick JSFiddle to show the result: https://jsfiddle.net/p6qnzs7t/18/

src/Translator/assets/dist/translator_controller.js
```diff
@@ -0,0 +1,229 @@
...
+    const map = {};
+    Object.entries(replacePairs).forEach(([from, to]) => {
+        regex.push(from.replace(/([-[\]{}()*+?.\\^$|#,])/g, '\\$1'));
+        map[from] = to;
+    });
+    if (regex.length === 0) {
+        return string;
+    }
+    return string.replace(new RegExp(regex.join('|'), 'g'), (matched) => map[matched].toString());
+}
```
Mikescops, line 26: You're rebuilding a map that is the exact same as your parameter, here is a suggestion to simplify:

```ts
function sample(string, replacePairs) {
  const regex = [];
  const map = {};
  Object.entries(replacePairs).forEach(([from, to]) => {
    regex.push(from.replace(/([-[\]{}()*+?.\\^$|#,])/g, '\\$1'));
    map[from] = to;
  });
  if (regex.length === 0) {
    return string;
  }
  return string.replace(new RegExp(regex.join('|'), 'g'), (matched) => map[matched].toString());
}

function rework(string, replacePairs) {
  const regex = Object.entries(replacePairs).map(([from, _to]) => {
    return from.replace(/([-[\]{}()*+?.\\^$|#,])/g, '\\$1');
  });
  if (regex.length === 0) {
    return string;
  }
  return string.replace(new RegExp(regex.join('|'), 'g'), (matched) => replacePairs[matched].toString());
}

const samples = [
        // https://github.com/php/php-src/blob/master/ext/standard/tests/strings/strtr_basic.phpt
        ['TEST STrTr', 'test strtr', {'t': 'T', 'e': 'E', 'st': 'ST'}],
        ['TEST STrTr', 'test strtr', {'t': 'T', 'e': 'E', 'st': 'ST'}],

        // https://github.com/php/php-src/blob/master/ext/standard/tests/strings/strtr_variation1.phpt
        ['a23', '123', {'1': 'a', 'a': '1', '2b3c': 'b2c3', 'b2c3': '3c2b'}],
        ['1bc', 'abc', {'1': 'a', 'a': '1', '2b3c': 'b2c3', 'b2c3': '3c2b'}],
        ['a1b2c3', '1a2b3c', {'1': 'a', 'a': '1', '2b3c': 'b2c3', 'b2c3': '3c2b'}],
        [`
a23
1bc
a1b2c3`, `
123
abc
1a2b3c`, {1: 'a', 'a': '1', '2b3c': 'b2c3', 'b2c3': '3c2b'}],

        // https://github.com/php/php-src/blob/master/ext/standard/tests/strings/strtr_variation2.phpt
        ['$', '%', {'$': '%', '%': '$', '#*&@()': '()@&*#'}],
        ['#%*', '#$*', {'$': '%', '%': '$', '#*&@()': '()@&*#'}],
        ['text & @()', 'text & @()', {'$': '%', '%': '$', '#*&@()': '()@&*#'}],
        [`
$
#%*&
text & @()`, `
%
#$*&
text & @()`, {'$': '%', '%': '$', '#*&@()': '()@&*#'}],
    ]


console.log('rework', rework(samples[7][1], samples[7][2]));

console.log('sample', sample(samples[7][1], samples[7][2]));
```

A JSFiddle to test quickly https://jsfiddle.net/kpfthv6u/8/

src/Translator/assets/src/utils.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * PHP strtr's equivalent, inspired and adapted from https://stackoverflow.com/a/37949642.
+ * @param string The string to replace in
+ * @param replacePairs The pairs of characters to replace
+ */
+export function strtr(string: string, replacePairs: Record<string, string | number>): string {
```
Mikescops, line 9: This function and the one in `src/Translator/assets/dist/translator_controller.js` seems identical, can they be merged?

Mikescops (changes requested):
Hello 👋 ,

I'm not used to Symfony at all but wanted to suggest some optimizations on the JS part.

src/Translator/assets/src/utils.ts
Mikescops: Dist folder, never mind.

src/Translator/assets/dist/translator_controller.js
Kocal: 1. Fixed, thanks
2. As said on Discord (:eyes:), the logic is taken from the Symfony Translator in its PHP version (https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Translation/Formatter/IntlFormatter.php#L45), so I prefer keep the same logic instead.
3. Usually you want to quickly check if one of the concerned character can be found before using the regex, to prevent performances loss if you directly use (parse + run) the regex. However I don't know how its behave with JavaScript.

src/Translator/assets/dist/translator_controller.js
Kocal: Fixed, thanks

kbond:
While I can see the need for this, my app doesn't have so many translations that I'd be concerned with the perf impact of dumping the entire catalog. Should we include this functionality here or are you thinking, in this case, use [BazingaJsTranslationBundle](https://github.com/willdurand/BazingaJsTranslationBundle)?

Kocal:
Some applications can have more than 1000 keys (we got ~2500 in our main app 😅):
image

In those 2500 keys, maybe 200 are used both in PHP and injected in JavaScript (we translate the messages with `|trans` and expose them in JavaScript, but it does not supports parameters). So we can't dump whole translations files nor creating a dedicated domain for JavaScript because we don't want to duplicate translations. 

I would like to see this functionality here because this package behave as a bridge between Symfony and JavaScript:
> Symfony UX: A set of PHP & JavaScript packages to solve every day frontend problems

I know that the BazingaJsTranslationBundle suffers from a lack of maintainability, the last release is 4 months old and some PRs are still opened. IINW @ willdurand does not maintain (or even use?) its bundle anymore.

Finally, I think it can makes Symfony UX more "valuable", as it provides a more simple and light installation + usage than Bazinga.

kbond:
> In those 2500 keys, maybe 200 are used both in PHP and injected in JavaScript

Is there an alternate way we can flag these translations for js? Adding each and every key in your twig template feels a bit cumbersome, no?

Message catalog metadata? Just thinking out loud.

> I know that the BazingaJsTranslationBundle suffers from a lack of maintainability

So should we consider moving it's primary functionality here (in addition to this new feature)?

> Finally, I think it can makes Symfony UX more "valuable", as it provides a more simple and light installation + usage than Bazinga.

I agree but clearly many people use Bazinga in it's current form w/o the perf issues you describe. I think we should provide both options.

Kocal:
> > In those 2500 keys, maybe 200 are used both in PHP and injected in JavaScript
> 
> Is there an alternate way we can flag these translations for js? Adding each and every key in your twig template feels a bit cumbersome, no?
> 
> Message catalog metadata? Just thinking out loud.

I can looks cumbersome yeah, but I don't find it shocking since we are not building an SPA.

Where the end-user can access to Catalogue Metadata? To me those metadata are internals to Symfony right?

> > I know that the BazingaJsTranslationBundle suffers from a lack of maintainability
> 
> So should we consider moving it's primary functionality here (in addition to this new feature)?

To me, yes. Each PHP + JS existing bundles should be put under the Symfony UX brand/repository.

> > Finally, I think it can makes Symfony UX more "valuable", as it provides a more simple and light installation + usage than Bazinga.
> 
> I agree but clearly many people use Bazinga in it's current form w/o the perf issues you describe.

I think that many people use Bazinga because it's the only one bundle to make re-using Symfony translations messages in JavaScript possible. 

For performance issues, I'm willing to bet my arm they either don't know or they don't care about these. Unfortunatelly, not every developer use profiling tools like Blackfire for PHP, and the tabs Analyzer/Performance Insights for JavaScript.

>  I think we should provide both options.

We can, this is something I planned for a 2nd PR if people had the issues. We can do it by making the 1st parameter of `load_translations_for_js()` optional, which will load all translations for a domain+locale. WDYT?

kbond:
> Where the end-user can access to Catalogue Metadata? To me those metadata are internals to Symfony right?

I think it can be used outside - the standard `MessageCatalog` implements [`MetadataAwareInterface`](https://github.com/symfony/symfony/blob/926b450c297a87c5673c45e2326a484f619d9f51/src/Symfony/Component/Translation/MetadataAwareInterface.php). Don't know if all the loaders support metadata - xliff for sure. Would be really neat if you could _tag_ translations for _js_. Outside of the scope of this current PR - just something to think about later.

> We can, this is something I planned for a 2nd PR if people had the issues. We can do it by making the 1st parameter of load_translations_for_js() optional, which will load all translations for a domain+locale. WDYT?

This, or I was thinking dumping a js file to be included in a user's encore asset build? I use `friendsofsymfony/jsrouting-bundle` and this pattern works well. You flag routes to be included in js and these are dumped (this is where I had the idea of flagging translation messages). You don't need any twig helpers - just include the `router.js` script provided by the bundle and the dumped `routes.js`.

For translations, we'd need the figure out what the current locale is. Can we use the `<html lang="x">` attribute to get this?

Kocal:
> > We can, this is something I planned for a 2nd PR if people had the issues. We can do it by making the 1st parameter of load_translations_for_js() optional, which will load all translations for a domain+locale. WDYT?
> 
> This, or I was thinking dumping a js file to be included in a user's encore asset build? I use `friendsofsymfony/jsrouting-bundle` and this pattern works well. You flag routes to be included in js and these are dumped (this is where I had the idea of flagging translation messages).

I've also used this pattern (with Bazinga and `friendsofsymfony/jsrouting-bundle`), it worked pretty fine but this was a bit tricky to setup:
- "should we commit those files?"
- you can't run Encore (or equivalent) before installing PHP deps and making your PHP app usable, so you must modify your installation steps to make things working, and that's not a easy thing to do with PaaS. Note: this argument is not so much valid anymore with Symfony UX, which seems have "popularized" this behavior.

The `load_translations_for_js()` is a solution that does not suffer about those _issues_.

kbond:
Certainly `load_translations_for_js()` could be an option but if we add the ability to dump all messages, we shouldn't suggest adding to every page's content.

Maybe we shouldn't even provide this. If you want to pick and choose which messages to include, use `load_translations_for_js()`. If you want to dump you're entire catalog, use the js-dump command and include in your assets.

Kocal:
> For translations, we'd need the figure out what the current locale is. Can we use the `<html lang="x">` attribute to get this?

Mmmh, maybe it can works if you use `<html lang="{{ app.request.locale }}">`, but this is not always true (e.g.: IIRC for SEO practices).

kbond:
Yeah, my thinking was to default the locale to the `html.lang` attribute and if `initialize_js_translator()` is used, override.

tgalopin:
Just my adding my two cents: I did something similar in the past (declaring translations in JS globally) but in the end, I reverted back to translations local to controllers. It's cumbersome to declare all translations to expose so my thinking was that I might as well declare them locally in a "labels" key in the controller to pass the translations directly from Twig:

```twig
{{ stimulus_controller('my_controller', {
    'labels': {
        'location': 'crm.list.location'|trans,
        'createdAt': 'crm.list.createdAt'|trans,
        ...
    }
) }}
```

This provides the following benefits:
* It naturally exposes only the translations actually used by the project ;
* It allows to have different translations for a same controller/component used in different places (think something like a confirm button)
* It allows to profit from Twig autocompletion and other similar tools to detect translation usages

So I'd say that if we don't have a big DX improvement using this component over this "local key" approach, we shouldn't release it.

There is a different way I can think of that could provide such great DX improvement: leveraging Webpack tree shaking to not have to care about marking translations to expose and easily importing them.

In such case, a translations module would be declared dynamically as a virtual module and would create one exported JS constant per translation key. We could then import these keys in our JS files (we could even have autocompletion if we create a stub file) and they would be linked to either EN or FR at runtime. Because of tree shaking, Webpack would remove the unused keys from the built JS automatically.

WDYT? I know it's very different from the current proposal but I really think we should be careful about the DX of this component, it's tricky to do right :) .

Kocal:
Hi @tgalopin, and thanks for your comment.

Your approach is interesting, however I can see three major issues in your code example:
1. You are "forced" to use Stimulus controller. It's not that I don't like them, quite the opposite, but it complicates the migrations from [BazingaJsTranslationBundle](https://github.com/willdurand/BazingaJsTranslationBundle) where developpers could use `Translator.trans('my_message')`   
2. You need to inject those translations everywhere you need them (i.e.: in a Vue.js component or a `utils.js`), where we can easily rely on a global state (like [BazingaJsTranslationBundle](https://github.com/willdurand/BazingaJsTranslationBundle) and this new package do)
3. You can't use translation parameters with values coming from JS since the message is already translated before, which can makes the UI less interactive

> There is a different way I can think of that could provide such great DX improvement: leveraging Webpack tree shaking to not have to care about marking translations to expose and easily importing them.

> In such case, a translations module would be declared dynamically as a virtual module and would create one exported JS constant per translation key. We could then import these keys in our JS files (we could even have autocompletion if we create a stub file) and they would be linked to either EN or FR at runtime. Because of tree shaking, Webpack would remove the unused keys from the built JS automatically.

I like this suggestion, however I have many questions:
1. Where those JS constants would be defined? In a file previously dumped by a Symfony command?
2. What if we have two same keys from different domains?
3. What if the locale is updated (through `setLocale()`), how the constants import + three-shaking will works? Should we import all translations from a given `id` ?
4. Are we ok to keep `trans(key, parameters, domain, locale)` function? 

Thanks!

tgalopin:
> Where those JS constants would be defined? In a file previously dumped by a Symfony command?

I was thinking about a virtual module or a Webpack loader so that it's dynamic and it doesn't need to be dumped in between.

It could generate multiple files (one per domain) like:

```typescript
export const CRM_LIST_LOCATION = {
    'en': '...',
    'fr': '...',
};
```

This would solve your 3 first questions (2: one file per domain, 3: we need to have locales in the key so that tree-shaking works as expected).

For 4, the usage could then be:

```typescript
CRM_LIST_LOCATION['en']
// or perhaps
trans(CRM_LIST_LOCATION, 'en')
```

There may be a better DX to be found though, expecially given how big the DX improvements can be.

Kocal:
About the virtual module, would it add too much complexity? 

I'm not familiar with them, but I wonder how we will deal with auto-completion in IDE? With TypeScript support? How do we get translations, from a Symfony command? If yes how do we run it (which `php` executable should be use?)

If we go to this direction, I prefer having a dedicated Symfony command that will dump translations (for all domains + enabled locales) somewhere in multiple `.ts` files (one file per domain). 

After running the command manually (ie. during a `composer install`), you will be able to import translations like this:
```ts
import { trans } from '@symfony/ux-translator';
import { CRM_LIST_LOCATION, FOO_BAR } from '@symfony/ux-translator/translations/messages';

trans(CRM_LIST_LOCATION, { param1: 'value1' }, 'en'); 
```

We can also imagine, maybe for later, thanks to TypeScript types, to have auto-completion on translations parameters given the translation id, WDYT?

/cc @weaverryan @kbond

kbond:
That'd be really cool! Could the imported messages be js objects? (I have minimal experience with js so not sure if this is even possible)

```js
import { CRM_LIST_LOCATION, FOO_BAR } from '@symfony/ux-translator/translations/messages';

CRM_LIST_LOCATION; // no parameters just is the value

FOO_BAR.with({ param1: 'value1' });
FOO_BAR.with({ param1: 'value1' }).locale('fr');

// maybe this would be required? (i'm thinking about __toString() in PHP above)
CRM_LIST_LOCATION.render();
FOO_BAR.with({ param1: 'value1' }).render();
FOO_BAR.with({ param1: 'value1' }).locale('fr').render();
```

Also, how are you thinking the locale-detection would work? I'm thinking the following _cascade_:
1. defaults to `en`
2. `html.lang` attr
3. some kind of twig function ( ie `{{ js_trans_locale('fr') }}`)
4. `trans()` js function

Kocal:
The following syntax would be possible, however I'm not a big of creating multiple objects like this because it can lead to performances issues.

kbond:
> creating multiple objects like this because it can lead to performances issues.

Ok, fair enough.

--- 

~I'm guessing each domain would/could be dumped by the command?~

```js
import { trans } from '@symfony/ux-translator';
import { FOO } from '@symfony/ux-translator/translations/messages';
import { BAR } from '@symfony/ux-translator/translations/domain1';
```
~Or how would that work?~

Nevermind, I see you specified this above :).

---

For small apps that don't have a perf problem loading all their translations, what about something like:

```js
import { trans } from '@symfony/ux-translator/translations/_all';

trans('key1');
trans('key2', { param1: 'value1' }, domain: 'security');
trans('key3', domain: 'security', locale: 'fr');
```

Kocal:
> For small apps that don't have a perf problem loading all their translations, what about something like:
> 
> ```js
> import { trans } from '@symfony/ux-translator/translations/_all';
> 
> trans('key1');
> trans('key2', { param1: 'value1' }, domain: 'security');
> trans('key3', domain: 'security', locale: 'fr');
> ```

The `trans()` function from `@symfony/ux-translator/translations/_all` would be a different function than `trans()` from `@symfony/ux-translator`?
 
If no, we can't make tree-shaking + translation id auto-completion + translation parameters working all together
If yes, I'm a bit afraid to confuse users with two different functions having the same name but different parameters type.

Constants imported from `@symfony/ux-translator/translations/messages` won't be a string but an object contains the following data:
- the translated message per local
- the parameters (maybe in a future iteration)

Thanks to TypeScript types, we could get parameters autocompletion given the constant name. This is also something that can work with your proposition, but you won't benefit from tree-shaking. 

If you really need to import all translations, per domain, you can still use `import * as messages from '@symfony/ux-translator/translations/messages';`, but I honestly don't think there is an interest in doing that.

WDYT?

kbond:
> The trans() function from @symfony/ux-translator/translations/_all would be a different function than trans() from @symfony/ux-translator?

I was thinking yes. 

> If you really need to import all translations, per domain, you can still use import * as messages from '@symfony/ux-translator/translations/messages';, but I honestly don't think there is an interest in doing that.

Your probably right but I'm trying to find an easy migration path for bazinga users.

Kocal:
> > The trans() function from @symfony/ux-translator/translations/_all would be a different function than trans() from @symfony/ux-translator?
> 
> I was thinking yes.

Alright, I will see what I can do.

> > If you really need to import all translations, per domain, you can still use import * as messages from '@symfony/ux-translator/translations/messages';, but I honestly don't think there is an interest in doing that.
> 
> Your probably right but I'm trying to find an easy migration path for bazinga users.

Oh, yeah that's true! I don't have a better suggestion for the moment, let's go for `_all` + special `trans()` function so.

Kocal:
This is what I've imagined for parameters/locales autocompletion, by using constants: 

```ts
interface Message<Translations = { [locale: string]: string }, Parameters = Record<string, unknown>> {
    domain: string;
    translations: Translations,
    parameters?: Parameters,
}

export type TranslationsOf<T> = T extends Message<infer TranslationsType, infer ParametersType> ? TranslationsType : never;
export type ParametersOf<T> = T extends Message<infer TranslationsType, infer ParametersType> ? ParametersType : never;

function trans<
    M extends Message,
    Parameters extends ParametersOf<M>,
    Translations extends TranslationsOf<M>,
>(message: M, parameters?: Parameters, locale?: keyof Translations): string {
    // ...
}
```

With the following constants file, which will be auto-generated by a Symfony command:
```ts 
const CRM_LIST_LOCATION: Message<Record<'en'|'fr', string>, { foo: string }> = {
    domain: 'messages',
    translations: {
        en: 'CRM List {foo}',
        fr: 'Liste CRM {foo}'
    },
}

const FOO_BAR: Message<Record<'en'|'fr'|'it', string>, { foo: string }> = {
    domain: 'messages',
    translations: {
        en: 'Foo bar (en)',
        fr: 'Foo bar (fr)',
        it: 'qsd',
    }
}
```

This way we can provide a very good DX: 
Capture d’écran 2022-12-23 à 08 54 13
image
image

tgalopin:
I'm fine with having a dumped file, but then it shouldn't be committed, otherwise it risks being unsynced from the actual translations.

With such an uncommited file, it becomes indeed easier to have great autocompletion without virtual module. I proposed a virtual module so that we are certain it will never be unsynced, but we may have checks to ensure that in dev instead.

Such generated file should IMO be in the assets directory of the project then, becoming very easy to instanciate. I'm thinking about something like `assets/translations.js` and have an option in the command-line to dump as Typescript (with advanced types). We could even provide the `trans` function through this file as well, to give it more typehinting capabilities (as you mentionned @Kocal).

Then in the code, this should be possible:

```typescript
import {trans, CRM_LIST_LOCATION} from './translations';

trans(CRM_LIST_LOCATION, { param1: 'value1' }, domain: 'security', locale: 'en');
```

Note that in order to avoid conflicts between domains, the internal data structure of a message will need to be something like:

```
const CRM_LIST_LOCATION = {
    messages: {
        en: 'CRM List {foo}',
        fr: 'Liste CRM {foo}'
    },
    user: {
        en: 'CRM List {foo}',
        fr: 'Liste CRM {foo}'
    },
}
```

tgalopin:
Also, @kbond for smaller projects, the `domain: "messages"` would be the default meaning you would be able to do:

```typescript
import {trans, CRM_LIST_LOCATION} from './translations';

trans(CRM_LIST_LOCATION, { param1: 'value1' });
```

Because of Webpack tree shaking, it doesn't matter that we import all translations in the generated file as it will be cleaned from unused translations anyway.

kbond:
@Kocal, that example looks pretty sweet!

> I'm fine with having a dumped file, but then it shouldn't be committed

Agree 100%, should we investigate hooking into Symfony's cache-warmup to create?

> or smaller projects, the domain: "messages" would be the default meaning you would be able to do:

I was thinking using strings as keys so you wouldn't need to import every translation constant. Again, I'm only thinking of a possible migration from bazinga to this - maybe that isn't or shouldn't be a concern.

tgalopin:
@kbond Having a dumped file always gives us the possibility to have an option for the dump to do that, something like `--as-strings` (name to be defined).

Kocal:
For the dumping command, I'm thinking we can get rid of `--typescript` options and generate both `.js` and `.d.ts` files at the same time.

This way, even people that does not use TypeScript would benefit from autocompletion through the IDE.

WDYT?

Kocal:
After a bit of work, I came up with something rather sexy, more specifically for the mandatory or non-mandatory translation parameters.

The two following files will be generated through a Symfony command:
```ts
// assets/translations.d.ts

type DomainType = string;
type LocaleType = string;
type NoParametersType = Record<string, never>;
type ParametersType = Record<string, string | number> | NoParametersType;

export interface Message<Domain extends DomainType, Locale extends LocaleType, Parameters extends ParametersType> {
    translations: Record<Domain, Record<Locale, string>>;
}

type DomainsOf<M> = M extends Message<infer Domain, LocaleType, ParametersType> ? Domain : never;
type LocaleOf<M> = M extends Message<DomainType, infer Locale, ParametersType> ? Locale : never;
type ParametersOf<M> = M extends Message<DomainType, LocaleType, infer Parameters> ? Parameters : never;

export function trans<M extends Message<DomainType, LocaleType, ParametersType>, P extends ParametersOf<M>>(
    ...args: P extends NoParametersType
        ? [message: M, parameters?: P, domain?: DomainsOf<M>, locale?: LocaleOf<M>]
        : [message: M, parameters: P, domain?: DomainsOf<M>, locale?: LocaleOf<M>]
);

// Those constants will be automatically generated
export const CRM_USER_FOO: Message<'messages', 'fr' | 'en', NoParametersType>;
export const FOO_BAR: Message<'messages', 'fr' | 'en', { foo: string; bar: string; count: number }>;
```

```js
// assets/translations.js
export function trans(message, parameters = {}, domain = 'messages', locale = null) {
    // ...
}

// Those constants will be automatically generated
export const CRM_USER_FOO = {
    translations: {
        messages: {
            fr: 'Crm user foo (fr)',
            en: 'Crm user foo (fr)',
        },
    },
};

export const FOO_BAR = {
    translations: {
        messages: {
            fr: '{foo} {bar}',
            en: '{foo} {bar}',
        },
    },
};
```

And for usage:
```js 
import { trans, CRM_USER_FOO, FOO_BAR } from './translations';

trans(CRM_USER_FOO);
trans(CRM_USER_FOO, {});
trans(CRM_USER_FOO, {}, 'foo');
trans(CRM_USER_FOO, {}, 'messages');
trans(CRM_USER_FOO, {}, 'messages', 'de');
trans(CRM_USER_FOO, {}, 'messages', 'fr');

trans(FOO_BAR);
trans(FOO_BAR, { count: 1 });
trans(FOO_BAR, { count: 1, foo: 'foo' });
trans(FOO_BAR, { count: 1, foo: 'foo', bar: 'bar' });
trans(FOO_BAR, { count: 1, foo: 'foo', bar: 'bar' }, 'qsd');
trans(FOO_BAR, { count: 1, foo: 'foo', bar: 'bar' }, 'messages');
trans(FOO_BAR, { count: 1, foo: 'foo', bar: 'bar' }, 'messages', 'de');
trans(FOO_BAR, { count: 1, foo: 'foo', bar: 'bar' }, 'messages', 'fr');
```

When using JavaScript, you won't have type-hinting issues (except with PHPStorm/WebStorm ?), but with TypeScript: 
image

tgalopin:
Very cool indeed! I very much like where this is going :)

> Agree 100%, should we investigate hooking into Symfony's cache-warmup to create?

Following this statement from @kbond I realized it's actually an amazing idea: we could create an always up-to-date file, without even having to implement any command, by hooking into the translations cache update process. Symfony already watches for changes on translations to dump them in PHP, we should definitely use that to generate the JS IMO. It would ease usage, integrates nicely in existing deployments and ensure always synced files.

Also, I was thinking about how we could put the generated files in var/cache and then have a file in assets that contains only an import from var/cache (and error handling): this would allow us to provide a better DX as this assets file would be committed and would warn the developer in the console that the translations are not generated when they are missing. This file would be committed to always be available in the other files.

WDYT?

Kocal:
> Following this statement from @kbond I realized it's actually an amazing idea: we could create an always up-to-date file, without even having to implement any command, by hooking into the translations cache update process. Symfony already watches for changes on translations to dump them in PHP, we should definitely use that to generate the JS IMO. It would ease usage, integrates nicely in existing deployments and ensure always synced files.

Yeah that could be nice, however we must really document that people needs to rebuild their assets (or use `encore dev-server` / `dev --watch`).
 
> Also, I was thinking about how we could put the generated files in var/cache and then have a file in assets that contains only an import from var/cache (and error handling): this would allow us to provide a better DX as this assets file would be committed and would warn the developer in the console that the translations are not generated when they are missing. This file would be committed to always be available in the other files.

This is a nice idea, by doing this way we keep the maximum of code in the Translator package and not in "userland". 
However I have two questions:
  1. How the file in `assets` would be created? By the end-user, or from a Symfony recipe? 
  2. Are you strictly speaking about `var/cache` or about `var/cache/<env>`?

tgalopin:
I was thinking about var/cache/env yes, it was a shortcut :)

I was thinking about using a recipe for the asset file indeed

> Yeah that could be nice, however we must really document that people needs to rebuild their assets

I agree with you, but I guess that won't be an issue because people will need to rebuild their assets to use the translations anyway.

Kocal:
Hi!

> I was thinking about var/cache/env yes, it was a shortcut :)
> 
> I was thinking about using a recipe for the asset file indeed

How would we know which environment to use?
What does this file looks like? Since we can't try/catch `export * from '../var/cache/<env>'` and try to load dumped translations from prod/dev env for examples.

IMO translations files should be dump in a non env-dependant directory, either `./assets/translations/` or `./var/translations/` for example.

tgalopin:
You're right, the env doesn't make sense, var/translations seems great to me

src/Translator/assets/jest.config.js
```diff
@@ -0,0 +1,7 @@
+const { defaults } = require('jest-config');
+const jestConfig = require('../../../jest.config.js');
+
+jestConfig.moduleFileExtensions = [...defaults.moduleFileExtensions, 'vue'];
+jestConfig.transform['^.+\\.vue$'] = ['@vue/vue3-jest'];
```
weaverryan: Is this needed?

src/Translator/assets/package.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "@symfony/ux-translator",
+    "description": "Symfony Translator for JavaScript",
+    "license": "MIT",
+    "version": "1.0.0",
+    "main": "dist/translator_controller.js",
+    "types": "dist/translator_controller.d.ts",
+    "peerDependencies": {
+        "intl-messageformat": "^10.2.5"
```
weaverryan, line 9: I'm curious about your thinking behind this as a peer dependency vs a normal dependency.

src/Translator/assets/src/translator.ts
```diff
@@ -0,0 +1,120 @@
...
+    ...args: P extends NoParametersType
+        ? [message: M, parameters?: P, domain?: D, locale?: LocaleOf<M>]
+        : [message: M, parameters: P, domain?: D, locale?: LocaleOf<M>]
+): string;
+export function trans<M extends Message<TranslationsType, LocaleType>, D extends DomainsOf<M>, P extends ParametersOf<M, D>>(
+    message: M,
+    parameters: P = {} as P,
+    domain: DomainsOf<M> = 'messages' as DomainsOf<M>,
+    locale: LocaleOf<M>|null = null
+): string {
```
weaverryan, line 123: Can you explain the double-export of `trans`? I'm sure this mysterious to me due to my lack lacking ts skills.

src/Translator/assets/src/translator_controller.ts
```diff
@@ -0,0 +1,12 @@
+/*
```
weaverryan, line 1: Is there any reason to include the word `controller` in the filename? The user will not even import this via `controllers.json`, right? This is a 100% normal, non-Stimulus package, correct?

src/Translator/assets/src/types.d.ts
```diff
@@ -0,0 +1,28 @@
...
+type ParametersType = Record<string, string | number> | NoParametersType;
+type RemoveIntlIcuSuffix<T extends string> = T extends `${infer U}+intl-icu` ? U : T;
+
+export type DomainsOf<M> = M extends Message<infer Translations, LocaleType> ? RemoveIntlIcuSuffix<keyof Translations>: never;
+export type LocaleOf<M> = M extends Message<TranslationsType, infer Locale> ? Locale : never;
+export type ParametersOf<M, D extends DomainType> = M extends Message<infer Translations, LocaleType>
+    ? (Translations[D] extends { parameters: infer Parameters }
+        ? Parameters
+        : never)
+    : never;
```
weaverryan: These types make my head spin a bit... did you write these by hand?

src/Translator/assets/translator-types-tests.ts
```diff
@@ -0,0 +1,80 @@
...
+            fr: 'Un simple message {parameter2}',
+            en: 'A simple message {parameter2}',
+        }
+    }
+};
+
+trans(MESSAGE_MULTI_DOMAINS_MULTI_LOCALES_WITH_PARAMETERS, {parameter1: 'foo', parameter2: 'bar'});
+trans(MESSAGE_MULTI_DOMAINS_MULTI_LOCALES_WITH_PARAMETERS, {parameter1: 'foo', parameter2: 'bar'}, 'messages');
+trans(MESSAGE_MULTI_DOMAINS_MULTI_LOCALES_WITH_PARAMETERS, {parameter1: 'foo', parameter2: 'bar'}, 'messages', 'fr');
+trans(MESSAGE_MULTI_DOMAINS_MULTI_LOCALES_WITH_PARAMETERS, {parameter1: 'foo', parameter2: 'bar'}, 'messages', 'en');
```
weaverryan: What's the purpose of this file?

src/Translator/composer.json
```diff
@@ -0,0 +1,51 @@
...
+        "symfony/filesystem": "^5.4|^6.0",
+        "symfony/string": "^4.4|^5.0|^6.0",
+        "symfony/translation": "^4.4|^5.0|^6.0",
+        "twig/twig": "^2.0|^3.0"
+    },
+    "require-dev": {
+        "symfony/framework-bundle": "^4.4|^5.0|^6.0",
+        "symfony/phpunit-bridge": "^5.2|^6.0",
+        "symfony/twig-bundle": "^4.4|^5.0|^6.0",
+        "symfony/var-dumper": "^4.4|^5.0|^6.0"
```
weaverryan:
```suggestion
        "symfony/var-dumper": "^5.4|^6.0"
```

Use this constraint for all of them - no need to support old stuff for a new package.

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,101 @@
...
+                'app.translation_4',
+                'app.translation_5',
+            ]) }}
+        {% endif %}
+    {% endblock %}
+
+
+.. note::
+    Specifying each translations keys one by one can be annoying, but it ensures that not every translations are loaded,
+    which could make your HTML page or your JavaScript files heavier.
```
weaverryan: Could/should there also be a way to dump an entire "domain" - e.g. maybe the user organizes things into smaller domains by feature, and they want to dump everything?

src/Translator/src/Intl/ErrorKind.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace Symfony\UX\Translator\Intl;
+
+/**
+ * Adapted from https://github.com/formatjs/formatjs/blob/590f1f81b26934c6dc7a55fff938df5436c6f158/packages/icu-messageformat-parser/error.ts#L9-L77
+ */
+final class ErrorKind
```
weaverryan, line 10: Did you adapt these classes? Or were they in the bazinga package?

weaverryan (commented):
This is very cool and promising - thanks for opening it! I left some notes.

My biggest concern is that there are a mountain of new, complex PHP classes... and I'm not sure where those came form. Same for a few .ts files and even the ts types already look very complex. I feel like I'm missing something about where those are coming from.

Also, I'm not yet following closely on the in-depth conversation in the comments. I don't personally have a lot of translation experience - so my value there might be limited... except for maybe some Webpack or JS parts :)

src/Translator/assets/translator-types-tests.ts
Kocal: This file was here to quickly tests `trans()` types, like we do in https://github.com/DefinitelyTyped/DefinitelyTyped (for example with [types for `assert`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/assert/assert-tests.ts)), but it's useless since https://github.com/JoshuaKGoldberg/eslint-plugin-expect-type is not installed.

Anyway, I've removed this file in favor of `src/Translator/assets/test/translator.test.ts`, this way we can have "real" + types test at the same time

src/Translator/composer.json
Kocal: Done, thanks

src/Translator/src/Intl/ErrorKind.php
Kocal: Classes in namespace `Symfony\UX\Translator\Intl` have been adapted from the https://github.com/formatjs/formatjs/tree/590f1f81b26934c6dc7a55fff938df5436c6f158/packages/icu-messageformat-parser JS package to PHP, as documented in the PHPDoc, but maybe it's not clear enough.

src/Translator/doc/index.rst
Kocal: The documentation is outdated, I'm rewriting the whole package (in order to have a better DX) and this won't apply anymore, since every translations from all possible domains will be dump to the filesystem.

And we won't have heavy JS bundles after compiling assets, thanks to the three-shaking.

src/Translator/assets/jest.config.js
Kocal: Nope, fixed!

src/Translator/assets/package.json
Kocal: I think it was the same logic of why `tom-select` is a peer dependency for Autocomplete package, why `cropperjs` is a peer dependency for Cropperjs package, `chartjs` for Chartjs package, ... 

But we can make it a normal dependency yea

src/Translator/assets/src/translator.ts
Kocal: This is called [Functions overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) in TypeScript, and I use it here to make argument `parameters` required or not, depending if the Message has parameters or not (`NoParametersType`): 

Capture d’écran 2023-01-14 à 18 58 44

Capture d’écran 2023-01-14 à 19 07 17

src/Translator/assets/src/translator_controller.ts
Kocal: Yes, this is a non-Stimulus package, but as we saw on Slack, it was necessary in order to be built by Rollup (`glob.sync('src/*/assets/src/*controller.ts')`). :/

The same trick is used for Turbo package, but it exports a `Controller` instance too.

src/Translator/assets/src/types.d.ts
Kocal: Yep, this was written by hand! This can be a bit complicated at first, but this is easier if you understand `infer`, `extends`, and `never` keywords.

For the given type: 
```ts
export type ParametersOf<M, D extends DomainType> = M extends Message<infer Translations, LocaleType>
    ? Translations[D] extends { parameters: infer Parameters }
        ? Parameters
        : never
    : never;
```
1. We define a type `ParametersOf<M, D extends DomainType>`, which use two generics:
    1. `M` (a `Message` or anything else) 
    2. and `D extends DomainType` (a translation domain)
2. If `M` is a `Message`: 
    1. then we fetch the translations by domain, thanks to `infer Translations`, 
    2. otherwise we stop, thanks to `never`
3. If `Translations[D]` (translations for the domain `D`) is a JS object containing one key `parameters`:
    1. then we returns the `Parameters`'s type, thanks to `infer Parameters`
    2. otherwise we stop, thanks to `never`

src/Translator/src/Intl/ErrorKind.php
tgalopin: Is that so that you can dump the translations in JS?

Kocal: Yes and no, without this I can still dump translations in JS, but Intl ICU messages won't have definition types for their parameters.

To support parameters extraction (and their possible values for `select`/`plural` control blocks), I had to migrate the JS package `icu-messageformat-parser` to PHP, since there is no existing library doing this (either `Intl` from PHP or `symfony/intl`).


image

Given this complex ICU message, we were able to extract the following parameters:
- `gender_of_host`, a string, which can takes `female`, `male` or anything else (because `other` is required) 
- `num_guest`, a number
- `host`, a string
- `guest`, a string

Kocal:
Hi everyone, this PR is ready for review again! 

The translations dumping system as JS constants, TypeScript types definitions, and locales fallback (which were missing from the previous iteration) have been implemented. 🎉 

/cc @weaverryan @tgalopin @kbond 

--- 

PS: I'm having issue with this CI check:

image

When running `yarn build` on my Mac (`rm -fr node_modules && yarn && yarn build`), I got a different `git status` output:
```
➜  symfony-ux git:(feat/translator) ✗ git status
On branch feat/translator
Your branch is up to date with 'origin/feat/translator'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/Autocomplete/assets/dist/controller.js
	modified:   src/LiveComponent/assets/dist/live_controller.js
	modified:   src/React/assets/dist/register_controller.d.ts
```

What can we do? Thanks :)

**EDIT:** my `yarn.lock` was outdated, removing it and reinstalling deps worked, thanks Ryan :)

src/Translator/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2020-2023 Fabien Potencier
```
Kocal: Should I change it to this?
```suggestion
Copyright (c) 2023-present Fabien Potencier
```

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+    $ npm run watch
+
+    # or use yarn
+    $ yarn install --force
+    $ yarn watch
+
+Usage
+-----
+
+When warming up the Symfony cache, all of your translations will be dumped as JavaScript into the ``var/translations/`` directory.
```
kbond, line 54: All locales or those enabled in `framework.enabled_locales`? `framework.enabled_locales: []` loads all, but `framework.enabled_locales: ['en', 'fr']` loads just en/fr. Would this follow the same logic?

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+For a better developer experience, TypeScript types definitions are also generated aside those JavaScript files.
+
+Then, you will be able to import those JavaScript translations in your assets.
+Don't worry about your final bundle size, only the translations you use will be included in your final bundle, thanks to the [tree shaking](https://webpack.js.org/guides/tree-shaking/).
+
+Configuring the default locale and locale fallbacks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, if the JavaScript translator is not configured, it will:
+- use the value of ``document.documentElement.lang`` or ``en`` (English) as the default locale
```
kbond: `document.documentElement.lang` is this set from `<html lang=...>`?

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+- not use any locale fallbacks
```
kbond: Should the fallback be `en` if the default isn't `en`?

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+
+This step is optional, but if you want to configure the default locale and locale fallbacks for the JavaScript translator, you must use the Twig function ``initialize_js_translator()``:
+
+.. code-block:: twig
+
+    {# templates/base.html.twig #}
+
+    {% block javascripts %}
+        {# It will use `Locale::getDefault()` or `en` as default locale #}
+        {{ initialize_js_translator() }}
```
kbond: Is this the only way to setup fallbacks? Could this be done as part of the cache:warmup (dump the fallbacks)?

src/Translator/src/Twig/TranslatorExtension.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function initializeJsTranslator(string $locale = null): string
+    {
+        $locale ??= class_exists(\Locale::class) ? \Locale::getDefault() : 'en';
+
+        return sprintf(
+            'window.__symfony_ux_translator = %s;',
+            json_encode([
+                'locale' => $locale,
+                'locales_fallbacks' => (object) $this->getLocalesFallbacks(...$this->translatorBag->getCatalogues()),
```
kbond: As stated elsewhere, `locales_fallbacks` isn't dependent on the passed locale (so unless I'm misunderstanding, it would be same for every request). Can we move this to the `cache:warmup` process?

kbond (commented):
This is looking really great!

Mostly reviewed the docs and had a few questions/comments.

src/Translator/doc/index.rst
Kocal: Yes! I will change it

src/Translator/doc/index.rst
Kocal: Currently this is the only way, but you're right I think we can configure fallbacks through `cache:warmup`.

src/Translator/doc/index.rst
Kocal: IMOthey are not related, or at least not here.

Kocal: IMO they are not related, or at least not here. 

The locale fallbacks (in this package) is represented as the following form: `{ locale: locale_fallback|null }`. 

If you have three locales `fr_FR`, `en_US` and `de_DE` in your apps, and the default locale is `en`, then this will be interpreted like this: 
```php
[
    'fr_FR' => 'fr',
    'fr' => 'en',
    'en' => null,
    'en_US' => 'en',
    'de_DE' => 'de',
    'de' => 'en',
]
```

If no translation is found for locale `fr_FR`, then we fallback for locale `fr`, and if necessary we fallback to `en`, and if still necessary then we display the translation's id. 

But maybe I missed your point...?

src/Translator/doc/index.rst
Kocal: The `TranslationsCacheWarmer` uses the `translator` service, so I believe it should uses enabled locales only.

I will double-check this later, and update tests if necessary, thanks :)

src/Translator/doc/index.rst
kbond: Or just add `(<html lang=...>)`

src/Translator/doc/index.rst
kbond: I was thinking, if `lang="fr"`, we'd want the `en` message to be used if a `fr` message doesn't exist. I'm a little rusty on how the fallbacks work though so maybe this already happens.

If we dump the fallbacks as suggested [here](https://github.com/symfony/ux/pull/616#discussion_r1141492637), then it wouldn't be an issue?

kbond: > not use any locale fallbacks

This is what threw me - I interpreted this as "no fallback will be used" so if the message key is `some.message`, the lang is `fr` and that key doesn't exist in the `fr` catalog, `some.message` would be rendered in js.

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+Don't worry about your final bundle size, only the translations you use will be included in your final bundle, thanks to the [tree shaking](https://webpack.js.org/guides/tree-shaking/).
+
+Configuring the default locale and locale fallbacks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, if the JavaScript translator is not configured, it will:
+- use the value of ``document.documentElement.lang`` or ``en`` (English) as the default locale
+- not use any locale fallbacks
+
+This step is optional, but if you want to configure the default locale and locale fallbacks for the JavaScript translator, you must use the Twig function ``initialize_js_translator()``:
```
weaverryan: Am I correct the the main way people SHOULD set up things is by setting the `<html lang="fr">`? If so, do we need to show a different way? I really like that, and I don't really like the `initialize_js_translator()` (right now, it feels like this is necessary and that it's doing some sort of important JavaScript setup... when in fact, none should be needed).

src/Translator/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+        {{ initialize_js_translator() }}
+
+        {# But you can manually specify which locale to use #}
+        {{ initialize_js_translator(locale=app.request.locale) }}
+    {% endblock %}
+
+Importing and using translations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you use the Symfony Flex recipe, you can import the ``trans()`` function and your translations in your assets from the file `assets/translator.js`.
```
weaverryan, line 71: I'm tempted to have a `Manual Installation` small section above where you show this file. We don't want people to actually FOLLOW that step in reality, but it would help show/explain how things work I think.

weaverryan:
> When running yarn build on my Mac (rm -fr node_modules && yarn && yarn build), I got a different git status output:

Hmm. Are the local changes just whitespace? Or something more significant? I'm on a mac also - my builds are clean locally on 2. last I checked.

weaverryan (commented):
I've only read the docs so far, but this is really, really interesting :).

And... so we also need to at least have some "plan" for import maps, which is coming very soon to Symfony. The big tricky part is the lack of tree shaking. You can import specific modules (e.g. `import { TRANSLATION_SIMPLE } from './translator'`), but the entire target `./translator` would be downloaded to make that happen. If you're using importmaps, you ARE opting out of tree shaking... but generally-speaking, it works well and there should be some reasonable path for UX translator to work with then. Any ideas about this?

src/Translator/doc/index.rst
kbond: Agreed! What about removing the twig function entirely to start (can always add later if there is a valid use-case)?

src/Translator/doc/index.rst
Kocal: I can confirm it uses locales enabled from `framework.enabled_locales`, as expected.

src/Translator/doc/index.rst
Kocal: I totally agree too, I'm removing it!

However, sometimes you can't use the lang from `<html lang="...">` because it is not always equals `app.request.local` (e.g.: for SEO reasons, where your app's locales does not follow the same format than `<html lang="...">`). 

I will add a new attribute `<html data-symfony-ux-translator-locale="...">` which will takes precedence over `<html lang="...">`.

src/Translator/doc/index.rst
Kocal: Yus!

Currently this file looks like this: 
```js
/*
 * This file is part of the Symfony UX Translator package.
 *
 * If folder "../var/translations" does not exist, or some translations are missing,
 * you must warmup your Symfony cache to refresh JavaScript translations.
 *
 */

import { trans, getLocale, setLocale, setLocaleFallbacks } from '@symfony/ux-translator';
import { localeFallbacks } from '../var/translations/configuration';

setLocaleFallbacks(localeFallbacks);

export { trans }
export * from '../var/translations';

```

src/Translator/doc/index.rst
Kocal: Yes, the fallback locale will be used if it exist and a translation is present.

If this is not the case, we will render `some.message` like the Symfony's Translator do.

I've added a test case for that: 
```ts 
setLocaleFallbacks({'fr_FR':'fr','fr':'en','en_US':'en','en_GB':'en','de_DE':'de','de':'en'});

const MESSAGE_FRENCH_ONLY: Message<{ messages: { parameters: NoParametersType } }, 'fr'> = {
    id: 'message_french_only',
    translations: {
        messages: {
            fr: 'Un message en français uniquement',
        }
    }
}

expect(trans(MESSAGE_FRENCH_ONLY, {}, 'messages', 'fr')).toEqual('Un message en français uniquement');
expect(trans(MESSAGE_FRENCH_ONLY, {}, 'messages', 'en' as 'fr')).toEqual('message_french_only');
```

Kocal:
Hi everyone, and thanks for the reviews! 

Your comments have been addressed, here are the updates:
- The locale fallbacks is now computed at cache warmup, it's located in `var/translations/configuration.js` file
- The Twig Extension (and the Twig dependency aswell) have been removed 


--- 

> And... so we also need to at least have some "plan" for import maps, which is coming very soon to Symfony. The big tricky part is the lack of tree shaking. You can import specific modules (e.g. `import { TRANSLATION_SIMPLE } from './translator'`), but the entire target `./translator` would be downloaded to make that happen. If you're using importmaps, you ARE opting out of tree shaking... but generally-speaking, it works well and there should be some reasonable path for UX translator to work with then. Any ideas about this?

Are you speaking about https://github.com/symfony/symfony/pull/48371? 

Unfortunately, this Translator package now **highly** relies on bundler's (Webpack or anything else) tree-shaking. 
It was not the case with the 1st implementation (loading translations you nedd through a Twig function), but Titouan proposed a solution with a better DX in https://github.com/symfony/ux/pull/616#issuecomment-1362554359 but it needs tree-shaking.

I don't know how works ImportMaps works, but if it does not support tree-shaking, then the Translator won't be usable. :/

tgalopin:
IMO we shouldn't care too much about importmaps for this PR.

My reasonning is that I think there are mostly two use cases for import maps (ie. two types of people using them):

1/ in dev environment it allows not to have to wait for webpack to build, meaning a faster iteration loop => we don't care that the whole translator is imported here, as long as it will work the same way when using Webpack.

2/ In prod environment, almost everyone with a slightly big app uses Webpack or another bundler. Webpack isn't only about building, it's also about productivity: many modules allow to inspect the code, to compile it, to provide DX tools, ... I generally compare it to the Symfony container being built: of course we could do without, but CompilerPass are hugely useful.

The only two cases where I can see someone using import maps in prod would be for small projects (=> fine to import the whole translator) or for projects having a large app but a tiny bit of front-end (in which case they most likely should just pass the translations from PHP as Stimulus values or similar).

I don't see a case where import maps are an issue, especially given how the DX is *so good* using the tree-shaking approach (you simply don't need to think about translations).

BTW I think a good tool to create in the future will be a Router ;)

Kocal:
> BTW I think a good tool to create in the future will be a Router ;)

That was planned :eyes:

weaverryan:
> The only two cases where I can see someone using import maps in prod would be for small projects (=> fine to import the whole translator) or for projects having a large app but a tiny bit of front-end (in which case they most likely should just pass the translations from PHP as Stimulus values or similar).

Yes, I think this is the key. I would be careful not to make it look like importmaps are not ok for production, because they are - even in large apps. So the point is that, if you're using importmaps, you are likely using Turbo + Stimulus to get your rich frontend, and so can pass translations as values to Stimulus.

kbond:
> BTW I think a good tool to create in the future will be a Router ;)

Ooo, like a replacement for https://github.com/FriendsOfSymfony/FOSJsRoutingBundle?

kbond (approved):
👍🏻 from me from a high-level perspective! I love the simplicity!

Kocal:
> > BTW I think a good tool to create in the future will be a Router ;)
> 
> Ooo, like a replacement for [FriendsOfSymfony/FOSJsRoutingBundle](https://github.com/FriendsOfSymfony/FOSJsRoutingBundle)?

Well, I had the same idea as Titouan, but I didn't want to spoil until this PR was merged 😛 

Now that the logic "warm => dump => import + tree-shaking" is mastered, that would be _easy_ to create a UX Router package.

tgalopin:
> Now that the logic "warm => dump => import + tree-shaking" is mastered, that would be easy to create a UX Router package.

Exactly my thinking as well :)

src/Translator/assets/src/formatters/formatter.ts
```diff
@@ -0,0 +1,236 @@
...
+        case 'cy':
+            return (1 == number) ? 0 : ((2 == number) ? 1 : (((8 == number) || (11 == number)) ? 2 : 3));
+        case 'ro':
+            return (1 == number) ? 0 : (((0 == number) || ((number % 100 > 0) && (number % 100 < 20))) ? 1 : 2);
+        case 'ar':
+            return (0 == number) ? 0 : ((1 == number) ? 1 : ((2 == number) ? 2 : (((number % 100 >= 3) && (number % 100 <= 10)) ? 3 : (((number % 100 >= 11) && (number % 100 <= 99)) ? 4 : 5))));
+        default:
+            return 0
+    }
+}
```
tgalopin: Missing EOF

src/Translator/assets/src/formatters/formatter.ts
```diff
@@ -0,0 +1,236 @@
...
+ * The right delimiter can be [ (exclusive) or ] (inclusive).
+ * Beside numbers, you can use -Inf and +Inf for the infinite.
+ *
+ * @see https://en.wikipedia.org/wiki/ISO_31-11
+ *
+ * @param id         The message id
+ * @param parameters An array of parameters for the message
+ * @param locale     The locale
+ */
+export function format(id: string, parameters: Record<string, string | number> = {}, locale: string): string {
```
tgalopin, line 50: From my initial understanding, this logic is basically private (it's used by your trans function but it's not meant to be used outside of the module)?

If that's true, maybe we can add a `@internal` annotation or something similar to discourage the direct use of this function outside of the package, to avoid BC breaks.

tgalopin (approved):
This is a great PR, thanks a lot for your work @Kocal

src/Translator/assets/src/formatters/formatter.ts
Kocal: Yes! This is something we can add or internal classes/functions.

Kocal: Looks like the equivalent tag for `@internal` is `@private`:
> or not meant for general use

See https://jsdoc.app/tags-private.html

Kocal:
(the PR has been rebased)

weaverryan:
Thank you @Kocal!

PR #617, opened 2022-12-18
imp: install PHP-CS-Fixer locally, allowing us to run in locally

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

After opening #616, the CI showed me a lot of style issues in the PHP code: 
image

I don't have PHP CS Fixer globally installed on my machine because that's a bad practice, and I don't want to fix those errors manually (the time is precious).

So, this PR adds PHP CS Fixer locally.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,6 +2,7 @@
     "name": "symfony/ux",
     "license": "MIT",
     "require-dev": {
-        "symfony/filesystem": "^5.2|^6.0"
+        "symfony/filesystem": "^5.2|^6.0",
+        "php-cs-fixer/shim": "^3.13"
     }
 }
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,9 +10,9 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: '8.0'
-                  tools: php-cs-fixer
+            - uses: ramsey/composer-install@v2
             - name: php-cs-fixer
-              run: php-cs-fixer fix --dry-run --diff
+              run: ./vendor/bin/php-cs-fixer fix --dry-run --diff
 
     coding-style-js:
         runs-on: ubuntu-latest


kbond (approved):
Makes sense.

composer.json
```diff
@@ -2,6 +2,7 @@
     "name": "symfony/ux",
     "license": "MIT",
     "require-dev": {
-        "symfony/filesystem": "^5.2|^6.0"
+        "symfony/filesystem": "^5.2|^6.0",
+        "friendsofphp/php-cs-fixer": "^3.13"
```
kbond: Will probably never be an issue here but might be better to use `php-cs-fixer/shim` to avoid any possible dep conflicts.

Kocal: Oh yeah, I forgot about this package.
Gonna change it ASAP.

Kocal: Done

weaverryan:
I think the recommended way is to install it into a `tools/` directory, right? https://github.com/PHP-CS-Fixer/PHP-CS-Fixer#installation

kbond:
The shim is new maybe? I only learned of it from the recipe (https://github.com/symfony/recipes/pull/1112) that was merged yesterday.

weaverryan:
This was the first I had heard of it. It seems like a good idea - is there some docs or mention of using this? And how it fits into best-practices?

Kocal:
As you know, installing this kind of tools into a separate composer.json (ie: in `tools/...` directory) allows to not pollute your app's deps/dev-deps with the tool's dependencies.

However I'm not a big fan of having multiple `composer.json` in a single project, and you can face some problems if you try to load/interpret your code with the tool (I remember to face this kind of issue because the are two autoloaders, but I may be wrong). 

This is something that can be _easily_ fixed by using a .phar, which can be installed through a shim-version of the package. 

To me the given installation methods for PHP-CS-Fixer are "wrong", `composer require --dev friendsofphp/php-cs-fixer` should download a `.phar` and that's all, no other dependencies, like PHPStan and Rector do since a long time.

Kocal:
Hi @weaverryan @kbond, is installing the shim still okay?

kbond (approved):
Works for me.

weaverryan:
This *does* make a lot of sense - thanks Hugo!

PR #618, opened 2022-12-18
[Live] Throw if a live component wrapped in other element

### LINKED ISSUE #612: [LiveComponent] Issues with custom Stimulus controller

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | #612
| License       | MIT

As how it works at the moment, it is intended for a live component to have a single root
as a live controller element. Maybe this can be checked earlier on server side, but that's not that easy.

### DIFF
--- a/src/LiveComponent/assets/src/Component/index.ts
+++ b/src/LiveComponent/assets/src/Component/index.ts
@@ -370,6 +370,10 @@ export default class Component {
         let newElement;
         try {
             newElement = htmlToElement(html);
+
+            if (!newElement.matches('[data-controller~=live]')) {
+                throw new Error('A live component template must contain a single root controller element.');
+            }
         } catch (error) {
             console.error('There was a problem with the component HTML returned:');
 


src/LiveComponent/assets/test/controller/render.test.ts
```diff
@@ -186,6 +186,22 @@ describe('LiveController rendering Tests', () => {
...
+            <div data-controller="some-other-not-live-controller">
+                <div ${initComponent(data)}></div>
+            </div>
+        `
+        );
+
+        test.expectsAjaxCall('get').expectSentData(test.initialData).init();
+
+        // @fixme how to assert this?
+        await waitFor(() => expect(test.component.render()).rejects.toThrow('I should fail'));
```
1ed: I don't know how to do this

weaverryan: Hmm. `test.component.render()` returns a `Promise`... and the `Promise, thanks to your exception, should get rejected. I've not done this before, but did you try `rejects.toEqual()`? https://jestjs.io/docs/tutorial-async#rejects

Btw, I like this error. If we get hung up on the test, I don't think that's too important.

1ed: Thanks, I tried a lot of variations, but none of them worked. Maybe a returned promise is not connected with that part of the code, that's the backend response which already resolved by then. Anyway I removed the test.

weaverryan:
Thanks @1ed!

PR #621, opened 2022-12-20
[Live] Adding ComponentRegistry as an easy way to fetch a Live Component

### LINKED ISSUE #612: [LiveComponent] Issues with custom Stimulus controller

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #612
| License       | MIT

Hi!

My solution for #612. It seems... to work perfectly well and it's dead-simple:

```js
const element = // find some Element that is your live controller rot

ComponentRegistry.get(element).then((component) => {
    // use the Component
    component.render()
});

// or use await
const component = await ComponentRegistry.get(element)l
component.render();
```

Cheers!

### DIFF
--- a/src/LiveComponent/assets/src/ComponentRegistry.ts
+++ b/src/LiveComponent/assets/src/ComponentRegistry.ts
@@ -1,0 +1,35 @@
+import Component from './Component';
+import { getElementAsTagText } from './dom_utils';
+
+const ComponentRegistry = class {
+    private components = new WeakMap<HTMLElement, Component>();
+
+    public registerComponent(element: HTMLElement, definition: Component) {
+        this.components.set(element, definition);
+    }
+
+    public unregisterComponent(element: HTMLElement) {
+        this.components.delete(element);
+    }
+
+    public getComponent(element: HTMLElement): Promise<Component> {
+        return new Promise((resolve, reject) => {
+            let count = 0;
+            const maxCount = 10;
+            const interval = setInterval(() => {
+                const component = this.components.get(element);
+                if (component) {
+                    resolve(component);
+                }
+                count++;
+
+                if (count > maxCount) {
+                    clearInterval(interval);
+                    reject(new Error(`Component not found for element ${getElementAsTagText(element)}`));
+                }
+            }, 5);
+        });
+    }
+};
+
+export default new ComponentRegistry();
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -17,8 +17,10 @@ import PollingPlugin from './Component/plugins/PollingPlugin';
 import SetValueOntoModelFieldsPlugin from './Component/plugins/SetValueOntoModelFieldsPlugin';
 import { PluginInterface } from './Component/plugins/PluginInterface';
 import getModelBinding from './Directive/get_model_binding';
+import ComponentRegistry from './ComponentRegistry';
 
 export { Component };
+export const getComponent = (element: HTMLElement): Promise<Component> => ComponentRegistry.getComponent(element);
 
 export interface LiveEvent extends CustomEvent {
     detail: {
@@ -104,6 +106,7 @@ export default class extends Controller<HTMLElement> implements LiveController {
             this.component.element.addEventListener(event, callback);
         });
 
+        ComponentRegistry.registerComponent(this.element, this.component);
         this._dispatchEvent('live:connect');
     }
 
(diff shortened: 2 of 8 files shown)


src/LiveComponent/assets/src/ComponentRegistry.ts
```diff
@@ -0,0 +1,35 @@
+import Component from './Component';
+import { getElementAsTagText } from './dom_utils';
+
+const ComponentRegistry = class {
+    private components = new Map<HTMLElement, Component>();
```
1ed: Shouldn't we use [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) here, as actually this will be global state holding references to components preventing them be garbage collected?

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -104,6 +105,7 @@ export default class extends Controller<HTMLElement> implements LiveController {
             this.component.element.addEventListener(event, callback);
         });
 
+        ComponentRegistry.registerComponent(this.element, this.component);
         this._dispatchEvent('live:connect');
```
1ed, line 110: Are there any use cases for listening to these events after adding the registry? Shouldn't we remove them for now?

src/LiveComponent/assets/src/ComponentRegistry.ts
```diff
@@ -0,0 +1,35 @@
...
+                if (component) {
+                    resolve(component);
+                }
+                count++;
+
+                if (count > maxCount) {
+                    clearInterval(interval);
+                    reject(new Error(`Component not found for element ${getElementAsTagText(element)}`));
+                }
+            }, 5);
```
1ed, line 30: To be honest, I don't know if I like this or not, IMHO it's kind of weird. What do you think about creating a message bus and dispatch events on it if something happens, so we can register listeners on a global bus related to live components, instead of waiting for something to happen. Something like this: 

```ts
ComponentRegistry.hook('live:connect', (component) => {})
```

Something similar what livewire does. https://github.com/livewire/livewire/blob/master/js/HookManager.js 
https://github.com/livewire/livewire/blob/master/js/Store.js#L116

Maybe I'm wrong and that's a different thing, it's kind of hard to get my head around this, as we have Stimulus as an extra layer.

Little bit off topic, but later we could implement some other stuff, that also could be useful, using this registry, like getting components by name (components should have a name imho, https://github.com/livewire/livewire/blob/master/js/Store.js#L116) or events https://laravel-livewire.com/docs/2.x/events. There is a great screencast about events in livewire https://laravel-livewire.com/screencasts/events
What do you think, maybe you have thought about this already?

1ed (commented):
Thank you for taking care of this. I would like to make a few comments if you allow me.

src/LiveComponent/assets/src/live_controller.ts
weaverryan: I wondered about that too, and you're probably correct. There might be some future need to check if a `Component` is connected or not, but we could easily add that to `Component` if needed later.

src/LiveComponent/assets/src/ComponentRegistry.ts
weaverryan: Yea... this might (?) be a different thing? First, about the Livewire events system, it's not something I've added in any way to LiveComponents yet. However, I assumed it was likely that it would be needed at some point.

> What do you think about creating a message bus and dispatch events on it if something happens, so we can register listeners on a global bus related to live components, instead of waiting for something to happen. Something like this:
>
> ```
> ComponentRegistry.hook('live:connect', (component) => {})
> ```

Two possible problems here. I'm not trying to shoot down your idea at all - just keep the conversation flowing. What I have in this PR might, indeed, not be the best solution. Anyways, the 2 possible problems:

A) How would you use this? For example, I believe your callback would be designed to be called when EVERY live component is connected. How would you figure out which `Component` is the correct one that you want if this is called multiple times? Perhaps this is where you're thinking a component name could come in handy? But if this code live inside of a Stimulus controller, perhaps it still makes more sense to match by element - e.g.

```
ComponentRegistry.hook('live:connect', (component) => {
    if (component.element === this.element) {
        // this is the one I want
    }
});
```

B) I think we still have the ordering problem. If you register this hook too later (after the LiveComponent has been connected), then you still won't be called.

Thanks for the convo!

1ed: Ohh, yes, you are totally right. Livewire is loaded differently, so something is working for them doesn't mean it will work for us too. This hook idea will not solve the problem at hand, so never mind. Your solution is simple and solves the problem :+1:

weaverryan:
Ok, final version of this is up. I simplified the export for userland to just the `getComponent` function.

PR #622, opened 2022-12-22
Update index.rst

wannevancamp:


kbond (approved):

kbond:
Good catch, thanks @wannevancamp.

PR #630, opened 2022-12-29
[Live] Fix model binding if a child has an empty select

### LINKED ISSUE #629: [Live] Bind to an empty select in a child is broken
Hi, the following test logs on console.error

```ts
    it('handles child with empty select', async () => {
        const test = await createTest({}, (data: any) => `
            <div ${initComponent(data)}>
                <div
                    ${initComponent({ selected: '' }, {}, {id: 'the-child-id'})}
                    data-model="foodName"
                    data-testid="child"
                >
                    <select data-model="selected"></select>
                </div>
            </div>
        `);

        await test.component.render()
        // @todo assert on console.error
        expect(test.element.querySelectorAll('select')).toHaveLength(1)
    });
```

```ts
 console.error
    Error initializing controller
    
    Error: The model name "selected" was never initialized. Did you forget to expose "selected" as a LiveProp? Available models values are: foodName
        at setDeepData (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/data_manipulation_utils.ts:71:19)
        at _default.set (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/Component/ValueStore.ts:52:21)
        at forEach (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts:65:38)
        at Proxy.forEach (<anonymous>)
        at _default.synchronizeValueOfModelFields (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts:33:60)
        at _default.attachToComponent (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts:15:14)
        at Component.addPlugin (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/Component/index.ts:88:16)
        at forEach (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/live_controller.ts:96:28)
        at Array.forEach (<anonymous>)
        at extended.initialize (/home/egabor/work/open-source/symfony-ux/src/LiveComponent/assets/src/live_controller.ts:95:17)
        at new Context (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1080:33)
        at Module.fetchContextForScope (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1296:27)
        at Module.connectContextForScope (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1282:34)
        at Router.scopeConnected (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1599:24)
        at ScopeObserver.elementMatchedValue (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1519:31)
        at ValueListObserver.tokenMatched (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:810:31)
        at TokenListObserver.tokenMatched (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:743:27)
        at /home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:737:42
        at Array.forEach (<anonymous>)
        at TokenListObserver.tokensMatched (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:737:20)
        at TokenListObserver.elementMatchedAttribute (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:726:18)
        at AttributeObserver.elementMatched (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:504:31)
        at ElementObserver.addElement (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:452:39)
        at ElementObserver.refresh (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:373:26)
        at ElementObserver.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:343:22)
        at AttributeObserver.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:480:34)
        at TokenListObserver.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:708:36)
        at ValueListObserver.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:791:36)
        at ScopeObserver.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1497:36)
        at Router.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1567:32)
        at Application.start (/home/egabor/work/open-source/symfony-ux/node_modules/@hotwired/stimulus/dist/stimulus.umd.js:1651:25)
        at processTicksAndRejections (node:internal/process/task_queues:96:5) {
      [stack]: 'Error: The model name "selected" was never initialized. Did you forget to expose "selected" as a LiveProp? Available models values are: foodName\n' +
```

There is a reproducer here: https://github.com/1ed/symfony-ux-demos/commit/9335701306f04b0d75a417a75987ab8b0b19d6e1

Using the child only without the parent context it works.

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #629
| License       | MIT

### DIFF
--- a/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts
@@ -61,7 +61,7 @@ export default class implements PluginInterface {
 
             // for select elements without a blank value, one might be selected automatically
             // https://github.com/symfony/ux/issues/469
-            if (element instanceof HTMLSelectElement && !element.multiple) {
+            if (element instanceof HTMLSelectElement && !element.multiple && element.length > 0) {
                 component.valueStore.set(modelName, getValueFromElement(element, component.valueStore));
             }
         })


src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts
```diff
@@ -61,7 +61,7 @@ export default class implements PluginInterface {
 
             // for select elements without a blank value, one might be selected automatically
             // https://github.com/symfony/ux/issues/469
-            if (element instanceof HTMLSelectElement && !element.multiple) {
+            if (element instanceof HTMLSelectElement && !element.multiple && element.length > 0) {
```
weaverryan, line 64: Totally unrelated to the PR, this is a bad spot for this code (and I'm the one that put it here) as it's about setting the value onto the component FROM the field... but the rest of the method is the opposite. Or, perhaps this whole thing should be called something more like "synchronize values and model fields" instead of "set value...".

Anyways, this PR makes sense: why try to "get" a value from a select that has zero options. But what if this `select` *did* have one option. Wouldn't the error still happen?

Really good debugging to see that `elementBelongsToThisComponent()` is really being called too early. THAT looks like the problem to me. And I'm wondering if we could fix it. Current idea:

A) In LiveController, initialize the children via the `initialize` method. Right now, children dispatch `live:connect` from `connect()` and we listen to that to discover that we have a child. But I think we could, instead, dispatch a `live:initialize` event from `initialize()` and listen to *that* to be notified of children (to make things a bit more complex, we would likely still ALSO need to listen on `live:connect` so that we're notified if any potential new children that are added later on - could be an edge case, but a controller element, later, could be initialized somewhere else, then appended into the parent element.)

B) Anyways, if we did this, it would be that, by the time `connect()` is called on the parent, we are aware of all of our children components. Then, perhaps plugins use the `connect` hook to run any initialization code (instead of having it that code directly inside the `attachToComponent` method.

Let me know if that makes any sense. And, it's all quite deep - so while I'd love if you made a PR with those changes, I don't expect it :).

1ed: > Anyways, this PR makes sense: why try to "get" a value from a select that has zero options. But what if this select did have one option. Wouldn't the error still happen?

Oops, right, the error still happens if the select has options.

> A) In LiveController, initialize the children via the initialize method. Right now, children dispatch live:connect from connect() and we listen to that to discover that we have a child. But I think we could, instead, dispatch a live:initialize event from initialize() and listen to that to be notified of children (to make things a bit more complex, we would likely still ALSO need to listen on live:connect so that we're notified if any potential new children that are added later on - could be an edge case, but a controller element, later, could be initialized somewhere else, then appended into the parent element.)

I'm not sure I understand it correctly. How should I get the children before they register themselves? Stimulus knows about descendant controllers or I should use `this.element.querySelectorAll('[data-controller~=live]')` (but that only returns a DOM element not a controller and not a component instance)?

AFAIK initialize and connect called in this order

```
live#initialize (component1)
live#connect (component1)
live#initialize (component2)
live#connect (component2)
```
and not like this (as I think it should be, as initialize should be called once, but because of lazy loading, I think, it is called multiple times?)

```
live#initialize
live#connect (component1)
live#connect (component2)
```
and not even like this

```
live#initialize (component1)
live#initialize (component2)
live#connect (component1)
live#connect (component2)
```
So I think currently we cannot get a descendant instance in a parent in these methods. Or am I missing something?

Doesn't we know the parent-child relation on server side already? Can't we use that information here somehow?

weaverryan: > and not even like this

> ```
> live#initialize (component1)
> live#initialize (component2)
> live#connect (component1)
> live#connect (component2)
> ```

Bah, I think you're right. I was TOTALLY thinking the order would be `initialize (component1)`, `initialize (component 2)`, then 2x connect. This, indeed, kills my idea.

> Doesn't we know the parent-child relation on server side already? Can't we use that information here somehow?

That's a really interesting idea. Yes, in theory, we could pass extra info to the parent Stimulus component controller about its children. We DO guarantee that every child component has a `data-live-id` attribute - https://github.com/symfony/ux/blob/c800637d20d51f5a84b995829e8315219b05e03f/src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php#L115-L118 - perhaps we could pass this into the parent Stimulus controller and use it to our advantage?

PR #633, opened 2023-01-03
Add autoload and replace rules for the root package

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | -
| License       | MIT

In order to use `symfony/ux` or more importantly a fork of `symfony/ux` as composer package we need these rules.
Whit this, one can use a fork like this:

```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/1ed/ux"
        }
    ],
    "require": {
        "symfony/ux": "dev-patches as 2.x-dev",
        "symfony/ux-autocomplete": "2.x-dev",
        "symfony/ux-live-component": "2.x-dev",
    }
}
```

There are 2 questions though

- [ ] should we add transitive dependencies here too?
- [ ] how to handle package.json and the flex recipe? should it have an own recipe?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,5 +3,37 @@
     "license": "MIT",
     "require-dev": {
         "symfony/filesystem": "^5.2|^6.0"
+    },
+    "replace": {
+        "symfony/ux-autocomplete": "self.version",
+        "symfony/ux-chartjs": "self.version",
+        "symfony/ux-cropperjs": "self.version",
+        "symfony/ux-dropzone": "self.version",
+        "symfony/ux-lazy-image": "self.version",
+        "symfony/ux-live-component": "self.version",
+        "symfony/ux-notify": "self.version",
+        "symfony/ux-react": "self.version",
+        "symfony/ux-swup": "self.version",
+        "symfony/ux-turbo": "self.version",
+        "symfony/ux-twig-component": "self.version",
+        "symfony/ux-typed": "self.version",
+        "symfony/ux-vue": "self.version"
+    },
+    "autoload": {
+        "psr-4": {
+            "Symfony\\UX\\Autocomplete\\": "src/Autocomplete/src/",
+            "Symfony\\UX\\Chartjs\\": "src/Chartjs/src/",
+            "Symfony\\UX\\Cropperjs\\": "src/Cropperjs/src/",
+            "Symfony\\UX\\Dropzone\\": "src/Dropzone/src/",
+            "Symfony\\UX\\LazyImage\\": "src/LazyImage/src/",
+            "Symfony\\UX\\LiveComponent\\": "src/LiveComponent/src/",
+            "Symfony\\UX\\Notify\\": "src/Notify/src/",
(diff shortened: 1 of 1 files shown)


weaverryan:
Hmm. I don't think we ever intended for someone to actually use `symfony/ux` itself - it's not even registered on Packagist (on purpose). If the intension is to use a fork, would the proper course be to fork the exact sub-tree split repository you need? - e.g. https://github.com/symfony/ux-autocomplete

kbond:
For development (working on a PR), I think using a fork of the subtree split wouldn't be ideal.

@1ed is this the intention?

1ed:
I think its OK to not publish a `symfony/ux` package on Packagist. My intention to deploy a working app, so I can use my fork until PRs are merged. For working on PRs the link utility works great, but that does not work in production. So I created a branch https://github.com/1ed/ux/tree/patches to merge patches and use that. I think to fork splits and do the same complicate things, especially for multiple components. I can add these replace rules to my fork locally only, but handling transitive dependencies and flex recipes complicate things even more. It would be nice to support this here, but I can accept that it's just a rare edge case and we don't want the extra work.

weaverryan:
If we did merge this, yea, I think it would basically be for more advanced situations, aimed at working on Symfony UX itself:

> * should we add transitive dependencies here too?

What do you mean?

> * how to handle package.json and the flex recipe? should it have an own recipe?

I think... nothing should be done - no flex recipe. But then, you would need to manually change all the paths to the packages in `package.json`, correct? It's a bit ugly...

1ed:
> What do you mean?

Components have transitive dependencies (e.g. Notify has mercure-bundle) which are needed them to work. With replace rules these dependencies are not installed as composer uses the root composer.json from symfony/ux, so we need to add these to that too.

> But then, you would need to manually change all the paths to the packages in package.json, correct?

When I used the configuration in the original post in an existing project, composer removed all ux packages and installed symfony/ux instead. The removal triggered flex to remove all the configuration of these packages and entries from package.json. I added them back manually. It's a little bit annoying but I can live with that, if we do not want to support this via flex, because I need to do it only once. But every time when I run composer update all entries in `assets/controllers.json` are removed by [PackageJsonSynchronizer.php#L226](https://github.com/symfony/flex/blob/c82477240111bfe41a1067c9f0ab91d40bafa5b6/src/PackageJsonSynchronizer.php#L226). This can fixed if we want by changing [PackageJsonSynchronizer.php#L237](https://github.com/symfony/flex/blob/c82477240111bfe41a1067c9f0ab91d40bafa5b6/src/PackageJsonSynchronizer.php#L237)

weaverryan:
Tbh, there is a growing list of complications with this - the flex/recipe problem, the transitive dependencies. If you need a fork for your app, I think forking the sub-tree split is the way to go. For local dev, we should use `link` and add any other tools that are needed to make that experience better.

Sorry - thanks for this, but I'm going to close :).

PR #611, opened 2022-12-14
[Autocomplete] typo in the Maker's template path

suasgn:
```
suabahasa ❯ php bin/console make:autocomplete-field

 The class name of the entity you want to autocomplete:
 > Pelaku

 Choose a name for your entity field class (e.g. PelakuAutocompleteField) [PelakuAutocompleteField]:
 >


In Generator.php line 176:

  Cannot find template "C:\workspace\git\saspol\vendor\symfony\ux-autocomplete\src\Maker./skeletons/AutocompleteField
  .tpl.php"


make:autocomplete-field
```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

PR #637, opened 2023-01-04
[Autocomplete] Fix maker autocomplete field template path

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | -
| License       | MIT


weaverryan:
Thanks - fixed already in #611... I just wasn't at a computer when I originally approved that (then, holidays).

PR #636, opened 2023-01-04
Add a new DataTable Component

KojoEnch:
This is a rework of [PR #605](https://github.com/symfony/ux/pull/605), replacing dependencies to Bootstrap & JQuery by a lightweight library, [Simple-DataTables](https://github.com/fiduswriter/Simple-DataTables).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #230 
| License       | MIT

I discussed the topic with @WebMamba from whom a large part of this code comes, and he was interested in exploring this way also. All credits to him :100: .

### Usage

To use Symfony UX DataTable, inject the ``TableBuilderInterface`` in your service and build a table.

**Minimalistic example :**

``` php
$table = $tableBuilder
        ->addData([
                ['id' => 1, 'pseudo' => 'Bob'],
                ['id' => 2, 'pseudo' => 'Kitty'],
            ])
            ->addOptions([
                'paging' => false,
                'searchable' => false,
            ])
        ->createTable();
```

**A more complete one :**
```php
#[Route('/usersList', name: 'app_users_list')]
public function index(TableBuilderInterface $tableBuilder, UserRepository $userRepository): Response
{
    $manifestPath = $this->getParameter('kernel.project_dir').'/public/build/manifest.json';
    $package = new Package(new JsonManifestVersionStrategy($manifestPath));

    $users = $userRepository->getUsersWithPostsAndCommentsCount();
    
    $table = $tableBuilder
                ->addData($users)
                ->addOptions([
                    'paging' => true,
                    'perPage' => 10,
                    'searchable' => true,
                    // Optionally import Simple-DataTables CSS file from CDN
                    'withPluginCss' => true,
                    // Optionally add custom CSS file, here through webpack entry and manifest.json
                    'withCss' => $package->getUrl('build/datatable.css'),
                    // Customize labels (@see https://github.com/fiduswriter/Simple-DataTables/wiki/labels)
                    'labels' => [
                        'placeholder' => 'Find a customer...',
                        'perPage' => 'Show {select} customers per page',
                    ],
                    // Customize layout (@see https://github.com/fiduswriter/Simple-DataTables/wiki/layout)
                    'layout' => [
                        'top' => '{search}{select}',
                        'bottom' => '{info}{pager}',
                    ],
                    // Customize columns (@see
                    'columns' => [
                        ['select' => 3, 'sort' => 'desc'],
                    ],
                    // Useful accessibility options
                    'rowNavigation' => true,
[...]

PR #649, opened 2023-01-15
DataTable component

### LINKED ISSUE #230: New component: Bootstrap Table?

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #230 
| License       | MIT

This PR come next to : #636 and #605.

After opening #605 lot of discussion happened. So here is one answer.

The thing is, it's really hard to find a perfect js library to handle every need. Some people want something fast, easy to set up, and powerful. Some other people don't want jquery at all in their projects and so on. So  I think it should not be the responsibility of symfony to choose what people want.

### So here is the idea: 
We should find a way to let the community build different components for a different library. And even if none of the components made by the community fit your project you should be able to create your own integration.

### How?:
What I propose here is to have a core component. This core component is not linked to any library. This core component simply gives an interface, an abstract builder, a twig extension, and some glue code. It gives a generic set of features like pagination, sort, and search...

And now if you wan't to create a new integration for a datatable library you like, you just need to write the code related to this library. So It's gonna be easy for the community to create new integration, or for you directly in to your project.

For exemple in this PR: I made the integration for the BootstrapTable library. For that, the only thing I need to do on the PHP side is to create a Renderer. A Renderer is an object called by twig, his role is to take the Table (a generic object defined in the core) and to return the HTML we need to render the table. Once we do that we can work on the js side inside a stimulus controller. This is all you need to create the integration to the library you want. 

### Try the BootstrapTable component:
1 - fork this branch
2 - create a new project and link the component to it
3 - add this line to your package.json in the dev-dependencies section: "@symfony/ux-bootstrap-table": "file:vendor/symfony/bootstrap-table/assets"
4 - add this lines to your controllers.json : 
``` json
{
    "controllers": {
        "@symfony/ux-chartjs": {
            "chart": {
                "enabled": true,
                "fetch": "eager"
            }
        }
    },
    "entrypoints": []
}
```
5- yarn install --force
6- yarn dev
7- use the bootstrapbuilder in your controller
```php
#[Route('/', name: 'app_index')]
    public function index(Request $request, BootstrapTableBuilder $tableBuilder): Response
    {
        $table = $tableBuilder
            ->handleRequest($request)
            ->addColumns([
                ['title' => 'ID', 'field' => 'id'],
                ['title' => 'Pseudo', 'field' => 'pseudo'],
                ['title' => 'Age', 'field' => 'age', 'sortable' => 'true'],
                ['title' => 'Gender', 'field' => 'gender']
            ])
            ->addData([
                ['id' => 1, 'pseudo' => 'Bob', 'age' => 23, 'gender' => 'M'],
                ['id' => 2, 'pseudo' => 'Kitty', 'age' => 36, 'gender' => 'F'],
                ['id' => 3, 'pseudo' => 'Spongy', 'age' => 34, 'gender' => 'M'],
                ['id' => 4, 'pseudo' => 'Slurp', 'age' => 20, 'gender' => 'F'],
                ['id' => 5, 'pseudo' => 'Zoom', 'age' => 23, 'gender' => 'M'],
            ])
            ->addOptions([
                'search' => 'true',
            ])
            ->setSort('age', 'DESC')
            ->createTable()
        ;

        return $this->render('index.html.twig', [
            'table' => $table,
        ]);
    }
```
8 - render the table
```twig 
{{ render_table(table) }}
```

### Create your own integration to the library you want
1 - fork this branch
2 - install the DataTable/Core component
3 - create the renderer 
```php
class CustomRenderer implements RendererInterface
{
    public function render(Environment $environment, TableInterface $table): string
    {
        $dto = new StimulusControllersDto($environment);
        $dto->addController('my_controller', $table->renderView());

        return '<table '.$dto.'></table>';
    }
}
```
4 - overwrite the getRenderMethod to return your CustomRenderer className
```php
class CustomTableBuilder extends AbstractTableBuilder
{
    public function getRenderer(): string
    {
        return BootstrapTableRenderer::class;
    }
}
```
5 - create your stimulus controller
```ts
    readonly rowsValue: any;
    readonly columnsValue: any;
    readonly optionsValue: any;

    static values = {
        rows: Array,
        columns: Array,
        options: {},
    };

    connect() {
        if (!(this.element instanceof HTMLTableElement)) {
            throw new Error('invalid Element');
        }

       renderMyTable();
    }
}
```
6 - and you are done! You have a dataTable with back-end support for pagination, search, sort, and many other features. I the js library you always wanted.

### End
Thanks for your attention. As in the first PR I will be happy to hear what you think about this approach. I really think we can make a great component! So let's keep moving forward 😁


weaverryan (commented):
I think this is the right approach: a core library for building tables that could then be used for Bootstrap tables, data tables or even as just a tool inside your own live component. Though, I think we could still create just one package, and put a few default table implementations into that one package. The user would choose which to activate by enabling the correct controller in `controllers.json`. Though, that's a detail that we could decide later.

I need to actually play with this so I can look at the specifics of the implementation. I'll try to do that soon :)

WebMamba:
The reason I separate things into deferents packages is to ease the management of js dependencies. Users can activate the controllers they want but if they need to manually install the dependencies they need, we lost the promise to have everything set up in one command.

WebMamba:
@weaverryan friendly ping here 😁

weaverryan:
Hi @WebMamba!

Thanks for your energy and sorry for my slowness. Data table is a complex problem. But this might be the right approach. To validate it, I think we should, in addition to the jQuery data tables use-case, build out a use-case where the core library is used without any JavaScript. My advice:

A) Build a test project to play with and push it to GH (you described the steps, but it still took me a bit to get all the pieces put together).

B) On this test project, build a Bootstrap data table on one page - e.g. `/bootstrap`.

C) Also build a no-JS table on another page - e.g. `/classic`. If the "Core" library is well-written, it should be a useful tool even if I simply want to render a table manually in Twig with a few nice features, like changing sort direction or having a filter. Here is some invented code as an example:

```twig
<form method="get">
    <input type="search" name="search">

    <button type="submit">
</form>

<table>
    <thead>
        {% for header in table.header %}
            <th>
                <a href="{{ table.changeSort(header.id)">
                    {{ header.text }} {% if header.isSorted %}<span class="fa-{{ header.sortDirection }}"></span>
               </a>
            </th>
        {% endfor %}
    </thead>
    ... something similar for the columns
</table>

{% if table.hasPrevious %}
    <a href="{{ table.previousPage }}">Prev</a>
{% endif %}
{% if table.hasNext %}
    <a href="{{ table.nextPage }}">Next</a>
{% endif %}
```

I "think" the value of this component is that it helps (A) manage a set of data including (B) pagination, fields, sorting & searching and maybe even (C) features like "export to CSV".

For the above or jQuery data tables, we'll need support for things like (1) pagination via Doctrine query builders and (2) searching/sorting via query builder, (3) grabbing info from the request - like query parameters - and changing sort/etc. There's a LOT of important stuff still missing.

Is it worth continuing? I can't promise you'll be able to get this to a point where we think it's worth being a UX package (because the problem is so complex). But if you're willing to put in some work, it's worth a shot - and this could fall to a 3rd-party bundle at worse.

WebMamba:
Thanks for your response Ryan! I closed that for now, I will try to make soon a Component on a separate repo, in order to have more flexibility to try things 😁 Cheers.

PR #653, opened 2023-01-19
[Turbo] Fix checking "topics" option in Broadcaster

wooky:
Mercure Broadcaster class was checking for the "topic" option, which is incorrect. This fixes that by checking for "topics" option instead.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT


weaverryan:
Thanks @wooky!

PR #663, opened 2023-01-24
Update _markdown.html.twig

94noni:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT

Small PR to "mimic" the showcase example


weaverryan:
Thanks :). I wish we didn't need to duplicate the code to these code snippets... but it's tricky to create a more "readable" version of the code in some cases (e.g. hide less important details for demo purposes). But then, we get duplication / de-sync

Issue #528, opened 2022-11-04
[RFC] Handling Data with Bad Types (i.e. that fails hydration)


weaverryan:
Hi!

Consider this component:

```php
#[AsLiveComponent('counter')]
final class CounterComponent
{
    #[LiveProp(writable: true)]
    public int $count = 0;
}
```

In the template, it's bound to an input:

```
<input data-model="count">
```

What should happen if I enter `apple` into that box? (Note, yes we could use `<input type="number">`, but there are other examples , like a `DateTime` property where an invalid date time string is passed).

I can only think of two options:

A) We advise to NOT use property-types for writeable properties. The idea is that the component represents the "client-side data". And so if you can type `apple` into a field, then we should allow for `apple` to be entered on the property. Then, you use `#[Assert\....]` validation to validate things.

B) If there is a type mismatch, we reject the new value then, somewhere, expose an "error" that the user can render (e.g. `errors.get('count')` would have a message). 

(B) is attractive, but tricky. It would require:

1) On the Ajax call, in addition to the "latest data" (e.g. the latest `count` value), we would also need to send all of the "original data". This is needed because, if count was previously `10` and the user types in `apple`, we need to know the original value (`10`) so we can set it onto the property (since we will not be setting the string `apple`). If we didn't do this, the invalid property would go back to its original default value (if it has one, e.g. 0, which would look weird) or, worse, it would remain "uninitialized" and result in an error.

2) On the frontend, we would need to be smart enough to realize that if we send a value of `apple` for `count`, when that Ajax call finishes, because `count` was invalid, that data is still in an "unsynchronized" state. This means that the the `<input` field should *not* be updated to hold the latest value from the server (which would be the original `10` at this point).

So, solution (B) is doable. But we should be sure that it's not overkill - i.e. that solution (A) doesn't work well enough or something else.

PR #670, opened 2023-02-01
[LiveComponents] Enhanced & formalized data hydration system

### LINKED ISSUE #425: [Live][RFC] Defining `LiveProp` writable/exposed permutations for non-scalar types
Had a lengthy discussion with @weaverryan about how writable/exposed works with objects/arrays. We came to the conclusion that we should formally define (and implement) how this should work.

Here are our conclusions but open for discussion!

For objects/arrays:
1. `LiveProp(writable: true)`: any manipulation is possible
2. `LiveProp(writable: false, exposed: ['k1', 'k2'])`: only property path `k1` and `k2` can be modified.

For Doctrine entity's, we'd have a bit of alternate logic because there could be a scenario where you want fields to be editable but not allow the entity to be _swapped for another_:

1. `LiveProp(writable: true)`: any property can be manipulated _AND_ can be swapped for another (I believe this is the behaviour now)
2. `LiveProp(writable: false, exposed: ['prop1', 'prop2'])`: Only `prop1` & `prop2` can be modified - the entity _CANNOT_ be swapped for another
3. `LiveProp(writable: false, exposed: ['@id', 'prop1', 'prop2'])`: Only `prop1` & `prop2` can be modified _AND_ the entity can be swapped for another. This adds a special `@id` syntax that let's the hydrator know it can be _swapped_.

I believe this does make `writable: true, exposed: ['k1', 'k2']` not really valid (`writable: true` means anything about the array/object can be manipulated). So I think we can deprecate `exposed` and allow `writable` to be an `array`:

1. For arrays/non-entity objects - `LiveProp(writable: ['k1', 'k2'])` would imply only property paths `k1`/`k2` can be modified.
2. For entity objects - `LiveProp(writable: ['k1', 'k2'])` would imply only property path `k1`/`k2` can be modified and the entity _CANNOT_ be swapped for another.
3. For entity objects - `LiveProp(writable: ['@id', 'k1', 'k2'])` would imply only property path `k1/`k2` can be modified and the entity _CAN_ be swapped for another.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #425 - may enhance to also address #528 
| License       | MIT

Hi!

This looks bigger than it is. In short, this makes sure that LiveComponents correctly hydrates/dehydrates all the different types of data we want to support (e.g. scalars, persisted entities, non-persisted entities, arrays, etc).

## End-user changes:

* A) `exposed` is now merged with `writable`:

```diff
// scalars are still simply "writable: true"
#[LiveProp(writable: true)]
public string $value = '';

// "exposed" now also goes into writable
-#[LiveProp(exposed: ['email', 'plainPassword'])]
+#[LiveProp(writable: ['email', 'plainPassword'])]
public User $user;
```

* B) arrays for `LiveProp` are now fully supported and solid. 

```php
    // allow only "character" to be writable 
    #[LiveProp(writable: ['character'])]
    public array $options = ['show' => 'Arrested development', 'character' => 'Michael Bluth'];

    // allow anything in the array to change
    #[LiveProp(writable: true)]
    public array $options = ['show' => 'Arrested development', 'character' => 'Michael Bluth'];
```

* C) A `writable` path can now include an object that needs to be (dehydrated) - e.g. `LiveProp(writable: ['createdAt'])` would work, even if `User.createdAt` is a `DateTime` object.

* D) Non-persisted entities can now be used on `LiveProp`: they are sent through the serializer for (de)hydration

TODOS:
* [x] Make the special-case "hydration" code work for writable paths also
* [x] update ux.symfony.com code
* [x] Add missing license headers, experimental status, `final` and `@internal` where needed on new classes
* [ ] possibly (in this PR) handle "bad hydration" situations - #528, but likely out of scope

Cheers!


src/LiveComponent/composer.json
```diff
@@ -38,6 +38,7 @@
         "symfony/dependency-injection": "^5.4|^6.0",
         "symfony/form": "^5.4|^6.0",
         "symfony/framework-bundle": "^5.4|^6.0",
+        "symfony/maker-bundle": "^1.48",
```
weaverryan: Seems to fix an issue with Foundry, where its `make:factory` command was exploding without this.

kbond:
```suggestion
```

The issue in foundry has been fixed

weaverryan: bam

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -41,126 +46,121 @@ public function __construct(
     ) {
     }
 
-    public function dehydrate(MountedComponent $mounted): DehydratedComponent
+    public function dehydrate(object $component, ComponentAttributes $attributes, LiveComponentMetadata $componentMetadata): array
```
weaverryan, line 50: Simplified, in part for testing. The big item here is: "dehydration" returns the final array of dehydrated props, that's it, nothing more.

That `LiveCompomentMetadata` is a new idea... again partially motivated by testing so that we could pass IN the metadata about each `LiveProp` (including `LiveProp` data itself, but also some reflection info about the "type" of the property) instead of it grabbing that information itself. This class is already so complex, I was trying to limit what it did.

The new `LiveComponentMetadata` DOES leave us with a, kind of odd system where "Twig components" Has a `ComponentMetadata`... then `Live Components" has a totally different additional `LiveComponentMetadata` which contains live-specific info. But, this is all quite internal anyways.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -41,126 +46,121 @@ public function __construct(
...
+                }
+                $dehydratedValue = $component->$method($dehydratedValue);
+            } elseif (\is_object($rawPropertyValue)) {
+                $dehydratedValue = $this->normalizer->normalize(
+                    $dehydratedValue,
+                    'json',
+                    [self::LIVE_CONTEXT => true, self::LIVE_IS_PROP_WRITABLE => $propMetadata->isIdentityWritable()]
+                );
             }
+            $this->preventWritableNonScalarIdentities($dehydratedValue, $propMetadata);
```
weaverryan: An explicit decision. We DO allow object-properties to dehydrate to an array (e.g. think of a DTO). But if you do this, you cannot set that property to `writable: true`. You CAN specify certain properties that you want to allow writable -like `writable: [foo, bar]` - but you can't make the whole object "replaceable".

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -41,126 +46,121 @@ public function __construct(
...
+                    $rawPropertyValue,
+                    $this->adjustPropertyPathForData($rawPropertyValue, $path)
+                );
+
+                if (\is_object($pathValue)) {
+                    $pathValue = $this->normalizer->normalize(
+                        $pathValue,
+                        'json',
+                        [self::LIVE_CONTEXT => true, self::LIVE_IS_PROP_WRITABLE => $propMetadata->isIdentityWritable()]
+                    );
```
weaverryan, line 117: Explicit support for having a writable path that is ALSO an object - e.g. you have a `$post` `LiveProp` and you set `LiveProp(writable: ['createdAt'])` where `createdAt` is a `DateTime` object. That is allowed.

weaverryan:
Low and high tests will fail until this is merged, due to the small change in Twig components that Live components relies on.

PR #672, opened 2023-02-02
[Autocomplete] Fix setData in AutocompleteField

### LINKED ISSUE #667: [Autocomplete] SetData in AutocompleteField
Hello everyone,
I have **unmapped** field with type `AutocompleteField`
inside `preSetData` event
i trying to `setData` to this is field:
```
$form->add('reference',ReferenceAutocompleteField::class,[
    'mapped' => false,
    'data' => $data
]);
```
but i found this is error:
`The form's view data is expected to be a "App\Entity\Reference", but it is a "string". You can avoid this error by setting the "data_class" option to null or by adding a view transformer that transforms "string" to an instance of "App\Entity\Reference".
`

Aziz403:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #667  
| License       | MIT

After debugging the code again, I think this is the best way to bypass the error,
unset `data_class` option from the `autocomplete` field

### DIFF
--- a/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
+++ b/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
@@ -43,7 +43,7 @@ final class AutocompleteEntityTypeSubscriber implements EventSubscriberInterface
         // pass to AutocompleteChoiceTypeExtension
         $options['autocomplete'] = true;
         $options['autocomplete_url'] = $this->autocompleteUrl;
-        unset($options['searchable_fields'], $options['security'], $options['filter_query']);
+        unset($options['searchable_fields'], $options['security'], $options['filter_query'], $options['data_class']);
 
         $form->add('autocomplete', EntityType::class, $options);
     }


weaverryan:
Hi!

Thanks for the debugging and the PR. To understand the issue and validate your solution, I want to reproduce the problem locally. I was trying to use your reproducer - https://github.com/Aziz403/test_ux-autocomplete. I did have to add a few things that seemed missing (e.g `symfony/ux-autocomplete` package was missing) - but that was no problem, and the updated code is here: https://github.com/weaverryan/test_ux-autocomplete

However, I can't reproduce the issue. On the homepage, the Ajax autocomplete works just fine. And if I submit, the data dumps correctly and looks correct.

Can you help me understand how to reproduce the issue?

Aziz403:
Yes sir,

i think my bad english don't help.
i see your repository and you canont found the problem because i'm forget to commit the fixutres , but i think you can see it with your fixtures and this is [PR](https://github.com/weaverryan/test_ux-autocomplete/pull/1).

to be clear it's not about unmapped fields or events !
the problems happens when found two things:

1. **AutoCompleteFiled** like [GroupAutocompleteField](https://github.com/weaverryan/test_ux-autocomplete/blob/main/src/Form/GroupAutocompleteField.php) in this is situation.
2. **Set field data from options** like :
```
$form->add('personGroup',GroupAutocompleteField::class,[
    'data' => $data // when data was `not null` the bug rendered
]);
```

Aziz403:
Hi @weaverryan,
I just want to remind you that the bug still exists !

norkunas:
`AutocompleteEntityTypeSubscriber` is now deprecated, so I think this should be closed

weaverryan:
Yup! If you have issues with the new system, let us know. And sorry for letting this PR sit

PR #674, opened 2023-02-02
[Website] Simplify some contents in the Swup example

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | -
| License       | MIT

Looking at the code on https://ux.symfony.com/swup it looks like you need to change things in your controller/templates to make it work. Maybe it's because of the naming of things. In practice, you don't have to change anything.

### DIFF
--- a/ux.symfony.com/code_snippets/_swup.html.twig
+++ b/ux.symfony.com/code_snippets/_swup.html.twig
@@ -1,8 +1,6 @@
-{# initialize the swup controller, then render normal links! #}
-<div
-    {{ stimulus_controller('symfony/ux-swup/swup') }}
-    id="swup"
->
+{# you only need to add the call to the Stimilus controller;
+   swup doesn't require any other changes in your template #}
+<div {{ stimulus_controller('symfony/ux-swup/swup') }}>
 
     {% for package in pager %}
         <a href="{{ path(package.route) }}" class="col-12">
--- a/ux.symfony.com/code_snippets/_SwupController.php
+++ b/ux.symfony.com/code_snippets/_SwupController.php
@@ -11,18 +11,19 @@ use Symfony\Component\Routing\Annotation\Route;
 
 class _SwupController extends AbstractController
 {
-    #[Route('/swup/{page<\d+>}', name: 'app_swup')]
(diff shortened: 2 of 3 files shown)


weaverryan (commented):
I understand: tweaked comments, and the example is made to be a bit more realistic. I like it :) 

The downside is that this doesn't quite match the page now. Perhaps we should make the URL something like `/swup/packages/{id}` and then also update the REAL code behind the page to match?

Basically, if we show them a controller that uses the URL `/packages/{page}`, then the demo is taking them to `/swup/{page}`, that might also be confusing.

javiereguiluz:
Thanks for the review. To be honest, I don't know exactly how to proceed because I'm pretty new to UX. Maybe you can take over this PR? Or maybe even close it if "fixing" this produces a worse result than the current situation. Thanks!

ux.symfony.com/code_snippets/_swup.html.twig
```diff
@@ -1,8 +1,6 @@
-{# initialize the swup controller, then render normal links! #}
-<div
-    {{ stimulus_controller('symfony/ux-swup/swup') }}
-    id="swup"
->
+{# you only need to add the call to the Stimilus controller;
+   swup doesn't require any other changes in your template #}
+<div {{ stimulus_controller('symfony/ux-swup/swup') }}>
```
javiereguiluz, line 3: How is the `symfony/ux-swup/swup` string formed?

* I guess `symfony/` is hardcoded
* `ux-swup/` not sure where it comes from
* `/swup` is the name of the controller action?

weaverryan: It's all "hardcoded". Or, to say it better, when you install the package, the package provides you with a Stimulus controller. And, it gets to choose what name to use for that controller.

In some packages, this info is a bit hidden - e.g. with chart.js, you use `{{ render_chart(chart) }}` and that outputs the element with the correct `data-controller` for you.

Knowing this, what would be your ideal way for this to look/work? I agree that this string looks a bit random.

weaverryan:
Thanks Javier!

PR #698, opened 2023-02-20
Fix copy-paste typo

coramba:
Current symfony/ux-vue documentations refers to React instead of Vue


kbond (approved):

weaverryan:
Thanks for fixing this! As fast as you were, someone else was even faster - but I really appreciate it.

PR #710, opened 2023-02-28
feat(twig): define the behaviour of boolean attribute values

### LINKED ISSUE #709: [TwigComponent] Conditionally pass HTML attributes
Sometimes I want to conditionally pass an HTML attribute to a Twig component, but that doesn't seem to work yet.

**Example**

```twig
<!-- components/button.html.twig -->

<button {{ attributes }}>
  {{ label }}
</button>
```

```twig
<!-- home/index.html.twig -->

{{ component('button', {
    label: 'Add feature',
    disabled: not feature.isEnabled,
}) }}
```

will always result in

```html
<!-- index.html -->

<button disabled>
  Add feature
</button>
```

No matter what value is passed (`''`, `false`, `null`), the `disabled` attribute will always be displayed.

My current workaround is this:

```twig
<!-- components/button.html.twig -->

<button {{ attributes ~ (isDisabled ? ' disabled') }}>
  {{ label }}
</button>
```

But this feels like a very dirty hack. I know that passing `null` will omit the value and just render the attribute, but I wish I was able to conditionally omit the whole attribute by passing `false` for example.

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #709
| License       | MIT

`true` attribute values now render just the attribute name, `false` excludes it entirely.

```twig
{# component template: #}
<div{{ attributes }}></div>

{# calling template: #}
{{ component('my_component', { autofocus: true, disabled: false }) }}

{# renders as: #}
<div autofocus></div>
```

~I've left the `null` behaviour as-is which is the same as `true`. This does feel weird so perhaps we should deprecate passing `null` as @AmProsius suggests in #709.~ Using `null` is now deprecated.

### DIFF
--- a/src/TwigComponent/tests/Unit/ComponentAttributesTest.php
+++ b/src/TwigComponent/tests/Unit/ComponentAttributesTest.php
@@ -26,7 +26,7 @@ final class ComponentAttributesTest extends TestCase
             'class' => 'foo',
             'style' => 'color:black;',
             'value' => '',
-            'autofocus' => null,
+            'autofocus' => true,
         ]);
 
         $this->assertSame(' class="foo" style="color:black;" value="" autofocus', (string) $attributes);
@@ -86,5 +86,29 @@ final class ComponentAttributesTest extends TestCase
             'data-live-data-value' => '{}',
             'data-foo-name-value' => 'ryan',
         ], $attributes->all());
+    }
+
+    public function testBooleanBehaviour(): void
+    {
+        $attributes = new ComponentAttributes(['disabled' => true]);
+
+        $this->assertSame(['disabled' => true], $attributes->all());
+        $this->assertSame(' disabled', (string) $attributes);
+
+        $attributes = new ComponentAttributes(['disabled' => false]);
+
+        $this->assertSame(['disabled' => false], $attributes->all());
+        $this->assertSame('', (string) $attributes);
+    }
+
+    /**
+     * @group legacy
+     */
+    public function testNullBehaviour(): void
+    {
+        $attributes = new ComponentAttributes(['disabled' => null]);
+
+        $this->assertSame(['disabled' => null], $attributes->all());
+        $this->assertSame(' disabled', (string) $attributes);
     }
 }
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.8.0
+
+-   `true` attribute values now render just the attribute name, `false` excludes it entirely.
+
 ## 2.7.0
 
 -   `PreMount` and `PostMount` hooks can now return nothing.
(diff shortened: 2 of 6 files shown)


src/TwigComponent/tests/Unit/ComponentAttributesTest.php
```diff
@@ -87,4 +87,22 @@ public function testCanAddStimulusController(): void
...
+
+        $attributes = new ComponentAttributes(['disabled' => false]);
+
+        $this->assertSame(['disabled' => false], $attributes->all());
+        $this->assertSame('', (string) $attributes);
+
+        $attributes = new ComponentAttributes(['disabled' => null]);
+
+        $this->assertSame(['disabled' => null], $attributes->all());
+        $this->assertSame(' disabled', (string) $attributes);
```
weaverryan, line 112: I agree that this is wrong (it should have the same behavior as `false` - but it should be allowed). The other 2 cases are correct.

Any thoughts on gently (from a BC pov) adjusting null's behavior so it's the same as false?

src/TwigComponent/tests/Unit/ComponentAttributesTest.php
```diff
@@ -87,4 +87,22 @@ public function testCanAddStimulusController(): void
...
+    {
+        $attributes = new ComponentAttributes(['disabled' => true]);
+
+        $this->assertSame(['disabled' => true], $attributes->all());
+        $this->assertSame(' disabled', (string) $attributes);
+
+        $attributes = new ComponentAttributes(['disabled' => false]);
+
+        $this->assertSame(['disabled' => false], $attributes->all());
+        $this->assertSame('', (string) $attributes);
```
weaverryan, line 101: Both `true` and `false` behaviors are BC breaks, no? Bug fixes, but breaks?

kbond: Yes, not sure how we should handle this...

src/TwigComponent/tests/Unit/ComponentAttributesTest.php
kbond: How do you suggest we do this?

src/TwigComponent/tests/Unit/ComponentAttributesTest.php
weaverryan: Actually, these are ok.

Currently:

```
true => disabled="1"
false => disabled
```

So changing `true` to just `disabled` looks NOT like a BC break to me/
The previous `false` was just plain incorrect. That’s a bug fix

src/TwigComponent/tests/Unit/ComponentAttributesTest.php
weaverryan: Let's "punt" on this. For now, let's trigger a deprecation saying that `null` isn't allowed.

Then in the next major, `null` could throw an exception. Or perhaps later we add a way to "opt into" allowing null with the new behavior.

kbond:
Ok, I believe this is ready. A deprecation is triggered for `null`.

weaverryan:
Looks perfect - thank you Kevin!

Issue #711, opened 2023-03-01
Add State Management (Pinia) to the Symfony/UX-Vue


deanilvincent:
Hi, 

First of all, I'd like to thank you for creating this awesome symfony/ux-vue plugin. Few configurations but has powerful usage especially if I want to a vue.js template and logics within a component in the Symfony Application. I also followed the documentation and completed the setup from here [https://symfony.com/bundles/ux-vue/current/index.html](https://symfony.com/bundles/ux-vue/current/index.html)

My question, is there a way to create a root state and access that state regardless of the components that want to access it? I'm using Pinia as my state management. [https://pinia.vuejs.org/](https://pinia.vuejs.org/).

On a regular vue.js app, we configured the pinia state at the `main.js` file.
```
import { createApp } from "vue";
import { createPinia } from "pinia";
import App from "./App.vue";

const app = createApp(App);
app.use(createPinia());
```

Thank you and I look forward to your response.

Kocal:
Hi, you can use the `vue:before-mount` event, see https://symfony.com/bundles/ux-vue/current/index.html#events

deanilvincent:
> Hi, you can use the `vue:before-mount` event, see https://symfony.com/bundles/ux-vue/current/index.html#events

Thank you! I'll try that. 
Just a quick question, where do we properly declare these line of codes? Is it within the app.js file or somewhere else?
```
document.addEventListener('vue:before-mount', (event) => {
    ...
    app.use(router);
});
```

Kocal:
Yep, inside the  app.js  file

Kocal:
Can be closed since #712 has been merged

PR #712, opened 2023-03-01
Add minor helpful words to the Symfony UX Vue Documentation

deanilvincent:
| Q             | A
| ------------- | ---
| Bug fix?      |  No
| New feature?  | No
| Tickets       | #711
| License       | MIT

- Add 'states' word as part of the examples for the "Event" section. This will easily help the new devs to determine where to declare the states e.g. Pinia state to the root app.
- Add `// assets/app.js` comment to which file to add the event.

### DIFF
--- a/src/Vue/doc/index.rst
+++ b/src/Vue/doc/index.rst
@@ -112,10 +112,11 @@ For example:
 Events
 ~~~~~~
 
-The event ``vue:before-mount`` is called before a component is mounted on the page. This is the event to listen if you need to modifiy the Vue application (e.g.: add plugins, add global directives, ...):
+The event ``vue:before-mount`` is called before a component is mounted on the page. This is the event to listen if you need to modifiy the Vue application (e.g.: add plugins, add global directives, states ...):
 
 .. code-block:: js
 
+    // assets/app.js
     document.addEventListener('vue:before-mount', (event) => {
         const {
             componentName, // The Vue component's name


src/Vue/doc/index.rst
```diff
@@ -193,6 +193,27 @@ used for all the Vue routes:
...
+    $ yarn add pinia
+    $ # or with npm
+    $ npm install pinia
+
+.. code-block:: js
+
+    import { createPinia } from 'pinia'
+    ...
+    document.addEventListener('vue:before-mount', (event) => {
+        app.use(createPinia());
```
memic84: Currently trying to get this working, but if i follow this part of the documentation the `app` in app.use is undefined.
Should we import the stimulus app here `import { app } from "./bootstrap";`? afaik this should be the Vue app, right?

Kocal: The app should be fetch from `event.detail`

deanilvincent: Hi @memic84,

Thank you for checking. I actually reverted that commit because I'm thinking to create a separate PR for that part of the docu regarding the optional implementation of Pinia. @Kocal is correct. It should be fetched from `event.detail` I'll update that commit on the future PR.

Here's what it looks like:
```
document.addEventListener('vue:before-mount', (event) => {
    const {
        componentName, // The Vue component's name
        component, // The resolved Vue component
        props, // The props that will be injected to the component
        app, // The Vue application instance
    } = event.detail;
    
    app.use(createPinia());
});
```
I tried this and it's working :)

deanilvincent: Or should I include it in this current PR? :) Please let me know. Thank you!

weaverryan:
Merging in the current state - thank you @deanilvincent!

PR #713, opened 2023-03-02
[Live] Robust handling of invalid user data

### LINKED ISSUE #528: [RFC] Handling Data with Bad Types (i.e. that fails hydration)

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Tickets       | Fix #528
| License       | MIT

Hi!

Another boring PR from me 😄.

t;dr When making an Ajax request, we now send the "original data" + "updated date". If the user sends something wacko (e.g. the word `apple` to a property with an `int` type), we can reject it and use the original data.

Cheers!


src/LiveComponent/src/Util/DehydratedProps.php
```diff
@@ -0,0 +1,236 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Util;
+
+class DehydratedProps
```
kbond, line 22:
```suggestion
/**
 * Helps organize the dehydrated props and data.
 *
 * @author Ryan Weaver <ryan@symfonycasts.com>
 *
 * @experimental
 *
 * @internal
 */
class DehydratedProps
```

src/LiveComponent/src/Util/DehydratedProps.php
```diff
@@ -0,0 +1,236 @@
+<?php
```
kbond, line 1:
```suggestion
<?php

/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */
```

norkunas:
As this affects hydration, I have a small proposition: add `dehydrationContext` to `AsLiveComponent` attribute which would be passed down to `serializer->denormalize` context.

Use case:
I have a root DTO for a form type and I declare on the DTO properties with docblock type hints and validation constraints, but without strict typing to avoid errors.

So now the result is that my initial DTO does not hold any data because it must be filled, and if I change a single field then the dehydration fails because the initial values are `null` and serializer gets from the property info that the value must be a string for example. (I know that I could initialize them as empty strings, so the type would match, but I can't initialize the prop if it expects an entity from an EntityType because it's not loaded at that time)

So my expectations that the form validation would handle that the value is not a string or an empty string, etc. and at the end I'd have a valid DTO which I could use.

Currently my workaround is to create a custom denormalizer and pass down `AbstractObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true` but that's not a friendly solution :)

weaverryan:
I'm not sure specifically about your use-case... Regardless, you're 100% correct about needing a way to pass in the context for the serializer.

weaverryan:
(de)normalization context added - usage for your situation

```php
#[LiveProp(
    denormalizationContext: [ObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true]
)]
public YourDto $yourDto;
```

kbond:
What do you think about reusing the context attribute from serializer?

```php
#[LiveProp]
#[Context([ObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true])]
public YourDto $yourDto;
```

norkunas:
> What do you think about reusing the context attribute from serializer?
> 
> ```php
> #[LiveProp]
> #[Context([ObjectNormalizer::DISABLE_TYPE_ENFORCEMENT => true])]
> public YourDto $yourDto;
> ```

Same thing to complete same task, I'd agree nevertheless ☺️

kbond:
After I wrote that I was thinking that a property on LiveProp would be more "discoverable"

weaverryan:
I don't think discoverability is very important as this is more of an edge-case feature. So I agree with using `#[Context]`. Btw, that feature (other than the "New in Symfony 5.3") looks undocumented - someone should fix that ;)

weaverryan:
Updated to use `#[Context]`.

Note: while we use the serialization system, we only read the "context": we're not reading and using other options like "maxDepth", etc. Also, if you have `#[Context]` on a live prop, then that context is used to (de)normalize any writable paths. For example, if the `LiveProp` is a `Post` object and it has a writable path that is a `Category` object, we don't also look on *that* property to see if IT also has a `Context`. I want to avoid getting in the weeds with complex serialization edge-cases. And, so far, other thank @norkunas' use-case, nobody has even asked for control over the serialization context yet. So giving some basic control, I think, is enough.

weaverryan:
Merging this - it's big and important - if there are any kinks, now we can work them out.

PR #717, opened 2023-03-04
[Live] Fix handling forms with unnamed root

### LINKED ISSUE #715: [LiveComponent] LiveCollectionType does not play well without block prefix
As soon as the form type returns an empty block prefix, `LiveCollectionType` fails to add or remove an item in the collection:
`Cannot read property "items" from an array. Maybe you intended to write the property path as "[items]" instead.`
That's because the property path for the action becomes `addCollectionItem(name=items)` instead of `addCollectionItem(name=todolist[items])`.

Should it be supported out of the box?

Reproducer: https://github.com/norkunas/sfux-tests/tree/live-form-dto-collection

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #715
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/LiveCollectionTrait.php
+++ b/src/LiveComponent/src/LiveCollectionTrait.php
@@ -27,30 +27,39 @@ trait LiveCollectionTrait
     #[LiveAction]
     public function addCollectionItem(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name): void
     {
-        if (str_starts_with($name, $this->formName)) {
-            $name = substr_replace($name, '', 0, mb_strlen($this->formName));
-        }
-
-        $data = $propertyAccessor->getValue($this->formValues, $name);
+        $propertyPath = $this->fieldNameToPropertyPath($name, $this->formName);
+        $data = $propertyAccessor->getValue($this->formValues, $propertyPath);
 
         if (!\is_array($data)) {
-            $propertyAccessor->setValue($this->formValues, $name, []);
+            $propertyAccessor->setValue($this->formValues, $propertyPath, []);
             $data = [];
         }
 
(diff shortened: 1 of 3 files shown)


kbond (approved):

src/LiveComponent/src/LiveCollectionTrait.php
```diff
@@ -27,30 +27,39 @@ trait LiveCollectionTrait
...
     }
 
     #[LiveAction]
     public function removeCollectionItem(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name, #[LiveArg] int $index): void
     {
-        if (str_starts_with($name, $this->formName)) {
-            $name = substr_replace($name, '', 0, mb_strlen($this->formName));
+        $propertyPath = $this->fieldNameToPropertyPath($name, $this->formName);
+        $data = $propertyAccessor->getValue($this->formValues, $propertyPath);
+        unset($data[$index]);
```
norkunas, line 47: What if array_splice would be used to reindex?

1ed: What would be the benefit of that? I think it's more readable this way. We are just removing an item here, not re-indexing the array.

norkunas: Most of the time you'd want a list, no?

1ed: If we re-index the array here that would trigger re-rendering all other items after the removed item on the client, AFAIK. I had no issues with this so far, so I rather not do that.

norkunas: Hm,okay,will see if that works afterwards

weaverryan:
Thank you Gábor!

PR #721, opened 2023-03-06
[Live] Adding failing test case for updating individual parts of deeply writable paths

### LINKED ISSUE #720: [LiveComponent] Fix bug where writable path does not respect frontend "fieldName"
I've installed `^2.x-dev` to test live component but have some issues with it.
In [this test repository](https://github.com/norkunas/sfux-tests/tree/illegal-form-names) I've added three test components with different configurations:
 - test1: `#[LiveProp(fieldName: 'data')]`
   Access the form, click add/delete collection items, no errors. As soon as I try to enter a value to any input it fails with `Invalid or missing checksum. This usually means that you tried to change a property that is not writable: true.` (okay, so probably this is expected, as we need to expose what is writable)
 - test2: `#[LiveProp(fieldName: 'data', writable: true)]`
   Access the form, immediately error is thrown: `An exception has been thrown during the rendering of a template ("The LiveProp path "invoice" is an object that was dehydrated to an array *and* it is writable. That's not allowed. You probably want to set writable to only the properties on your class that should be writable (e.g. writable: ['name', 'price']).").`
 - test3: `#[LiveProp(fieldName: 'data', writable: ['number', 'lineItems'])]`
   Access the form, enter value to any input - same errors as in the first test.
   Adding a single item to collection works, adding one more or removing first one throws: `The name "0.name" contains illegal characters. Names should start with a letter, digit or underscore and only contain letters, digits, numbers, underscores ("_"), hyphens ("-") and colons (":").`

What I miss here to make it work?

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #720
| License       | MIT

In the unreleased code, writing a "deep" path to an array `LiveProp` is broken (e.g. `formValues` in `ComponentWithFormTrait` when you have an embedded form). Test case here, solution coming shortly.

### DIFF
--- a/src/LiveComponent/assets/dist/Component/ElementDriver.d.ts
+++ b/src/LiveComponent/assets/dist/Component/ElementDriver.d.ts
@@ -1,6 +1,9 @@
 export interface ElementDriver {
     getModelName(element: HTMLElement): string | null;
-    getComponentProps(rootElement: HTMLElement): any;
+    getComponentProps(rootElement: HTMLElement): {
+        props: any;
+        nestedProps: any;
+    };
     findChildComponentElement(id: string, element: HTMLElement): HTMLElement | null;
     getKeyFromElement(element: HTMLElement): string | null;
 }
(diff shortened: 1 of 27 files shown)


weaverryan:
This works now on the reproducer from #720. I'm playing a bit of whack-a-mole with hydration edge-cases, but I don't think it's a problem: each adds a new test case, and I think remaining problems are finite.

PR #725, opened 2023-03-08
[LiveComponent] Fix calculateFingerprint with big objects

LunkLoafGrumble:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

This pull request will fix an issue for large objects (for example entities with a large relations) with in the FingerprintCalculator

While working with liveComponents I noticed my project getting slower and slower. After some digging I found out that in the calculateFingerprint function the data is serialized. When an entity with relations passes by it will serialize the entity and its child objects. This takes a considerable amount of time and is not needed.

### DIFF
--- a/src/LiveComponent/src/Util/FingerprintCalculator.php
+++ b/src/LiveComponent/src/Util/FingerprintCalculator.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of the Symfony package.
  *
@@ -11,6 +13,9 @@
 
 namespace Symfony\UX\LiveComponent\Util;
 
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\UX\LiveComponent\LiveComponentHydrator;
+
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
  *
@@ -20,12 +25,16 @@ namespace Symfony\UX\LiveComponent\Util;
  */
 class FingerprintCalculator
 {
-    public function __construct(private string $secret)
-    {
+    public function __construct(
+        private NormalizerInterface $objectNormalizer,
+        private string $secret,
+    ) {
     }
 
     public function calculateFingerprint(array $data): string
     {
-        return base64_encode(hash_hmac('sha256', serialize($data), $this->secret, true));
+        $normalizedData = $this->objectNormalizer->normalize($data, context: [LiveComponentHydrator::LIVE_CONTEXT => true]);
+
+        return base64_encode(hash_hmac('sha256', serialize($normalizedData), $this->secret, true));
     }
 }
--- a/src/LiveComponent/tests/Integration/Util/FingerprintCalculatorTest.php
+++ b/src/LiveComponent/tests/Integration/Util/FingerprintCalculatorTest.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Tests\Integration\Util;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\UX\LiveComponent\Tests\Fixtures\Entity\Entity1;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Test\ResetDatabase;
+
+use function Zenstruck\Foundry\create;
+
+final class FingerprintCalculatorTest extends KernelTestCase
+{
+    use Factories;
+    use ResetDatabase;
+
+    public function testFingerprintEqual()
+    {
+        $fingerprintCalculator = self::getContainer()->get('ux.live_component.fingerprint_calculator');
+
+        $entityOne = create(Entity1::class)->object();
+
+        $entityTwo = clone $entityOne;
+
+        self::assertNotSame(
+            spl_object_id($entityOne),
+            spl_object_id($entityTwo)
+        );
+
+        self::assertSame(
+            $fingerprintCalculator->calculateFingerprint([$entityOne]),
+            $fingerprintCalculator->calculateFingerprint([$entityTwo])
+        );
+    }
+
+    public function testFingerprintNotEqual()
+    {
+        $fingerprintCalculator = self::getContainer()->get('ux.live_component.fingerprint_calculator');
+
+        $entityOne = create(Entity1::class)->object();
+
+        $entityTwo = create(Entity1::class)->object();
+
+        self::assertNotSame(
+            $fingerprintCalculator->calculateFingerprint([$entityOne]),
+            $fingerprintCalculator->calculateFingerprint([$entityTwo])
+        );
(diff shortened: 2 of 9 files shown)


kevinmade:
I'm having the same issue.

src/LiveComponent/src/Util/FingerprintCalculator.php
```diff
@@ -26,6 +26,21 @@ public function __construct(private string $secret)
 
     public function calculateFingerprint(array $data): string
     {
-        return base64_encode(hash_hmac('sha256', serialize($data), $this->secret, true));
+        return base64_encode(hash_hmac('sha256', serialize($this->normalizeData($data)), $this->secret, true));
+    }
+
+    private function normalizeData(array $data): array
```
kbond:
```suggestion
    private static function normalizeData(array $data): array
```

kbond (commented):
Makes sense to me!

src/LiveComponent/src/Util/FingerprintCalculator.php
LunkLoafGrumble: Done

kbond (approved):

src/LiveComponent/src/Util/FingerprintCalculator.php
```diff
@@ -26,6 +26,21 @@ public function __construct(private string $secret)
 
     public function calculateFingerprint(array $data): string
     {
-        return base64_encode(hash_hmac('sha256', serialize($data), $this->secret, true));
+        return base64_encode(hash_hmac('sha256', serialize($this->normalizeData($data)), $this->secret, true));
```
kbond:
```suggestion
        return base64_encode(hash_hmac('sha256', serialize(self::normalizeData($data)), $this->secret, true));
```

src/LiveComponent/src/Util/FingerprintCalculator.php
```diff
@@ -26,6 +26,21 @@ public function __construct(private string $secret)
...
+
+    private static function normalizeData(array $data): array
+    {
+        return array_map(function (mixed $value) {
+            if (\is_object($value)) {
+                return spl_object_id($value);
+            }
+
+            if (\is_array($value)) {
+                return $this->normalizeData($value);
```
kbond:
```suggestion
                return self::normalizeData($value);
```

kbond (changes requested):

kbond (approved):

weaverryan:
Great catch on this!

These fingerprints need to be a deterministic string that represents "the input passed into a component" - https://github.com/symfony/ux/blob/e45a8edfcd6d6ab62248d1ffe42ef13e9dcc11b3/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php#L68 - and they need to always return the same string for the same input.

If you pass an object (e.g. a `Post` entity) into a component, the `spl_object_id()` of that might be different the next time the component is re-rendered via an Ajax call, right?

I'm open to ideas on how to fix this. In practice, if we added some special code for entity objects that, if persisted, used their id instead of serializing them, that might solve things. For example, if one of the items in `$data` is a `Post` entity, then instead of serializing that or `spl_object_id()`, we use `App\Entity\Post:{THE_ID}`.

kbond:
> These fingerprints need to be a deterministic string that represents

Oops, didn't realize this. Could `spl_object_hash()` be used? I don't know if this would still have a perf problem.

weaverryan:
I think `spl_object_hash()` has the same problem - not 100% certain on that, but pretty sure.

kbond:
You're right. From php.net:

> When an object is destroyed, its hash may be reused for other objects.

kbond:
> These fingerprints need to be a deterministic string that represents "the input passed into a component" -

We should have a test to prove this behaviour.

kevinmade:
Perhaps it's an idea to add `if (\is_object($value) && method_exists($value, 'getId')) {
                return $value->getId();
            }` instead of `\is_object($value)`.

src/LiveComponent/src/Util/FingerprintCalculator.php
```diff
@@ -20,12 +25,29 @@
...
 
     public function calculateFingerprint(array $data): string
     {
-        return base64_encode(hash_hmac('sha256', serialize($data), $this->secret, true));
+        return base64_encode(hash_hmac('sha256', serialize($this->normalizeData($data)), $this->secret, true));
+    }
+
+    private function normalizeData(array $data): array
+    {
+        return array_map(function (mixed $value) {
```
kbond: Can we just call `return $this->objectNormalizer->normalize($value, context: [LiveComponentHydrator::LIVE_CONTEXT => true]);` (or inline above)?

kbond (changes requested):
This is probably the best solution.

@LunkLoafGrumble does this method improve the speed in your app?

Could you add a basic integration test for `FingerprintCalculator` that proves the generated fingerprint is the same two different object instances that are equal?

src/LiveComponent/src/Util/FingerprintCalculator.php
LunkLoafGrumble: Good suggestion. changed it.

LunkLoafGrumble:
Using objectNormalizer greatly improves the performance over just serializing the object.

src/LiveComponent/tests/Integration/Util/FingerprintCalculatorTest.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function testFingerprintEqual()
+    {
+        $fingerprintCalculator = self::getContainer()->get('ux.live_component.fingerprint_calculator');
+
+        $entityOne = create(Entity1::class)->object();
+
+        $entityTwo = clone $entityOne;
+
+        self::assertNotEquals(
```
kbond:
```suggestion
        self::assertNotSame(
```

src/LiveComponent/tests/Integration/Util/FingerprintCalculatorTest.php
```diff
@@ -0,0 +1,58 @@
...
+            spl_object_id($entityOne),
+            spl_object_id($entityTwo)
+        );
+
+        self::assertEquals(
```
kbond:
```suggestion
        self::assertSame(
```

src/LiveComponent/tests/Integration/Util/FingerprintCalculatorTest.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function testFingerprintNotEqual()
+    {
+        $fingerprintCalculator = self::getContainer()->get('ux.live_component.fingerprint_calculator');
+
+        $entityOne = create(Entity1::class)->object();
+
+        $entityTwo = create(Entity1::class)->object();
+
+        self::assertNotEquals(
```
kbond:
```suggestion
        self::assertNotSame(
```

kbond (changes requested):
A few remaining minor things, then good to go by me!

kbond (approved):

weaverryan:
Let's give this version a try - thanks @LunkLoafGrumble!

weaverryan:
FYI - I'm reverting this in #832. But I'm hoping that it's not a problem anymore because usually zero LiveProps are used to calculate the fingerprint. But please tell me if you still have issues.

PR #726, opened 2023-03-09
[TwigComponent] Move DataModelPropsSubscriber to LiveComponent

onEXHovia:
Parsing the "data-model" key useful only use LiveComponent, so let's move services from TwigComponent.


weaverryan:
Fair enough - this is definitely true on practice. Thanks @onEXHovia!

PR #728, opened 2023-03-13
[Live] Smart Rendering! Mix/match live components with external JavaScript

### LINKED ISSUE #270: [LiveComponent] data-live-ignore and nested elements

### LINKED ISSUE #354: [LiveComponent] Using other UX components in Live form

### LINKED ISSUE #490: [LiveComponent] Better control over re-rendering behavior (relevant for js widgets)

### LINKED ISSUE #515: [Live+Autocomplete] Wrong behavior of Autocomplete with dynamic forms

### LINKED ISSUE #702: [LiveComponent] Prevent uncollapse item on refresh
Hello, I'm using LiveComponent to render an hardcore form with a lot of collections.

This form has collections inside collapsed div.
When clicking the button to "add" or "remove" a collection item, the item is added/removed but all is "uncollapsed".

Example : 

https://user-images.githubusercontent.com/53140475/220677572-1852e200-3e9e-4e8b-bd86-7109fbf66755.mp4

Unfortunately, I cannot add "data-live-ignore" in the div containing the collapsed content since otherwise the buttons allowing to add/remove an element from a collection will no longer refresh the fields.

I had another idea which was to set up a system that would save in the entities whether it's collapsed or not, and at refresh, would allow me to put everything back as it should be, but I find it restrictive, and I wonder if there isn't a better way?

This is how I render the form : 

```twig
<ul class="list-group">
        <li class="list-group-item bg-danger text-white">
            <span class="badge bg-success rounded-pill"></span>
            <span class="ms-2 fs-5">Éléments de preuve et questionnaire</span>
        </li>
        {% for sous_dossier_index, sous_dossier_form in form.sousDossiers.children %}
            {% set sous_dossier = sous_dossier_form.vars.data %}
            <li class="list-group-item" style="background-color: #f8f9fa;" data-bs-toggle="collapse" data-bs-target="#sous_dossier_{{ sous_dossier_index }}" aria-expanded="false" aria-controls="sous_dossier_{{ sous_dossier_index }}">
                <div class="d-flex justify-content-between align-items-center" role="button">
                    <span class="fs-4">{{ sous_dossier.name }}</span>
                    <span class="badge bg-warning rounded-pill">&nbsp;</span>
                </div>
                <div id="sous_dossier_{{ sous_dossier_index }}" class="collapse">
                        <div class="row row-cols-1 row-cols-md-2 g-4">
                            {% for sous_dossier_item_index, sous_dossier_item_form in sous_dossier_form.items %}
                                {% set sous_dossier_item = sous_dossier_item_form.vars.data %}
                                <div class="col">
                                    <div class="card">
                                        <div class="card-header">{{ sous_dossier_item }}</div>
                                        <div class="card-body">
                                            {% set files_item = sous_dossier_item_form.files.children %}

                                            {% for file_index, file_item in files_item %}
                                                <button type="button"
                                                        class="btn btn-link text-danger"
                                                        data-action="live#action"
                                                        data-action-name="removeItem(sousDossierIndex={{ sous_dossier_index }}, sousDossierItemIndex={{ sous_dossier_item_index }}, fileIndex={{ file_index }})">
                                                    Retirer
                                                </button>
                                                {{ form_widget(file_item, {'attr': {'data-model': 'on(change)|norender|' ~ file_item.vars.full_name }}) }}
                                            {% endfor %}
                                        </div>
                                        <div class="card-footer">
                                            <button type="button"
                                                    class="btn btn-link float-start"
                                                    data-action="live#action"
                                                    data-action-name="addItem(sousDossierIndex={{ sous_dossier_index }}, sousDossierItemIndex={{ sous_dossier_item_index }})">
                                                Ajouter un élément
                                            </button>
                                            <button type="button" class="btn btn-sm btn-secondary float-end">
                                                <i class="fas fa-arrow-right"></i>
                                            </button>
                                        </div>
                                    </div>
                                </div>
                            {% endfor %}
                        </div>
                </div>
            </li>
        {% endfor %}
    </ul>
```

And in the component, my two liveAction to add or remove an item : 

```php
    #[LiveAction]
    public function removeItem(#[LiveArg] int $sousDossierIndex, #[LiveArg] int $sousDossierItemIndex, #[LiveArg] int $fileIndex): void
    {
        if (isset($this->formValues['sousDossiers'][$sousDossierIndex]['items'][$sousDossierItemIndex]['files'][$fileIndex])) {
            unset($this->formValues['sousDossiers'][$sousDossierIndex]['items'][$sousDossierItemIndex]['files'][$fileIndex]);
        }
    }

    #[LiveAction]
    public function addItem(#[LiveArg] int $sousDossierIndex, #[LiveArg] int $sousDossierItemIndex): void
    {
        $this->formValues['sousDossiers'][$sousDossierIndex]['items'][$sousDossierItemIndex]['files'][] = [];
    }
```

Thanks for your help!

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yesish
| New feature?  | yes
| Tickets       | Fixes #702, Fixes #490, Fixes #515, Fixes #270, Fixes Bug H on https://github.com/symfony/ux/issues/102, Fixes #354 Replaces #519
| License       | MIT

Hi!

This fixes the biggest pain-point with LiveComponents: mixing them with custom JavaScript. Before this PR, a lot of things were difficult/weird: keeping modals open, collapsing content (#702), using JS widgets like Autocomplete or a date picker, and more. The best part? Everything works automatically with you needing to do anything.

With this PR, a new philosophy is born: **start with a live component, then mix in any "extras" you want with Stimulus controllers**.

The new "smart rendering" system isn't designed to catch everything (see the docs), but I think it will catch 95%+ of the situations. Here are 2 cool things that now work out-of-the-box:

A) ux-autocomplete + Live components works perfectly now - even if the available options completely change - the "dependent form fields" example has been updated to use autocomplete on both fields 🔥 

https://user-images.githubusercontent.com/121003/224790239-3b723fa0-1eb7-4f27-bd71-a04bf32687f1.mp4

B) I added a new demo of Live Components + Autocomplete + Chartjs. What you see is done without needing to write any JavaScript:

https://user-images.githubusercontent.com/121003/224789304-ac1228a8-57eb-4797-8df0-245bd767872b.mp4

### How Does it Work?

If you care about the technical details, we leverage a `MutationObserver` to watch your component for changes made by anything *other* than live components. Those changes are "tracked". Then, during re-render, we use that info to *keep* those custom changes.

Cheers!

### DIFF
--- a/src/Autocomplete/assets/dist/controller.d.ts
+++ b/src/Autocomplete/assets/dist/controller.d.ts
@@ -1,5 +1,12 @@
 import { Controller } from '@hotwired/stimulus';
 import TomSelect from 'tom-select';
+export interface AutocompletePreConnectOptions {
+    options: any;
+}
+export interface AutocompleteConnectOptions {
+    tomSelect: TomSelect;
+    options: any;
+}
 export default class extends Controller {
     #private;
     static values: {
@@ -23,11 +30,20 @@ export default class extends Controller {
     readonly hasPreloadValue: boolean;
     readonly preloadValue: string;
     tomSelect: TomSelect;
+    private mutationObserver;
+    private isObserving;
     initialize(): void;
     connect(): void;
     disconnect(): void;
+    private getMaxOptions;
     get selectElement(): HTMLSelectElement | null;
     get formElement(): HTMLInputElement | HTMLSelectElement;
     private dispatchEvent;
     get preload(): string | boolean;
+    private resetTomSelect;
+    private changeTomSelectDisabledState;
+    private updateTomSelectPlaceholder;
+    private startMutationObserver;
+    private stopMutationObserver;
+    private onMutations;
 }
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -27,14 +27,13 @@ class default_1 extends Controller {
     constructor() {
         super(...arguments);
         _default_1_instances.add(this);
+        this.isObserving = false;
     }
     initialize() {
-        this.element.setAttribute('data-live-ignore', '');
-        if (this.element.id) {
-            const label = document.querySelector(`label[for="${this.element.id}"]`);
-            if (label) {
-                label.setAttribute('data-live-ignore', '');
-            }
+        if (!this.mutationObserver) {
+            this.mutationObserver = new MutationObserver((mutations) => {
+                this.onMutations(mutations);
+            });
         }
     }
     connect() {
@@ -47,10 +46,14 @@ class default_1 extends Controller {
             return;
         }
         this.tomSelect = __classPrivateFieldGet(this, _default_1_instances, "m", _default_1_createAutocomplete).call(this);
+        this.startMutationObserver();
     }
     disconnect() {
-        this.tomSelect.revertSettings.innerHTML = this.element.innerHTML;
+        this.stopMutationObserver();
         this.tomSelect.destroy();
+    }
+    getMaxOptions() {
+        return this.selectElement ? this.selectElement.options.length : 50;
     }
     get selectElement() {
         if (!(this.element instanceof HTMLSelectElement)) {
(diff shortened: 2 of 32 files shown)


jcrombez:
Can't wait to get back to work tomorrow to try this black magic, good job <3

weaverryan:
Fixing a bug I just found - where the `live` controller needs to be registered before any other controllers that might make custom changes in their `connect()` method

weaverryan:
And... here's the fix: https://github.com/symfony/stimulus-bridge/pull/81 - using the latest version of `@symfony/stimulus-bridge` will be a requirement if you want to take advantage of this new feature in Live Components.

norkunas:
I've got a problem with this philosophy:

In live component i have a list of checkboxes. Each checkbox has it's own stimulus controller which has configured custom `checked` class. So when I check the checkbox, the class is added, live component rerenders - everything fine.

Simultaneously I have an live component action which changes the `checked` value in component code, so after rerender checkbox becomes unchecked but the class is still left here even backend returns the element without custom `checked` class. Even returns with the `checked` attribute..

weaverryan:
Thanks for bringing this up @norkunas - I figured there would be some real-world hiccups we'd need to work through.

> so after rerender checkbox becomes unchecked but the class is still left here even backend returns the element without custom checked class.

That makes sense. Question: BEFORE the new re-rendering system, did you have the opposite problem? When the Stimulus controller added the checked class, when the live component re-rendered (not via the action, just a normal re-render), didn't it NOT include the custom class? And so, didn't the re-render cause the custom class to be lost? Or do you have the "if checked -> add custom class" logic both inside of Twig and also include if your Stimulus controller? And if so, why do you have it in both places? Is it just to make the experience feel faster / more responsive (i.e. as SOON as the user checks the box, the class is added, without needing to wait for the re-render)?

Btw, feel free to open a new issue - I don't want this to get lost on a merged PR - I sometimes catch these notifications, and sometimes not :)

Enitnelav:
Hi,
Do you know when the 2.8 version will be stable and available ?
Thanks

weaverryan:
Hoping within a week - only waiting for 2 minor bug fixes with new twig syntax

ag-erwan:
Hello community

I come to you with a similar problem. 
In addition to the **Live Component** and **Autocomplete** libraries I also use the **symfonycast/dynamics-form** library, I manage to get my fields to display at the first change with the data in them, but when I modify my select field again, my other fields do not appear. do not modify. In fact it's a rendering problem, meaning it doesn't render again. Because when I do a dd of the field I can clearly see the data modify.
Thank you in advance for your help

antoniovj1:
@ag-erwan Did yo solve you issue? I'm facing the same problem I think

PR #734, opened 2023-03-15
Drop deprecated namedExports option

bocharsky-bw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #733 
| License       | MIT

PR #735, opened 2023-03-16
Fix TypeScript warnings

bocharsky-bw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

If you run `yarn build` there are quite a few TS warnings. This happened because we originally didn't use TypeScript, so when we switched, a lot of "types" were missing.

I'm not sure how to fix the last one remaining here:
```
src/Notify/assets/src/controller.ts → src/Notify/assets/dist/controller.js...
(!) Plugin typescript: @rollup/plugin-typescript TS2769: No overload matches this call.
  Overload 1 of 3, '(type: "message", listener: (this: EventSource, ev: MessageEvent<any>) => any, options?: boolean | EventListenerOptions | undefined): void', gave the following error.
    Argument of type '(content: string | undefined) => void' is not assignable to parameter of type '(this: EventSource, ev: MessageEvent<any>) => any'.
      Types of parameters 'content' and 'ev' are incompatible.
        Type 'MessageEvent<any>' is not assignable to type 'string'.
  Overload 2 of 3, '(type: string, listener: (this: EventSource, event: MessageEvent<any>) => any, options?: boolean | EventListenerOptions | undefined): void', gave the following error.
    Argument of type '(content: string | undefined) => void' is not assignable to parameter of type '(this: EventSource, event: MessageEvent<any>) => any'.
      Types of parameters 'content' and 'event' are incompatible.
        Type 'MessageEvent<any>' is not assignable to type 'string'.
  Overload 3 of 3, '(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | EventListenerOptions | undefined): void', gave the following error.
    Argument of type '(content: string | undefined) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.
      Type '(content: string | undefined) => void' is not assignable to type 'EventListener'.
        Types of parameters 'content' and 'evt' are incompatible.
          Type 'Event' is not assignable to type 'string'.
src/Notify/assets/src/controller.ts: (61:56)

61             eventSource.removeEventListener('message', this._notify);
                                                          ~~~~~~~~~~~~
created src/Notify/assets/dist/controller.js in 926ms
```

If anyone has ideas how to fix - please, share :)

Depreciations like "(!) Plugin commonjs: The namedExports option from "@rollup/plugin-commonjs" is deprecated. Named exports are now handled automatically." were fixed in https://github.com/symfony/ux/pull/734

### DIFF
--- a/src/LiveComponent/assets/src/ComponentRegistry.ts
+++ b/src/LiveComponent/assets/src/ComponentRegistry.ts
@@ -1,7 +1,7 @@
 import Component from './Component';
 import { getElementAsTagText } from './dom_utils';
 
-const ComponentRegistry = class {
+class ComponentRegistry {
     private components = new WeakMap<HTMLElement, Component>();
 
     public registerComponent(element: HTMLElement, definition: Component) {
@@ -30,6 +30,6 @@ const ComponentRegistry = class {
             }, 5);
         });
     }
-};
+}
 
 export default new ComponentRegistry();
--- a/src/Notify/assets/src/controller.ts
+++ b/src/Notify/assets/src/controller.ts
@@ -19,6 +19,10 @@ export default class extends Controller {
         hub: String,
         topics: Array,
     };
+    declare hubValue: string;
+    declare topicsValue: Array<string>;
+    declare readonly hasHubValue: boolean;
+    declare readonly hasTopicsValue: boolean;
 
     eventSources: Array<EventSource> = [];
 
--- a/src/Notify/assets/src/controller.ts
+++ b/src/Notify/assets/src/controller.ts
@@ -25,6 +25,7 @@ export default class extends Controller {
     declare readonly hasTopicsValue: boolean;
 
     eventSources: Array<EventSource> = [];
+    listeners: WeakMap<EventSource, (event: MessageEvent) => void> = new WeakMap();
 
     initialize() {
         const errorMessages: Array<string> = [];
@@ -50,7 +51,9 @@ export default class extends Controller {
         }
 
         this.eventSources.forEach((eventSource) => {
-            eventSource.addEventListener('message', (event) => this._notify(JSON.parse(event.data).summary));
+            const listener = (event: MessageEvent) => this._notify(JSON.parse(event.data).summary);
+            eventSource.addEventListener('message', listener);
+            this.listeners.set(eventSource, listener);
         });
 
         this.dispatchEvent('connect', { eventSources: this.eventSources });
@@ -58,7 +61,11 @@ export default class extends Controller {
 
     disconnect() {
         this.eventSources.forEach((eventSource) => {
-            eventSource.removeEventListener('message', this._notify);
+            const listener = this.listeners.get(eventSource);
+            if (listener) {
+                eventSource.removeEventListener('message', listener);
+            }
+
             eventSource.close();
         });
 
--- a/src/Swup/assets/src/index.d.ts
+++ b/src/Swup/assets/src/index.d.ts
@@ -1,0 +1,5 @@
+declare module 'swup';
+declare module '@swup/debug-plugin';
+declare module '@swup/forms-plugin';
+declare module '@swup/fade-theme';
+declare module '@swup/slide-theme';
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,6 +15,7 @@ jobs:
               run: ./vendor/bin/php-cs-fixer fix --dry-run --diff
 
     coding-style-js:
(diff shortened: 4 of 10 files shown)


src/LiveComponent/assets/src/ComponentRegistry.ts
```diff
@@ -2,22 +2,22 @@ import Component from './Component';
 import { getElementAsTagText } from './dom_utils';
 
 const ComponentRegistry = class {
-    private components = new WeakMap<HTMLElement, Component>();
+    #components = new WeakMap<HTMLElement, Component>();
```
weaverryan: private should be ok - it’s a typescript thing (whereas # is pure js, so that’s also valid, but private should be ok in ts). But I see that this fixed a warning:

> Property 'components' of exported class expression may not be private or protected.

Apparently the problem is that this is an anonymous class - https://github.com/microsoft/TypeScript/issues/30355#issuecomment-473060190

I'll comment the fix above

src/LiveComponent/assets/src/morphdom.ts
```diff
@@ -41,8 +41,8 @@ export function executeMorphdom(
             }
 
             if (
-                !(fromEl instanceof HTMLElement || fromEl instanceof SVGElement) ||
-                !(toEl instanceof HTMLElement || toEl instanceof SVGElement)
+                !(fromEl instanceof SVGElement || fromEl instanceof HTMLElement) ||
+                !(toEl instanceof SVGElement || toEl instanceof HTMLElement)
```
weaverryan: Did this fix a ts warning? That would be wild - it looks logically identical

src/Notify/assets/src/controller.ts
```diff
@@ -19,6 +19,10 @@ export default class extends Controller {
         hub: String,
         topics: Array,
     };
+    declare hubValue: string;
+    declare topicsValue: Array<string>;
+    declare readonly hasHubValue: boolean;
+    declare readonly hasTopicsValue: boolean;
```
weaverryan, line 25: Awesome 👍

src/Swup/assets/src/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'swup';
+declare module '@swup/debug-plugin';
+declare module '@swup/forms-plugin';
+declare module '@swup/fade-theme';
+declare module '@swup/slide-theme';
```
weaverryan, line 5: Can you tell more about this? We are not declaring any .d.ts files by hand at the moment. Are you fixing missing types from those swup libs?

src/LiveComponent/assets/src/ComponentRegistry.ts
```diff
@@ -2,22 +2,22 @@ import Component from './Component';
 import { getElementAsTagText } from './dom_utils';
 
 const ComponentRegistry = class {
```
weaverryan:
```suggestion
class ComponentRegistry {
```

src/LiveComponent/assets/src/morphdom.ts
bocharsky-bw: It did... but don't ask why :D My guess is that somehow the TS thinks it's pointless to check for `SVGElement` because it expects only HTMLElement. But somehow is fixed. Though I see you refactored that if without SVGElement check now :)

src/Swup/assets/src/index.d.ts
bocharsky-bw: Fairly speaking, I don't know for sure, but that's something that the warning suggested to do:
```
(!) Plugin typescript: @rollup/plugin-typescript TS7016: Could not find a declaration file for module 'swup'. '/Users/victor/www/symfony/ux/node_modules/swup/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/swup` if it exists or add a new declaration (.d.ts) file containing `declare module 'swup';`
```

And since no `@types/swup` package, I did the 2nd suggestion. I just googled how to do it and it worked in this case

src/LiveComponent/assets/src/ComponentRegistry.ts
bocharsky-bw: I see it now, thanks!

src/LiveComponent/assets/src/ComponentRegistry.ts
bocharsky-bw: Reverted in favor of `class ComponentRegistry`

weaverryan:
Thanks Victor!

PR #765, opened 2023-03-29
[Live] Emit system for component-to-component communication!

### LINKED ISSUE #163: [LiveComponent] How to reload parent component when child component form submitted?
I have a child component which has an "add item" form, and when this submits, it persists a new entity. This works fine.

However, the parent component is showing a list of the persisted entities - how can I reload the parent component to show the updated list when the child component finished submitting? (it's a "Choose product" form, and then "Add to basket" type thing.)

I've read the docs about the parent and child both having a model property with the same name etc, but there is no model to share here. I would have thought this is a common behaviour, but I just cannot find a way to get it working. Any ideas?

Thanks

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #163
| License       | MIT

Hi!

This PR - which is modeled after Livewire's excellent emitting system - adds a way for a component to communicate to another components by emitting an event. You could use this to, for example, emit an event in ComponentA to trigger ComponentB to re-render itself... which is exactly what's described in #163.

* [ ] Possibly add a way to dispatch a DOM event from PHP

Cheers!

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -1982,6 +1982,162 @@ You can also trigger a specific "action" instead of a normal re-render:
         #}
     >
 
+.. _emit:
+
+Communication Between Components: Emitting Events
+-------------------------------------------------
+
+Events allow you to communicate between any two components that live
+on your page.
+
+Emitting an Event
+~~~~~~~~~~~~~~~~~
+
+There are three ways to emit an event:
+
+1. From Twig:
+
+.. code-block:: twig
+
+    <button
+        data-action="live#emit"
+        data-event="productAdded"
+    >
+
+2. From your PHP component via ``ComponentEmitsTrait``::
+
+    use Symfony\UX\LiveComponent\ComponentEmitsTrait;
+
+    class MyComponent
+    {
+        use ComponentEmitsTrait
+
+        #[LiveAction]
+        public function saveProduct()
+        {
+            // ...
+
+            $this->emit('productAdded');
+        }
+    }
+
+3. :ref:`From JavaScript <working-in-javascript>`, using your component:
+
+.. code-block:: javascript
+
+    this.component.emit('productAdded');
+
+Listen to Events
+~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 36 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -1982,6 +1982,162 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+.................................................
+
+If you want to emit an event to only components with a specific name,
+use the ``name()`` modifier:
+
+.. code-block:: twig
+
+    <button
+        data-action="live#emit"
+        data-event="component(product_list)|productAdded"
```
```suggestion
        data-event="name(product_list)|productAdded"
```
?

kbond (commented):
Looks great!

Would `emitDown` be useful?

src/LiveComponent/src/ComponentEmitsTrait.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+trait ComponentEmitsTrait
+{
+    private LiveResponder $liveResponderForEmit;
+
+    /**
+     * @internal
+     *
+     * @required
+     */
```
kbond:
```suggestion
    /**
     * @internal
     */
    #[Required]
```

(import needed)

src/LiveComponent/src/ComponentEmitsTrait.php
```diff
@@ -0,0 +1,47 @@
...
+    public function setLiveResponderForEmit(LiveResponder $liveResponder): void
+    {
+        $this->liveResponderForEmit = $liveResponder;
+    }
+
+    public function emit(string $eventName, array $data, string $componentName = null): void
```
kbond:
```suggestion
    public function emit(string $eventName, array $data = [], string $componentName = null): void
```

I believe `$data` is optional

wannevancamp:
Any ETA when this cool feature will be released?

weaverryan:
Hopefully it’s a matter of days or a week - but it could get delayed if we find some bugs in the giant 2.8 changes

PR #773, opened 2023-04-07
fix(autocomplete): Use absolute path to generate next page url

### LINKED ISSUE #772: [Autocomplete]  `next_page` URL is absolute
Hi, 

I encounter an error due to the fact that the URL generated to retrieve the next page of results is absolute. 

I know this is a special use case because the application I'm working on is under a proxy, and the hostname in the generated  URL is different from the main one. This leads to an error because the server is not directly reachable from the outside.

So, is there a reason to generate absolute URL? 

If not, I can provide a PR to change this behaviour.

Regards,

Sylvain

syl20b:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #772 
| License       | MIT

This will generate `next_page` URL based on absolute path instead of absolute URL.

### DIFF
--- a/src/Autocomplete/src/Controller/EntityAutocompleteController.php
+++ b/src/Autocomplete/src/Controller/EntityAutocompleteController.php
@@ -46,7 +46,7 @@ final class EntityAutocompleteController
         if ($data->hasNextPage) {
             $parameters = array_merge($request->attributes->all('_route_params'), $request->query->all(), ['page' => $page + 1]);
 
-            $nextPage = $this->urlGenerator->generate($request->attributes->get('_route'), $parameters, UrlGeneratorInterface::ABSOLUTE_URL);
+            $nextPage = $this->urlGenerator->generate($request->attributes->get('_route'), $parameters);
         }
 
         return new JsonResponse([


src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -46,7 +46,7 @@ public function __invoke(string $alias, Request $request): Response
         if ($data->hasNextPage) {
             $parameters = array_merge($request->attributes->all('_route_params'), $request->query->all(), ['page' => $page + 1]);
 
-            $nextPage = $this->urlGenerator->generate($request->attributes->get('_route'), $parameters, UrlGeneratorInterface::ABSOLUTE_URL);
+            $nextPage = $this->urlGenerator->generate($request->attributes->get('_route'), $parameters);
```
weaverryan, line 49: @norkunas you originally added this in #502 and made it absolute. I agree with @syl20b about removing this - but I wanted to make sure you didn't have some specific reason that you make it absolute originally

norkunas: I agree to remove it.

syl20b: @weaverryan @norkunas Good news :) 
Thanks

weaverryan:
Thanks @syl20b!

PR #778, opened 2023-04-11
React components : working twig example

hedocode:
Hello ! The tutorial didn't work because I had not the following lines in my twig : 

{% block stylesheets %}{{ encore_entry_link_tags('app') }}{% endblock %}
{% block javascripts %}{{ encore_entry_script_tags('app') }}{% endblock %}

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | N/A 
| License       | MIT

### DIFF
--- a/src/React/doc/index.rst
+++ b/src/React/doc/index.rst
@@ -62,7 +62,7 @@ Finally, to load your React components, add the following lines to ``assets/app.
     // they are not necessary.
     registerReactControllerComponents(require.context('./react/controllers', true, /\.(j|t)sx?$/));
 
-That's it! Create an `assets/react/controllers/` directory and start creating your
+That's it! Create an ``assets/react/controllers/`` directory and start creating your
 React components.
 
 Usage
@@ -77,6 +77,7 @@ React components located in the directory ``assets/react/controllers`` are regis
 React controller components.
 
 You can then render any React controller component in Twig using the ``react_component``.
+
 For example:
 
 .. code-block:: javascript
@@ -88,11 +89,19 @@ For example:
         return <div>Hello {props.fullName}</div>;
     }
 
-
 .. code-block:: html+twig
 
     {# templates/home.html.twig #}
-    <div {{ react_component('MyComponent', { 'fullName': app.user.fullName }) }}></div>
+    {% extends 'base.html.twig' %}
+
+    {% block body %}
+        <div {{ react_component('MyComponent', { 'fullName': number }) }}>
+            Loading... <i class="fas fa-cog fa-spin fa-3x"></i>
+        </div>
+        
+        {# Component living in a subdirectory: "assets/react/controllers/Admin/OtherComponent" #}
+        <div {{ react_component('Admin/OtherComponent') }}></div>
+    {% endblock %}
 
 Backward Compatibility promise
 ------------------------------


src/React/doc/index.rst
```diff
@@ -89,9 +89,13 @@ For example:
     }
 
 
+
 .. code-block:: html+twig
 
     {# templates/home.html.twig #}
+    
+    {% block stylesheets %}{{ encore_entry_link_tags('app') }}{% endblock %}
+    {% block javascripts %}{{ encore_entry_script_tags('app') }}{% endblock %}
```
weaverryan: Hmm. These ARE needed, but they're supposed to live in your `base.html.twig` file and they *should* already be there by default (when you install Twig, it ships with a `base.html.twig` that has these two lines). Did you not have these lines already in your `base.html.twig`?

hedocode: You are right ! 
Thank you for the precision, I didn't check the generated base.html.twig yet as the tutorial didn't extend it.

I added more informations so new users can make it work by copy-pasting, and still understand why it's needed with a link to related Encore documentation.

Sorry for the huge quantity of commits, I have been fighting with DOCtor RST, and he won by saying there is an error on a file I didn't change, on a line that does not have the problematic code : 

src/LiveComponent/doc/index.rst ✘
  581: Please use "html+twig" instead of "twig"


Have a good day !
Best regards

hedocode:
Another thing, I noticed that they are no explanation for components in subfolders, I expected it to work by writing the component file name but we actually need to write the path to it from the root folder, can I document it in this pull request or may I create a new one for this ?

src/React/doc/index.rst
weaverryan: Ok, let's see. Hmm. Earlier, on line 16 (first line under installation), we say:

> Before you start, make sure you have `Symfony UX configured in your app`_.

That is where, in theory, we're trying to point you to "make sure you have Encore fully installed and working in your app". By putting that note there, we don't need to include those details later on. Adding them here "muddies" the docs because these notes are really about setting up Encore/Symfony UX, and not about actually installing React.

Did you see that note on line 16? Is it too subtle? Could you suggest a change to it? I'd very much like to improve that as needed, and not include the new notes here (and, btw, absolutely no problem with many commits - keep going - we automatically squash when we merge).

src/React/doc/index.rst
```diff
@@ -89,10 +89,25 @@ For example:
...
 
     {# templates/home.html.twig #}
-    <div {{ react_component('MyComponent', { 'fullName': app.user.fullName }) }}></div>
+    {% extends 'base.html.twig' %}
+
+    {% block body %}
+        <div {{ react_component('MyComponent', { 'fullName': number }) }}>
+            Loading... <i class="fas fa-cog fa-spin fa-3x"></i>
+        </div>
+    {% endblock %}
```
weaverryan, line 104: I'm totally fine with this change - it makes it more complete, without distracting too much

weaverryan (commented):
> Another thing, I noticed that they are no explanation for components in subfolders, I expected it to work by writing the component file name but we actually need to write the path to it from the root folder, can I document it in this pull request or may I create a new one for this ?

Yea, this is something that should be mentioned - totally ok to include it here 👍

src/React/doc/index.rst
hedocode: > we're trying to point you to "make sure you have Encore fully installed and working in your app"

I noticed that after, that's a bit of  "recursion"  but with attentive read it's clear enough.

> these notes are really about setting up Encore/Symfony UX, and not about actually installing React

I agree with you that this page's responsibility is about Symfony React UX and therefore don't need to remind every dependencies such as Encore on it. According to that, I removed the Encore note and just kept the modified example extending base.html.twig.

hedocode:
> Yea, this is something that should be mentioned - totally ok to include it here +1

Done ! :heavy_check_mark:

src/React/doc/index.rst
```diff
@@ -77,6 +77,9 @@ React components located in the directory ``assets/react/controllers`` are regis
 React controller components.
 
 You can then render any React controller component in Twig using the ``react_component``.
+The first parameter is the path to the component from the root context folder, without extension at the end.
+The second is the component props object.
```
weaverryan: What about, instead of this note, we add a commented-out code example below - I'll suggest below

src/React/doc/index.rst
```diff
@@ -88,11 +91,16 @@ For example:
...
 .. code-block:: html+twig
 
     {# templates/home.html.twig #}
-    <div {{ react_component('MyComponent', { 'fullName': app.user.fullName }) }}></div>
+    {% extends 'base.html.twig' %}
+
+    {% block body %}
+        <div {{ react_component('MyComponent', { 'fullName': number }) }}>
+            Loading... <i class="fas fa-cog fa-spin fa-3x"></i>
+        </div>
```
weaverryan, line 100:
```suggestion
        </div>
        
        {# component living in a subdirectory: "assets/react/controllers/Admin/OtherComponent" #}
        <div {{ react_component('Admin/OtherComponent') }}></div>
```

The name `Admin/OtherComponent` is correct here, right? 😅

weaverryan:
Thank you @hedocode!

src/React/doc/index.rst
hedocode: That's correct ! I just verified it :)

hedocode:
You're welcome !

PR #786, opened 2023-04-13
[TwigComponent] Change component naming standard + optional name

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | None
| License       | MIT

Changes the naming convention of components to upper camel case + makes the `name` argument optional (defaults to class name). Components now look like:

```php
    #[AsTwigComponent]
    class Alert
    {
        public string $message;
    }
```

The template is `templates/components/Alert.html.twig` and you render with `{{ component('Alert') }}`. This removes the need for the user to invent & keep track of an arbitrary name, but they *can* still pass a name to `AsTwigComponent`. No BC break on this.

Relates to a conversation on https://github.com/symfony/ux/pull/771#issuecomment-1503915543

TODO: 

* [ ] Update MakerBundle to follow this convention


WebMamba:
I actually always follow this convention. This felt way more natural to have the same name on the PHP files and on the views side. So 100% agree here.

PR #794, opened 2023-04-15
[Live] Dispatch browser events

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | None
| License       | MIT

Hi!

I figured we'd need this eventually :). While building an component that opens/closes a modal, I needed a way to close the modal from a `LiveAction` in PHP. The way to do that is to dispatch a DOM event from the action, then listen to that with a tiny bit of JavaScript that closes the modal. Modeled after Livewire, like with emit(). Super simple.

Also fixed a bug with `attributes.add()` on non-live components.

Cheers!

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2114,13 +2114,13 @@ There are three ways to emit an event:
         data-event="productAdded"
     >
 
-2. From your PHP component via ``ComponentEmitsTrait``::
+2. From your PHP component via ``ComponentToolsTrait``::
 
-    use Symfony\UX\LiveComponent\ComponentEmitsTrait;
+    use Symfony\UX\LiveComponent\ComponentToolsTrait;
 
     class MyComponent
     {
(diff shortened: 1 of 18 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -2248,6 +2248,71 @@ Or, in PHP::
...
+            this.modal = Modal.getOrCreateInstance(this.element);
+            window.addEventListener('modal:close', () => this.modal.hide());
+        }
+    }
+
+Just make sure this controller is attached to the modal element:
+
+.. code-block:: html+twig
+
+    <div class="modal fade" {{ stimulus_controller('bootstrap-modal') }}">
```
norkunas, line 2296: Redundant quote at the end?

weaverryan: Absolutely - thanks for mentioning that :) - fixed in an upcoming PR

PR #796, opened 2023-04-16
[Autocomplete] Opt-out from auto updates if the root has a live id

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | -
| License       | MIT

When I have a form with dependent fields I can tell when a dependency changes
and if I use a custom `data-live-id` attribute which calculated based on the
dependencies the dynamic fields will re-render correctly and fixes the issue
with multi-select fields.

For example this seems to work well:

```php
    {
        $builder
            ->add('department', null, [
                'autocomplete' => true,
            ])
        ;

        $formModifier = function (FormInterface $form, Department $department = null) {
            $members = $department ? $this->organizationMemberRepository->findByDepartment($department) : [];
            $form->add('participatingMembers', null, [
                'autocomplete' => true,
                'choices' => $members,
                'attr' => [
                    'data-live-id' => 'participating-members-'.$department?->getId(),
                ],
            ]);
        };

        /// ...
    }

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -30,6 +30,9 @@ class default_1 extends Controller {
         this.isObserving = false;
     }
     initialize() {
+        if (this.element.hasAttribute('data-live-id')) {
+            return;
+        }
         if (this.requiresLiveIgnore()) {
             this.element.setAttribute('data-live-ignore', '');
             if (this.element.id) {
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -36,6 +36,10 @@ export default class extends Controller {
     private isObserving = false;
 
     initialize() {
+        if (this.element.hasAttribute('data-live-id')) {
+            return;
+        }
+
         if (this.requiresLiveIgnore()) {
             // unfortunately, TomSelect does enough weird things that, for a
             // multi select, if the HTML in the `<select>` element changes,


weaverryan:
Thanks for this @1ed! Question: does the `MutationObserver` stuff cause a problem? Your PR definitely makes sene for the `multiple` autocomplete case (where that simply doesn't work - so we add `data-live-ignore` and you'd want to avoid that since you're handling it manually). But for the "single" case, even if you ARE adding the `data-live-id` manually, was the `MutationObserver` causing some undesirable side effects?

1ed:
> But for the "single" case, even if you ARE adding the data-live-id manually, was the MutationObserver causing some undesirable side effects?

Actually, I didn't tried that scenario, yet. I think it will not cause any problems, but I can check it later. When I add a live-id and it changes it will re-render anyway, won't it? But if the live-id does not change should it have an observer looking for changes? I don't know, I don't have a use case in mind. I've thought, when I add a live-id I want to manually control when the component should re-render, but I'm fine with both ways (with or without the observer).

weaverryan:
> I think it will not cause any problems, but I can check it later

Yes, please let me know! The MutationObserver is a bit of new complexity on Autocomplete, and ideally we won't break existing behavior.

> When I add a live-id and it changes it will re-render anyway, won't it?

Yes. Live components *should* see this as "the old element was removed and a new element was added". In that case, the old element (which would contain a record of the mutations that TomSelect made to it when it was initializing) will be removed and its mutations will be ignored. But, this is something we should actually check ;) 

> I don't know, I don't have a use case in mind.

I DO think this change is likely valid for the `multiple` situation. In that case, you might be purposely using `data-live-id` to "take control"... and so you wouldn't want `data-live-ignore` added, I think.

Kocal:

Thanks!

Status: Needs Work

PR #792, opened 2023-04-14
Add docs for the new Twig Component syntax

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       |  
| License       | MIT

see #771

PR #799, opened 2023-04-16
[Twig] Tweaking docs for new HTML syntax

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | None
| License       | MIT

Just following up with a few more tweaks to @WebMamba's docs in #792.

Cheers!

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -778,9 +778,10 @@ child components works. Read `Live Nested Components`_.
 Embedded Components
 -------------------
 
-.. versionadded:: 2.2
+.. tip::
 
-    Embedded components were added in TwigComponents 2.2.
+    Embedded components (i.e. components with blocks) can be written in a more
+    readable way by using the `Component HTML Syntax`_.
 
 You can write your component's Twig template with blocks that can be overridden
 when rendering using the ``{% component %}`` syntax. These blocks can be thought of as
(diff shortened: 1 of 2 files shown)


WebMamba:
Thanks @weaverryan! Looks good to me! 😁

javiereguiluz (approved):

src/TwigComponent/doc/index.rst
```diff
@@ -840,86 +841,96 @@ The ``with`` data is what's mounted on the component object.
...
 
-    // "withActions" property will be set to true
-    <twig:Alert withActions message="hello"></:Alert>
-        
-You can add the ':' prefix to your attribute to indicate that the value 
-should be compiled by Twig
+    // withActions will be set to true
+    <twig:Alert type="info" message="hello!" withActions />
+
+To pass in a dynamic value, prefix the attribute with ``:`` or use the
```
javiereguiluz, line 880: Just asking ... is this convention based on the `:` prefix imposed somehow to us? Otherwise, it looks strange to introduce this new convention when we all know and use the `{{ ... }}` to output the value of variables or expressions. Thanks.

weaverryan: It's borrowed from Vue, and also is used [Blade Components](https://laravel.com/docs/10.x/blade#component-attributes). I agree that it feels like an "extra" thing, but it's also a very common syntax to see.

WebMamba: And you can still use the `{{ ... }}`, the `:` is just a small shortcut.

javiereguiluz: OK, so this is a very common syntax outside of Symfony. Then, it makes sense to have it too. Thanks.

src/TwigComponent/doc/index.rst
```diff
@@ -840,86 +841,96 @@ The ``with`` data is what's mounted on the component object.
...
     // or use a self-closing tag
-    <twig:Alert/>
-    
-You can pass props to your component by using HTML attributes. Suppose you have the following component:
+    <twig:Alert />
+
+Passing Props as HTML Attributes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Passing props is done with HTML attributes. For example if you have this component::
```
jdreesen, line 863:
```suggestion
Passing props is done with HTML attributes. For example if you have this component:
```

javiereguiluz: @jdreesen actually, the double `::` is needed. We use RST format instead of Markdown. code blocks are declared as `.. code-block:: php`. But since we have so many PHP code blocks, it's boring to do that. That's why we use a RST trick to tell it that any code following two `::` is to be considered as PHP code.

jdreesen: Oh, I didn't know that. Thx for the explanation.

kbond (approved):

src/TwigComponent/doc/index.rst
weaverryan: Btw, I found that it is more or less "needed" when you want to pass non-string values - like booleans or arrays:

```html
<twig:Alert :showClose="false" :messages="['hello', 'world']" />
```

PR #802, opened 2023-04-17
Anonymous TwigComponent

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | 
| License       | MIT

Hey all! 
TwigComponent getting really cool: a custom syntax, embedded component, nested component, and so on. Why not extends all of this to anonymous components?
What I call an anonymous component is a simple template.
For example if you have the following template:
```php
  // {% components/button/button.html.twig %}
  <button class="primary large actions">
    <span>
        {% block content %}
        {% endblock %}
    </span>
</button>
```

You can now use it like so:
```php

  {% index.html.twig %}

 ...
<div class="cart-actions">
   <twig:Button:Button>
       Click Me!
   </twig:Button:Button>
</div>

```
As can see you can use `:` to navigate through the directories.
And this PR came also with twig_entension configuration key (default to .html.twig), to find your template with the extension you want. 😁

You now have the {% props %} tag. This tag allows you to define required props for a static component or to set a default value.

```php
 // {% components/button/button.html.twig %}

{% props label, primary = true %}

  <button class="{{ primary ? = 'primary' ? 'secondary' }} large actions">
    <span>
        {{ label }}
    </span>
</button>
```

So here you have one require prop `label`, and one optional prop primary with a default value set to true. You can now use your anonymous component like so:

 ```php

  {% index.html.twig %}

 ...
<div class="cart-actions">
   <twig:Button.Button label="Click me">
</div>


kbond:
Love this! (attempted this in https://github.com/symfony/ux/pull/283 a while back but now that we have html-syntax and default slots, it's much more useful).

What about component properties/attributes?

```html
<twig:button.button class="foo" :username="user.name" :isAdmin>
    Click Me!
</twig:button.button>
```

weaverryan:
Ah, awesome! I'd really like to get this working too! To follow-up on Kevin's comment:

We need a way to differentiate a "variable" being passed to the template vs an attribute. Fortunately, we are not the first to face this problem ;) https://laravel.com/docs/10.x/blade#data-properties-attributes - so it may require some new tag on top of the component template - like `{% prop type='info', message %}` for a `type` prop that defaults to `info` and also a `message` prop with no default. I'm open to other syntax ideas! And can we get this working?

I'm also not sure, as a convention, to start mixing `<twig:SuccessButton/>` uppercase syntax with `<twig:button.success>` lowercase syntax. I can see how this differentiation could help you recognize an anonymous component from a class-backed component. But I'd like it to be easy to switch from an anonymous component to a class-backed one, without needing to go find everywhere you use that component in your templates and update the name.

kbond:
> I'm open to other syntax ideas! And can we get this working?

I like this syntax (if it's possible).

> I'm also not sure, as a convention, to start mixing <twig:SuccessButton/> uppercase syntax with <twig:button.success> lowercase syntax.

If I understand the new html syntax correctly, upper camel case is not required. `<twig:SuccessButton/>`, `<twig:success-button/>` and `<twig:success_button/>` are all valid (as long it matches your components name).

Assuming that's true, I don't know we can have a universal way to easily migrate from anon to non-anon. I feel like find-and-replace would be the best migration path.

I wonder if, to start, we should pass the full template?

```html
<twig:path/to/button.html.twig type="info" class="mb-1"/>
```

Or maybe we can drop the `.html.twig`:

```html
<twig:path/to/button type="info" class="mb-1"/>
```

weaverryan:
If we, as a standard, always used upper camel case - like <twig:SuccessButton> - even for anonymous components (so, the template would be SuccessButton.html.twig, then we’re good. Simply introducing a SuccessButton component class would be enough.

What likely wouldn’t be as clean are anonymous components that live in sub-directories.

kbond:
I think this would still be possible with what I propose.

For `<twig:{name}/>`, the following chain could occur (first match wins):
1. Check if `{name}` is a _real_ component
2. `{name}` exists as an _absolute_ twig template
3. `{name}.html.twig` exists as an _absolute_ twig template
4. `components/{name}` exists as twig template
5. `components/{name}.html.twig` exists as twig template

WebMamba:
>  so it may require some new tag on top of the component template - like {% prop type='info', message %} for a type prop that defaults to info and also a message prop with no default. I'm open to other syntax ideas!

I was thinking about something similar, but I think it's important to let the ability to type checking. So my proposal is something like so:
```php
{% prop 'info' is 'string'  default 'info' %}
{% prop 'foo' is 'bool' default 'false' %}
```

>  But I'd like it to be easy to switch from an anonymous component to a class-backed one,

This is a really good point! Because in my use the front-end developer first is the template, and then a Symfony developer as the back-end logic. So I think we should let the ability to add the component class without touching the templates.

> What likely wouldn’t be as clean are anonymous components that live in sub-directories

And why not also look at every subdirectory, and we just force to have every template with a different name under the components directory?

> I think this would still be possible with what I propose.

Hum yes, this is interesting, I like this priority check idea!

By mixing your comments the solution could be:
 
first, imagine you have the following template: components/Button/Primary.html.twig

then in an other template you use like so:
```php
   <twig:Button.Primary/>
```

and then if you create a new component class:
```php
#[AsTwigComponent('Button.Primary')]
class ButtonPrimary
{
}
```

weaverryan:
> I was thinking about something similar, but I think it's important to let the ability to type checking. So my proposal is something like so:

That syntax looks complex to me :). My opinion: if you are about types, use a class-backed component. There is nowhere else in Twig (e.g. `include()`) where there is type-safety when calling a template. Besides, we don't want to start seeing things like `{% prop 'foo' is 'App\\Entity\\Foo' default 'null' %}` ;)

So, I'd go back to my original proposal (well, I've modified to `props` with an `s`):

```twig
{% props type='info', message %}
```

> For <twig:{name}/>, the following chain could occur (first match wins):
> 
> 1. Check if `{name}` is a _real_ component
> 2. `{name}` exists as an _absolute_ twig template
> 3. `{name}.html.twig` exists as an _absolute_ twig template
> 4. `components/{name}` exists as twig template
> 5. `components/{name}.html.twig` exists as twig template

I might remove 2 & 3 (motivation: keep these anonymous components looking/feeling like everything else)? And is 4 really needed - I guess you'd have `<twig:foo.html.twig>`, which looks weird to me.

For 2 & 3, if we want to allow Twig components to live in other paths, probably we could add a different extension point for that - e.g. Blade components allow you to configure a "namespace" of sorts - e.g. a `Dashboard` namespace could be configured to look in `templates/dashboard/` and then if you use `<twig:dashboard:Panel>`, that would look in `templates/dashboard/Panel.html.twig`. But... that's to think about later.

src/TwigComponent/src/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,22 @@
...
+
+class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $treeBuilder = new TreeBuilder('twig_component');
+        $rootNode = $treeBuilder->getRootNode();
+        $rootNode
+            ->children()
+                ->scalarNode('template_extension')->defaultValue('.html.twig')->end()
```
kbond: I think we should drop this config option (for now). If we use some form of the chain I suggest, this should not be needed anyway.

Probably not realistic but could be a problem if a 3rd party bundle expects `.html.twig` but you've customized it in your app.

kbond:
> first, imagine you have the following template: components/Button/Primary.html.twig

I think we should keep the directory separator as `/` for now so there's no need for any string manipulation.

> I might remove 2 & 3 (motivation: keep these anonymous components looking/feeling like everything else)? And is 4 really needed - I guess you'd have <twig:foo.html.twig>, which looks weird to me.

As long as (for an end-app), `<twig:form/input/text />` resolves to `components/form/input/text.html.twig` I'm good but...

While it looks weird, I think this is the easiest way to allow including anon. templates from 3rd party bundles: `<twig:@SomeBundle/form/input/text />`

Is the performance of running the chain part of your objection? I _do_ think we should put the best-practice/most-common scenarios highest on the chain.

kbond:
> (motivation: keep these anonymous components looking/feeling like everything else)

I think a lot of people, myself included, would want to use these as super-charged include tags and may not have any intention of using _real_ components.

src/TwigComponent/src/DependencyInjection/Configuration.php
weaverryan: less config is better to start - then we can collect real use-cases on what's needed 👍

weaverryan:
Ok fair - we can keep 2 & 3 :)

AmProsius:
I don't want to wake sleeping dogs, but why is the directory separator different from the [template namespace](https://symfony.com/bundles/ux-twig-component/current/index.html#twig-template-namespaces)?

That would mean an anonymous twig component would be written differently than a class component:

```twig
{# components/button/button.html.twig #}

<button class="btn btn-{{ variant }}">
</button>

{# Anonymous component #}

<twig:button.button variant="primary">Click me</twig:button.button>
{# or #}
<twig:button/button variant="primary">Click me</twig:button/button>

{# Class component #}

<twig:button:button variant="primary">Click me</twig:button:button>
```

It would be great if the directory separators would be similar. That way, devs will be able to just remove the component class and the components would still work.
Plus it removes mental overload, because you don't have to remember when to use which separator.

WebMamba:
Hey @AmProsius! Thanks for your comment! I think you are right and this is the direction we gonna took here! I will keep working on this PR this week! 😁 I was just focused on the bug we have with the new twig component syntax 😁 Cheers!

seb-jean:
For this PR you could take help from https://laravel.com/docs/10.x/blade#anonymous-components 😄

WebMamba:
Hey all! I think we need to work first on what @weaverryan talks about in this comment https://github.com/symfony/ux/issues/844#issuecomment-1543805445 before continuing this PR 😁

WebMamba:
Hey all! I am back on this! I updated the first comment if you want some examples, but what I did is:
- now you put your static components in template templates/*, templates/*.html.twig,  templates/component/*,   templates/component/*.html.twig
- You have new props that let you define optional or require props for your component

Tell me what you think about this change! Cheers 🧡

src/TwigComponent/src/DependencyInjection/Configuration.php
kbond: I still think we should drop the config to start.

Edit: I think this is not used anyway, dead code?

src/TwigComponent/src/StaticComponent.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Symfony\UX\TwigComponent;
+
+class StaticComponent
```
kbond:
```suggestion
class AnonymousComponent
```

? I see you are using the term _static_ now - are you wanting to use this term instead of _anonymous_?

And we should mark this as internal.

src/TwigComponent/tests/Fixtures/templates/render_static_component.html.twig
```diff
@@ -0,0 +1 @@
+{{ component('static.button') }}
```
kbond: I'm confused what this is for. Is this being used anywhere?

src/TwigComponent/tests/Fixtures/templates/anonymous_component.html.twig
```diff
@@ -0,0 +1 @@
+<twig:Button label='Click me'/>
```
kbond, line 1: I'm thinking we should add a test proving the nested directory system.

kbond (commented):
Looking good, I'm excited! I'm happy you got the prop system working.

> now you put your static components in template templates/, templates/.html.twig, templates/component/, templates/component/.html.twig

Does this mean you can use `<twig:Form.Button.html.twig>`?

kbond:
From your description above:

> As can see you can use `.` to navigate through the directories.

This should be `:`, correct?

WebMamba:
Ha yes good catch! I changed to :

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -43,7 +45,14 @@ public function metadataFor(string $name): ComponentMetadata
...
         if (!$config = $this->config[$name] ?? null) {
-            $this->throwUnknownComponentException($name);
+            if (($template = $this->findAnonymousComponentTemplate($name)) !== null) {
+                return new ComponentMetadata([
+                    'key' => $name,
+                    'template' => $template,
+                ]);
+            }
+
+            throw new \InvalidArgumentException(sprintf('Unknown component "%s". The registered components are: %s. And no template anonymous component founded', $name, implode(', ', array_keys($this->config))));
```
WebMamba: If you have some idea to make this error more helpful with the anonymous component...

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -149,7 +164,11 @@ private function getComponent(string $name): object
         $name = $this->classMap[$name] ?? $name;
 
         if (!$this->components->has($name)) {
-            $this->throwUnknownComponentException($name);
+            if ($this->isAnonymousComponent($name)) {
+                return new AnonymousComponent();
+            }
+
+            throw new \InvalidArgumentException(sprintf('Unknown component "%s". The registered components are: %s. And no anonymous component founded', $name, implode(', ', array_keys($this->components->getProvidedServices()))));
```
WebMamba: here to

src/TwigComponent/src/Twig/PropsNode.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        foreach ($this->getAttribute('names') as $name) {
+            $compiler
+                ->addDebugInfo($this)
+                ->write('if (!isset($context[\''.$name.'\'])) {')
+            ;
+
+            if (!$this->hasNode($name)) {
+                $compiler
+                    ->write('throw new \Exception("'.$name.' should be defined for component '.$this->getTemplateName().'");')
```
WebMamba: And here to make something more helpful

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -30,11 +31,12 @@ final class ComponentFactory
      * @param array<class-string, string> $classMap
      */
     public function __construct(
+        private Environment $environment,
```
weaverryan: WDYT about passing something smaller here - a new `ComponentTemplateFinderInterface` - with a single method on it.

src/TwigComponent/src/ComponentFactory.php
weaverryan:
```suggestion
            throw new \InvalidArgumentException(sprintf('Unknown component "%s". The registered components are: %s. And no matching anonymous component template was found.', $name, implode(', ', array_keys($this->config))));
```

And we can move it into `throwUnknownComponentException()`, I think, and call it from both places.

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -124,6 +133,12 @@ private function mount(object $component, array &$data): void
             return;
         }
 
+        if ($component instanceof AnonymousComponent) {
+            $component->mount($data);
+
+            return;
+        }
```
weaverryan, line 139: Hmm. It's not mandatory, but I'd love for `AnonymousComponent` to not need a special case. Maybe this is possible:

```php
public function mount(...$props = [])
{

}
```

Tell me if this makes sense: if the final argument of `mount()` is a `...$props` variadic-style argument, then we pass everything. This should also be possible:

```
public function mount(string $name, ...$props = [])
{

}
```

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -189,6 +208,35 @@ private function postMount(object $component, array $data): array
...
+        if ($loader->exists($componentPath.'.html.twig')) {
+            return $componentPath.'.html.twig';
+        }
+
+        if ($loader->exists('components/'.$componentPath)) {
+            return 'components/'.$componentPath;
+        }
+
+        if ($loader->exists('/components/'.$componentPath.'.html.twig')) {
+            return '/components/'.$componentPath.'.html.twig';
```
weaverryan: Are the starting `/` there on purpose? I'm not familiar with Twig template paths starting with `/`

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -105,6 +105,7 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
 
             // expose public properties and properties marked with ExposeInTemplate attribute
             iterator_to_array($this->exposedVariables($component, $metadata->isPublicPropsExposed())),
+            $component instanceof AnonymousComponent ? $component->getProps() : []
```
weaverryan: WDYT about somehow reusing `#[ExposeInTemplate]` to do this, instead of this special case? We would need some new option on that attribute, which would mean "I am returning an associative array of variable names" - I'm not sure about a good name:

```php
#[ExposeInTemplate(destructure: true)
public function getProps(): array
{
    return $this->props;
}
```

src/TwigComponent/src/Twig/PropsNode.php
weaverryan: > The prop "$name" defined in the {% props %} tag was not passed when rendering the component.

Also, I'm not sure which type of `\Exception` should be thrown - but maybe you can check out some similar code.

src/TwigComponent/src/Twig/PropsNode.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        parent::__construct($values, ['names' => $propsNames], $lineno, $tag);
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        foreach ($this->getAttribute('names') as $name) {
+            $compiler
+                ->addDebugInfo($this)
+                ->write('if (!isset($context[\''.$name.'\'])) {')
```
weaverryan, line 34: Probably need some check in case this is used entirely outside of a component.

weaverryan (commented):
This is looking VERY promising - thanks for your work on this moving target!

src/TwigComponent/src/ComponentFactory.php
WebMamba: It's a good idea and that let more flexibility for people to overwrite and structure differently their project

src/TwigComponent/src/ComponentFactory.php
WebMamba: I tried but looks like we just need a little tweak here if we don't wan't to move too much code.

src/TwigComponent/src/ComponentRenderer.php
WebMamba: I made something, tell me what you think about it

WebMamba:
I don't think the failure is related to this PR.

I made some change it's ready for review again 😁

Thats was super good idea to add this props, I tried on a real project and it's super usefull 😁

WebMamba:
@kbond @weaverryan friendly ping here. I think we are really close, I am so thrilled to have this feature 😁

src/TwigComponent/src/AnonymousComponent.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Symfony\UX\TwigComponent;
+
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+
+/**
+ * @author Matheo Daninos <matheo.daninos@gmail.com>
+ *
+ * @internal
+ */
+class AnonymousComponent
```
weaverryan:
```suggestion
final class AnonymousComponent
```

src/TwigComponent/src/Attribute/ExposeInTemplate.php
```diff
@@ -22,12 +22,14 @@
...
     /**
-     * @param string|null $name   The variable name to expose. Leave as null
-     *                            to default to property name.
-     * @param string|null $getter The getter method to use. Leave as null
-     *                            to default to PropertyAccessor logic.
+     * @param string|null $name     The variable name to expose. Leave as null
+     *                              to default to property name.
+     * @param string|null $getter   The getter method to use. Leave as null
+     *                              to default to PropertyAccessor logic.
+     * @param bool        $destruct The content should be used as array of variables
```
weaverryan:
```suggestion
     * @param bool        $destruct The content should be used as array of variable
```

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -148,6 +148,14 @@ private function exposedVariables(object $component, bool $exposePublicProps): \
                 throw new \LogicException(sprintf('Cannot use %s on methods with required parameters (%s::%s).', ExposeInTemplate::class, $component::class, $method->name));
             }
 
+            if ($attribute->destruct) {
```
weaverryan, line 157: We need similar code near the top of the function when looping over properties. You could have a property with the attribute on it and this new option.

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+use Twig\Environment;
+
+/**
+ * @author Matheo Daninos <matheo.daninos@gmail.com>
+ */
+class ComponentTemplateFinder implements ComponentTemplateFinderInterface
```
weaverryan:
```suggestion
final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
```

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -68,7 +79,7 @@ class_exists(AbstractArgument::class) ? new AbstractArgument(sprintf('Added in %
             ])
         ;
 
-        $container->setAlias(ComponentRendererInterface::class, 'ux.twig_component.component_renderer');
+        $container->setAlias(ComponentTemplateFinderInterface::class, 'ux.twig_component.component_template_finder');
```
weaverryan: I think this isn't needed. This isn't something I'd expect users to need to autowire.

src/TwigComponent/tests/Fixtures/templates/components/Button.html.twig
```diff
@@ -0,0 +1,5 @@
+{% props label, primary = true %}
+
+<button class="{{ primary ? 'primary' : 'secondary' }}">
```
weaverryan: We should use `attributes` here:

```twig
<button {{ attributes.defaults({ class: primary ? 'primary' : 'secondary' }) }}>
```

The reason is that, we need to test that anything that is a prop is NOT added as an attribute (and I think this may not be implemented yet?). So given this example:

```html
<twig:Button label='Click me' :primary='false'/>
```

We need to assert that, even if we use `{{ attributes }}`, both `label` and `primary` don't show up as attributes.

weaverryan (commented):
Getting very close now! I dropped some comments - but we also need documentation!

src/TwigComponent/tests/Fixtures/templates/components/Button.html.twig
WebMamba: Hum, I am not sure we should have a difference between argument and prop. Why do you think we should differentiate theses two notions?

WebMamba:
@weaverryan for the doc I plan to let a colleague do it, it's her first PR 😁

src/TwigComponent/doc/.idea/.gitignore
Kocal, line 1: The `.idea` folder must be removed by you or your colleague, it must be globally ignored instead.

Kocal (changes requested):

kbond:
I think it's been decided that `:` is the dir separator? Do you mind updating the PR description?

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+use Twig\Environment;
+
+/**
+ * @author Matheo Daninos <matheo.daninos@gmail.com>
+ */
+final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
```
kbond, line 19: This feels a bit like a premature abstraction, what about in-lining in `ComponentFactory` for now?

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -0,0 +1,49 @@
...
+        private Environment $environment
+    ) {
+    }
+
+    public function findAnonymousComponentTemplate(string $name): ?string
+    {
+        $loader = $this->environment->getLoader();
+        $componentPath = rtrim(str_replace(':', '/', $name));
+
+        if ($loader->exists($componentPath)) {
```
kbond: Maybe a micro-optimization but should we order these checks to make what we call the best practice first? Here's my thinking:

1. `components/{$name}.html.twig`
2. `{$name}.html.twig`
3. `components/{$name}`
4. `{$name}`

src/TwigComponent/src/ComponentTemplateFinder.php
WebMamba: @weaverryan ask me to do it in this comment: https://github.com/symfony/ux/pull/802#discussion_r1253396681.
To not have to autowire the Environment service

src/TwigComponent/doc/.idea/.gitignore
WebMamba: Houuuuuch good catch! Thanks for it! 😁

WebMamba:
Hey all! I am back from holiday! And back to making this PR live 😎

@kbond PR description updated 👍

src/TwigComponent/src/ComponentTemplateFinder.php
weaverryan: Yea, it felt odd to inject ALL of `Environment` into `ComponentFactory`, which is otherwise unaware of Twig iirc

weaverryan:
Thank you for your hard work @WebMamba! Now, can you take a shot at writing the documentation?

WebMamba:
Woup Woup 🕺🏽 !! So happy to have this merge! 😁🎊🎉

@weaverryan I am on it 😎

seb-jean:
Are arrays taken into account?

Example:
```twig
<twig:TabbedCodeBlocks :files="[
    'src/Twig/MealPlanner.php',
    'templates/components/MealPlanner.html.twig',
    'src/Form/MealPlannerForm.php',
]" />
```

I have an error
`Error rendering "TabbedCodeBlocks" component: A "files" prop was passed when creating the "TabbedCodeBlocks" component. No matching files property or mount() argument was found, so we attempted to use this as an HTML attribute. But, the value is not a scalar (it's a array). Did you mean to pass this to your component or is there a typo on its name?`

seb-jean:
I tried to make a simple example

```twig
<twig:Button label='Click me' :primary='false'/>
```

```twig
{% props label, primary = true %}

<button {{ attributes.defaults({class: primary ? 'primary' : 'secondary'}) }}>
    {{ label }}
</button>
```

The render is as follows but I have `label="Click me"` in `button` element.  It’s too bad! Is there a way not to have the `label` present?
```html
<button class="secondary" label="Click me">
    Click me
</button>
```

WebMamba:
@seb-jean for the first issue you point out I open a fix: #1041 

for the second issue the function `attributes.only` or `attributes.without` should do the fix.

seb-jean:
thanks @WebMamba

seb-jean:
I would like to know if it is possible to have a dynamic component name.
See the code below:

```twig
{% set items = [
    {icon: 'Home'},
    {icon: 'MagnifyingGlass'},
    {icon: 'Bell'},
    {icon: 'Users'},
    {icon: 'Heart'},
] %}
{% for item in items %}
    <twig:Icon:{item.icon} class="h-6 w-6" />
{% endfor %}
```

PR #806, opened 2023-04-19
Fix error for component name with :

### LINKED ISSUE #803: [TwigComponent] namespace double dot isn't parsed correctly
First of all, thanks for this awesome component!

I noticed that create a twig component using a namespace wasn't working as expected using some syntaxes.

Let's assume a component named `foo:bar`.
Using quotes, it works as expected:

```twig
{% component 'foo:bar' %}
  {# ... #}
{% endcomponent %}
```

But if I try the same without quotes or using the html syntax:

```twig
{% component foo:bar %}
  {# ... #}
{% endcomponent %}

{# OR #}

<twig:foo:bar>
  {# ... #}
</twig:foo:bar>
```

This throw a `SyntaxError`:
```
An exception has been thrown during the compilation of a template ("Unknown component "foo". The registered components are: foo:bar").
```

As you can see, it is only taking in account the namespace as component name.

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #803 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -58,7 +58,7 @@ class TwigPreLexer
                     $this->currentComponents[] = ['name' => $componentName, 'hasDefaultBlock' => false];
                 }
 
-                $output .= "{% component {$componentName}".($attributes ? " with { {$attributes} }" : '').' %}';
+                $output .= "{% component '{$componentName}'".($attributes ? " with { {$attributes} }" : '').' %}';
                 if ($isSelfClosing) {
                     $output .= '{% endcomponent %}';
                 }
@@ -117,7 +117,7 @@ class TwigPreLexer
     private function consumeComponentName(): string
     {
         $start = $this->position;
-        while ($this->position < $this->length && preg_match('/[A-Za-z0-9_]/', $this->input[$this->position])) {
+        while ($this->position < $this->length && preg_match('/[A-Za-z0-9_:@\-\/.]/', $this->input[$this->position])) {
             ++$this->position;
         }
         $componentName = substr($this->input, $start, $this->position - $start);
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -29,47 +29,67 @@ final class TwigPreLexerTest extends TestCase
     {
         yield 'simple_component' => [
             '<twig:foo />',
-            '{% component foo %}{% endcomponent %}',
+            '{% component \'foo\' %}{% endcomponent %}',
         ];
 
         yield 'component_with_attributes' => [
             '<twig:foo bar="baz" with_quotes="It\'s with quotes" />',
-            "{% component foo with { bar: 'baz', with_quotes: 'It\'s with quotes' } %}{% endcomponent %}",
+            "{% component 'foo' with { bar: 'baz', with_quotes: 'It\'s with quotes' } %}{% endcomponent %}",
         ];
 
         yield 'component_with_dynamic_attributes' => [
             '<twig:foo dynamic="{{ dynamicVar }}" :otherDynamic="anotherVar" />',
-            '{% component foo with { dynamic: dynamicVar, otherDynamic: anotherVar } %}{% endcomponent %}',
+            '{% component \'foo\' with { dynamic: dynamicVar, otherDynamic: anotherVar } %}{% endcomponent %}',
         ];
 
         yield 'component_with_closing_tag' => [
             '<twig:foo></twig:foo>',
-            '{% component foo %}{% endcomponent %}',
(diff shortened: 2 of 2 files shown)


src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -117,7 +117,7 @@ public function preLexComponents(string $input): string
     private function consumeComponentName(): string
     {
         $start = $this->position;
-        while ($this->position < $this->length && preg_match('/[A-Za-z0-9_]/', $this->input[$this->position])) {
+        while ($this->position < $this->length && preg_match('/[A-Za-z0-9_:]/', $this->input[$this->position])) {
```
WebMamba: Do you think about other characters that should be allowed here?

WebMamba:
This is a fix on the HTML twig component syntax but do you think we should extend the fix to allow this
`` {% component foo:bar `` for the moment only ``{% component 'foo:bar' %}`` is allow?

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -58,7 +58,7 @@ public function preLexComponents(string $input): string
                     $this->currentComponents[] = ['name' => $componentName, 'hasDefaultBlock' => false];
                 }
 
-                $output .= "{% component {$componentName}".($attributes ? " with { {$attributes} }" : '').' %}';
+                $output .= "{% component '$componentName'".($attributes ? " with { {$attributes} }" : '').' %}';
```
jdreesen:
```suggestion
                $output .= "{% component '{$componentName}'".($attributes ? " with { {$attributes} }" : '').' %}';
```

WebMamba: good catch thanks

src/TwigComponent/src/Twig/TwigPreLexer.php
kbond: Should we future proof this by allowing `-`, `/`, `@` and `.`?

WebMamba: and - ?

kbond: `<twig:my-component/>`?

WebMamba: yeah why not! We never know what people can bring 😅

WebMamba: ha ok fine that was the first on your list! I didn't see it! Sorry about that

kbond: (I edited the comment) 😁

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
```diff
@@ -29,47 +29,47 @@ public function getLexTests(): iterable
...
             '<twig:foo>Foo</twig:foo>',
-            '{% component foo %}{% block content %}Foo{% endblock %}{% endcomponent %}',
+            '{% component \'foo\' %}{% block content %}Foo{% endblock %}{% endcomponent %}',
         ];
 
         yield 'component_with_default_block_that_holds_a_component_and_multi_blocks' => [
             '<twig:foo>Foo <twig:bar /><twig:block name="other_block">Other block</twig:block></twig:foo>',
-            '{% component foo %}{% block content %}Foo {% component bar %}{% endcomponent %}{% endblock %}{% block other_block %}Other block{% endblock %}{% endcomponent %}',
+            '{% component \'foo\' %}{% block content %}Foo {% component \'bar\' %}{% endcomponent %}{% endblock %}{% block other_block %}Other block{% endblock %}{% endcomponent %}',
         ];
```
weaverryan, line 73: Let's add tests specifically for any of the new character we allow - e.g. `<twig:foo:bar` or `<twig:foo-bar`, etc

weaverryan (commented):
Thanks for getting this rolling ❤️

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -58,7 +58,7 @@ public function preLexComponents(string $input): string
                     $this->currentComponents[] = ['name' => $componentName, 'hasDefaultBlock' => false];
                 }
 
-                $output .= "{% component {$componentName}".($attributes ? " with { {$attributes} }" : '').' %}';
+                $output .= "{% component '{$componentName}'".($attributes ? " with { {$attributes} }" : '').' %}';
```
kbond, line 61: Does `{% component @foo/bar-baz.html.twig %}` (no `'`) not work?

WebMamba: Yes It works with both

kbond (approved):

weaverryan:
Thanks Matheo!

Jupi007:
Thanks a lot for the fix @WebMamba !

Issue #810, opened 2023-04-19
[TwigComponent] New HTML syntax marked as embedded in all cases - fails for live components


weaverryan:
Take this component render:

```html
<twig:PackageSearch />
```

Internally, we remap this to:

```twig
{% component PackageSearch %}{% endcomponent%}
```

The key here is that there is NO content inside of the component: we are not defining any blocks. Even still, `ComponentRenderer::embeddedContext()` is called, which marks the component as "embedded". And so, trying to use with a live component fails as this is not allowed

@kbond WDYT? We could work around this with the HTML syntax to map self-closing tags to `{{ component(...) }}`. But it also feels more correct to allow `{% component PackageSearch %}{% endcomponent%}` to realize that it's not really an "embedded" component.

kbond:
> But it also feels more correct to allow {% component PackageSearch %}{% endcomponent%} to realize that it's not really an "embedded" component.

True but I'm not really concerned. We're almost at the point where we can make the `{% component %}` tag _internal_ (remove all references to it in the docs)

weaverryan:
Closing this now - it's effectively fixed

PR #811, opened 2023-04-19
[TwigComponent] Making self-closing HTML syntax render with simpler component() syntax

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (partial)
| New feature?  | no
| Tickets       | Workaround for #810
| License       | MIT


WebMamba (approved):
Completly agree here! If you didn't make this PR, I planed to do the exact same one 😁

kbond:
So to clarify, using a self closing twig tag allows it to be a live component?

weaverryan:
> So to clarify, using a self closing twig tag allows it to be a live component?

Exactly right :)

PR #814, opened 2023-04-20
remove / character in the list of allowed character

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | 
| License       | MIT

By allowing the / character in #806, we introduce a bug. You can't now use the self-closing syntax because it enters in conflict with the component name.
```php
<twig:foo/>
```
The Lexer thinks now that the name is `foo/`
I just remove the / from the allowing list, tell me if you think we should go a bit further.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -123,7 +123,7 @@ class TwigPreLexer
     private function consumeComponentName(string $customExceptionMessage = null): string
     {
         $start = $this->position;
-        while ($this->position < $this->length && preg_match('/[A-Za-z0-9_:@\-\/.]/', $this->input[$this->position])) {
+        while ($this->position < $this->length && preg_match('/[A-Za-z0-9_:@\-.]/', $this->input[$this->position])) {
             ++$this->position;
         }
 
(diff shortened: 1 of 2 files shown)


kbond:
Hmm, the `/` would be required for #802. Is is possible to allow it for any character but the last?

weaverryan:
> Hmm, the / would be required for https://github.com/symfony/ux/pull/802. Is is possible to allow it for any character but the last?

Unless we force a `.` syntax instead - that's another thing Blade components do. This would be my preference, at least for now. It'll make life simpler - and we could consider adding `/` back later.

weaverryan:
> By allowing the / character in https://github.com/symfony/ux/pull/806, we introduce a bug. You can't now use the self-closing syntax because it enters in conflict with the component name

The tests on #806 were passing, so we must be missing a test-case for this. Can you add one in this PR to "prove" the bug that you're fixing?

kbond:
> Unless we force a . syntax instead

This would disallow adding .html.twig but I'm fine with that. "Anonymous twig template files must end in .html.twig". If someone has a use-case where this doesn't work for them we can re-evaluate.

WebMamba:
> The tests on https://github.com/symfony/ux/pull/806 were passing, so we must be missing a test-case for this. Can you add one in this PR to "prove" the bug that you're fixing?

I just push a test for it 👍

This is also my thought : 
> This would disallow adding .html.twig but I'm fine with that. "Anonymous twig template files must end in .html.twig"

I just want to avoid the code being more complex for edge cases. Forcing the `.` syntax as Blade does is the way to go for me.

weaverryan:
Thank you Matheo!

PR #813, opened 2023-04-19
[LiveComponent] Adding more detail about loops & keys behavior

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | None
| License       | MIT

One of the trickier things in any frontend system like this is dealing with loops of elements or components. This expands on the existing docs to clarify that `data-live-id` is sometimes needed for loops of elements and also to explain an expected situation that might cause a WTF moment for people.

Cheers!

PR #815, opened 2023-04-20
Add doc for another usage of "data-live-id"

jcrombez:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2693,6 +2693,7 @@ The system doesn't handle every edge case, so here are some things to keep in mi
 * If an element is moved from one location in the component to another,
   that change is **lost**: the element will be re-added in its original location
   during the next re-render.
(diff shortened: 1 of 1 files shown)


weaverryan:
After a bit more thinking, I'm going to close this in favor of #813. The issue you were having is not actually related to the smart component re-rendering system: it's just the classic "if you're looping over elements, you may need a `data-live-id`". In your case, if you have 3 `<div>` elements (let's call them A, B , C), and they are reordered (on the client side or on the server),  to B, A, C, then when the component re-renders, it will actually:

* Take div A and change it to have the contents of div B
* Take div B and change it to have the contents of div A

The problem is that, if you have a Stimulus controller attached to each of these elements and you're storing some sort of state on the controller instance - e.g. `this.isHighlighted = true`, then suddenly your controller A instance will live on what appears to be the div B, and controller B instance will live on what appears to be the div A.

Anyways, thank you for bringing this up - you inspired the new docs there. If/when this bites people, we can continue to work on it.

PR #817, opened 2023-04-22
Fix error with nested component

### LINKED ISSUE #805: [TwigComponent] Nested component doesn't work with html syntax
Let's assume two component `foo` and `bar`, which both define a content block.

Using the normal syntax, it works as expected:

```twig
{% component foo %}
  {% block content %}
    {% component bar %}
      {% block content %}
        {# ... #}
      {% endblock %}
    {% endcomponent %}
  {% endblock %}
{% endcomponent %}
```

But using the html syntax: 

```twig
<twig:foo>
  <twig:block name="content">
    <twig:bar>
      <twig:block name="content">
        {# ... #}
      </twig:block>
    </twig:bar>
  </twig:block>
</twig:foo>
```

This throw a `SyntaxError`:

```
An exception has been thrown during the compilation of a template ("Expected closing tag '</twig:block>' for block 'content' at line 13").
```

---

Btw, simply write it like this doesn't work too:

```twig
<twig:foo>
  <twig:bar>
    {# ... #}
  </twig:bar>
</twig:foo>
```

This one throw:

```
A template that extends another one cannot include content outside Twig blocks. Did you forget to put the content inside a {% block %} tag?
```

But maybe it is how it is expected to work?

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #805 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -314,7 +314,7 @@ class TwigPreLexer
         if (!$this->doesStringEventuallyExist($closingTag)) {
             throw new SyntaxError("Expected closing tag '{$closingTag}' for block '{$blockName}'.", $this->line);
         }
-        $blockContents = $this->consumeUntil($closingTag);
+        $blockContents = $this->consumeUntilEndBlock();
 
         $subLexer = new self($this->line);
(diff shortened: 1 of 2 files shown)


weaverryan (approved):
👏 ⭐

weaverryan:
Thank you for tackling this tricky one!!

PR #820, opened 2023-04-27
Resolve twig syntax in twig component arguments

### LINKED ISSUE #812: [TwigComponent] Twig HTML syntax in attribute can't be mixed
Hi!

Still finding edge cases in the new HTML twig syntax:

```
<twig:Something
    key="new-feature-{{ features|length }}"
/>
```

In this case, `key` will pass in as the literal string `new-feature-{{ features|length }}`. We're handling attributes that *start* with `{{`, but not strings with a `{{` later, or even multiple `{{`. Probably easy to fix - I think I remember introducing this shortcoming.

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #812 
| License       | MIT

### DIFF
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -96,5 +96,9 @@ final class TwigPreLexerTest extends TestCase
             '<twig:foo><twig:block name="child"><twig:bar><twig:block name="message">Hello World!</twig:block></twig:bar></twig:block></twig:foo>',
             '{% component \'foo\' %}{% block child %}{% component \'bar\' %}{% block message %}Hello World!{% endblock %}{% endcomponent %}{% endblock %}{% endcomponent %}',
         ];
+        yield 'component_with_html_syntaxt_in_argument' => [
+            '<twig:foo text="Hello {{ name }}!"/>',
+            "{{ component('foo', { text: 'Hello '~(name)~'!' }) }}",
+        ];
     }
 }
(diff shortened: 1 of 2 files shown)


weaverryan:
Thank you Matheo!

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
```diff
@@ -96,5 +96,9 @@ public function getLexTests(): iterable
             '<twig:foo><twig:block name="child"><twig:bar><twig:block name="message">Hello World!</twig:block></twig:bar></twig:block></twig:foo>',
             '{% component \'foo\' %}{% block child %}{% component \'bar\' %}{% block message %}Hello World!{% endblock %}{% endcomponent %}{% endblock %}{% endcomponent %}',
         ];
+        yield 'component_with_html_syntaxt_in_argument' => [
+            '<twig:foo text="Hello {{ name }}!"/>',
+            "{{ component('foo', { text: 'Hello '~(name)~'!' }) }}",
```
Kocal, line 101: What happens if we have a single quote like this `<twig:foo text="Hello {{ name }}, I'm Theo!"/>`?, is it properly escaped?

weaverryan: We have a test for a similar situation https://github.com/symfony/ux/blob/2.x/src/TwigComponent/tests/Unit/TwigPreLexerTest.php#L37 - which is passing. But oddly, I DOES look like we may have lost this escaping: https://github.com/symfony/ux/pull/820/files#diff-325c6a70c3103b97d203266f5b3b33ff88552c3d50c6219340b90ea38c77a843L196

Test case needed to confirm.

weaverryan: Just tested locally - it works. I'll include a test case in a PR

PR #821, opened 2023-04-28
[Twig] add test helper

### LINKED ISSUE #818: [TwigComponent] Provide class/trait for testing component
It would be great to provide a `ComponentTestCase` class or `ComponentAssertionTrait` trait to help testing component.

I was thinking of something similar to [MailerAssertionsTrait](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Bundle/FrameworkBundle/Test/MailerAssertionsTrait.php) so we can:

- get an instance of the component from its name and parameters
- assert that the output of the component is similar to an expected value
- etc ...

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #818
| License       | MIT

_This works with Live components out of the box (not with slots though as they aren't supported)._

```php
use Symfony\UX\TwigComponent\Test\InteractsWithTwigComponents;

class MyComponentTest extends KernelTestCase
{
    use InteractsWithTwigComponents;

    public function testComponentMount(): void
    {
        $component = $this->mountTwigComponent(
            name: 'MyComponent', // can also use FQCN (MyComponent::class)
            data: ['foo' => 'bar'],
        );

        $this->assertInstanceOf(MyComponent::class, $component);
        $this->assertSame('bar', $component->foo);
    }

    public function testComponentRenders(): void
    {
        $rendered = $this->renderTwigComponent(
            name: 'MyComponent', // can also use FQCN (MyComponent::class)
            data: ['foo' => 'bar'],
        );

        $this->assertStringContainsString('bar', $rendered);
    }

    public function testEmbeddedComponentRenders(): void
    {
        $rendered = $this->renderTwigComponent(
            name: 'MyComponent', // can also use FQCN (MyComponent::class)
            data: ['foo' => 'bar'],
            content: '<div>My content</div>', // "content" (default) block
            blocks: [
                'header' => '<div>My header</div>',
                'menu' => $this->renderTwigComponent('Menu'), // can embed other components
            ],
        );

        $this->assertStringContainsString('bar', $rendered);
    }
}
```

### DIFF
--- a/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
+++ b/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
@@ -1,0 +1,67 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Test;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+trait InteractsWithTwigComponents
+{
+    protected function mountTwigComponent(string $name, array $data = []): object
+    {
+        if (!$this instanceof KernelTestCase) {
+            throw new \LogicException(sprintf('The "%s" trait can only be used on "%s" classes.', __TRAIT__, KernelTestCase::class));
+        }
+
+        return static::getContainer()->get('ux.twig_component.component_factory')->create($name, $data)->getComponent();
+    }
+
+    /**
+     * @param array<string,string> $blocks
+     */
+    protected function renderTwigComponent(string $name, array $data = [], ?string $content = null, array $blocks = []): RenderedComponent
+    {
+        if (!$this instanceof KernelTestCase) {
+            throw new \LogicException(sprintf('The "%s" trait can only be used on "%s" classes.', __TRAIT__, KernelTestCase::class));
+        }
+
+        $blocks = array_filter(array_merge($blocks, ['content' => $content]));
+
+        if (!$blocks) {
+            return new RenderedComponent(self::getContainer()->get('twig')
+                ->createTemplate('{{ component(name, data) }}')
+                ->render([
+                    'name' => $name,
+                    'data' => $data,
+                ])
+            );
+        }
+
+        $template = sprintf('{%% component "%s" with data %%}', addslashes($name));
+
+        foreach (array_keys($blocks) as $blockName) {
+            $template .= sprintf('{%% block %1$s %%}{{ blocks.%1$s|raw }}{%% endblock %%}', $blockName);
+        }
+
+        $template .= '{% endcomponent %}';
+
+        return new RenderedComponent(self::getContainer()->get('twig')
+            ->createTemplate($template)
+            ->render([
+                'data' => $data,
+                'blocks' => $blocks,
+            ])
+        );
+    }
+}
--- a/src/TwigComponent/src/Test/RenderedComponent.php
+++ b/src/TwigComponent/src/Test/RenderedComponent.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Test;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class RenderedComponent implements \Stringable
+{
+    /**
+     * @internal
+     */
+    public function __construct(private string $html)
+    {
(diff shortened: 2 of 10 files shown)


src/TwigComponent/src/Test/RenderedComponent.php
```diff
@@ -0,0 +1,27 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Test;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class RenderedComponent implements \Stringable
```
kbond, line 17: I added this stringable object to wrap the rendered html so we can possibly add something like described in https://github.com/symfony/ux/issues/818#issuecomment-1527750014 in the future.

I wasn't sure if changing the type-hint from string -> stringable object as the return type of `InteractsWithTwigComponent::renderTwigComponent()` would be considered a BC break. If not, I can change to just return `string`.

src/TwigComponent/src/Test/InteractsWithTwigComponents.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+namespace Symfony\UX\TwigComponent\Test;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+trait InteractsWithTwigComponents
```
kbond, line 19: This is my preferred name (:wink:) but I have no problem adding the Trait suffix or naming something else that jives with Symfony standards.

kbond:
My plan is to eventually add an `InteractsWithLiveComponents` trait (that'll probably use/extend this one) that adds _live-specific_ testing helpers.

kbond:
I've added the ability to use the component class name with these helpers (ie `$this->renderTwigComponent(MyComponent::class)`). Not sure if we should though as we aren't testing that the name (which will be used in twig) renders the expected component. For instance, if you use your IDE to rename a component, the tests will still pass but the template might not (if not manually configuring a component name).

src/TwigComponent/src/Test/InteractsWithTwigComponents.php
WebMamba, line 19: I am just wondering why do we use trait here? In my use case, I gonna create a dedicated test class for my component, and I not gonna mixed my components tests with other tests

WebMamba (commented):
Thanks @kbond that's super handy! 😁

src/TwigComponent/src/Test/InteractsWithTwigComponents.php
kbond: This is just a pattern I've found useful with my other testing libraries. I like using traits to avoid forcing a single _type_ of test.

```php
class MyComponentTest extends KernelTestCase
{
    use InteractsWithTwigComponents, ResetDatabase, Factories;

    // ...
}
```

I get the above would still work if we made this trait into and abstract `TwigComponentTestCase`. I'm just trying to be consistent: _Make a standard kernel test, and inject the test features you need via traits._

WebMamba:
@kbond friendly ping here! What is needed in order to merge it? I feel like it's super close. I already start to use this trait on my project and it's super handy! 😁

src/TwigComponent/src/Test/InteractsWithTwigComponents.php
WebMamba: Ho yes using this trait with foundry just rocks! Thanks for all of this 😁

kbond:
I feel this is basically ready. I'll rebase this week to ensure all's still good. I'll add a test for an embedded component using html syntax too.

I guess I was looking for opinions on the following:

- https://github.com/symfony/ux/pull/821#issuecomment-1528082967
- https://github.com/symfony/ux/pull/821#discussion_r1180758774 (would be cool to add a full HTML selector API in the future I think)

/cc @weaverryan

src/TwigComponent/src/Test/RenderedComponent.php
weaverryan: Let's start with this - easier to remove this later if we end up not needing it or it's a pain for some reason than to go in the other direction.

weaverryan:
> I've added the ability to use the component class name with these helpers (ie $this->renderTwigComponent(MyComponent::class))

I like this. I see what you're saying... but we can't do any better. To me, it feels like you've added a way to unit test a class... but sure, ultimately, someone could call it with invalid arguments or forget to call it at all on page. A higher-level test would be needed for that.

src/TwigComponent/src/Test/InteractsWithTwigComponents.php
```diff
@@ -0,0 +1,67 @@
...
+        $blocks = array_filter(array_merge($blocks, ['content' => $content]));
+
+        if (!$blocks) {
+            return new RenderedComponent(self::getContainer()->get('twig')
+                ->createTemplate('{{ component(name, data) }}')
+                ->render([
+                    'name' => $name,
+                    'data' => $data,
+                ])
+            );
```
weaverryan, line 48: this is pretty nuts 👍

weaverryan (approved):

weaverryan:
Thanks Kevin - this looks awesome!

PR #822, opened 2023-04-29
[Translator] Type and documentation fixes

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #... 
| License       | MIT

When working on the demo page, I've faced an issue about `Date` which was unuasable as translation parameter, but it should as we can use it to format date/datetime.

The documentation rendering was a bit broken too, I'm not familiar with RST, but it should be fixed.

Also, when running `yarn install`, the `yarn.lock` change, so I commited it aswell.


kbond (approved):

weaverryan:
Thanks @Kocal!

PR #823, opened 2023-04-29
[Live] add test helper

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Continuation of #821 
| License       | MIT

```php
use Symfony\UX\LiveComponent\Test\InteractsWithLiveComponents;

class MyComponentTest extends KernelTestCase
{
    use InteractsWithLiveComponents;

    public function testThisComponent(): void
    {
        $testComponent = $this->createLiveComponent('my_component');
        $testComponent = $this->createLiveComponent(Component::class); // or use the FQCN
        
        (string) $testComponent->render(); // string - initial state html
        $testComponent->component(); // object - component instance at initial state
        
        $testComponent
            ->call('increase') // call a live action
            ->call('increase', ['amount' => 2]) // call a live action with arguments
            ->set('count', 5) // set a live prop
            ->emit('inceaseEvent') // emit a live event
            ->emit('inceaseEvent', ['amount' => 2]) // emit a live event with arguments
            ->refresh() // simply refresh the component
        ;
        
        (string) $testComponent->render(); // string - updated state html
        $testComponent->component(); // object - component instance at updated state

        /** @var Symfony\Component\HttpFoundation\Response $response */
        $response = $testComponent->call('actionThatRedirects')->response();
    }
}
```

TODO:
- [x] handle action responses
- [x] `->refresh()` method?
- [x] Docs
- [x] Changelog

### DIFF
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -1,0 +1,149 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Test;
+
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\UX\LiveComponent\LiveComponentHydrator;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\TwigComponent\ComponentFactory;
+use Symfony\UX\TwigComponent\ComponentMetadata;
+use Symfony\UX\TwigComponent\MountedComponent;
+use Symfony\UX\TwigComponent\Test\RenderedComponent;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
+final class TestLiveComponent
+{
+    /**
+     * @internal
+     */
+    public function __construct(
+        private ComponentMetadata $metadata,
+        array $data,
+        private ComponentFactory $factory,
+        private KernelBrowser $client,
+        private LiveComponentHydrator $hydrator,
+        private LiveComponentMetadataFactory $metadataFactory,
+        private UrlGeneratorInterface $router,
+    ) {
+        $this->client->catchExceptions(false);
+
+        $mounted = $this->factory->create($this->metadata->getName(), $data);
+        $props = $this->hydrator->dehydrate(
+            $mounted->getComponent(),
+            $mounted->getAttributes(),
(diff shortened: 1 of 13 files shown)


kbond:
Again, with the system I describe in https://github.com/symfony/ux/issues/818#issuecomment-1527750014, we could make a really nice chainable test experience:

```php
$this->createLiveComponent('my_component')
    ->assertSeeIn('div.count', 1)
    ->call('increase')
    ->assertSeeIn('div.count', 2)
    ->call('increase', ['amount' => 3])
    ->assertSeeIn('div.count', 5)
;
```

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        return $this->request(['args' => $arguments], $action);
+    }
+
+    /**
+     * @param array<string,mixed> $arguments
+     */
+    public function emit(string $event, array $arguments = []): self
+    {
+        return $this->call($event, $arguments);
```
kbond, line 90: @weaverryan, does any consideration need to be made for emitting events?

weaverryan: Hmm, maybe? The point of emitting is to communicate between components. So if ComponentA emits a `Foo` event, then ComponentB might be listening to it.  But are we supporting such multiple-component communication? We would need to keep track of all of the components that are "rendered" currently in order to know if any are listening to `Foo`. Worse, if the component does `emitUp()`, we would need to understand parent-child relationships.

kbond: oo boy, that would be complex.

weaverryan: Right - that's what I was thinking :). From some quick looking at Livewire, I don't think they do this either. I found a forum somewhere where a user was asserting that one component emitted an event. Then, right after, they were calling the action on a second component that listens to that event. In other words, they were "triggering" the emit on the 2nd component manually - not expecting the system to magically do it.

WebMamba:
Hey @kbond, I am using this PR a lot to test my component and it's working great! Do you need a hand in order to finish it?

kbond:
> I am using this PR a lot to test my component and it's working great! Do you need a hand in order to finish it?

I was really just waiting for this feedback! Think it's current state is enough to start?

WebMamba:
> Think it's current state is enough to start?

I guess, I mean I didn't have a complex need yet but for my current need, it's working perfectly fine.

kbond:
Ok, I'll add `refresh()`/`response()` methods + docs and I think we should be good to go here.

kbond:
Ok, I believe this to be ready!

I also added 586a572b463cfb6d085829c33501592289512a5b but can move to another PR if desired.

src/LiveComponent/src/Test/TestLiveComponent.php
kbond: Feels like the best/easiest way to handle this.

weaverryan:
Thank you Kevin!

PR #824, opened 2023-04-30
[TwigComponent] Html syntax bug test case + fix

### LINKED ISSUE #805: [TwigComponent] Nested component doesn't work with html syntax

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #805
| License       | MIT

Not a fix, just a test case for the bug described in #805. It's *probably* easy to fix.

This fails to add the correct `content` block inside of `twig:foo`:

```
<twig:foo>
    <twig:bar>bar content</twig:bar>
</twig:foo>
```

But if you add *any* non-whitespace character before `<twig:bar>` - e.g. `foo <twig:bar>` - then it WILL add the block correctly. So, just some small thing not looking for an embedded component as valid "content" that should open that content block.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -30,7 +30,10 @@ class TwigPreLexer
         $this->line = $startingLine;
     }
 
-    public function preLexComponents(string $input): string
+    /**
+     * @param bool $insideOfBlock Are we sub-parsing the content inside a block?
+     */
+    public function preLexComponents(string $input, bool $insideOfBlock = false): string
     {
         $this->input = $input;
         $this->length = \strlen($input);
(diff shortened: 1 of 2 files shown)


weaverryan:
Now this also has the fix :)

PR #836, opened 2023-05-03
[Twig] Working around bug where the new lexer causes the extension set to load too early

### LINKED ISSUE #835: [TwigComponent] Twig ExtensionSet is initialized to soon since ux-twig-component 2.8.0

6.2.10

### Description

The problem occurred in combination with https://github.com/php-translation/symfony-bundle and the translation:extract command.
I'm getting this exception:

>  Unable to register extension "Translation\Extractor\FileExtractor\TwigFileExtractor" as extensions have already been initialized.

In `Symfony\UX\TwigComponent\DependencyInjection\TwigComponentExtension` the service `twig.configurator.environment` gets decorated by `Symfony\UX\TwigComponent\Twig\TwigEnvironmentConfigurator` and there the `Symfony\UX\TwigComponent\Twig\ComponentLexer` is initialized which will lead to the initialization of `Twig\ExtensionSet`:

Symfony\UX\TwigComponent\Twig\TwigEnvironmentConfigurator (31):
`$environment->setLexer(new ComponentLexer($environment));`

Twig\Lexer (457):
`array_keys($this->env->getUnaryOperators()),`

Twig\ExtensionSet (396):
```
    public function getUnaryOperators(): array
    {
        if (!$this->initialized) {
            $this->initExtensions();
        }

        return $this->unaryOperators;
    }
```

### How to reproduce

1. Add symfony/ux-twig-component and php-translation/symfony-bundle as dependency
2. Run `php bin/console translation:extract`

### Possible Solution

_No response_

### Additional Context

_No response_

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #835
| License       | MIT

Well, I don't love this fix :/. Normally, Twig's `Lexer` class is not instantiated inside `Environment` until it's needed. However, if you inject your own `Lexer`, then you must instantiate it early... AND it needs to extend `Lexer`, whose constructor cases the extension set to initialize. This is an ugly fix - but it works. 

Should this be considered a Twig bug? - e.g. perhaps all the work done inside the Lexer's constructor should be delayed until it's needed - https://github.com/twigphp/Twig/blob/f5ee1b6815a28fbab6e6fbb8c48b7964f0d93dd4/src/Lexer.php#L51-L152

Either way, it seems harmless to work around it here for the moment.

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentLexer.php
+++ b/src/TwigComponent/src/Twig/ComponentLexer.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\TwigComponent\Twig;
 
+use Twig\Environment;
 use Twig\Lexer;
 use Twig\Source;
 use Twig\TokenStream;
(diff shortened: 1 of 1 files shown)


WebMamba:
Hou yes super odd one! 😳 I think you referenced the wrong issue in your comment description - #835 instead of #83

weaverryan:
This is now fixed in Twig! 2.15.5 contains it. It looks like 3.6.0 WILL fix it also, but it's not released yet. You can bring in the 3.x-dev branch in your project until then to work around the issue.

Cheers!

fabpot:
Twig 3.6.0 released now.

PR #837, opened 2023-05-03
[TwigComponent] Fix lexer to escape attribute name when it contains dashes

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #... 
| License       | MIT

Hello :) back from holidays and now testing new features, so this is the second thing I've found.
Will open an issue about the first one.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -200,7 +200,7 @@ class TwigPreLexer
                 $attributeValue = $this->consumeAttributeValue($quote);
             }
 
-            $attributes[] = sprintf('%s: %s', $key, $attributeValue);
+            $attributes[] = sprintf('%s: %s', preg_match('/[-:]/', $key) ? "'$key'" : $key, $attributeValue);
 
             $this->expectAndConsumeChar($quote);
             $this->consumeWhitespace();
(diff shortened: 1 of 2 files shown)


src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -200,7 +200,7 @@ private function consumeAttributes(string $componentName): string
                 $attributeValue = $this->consumeAttributeValue($quote);
             }
 
-            $attributes[] = sprintf('%s: %s', $key, $attributeValue);
+            $attributes[] = sprintf('%s: %s', str_contains($key, '-') ? "'$key'" : $key, $attributeValue);
```
WebMamba: Maybe we should escape all the attributes. We can have the same issue with names containing `:` for example

norkunas: Yeah i forgot about colons. Tried to escape always but then some of tests throws:
```
2) Symfony\UX\TwigComponent\Tests\Unit\TwigPreLexerTest::testPreLex with data set "component_with_embedded_component_inside_block" ('<twig:foo><twig:block name="f...g:foo>', '{% component 'foo' %}{% block...ent %}')
Twig\Error\SyntaxError: Expected block name at line 1.
```

weaverryan:
Very nice - thank you @norkunas! We may hit other edge cases and need to escape more. I'm not too worried - that test suite is getting more and more cases all the time.

PR #854, opened 2023-05-08
[TwigComponent] Fix MountedComponent::$extraMetadata type doc

mttsch:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | -
| License       | MIT

MountedComponent::getExtraMetadata() has `mixed` as return type and it returns elements from this array.


weaverryan:
You're right - thanks Matthias!

PR #856, opened 2023-05-08
[LiveComponent] Throw clear error when using union types for LiveProps

### LINKED ISSUE #853: [LiveComponent] Support UnionTypes for LiveProps
## Problem

```
Call to undefined method ReflectionUnionType::getName()
```

## Cause

https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php#L77 assuming to work with a `ReflectionNamedType`.

## Reproduce

Create a LiveComponent with
```
#[LiveProp] public int|string $someProperty;
```
Use it somewhere

## Expected behavior

Either:

1. Support using UnionTypes for LiveProps
2. Throw clear error that UnionTypes are not supported

---
I want to contribute, but am not sure if option 1 or 2 is desired. The current `$metadatas` array is keyed by the property name and is flat, i.e. does not allow multiple types for one property (nor does the `LivePropMetadata` object), so allowing union types (option 1) will require some changes.

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #853
| License       | MIT

As mentioned in https://github.com/symfony/ux/issues/853#issuecomment-1538949120, supporting union types for LiveProps will be quite tricky. Given that, at least a clear error should be thrown.

### DIFF
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -71,6 +71,9 @@ class LiveComponentMetadataFactory
             }
 
             $type = $property->getType();
+            if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
+                throw new \LogicException(sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
+            }
             $metadatas[$property->getName()] = new LivePropMetadata(
                 $property->getName(),
                 $attribute->newInstance(),


kbond (commented):
Does this problem exist for intersection types as well? (`\ReflectionIntersectionType`)

sneakyvv:
I considered that, but  I wrongly assumed that intersection types are only used for method argument types, to enforce an argument to implement two interfaces, and thus not for property types, which does not make much sense. However, I checked now, and it is allowed (although the use case might be less common imo).

I updated the code, since it _is_ possible for properties as well.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -71,6 +71,9 @@ public function createPropMetadatas(\ReflectionClass $class): array
             }
 
             $type = $property->getType();
+            if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
+                throw new \LogicException(sprintf('Union nor intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
```
weaverryan:
```suggestion
                throw new \LogicException(sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
```

Just to avoid the double-negative of `nor` and `not supported`.

weaverryan (commented):
One tiny wording tweak - but I think this is the correct approach. And we could always revisit later 👍

sneakyvv:
@weaverryan that's indeed a double negative 🙈 
Fixed it!

kbond (approved):

weaverryan:
Thanks @sneakyvv!

Issue #844, opened 2023-05-04
[TwigComponent] Impossible to declare higher level component content block


norkunas:
So I have a component A which defines it's content block. Then I have a higher level component B which also defines it and to forward to the A component, but Twig says: `The block 'content' has already been defined line 6.`

```twig
{# A.html.twig #}
<{{ element }} class="{{ classes }}"{{ attributes }}>
  {%- block content -%}{%- endblock -%}
</{{ element }}>
```
```twig
{# B.html.twig #}
<twig:A>
  {%- block content -%}{%- endblock -%}
</twig:A>
```

WebMamba:
Humm but it can be also that `content` is a reserved word since the new syntax. When you use the new syntax the content block is automatically added. 🤔

norkunas:
So I should use `block('content')` ?
Edit: no this way does not work (probably some loop goes on because got 504)

norkunas:
Works only this way:

```twig
{# A.html.twig #}
<{{ element }} class="{{ classes }}"{{ attributes }}>
  {%- block content -%}{%- endblock -%}
</{{ element }}>
```

```twig
{# B.html.twig #}
{% set passthrough %}{% block content %}{% endblock %}{% endset %}
<twig:A>
  {{ passthrough }}
</twig:A>
```

weaverryan:
I've created a more realistic example to help me not get confused :).

```twig
{# anywhere #}
<twig:DivComponent>
    Hello!
</twig:DivComponent>
```

```
{# DivComponent.html.twig #}
<twig:GenericElement element="div">
    {% block content %}{% endblock %}
</twig:GenericElement>
```

```twig
{# GenericElement.html.twig #}
<{{ element }} class="{{ classes }}"{{ attributes }}>
  {%- block content -%}{%- endblock -%}
</{{ element }}>
```

Is that accurate?

@WebMamba In my `DivComponent.html.twig`, are we taking into account the possibility that the user may define normal, Twig-style blocks - e.g. `{% block content %}{% endblock %}` - inside the HTML syntax? I think we're not. I think we're seeing `{% block content` and thinking its Twig "output" code, and so we're starting the block content. So we're pre-lexing like this:

```
Original
<twig:GenericElement element="div">
    {% block content %}{% endblock %}
</twig:GenericElement>

After pre-lexing:
<twig:GenericElement element="div">
    {% block content %}{% block content %}{% endblock %}{% endblock %}
</twig:GenericElement>
```

[...]

PR #859, opened 2023-05-10
[TwigComponent] Fixing bug where traditional blocks aren't handled correctly

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fixes part of https://github.com/symfony/ux/issues/844#issuecomment-1542637242
| License       | MIT

Using `{% block traditional_block %}` inside of a `<twig:Component` syntax should now work. However, even though I tried to make the TwigPreLexer a bit more intelligent than just a regex parser, it's reaching its limits of complexity. Even this fix will break down if the user adds extra whitespace - e.g. `{%       block traditional_block %}`. It's likely that `TwigPreLexer` will need to be converted to an actual Lexer -> token stream -> parser type of a system. On the bright side, that would make it easier to integrate into Twig core if we ever chose to do that (the parser wouldn't convert over directly, but the lexer & tokens in theory would).

Cheers!

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -52,8 +52,10 @@ class TwigPreLexer
                 }
             }
 
-            if ($this->consume('<twig:')) {
-                $componentName = $this->consumeComponentName();
+            $isTwigHtmlOpening = $this->consume('<twig:');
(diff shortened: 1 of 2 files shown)


weaverryan:
As I mentioned, this isn't the most robust code, but the test suite is passing. So let's merge and see if any edge-cases spring up.

PR #861, opened 2023-05-11
[LiveComponent] Added type mixed to $value property and added string $className to mach HydrationExtensionInterface

mwansinck:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| License       | MIT

Documentation example for writing the hydrate method was not compatible with the HydrationExtensionInterface. 
Corrected the example so the code can be copy-pasted without errors.


weaverryan:
Absolutely correct! Thank you Marco!

PR #866, opened 2023-05-12
[LiveComponent] Smarter form trait data extracting

### LINKED ISSUE #864: [LiveComponent] `ComponentWithFormTrait::extractValues` does not properly flattens data
Comment from the trait:

> This is used to pass the initial values into the live component's
    frontend, and it's meant to equal the raw POST data that would
    be sent if the form were submitted without modification.

But "would be sent if the form were submitted without modification" is not correct. When you render an empty form and select entity in autocomplete the value is an string ID which is converted to an entity in backend and not the entity itself. So when using form with DTO and ux autocomplete when property in DTO is predefined - aka entity is set on it the value is not flattened to the ID.

Now after upgrade to the latest dev in tests we started to get:

`Uncaught PHP Exception Twig\Error\RuntimeError: "Error rendering "InvoiceCreateForm" component: The LiveProp "formValues" on component "InvoiceCreateFormComponent" is an array, but it contains one or more keys that are not scalars: formValues.customer: Proxies\__CG__\App\Entity\User" at /var/www/html/templates/Invoice/create.html.twig`

Also based on the autocomplete structure it should be flattened to `formValues.customer[autocomplete]` but I see only `formValues.customer`

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #864
| License       | MIT

Try this out @norkunas :). The problem was actually a tricky thing that UX autocomplete does. Thanks for reporting this so quickly.

Cheers!


norkunas:
Nice,thank you. Will try it out on monday 🙂

norkunas:
Tried this, but now I get: `Error rendering "InvoiceCreateForm" component: Child "_token" does not exist.`

weaverryan:
Ok, give it another try. We have CSRF disabled in the tests - I still need to fix that - but I bet it'll work now

norkunas:
> Ok, give it another try. We have CSRF disabled in the tests - I still need to fix that - but I bet it'll work now

works :+1:

PR #884, opened 2023-05-18
[LiveComponent] resetForm() method to get a fresh form

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | None - but from conversation on Slack
| License       | MIT

Hi!

Makes it possible to submit a form to a `LiveAction` then "reset" the form so that you can allow the component to render again with a fresh form.

Cheers!


94noni:
This one is nice 👍🏻

PR #885, opened 2023-05-18
 [attributes.add()] the Add() need to have an example and clarificati…

mina20088:
…on of usage and how to used

the [attributes.add()] need to have an example and clarification of usage and how to used and why it should be used

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | no issue
| License       | MIT

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -601,6 +601,8 @@ Set an attribute's value to ``false`` to exclude the attribute:
 
 .. versionadded:: 2.7
 
+    #this part needs more calrification and an example of how to use it and why 
+
     The ``add()`` method was introduced in TwigComponents 2.7.
 
 To add a custom Stimulus controller to your root component element:


src/TwigComponent/doc/index.rst
```diff
@@ -601,6 +601,8 @@ Set an attribute's value to ``false`` to exclude the attribute:
 
 .. versionadded:: 2.7
 
+    #this part needs more calrification and an example of how to use it and why 
+
```
weaverryan, line 605: The purpose of the `add()` method is to be able to render a Stimulus controller on your root component element. We mention it below. Do you have a suggestion about how it could be more clear?

mina20088: just give more example like for beginners they would not know what are the Stimulus controller

mina20088:
you can jsut add this link to let beginners know about the  Stimulus javascript framwork

https://symfony.com/doc/current/frontend/encore/simple-example.html#stimulus-symfony-ux

weaverryan:
Thanks for the clarification - I've added it over on #887 

Cheers!

PR #910, opened 2023-05-30
[LiveComponent] Fix array valued checkboxes change event handling

### LINKED ISSUE #906: [LiveComponent] Checkbox check/uncheck  doesn't trigger component update
Based on the [following documentation](https://symfony.com/bundles/ux-live-component/current/index.html#checkboxes-select-elements-radios-arrays) I tried to build a simple list filter based on checkboxes:

## Code

FilterMachines.php

```php
<?php

#[AsLiveComponent('filter_machines')]
class FilterMachines
{
    use DefaultActionTrait;

    #[LiveProp(writable: true)]
    public array $types = ['HYDRAULIC', 'AUTOMATIC', 'REPRISE'];

    public function __construct(
    ) {
    }

    public function getMachines(): array
    {
        return array_filter([
            ['name' => 'Machine 1', 'type' => 'HYDRAULIC'],
            ['name' => 'Machine 2', 'type' => 'AUTOMATIC'],
            ['name' => 'Machine 3', 'type' => 'REPRISE'],
        ], fn($machine) => in_array($machine['type'], $this->types));
    }
}
```

filter_machines.html.twig
```twig
<div {{ attributes }}>
    <label for="type-hydraulic">hydraulic</label>
    <input type="checkbox" data-model="types" id="type-hydraulic" value="HYDRAULIC">
    <label for="type-automatic">automatic</label>
    <input type="checkbox" data-model="types" id="type-automatic" value="AUTOMATIC">
    <label for="type-reprise">reprise</label>
    <input type="checkbox" data-model="types" id="type-reprise" value="REPRISE">
    {% if computed.machines|length > 0 %}
        <ul>
            {% for machine in computed.machines %}
                <li>
                    <strong>{{ machine.name }}</strong>
                    <p>{{ machine.type }}</p>
                </li>
            {% endfor %}
        </ul>
    {% else %}
        <div>Ah! No machines found matching "{{ types }}"</div>
    {% endif %}
</div>
```

index.html.twig (rendered by a dead simple FooController)
```twig

{% block title %}Hello HomeController!{% endblock %}

{% block body %}
    {{ component('filter_machines') }}
{% endblock %}
```

To reproduce:

- Run the previous code
- Uncheck any checkbox
- See that nothing happens (no AJAX request, no component render)

## Versions:

Symfony 6.2
Webpack Encore Bundle 1.17
UX LiveComponent 2.9

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | Fix https://github.com/symfony/ux/issues/906
| License       | MIT

As described in the linked issue, the event `change` of checkboxes weren't fully triggered (it didn't trigger AJAX request). It was because of the always-equals values of the pre-checked value attribute of the element and the after-change-event value of the element.

Thanks @Jim56450 for the help!

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -311,18 +311,19 @@ function getElementAsTagText(element) {
         : element.outerHTML;
 }
 const getMultipleCheckboxValue = function (element, currentValues) {
+    const finalValues = [...currentValues];
     const value = inputValue(element);
     const index = currentValues.indexOf(value);
     if (element.checked) {
         if (index === -1) {
-            currentValues.push(value);
+            finalValues.push(value);
         }
-        return currentValues;
+        return finalValues;
     }
     if (index > -1) {
-        currentValues.splice(index, 1);
+        finalValues.splice(index, 1);
     }
-    return currentValues;
+    return finalValues;
 };
 const inputValue = function (element) {
     return element.dataset.value ? element.dataset.value : element.value;
--- a/src/LiveComponent/assets/src/dom_utils.ts
+++ b/src/LiveComponent/assets/src/dom_utils.ts
@@ -266,24 +266,25 @@ export function getElementAsTagText(element: HTMLElement): string {
 }
 
 const getMultipleCheckboxValue = function (element: HTMLInputElement, currentValues: Array<string>): Array<string> {
+    const finalValues = [...currentValues];
     const value = inputValue(element);
     const index = currentValues.indexOf(value);
 
     if (element.checked) {
         // Add value to an array if it's not in it already
         if (index === -1) {
-            currentValues.push(value);
+            finalValues.push(value);
         }
 
-        return currentValues;
+        return finalValues;
     }
 
     // Remove value from an array
     if (index > -1) {
-        currentValues.splice(index, 1);
+        finalValues.splice(index, 1);
     }
 
-    return currentValues;
+    return finalValues;
 };
 
 const inputValue = function (element: HTMLInputElement): string {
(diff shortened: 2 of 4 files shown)


src/LiveComponent/assets/src/Component/ValueStore.ts
```diff
@@ -50,6 +50,9 @@ export default class {
         }
 
         if (this.props[normalizedName] !== undefined) {
+            if (Array.isArray(this.props[normalizedName])) {
+                return [...this.props[normalizedName]];
+            }
```
weaverryan: Instead of this change, I'm thinking we should:

```diff
--- a/src/LiveComponent/assets/src/dom_utils.ts
+++ b/src/LiveComponent/assets/src/dom_utils.ts
@@ -266,24 +266,25 @@ export function getElementAsTagText(element: HTMLElement): string {
 }

 const getMultipleCheckboxValue = function (element: HTMLInputElement, currentValues: Array<string>): Array<string> {
+    const finalValues = [...currentValues]
     const value = inputValue(element);
     const index = currentValues.indexOf(value);

     if (element.checked) {
         // Add value to an array if it's not in it already
         if (index === -1) {
-            currentValues.push(value);
+            finalValues.push(value);
         }

-        return currentValues;
+        return finalValues;
     }

     // Remove value from an array
     if (index > -1) {
-        currentValues.splice(index, 1);
+        finalValues.splice(index, 1);
     }

-    return currentValues;
+    return finalValues;
 };

 const inputValue = function (element: HTMLInputElement): string {
```

Currently (right or wrong), `ValueStore` returns (in the case of arrays/objects) a reference to the *actual* data in the `ValueStore`. This changes that to return a clone. The above change is smaller/safer for the time being. It fixes this exact checkbox case where we were accidentally modifying the value store directly.

I think modifying the `ValueStore` to not return references might, indeed, be a good idea (it would avoid an accident like this). But safer to do it later.

weaverryan:
Thank you Mathieu!

PR #929, opened 2023-06-06
Added warning for list of embedded components

florentdestremau:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       |  
| License       | MIT

As discussed on slack with @weaverryan , I think the `key` trick for lists of embedded components is an easy footgun and thus should be pointed as soon as the docs introduce the subject. I'm open for better wording or placement.


94noni (approved):
👍🏻

weaverryan:
This makes it just a LITTLE bit more visible - let's try it! Thanks Florent!

florentdestremau:
I didn't want to make it too obvious, but there sure is margin for a bigger warning

PR #931, opened 2023-06-08
[LiveComponent] Add StimulusBundle dependency

stof:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/
| New feature?  | no
| Tickets       | 
| License       | MIT

As this bundle implements its feature using stimulus, it should enforce that the bundle is available.

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -29,6 +29,7 @@
         "php": ">=8.0",
         "symfony/property-access": "^5.4.5|^6.0",
         "symfony/serializer": "^5.4|^6.0",
+        "symfony/stimulus-bundle": "^2.9",
         "symfony/ux-twig-component": "^2.8"
     },
     "require-dev": {


weaverryan:
Hey Stof!

There is no actual reference or usage of the Stimulus classes inside of LiveComponents. I agree that you'll need StimulusBundle in practice (because its recipe will set up the app for you), but I can't see where it's strictly required. Can you point?

Thanks!

stof:
The whole JS logic is implemented through a Stimulus controller: https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/src/live_controller.ts

Without stimulus, the live component will not be live. This is exactly what has been reported by someone on Slack.

stof:
hmm, apparently, the min PHP version for StimulusBundle is higher than for other UX bundles. This does not help defining such dependencies.

stof:
Actually, many other UX packages are relying on Stimulus for their JS logic, so it might make sense for them to have this dependency as well.

weaverryan:
Hmm, I see. It’s weird to me to make a package like this rely on StimulusBundlewhen it doesn’t *really* need it (you could get stimulus running manually and it would work). In practice, that’s not actually doing to happen, so your PR would make life easier in practice.

smnandre:
Done in #1946

PR #936, opened 2023-06-11
[TwigComponent] Add variable to refer outer scope when rendering embedded components

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | 
| License       | MIT

This PR's is an alternative approach to the problem stated in #863.

Due to the changes proposed in #920, which deals with passing blocks down to nested embedded components, I realized that just being able to access the host (or parent) component isn't sufficient if blocks can be passed down multiple levels deep (as already mentioned in https://github.com/symfony/ux/pull/863#issuecomment-1575123112).

This PR adds a proxy object `Hierarchy` which allows this syntax:

`Component.php`
```php
final class Component
{
    public function scream(int $screamHowHard = 1): string
    {
        return 'sf-ux rules' . str_repeat('!', $screamHowHard);
    }
}
```

```twig
{# component.html.twig #}

<twig:Foo :someProp="this.something(1)">
    {{ someProp }}
    {{ outerScope.this.something(3) }}
</twig:Foo>
```

```twig
{# Foo.html.twig #}

<div>{% block content %}{% endblock %}</div>
```

Resulting in 

```twig
<div>
    sf-ux rules!
    sf-ux rules!!!
</div>
```

Adding another layer below Foo, would ask for something like

```twig
{# component.html.twig #}

<twig:Foo>
    {{ outerScope.outerScope.this.something(3) }}
</twig:Foo>
```

to still point to Foo's function.

---

### Questions

There's also a `ComputedPropertiesProxy` object. Should we have a cached version of the `Hierarchy` proxy?

--- 
 
### NOTE

This PR continues on the changes of #920, because having a hierarchy only makes sense if blocks can be passed down multiple times. So #920 may have to be merged first, but if needed, and the `outerScope.this` syntax is preferred over the syntax of #863, I could push my branch that starts of the main branch instead.

Edited: using outerScope variable now instead Hierarchy proxy

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -951,6 +951,46 @@ When overriding the ``alert_message`` block, you have access to the ``message`` 
         {% endblock %}
     {% endcomponent %}
 
+
+.. versionadded:: 2.12
+
+    The ability to refer to the scope of higher components via the ``outerScope`` variable was added in 2.12.
+
+As mentioned before, variables from lower components are merged with those from upper components. When you need
+access to some properties or functions from higher components, that can be done via the ``outerScope...`` variable:
+
+.. code-block:: twig
+
+    {# templates/SuccessAlert.html.twig #}
+    {% set name = 'Fabien' %}
+    {% set message = 'Hello' %}
+    {% component Alert with { type: 'success', name: 'Bart' } %}
+        Hello {{ name }} {# Hello Bart #}
+
+        {{ message }} {{ outerScope.name }} {# Hello Fabien #}
+
+        {{ outerScope.this.someFunction }} {# this refers to SuccessAlert #}
+
+        {{ outerScope.this.someProp }} {# references a "someProp" prop from SuccessAlert #}
+    {% endcomponent %}
+
+You can keep referring to components higher up as well. Just add another ``outerScope``.
+Remember though that the ``outerScope`` reference only starts once you're INSIDE the (embedded) component.
+
+.. code-block:: twig
+
+    {# templates/FancyProfileCard.html.twig #}
+    {% component Card %}
+        {% block header %}
+            {% component Alert with { message: outerScope.this.someProp } %} {# not yet INSIDE the Alert template #}
+                {% block content %}
+                    {{ message }} {# same value as below, indirectly refers to FancyProfileCard::someProp #}
+                    {{ outerScope.outerScope.this.someProp }} {# directly refers to FancyProfileCard::someProp #}
+                {% endblock %}
+            {% endcomponent %}
+        {% endblock %}
+    {% endcomponent %}
+
 Component HTML Syntax
 ---------------------
 
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -121,6 +121,9 @@ final class ComponentRenderer implements ComponentRendererInterface
             // first so values can be overridden
             $context,
 
+            // keep reference to old context
+            ['outerScope' => $context],
+
             // add the component as "this"
             ['this' => $component],
 
--- a/src/TwigComponent/tests/Fixtures/Component/DivComponentWrapper3.php
+++ b/src/TwigComponent/tests/Fixtures/Component/DivComponentWrapper3.php
@@ -17,6 +17,12 @@ use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
  * @author Bart Vanderstukken <bart.vanderstukken@gmail.com>
  */
 #[AsTwigComponent]
-final class DivComponentWrapper
+final class DivComponentWrapper3
 {
+    public string $divComponentWrapperName = 'bar';
+
+    public function someFunction(): string
+    {
+        return 'calling DivComponentWrapper';
+    }
 }
--- a/src/TwigComponent/tests/Fixtures/templates/components/DivComponent5.html.twig
+++ b/src/TwigComponent/tests/Fixtures/templates/components/DivComponent5.html.twig
@@ -2,6 +2,7 @@
     I can access my own properties: {{ divComponentName }}.
     I can access the id of the Generic Element: {{ id }}.
     This refers to the Generic Element: {{ this.someFunction }}.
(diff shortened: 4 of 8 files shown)


Kocal:
IMO a child component should not have access to its parent.

weaverryan:
> IMO a child component should not have access to its parent.

Well... the tricky thing is that you ARE still kind of "inside" the parent template - i.e. you're inside `component.html.twig` and want to access the `Component` object... but you currently can't if you're coding inside the `<twig:Foo>` section of that template.

I'm not on one side or the other yet.

sneakyvv:
Since #920 is merged now, I rebased this branch on the current HEAD.

src/TwigComponent/doc/index.rst
```diff
@@ -951,6 +951,42 @@ When overriding the ``alert_message`` block, you have access to the ``message``
...
+Sometimes, you may want to override a block of a deeply nested component and you need access to
+some properties or functions from higher components. That can be done via the ``this.parent...`` hierarchy:
+
+.. code-block:: twig
+
+    {# templates/SuccessAlert.html.twig #}
+    {% component Alert with { type: 'success' } %}
+        {{ this.parent.someFunction }} {# this refers to SuccessAlert #}
+
+        {{ this.parent.someProp }} {# references a "someProp" prop from SuccessAlert #}
```
weaverryan: Instead of `this.parent` (which requires us to use a wrapper `Hierarchy` object and magic methods), what about following our `outerBlocks` convention and adding an `outerContext` variable?

```
{{ outerContext.this.someProp }} 
```

sneakyvv: How would you refer this.parent.parent then?

weaverryan: `outerContext.outerContext.this.someProp`? I think this would naturally occur if we started adding the `outerContext` variable, right?

sneakyvv: Changed and using `outerScope` variable instead of `outerContext` as discussed via Slack.

src/TwigComponent/doc/index.rst
```diff
@@ -951,6 +951,48 @@ When overriding the ``alert_message`` block, you have access to the ``message``
...
+As mentioned before, variables from lower components are merged with those from upper components. When you need
+access to some properties or functions from higher components, that can be done via the ``outerScope...`` variable:
+
+.. code-block:: twig
+
+    {# templates/SuccessAlert.html.twig #}
+    {% set name = 'Fabien' %}
+    {% set message = 'Hello' %}
+    {% component Alert with { type: 'success', name: 'Bart' } %}
+        Hello {{ name }} {# name = Bart #}
```
weaverryan: It might be slightly more clear to just have:

```suggestion
        Hello {{ name }} {# Hello Bart #}
```

Especially in the next example, where `name = Fabien` isn't actually quite right. But showing the printed result, I think, is clear.

src/TwigComponent/doc/index.rst
```diff
@@ -951,6 +951,48 @@ When overriding the ``alert_message`` block, you have access to the ``message``
...
+
+        {{ message }} {{ outerScope.name }} {# message = 'Hello', name = Fabien #}
```
```suggestion
        {{ message }} {{ outerScope.name }} {# Hello Fabien #}
```

src/TwigComponent/doc/index.rst
```diff
@@ -951,6 +951,48 @@ When overriding the ``alert_message`` block, you have access to the ``message``
...
+
+        {{ outerScope.this.someFunction }} {# this refers to SuccessAlert #}
```
weaverryan, line 972: But this and the next are good 👍

src/TwigComponent/doc/index.rst
```diff
@@ -951,6 +951,48 @@ When overriding the ``alert_message`` block, you have access to the ``message``
...
+
+        {{ outerScope.this.someProp }} {# references a "someProp" prop from SuccessAlert #}
+    {% endcomponent %}
+
+To keep the generic Alert component unaware of the ``someProp`` prop it is better to use ``outerScope.this``
+instead of passing that info along via Alert's attributes.
```
weaverryan: I'd delete this. It's true - but I think people might not understand what you mean, and will second guess themselves. Let's focus on showing them the tools they have.

src/TwigComponent/doc/index.rst
```diff
@@ -951,6 +951,48 @@ When overriding the ``alert_message`` block, you have access to the ``message``
...
+                    {{ message }} {# same value as below, indirectly refers to FancyProfileCard::someProp #}
+                    {{ outerScope.outerScope.this.someProp }} {# directly refers to FancyProfileCard::someProp #}
+                {% endblock %}
+            {% endcomponent %}
+        {% endblock %}
+    {% endcomponent %}
+
+.. versionadded:: 2.12
+
+    The ability to refer to the scope of higher components via the ``outerScope`` variable was added in 2.12.
```
weaverryan: Move this to the top of the new docs

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -121,6 +121,9 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
             // first so values can be overridden
             $context,
 
+            // keep reference to old context
+            ['outerScope' => $context],
```
weaverryan, line 125: Boy, that's quite simple!

weaverryan (commented):
This is looking good - big set of tests 👍

src/TwigComponent/doc/index.rst
sneakyvv: changed

src/TwigComponent/doc/index.rst
sneakyvv: changed

src/TwigComponent/doc/index.rst
sneakyvv: done

src/TwigComponent/doc/index.rst
sneakyvv: removed

weaverryan:
Thanks Bart!

PR #942, opened 2023-06-14
Stimulus controllers: allow to define outlets

jmsche:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | N/A
| License       | MIT

This PR allows to define outlets to Stimulus controllers. See https://stimulus.hotwired.dev/reference/outlets

Also updated .gitignore as running PHPUnit locally generated a `.phpunit.result.cache` file.

### DIFF
--- a/src/StimulusBundle/doc/index.rst
+++ b/src/StimulusBundle/doc/index.rst
@@ -162,6 +162,7 @@ stimulus_controller
 
 This bundle ships with a special ``stimulus_controller()`` Twig function
 that can be used to render `Stimulus Controllers & Values`_ and `CSS Classes`_.
+Stimulus Controllers can also reference other controllers by using `Outlets`_.
 
 For example:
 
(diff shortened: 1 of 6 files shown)


jmsche:
I've just seen that [the PR on Webpack Encore bundle](https://github.com/symfony/webpack-encore-bundle/pull/206) adds an `addOutlet()` function instead, but I feel it's more meaningful and less error prone to add it to the controller directly.

Also, should I PR against Encore bundle (v1) if this PR gets merged?

src/StimulusBundle/doc/index.rst
```diff
@@ -203,6 +204,30 @@ If you want to set CSS classes:
...
 
+And with outlets:
+
+.. code-block:: html+twig
+
+    <div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }, { 'loading': 'spinner' }, { 'other': '.target' ) }}>
+        Hello
+    </div>
+
+    <!-- would render -->
```
94noni, line 215: -would+will ?

94noni (approved):

src/StimulusBundle/doc/index.rst
jmsche: I just followed the other examples (that I didn't write); should I change them all?

94noni: I dont know :/
I was just reviewing unitary lol

weaverryan:
Thanks @jmsche! This looks great

jmsche:
@weaverryan Should I also PR against Encore bundle (v1)?

weaverryan:
@jmsche I don't think so. My thinking is that upgrading from EncoreBundle 1 to 2 is trivial. And 1 should be considered in maintenance - bug fixes only.

jmsche:
I guess symfony/webpack-encore-bundle#205 & symfony/webpack-encore-bundle#206 can be closed then?

PR #944, opened 2023-06-14
LiveComponents: nested dependent fields example

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | instead -->
| License       | MIT

This PR update the LiveComponents dependant form fields example to include a 3rd dependent field which is dependent on the 2nd.

Here, we need to use `FormFactoryInterface::createNamedBuilder()` to get access to `addEventListener()`,
as that method is not available on `FormInterface`.

Some refactoring was done to introduce enums to make this a bit more managable.

pizza-size


weaverryan:
Thanks Ben!

PR #945, opened 2023-06-15
fix(translator): generate unique constants name (fix #938)

### LINKED ISSUE #938: [Translator] TranslationsDumper generates duplicate keys in JS translation files
When I generate JS translation files, I cannot use it because I have duplicate keys. 

For instance, if I have 2 different translations ids like `Error` and `Error.` in my translation files, it generates the same key in JS file as `ERROR` so I have it twice in my translation file and I cannot build my JS files

Capture d’écran 2023-06-13 à 16 09 55

The problem comes from the `TranslationsDumper` class which using `translation id` instead of `constant name` in `getTranslations` method

The `id` should be the final `constant name` to exclude duplicate keys instead of the translation id like below 
```
$constantName = s($id)->ascii()->snake()->upper()->toString();

$translations[$constantName] ??= [];
$translations[$constantName][$realDomain] ??= [];
$translations[$constantName][$realDomain][$locale] = $message;
```

Hope this is clear for everyone

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #938 
| License       | MIT

Generating constants name with a automatic-prefix, in order to prevent conflicts when they are dumped in JS.

### DIFF
--- a/src/Translator/src/TranslationsDumper.php
+++ b/src/Translator/src/TranslationsDumper.php
@@ -53,7 +53,7 @@ class TranslationsDumper
         $translationsTs = "import { Message, NoParametersType } from '@symfony/ux-translator';\n\n";
 
         foreach ($this->getTranslations(...$catalogues) as $translationId => $translationsByDomainAndLocale) {
-            $constantName = s($translationId)->ascii()->snake()->upper()->toString();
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks Hugo!

PR #946, opened 2023-06-15
bump to php >=8.1 everywhere

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  |no
| Tickets       | 
| License       |

The versions of php were a bit disparate. they should be the same everywhere.

workflows for for 7.2 and 8.0 removed.

testing for php 8.2 should probably be added

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,7 +11,7 @@ jobs:
             - uses: actions/checkout@master
             - uses: shivammathur/setup-php@v2
               with:
-                  php-version: '8.0'
+                  php-version: '8.1'
             - uses: ramsey/composer-install@v2
             - name: php-cs-fixer
               run: ./vendor/bin/php-cs-fixer fix --dry-run --diff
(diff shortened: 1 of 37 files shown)


.github/workflows/test.yaml
```diff
@@ -67,140 +67,6 @@ jobs:
...
-              working-directory: src/LiveComponent
-              run: php vendor/bin/simple-phpunit
-
-            - name: Autocomplete Dependencies
-              uses: ramsey/composer-install@v2
-              with:
-                working-directory: src/Autocomplete
-            - name: Autocomplete Tests
-              working-directory: src/Autocomplete
-              run: php vendor/bin/simple-phpunit
```
weaverryan, line 202: This all should be moved to the 8.1-equivalent jobs. Each package should test on high deps and low deps.

And yes, this is annoyingly repetitive. We need to make the matrix dynamic. Getting everything on the same php version is a step towards making that easier.

bendavies: ah i see, only some components were tested on 7.2 and 8.0. ok

bendavies:
@weaverryan this is ready for review.

bendavies:
php-cs-fixer is failing because v3.18 was released, i think

src/Autocomplete/composer.json
```diff
@@ -24,19 +24,22 @@
...
     "require": {
-        "php": ">=8.0",
+        "php": ">=8.1",
         "symfony/dependency-injection": "^5.4|^6.0",
         "symfony/http-foundation": "^5.4|^6.0",
         "symfony/http-kernel": "^5.4|^6.0",
         "symfony/property-access": "^5.4|^6.0",
-        "symfony/string": "^5.4|^6.0"
+        "symfony/string": "^5.4|^6.0",
+        "twig/twig": "^2.14.7|^3.0.4"
```
weaverryan: Should be in require-dev

src/Autocomplete/tests/Fixtures/Kernel.php
```diff
@@ -66,6 +69,16 @@ public function registerBundles(): iterable
         yield new ZenstruckFoundryBundle();
     }
 
+    protected function build(ContainerBuilder $container): void
+    {
+        $container->addCompilerPass(new class() implements CompilerPassInterface {
+            public function process(ContainerBuilder $container): void
+            {
+                $container->removeDefinition('doctrine.orm.listeners.pdo_session_handler_schema_listener');
```
weaverryan, line 78: Can you add a comment about what this is working around?

src/Cropperjs/composer.json
```diff
@@ -28,13 +28,15 @@
...
+        "php": ">=8.1",
         "intervention/image": "^2.5",
         "symfony/config": "^5.4|^6.0",
         "symfony/dependency-injection": "^5.4|^6.0",
         "symfony/form": "^5.4|^6.0",
         "symfony/http-kernel": "^5.4|^6.0",
-        "symfony/validator": "^5.4|^6.0"
+        "symfony/options-resolver": "^5.4|^6.0",
+        "symfony/validator": "^5.4|^6.0",
+        "twig/twig": "^2.14.7|^3.0.4"
```
weaverryan: I'm ok with the first 2, but `twig/twig` should be in `require-dev`.

src/Dropzone/composer.json
```diff
@@ -28,11 +28,13 @@
...
     "require": {
-        "php": ">=7.2.5",
+        "php": ">=8.1",
         "symfony/config": "^5.4|^6.0",
         "symfony/dependency-injection": "^5.4|^6.0",
         "symfony/form": "^5.4|^6.0",
-        "symfony/http-kernel": "^5.4|^6.0"
+        "symfony/http-kernel": "^5.4|^6.0",
+        "symfony/options-resolver": "^5.4|^6.0",
+        "twig/twig": "^2.14.7|^3.0.4"
```
weaverryan: require-dev for the last one

src/LiveComponent/composer.json
```diff
@@ -26,27 +26,30 @@
...
         "symfony/dependency-injection": "^5.4|^6.0",
         "symfony/form": "^5.4|^6.0",
+        "symfony/options-resolver": "^5.4|^6.0",
         "symfony/framework-bundle": "^5.4|^6.0",
         "symfony/phpunit-bridge": "^6.0",
         "symfony/property-info": "^5.4|^6.0",
         "symfony/security-csrf": "^5.4|^6.0",
         "symfony/twig-bundle": "^5.4|^6.0",
         "symfony/validator": "^5.4|^6.0",
+        "twig/twig": "^2.14.7|^3.0.4",
```
weaverryan: I think this could actually be moved to `require`: there are a few places where we rely on `Twig` code directly. It's redundant in practice, as `ux-live-component` requires Twig, bug strictly speaking, I think this packages DOES directly need Twig.

src/LiveComponent/tests/Fixtures/Kernel.php
```diff
@@ -57,6 +61,16 @@ public function registerBundles(): iterable
...
     }
 
+    protected function build(ContainerBuilder $container): void
+    {
+        $container->addCompilerPass(new class() implements CompilerPassInterface {
+            public function process(ContainerBuilder $container): void
+            {
+                $container->removeDefinition('doctrine.orm.listeners.pdo_session_handler_schema_listener');
+            }
+        }, PassConfig::TYPE_BEFORE_OPTIMIZATION, 1);
```
weaverryan, line 72: And same comment here.

weaverryan (commented):
Just a few minor things - very good stuff 👏 

About php-cs-fixer, I bet you're correct. But the fix looks wrong to me. However, that code is not important - let's delete this commented-out methods entirely as a workaround.

src/Autocomplete/tests/Fixtures/Kernel.php
```diff
@@ -66,6 +69,17 @@ public function registerBundles(): iterable
         yield new ZenstruckFoundryBundle();
     }
 
+    protected function build(ContainerBuilder $container): void
+    {
+        // workaround https://github.com/symfony/symfony/igssues/50322
```
weaverryan: small typo here!

weaverryan:
Thank you Ben!

PR #951, opened 2023-06-16
[Turbo] Fixing a bug where saving a proxy would not trigger Broadcasts

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | None
| License       | MIT

If the object you're updating, for example, happens to be wrapped in a Proxy object, then `BroadcastListener` would fail to find the `#[Broadcast]` attribute and no broadcasting would happen.

Cheers!


chalasr (approved):
👍 modulo some coding style and test fixes eventually - not sure they are related

nicolas-grekas (approved):

PR #953, opened 2023-06-19
[StimulusBundle] Fix controller name conversion

jon-ht:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

Stimulus controllers must follow some naming convention/normalization according to https://stimulus.hotwired.dev/reference/controllers#identifiers

There was a case when controllers file with underscores in their name would not be a valid entry for Stimulus.

I wasn't sure about using `Symfony\UX\StimulusBundle\Dto\StimulusAttributes::normalizeControllerName` because it seems related to Twig (due to its `private Environment $env` in construct)


weaverryan:
Thank you @jon-ht - very thorough work!

PR #969, opened 2023-06-25
[TwigComponent] Fix escaping stimulus attributes

### LINKED ISSUE #968: Attributes added by `stimulus_controller`  should be escaped
When I write

```
<div {{ attributes.defaults(stimulus_controller('my-controller', { params: ['a', 'b'] })) }}>
```
then params' value should be escaped as html attributes.

Now it renders 

```
data-controller="my-controller live" data-my-controller-params-value="["a","b"]" ...
```
instead of
```
data-controller="my-controller live" data-my-controller-params-value="&#x5B;&quot;a&quot;,&quot;b&quot;&#x5D;" ...
```

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #968
| License       | MIT

I think we should do this as late as possible, e.g. in `__toString` or only in twig and maybe for all the attributes.
So this feels like a naive solution, but it works for now.

### DIFF
--- a/src/StimulusBundle/src/Dto/StimulusAttributes.php
+++ b/src/StimulusBundle/src/Dto/StimulusAttributes.php
@@ -179,6 +179,16 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
         return array_merge($attributes, $this->attributes);
     }
 
+    public function toEscapedArray(): array
+    {
+        $escaped = [];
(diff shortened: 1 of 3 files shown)


weaverryan:
I don't love this solution, but let's do it for now to fix things. Thanks @1ed!

PR #972, opened 2023-06-27
Update namespace for PackageRepository

nicodemuz:


norkunas:
Then the `PackageRepository` should be moved to `src/Repository` :)

weaverryan:
Yup, @norkunas is correct. I put this NOT in the `src/Repository/` originally just to not confuse it with "entity repositories", which this is not one.

PR #976, opened 2023-06-29
Allow removing options that `EntityType` can't accept

### LINKED ISSUE #420: [Autocomplete] Impossible to define custom options in a EntityAutocompleteType class
### Description

I followed the documentation of UX-Autocomplete and I created my custom form type. Problem is that I expect to be able to use some custom options. I can't. It looks like the `AutocompleteEntityTypeSubscriber` is just assuming you don't use any custom options and it's passing all the options to an "autocomplete" field. See https://github.com/symfony/ux-autocomplete/blob/2.x/src/Form/AutocompleteEntityTypeSubscriber.php#L44

### How to reproduce

Just follow the instructions provided in documentation https://symfony.com/bundles/ux-autocomplete/current/index.html#usage-in-a-form-with-ajax and add a custom option, like "allowed_states" in the example provided in https://symfony.com/doc/current/form/create_custom_field_type.html#adding-configuration-options-for-the-form-type.
As soon as you try to use the form with your custom form type (no matter if you use your custom option or not), you get the following exception:

>An error has occurred resolving the options of the form "Symfony\Bridge\Doctrine\Form\Type\EntityType": The option "allowed_states" does not exist. Defined options are: "action", "allow_extra_fields", "allow_file_upload", "allow_options_create", "attr", "attr_translation_parameters", "auto_initialize", "autocomplete", "autocomplete_url", "block_name", "block_prefix", "by_reference", "choice_attr", "choice_filter", "choice_label", "choice_loader", "choice_name", "choice_translation_domain", "choice_translation_parameters", "choice_value", "choices", "class", "compound", "constraints", "csrf_field_name", "csrf_message", "csrf_protection", "csrf_token_id", "csrf_token_manager", "data", "data_class", "disabled", "em", "empty_data", "error_bubbling", "error_mapping", "expanded", "extra_fields_message", "form_attr", "getter", "group_by", "help", "help_attr", "help_html", "help_translation_parameters", "id_reader", "inherit_data", "invalid_message", "invalid_message_parameters", "is_empty_callback", "label", "label_attr", "label_format", "label_html", "label_translation_parameters", "mapped", "method", "multiple", "no_more_results_text", "no_results_found_text", "options_as_html", "placeholder", "post_max_size_message", "preferred_choices", "priority", "property_path", "query_builder", "required", "row_attr", "setter", "tom_select_options", "translation_domain", "trim", "upload_max_size_message", "validation_groups".

### Possible Solution

I don't know if the current implementation could be fixed.
If not, the documentation should be updated, adding a big warning that the developer must not define any custom option in their custom form type.

janklan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #420
| License       | MIT

If you implement a custom Autocompleter class as [described in the docs](https://symfony.com/bundles/ux-autocomplete/current/index.html#usage-in-a-form-with-ajax) and happen to require some non-standard options for your custom field, the Autocompleter (currently) tries to pass your custom option to the internal `EntityType` field. 

This PR adds the ability to list the options that should NOT be passed to that `EntityType`.

### DIFF
--- a/src/Autocomplete/src/Form/ParentEntityAutocompleteType.php
+++ b/src/Autocomplete/src/Form/ParentEntityAutocompleteType.php
@@ -82,6 +82,8 @@ final class ParentEntityAutocompleteType extends AbstractType implements DataMap
             'security' => false,
             // set the max results number that a query on automatic endpoint return.
             'max_results' => 10,
+            /* @see AutocompleteEntityTypeSubscriber::withoutExcludedOptions() */
+            'autocomplete_excluded_options' => [],
         ]);
 
         $resolver->setRequired(['class']);
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## Next
+
+-   Allow not passing custom options to the nested `EntityType` by passing a list through another custom option
+    called `autocomplete_excluded_options` - #420
+
 ## 2.9.0
 
 -   Add support for symfony/asset-mapper
--- a/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
+++ b/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
@@ -43,9 +43,8 @@ final class AutocompleteEntityTypeSubscriber implements EventSubscriberInterface
         // pass to AutocompleteChoiceTypeExtension
         $options['autocomplete'] = true;
         $options['autocomplete_url'] = $this->autocompleteUrl;
-        unset($options['searchable_fields'], $options['security'], $options['filter_query']);
 
-        $form->add('autocomplete', EntityType::class, $options);
+        $form->add('autocomplete', EntityType::class, $this->withoutExcludedOptions($options));
     }
 
     public function preSubmit(FormEvent $event)
@@ -103,5 +102,30 @@ final class AutocompleteEntityTypeSubscriber implements EventSubscriberInterface
             FormEvents::PRE_SET_DATA => 'preSetData',
             FormEvents::PRE_SUBMIT => 'preSubmit',
         ];
+    }
+
+    /**
+     * If you create a custom Form Type based on {@link ParentEntityAutocompleteType} and make it accept any non-standard
+     * options, these options need to be removed from the $options array passed to EntityType because it doesn't
+     * accept those custom options.
+     *
+     * @see https://github.com/symfony/ux/issues/420
+     */
+    private function withoutExcludedOptions(array $options): array
+    {
+        $optionBlacklist = [
+            // Options added by ParentEntityAutocompleteType
+            'searchable_fields', 'security', 'filter_query', 'property',
+            // Any other options that were added to the custom AutocompleteType that EntityType doesn't understand
+            ...$options['autocomplete_excluded_options'],
+            // The list of blacklisted options itself (EntityType doesn't understand that either!)
+            'autocomplete_excluded_options',
+        ];
+
+        foreach ($optionBlacklist as $option) {
+            unset($options[$option]);
+        }
+
+        return $options;
     }
 }


janklan:
A side note: if I knew how to get a list of options of that `EntityType` field's instance, we wouldn't need to list them manually. It would be enough to just take the current `$options` and remove any unsupported values from it before passing it on. 

If there is a way to do that, please let me know, I'm happy to tweak things!

garak:
@janklan we should avoid words like "blacklist" and "whitelist", and prefer words that are more inclusive and convey real meanings. 
Here are some good alternatives:
* deny list
* avoid list
* block list 
* exclude list

janklan:
@garak I think we don't need to get into politics of offensive language before this gets any real indication of whether or not it is even viable. Thanks for the feedback though.

garak:
@janklan it's not politics, it's just trying to be nice.
Symfony is making an effort for it, see for example https://symfony.com/doc/current/contributing/diversity/index.html

Anyway, even if you don't care about it and you want to consider it as "politics", I provided you with a very practical reason: the word "blacklist" simply doesn't convey the meaning you want unless you leverage a stereotype.

janklan:
Using the word "blacklist" shouldn't be seen as "being not nice". The official definition in the Oxford Dictionary is: "a list of people or things that are regarded as unacceptable or untrustworthy and should be excluded or avoided.".

In the current context, it contains a list of options we wish to exclude from passing to the `FormEntity`, and as such, I think the name conveys the meaning ~perfectly~ reasonably well. 

With that said, let's please go back to the technical aspects of this PR.

garak:
Thanks for your understanding
Just a final word to point out the false dichotomy of nice and not nice. The dictionary entry you mentioned is neutral.

src/Autocomplete/src/Form/ParentEntityAutocompleteType.php
```diff
@@ -82,6 +82,8 @@ public function configureOptions(OptionsResolver $resolver)
             'security' => false,
             // set the max results number that a query on automatic endpoint return.
             'max_results' => 10,
+            /* @see AutocompleteEntityTypeSubscriber::removeBlacklistedOptions() */
```
jmsche:
```suggestion
            /* @see AutocompleteEntityTypeSubscriber::withoutExcludedOptions() */
```

jmsche (changes requested):

src/Autocomplete/src/Form/ParentEntityAutocompleteType.php
janklan: Yeah I'm pushing that change now. Thanks!

janklan:
@garak correct; neutral does not mean "not nice". I think neutral is great, especially in software. It's not a poem. It's code. It should be precise above all. `autocomplete_excluded_options` is more precise than `autocomplete_blacklist`, so that makes it better. Not the absence of the word `blacklist`.

Anyway, I have an opinion on these things that you would probably disagree with, and that's OK. Consider my opinion suppressed in favour of your opinion :)

garak:
The problem is that if you're neutral in an unbalanced situation, you're on the same side as the "not nice" ones.
And software is used by people, so it can't be neutral, not as long as you want to try to be nice.

janklan:
Let's agree to disagree and let's get back to focusing on the subject of this PR. Thanks.

weaverryan:
@janklan I like the idea behind this. It's not automatic, but it makes something work that doesn't currently work. If I'm understanding the use-case, it would be this:

A) I create an "entity autocomplete field" - like https://github.com/symfony/ux/blob/2.x/ux.symfony.com/src/Form/FoodAutocompleteField.php - but I add a custom option - e.g. `hideJunkFood`. Perhaps I use this to customize the `query_builder` option I'm using.

B) When using the field - like https://github.com/symfony/ux/blob/922a422899b45e972ff70119e245d9d7bf4610c3/ux.symfony.com/src/Form/TimeForAMealForm.php#L14 - I would pass that as a custom option:

```
->add('foods', FoodAutocompleteField::class, ['hideJunkFood' => true])
```

Am I thinking about this correctly? If so, then when the Ajax request is made to the autocomplete endpoint, won't we recreate the `FoodCompleteField` there but *without* this option? And so `hideJunkFood` would be true and we may return the wrong autocomplete options? If that is all true, then we should tackle the other half of this problem now: allowing some options (like `hideJunkFood`) to be passed as a query parameter on the autocomplete Ajax URL and passed back into the form type there.

Cheers!

norkunas:
I think this can be closed as there are no nested form type now, and you can provide everything same as with other form types

weaverryan:
Yup & yay! Let us know if you have any issues with the new system

PR #983, opened 2023-07-02
new QuillJs component

Ehyiah:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       |  N/A
| License       | MIT

This PR try to make a WYSIWYG into a new Symfony UX component, the famous QuillJs https://quilljs.com/

Usage is pretty simple, just add a QuillType as any regular FormType to get the most simple WYSIWYG.
Or use the configurations options to add other possible options.

It also comes with an EasyAdmin integration by creating a specific QuillAdminField.

Also Currently adding some features not availables out of the box by quill like custom image upload endpoint if you don't want images to be stored as base64 in your database.

Would you like to see this component be added into symfony UX ?


ahmedyakoubi:
Nice job man but I have other thoughts about QuillJs. quilljs is not well maintained and if we have to make it happen we need to wait until version 2 is released and see where things are going. Also, I'd prefer to go with something smaller (size) and more potent like [Lexical](https://github.com/facebook/lexical).

Ehyiah:
Thank you :-) 
Pretty happy you found it great, this is my first try contributing in symfony ecosystem.

And hat was one of my concern too when I choose to start this WYSIWYG component with QuillJs.
Version 2.0 seems to be still in developpement.

And Lexical seems pretty nice too.

I'm totally open to make it with another tool if the majority prefer it.

weaverryan:
Ah yea, I see that Quill tagged 2.0.0.dev0 5 years ago! 

I have not used Lexical, but it looks high quality, is backed by Facebook. On #1 (yes, issue 1!) some other options were discussed. 

Also, there is https://trix-editor.org/ - which is from the same people as Stimulus / Turbo. I would probably vote for trying Trix first, then Lexical if Trix is lacking in some way.

smnandre:
I spent a lot of time comparing various RTE last couple of years and.. _in my humble opinion_ the most updated/professional one is [TipTapDev](https://tiptap.dev/).

It allows more complex content structures like [editor.js](https://editorjs.io/) but even in its simplest form it's already a great / profesional editor with advanced capacities.

I'd love to see it integrated in Symfony UX

Ehyiah:
I'll take a few days to try some of these, at this point I keep in mind 4 alternatives to QuillJs :

**Lexical** which seems pretty complete,
**TipTap** seems pretty well (also very complete and configurable, i like this point).
**Trix** seems very easy to use and to implement (I think), but maybe lack of configurability at first glance).
**suneditor** looks very nice also (complete and configurable and very well documented)

This is going to be complicated, but at this moment, my preference goes to suneditor. 

Maybe we could start a polling ? What do you think ?

WebMamba:
Tiptap already has a Laravel integration: https://tiptap.dev/installation/php. And like what they did by wrapping everything in a component:
```html
 <x-editor
  wire:model="foo"
  wire:poll.10000ms="autosave"
></x-editor>
```

It can be cool to endup with something like this:
```
 <twig:editor 
   wire:model="foo"
  wire:poll.10000ms="autosave"
  ></x-editor>
```

The 4 alternatives look good, but Lexical and TipTap have stronger communities AND GOOD DOCUMENTATION. suneditor is a no-go for me for these two reasons.

weaverryan:
Both tiptap and lexical look great. Both have a ton of GH stars and are very active. Both have MIT licenses. As someone who hasn't used either, if I had to pick one, I'd pick tiptap. Its docs/site look better and it's initialization looks clean.

Ehyiah:
Well, looks like we have 2 finalists.
I'm going to have a deeper look into these 2, then start a new component from one of these :-)

Lets get back to work again !

Jesmaster:
With the release of Symfony 7 there is currently no supported wysiwg editor bundle for Symfony projects.

The FOSCKEditor bundle is still using CKEditor 4 which was EoL back in June 2023 and the bundle itself seems to have been abandoned shortly after Symfony 6 support was added https://github.com/FriendsOfSymfony/FOSCKEditorBundle/issues/254

I'm currently working on a project rebuilding a Symfony 3 site to the latest version and I've been using some of the new UX components and they are working great. I was looking to see if a new wysiwyg editor integration was part of the UX initiative and found this issue. My new site rebuild is stuck on Symfony 6 until I can remove the dependency on the FOSCKeditor package so I'm highly anticipating a new official UX wysiwyg component that will be supported for many versions to come!

simondaigre:
Quill 2.0 [is available](https://github.com/quilljs/quill/releases/tag/v2.0.0). Maybe we can reconsider this PR ?

Ehyiah:
I'm working on it for a few weeks/months already,
The bundle is already compatible with quill v2 (release candidate 5 atm) with some new things.
I will bump the quill to the 2.0 release in the next few days.

PR #985, opened 2023-07-02
[TwigComponent] Add documentation about passing blocks to embedded components

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | Fix #...
| License       | MIT

This adds documentation of the feature added in #920

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -832,6 +832,136 @@ The ``with`` data is what's mounted on the component object.
 
     Embedded components *cannot* currently be used with LiveComponents.
 
+Passing blocks to embedded components
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By combining nested components and embedded components very powerful structures can be created.
+There's one important thing to remember: **each embedded component, a.k.a. a component with
+blocks, inside a template, a.k.a. a nested component, starts its own template**.
+Any blocks that are available in your component template are *NOT* available *inside* the embedded component.
+
+Imagine this simple Alert component, which allows its content to be set by a higher component,
+for example a SuccessAlert component.
+
+.. code-block:: html+twig
+
+    {# templates/alert.html.twig #}
+    <div class="alert alert-{{ type }}">
+        {% block content %}{% endblock %}
+    </div>
+
+.. code-block:: twig
+
+    {# templates/successAlert.html.twig #}
+    {% component Alert with { type: 'success' } %}
+        I can override the content, but that's pretty dull.
+    {% endcomponent %}
+
+What if this SuccessAlert component wants to make the content of its nested Alert component configurable?
+Well, you can use the special `outerBlocks` variable for those cases. This makes it possible to refer
+to blocks that are passed in from another higher component.
+
+.. code-block:: twig
+
+    {# templates/successAlert.html.twig #}
+    {% component Alert with { type: 'success' } %}
+        {{ blocks(outerBlocks.content) }}
+    {% endcomponent %}
+
+.. code-block:: twig
+
+    {# templates/some_page.html.twig #}
+    {% component SuccessAlert %}
+        Look mom, I'm a complex component!
+    {% endcomponent %}
+
+Note that to pass a block multiple components down, each component would need to pass it along.
+
+.. code-block:: twig
+
+    {# templates/level1.html.twig #}
+    {% component Level2 %}
+        {% block foo %}
+            I'm providing/overriding the foo block of Level2
+        {% endblock %}
+    {% endcomponent %}
+
+.. code-block:: twig
+
+    {# templates/level2.html.twig #}
+    {% component Target %}
+        {% block foo %}
+            {# overriding the foo block of Target #}
+            {# but actually passing it along from the upper component #}
+            {# it could add some extra stuff here of course #}
+            {# and even append/prepend the content of Target's foo block by using {{ parent() }} #}
+
+            {{ blocks(outerBlocks.foo) }}
+        {% endblock %}
+    {% endcomponent %}
+
+.. code-block:: html+twig
+
+    {# templates/target.html.twig #}
+    <div>{% block foo %}default content for foo{% endblock %}</div>
+
+Context of embedded components
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Every embedded component can be viewed as another template inside a template, hence the name.
+Or in other words, in the other direction, it's as if the embedded component's blocks are copy-pasted
+into that component's template. This has a few interesting consequences.
+
+First of all, although `this` is referring to the component of the template into which an (embedded)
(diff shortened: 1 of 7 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
 
     Embedded components *cannot* currently be used with LiveComponents.
 
+Passing blocks to embedded components
```
weaverryan: Uppercase Blocks, Embedded and Components

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.10
+
+    The `outerBlocks` variable has been added since 2.10.
```
```suggestion
    The ``outerBlocks`` variable was added in 2.10.
```

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+
+.. code-block:: twig
+
+    {# templates/successAlert.html.twig #}
+    {% component Alert with { type: 'success' } %}
+        I can override the content, but that's pretty dull.
+    {% endcomponent %}
+
+What if this SuccessAlert component wants to make the content of its nested Alert component configurable?
+Well, you can use the special `outerBlocks` variable for those cases. This makes it possible to refer
```
weaverryan: In rst, unlike markdown, all technical words have 2 “ticks” around them, not just 1

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+    {% endcomponent %}
+
+.. code-block:: twig
+
+    {# templates/some_page.html.twig #}
+    {% component SuccessAlert %}
+        Look mom, I'm a complex component!
+    {% endcomponent %}
+
+Note that to pass a block multiple components down, each component would need to pass it along.
```
weaverryan: I’m not sure about this part - it seems obvious enough to me, and the code below REALLY makes this look complex. I vote for mentioning this perhaps, but not showing it.

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+            {{ blocks(outerBlocks.foo) }}
+        {% endblock %}
+    {% endcomponent %}
+
+.. code-block:: html+twig
+
+    {# templates/target.html.twig #}
+    <div>{% block foo %}default content for foo{% endblock %}</div>
+
+Context of embedded components
```
weaverryan: Uppercase everything except of

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Every embedded component can be viewed as another template inside a template, hence the name.
+Or in other words, in the other direction, it's as if the embedded component's blocks are copy-pasted
+into that component's template. This has a few interesting consequences.
```
weaverryan: into the target component’s template

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+
+First of all, although `this` is referring to the component of the template into which an (embedded)
```
weaverryan: First, although ``this`` refers to…

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+component is nested, *INSIDE* an embedded component it is referring to the *embedded* component,
+*NOT* the component into which it's being nested.
```
weaverryan: The wording on this is really tricky: nested vs embedded. Maybe say less and let the code do the talking.

src/TwigComponent/doc/index.rst
```diff
@@ -832,6 +832,140 @@ The ``with`` data is what's mounted on the component object.
...
+lower components. (That's also why `this` refers to the embedded, or "current" component)
+
+Finally, the most interesting thing is that, because the embedded components are literally
+resolved when rendering the nested component's template, you can even access the variables
+from THAT template as well. Again, it's as if the embedded component's blocks are being
+copy-pasted into the nested component's template.
+
+.. code-block:: twig
+
+    {# templates/successAlert.html.twig #}
```
weaverryan: SuccessAlert.html.twig - I think that’s in a few places.

weaverryan (commented):
Parts are a bit more technically worded than I’d like - and I dropped you an idea on slack, but this is very nice: you chose excellent examples to show off what’s possible. Thanks!

src/TwigComponent/doc/index.rst
sneakyvv: I guess you're right. Essentially, it was kinda explaining how blocks work, which is not the scope of this documentation.

weaverryan:
Hey! I just did a quick proof and edit of the docs so far - a team effort to explain these important concepts on a simple way. I've tried to remove the wording "embedded component". Even I mix up "embedded" and "nested" in my mind. It's not as succinct, but saying "when you pass a block to a component" instead of "when you render an embedded component" I think is much more clear.

sneakyvv:
> Hey! I just did a quick proof and edit of the docs so far - a team effort to explain these important concepts on a simple way. I've tried to remove the wording "embedded component". Even I mix up "embedded" and "nested" in my mind. It's not as succinct, but saying "when you pass a block to a component" instead of "when you render an embedded component" I think is much more clear.

Love it. Indeed I also always have to think twice to correctly separate nested vs embedded components. It took a while to get this terminology right in my head, but it will take perhaps equally long to stop referring to them as embedded components :-) 
Perhaps on a technical level, calling them that is actually still the most correct, especially seeing their similarity to the `embed` syntax, and even more so since the `ComponentNode` is in fact extending the `EmbedNode`. 🤓 
But again, I'm all for simplicity and even more for clarity in the documentation!

weaverryan:
> Perhaps on a technical level, calling them that is actually still the most correct

I think so yes. So we may sometimes still use that term - but let's see if we can avoid it for the "outside" world :)

sneakyvv:
@weaverryan As discussed on Slack, I also allowed references to outer blocks from with a regular (non-nested) component. 
Note that these blocks probably only make sense within an extending template, because otherwise the block would be rendered where it is defined ànd where it is referred as an outerblock. It even that might be desirable in some specific use cases, who knows 🤷

weaverryan:
Thanks Bart!

PR #991, opened 2023-07-07
[Docs] Update to <twig:component />

MattMaes:
Explain on using <twig:component />

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | Fix #... 
| License       | MIT

We're using Symfony UX a lot with our team and fully switched from using {{ component() }} to a more html-like approach with <twig:component />
They both do the same thing but the lather part leans in more with other front-end frameworks. Aside from that, makes using Symfony UX live components more enjoyable!

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -1402,7 +1402,7 @@ a :ref:`LiveAction <actions>`::
 
 Next, tell the ``form`` element to use this action:
 
-.. code-block:: html+twig
+.. code-block:: twig
 
     {# templates/components/PostForm.html.twig #}
     {# ... #}
@@ -1959,7 +1959,7 @@ and ``live_collection_entry_*`` as prefixes instead.
 
 For example, by default the add button is placed after the items (the comments in our case). Let's move it before them.
 
-.. code-block:: html+twig
+.. code-block:: twig
 
     {%- block live_collection_widget -%}
         {%- if button_add is defined and not button_add.rendered -%}
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -62,12 +62,18 @@ Done! Now render it wherever you want:
 
 .. code-block:: twig
 
-    {{ component('ProductSearch') }}
+    <twig:ProductSearch />
 
 As a user types into the box, the component will automatically re-render
 and show the new results!
 
 Want some demos? Check out https://ux.symfony.com/live-component#demo
+
+.. versionadded:: 2.8
+
+    Before version 2.8, you could not make use of ``<twig:... />`` and instead 
+    components like this had to be rendered as ``{{ component('ProductSearch') }}``.
+    Using ``<twig:... />`` makes them feel more like real html-components!
 
 Installation
 ------------
@@ -221,7 +227,7 @@ the component:
 
 .. code-block:: twig
 
-    {{ component('RandomNumber', { max: 500 }) }}
+    <twig:RandomNumber max="500" />
 
 But what's up with the ``LiveProp`` attribute? A property with the
 ``LiveProp`` attribute becomes a "stateful" property for this component.
@@ -1275,9 +1281,7 @@ Great! In the template for some page (e.g. an "Edit post" page), render a
     {% block body %}
         <h1>Edit Post</h1>
 
-        {{ component('PostForm', {
-            initialFormData: post,
-        }) }}
+        <twig:PostForm :initialFormData="post" />
     {% endblock %}
 
 Ok: time to build that ``PostForm`` component! The Live Components
@@ -1360,9 +1364,7 @@ or omit it entirely to let the ``initialFormData`` property default to ``null``:
     {# templates/post/new.html.twig #}
     {# ... #}
 
-    {{ component('PostForm', {
-        form: form
-    }) }}
+    <twig:PostForm :form="form" />
 
 Submitting the Form via a LiveAction
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -2877,11 +2869,9 @@ To fix this, you have two options:
 
 \1) Make the ``key`` dynamic so it will be different after adding a new item:
 
-.. code-block:: twig
+.. code-block:: html+twig
 
-    {{ component('InvoiceLineItemForm', {
-        key: 'new_line_item_'~lineItems|length,
-    }) }}
+    <twig:InvoiceLineItemForm :key="'new_line_item_'~lineItems|length" />
```
weaverryan, line 2874: Does this work?

```suggestion
    <twig:InvoiceLineItemForm key="new_line_item_{{ lineItems|length }}" />
```

I wrote the pre-lexer and I honestly can't remember if that'll work or not. It sure looks nicer.

weaverryan (commented):
What do other people think about this idea? @WebMamba? @kbond? @sneakyvv?

I also prefer the new `<twig:html` syntax. However, it's experimental and has some edge cases. At the same time, live components is also experimental... and if this makes it more enjoyable, that's worth something.

WebMamba:
Yeap, I completly agree with this PR. But there are still some edge case, maybe we can add some red flag on the doc so people are aware of this edge cases

kbond:
Yep, I'm in favor of showing the html-syntax everywhere.

src/LiveComponent/doc/index.rst
```diff
@@ -60,15 +60,21 @@ A real-time product search component might look like this::
...
 As a user types into the box, the component will automatically re-render
 and show the new results!
 
 Want some demos? Check out https://ux.symfony.com/live-component#demo
 
+.. versionadded:: 2.8
+
+    Before version 2.8, you could not make use of ``<twig:... />`` and instead 
+    components like this had to be rendered as ``{{ component('ProductSearch') }}``.
+    Using ``<twig:... />`` makes them feel more like real html-components!
```
weaverryan, line 76: How about:

```rst
The `<twig:... />`_ syntax for rendering components is experimental, but clear and concise. You
can also choose to render components via `{{ component('ProductSearch') }}`.

... at the bottom:

.. _<twig:... />: https://symfony.com/bundles/ux-twig-component/current/index.html#component-html-syntax
```

src/LiveComponent/doc/index.rst
```diff
@@ -1446,13 +1448,10 @@ If validation fails, you'll want the live component to render with the form
...
 
-.. code-block:: twig
+.. code-block:: html+twig
 
     {# templates/post/edit.html.twig #}
-    {{ component('PostForm', {
-        initialFormData: post,
-        form: form
-    }) }}
+    <twig:PostForn :initialFormData="post" :form="form" />
```
weaverryan, line 1454:
```suggestion
    <twig:PostForm :initialFormData="post" :form="form" />
```

src/LiveComponent/doc/index.rst
```diff
@@ -2531,10 +2528,11 @@ that will change if the component should be totally re-rendered:
     <div {{ attributes }}>
         <!-- ... -->
 
-        {{ component('TodoFooter', {
-            count: todos|length,
-            'data-live-id': 'todo-footer-'~todos|length
-        }) }}
+        <twig:TodoFooter 
+            :count="todos|length"
+            :data-live-id="'todo-footer-'~todos|length"
```
weaverryan, line 2533: Same question as below: does this work:

```suggestion
            data-live-id="todo-footer-{{ todos|length }}"
```

src/LiveComponent/doc/index.rst
```diff
@@ -2593,13 +2591,13 @@ child model changes, that should also update a model on the
...
+.. code-block:: html+twig
 
     {# templates/components/PostForm.html.twig #}
-    {{ component('TextareaField', {
-        dataModel: 'content',
-        error: _errors.get('content'),
-    }) }}
+    <twig:TextareaField 
+        data-model="content"
+        :error="_errors.get('content')"
```
weaverryan, line 2599: Subjective (I wonder what others think), but I'd prefer this here:

```suggestion
        error="{{ _errors.get('content') }}"
```

src/LiveComponent/doc/index.rst
```diff
@@ -2593,13 +2591,13 @@ child model changes, that should also update a model on the
 parent. To do this, pass a ``dataModel`` (or ``data-model``)
```
weaverryan, line 2591:
```suggestion
parent. To do this, pass a ``data-model`` (or ``dataModel``)
```

sneakyvv:
Also in favor of this! Btw @MattMaes is a colleague of mine 😄 (but wasn't aware of this PR 👍)

sneakyvv:
> and if this makes it more enjoyable, that's worth something.

I also think that it's particularly more enjoyable because we're _also_ doing live _embedded_ components already a lot in our company ;-)

sneakyvv:
Shouldn't the same change also be done to the twig components documentation (in one go)?

weaverryan:
> Shouldn't the same change also be done to the twig components documentation (in one go)?

Yes, I think it should. Though, we need to make sure that, in the TwigComponent docs, the old `{{` and `{% component` syntaxes have a section where they're properly & fully documented.

src/LiveComponent/doc/index.rst
WebMamba: Yes it works

MattMaes:
Just a quick message that I will have time tomorrow/this weekend to look and apply everything, but do keep feedback coming :D

weaverryan:
Friendly ping @MattMaes :)

src/LiveComponent/doc/index.rst
tacman: Another option. I wish there were a shorthand to the format filter, I use this syntax for clarity instead of ~, although it often ends up being longer.

```twig
:data-live-id="'todo-footer-%d'|format(todos|length)"
```

PR #1000, opened 2023-07-11
new TogglePassword component

feymo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | N/A
| License       | MIT

Hello to the Symfony UX community !
This PR propose a new Symfony UX Component.
The `TogglePassword` component allows users to switch visibility on password inputs through a dedicated button.
Default style is provided and english language is used for labels but everything is customizable to fit projects requirements.

Here are screenshots previews for the default rendering : 
image
image

Thanks you all in advance for your feedbacks and comments on the relevance of a component like this within Symfony UX ! 😄 

_Edit:_ The TogglePassword component propose : 
- A `FormExtension` transforming any `PasswordType` field into a Toggle Password by adding the `toggle` option ;
- A form theme with complete style, English labels and SVG icons for "toggable" password field "out of the box" ;
- Translation support for both toggle labels (with `TranslatableMessage` object or translation key string with domain) and the option to disable translation ;
- Customization for : labels, icons, classes for the toggle button element, and classes for the container element.


norkunas:
Icon customization could be useful too

jmsche:
Hi,

I'm wondering:

* could it be possible to have a form extension instead of a new form type? and so have an option available for the PasswordType like `'toggle' => true`
* like @norkunas, I think being allowed to customize the icons would be great
* about the hidden/visible labels, it would be nice to be able to pass a TranslatableMessage that would be translated automatically, else it would require to inject the TranslatorInterface or override the template
* maybe allow labels to be nullable (would display icons only)

feymo:
Thanks for your feedbacks @norkunas and @jmsche  ! 👍🏼 
I'm working on adding icons customization as well right now

About the `TranslatableMessage` support, my first idea was to, when translation is used, either indeed inject TranslatorInterface in the Form object, or pass labels options from twig view using trans filter (as shown in `doc/index.rst` code such as 
```twig
{{ form_row(form.password, { 'hidden_label': 'password_hide'|trans, 'visible_label': 'password_show'|trans }) }}
```
will work with provided theme. But i'll look into `TranslatableMessage` support, since it could ease even more translation support 👍🏼 
About using a `FormExtension` instead of a new `FormType`, i chose to stick with the others packages logic (such as `Dropzone`) and define a new `FormType`. But we could totally use a `FormExtension` if needed or preferred, no problem for me 😃

jmsche:
I think Autocomplete uses an extension as well :)

src/TogglePassword/assets/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "main": "dist/controller.js",
+    "types": "dist/controller.d.ts",
+    "config": {
+        "css_source": "src/style.css"
+    },
+    "symfony": {
+        "controllers": {
+            "toggle-password": {
+                "main": "dist/controller.js",
+                "webpackMode": "eager",
```
weaverryan: you can remove this line - it was for legacy versions of stimulus-bridge

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,69 @@
...
+    connect() {
+        this.element.insertAdjacentElement('afterend', this._createBtn());
+        this.dispatchEvent('connect', { element: this.element });
+    }
+
+    /**
+     * Create new button element
+     * @returns {HTMLButtonElement}
+     */
+    _createBtn(): HTMLButtonElement {
```
weaverryan:
```suggestion
    private createButton(): HTMLButtonElement {
```

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,69 @@
...
+export default class extends Controller {
+    declare readonly visibleLabelValue: string;
+    declare readonly hiddenLabelValue: string;
+
+    static values = {
+        visibleLabel: String,
+        hiddenLabel: String,
+    };
+
+    display = false;
```
weaverryan: Maybe:

```suggestion
    isDisplayed = false;
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+                }
+            }
+        },
+        "entrypoints": []
+    }
+
+.. note::
+
+   *Note*: you should put the value to ``false`` and not remove the line
+   so that Symfony Flex won’t try to add the line again in the future.
```
weaverryan:
```suggestion
   so that Symfony Flex won't try to add the line again in the future.
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+
+Once done, the default stylesheet won’t be used anymore and you can
+implement your own CSS on top of the TogglePassword.
+
+Extend the default behavior
```
weaverryan:
```suggestion
Extend the Default Behavior
```

src/TogglePassword/assets/dist/controller.js
```diff
@@ -0,0 +1,47 @@
...
+<path d="M10 12a2 2 0 100-4 2 2 0 000 4z" />
+<path fill-rule="evenodd" d="M.458 10C1.732 5.943 5.522 3 10 3s8.268 2.943 9.542 7c-1.274 4.057-5.064 7-9.542 7S1.732 14.057.458 10zM14 10a4 4 0 11-8 0 4 4 0 018 0z" clip-rule="evenodd" />
+</svg>`;
+        this.hiddenIcon = `<svg xmlns="http://www.w3.org/2000/svg" class="toggle-password-icon" viewBox="0 0 20 20" fill="currentColor">
+<path fill-rule="evenodd" d="M3.707 2.293a1 1 0 00-1.414 1.414l14 14a1 1 0 001.414-1.414l-1.473-1.473A10.014 10.014 0 0019.542 10C18.268 5.943 14.478 3 10 3a9.958 9.958 0 00-4.512 1.074l-1.78-1.781zm4.261 4.26l1.514 1.515a2.003 2.003 0 012.45 2.45l1.514 1.514a4 4 0 00-5.478-5.478z" clip-rule="evenodd" />
+<path d="M12.454 16.697L9.75 13.992a4 4 0 01-3.742-3.741L2.335 6.578A9.98 9.98 0 00.458 10c1.274 4.057 5.065 7 9.542 7 .847 0 1.669-.105 2.454-.303z" />
+</svg>`;
+    }
+    connect() {
+        this.element.insertAdjacentElement('afterend', this._createBtn());
```
weaverryan: What if we set `_createBtn()` to a new variable - e.g. `btn` then also passed that into the event? That way, someone could use `event.detail.btn` to modify the HTML of the button in any way they want.

weaverryan (commented):
Honestly, this looks pretty sweet 😎. I agree with the suggestions (including the form type extension) and about making a few things more overridable + the translations stuff. Once we've got that, I'll try this out for real. But I like how this aims to just "drop in and work" for most cases but then can be configured by overriding the CSS.

Also, congrats on PR #1000! 🥳

seb-jean:
We can help with that: https://stimulus-password-visibility.stimulus-components.com/

Their template is:

```html
<div data-controller="password-visibility">
  <input type="password" data-password-visibility-target="input" spellcheck="false" />

  <button type="button" data-action="password-visibility#toggle">
    <span data-password-visibility-target="icon">Eye</span>
    <span data-password-visibility-target="icon" class="hidden">Eye Slash</span>
  </button>
</div>
```

src/TogglePassword/assets/dist/controller.js
feymo: Indeed, great idea 👍🏼

src/TogglePassword/assets/package.json
feymo: It's done, thanks 👍🏼

src/TogglePassword/assets/src/controller.ts
feymo: It's done, thanks 👍🏼

feymo:
Thanks for the suggestion @seb-jean 👍🏼  but i must admit i'm not fully on board with their template approach on this one. I'd rather instantiate the stimulus controller on the password input (which is the main "actor" in this case), then create the action button and update its content without having all options loaded in the browser by default all the time. To me, the logic belongs to the stimulus controller scope, not the DOM

feymo:
Based on your reviews and suggestions (thanks again for those 🙏🏼 ) : 
- I've changed the `FormType` to use a `FormExtension`
- Icons are now customizable (and passing `null` hide the default icon)
- Labels are now nullable too

I'm still working on the translation support for labels 👍🏼

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Create new button element
+     * @returns {HTMLButtonElement}
+     */
+    private createButton(): HTMLButtonElement {
+        const button: HTMLButtonElement = document.createElement('button');
+        button.type = 'button';
+        button.classList.add('toggle-password-btn');
```
jmsche: Could you make the button classlist configurable? Would be useful e.g. to have a button appended to the input for Bootstrap like this:

image

Maybe the class of the div containing the form_widget could also be customized?

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+    class CredentialFormType extends AbstractType
+    {
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder
+                // ...
+                ->add('password', TogglePasswordType::class)
+                // ...
+            ;
+        }
```
jmsche: This part should be updated since you switched to a form extension :)

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+
+        // ...
+    }
+
+Customizing labels
+~~~~~~~~~~~~~~~~~~
+
+Symfony UX TogglePassword uses english by default on labels label for the toggle button.
+You can customize them by passing a ``hidden_label``  or a ``visible_label`` option to the field::
```
jmsche:
```suggestion
Symfony UX TogglePassword uses English by default for the toggle button's labels.
You can customize them by passing a ``hidden_label`` or a ``visible_label`` option to the field::
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+        {
+            $builder
+                // ...
+                ->add('password', TogglePasswordType::class, [
+                    'hidden_label' => 'Masquer',
+                    'visible_label' => 'Afficher',
+                ])
+                // ...
+            ;
+        }
```
jmsche: This part should be updated since you switched to a form extension :)

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+        "entrypoints": []
+    }
+
+.. note::
+
+   *Note*: you should put the value to ``false`` and not remove the line
+   so that Symfony Flex won't try to add the line again in the future.
+
+Once done, the default stylesheet won’t be used anymore and you can
+implement your own CSS on top of the TogglePassword.
```
jmsche:
```suggestion
Once done, the default stylesheet won't be used anymore and you can
implement your own CSS on top of the TogglePassword.
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,186 @@
...
+
+    class CredentialFormType extends AbstractType
+    {
+        public function buildForm(FormBuilderInterface $builder, array $options)
+        {
+            $builder
+                // ...
+                ->add('password', TogglePasswordType::class, [
+                    'attr' => ['data-controller' => 'my-toggle-password'],
+                ])
```
jmsche: This part should be updated since you switched to a form extension :)

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefined([
+            'toggle',
+            'hidden_label',
+            'visible_label',
+            'hidden_icon',
+            'visible_icon',
+        ]);
```
jmsche: As you use `setDefaults()`, I don't think using `setDefined()` is useful.

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $attr = $view->vars['attr'] ?? [];
+        $controllerName = 'symfony--ux-toggle-password--toggle-password';
+        $attr['data-controller'] = trim(sprintf('%s %s', $attr['data-controller'] ?? '', $controllerName));
+
+        $controllerValues = [];
+        $controllerValues['hidden-label'] = $options['hidden_label'] ?? null;
+        $controllerValues['visible-label'] = $options['visible_label'] ?? null;
+        $controllerValues['hidden-icon'] = $options['hidden_icon'] ?? null;
+        $controllerValues['visible-icon'] = $options['visible_icon'] ?? null;
```
jmsche: There are default values, so `?? null` is not required I guess.

jmsche (changes requested):

src/TogglePassword/assets/src/controller.ts
feymo: Indeed,  i didn't think of that use case, let's do that 😃 👍🏼

src/TogglePassword/doc/index.rst
feymo: Yes, i wanted to implement translation support first before updating doc 👍🏼 but i might do that first

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: That's right, but default values are not `null`, and if you choose to set `null` for one of these options (e.g to only display the icon) then the option is unset of the `$options` array. So i used `?? null `to ensure that `null` values are passed to the controller values

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: If i remove `setDefined()` and keep only `setDefaults()`, the options are not found by the `OptionsResolver` 😅

src/TogglePassword/doc/index.rst
feymo: Documentation updated 👍🏼  (will do another update once translation support and button classlist customization will be updated 😄 )

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
jmsche: What if you move the `setDefaults()` call above the `setAllowedTypes()` calls?

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
jmsche: I haven't tested in the context of a Form extension, but in a Form type, passing `null` as a value does not unset the option (just tested right now).

If the option is really unset, that I think it should be considered a bug in Symfony? I mean, `null` is still a real value.

feymo: Must have missed it but after checking again, you're right ! 
With default values set, if we pass `null` from our Form, the option isn't unset we get the `null` value 👍🏼

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: Yes indeed, make sense and work in that order 👍🏼

feymo:
Thanks again @jmsche for your review and suggestions 👍🏼 
- Toggle button classlist is now customizable
- Translation support has been implemented (with translation key or `TranslatableMessage`)
- Documentation is up-to-date with the latest changes & improvements made

jmsche:
Some quick notes as I'm on a phone, I'll review this more deeply tomorrow:

- maybe write "button" instead of "bon" everywhere?
- it's missing a .gitattributes file to lighten vendor installation
- the option about classes could end with "classes" instead of "class" as it may contain several classes

Thanks for all your work :)

src/TogglePassword/.gitattributes
```diff
@@ -0,0 +1,8 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.symfony.bundle.yaml export-ignore
+/phpunit.xml.dist export-ignore
+/assets/src/*.ts export-ignore
```
jmsche: I think the whole assets/src can be excluded as the min.css is included in the assets/dist dir

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,96 @@
...
+        $controllerValues['visible-label'] = $this->translateLabel($options['visible_label']);
+        $controllerValues['hidden-icon'] = $options['hidden_icon'];
+        $controllerValues['visible-icon'] = $options['visible_icon'];
+        if (isset($attr['toggle_classes'])) {
+            $controllerValues['button-classes'] = json_encode(
+                explode(' ', $attr['toggle_classes']),
+                \JSON_THROW_ON_ERROR
+            );
+            unset($attr['toggle_classes']);
+        }
```
jmsche: I think it would be better to have a proper option (e.g. `button_classes` requiring an array of strings) instead of looking for a specific attr.

src/TogglePassword/templates/form_theme.html.twig
```diff
@@ -0,0 +1,8 @@
+{%- block password_widget -%}
+    {%- set type = type|default('password') -%}
+    {%- if toggle is defined -%}
+        <div class="toggle-password-container">{{ block('form_widget') }}</div>
```
jmsche: Maybe make this class configurable as well?

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,86 @@
...
+     * @returns {HTMLButtonElement}
+     */
+    private createButton(): HTMLButtonElement {
+        const button: HTMLButtonElement = document.createElement('button');
+        button.type = 'button';
+        if (this.buttonClassesValue.length > 0) {
+            button.classList.add(...this.buttonClassesValue);
+        } else {
+            button.classList.add('toggle-password-button');
+        }
```
jmsche: The default class could be the default option in the form extension I think.

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+'use strict';
+
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
```
jmsche: Maybe the controller can be typed as `<HTMLInputElement>`?

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,207 @@
...
+            ;
+        }
+
+        // ...
+    }
+
+Customizing labels and icons
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony UX TogglePassword uses English by default and svg icons from `Heroicons`_ for the toggle button.
```
jmsche:
```suggestion
Symfony UX TogglePassword uses English by default and SVG icons from `Heroicons`_ for the toggle button.
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,207 @@
...
+You can customize them by passing a ``hidden_label``, ``visible_label``, ``hidden_icon``, or a ``visible_icon`` option
+to the field (use ``null`` to disable label or icon)::
```
jmsche:
```suggestion
You can customize them by passing ``hidden_label``, ``visible_label``, ``hidden_icon`` and ``visible_icon`` options
to the field (use ``null`` to disable label or icon)::
```

jmsche: Maybe specify that the labels are translated by default, and that one can pass a TranslatableMessage instance?

Also, I'm wondering if there should be a `toggle_translation_domain` when TranslatableMessage is not used for some reason (also, could help disabling translating the labels if set to `false` like it's done with the `choice_translation_domain` for the ChoiceType).

jmsche (changes requested):

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: Ok, actually i was hesitating between this approach (closer to classic Form usage) and your suggestion 😄 so let's change that !

src/TogglePassword/assets/src/controller.ts
feymo: You're right, doing so will simplify this step 👍🏼

src/TogglePassword/doc/index.rst
feymo: Interesting, i like this 👍🏼 i'm on it

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
jmsche: Well, it does not really apply to the input in the end :)

src/TogglePassword/templates/form_theme.html.twig
feymo: You're right, it was the last element unavailable for customization i think 😄 it's done 👍🏼

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,209 @@
...
+
+.. note::
+
+   *Note*: you should put the value to ``false`` and not remove the line
+   so that Symfony Flex won't try to add the line again in the future.
+
+Once done, the default stylesheet won't be used anymore and you can
+implement your own CSS on top of the TogglePassword.
+
+You can also only customize specific TogglePassword elements by overriding the defaults classes used.
```
jmsche:
```suggestion
You can also only customize specific TogglePassword elements by overriding the default classes.
```

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,95 @@
...
+            $attr[sprintf('data-%s-%s-value', $controllerName, $name)] = $value;
+        }
+        $view->vars['attr'] = $attr;
+        $view->vars['toggle_container_classes'] = implode(' ', $options['toggle_container_classes']);
+        $view->vars['toggle'] = true;
+    }
+
+    private function translateLabel(string|TranslatableMessage|null $label): ?string
+    {
+        if (null === $this->translator | null === $label) {
```
jmsche:
```suggestion
        if (null === $this->translator || null === $label) {
```

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,95 @@
...
+            return $label;
+        }
+
+        if ($label instanceof TranslatableMessage) {
+            return $this->translator->trans($label->getMessage(), domain: $label->getDomain());
```
jmsche: Hmm you could use the trans() method from the translatable message instead

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,95 @@
...
+            'button_classes' => ['toggle-password-button'],
+            'toggle_container_classes' => ['toggle-password-container'],
+        ]);
+        $resolver->setAllowedTypes('toggle', ['boolean']);
+        $resolver->setAllowedTypes('hidden_label', ['string', TranslatableMessage::class, 'null']);
+        $resolver->setAllowedTypes('visible_label', ['string', TranslatableMessage::class, 'null']);
+        $resolver->setAllowedTypes('hidden_icon', ['string', 'null']);
+        $resolver->setAllowedTypes('visible_icon', ['string', 'null']);
+        $resolver->setAllowedTypes('button_classes', ['array']);
+        $resolver->setAllowedTypes('toggle_container_classes', ['array']);
```
jmsche:
```suggestion
        $resolver->setAllowedTypes('button_classes', ['string[]']);
        $resolver->setAllowedTypes('toggle_container_classes', ['string[]']);
```

jmsche (changes requested):
Added some minor changes.

Still missing a note in the docs about translating labels. And what about the translation domain option for labels? Would you add it, or make the developer consider using TranslatableMessage instead?

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: nice catch ! thanks 🙏🏼

feymo:
Absolutely @jmsche ! Commiting your suggestion resolved the conversation and i forgot about it but i agree, i'll add a translation domain option for labels 👍🏼

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,97 @@
...
+            'toggle_translation_domain' => null,
+        ]);
+        $resolver->setAllowedTypes('toggle', ['boolean']);
+        $resolver->setAllowedTypes('hidden_label', ['string', TranslatableMessage::class, 'null']);
+        $resolver->setAllowedTypes('visible_label', ['string', TranslatableMessage::class, 'null']);
+        $resolver->setAllowedTypes('hidden_icon', ['string', 'null']);
+        $resolver->setAllowedTypes('visible_icon', ['string', 'null']);
+        $resolver->setAllowedTypes('button_classes', ['string[]']);
+        $resolver->setAllowedTypes('toggle_container_classes', ['string[]']);
+        $resolver->setAllowedTypes('toggle_translation_domain', ['string', 'null']);
```
jmsche: Should be a boolean too, to disable translating it (see ChoiceType)

jmsche (changes requested):

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: Indeed, my bad, it's fixed 😃

jmsche (approved):
Seems great!

Thanks for all the changes, and sorry for all the comments :smile:

feymo:
No worries at all, it was super relevant and interesting ! Thank you for your time and reactivity ! 👍🏼

feymo:
@weaverryan i've updated PR description to include all changes added with reviews and suggestions made, if you want to take a look and try it out, i think it looks quite clean and ready now 😄 👍🏼 🚀

src/TogglePassword/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2020-2021 Fabien Potencier
```
weaverryan:
```suggestion
Copyright (c) 2023-present Fabien Potencier
```

We will need to add that `-present` to the rest of the packages in another PR, but let's do it correctly here :)

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
+Symfony UX TogglePassword
+=========================
+
+Symfony UX TogglePassword is a Symfony bundle providing visibility toggle for password inputs
+in Symfony Forms. It is part of `the Symfony UX initiative`_.
+
+It allows visitors to switch the type of password field to text and vice versa.
```
weaverryan, line 7: We'll want to add a short gif to the docs right here

src/TogglePassword/assets/src/style.css
```diff
@@ -0,0 +1,28 @@
...
+    cursor: pointer;
+    display: flex;
+    flex-direction: row;
+    font-size: 0.875rem;
+    justify-items: center;
+    height: 1rem;
+    line-height: 1.25rem;
+    position: absolute;
+    right: 0.5rem;
+    top: -1.25rem;
```
weaverryan, line 21: This is going to be the trickiest part of this PR. Can we make some CSS that works in "most" cases. I was testing locally on https://ux.symfony.com/live-component/demos/auto-validating-form

I see the form theme template, which contains the `toggle-password-container` class, which is the key to making this all work "out of the box".  I also see that you prepended the config to add this form theme automatically - smart! Unfortunately... because ux.symfony.com already uses a form theme, by prepending the config, the custom `bootstrap_5_layout.html.twig` from that project shows up last in the list, and so its form theme wins (and the one from this bundle isn't used). I also think that there should probably be a way to "opt out" of the form theme - you may want to use your normal `password_widget` rendering (without the new div) and then use your own CSS to place things correctly.

I'm not convinced I have the correct solution for this. But my idea is:

A) Remove the prepend and instead document the user adding the form theme manually. It kinda sucks to do this... but then the user can opt out of it.

B) Document that, instead of the form theme, you can also style manually.

Another idea, might be to:

1) In `TogglePasswordTypeExtension`, add a new `toggle_form_theme` option, which defaults to `true`.
2) In `TogglePasswordTypeExtension`, if that's true, add a new item to the `block_prefixes` variable - like this https://github.com/symfony/ux/blob/24112fa46bb7e192d69de1b1e5f4365186bb296b/src/Autocomplete/src/Form/ParentEntityAutocompleteType.php#L56 - something like `ux_password_toggle`.
3) Change the form theme template to override *just* that block - e.g. `ux_password_toggle_widget`.

The idea would be that, if `toggle_form_theme` is true, then it will use your form theme. If false, it will skip and use the normal. Additionally, in your form theme (`form_theme.html.twig`), I believe you could then:

```diff
-{ block('form_widget') }}
+{ block('password_widget') }}
```

So in the event the user is overriding the `password_widget` in a custom form theme, we will still use that.

Also, we should document how this all works a bit - basically communicating:

A) There is a custom form theme that's activated, which wraps your widget in a `<div class="toggle-password-container">`.
B) You can turn this off with the `toggle_form_theme` option, but then you're on your own to style.

src/TogglePassword/assets/src/style.css
```diff
@@ -0,0 +1,28 @@
...
+}
+.toggle-password-button:hover {
+    color: #1d4ed8;
+    cursor: pointer;
```
weaverryan: Is this needed since it's already a `button`?

Also, it might be better to NOT specify a `color` here or on the non-hover variant. We can just use the default color from the user's CSS, right?

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,82 @@
...
+     * Toggle input type between "text" or "password" and update label accordingly
+     */
+    toggle(event: any): void {
+        this.isDisplayed = !this.isDisplayed;
+        const toggleButtonElement: HTMLButtonElement = event.currentTarget;
+        toggleButtonElement.innerHTML = this.isDisplayed
+            ? this.hiddenIcon + ' ' + this.hiddenLabelValue
+            : this.visibleIcon + ' ' + this.visibleLabelValue;
+        this.element.setAttribute('type', this.isDisplayed ? 'text' : 'password');
+        this.dispatchEvent('toggle', { element: this.element, display: this.isDisplayed, button: toggleButtonElement });
```
weaverryan: Instead of `toggle`, it is probably more useful to dispatch a `show` or `hide` event depending on which is happening.

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,82 @@
...
+     * Create new button element
+     * @returns {HTMLButtonElement}
+     */
+    private createButton(): HTMLButtonElement {
+        const button: HTMLButtonElement = document.createElement('button');
+        button.type = 'button';
+        button.classList.add(...this.buttonClassesValue);
+        button.setAttribute('tabindex', '-1');
+        button.addEventListener('click', this.toggle.bind(this));
+        button.innerHTML = this.visibleIcon + ' ' + this.visibleLabelValue;
```
weaverryan, line 60: If you test with the registration form on ux.symfony.com, you'll notice the hide/show button will disappear after filling in the password field. That is not a problem with this code - I need to make the `live` controller not lazy in `controllers.json`. Just an FYI :)

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+        const button = this.createButton();
+        this.element.insertAdjacentElement('afterend', button);
+        this.dispatchEvent('connect', { element: this.element, button: button });
+    }
+
+    /**
+     * Create new button element
+     * @returns {HTMLButtonElement}
+     */
```
weaverryan, line 53: No need for this comment - it doesn't add more information that's not obvious already by the method name or the return type.

src/TogglePassword/assets/src/controller.ts
```diff
@@ -0,0 +1,82 @@
...
+        button.setAttribute('tabindex', '-1');
+        button.addEventListener('click', this.toggle.bind(this));
+        button.innerHTML = this.visibleIcon + ' ' + this.visibleLabelValue;
+
+        return button;
+    }
+
+    /**
+     * Toggle input type between "text" or "password" and update label accordingly
+     */
```
weaverryan, line 67: But this IS a good comment - it adds a bit more info :)

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+                // ...
+                ->add('password', PasswordType::class, ['toggle' => true])
+                // ...
+            ;
+        }
+
+        // ...
+    }
+
+Customizing labels and icons
```
weaverryan:
```suggestion
Customizing Labels and Icons
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony UX TogglePassword uses English by default and SVG icons from `Heroicons`_ for the toggle button.
```
weaverryan:
```suggestion
The field uses the words "Show" and "Hide" by default and SVG icons from `Heroicons`_ for the toggle button.
```

I'm removing the part about English because if you need to translate, the more important thing is setting up the translations, which is mentioned below.

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        $resolver->setDefaults([
+            'toggle' => false,
+            'hidden_label' => 'Hide',
+            'visible_label' => 'Show',
+            'hidden_icon' => 'Default',
+            'visible_icon' => 'Default',
+            'button_classes' => ['toggle-password-button'],
+            'toggle_container_classes' => ['toggle-password-container'],
+            'toggle_translation_domain' => null,
```
weaverryan, line 49: I'm legitimately not sure, but do we need to make this configurable? I think it's more common to hardcode one: you're grouping your translations into a group, then telling the user which group/domain to use when translating.

If we used something like `ux_toggle_password`, we would then provide a `ux_toggle_password.en.xlf` file. It would simply translate "Show" to "Show".

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+
+
+.. note::
+
+   *Note*: Translation is supported for both label options.
+   You can either use a translation key string (and provide a translation domain with the ``toggle_translation_domain`` option)
+   or a ``Symfony\Component\Translation\TranslatableMessage`` object.
+   Passing ``false`` to the ``toggle_translation_domain`` option will disable translation for the labels.
+
+Customizing the design
```
weaverryan:
```suggestion
Customizing the Design
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+.. note::
+
+   *Note*: you should put the value to ``false`` and not remove the line
+   so that Symfony Flex won't try to add the line again in the future.
+
+Once done, the default stylesheet won't be used anymore and you can
+implement your own CSS on top of the TogglePassword.
+
+You can also only customize specific TogglePassword elements by overriding the default classes.
+Simply add your custom class name(s) using the ``button_classes`` option for the toggle element.
```
weaverryan:
```suggestion
Add your custom class name(s) using the ``button_classes`` option for the toggle element.
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+
+   *Note*: Translation is supported for both label options.
+   You can either use a translation key string (and provide a translation domain with the ``toggle_translation_domain`` option)
+   or a ``Symfony\Component\Translation\TranslatableMessage`` object.
+   Passing ``false`` to the ``toggle_translation_domain`` option will disable translation for the labels.
+
+Customizing the design
+~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony UX TogglePassword provides a default stylesheet in order to ease
```
weaverryan:
```suggestion
The package provides a default stylesheet in order to ease
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+            ;
+        }
+
+        // ...
+    }
+
+Extend the Default Behavior
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony UX TogglePassword allows you to extend its default behavior using a
```
weaverryan: If you need to additional control from JavaScript, you can leverage a few events dispatched by this package:

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+custom Stimulus controller:
+
+.. code-block:: javascript
+
+    // my-toggle-password_controller.js
```
weaverryan:
```suggestion
    // assets/controllers/my-toggle-password_controller.js
```

src/TogglePassword/doc/index.rst
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        disconnect() {
+            // You should always remove listeners when the controller is disconnected to avoid side-effects
+            this.element.removeEventListener('toggle-password:connect', this._onConnect);
+            this.element.removeEventListener('toggle-password:toggle', this._onToggle);
+        }
+
+        _onConnect(event) {
+            // The TogglePassword was just created
```
weaverryan: Show a quick example of the relevant `event.detail` stuff they will have access to

weaverryan (commented):
Looking really good!

src/TogglePassword/assets/src/controller.ts
feymo: make sense that's right 👍🏼

src/TogglePassword/assets/src/style.css
feymo: actually i kept this just to make sure that even if a CSS normalizer is used, the `cursor: pointer;` instruction will be added. I agree about the color instruction so i think it might be better to follow what the devs want to do on their project and remove both of these instructions here 👍🏼

src/TogglePassword/doc/index.rst
feymo: sure, on it 👍🏼

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
feymo: The approach was to allow users to implement their own custom labels with custom translations, to avoid implementing X translations files inside the bundle (for only 2 words each). Allowing custom translation domain would allow user to define or follow their own translation domain implementation, and it sorts of follow the same approach as label_translation_domain in others `FormType` objects

src/TogglePassword/assets/src/style.css
feymo: Nice catch ! Indeed, that's tricky but absolutely relevant !!
I'd go for your second suggestion, adding a toggle_form_theme boolean option, which would define a block prefix 👍🏼 
Let's go for that 😄

src/TogglePassword/src/Form/TogglePasswordTypeExtension.php
weaverryan: Very point about the label translation_domain! Ok, let's keep it

src/TogglePassword/assets/src/controller.ts
feymo: Thanks for the warning, i would have been really confused otherwise 🤣

src/TogglePassword/assets/src/style.css
feymo: Ok, done ! 
I went for `use_toggle_form_theme` option, with `true` as default value so it can add a custom block prefix : `ux_toggle_password` (to keep consistency with the names used everywhere else in the package) and i add documentation about the behaviour to expect.
I tried to be as clear as possible on this one while while remaining as concise as possible. I hope it's clear enough, otherwise i'll try another way 😄

src/TogglePassword/doc/index.rst
feymo: GIF added 👍🏼

src/TogglePassword/assets/package.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "@symfony/ux-toggle-password",
+    "description": "Toggle visibility of password inputs for Symfony Forms",
+    "license": "MIT",
+    "version": "1.0.0",
+    "main": "dist/controller.js",
+    "types": "dist/controller.d.ts",
+    "config": {
+        "css_source": "src/style.css"
+    },
```
weaverryan, line 10: You can remove this `config` key. I had never noticed that we had this on a few other, old packages. I believe it was added by accident there.

weaverryan (approved):
Small comment - but I'm going to merge. This is very high-quality work - it was a pleasure to review and test. Thank you! Once we have a sub-tree split, would you mind creating a PR to ux.symfony.com to add a page for it? Also, once released, we should add the option to https://symfony.com/doc/current/reference/forms/types/password.html with a link over to the bundle

src/TogglePassword/assets/package.json
feymo: sure, i'll remove it 👍🏼 
I thought it was needed for the css file provided with the bundle !

weaverryan: It probably was in a very early WIP version of UX + Flex - that's my best guess :)

feymo:
Thanks a lot to you too for your time and review @weaverryan 👍🏼 
I'll sure take care of the PR for adding a page on ux.symfony.com and i'll also suggest adding a reference on the PesswordType symfony doc once released 👍🏼

weaverryan:
Thank you! The sub-tree split and packagist package now exist - https://github.com/symfony/ux-toggle-password - so you should be able to create the ux.symfony.com page (we don't need a release for this - we bring in the latest, dev versions of packages inside the ux.symfony.com directory anyway).

feymo:
Awesome 👍🏼 i'll create the ux.symfony.com page in a couple of hours and submit a PR for it once ready 😄

weaverryan:
@feymo ux.symfony.com page is live! The release is out and the docs are up. 

Would you mind creating a PR to https://github.com/symfony/symfony-docs to:

A) Document the new option (with a link to the component) on https://symfony.com/doc/current/reference/forms/types/password.html
B) Also add the component to this list - https://symfony.com/bundles/StimulusBundle/current/index.html - that's actually in this repository - we should have done it earlier :) https://github.com/symfony/stimulus-bundle/blob/2.x/doc/index.rst

Thanks!

feymo:
@weaverryan Thanks ! i'll work on these as soon as possible 👍🏼

PR #1001, opened 2023-07-11
Fix custom controller scale documentation

DennisdeBest:
In a custom stimulus controller to update the configuration for the scales it is not an array but just an object, if not you get the following error : 

Invalid scale configuration for scale

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT


weaverryan:
It looks like you're right, thanks Dennis!

PR #1005, opened 2023-07-11
[LiveComponent] fix: respect `data-turbo="false"` when performing redirects

phramz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Tickets       | no
| License       | MIT

Any redirect response from the backend will trigger a `Turbo.visit()` regardless if the frontend component lives within a `data-turbo="false"` context.

This might lead to several problems when it comes to cross-site redirects since they will be dispatched via javascript `fetch()` where stricter rule apply (e.g. CORS etc).


weaverryan:
Nice one - thanks @phramz!

PR #1011, opened 2023-07-15
Update index.rst regex matching file extensions fix

gpetraroli:
/\\.(j|t)sx?$/. It seems to be using double backslashes (\\) to escape the dot (.) character, which is not necessary. The correct way to escape the dot character in a regular expression is by using a single backslash (\), like this: /\.(j|t)sx?$/.


weaverryan:
Thanks! This is what we had before - but something was lost in translation when moving it to the recipe. I appreciate you catching that!

PR #1018, opened 2023-07-22
[Svelte] Svelte v4 support

### LINKED ISSUE #1012: [Svelte] Warning / suggested configuration when installing via composer
Installing ux-svelte via `composer require symfony/ux-svelte` have warning / missing webpack config entry. I'm not sure though if this is related to Svelte or UX, since it does not export the path `./internal` when adding `conditionNames` to the Webpack configuration.

## The problem

- Create a new project by `symfony new --webapp my-app`
- Add Svelte to project `composer require symfony/ux-svelte`
- Build dependencies using yarn `yarn install` && `yarn build`

Dependencies added and installed:
```json
{
    "devDependencies": {
        ...
        "@symfony/stimulus-bridge": "^3.2.0",
        "@symfony/ux-svelte": "file:vendor/symfony/ux-svelte/assets",
        ...
        "svelte": "^3.0",
        "svelte-loader": "^3.1.9",
        ...
    },
   ...
}
```

This produces a warning, but still build successfully:
```bash
WARNING: You should add "svelte" to the "resolve.conditionNames" array in your webpack config. See https://github.com/sveltejs/svelte-loader#resolveconditionnames for more information
```

Attempting to fix the warning by changing default webpack configuration:
```js
let config = Encore.getWebpackConfig();
config.resolve.conditionNames = ["svelte"];
module.exports = config;
```

This will produce, if attempting to build again by `yarn build`, the following error message:
```bash
ERROR  Failed to compile with 1 errors                                                                                                                                                                                                   
error  in ./assets/svelte/controllers/Hello.svelte                                                                                                                                                                                       

Module not found: Error: Package path ./internal is not exported from package ./node_modules/svelte (see exports field in ./node_modules/svelte/package.json)
```

Which can be fixed by updating Svelte to ^4.0 in `packages.json`, this is however probably a "solution" unless `ux-svelte` can increment Svelte version without breaks:
```json
{
    "devDependencies": {
        ...
        "svelte": "^3.0",
        ...
    },
   ...
}
```

With a single dependency warning from ux-svelte:
```bash
warning " > @symfony/ux-svelte@1.0.0" has incorrect peer dependency "svelte@^3.0".
```

ChqThomas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | Fix #1012
| License       | MIT

This PR updates the peerDependencies of ux-svelte to support Svelte 4

This includes a slight changes with the typescript types : https://svelte.dev/docs/v4-migration-guide#sveltecomponenttyped-is-deprecated

There are no BC breaks, Svelte 3 is still supported :
```json
"peerDependencies": {
    "svelte": "^3.0 || ^4.0"
}
```

### DIFF
--- a/src/Svelte/assets/package.json
+++ b/src/Svelte/assets/package.json
@@ -21,12 +21,12 @@
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "svelte": "^3.0"
+        "svelte": "^3.0 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
         "@sveltejs/vite-plugin-svelte": "^2.4.6",
         "@types/webpack-env": "^1.16",
-        "svelte": "^3.0"
+        "svelte": "^3.0 || ^4.0"
     }
 }
--- a/src/Svelte/doc/index.rst
+++ b/src/Svelte/doc/index.rst
@@ -8,7 +8,7 @@ Svelte is a JavaScript framework for building user interfaces.
 Symfony UX Svelte provides tools to render Svelte components from Twig,
 handling rendering and data transfers.
 
-Symfony UX Svelte supports Svelte 3 only.
+Symfony UX Svelte supports Svelte 3 and Svelte 4.
 
 Installation
 ------------
@@ -41,10 +41,10 @@ Next, install a package to help Svelte:
 
 .. code-block:: terminal
 
-    $ npm install svelte svelte-loader --save-dev
+    $ npm install svelte-loader --save-dev
 
     # or use yarn
-    $ yarn add svelte svelte-loader --dev
+    $ yarn add svelte-loader --dev
 
 That's it! Any files inside ``assets/svelte/controllers/`` can now be rendered as
 Svelte components.
@@ -97,7 +97,7 @@ Svelte client-side component API:
         export let name = "Svelte";
     </script>
 
-    <div transition:fade>Hello {name}</div>
+    <div transition:fade|global>Hello {name}</div>
 
 
 .. code-block:: html+twig
--- a/src/Svelte/assets/dist/loader.d.ts
+++ b/src/Svelte/assets/dist/loader.d.ts
@@ -1,9 +1,9 @@
 import type { SvelteComponent } from 'svelte';
 import { ComponentCollection } from './components.js';
 declare global {
-    function resolveSvelteComponent(name: string): typeof SvelteComponent;
+    function resolveSvelteComponent(name: string): typeof SvelteComponent<any>;
     interface Window {
-        resolveSvelteComponent(name: string): typeof SvelteComponent;
+        resolveSvelteComponent(name: string): typeof SvelteComponent<any>;
     }
 }
 export declare function registerSvelteControllerComponents(svelteComponents?: ComponentCollection): void;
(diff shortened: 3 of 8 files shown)


src/Svelte/assets/package.json
```diff
@@ -21,7 +21,7 @@
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "svelte": "^3.0"
+        "svelte": "^3.0 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
```
weaverryan, line 27: Can you update the `devDependencies` svelte also below? That will, in effect, mean that we'll be testing now with svelte 4... but our CI isn't rich enough (and maybe shouldn't be, as the matrix could get huge) to test multiple versions of the JS.

ChqThomas: Sorry for the delay!

I can't get the test suite to work with svelte 4 :(
Svelte 4 requires ESM and that would require running jest with the --experimental-vm-modules options.
I followed [the guide from svelte-jester](https://github.com/svelteness/svelte-jester/issues/121#issuecomment-1646119671) and tried a lot of options but can't find a working transformer config for our monorepo + ts setup

aqmebrak: Any update on this ? I would very much like to use Svelte but not being in v4 is a bit of a blocker.
Any chance to try to use vitest as the testing lib only for the Svelte package ?
The majority of svelte devs use Vite as it is part of the Svelte docs, so I don't think sticking with jest is a good choice for the maintainability

weaverryan: This might be the same problem we're facing with chart.js v4 (it works fine, but we can't get the tests to pass because jest doesn't like esm). https://github.com/symfony/ux/pull/610#discussion_r1064788538

I'd love if someone experimented with switching to vitest and opened a PR - at least just to get things rolling and learn more.

ChqThomas: I quickly tried to install Vitest and just by following a migration guide 75% of the test suite pass
Maybe it will take a lot of work for the remaining 25% (some packages use specific jest mock libraries) but it may be possible!

weaverryan: That's great! Would you mind pushing your work on a separate PR? Then we could take a look and see what the remaining 25% is.

ChqThomas: @weaverryan Here you go #1202 😉

ChqThomas:
Thanks to Vitest it's now ready!

src/Svelte/doc/index.rst
```diff
@@ -41,10 +41,10 @@ Next, install a package to help Svelte:
 
 .. code-block:: terminal
 
-    $ npm install svelte svelte-loader --save-dev
+    $ npm install svelte-loader --save-dev
 
     # or use yarn
-    $ yarn add svelte svelte-loader --dev
+    $ yarn add svelte-loader --dev
```
weaverryan, line 47: `svelte` itself is no longer needed in the user's project?

ChqThomas: It is automatically added when you install `symfony/ux-svelte`, I think I removed it to be consistent with the other docs

Huluti:
Guys! Thank you very much for taking care of this 🙏🙏
https://github.com/symfony/ux/pull/610 will be unblocked also right?

weaverryan:
Thank you for pushing on this Thomas!

PR #1023, opened 2023-07-26
[TwigComponent] Support ...spread operator with html syntax

### LINKED ISSUE #845: [TwigComponent] Let HTML components work together with stimulus helpers
First and for all: the new HTML syntax is very nice! 💣 
There are some bugs in 2.8 (which seem to be resolved on master already)

Yet this on was not reported yet:

When using the twig `stimulus_` helper functions, you get an error:

```twig
<twig:atoms:button theme="danger" class="flex-no-shrink"
    {{ stimulus_controller('confirm', {
        message: 'Are you sure you want to delete task: "'~todo.task~ '"?',
    }) }}
    {{ stimulus_action('confirm', 'ask')|stimulus_action('live', 'action') }}
    data-action-name="remove"
>
    <twig:block name="children">Remove</twig:block>

    <twig:block name="icon">
        <twig:atoms:icon icon="garbage" class="fill-gray" />
    </twig:block>

</twig:atoms:button>
```

> Error rendering "molecules:xxx" component: Expected attribute name when parsing the "<twig:atoms:button" syntax.

The expectation here is that the data attributes should be rendered prior to parsing the html.

Maybe in the same line:

```twig
{% set attrs = {class: 'bg-black'}%}
<twig:atoms:button {{ attrs }} />
```

One might expect this to work as well.

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #845
| License       | MIT

I think that's a new feature like in twig.

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.11.0
+
+-   Support ...spread operator with html syntax (requires Twig 3.7.0 or higher)
+
 ## 2.9.0
 
 -   The `ComponentAttributes::defaults()` method now accepts any iterable argument.
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -22,7 +22,9 @@ class TwigPreLexer
     private int $length;
     private int $position = 0;
     private int $line;
-    /** @var array<string: name, bool: hasDefaultBlock> */
+    /**
+     * @var array<array{name: string, hasDefaultBlock: bool}>
+     */
     private array $currentComponents = [];
 
     public function __construct(int $startingLine = 1)
@@ -199,6 +201,15 @@ class TwigPreLexer
             $this->consumeWhitespace();
             if ($this->check('>') || $this->check('/>')) {
                 break;
+            }
+
+            if ($this->check('{{...') || $this->check('{{ ...')) {
+                $this->consume('{{...');
+                $this->consume('{{ ...');
+                $attributes[] = '...'.trim($this->consumeUntil('}}'));
+                $this->consume('}}');
+
+                continue;
             }
 
             $isAttributeDynamic = false;
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -216,5 +216,31 @@ final class TwigPreLexerTest extends TestCase
             '{% verbatim %}<twig:Alert/>{% endverbatim %}',
             '{% verbatim %}<twig:Alert/>{% endverbatim %}',
         ];
+
+        yield 'component_attr_spreading_self_closing' => [
+            '<twig:foobar bar="baz"{{...attr}}/>',
+            '{{ component(\'foobar\', { bar: \'baz\', ...attr }) }}',
+        ];
+        yield 'component_attr_spreading_self_closing2' => [
+            '<twig:foobar bar="baz"{{ ...customAttrs }} />',
+            '{{ component(\'foobar\', { bar: \'baz\', ...customAttrs }) }}',
+        ];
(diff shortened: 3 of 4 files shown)


src/TwigComponent/tests/Unit/TwigPreLexerTest.php
```diff
@@ -216,5 +216,31 @@ public function getLexTests(): iterable
             '{% verbatim %}<twig:Alert/>{% endverbatim %}',
             '{% verbatim %}<twig:Alert/>{% endverbatim %}',
         ];
+
+        yield 'component_attr_spreading_self_closing' => [
+            '<twig:foobar bar="baz"{{...attr}}/>',
```
WebMamba, line 221: Instead of having ... between {{ why not just do:
```html
<twig:foobar bar="baz" ...attr />
```

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -201,6 +203,15 @@ private function consumeAttributes(string $componentName): string
                 break;
             }
 
+            if ($this->check('{{...') || $this->check('{{ ...')) {
```
WebMamba, line 206: I think you should consume the whitespace before doing this check, tab, or multiples tab can be valid syntax too.

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
norkunas: For me eyes it's easier to find that it's a variable like in twig

src/TwigComponent/src/Twig/TwigPreLexer.php
norkunas: It is consumed 4 lines before, isn't it?

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
kbond: A little funky looking but what about `<twig:foobar bar="baz" :...attr />`? (in addition to `{{`). This sort of lines up with dynamic attribute values: https://symfony.com/bundles/ux-twig-component/current/index.html#passing-props-as-html-attributes

norkunas: My proposed solution is already in our prod and working, so I'd prefer as is 😁

weaverryan: Looking at React:

```jsx
<button className={className} {...other} />
```

And Vue:

```
<button :className="className" v-bind="other">
```

Given these, the logical options are:

```
<twig:Button className={{ className }} {{...other}} />

or

<twig:Button live-bind="other">
```

And that `live-bind` thing looks silly :). So I agree with the proposed syntax.

src/TwigComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.11.0
+
+-   Support ...spread operator with html syntax
```
weaverryan:
```suggestion
-   Support ...spread operator with html syntax (requires Twig 3.7.0 or higher)
```

weaverryan (commented):
Wow! Fast turn-around on this :). We also need a mention in the docs - probably one example at the bottom of https://github.com/symfony/ux/blob/2.x/src/TwigComponent/doc/index.rst#passing-props-as-html-attributes would do it

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
Kocal: Note that JSX for Vue.js also use `{... }` syntax.

`{{... }}` is nice because the two brackets reminds us about Twig

norkunas:
> Wow! Fast turn-around on this :). We also need a mention in the docs - probably one example at the bottom of https://github.com/symfony/ux/blob/2.x/src/TwigComponent/doc/index.rst#passing-props-as-html-attributes would do it

Added

weaverryan:
Thank you Tomas!

PR #1031, opened 2023-08-02
[TwigComponent] Improve exception message when component not found

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #...
| License       | MIT

I have over 150 components and to me it's just useless to list all component names in exception message and to look for the right name in a very long list.

Used same code to get alternatives like in Symfony codebase.

### DIFF
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -217,6 +217,32 @@ final class ComponentFactory
      */
     private function throwUnknownComponentException(string $name): void
     {
-        throw new \InvalidArgumentException(sprintf('Unknown component "%s". The registered components are: %s. And no matching anonymous component template was found', $name, implode(', ', array_keys($this->config))));
+        $message = sprintf('Unknown component "%s".', $name);
+        $lowerName = strtolower($name);
+        $nameLength = \strlen($lowerName);
+        $alternatives = [];
+
+        foreach (array_keys($this->config) as $type) {
+            $lowerType = strtolower($type);
+            $lev = levenshtein($lowerName, $lowerType);
+
+            if ($lev <= $nameLength / 3 || str_contains($lowerType, $lowerName)) {
+                $alternatives[] = $type;
+            }
+        }
+
+        if ($alternatives) {
+            if (1 === \count($alternatives)) {
+                $message .= ' Did you mean this: "';
+            } else {
+                $message .= ' Did you mean one of these: "';
+            }
+
+            $message .= implode('", "', $alternatives).'"?';
+        } else {
+            $message .= ' And no matching anonymous component template was found.';
+        }
+
+        throw new \InvalidArgumentException($message);
     }
 }
--- a/src/TwigComponent/tests/Integration/ComponentFactoryTest.php
+++ b/src/TwigComponent/tests/Integration/ComponentFactoryTest.php
@@ -152,17 +152,24 @@ final class ComponentFactoryTest extends KernelTestCase
     public function testCannotGetConfigByNameForNonRegisteredComponent(): void
     {
         $this->expectException(\InvalidArgumentException::class);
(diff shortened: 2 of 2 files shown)


kbond (approved):
Great idea!

norkunas:
Should we list all components in case it didn't find any alternatives? :slightly_smiling_face:

kbond:
I'm thinking no. I don't think the Symfony code base does that.

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -194,6 +194,31 @@ private function postMount(object $component, array $data): array
      */
     private function throwUnknownComponentException(string $name): void
     {
-        throw new \InvalidArgumentException(sprintf('Unknown component "%s". The registered components are: %s', $name, implode(', ', array_keys($this->config))));
+        $message = sprintf('Unknown component "%s".', $name);
+
+        $alternatives = [];
+
+        foreach (array_keys($this->config) as $type) {
+            $lowerName = strtolower($name);
```
stloyd: Could be moved out of loop

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -194,6 +194,31 @@ private function postMount(object $component, array $data): array
...
+            $lowerType = strtolower($type);
+
+            $lev = levenshtein($lowerName, $lowerType);
+
+            if ($lev <= \strlen($lowerName) / 3 || str_contains($lowerType, $lowerName)) {
```
stloyd: Length is static, could be moved out of loop

weaverryan (approved):
levenshtein! I like this - can you rebase?

norkunas:
Rebased

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -217,6 +217,32 @@ private function isAnonymousComponent(string $name): bool
...
+        $nameLength = \strlen($lowerName);
+        $alternatives = [];
+
+        foreach (array_keys($this->config) as $type) {
+            $lowerType = strtolower($type);
+            $lev = levenshtein($lowerName, $lowerType);
+
+            if ($lev <= $nameLength / 3 || str_contains($lowerType, $lowerName)) {
+                $alternatives[] = $type;
+            }
```
stloyd, line 231: What about low cost check first instead of doing both?
```suggestion
            if (str_contains($lowerType, $lowerName)) {
                $alternatives[] = $type;
            } elseif (levenshtein($lowerName, $lowerType) <= $nameLength / 3) {
                $alternatives[] = $type;
            }
```

norkunas: Are you going to get this exception all the time? :slightly_smiling_face:

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -217,6 +217,32 @@ private function isAnonymousComponent(string $name): bool
...
+        if ($alternatives) {
+            if (1 === \count($alternatives)) {
+                $message .= ' Did you mean this: "';
+            } else {
+                $message .= ' Did you mean one of these: "';
+            }
+
+            $message .= implode('", "', $alternatives).'"?';
+        } else {
+            $message .= ' And no matching anonymous component template was found.';
```
WebMamba, line 243: Nice! Just tried it out, it's super useful! Do you think we can apply the Levenshtein to look also for AnonymousComponent?

norkunas: :thinking: probably all available twig templates then should be iterated :)

norkunas: As I'm not using anonymous components, then I'll leave it for you as that was your feature :)

weaverryan:
This is great - thanks Tomas!

PR #1040, opened 2023-08-07
add swup progress bar plugin to docs

### LINKED ISSUE #1039: [Swup] Add support for progress bar plugin
it would be nice to have the swup progress bar plugin supported.
and having it disabled by default unless passing a useProgressBar option as true i guess

yassinefikri:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | none
| License       | MIT

updating docs to include swup progress bar plugin as example

### DIFF
--- a/src/Swup/doc/index.rst
+++ b/src/Swup/doc/index.rst
@@ -144,6 +144,7 @@ Stimulus controller:
     // assets/controllers/myswup_controller.js
 
     import { Controller } from '@hotwired/stimulus';
+    import SwupProgressPlugin from '@swup/progress-plugin';
 
     export default class extends Controller {
         connect() {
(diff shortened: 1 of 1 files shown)


weaverryan:
Hi!

This is a nice PR, but the issue is tricky: each plugin we add increases the download size of the controller. So, even if you aren't using the progress bar plugin, it will now be downloaded by the user (with Encore, it'll be packaged into the final built files & with asset mapper, the `import` will be followed and downloaded).

Instead, I think we should encourage people to create a small swup controller where they configure the exact behavior they want - so enhancing the docs here https://symfony.com/bundles/ux-swup/current/index.html#extend-the-default-behavior to show hooking up the progress bar would be 🙌 .

yassinefikri:
I agree, i also think that extending the controller will give easier access to customize the progress bar plugin options
i will update the PR as you mentioned

weaverryan:
Awesome - thank you Yassine!

PR #1041, opened 2023-08-08
Props tag remove props from the attribute list

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | none
| License       | MIT

An assertion in ComponentFactory is blocking us to use non-scalar values for anonymous components. @weaverryan do you know why this check was made for?


WebMamba:
@weaverryan friendly ping here. I think we should merge a solution for this issue before the next release. Sorry about that 😔

weaverryan:
This originates from https://github.com/symfony/ux/pull/255/files#diff-ad19b699a899e152e1619ea828ba39d72a419702a33769092f35f39d65c23073R75 - because any "extra" props become attributes, which need to be strings. The idea was to give a better error now than some "cannot be cast to a string" type of error when the attributes are rendered. Perhaps we could put some check into `ComponentAttributes` for this.

src/TwigComponent/tests/Fixtures/templates/components/UserCard.html.twig
```diff
@@ -0,0 +1,4 @@
+<div class='user-card'>
+    <p>{{ user.name }}</p>
+    <p>{{ user.email }}</p>
+</div>
```
weaverryan, line 7: I thought that props that were passed to anonymous components became attributes *unless* you used the `{% props` tag. Did I misunderstand that?

WebMamba: No the props tag only allow to set default value on attributes or to required attributes, but all props are at first an attribute

weaverryan: Does this mean that, if we added `<div {{ attributes }}>` in this component, it would break because it would try to "print" the `user` as an attribute?

kbond: I also misunderstood. I think this could make `{{attributes}}` non-stringable like Ryan says(?)

> I thought that props that were passed to anonymous components became attributes unless you used the {% props tag.

Feel like this _should_ be the case as I don't think there's any other way to make the distinction.

src/TwigComponent/src/AnonymousAnonymousComponentTemplateParser.php
```diff
@@ -46,4 +46,23 @@ public function findAnonymousComponentTemplate(string $name): ?string
 
         return null;
     }
+
+    public function findComponentProps(string $name): array
+    {
+        $loader = $this->environment->getLoader();
+        $templateContent = $loader->getSourceContext($this->findAnonymousComponentTemplate($name))->getCode();
+
+        $pattern = '/{%\s*props\s*(.*?)\s*%}/';
```
WebMamba: I'm open for a better way extract the props from the template

WebMamba:
Ok so here is how I trying to do this now: before mounting I extract the props defined in the component template, then when we mount the component, the component only takes as props the props defined in the tag, the rest goes to the attributes object.

weaverryan:
> Ok so here is how I trying to do this now: before mounting I extract the props defined in the component template, then when we mount the component, the component only takes as props the props defined in the tag, the rest goes to the attributes object.

Thanks for this :). I have an idea - I'm wondering if it is feasible or not. What if:

A) We move the check for scalar props into `ComponentAttributes`. So, we move this - https://github.com/symfony/ux/pull/1041/files#diff-ad19b699a899e152e1619ea828ba39d72a419702a33769092f35f39d65c23073R108 - into `ComponentAttributes::__toString()` (https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/ComponentAttributes.php#L31) (note, back in the day throwing an exception from `__string()` caused a nasty, non-obvious exception - I can't remember if that's true anymore.

B) At this point, if I understand correctly, ALL props I've passed to my anonymous component will be inside `ComponentAttributes`. We fix this entirely inside `PropsNode`: when we see `{% props`, we do 2 things:
 * 1) Like now, if the variable doesn't exist, we use the default from `{% props`. So, this is what we have already.
 * 2) We also *remove* any props in `{% props` from `ComponentAttributes`.
 
 And, voila! As long as you have `{% props` at the top of your template (a requirement), then by the time you render `{{ attributes }}`, anything that are actually props will have already been removed. If we do this, I think `AnonymousComponent` won't even need to mount `$props` anymore or put it into a property - I think it could be empty.

WDYT?

WebMamba:
@weaverryan I think I manage to implement your solution! Tell me what you think about it! 😁

src/TwigComponent/tests/Fixtures/templates/anonymous_component_none_scalar_prop.html.twig
```diff
@@ -0,0 +1 @@
+<twig:UserCard :user='user'/>
```
weaverryan: We should pass some real attribute here also - like `class="foo"` and test that it is rendered

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -146,4 +150,9 @@ public function add($stimulusDto): self
         // add the remaining attributes for values/classes
         return $clone->defaults($controllersAttributes);
     }
+
+    public function remove($key): void
+    {
+        unset($this->attributes[$key]);
+    }
```
weaverryan, line 159: The other methods in this class are immuatable - so we should probably do the same thing here: clone and return a modified version of ourself. Then in the Twig magic, it'll need to be:

```
$context[\'attributes\'] = $context[\'attributes\']->remove(...)
```

weaverryan (commented):
I like this! I do see one test failing

WebMamba:
Thanks for your review! Looks like the tests failing are related to the form changed but not this PR

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -35,6 +35,10 @@ public function __toString(): string
             function (string $carry, string $key) {
                 $value = $this->attributes[$key];
 
+                if (!\is_scalar($value) && null !== $value) {
+                    throw new \LogicException(sprintf('A "%s" prop was passed when creating the component. No matching %s property or mount() argument was found, so we attempted to use this as an HTML attribute. But, the value is not a scalar (it\'s a %s). Did you mean to pass this to your component or is there a typo on its name?', $key, $key, get_debug_type($value)));
```
weaverryan:
```suggestion
                    throw new \LogicException(sprintf('A "%s" prop was passed when creating the component. No matching "%s" property or mount() argument was found, so we attempted to use this as an HTML attribute. But, the value is not a scalar (it\'s a %s). Did you mean to pass this to your component or is there a typo on its name?', $key, $key, get_debug_type($value)));
```

src/TwigComponent/tests/Fixtures/templates/components/UserCard.html.twig
```diff
@@ -0,0 +1,6 @@
+{% props user %}
+
+<div {{ attributes }}>
+    <p>{{ user.name }}</p>
+    <p>{{ user.email }}</p>
```
weaverryan, line 5: Also add and test for this:

```
<p>class variable defined? {{ class is defined ? 'yes': 'no' }}
```

This is the last thing to fix. I just tested locally. Everything works great (`user` is removed from `attributes`, `{% props` can be used to define defaults, etc) except that if something is NOT defined in `{% props` (i.e. `class`) is IT being made available as a variable, and it should not be. So, in the test, we want to assert that we see:

```
class variable defined? no
```

weaverryan (commented):
One last thing! I just tried it locally - it was simple and fun to play with.

weaverryan:
Friendly ping @WebMamba - I think that last tweak, relative to the work you've already done, is quite minor :)

WebMamba:
Sorry for the delay I am on holiday with all my family so I don't have a lot of time for coding! Here is a fix tell me what you think about it 😁

weaverryan:
Sorry for the bother on holiday - but thank you for this! It looks great now!

PR #1052, opened 2023-08-11
fix(live): ensure non-class service id can be used

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | n/a
| License       | MIT

Without this fix, you cannot register live components with a non-class service id.


weaverryan:
Thanks Kevin!

PR #1054, opened 2023-08-16
[Live][Doc] ComponentWithFormTrait::getFormInstance -> getForm

jcrombez:


kbond (approved):

kbond:
Good catch, thanks @jcrombez.

PR #1057, opened 2023-08-18
[LiveComponent] Add missing typehints to HydrationExtensionInterface

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no /not sure
| New feature?  | no /not sure
| Tickets       | N/A
| License       | MIT

I have a custom hydration extension that in `supports` method checks if the provided class has my custom marker attribute through Reflection, but SA complains:

> Parameter `#1` $objectOrClass of class ReflectionClass constructor expects class-string<T of object>|T of object, string given.


weaverryan:
Thanks Tomas!

PR #1058, opened 2023-08-18
[Docs] Anonymous twig components

mariecharles:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | none
| License       | MIT

Add documentation for the recently added support of anonymous twig components [https://github.com/symfony/ux/pull/802](https://github.com/symfony/ux/pull/802)

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -1050,6 +1050,70 @@ And in your component template you can access your embedded block
         {% block footer %}{% endblock %}
      </div>
 
+Anonymous Components
+--------------------
+
+Sometimes a component is simple enough that it doesn't have any complex logic or injected services.
+In this case, you can skip the class and only create the template. The component name is determined
+by the location of the template (see `Twig Template Namespaces`_):
+
+.. code-block:: html+twig
+
+    {# templates/components/Button/Primary.html.twig #}
+    <button {{ attributes.defaults({ class: 'primary' }) }}>
+        {% block content %}{% endblock %}
+    </button>
+
+Then use your component with ``:`` to navigate through sub-directories (if there are any):
+
+.. code-block:: html+twig
+
+    {# index.html.twig #}
+    ...
+    <div>
+       <twig:Button:Primary>Click Me!</twig:Button:Primary>
+    </div>
+
+    {# renders as: #}
+    <button class="primary">Click Me!</button>
+
+Like normal, you can pass extra attributes that will be rendered on the element:
+
+.. code-block:: html+twig
+
+    {# index.html.twig #}
+    ...
+    <div>
+       <twig:Button:Primary type="button" name="foo">Click Me!</twig:Button:Primary>
+    </div>
+
+    {# renders as: #}
+    <button class="primary" type="button" name="foo">Click Me!</button>
+
+You can also pass a variable (prop) into your template:
+
+.. code-block:: html+twig
+
+    {# index.html.twig #}
+    ...
+    <div>
+        <twig:Button icon="fa-plus" type="primary" role="button">Click Me!</twig:Button>
+    </div>
+
+To tell the system that ``icon`` and ``type`` are props and not attributes, use the ``{% props %}`` tag at the top of your template.
+
+.. code-block:: html+twig
+
+    {# templates/components/Button.html.twig #}
+    {% props icon = null, type = 'primary' %}
+
+    <button {{ attributes.defaults({ class: 'btn btn-'~type }) }}>
+        {% block content %}{% endblock %}
+        {% if icon %}
+            <span class="fa-solid fa-{{ icon }}"></span>
+        {% endif %}
+    </button>
+
 Test Helpers
 ------------
 


src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,79 @@ And in your component template you can access your embedded block
...
+
+.. code-block:: html+twig
+
+     {# index.html.twig #}
+     ...
+    <div>
+       <twig:Button label="Click Me!" />
+    </div>
+
+You can also define your component to allow overwriting attributes:
```
WebMamba: Since this PR #1041, this is not true anymore. Now if you defined props in your component {% props label %} the label is not available by the attributes object but only with {{ label }}. The attributes object contains only passed to the object without a defined props

WebMamba (approved):
Well done Marie 🎉😁

src/TwigComponent/doc/index.rst
weaverryan: But won't the code below still work after #1041?

* `primary` is in `{% props` so it will NOT be inside `attributes` but we WILL have a `primary` varaible
* Then, we use that `primary` variable (boolean) to set the class to the string `primary` or `secondary`.

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,79 @@ And in your component template you can access your embedded block
         {% block footer %}{% endblock %}
      </div>
 
+Anonymous TwigComponent
+-----------------------
+
+Sometimes, reusable elements do not require complex logic or injected service to render what it could be processed
+from one template itself. (e.g. retrieving a customized primary button that could take a different label).
```
weaverryan: How about:

```
Sometimes a component is simple enough that it doesn't have any complex logic or injected services.
In this case, you can skip the class and only create the template. The component name is determined
by the location of the template (see `Twig Template Namespaces`_):
```

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,79 @@ And in your component template you can access your embedded block
...
+
+.. code-block:: html+twig
+
+    {# templates/components/Button/Primary.html.twig #}
+    <button class="primary">
+        {% block content %}
+        {% endblock %}
+    </button>
+
+Then use your component with ``:`` to navigate through the directories:
```
```suggestion
Then use your component with ``:`` to navigate through sub-directories (if there are any):
```

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,79 @@ And in your component template you can access your embedded block
...
+
+     {# index.html.twig #}
+     ...
+    <div>
+       <twig:Button:Primary>
+           Click Me!
+       </twig:primary>
+    </div>
+
+Now let's pass props to your button by defining a ``{% props %}`` tag in its view.
```
weaverryan: I think we should first show passing a normal "attribute" and show how that works very normally (show the `{{ attributes }}` being printed in the template). Maybe a `role` attribute, for example.

Then we can show how you can, in addition to an attribute, pass a prop using `{% props}`.

weaverryan (commented):
Look great - I had just noticed this was missing moments ago when I looked for the docs :)

src/TwigComponent/doc/index.rst
WebMamba: It was more about the sentence do you think overwritting is the right word to use here, or should we have more detail ?

weaverryan:
Related PR is merged - the final tweaks on this are ready to be done :)

seb-jean:
The closing tag is not good:
```twig
</twig:primary>
```

You should put:
```twig
</twig:Button:Primary>
```

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,108 @@ And in your component template you can access your embedded block
         {% block footer %}{% endblock %}
      </div>
 
+Anonymous TwigComponent
```
weaverryan: Anonymous Components

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,108 @@ And in your component template you can access your embedded block
...
+-----------------------
+
+Sometimes a component is simple enough that it doesn't have any complex logic or injected services.
+In this case, you can skip the class and only create the template. The component name is determined
+by the location of the template (see `Twig Template Namespaces`_):
+
+.. code-block:: html+twig
+
+    {# templates/components/Button/Primary.html.twig #}
+    <button class="primary">
```
weaverryan: It would be more flexible to immediately how `attributes` being used:

```
<button {{ attributes.defaults({ class: 'primary' }) }}>
```

We don't need to introduce the concept of attributes here, since it's not unique to anonymous components :)

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,108 @@ And in your component template you can access your embedded block
...
+
+.. code-block:: html+twig
+
+    {# templates/components/Button.html.twig #}
+    <button {{ attributes }}>
+        {% block content %}
+        {% endblock %}
+    </button>
+
+Then pass any further data to your reusable "button" like so:
```
weaverryan: Like normal, you can pass extra attributes that will be rendered on the element:

src/TwigComponent/doc/index.rst
```diff
@@ -1046,6 +1046,108 @@ And in your component template you can access your embedded block
...
+    <div>
+        <twig:Button role="button" class="foo">
+            Click Me!
+        </twig:Button>
+    </div>
+
+    {# renders as: #}
+    <button role="button" class="foo">Click Me!</button>
+
+Now let's pass props to your "button" by defining a ``{% props %}`` tag in its view.
```
weaverryan: Let's reverse this section: show the usage first, then show the `{% props` syntax. Something like:

You can also pass a variable (prop) into your template:

<twig:Button role="button" icon="fa-plus" type="primary">Buy now!</twig:Button>

To tell the system that `icon` and `type` are props and not attributes, use the `{% props` tag at the top of your template:

```
{% props icon, type = 'primary' %}

<button {{ attributes.defaults({
    class: 'btn btn-'~type
} }}>
    {% if icon %}
        <span class="fa-solid fa-{{ icon }}"></span>
    {% endif %}
</button>
```

I don't love the `attributes.defaults()` syntax, but it does make the components more flexible.

mariecharles:
Thank you @weaverryan for your help

weaverryan:
Gorgeous 😍 - great work Marie!

PR #1064, opened 2023-08-20
[LiveComponent] Fix tests now that a live id is always added, even for embedded components

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | 
| License       | MIT

Embedded components can also be live now, so they also need a live-id.

The `TestLiveComponent` trait was added after the live-id change had been introduced in the embedded live components branch, so it now needed to be synced with those changes.

The conflicting change: https://github.com/symfony/ux/pull/913/files#diff-c1c2695c57da687bdba39e2b9e5285f30e9b87f50886bf645a048545575f6659

### DIFF
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -40,6 +40,8 @@ final class TestLiveComponent
     ) {
         $this->client->catchExceptions(false);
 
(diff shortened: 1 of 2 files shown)


weaverryan:
Thanks @sneakyvv!

PR #1070, opened 2023-08-24
[TwigComponent] Add support for namespaced templates in TemplateMap

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | 
| License       | MIT

## Problem
If a template is rendered as "_@MyNamespace/some/template.html.twig_", and it contains an embedded live component, then that name is used for the host template attribute. When obscuring this name, an error is thrown, because it cannot be found in the template map (which consists of real file paths).

## Solution
Parse the name when compiling a ComponentNode, just like the FilesystemLoader would, so it becomes "_some/template.html.twig_"

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentNode.php
+++ b/src/TwigComponent/src/Twig/ComponentNode.php
@@ -68,7 +68,7 @@ final class ComponentNode extends EmbedNode
             ->raw('), ')
             ->raw($this->getAttribute('only') ? '[]' : '$context')
             ->raw(', ')
-            ->string($this->getAttribute('name'))
+            ->string($this->parseTemplateName($this->getAttribute('name')))
             ->raw(', ')
             ->raw($this->getAttribute('index'))
             ->raw(");\n")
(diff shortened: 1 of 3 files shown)


src/TwigComponent/src/Twig/ComponentNode.php
```diff
@@ -91,4 +91,20 @@ public function compile(Compiler $compiler): void
             ->raw("\n")
         ;
     }
+
+    /**
+     * Copied from Twig\Loader\FilesystemLoader, and adjusted to needs for this class
+     */
+    private function parseTemplateName(string  $name): mixed
+    {
+        if (isset($name[0]) && '@' == $name[0]) {
```
stloyd, line 100:
```suggestion
        if ('@' === $name[0] ?? '') {
```
but also test is missing to prove the fix is working.

weaverryan: Yea - a test would be ideal

sneakyvv: Test added, and branch rebased which now incorporates @weaverryan's latest test fixes. All green now here as well!

@stloyd ~~I corrected the loose comparison.~~ The (copied) code from Twig also used loose comparison. I guess because in this case loose comparison would be faster ~~and if the type didn't match it would also fail (i.e. nothing would equal '@' anyway)~~ and the type is already guaranteed because of the parameter type hint. For the same reason I kept the `isset` instead of `??` because it only needs one check to know to fail while the `??` would always do the assignment and then do the comparison always.

weaverryan:
Thank you Bart!

PR #1072, opened 2023-08-24
[TwigComponent][LiveComponent] Add Embedded Live Components doc

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | 
| License       | MIT

Adds documentation for the feature added by #913

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2911,6 +2911,36 @@ To fix this, you have two options:
         }
     }
 
+.. _passing-blocks:
+
+Passing Content (Blocks) to Components
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Passing content via blocks to Live components works completely the same way you would `pass content to Twig Components`_.
+Except with one important difference: when a component is re-rendered, any variables defined only in the
+"outside" template will not be available. For example, this won't work:
+
+.. code-block:: twig
+
+    {# templates/some_page.html.twig #}
+    {% set message = 'Variables from the outer part of the template are only available during  the initial render' %}
+
+    {% component Alert %}
+        {% block content %}{{ message }}{% endblock %}
+    {% endcomponent %}
+
+Local variables do remain available:
+
+.. code-block:: twig
+
+    {# templates/some_page.html.twig #}
+    {% component Alert %}
+        {% block content %}
+            {% set message = 'this works during re-rendering!' %}
+            {{ message }}
+        {% endblock %}
+    {% endcomponent %}
+
 Advanced Functionality
 ----------------------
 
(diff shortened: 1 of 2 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -2453,6 +2453,40 @@ The parent-child system is *smart*. And with a few tricks
 (:ref:`such as the key prop for lists of embedded components <rendering-quirks-with-list-of-embedded-components>`),
 you can make it behave exactly like you need.
 
+.. _passing-blocks:
+
+Passing Blocks to Components
```
weaverryan: I think this is in the wrong section - it's right in the middle of the docs on "nested components" - i.e. parent-child setup. 

Also, WDYT about:

`Passing Content (Blocks) to Components`

From a user's perspective, I am "passing content into my component". Blocks are the mechanism, but not the star. We should update the same title in Twig components.

src/LiveComponent/doc/index.rst
```diff
@@ -2453,6 +2453,40 @@ The parent-child system is *smart*. And with a few tricks
...
+----------------------------
+
+Passing blocks to Live components works completely the same way you would `pass blocks for Twig Components`_. Except...
```
weaverryan: Passing content via blocks to Live...

Also, the next paragraph is a bit thick. We know the technical word is "context", users think "variables". Suggestion to replace the entire paragraph with:

... Except with one important difference: when a component is re-rendered, any variables defined
only in the "outside" template will note be available. For example, this won't work:

src/LiveComponent/doc/index.rst
```diff
@@ -2453,6 +2453,40 @@ The parent-child system is *smart*. And with a few tricks
...
+When re-rendering a Live component, only the template for thàt component is rendered. The blocks defined in the
+"parent" template are grabbed via the smart Twig templating system, but the context from the parent template is *NOT*
+available during re-rendering. So, keep in mind to not refer to any outer context variables within blocks of a live component.
+
+This won't work:
+
+.. code-block:: twig
+
+    {# templates/some_page.html.twig #}
+    {% set message = 'Outer context only works on first render' %}
```
```suggestion
    {% set message = 'Variables from the outer part of the template are only available during  the initial render' %}
```

src/LiveComponent/doc/index.rst
sneakyvv: Oh yeah, you're right 🙈 
I just put it right after nested blocks, because that's also where it is in the Twig Component docs, but indeed in the Live Components there are subsections. I placed it behind those subsections now.

src/LiveComponent/doc/index.rst
sneakyvv: yep agreed, changed

src/LiveComponent/doc/index.rst
sneakyvv: I also changed the section titles.

src/LiveComponent/doc/index.rst
```diff
@@ -2911,6 +2911,36 @@ To fix this, you have two options:
...
     }
 
+.. _passing-blocks:
+
+Passing Content (Blocks) to Components
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Passing content via blocks to Live components works completely the same way you would `pass content to Twig Components`_.
+Except with one important difference: when a component is re-rendered, any variables defined only in the
+"outside" template will note be available. For example, this won't work:
```
```suggestion
"outside" template will not be available. For example, this won't work:
```

kbond (changes requested):
Looks good!

weaverryan:
Thanks Bart!

PR #1079, opened 2023-08-30
[LiveComponent] Use `display:revert` for `data-loading` style

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

Using `display:inline-block` breaks the inner workings if for example `data-loading` was used on `<tr>` element, so the table row contents are incorrectly displayed.


weaverryan:
I was waiting for this problem to come up. But woh! `revert`? I had no idea that existed - but it looks perfect!

weaverryan:
Thank you Tomas!

PR #1082, opened 2023-08-30
[TwigComponent][LiveComponent] Fix Live embedded component within namespaced template

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | 
| License       | MIT

This fix is related to #1070 .
That PR made sure that a namespaced template could also be found in the TemplateMap, which contains paths without the namespace, so that the right template name (or actually obscuring hash) is used for the host template reference for live embedded components within such a template (which is used to find the right blocks during a re-render). 

However, because the Parser was still using the namespaced template name to calculate the deterministic index number for an embedded template, it would not match with the index of the stripped / namespace-less template name, causing it to have a different index for the embedded template during re-render then during parsing, resulting in a template not found error.

This PR now also strips namespaces from template names during parsing.

### DIFF
--- a/src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
+++ b/src/LiveComponent/tests/Functional/EventListener/LiveComponentSubscriberTest.php
@@ -311,6 +311,45 @@ final class LiveComponentSubscriberTest extends KernelTestCase
         ;
     }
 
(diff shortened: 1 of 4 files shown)


weaverryan:
Fantastic - thank you for the fix and the test!

PR #1086, opened 2023-08-31
[Site] Use new symfonycasts/dynamic-forms for dependent fields

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | None
| License       | MIT

After @bendavies solved the form event puzzle of handling multiple levels of dependent form fields, we set out to try to create a library to wrap that complexity. I'm pretty darn excited to announce a v0.1.0 version of that! https://github.com/SymfonyCasts/dynamic-forms - which works especially great with live components.

The library may have edge cases and might change - but a good first step for exposure is to use it in our demo. It works GREAT (even better than the old implementation, where a validation error was sometimes triggered that we didn't want!).

Cheers!

### DIFF
--- a/ux.symfony.com/composer.json
+++ b/ux.symfony.com/composer.json
@@ -49,6 +49,7 @@
         "symfony/ux-vue": "2.x-dev",
         "symfony/validator": "^6.3.1",
         "symfony/yaml": "^6.3.0",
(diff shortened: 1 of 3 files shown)


bendavies:
nice!

PR #1088, opened 2023-09-01
Add `debug:component` command

StevenRenaux:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | #102 
| License       | MIT

Related to the Future Ideas section of #102 
I added a debug:component command.

### Debugging component

The `debug:component` command lists all your application components (TwigComponent and LiveComponent) who lives in templates/components directory:

```
$ php bin/console debug:component

+---------------+-----------------------------+------------------------------------+
| Component     | Class                       | Template                           |
+---------------+-----------------------------+------------------------------------+
| Coucou        | App\Components\Alert        | components/Coucou.html.twig        |
| RandomNumber  | App\Components\RandomNumber | components/RandomNumber.html.twig  |
| Test          | App\Components\foo\Test     | components/foo/Test.html.twig      |
| Button        | Anonymous component         | components/Button.html.twig        |
| foo:Anonymous | Anonymous component         | components/foo/Anonymous.html.twig |
+---------------+-----------------------------+------------------------------------+

```

Pass the name of some component to this argument to print the component details:

```
$ php bin/console debug:component Test

+---------------------------------------------------+-------------------------------+
| Property                                          | Value                         |
+---------------------------------------------------+-------------------------------+
| Component                                         | Test                          |
| Class                                             | App\Components\foo\Test       |
| Template                                          | components/foo/Test.html.twig |
| Properties (type / name / default value if exist) | string $type = success        |
|                                                   | string $message               |
+---------------------------------------------------+-------------------------------+
```

To get the details about an anonymous component who is rendered with sub directory, just add it to the name:

```
    <div>
        <twig:foo:Anonymous label="Click Me!" :disabled="true" />
    </div>
```

```
$ php bin/console debug:component foo:Anonymous

+--------------------------------------------+------------------------------------+
| Property                                   | Value                              |
+--------------------------------------------+------------------------------------+
| Component                                  | foo:Anonymous                      |
| Class                                      | Anonymous component                |
| Template                                   | components/foo/Anonymous.html.twig |
| Properties (name / default value if exist) | label                              |
|                                            | name = toto                        |
+--------------------------------------------+------------------------------------+
```

### DIFF
--- a/src/TwigComponent/composer.json
+++ b/src/TwigComponent/composer.json
@@ -34,6 +34,7 @@
         "twig/twig": "^2.14.7|^3.0.4"
     },
     "require-dev": {
+        "symfony/console": "^5.4|^6.0",
         "symfony/css-selector": "^5.4|^6.0",
         "symfony/dom-crawler": "^5.4|^6.0",
         "symfony/framework-bundle": "^5.4|^6.0",
--- a/src/TwigComponent/src/Command/ComponentDebugCommand.php
+++ b/src/TwigComponent/src/Command/ComponentDebugCommand.php
@@ -1,0 +1,273 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Command;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Finder\Finder;
+use Symfony\UX\LiveComponent\Attribute\LiveAction;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\TwigComponent\Attribute\PostMount;
+use Symfony\UX\TwigComponent\Attribute\PreMount;
+use Symfony\UX\TwigComponent\ComponentFactory;
+use Symfony\UX\TwigComponent\ComponentMetadata;
+use Symfony\UX\TwigComponent\Twig\PropsNode;
+use Twig\Environment;
+
+#[AsCommand(name: 'debug:twig-component', description: 'Display current components and them usages for an application')]
+class ComponentDebugCommand extends Command
+{
+    public function __construct(private string $twigTemplatesPath, private ComponentFactory $componentFactory, private Environment $twigEnvironment, private iterable $components)
+    {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setDefinition([
+                new InputArgument('name', InputArgument::OPTIONAL, 'A component name'),
+                new InputOption('dir', null, InputOption::VALUE_REQUIRED, 'Show all components with a specific directory in templates', 'components'),
+            ])
+            ->setHelp(<<<'EOF'
+                The <info>%command.name%</info> display all components in your application:
+
+                    <info>php %command.full_name%</info>
+
+                Find all components within a specific directory in templates by specifying the directory name with the <info>--dir</info> option:
+
+                    <info>php %command.full_name% --dir=bar/foo</info>
+
+                EOF
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $name = $input->getArgument('name');
+        $componentsDir = $input->getOption('dir');
+
+        if (null !== $name) {
+            try {
+                $metadata = $this->componentFactory->metadataFor($name);
+            } catch (\Exception $e) {
+                $io->error($e->getMessage());
+
+                return Command::FAILURE;
+            }
+
+            $class = $metadata->get('class');
+            $live = null;
+            $allProperties = [];
+
+            if ($class) {
+                if ($metadata->get('live')) {
+                    $live = 'X';
+                }
+
+                $reflectionClass = new \ReflectionClass($class);
+                $properties = $reflectionClass->getProperties();
+                $allLiveProperties = [];
+
+                foreach ($properties as $property) {
+                    if ($property->isPublic()) {
+                        $visibility = $property->getType()?->getName();
+                        $propertyName = $property->getName();
+                        $value = $property->getDefaultValue();
+                        $propertyAttributes = $property->getAttributes(LiveProp::class);
+
+                        $propertyDisplay = $visibility.' $'.$propertyName.(null !== $value ? ' = '.$value : '');
+
+                        if (\count($propertyAttributes) > 0) {
+                            $allLiveProperties[] = $propertyDisplay;
+                        } else {
+                            $allProperties[] = $propertyDisplay;
+                        }
+                    }
+                }
+
+                $methods = $reflectionClass->getMethods();
+                $allEvents = [];
+                $allActions = [];
+
+                foreach ($methods as $method) {
+                    if ('mount' === $method->getName()) {
+                        $allEvents[] = 'Mount';
+                    }
+
+                    foreach ($method->getAttributes() as $attribute) {
+                        if (PreMount::class === $attribute->getName()) {
+                            $allEvents[] = 'PreMount';
+                            break;
+                        }
+
+                        if (PostMount::class === $attribute->getName()) {
+                            $allEvents[] = 'PostMount';
+                            break;
+                        }
+
+                        if (LiveAction::class === $attribute->getName()) {
+                            $allActions[] = $method->getName();
+                            break;
+                        }
+                    }
+                }
+            } else {
+                $allProperties = $this->getPropertiesForAnonymousComponent($metadata);
+            }
+
+            $componentInfos = [
+                ['Component', $name],
+                ['Live', $live],
+                ['Class', $class ?? 'Anonymous component'],
+                ['Template', $metadata->getTemplate()],
+                ['Properties', \count($allProperties) > 0 ? implode("\n", $allProperties) : null],
+            ];
+
+            if (isset($allLiveProperties) && \count($allLiveProperties) > 0) {
+                $componentInfos[] = ['Live Properties', implode("\n", $allLiveProperties)];
+            }
+            if (isset($allEvents) && \count($allEvents) > 0) {
+                $componentInfos[] = ['Events', implode("\n", $allEvents)];
+            }
+            if (isset($allActions) && \count($allActions) > 0) {
+                $componentInfos[] = ['LiveAction Methods', implode("\n", $allActions)];
+            }
+
+            $table = new Table($output);
+            $table->setHeaders(['Property', 'Value'])->setRows($componentInfos);
+            $table->render();
+
+            return Command::SUCCESS;
+        }
+
+        $finderTemplates = new Finder();
+        $finderTemplates->files()->in("{$this->twigTemplatesPath}/components");
+
+        $anonymousTemplatesComponents = [];
+        foreach ($finderTemplates as $template) {
+            $anonymousTemplatesComponents[] = $template->getRelativePathname();
+        }
+
+        $componentsWithClass = [];
+        foreach ($this->components as $class) {
+            $reflectionClass = new \ReflectionClass($class);
+            $attributes = $reflectionClass->getAttributes();
+
+            foreach ($attributes as $attribute) {
+                $arguments = $attribute->getArguments();
+
+                $name = $arguments['name'] ?? $arguments[0] ?? null;
+                $template = $arguments['template'] ?? $arguments[1] ?? null;
+
+                if (null !== $template || null !== $name) {
+                    if (null !== $template && null !== $name) {
+                        $templateFile = str_replace('components/', '', $template);
+                        $metadata = $this->componentFactory->metadataFor($name);
+                    } elseif (null !== $name) {
+                        $templateFile = str_replace(':', '/', "{$name}.html.twig");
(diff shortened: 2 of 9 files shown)


kbond (commented):
This is great work @StevenRenaux!

Couple things:
1. In the listing of components, could you add a Live? (yes/no) column?
2. When showing details about a component, can you indicate whether it is live or not?
3. Also in the details, can you list the live props?

kbond:
4. Could you also add a section to the docs (twig & live)

StevenRenaux:
@kbond I will try to work on your suggestions

1. In the listing of components, could you add a Live?
I suggest to add a column `Type` directly with AsTwigComponent, AsLiveComponent or empty if Anonymous.
I can get it with the ReflectionClass and the method `getAttributes`.

2. When showing details about a component, can you indicate whether it is live or not?
As I suggest for the 1. I can add a property `Type` 

3. Also in the details, can you list the live props?
I can try to divide in two sections the Props part like  `Properties (name / default value if exist)` and after `Live Properties` 

4. Could you also add a section to the docs (twig & live)
Yeah it will be a TODO ;)

kbond:
Sounds good!

> I can get it with the ReflectionClass and the method getAttributes.

I think this is available in the metadata (has a `live` key) - this would be a more robust way to check.

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,214 @@
...
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $name = $input->getArgument('name');
+
+        if ($name) {
```
WebMamba:
```suggestion
        if ($name !== null) {
```

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,214 @@
...
+                $properties = $reflectionClass->getProperties();
+
+                foreach ($properties as $property) {
+                    if ($property->isPublic()) {
+                        $visibility = $property->getType()?->getName();
+                        $propertyName = $property->getName();
+                        $value = $property->getDefaultValue();
+
+                        $allProperties = [
+                            ...$allProperties,
```
WebMamba: Funny to see that in PHP, I expected something more like:
```suggestion
                          $allProperties[] = $visibility.' $'.$propertyName.(null !== $value ? ' = '.$value : '')
```

src/TwigComponent/composer.json
```diff
@@ -34,6 +34,7 @@
         "twig/twig": "^2.14.7|^3.0.4"
     },
     "require-dev": {
+        "symfony/console": "^5.4|^6.0",
```
WebMamba, line 37: Is it the right place to add this command? Maybe in the StimulusBundle?

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,214 @@
...
+                $table->addRow($info);
+            }
+
+            $table->render();
+
+            return Command::SUCCESS;
+        }
+
+        $finderTemplates = new Finder();
+        $finderTemplates->files()->in("{$this->twigTemplatesPath}/components");
```
WebMamba, line 162: @kbond @StevenRenaux here made the choice to only look for anonymous components inside the `components` directory, even if it's not really true. The convention is to put your components inside the `components` directory, but do you think the command should look even for none conventional component?

StevenRenaux:
Hi @kbond 

I updated the code with your suggestions.
To know if it's a LiveComponent I used `$metadata->get('live')` as you suggested to get one asked component.
But to get the list of all of them I was originaly in a loop after a `$reflectionClass->getAttributes()`  so I took the position to just make a check if the attribute depend on AsTwigComponent::class or AsLiveComponent::class. Let me know if I also need to update this part :wink: 

For one component :

```
$ php bin/console debug:component RandomNumber
+---------------------------------------------------+-----------------------------------+
| Property                                          | Value                             |
+---------------------------------------------------+-----------------------------------+
| Component                                         | RandomNumber                      |
| Type                                              | AsLiveComponent                   |
| Class                                             | App\Components\RandomNumber       |
| Template                                          | components/RandomNumber.html.twig |
| Properties (type / name / default value if exist) | string $name = toto               |
|                                                   | string $type = test               |
| Live Properties                                   | int $max = 1000                   |
|                                                   | int $min = 10                     |
+---------------------------------------------------+-----------------------------------+
```


And for all :

```
$ php bin/console debug:component
+---------------+-----------------------------+------------------------------------+-----------------+
| Component     | Class                       | Template                           | Type            |
+---------------+-----------------------------+------------------------------------+-----------------+
| Coucou        | App\Components\Alert        | components/Coucou.html.twig        | AsTwigComponent |
| RandomNumber  | App\Components\RandomNumber | components/RandomNumber.html.twig  | AsLiveComponent |
| Test          | App\Components\foo\Test     | components/foo/Test.html.twig      | AsTwigComponent |
| Button        | Anonymous component         | components/Button.html.twig        |                 |
| foo:Anonymous | Anonymous component         | components/foo/Anonymous.html.twig |                 |
+---------------+-----------------------------+------------------------------------+-----------------+
```

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,214 @@
...
+            } catch (\Exception $e) {
+                $io->error($e->getMessage());
+
+                return Command::FAILURE;
+            }
+
+            $class = $metadata->get('class');
+            $allProperties = [];
+
+            if ($class) {
```
WebMamba, line 80: I think it can be useful to display: 
- properties with the ExposeInTemplate attribute, and if the name is set show the name.
- if the component has a mount, premount, and postmount hooks

And for LiveConponents:
- as @kbond said show if the components is live
- show live props
- show live actions

WebMamba (commented):
Whoooo thanks a lot for it @StevenRenaux! 🧡

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,250 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Finder\Finder;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+use Symfony\UX\TwigComponent\ComponentFactory;
+use Symfony\UX\TwigComponent\Twig\PropsNode;
+use Twig\Environment;
+
+#[AsCommand(name: 'debug:component', description: 'Display current components and them usages for an application')]
```
94noni: just passing by, this name `debug:component` sounds to wide to me

perhaps `debug:twig-component` or `debug:ux-component` ?

src/TwigComponent/src/Command/ComponentDebugCommand.php
StevenRenaux: Hi @WebMamba and @kbond, an update with `Events` and `LiveAction`, it could be look like this for a Live component: 
 
```
 $ php bin/console debug:component RandomNumber
+--------------------+-----------------------------------+
| Property           | Value                             |
+--------------------+-----------------------------------+
| Component          | RandomNumber                      |
| Type               | AsLiveComponent                   |
| Class              | App\Components\RandomNumber       |
| Template           | components/RandomNumber.html.twig |
| Properties         | string $name = toto               |
|                    | string $type = test               |
| Live Properties    | int $max = 1000                   |
|                    | int $min = 10                     |
| Events             | Mount                             |
|                    | PreMount                          |
| LiveAction Methods | resetMax                          |
+--------------------+-----------------------------------+
```

And for a TwigComponent

```
$ php bin/console debug:component Coucou
+------------+-----------------------------+
| Property   | Value                       |
+------------+-----------------------------+
| Component  | Coucou                      |
| Type       | AsTwigComponent             |
| Class      | App\Components\Alert        |
| Template   | components/Coucou.html.twig |
| Properties | string $type = success      |
|            | string $message             |
| Events     | Mount                       |
|            | PreMount                    |
+------------+-----------------------------+
```

But for the `#[ExposeInTemplate]` what could I render for you? 
The name of the property in the class or the one one displayed in the template? And with the getter or not?

src/TwigComponent/src/Command/ComponentDebugCommand.php
StevenRenaux: Let see what the others think about that but IMHO it would be a replication of code

src/TwigComponent/composer.json
StevenRenaux: Doesn't know either and it's the reason why i putted here

src/TwigComponent/src/Command/ComponentDebugCommand.php
weaverryan: Hmm, yea probably it should be `debug:twig-component` - we already have a `make:twig-component`. It's annoyingly longer, but "component" is a pretty generic word - e.g. "Symfony components".

src/TwigComponent/src/Command/ComponentDebugCommand.php
kbond: I was thinking about this part too... What about a command option to customize (with `/components` being the default)?

kbond:
> But to get the list of all of them I was originaly in a loop after a $reflectionClass->getAttributes() so I took the position to just make a check if the attribute depend on AsTwigComponent::class or AsLiveComponent::class. Let me know if I also need to update this part 😉

The issue with using these attributes is components can be registered without (ie 3rd party bundle registering components purely with a tag).

src/TwigComponent/src/Command/ComponentDebugCommand.php
StevenRenaux: Added  --dir option in the command

StevenRenaux:
> > But to get the list of all of them I was originaly in a loop after a $reflectionClass->getAttributes() so I took the position to just make a check if the attribute depend on AsTwigComponent::class or AsLiveComponent::class. Let me know if I also need to update this part 😉
> 
> The issue with using these attributes is components can be registered without (ie 3rd party bundle registering components purely with a tag).

Updated as you submitted ;)

StevenRenaux:
> I was thinking about this part too... What about a command option to customize (with `/components` being the default)?

If you want components with a specific dir:
```

php bin/console debug:twig-component --dir=bar
+----------------+-------------------------------+------------------------------+-----------------+
| Component      | Class                         | Template                     | Type            |
+----------------+-------------------------------+------------------------------+-----------------+
| OtherDirectory | App\Components\OtherDirectory | bar/OtherDirectory.html.twig | AsTwigComponent |
+----------------+-------------------------------+------------------------------+-----------------+

```

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,265 @@
...
+
+    protected function configure(): void
+    {
+        $this
+            ->setDefinition([
+                new InputArgument('name', InputArgument::OPTIONAL, 'A component name'),
+                new InputOption('dir', null, InputOption::VALUE_REQUIRED, 'Show all components with a specific directory in templates', 'components'),
+            ])
+            ->setHelp(<<<'EOF'
+                The <info>%command.name%</info> display current components and them usages for an application:
```
weaverryan: Just this?

```suggestion
                The <info>%command.name%</info> displays all components in your application:
```

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,265 @@
...
+
+                    <info>php %command.full_name%</info>
+
+                Find all components with a specific directory in templates by specifying the directory name with the <info>--dir</info> option:
```
weaverryan:
```suggestion
                Find all components within a specific directory in templates by specifying the directory name with the <info>--dir</info> option:
```

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,265 @@
...
+            $class = $metadata->get('class');
+            $type = null;
+            $allProperties = [];
+
+            if ($class) {
+                $type = 'AsTwigComponent';
+
+                if ($metadata->get('live')) {
+                    $type = 'AsLiveComponent';
+                }
```
weaverryan, line 83: What about making the column title just "Live?". And if you are live, we print an "X". If not live, leave that cell empty.

src/TwigComponent/src/Command/ComponentDebugCommand.php
```diff
@@ -0,0 +1,265 @@
...
+                        }
+
+                        if (LiveAction::class === $attribute->getName()) {
+                            $allActions[] = $method->getName();
+                            break;
+                        }
+                    }
+                }
+            } else {
+                $source = $this->twigEnvironment->load($metadata->getTemplate())->getSourceContext();
```
weaverryan: Can we extract this block to a `private getPropertiesForAnonymousComponent()` and the above block a `private `getPropertiesForClassComponent()`? Just for readability.

weaverryan (commented):
This is cool! Though, tbh, it makes me wish we had more centralized ways to get metadata about a component. It feels like we are reimplementing a lot of logic - like for finding what public properties will be made into variables, etc.

kbond (approved):
Agree with Ryan there's a lot of duplication but this can be refactored to some kind of metadata registry.

We should probably do this sooner, rather than later so things don't get out of sync. For now, it's out of the scope of this PR.

WebMamba:
Yes I was thinking the same thing when I read for the first time this PR. Thanks @StevenRenaux for digging into all of this

weaverryan:
Ok then. But I'll be watching for some PR's to centralize some of that stuff ;). 

@StevenRenaux Can you fix the conflict? I can usually do that with our merge tool, but the rebase is looking oddly ugly.

StevenRenaux:
@weaverryan Conflict resolution done and let me know when it will be done and I will update this command :wink:

weaverryan:
Thank you for your hard work Steven and congrats on this great PR. I forgot that I hadn't merged it yet, and I was needing it yesterday and trying to use it. So today my work will be better ;)

PR #1090, opened 2023-09-03
[LiveComponent] Implement hydratation of DTO object

### LINKED ISSUE #955: [Live] Better DTO Dehydration
Hi!

LiveComponent can dehydrate many different `LiveProp` data types - scalars, DateTime objects, entities, collection of entities, etc. But it can't dehydrate DTO's out-of-the-box.

### The Setup:

Suppose you have a live component and you want to leverage DTO. This example uses a form, but that's not a requirement:

```php
CustomerDetails
{
    use DefaultActionTrait;
    use ComponentWithFormTrait;
​
    #[LiveProp]
    public CustomerInfoDto $customerInfoDto;

    {
        return $this->createForm(CustomInfoType::class, $this->customerInfoDto);
    }
}
```

And the DTO:

```php
class CustomerInfoDto
{
    public string $name;
    public string $address1;
    private string $city;

    public User $owner;

    // getter and setter for city since it's private
}
```

Currently, the `$customerDto` can't be dehydrated. You COULD set `useSerializerForHydration: true`, but that would result in the `CustomerInfoDto.owner` entity in being serialized... which could expose data and can result in a circular reference if that entity has more relationships where both sides are mapped.

### Proposed Solution

Introduce a new "mode" where we "cascade" the (de)hydration process onto the DTO.

```diff
CustomerDetails
{
-    #[LiveProp]
+    #[LiveProp(cascadeHydration: true)]
    public CustomerInfoDto $customerInfoDto;
}
```

With this set, live components WOULD dehydrate `CustomInfoDto` using the same logic it uses for LiveProp. Specifically, it would:

A) find all properties that are both readable & writable inside `CustomerInfoDto` and (de)hydrate these (using the property access component)
B) For (de)hydration, read the property type, like with a normal `LiveProp`.

There will likely be some edge-cases that we wouldn't handle on purpose - e.g. there would not be any way to "exclude" a property from being (de)hydrated. We don't want to reinvent the serializer.

WDYT - helpful? Should this, perhaps, be the default mode for `LiveProp` objects?

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #955
| License       | MIT

With this PR you van easily use DTO with your LiveComponents.

```php
class CustomerDetails
{
    public string $name;
    public Address $address;
    public string $city;
}
```

```php
class Address
{
    public string $street;
    public string $postCode;
}
```

```php
#[AsLiveComponent(name: 'CustomerDetails')]
class CustomerDetailsComponent
{
    use DefaultActionTrait;

    #[ExposeInTemplate]
    public string $hello = 'hello';

    #[LiveProp(writable: true)]
    public ?CustomerDetails $customerDetails = null;

    public function mount(): void
    {
        $this->customerDetails = new CustomerDetails();
        $this->customerDetails->name = 'Matheo';
        $this->customerDetails->city = 'Paris';
        $this->customerDetails->address = new Address();
        $this->customerDetails->address->street = '3 rue de la Paix';
        $this->customerDetails->address->postCode = '92270';
    }

    #[LiveAction]
    public function switch(): void
    {
        $this->customerDetails = new CustomerDetails();
        $this->customerDetails->name = 'Paul';
        $this->customerDetails->city = 'Paris';
        $this->customerDetails->address = new Address();
        $this->customerDetails->address->street = '3 rue des mimosas';
        $this->customerDetails->address->postCode = '92270';
    }
}
```

```twig
<div {{ attributes }}>
    <p>{{ customerDetails.name }}</p>
    <p>{{ customerDetails.address.street }}</p>
    <button
            data-action="live#action"
            data-action-name="switch"
    >Switch</button>
</div>
```

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -242,7 +242,7 @@ LiveProp Data Types
 
 LiveProps must be a value that can be sent to JavaScript. Supported values
 are scalars (int, float, string, bool, null), arrays (of scalar values), enums,
-DateTime objects, Doctrine entity objects, DTO, or array of DTO.
+DateTime objects, Doctrine entity objects, DTOs, or array of DTOs.
 
 See :ref:`hydration` for handling more complex data.
 
@@ -622,25 +622,21 @@ Note that being able to change the "identity" of an object is something
 that works only for objects that are dehydrated to a scalar value (like
 persisted entities, which dehydrate to an ``id``).
 
-Hydration, DTO's & the Serializer
+Using DTO's on a LiveProp
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. versionadded:: 2.11
 
-    Dto are now a supported type for LiveComponent
+    The automatic (de)hydration of DTO objects was introduced in LiveComponents 2.11.
 
-To use a ``LiveProp`` with DTO simply type properly your property
-
-.. code-block:: php
+You can also use a DTO (i.e. data transfer object / any simple class) with LiveProp as long as the property has the correct type::
 
     class ComponentWithAddressDto
     {
         public AddressDto $addressDto;
     }
 
-And to work with DTO collection, you simply need to specified the collection type in your phpdoc
-
-.. code-block:: php
+To work with a collection of DTOs, specify the collection type inside PHPDoc::
 
     class ComponentWithAddressDto
     {
@@ -650,11 +646,18 @@ And to work with DTO collection, you simply need to specified the collection typ
         public array $addressDtoCollection;
     }
 
+Here is how the (de)hydration of DTO objects works:
+
+- It finds all properties on your DTO that are readable and writable and dehydrates each one.
+- the PropertyAccess component is used, which means getter and setter methods are supported, in addition to public properties.
+- The DTO cannot have any constructor arguments.
+
 If this solution doesn't feat your need
+
 there are two others options to make this work:
 
-1) Hydrating with the Serializer
-................................
+Hydrating with the Serializer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. versionadded:: 2.8
 
@@ -671,8 +674,8 @@ option::
 
 You can also set a ``serializationContext`` option on the ``LiveProp``.
 
-2) Hydrating with Methods: hydrateWith & dehydrateWith
-......................................................
+Hydrating with Methods: hydrateWith & dehydrateWith
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You can take full control of the hydration process by setting the ``hydrateWith``
 and ``dehydrateWith`` options on ``LiveProp``::
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -242,7 +242,7 @@ LiveProp Data Types
 
 LiveProps must be a value that can be sent to JavaScript. Supported values
 are scalars (int, float, string, bool, null), arrays (of scalar values), enums,
-DateTime objects & Doctrine entity objects.
+DateTime objects, Doctrine entity objects, DTO, or array of DTO.
 
 See :ref:`hydration` for handling more complex data.
 
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
 Hydration, DTO's & the Serializer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you try to use a ``LiveProp`` for some unsupported type (e.g.a DTO object),
-it will fail. A best practice is to use simple data.
+.. versionadded:: 2.11
 
-But there are two options to make this work:
+    Dto are now a supported type for LiveComponent
+
+To use a ``LiveProp`` with DTO simply type properly your property
+
+.. code-block:: php
+
+    class ComponentWithAddressDto
+    {
+        public AddressDto $addressDto;
+    }
+
+And to work with DTO collection, you simply need to specified the collection type in your phpdoc
+
+.. code-block:: php
+
+    class ComponentWithAddressDto
+    {
+        /**
+         * @var AddressDto[]
+        /*
+        public array $addressDtoCollection;
+    }
+
+If this solution doesn't feat your need
+there are two others options to make this work:
 
 1) Hydrating with the Serializer
 ................................
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -623,7 +623,7 @@ that works only for objects that are dehydrated to a scalar value (like
 persisted entities, which dehydrate to an ``id``).
 
 Using DTO's on a LiveProp
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. versionadded:: 2.11
 
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -16,6 +16,8 @@ use Symfony\Component\PropertyAccess\Exception\ExceptionInterface as PropertyAcc
 use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\Exception\UninitializedPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
+use Symfony\Component\PropertyInfo\PropertyInfoExtractor;
 use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
@@ -358,14 +360,14 @@ final class LiveComponentHydrator
             if (!$this->isValueValidDehydratedValue($value)) {
                 $badKeys = $this->getNonScalarKeys($value, $propMetadata->getName());
                 $badKeysText = implode(', ', array_map(fn ($key) => sprintf('%s: %s', $key, $badKeys[$key]), array_keys($badKeys)));
-                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" is an array, but it contains one or more keys that are not scalars: %s', $propMetadata->getName(), $parentObject::class, $badKeysText));
+                throw new \LogicException(throw new \LogicException(sprintf('Unable to dehydrate value of type "%s" for property "%s" on component "%s". Change this to a simpler type of an object that can be dehydrated. Or set the hydrateWith/dehydrateWith options in LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer.', get_debug_type($value), $propMetadata->getName(), $parentObject::class)));
             }
 
             return $value;
         }
 
         if (!\is_object($value)) {
-            throw new \LogicException(sprintf('The "%s" object has a hydrateMethod of "%s" but the method does not exist.', $parentObject::class, $propMetadata->hydrateMethod()));
+            throw new \LogicException(sprintf('Unable to dehydrate value of type "%s" for property "%s" on component "%s". Change this to a simpler type of an object that can be dehydrated. Or set the hydrateWith/dehydrateWith options in LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer.', get_debug_type($value), $propMetadata->getName(), $parentObject::class));
         }
 
         if (!$propMetadata->getType() || $propMetadata->isBuiltIn()) {
@@ -395,10 +397,10 @@ final class LiveComponentHydrator
         }
 
         $dehydratedObjectValues = [];
-        foreach ((new \ReflectionClass($classType))->getProperties() as $property) {
-            $propertyValue = $this->propertyAccessor->getValue($value, $property->getName());
-            $propMetadata = $this->liveComponentMetadataFactory->createLivePropMetadata($classType, $property->getName(), $property, new LiveProp());
-            $dehydratedObjectValues[$property->getName()] = $this->dehydrateValue($propertyValue, $propMetadata, $parentObject);
+        foreach ((new PropertyInfoExtractor([new ReflectionExtractor()]))->getProperties($classType) as $property) {
+            $propertyValue = $this->propertyAccessor->getValue($value, $property);
+            $propMetadata = $this->liveComponentMetadataFactory->createLivePropMetadata($classType, $property, new \ReflectionProperty($classType, $property), new LiveProp());
+            $dehydratedObjectValues[$property] = $this->dehydrateValue($propertyValue, $propMetadata, $parentObject);
         }
 
         return $dehydratedObjectValues;
@@ -479,8 +481,8 @@ final class LiveComponentHydrator
         if (\is_array($value)) {
             $object = new $className();
             foreach ($value as $propertyName => $propertyValue) {
-                $reflexionClass = new \ReflectionClass($className);
-                $property = $reflexionClass->getProperty($propertyName);
+                $reflectionClass = new \ReflectionClass($className);
+                $property = $reflectionClass->getProperty($propertyName);
                 $propMetadata = $this->liveComponentMetadataFactory->createLivePropMetadata($className, $propertyName, $property, new LiveProp());
                 $this->propertyAccessor->setValue($object, $propertyName, $this->hydrateValue($propertyValue, $propMetadata, $component));
             }
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -16,7 +16,7 @@ use Symfony\Component\PropertyAccess\Exception\ExceptionInterface as PropertyAcc
 use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\Exception\UninitializedPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
-use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
+use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
 use Symfony\Component\PropertyInfo\Type;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
@@ -47,6 +47,7 @@ final class LiveComponentHydrator
     public function __construct(
         private iterable $hydrationExtensions,
         private PropertyAccessorInterface $propertyAccessor,
+        private PropertyTypeExtractorInterface $propertyTypeExtractor,
         private NormalizerInterface|DenormalizerInterface $normalizer,
         private string $secret
     ) {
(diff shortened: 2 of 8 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -392,7 +394,22 @@ private function dehydrateObjectValue(object $value, string $classType, ?string
...
+            if ($reflexionExtractor->isReadable($classType, $property)) {
+                $propertyValue = $this->propertyAccessor->getValue($value, $property);
+                $type = $reflexionExtractor->getTypes($classType, $property)[0]->getBuiltinType();
+                if ($type === 'object') {
+                    $type = $reflexionExtractor->getTypes($classType, $property)[0]->getClassName();
+                }
+                $propMetadata = new LivePropMetadata($property, new LiveProp(true), $type, false, true, null);
+                $propertiesValues[$property] = $this->dehydrateValue($propertyValue, $propMetadata, $component);
+            }
+        }
```
weaverryan, line 404: Yes, I think this is definitely what I had in mind!!! Details:

* A) What about support for collections / arrays? So, something like this should also be allowed to work in the DTO

```php
class CustomerDetails
{
    public string $name;
    /** @var Address[] */
    public array $addresses;
    public string $city;
}
```

This type of thing DOES work on the top-level component (e.g. you could have a `LiveProp` that is an array of entity objects). 

* B) Your extracting the "types" and properties using the `ReflectionExtractor`. Can we use the PropertyInfo component services instead? That's especially important for the types: we need to use the same logic for determining the types in the DTOs as we do on the top-level `LiveProp`s - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php#L62-L84

Ideally, we would not only match that, but centralize it somehow: whenever we need a "type" for any property that we (de)hydrate, we use the same code.

weaverryan:
Fantastic!!!

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -559,4 +572,35 @@ private function recursiveKeySort(array &$data): void
...
+
+        return new LivePropMetadata(
+            $property->getName(),
+            new LiveProp(true),
+            $type ? $type->getName() : null,
+            $type ? $type->isBuiltin() : false,
+            $type ? $type->allowsNull() : true,
+            $collectionValueType,
+        );
+    }
```
weaverryan: We should centralize this - as it's repeated in https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -341,22 +344,15 @@ private function dehydrateValue(mixed $value, LivePropMetadata $propMetadata, ob
...
+                $type = \gettype($objectItem);
+                if ('object' === $type) {
+                    $type = $objectItem::class;
                 }
-            }
 
-            if (!$this->isValueValidDehydratedValue($value)) {
-                $badKeys = $this->getNonScalarKeys($value, $propMetadata->getName());
-                $badKeysText = implode(', ', array_map(fn ($key) => sprintf('%s: %s', $key, $badKeys[$key]), array_keys($badKeys)));
+                $propMetadata = new LivePropMetadata($key, new LiveProp(true), $type, false, true, null);
```
weaverryan: Any reason for this change and the changes above? We already have a `$propMetadata` variable - is it not correct for some reason?

WebMamba: The goal here is for each item of the array to go through the dehydrate method. I did that to support array of DTOs

weaverryan: That makes sense - but I think the old code already did that. The main difference between the old and new code just seems to be that the old called `->dehydrateObjectValue()` on each item and the new code calls `->dehydrateValue()`.

But there is an important detail in the original code: to determine the "type" of each item in the array, we should NOT use `$objectItem::class`. Instead, we need to read the "property info" off of the property. Otherwise, we may allow for a value to be dehydrated that won't be able to be rehydrated later. Example:

```
class Foo
{
    public array $dtos1 = [];

    /** @var Dto[] */
    public array $dtos2 = [];
}
```

```php
$foo = new Foo();
$foo->dtos1 = [new Dto()];
$foo->dtos2 = [new Dto()];
```

In this situation, the `dtos1` property is *invalid*: we will *not* be able to hydrate this. We know that `dtos1` holds an array, but an array of what? That is the purpose of the `$propMetadata->collectionValueType()->getClassName();`: it tells us what the "metadata" (e.g. phpdoc) tells us the "type" of each item in the collection is. If that's missing, we COULD still successfully dehydrate, but hydration would fail. So, instead, we fail during dehydration to warn the user early.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -348,14 +351,13 @@ private function dehydrateValue(mixed $value, LivePropMetadata $propMetadata, ob
                         throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" is an array. We determined the array is full of %s objects, but at least on key had a different value of %s', $propMetadata->getName(), $component::class, $collectionClass, get_debug_type($objectItem)));
```
weaverryan: This `dehydrateValue()` method is now called both for top-level `LiveProp` on the component object, but it's also called for properties on any nested DTO's. Let's rename the `$component` argument to `$parentObject`. Same in `dehydrateObjectValue()`. Some of the exception messages that use the `$component` variable may also need to be tweaked: some reference `LiveProp`. If the `$parentObject` IS actually the top-level `$component`, then this message is still accurate. Otherwise, it needs to be tweaked to be more clear.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -392,7 +394,14 @@ private function dehydrateObjectValue(object $value, string $classType, ?string
             }
         }
 
-        throw new \LogicException(sprintf('Unable to dehydrate value of type "%s" for property "%s" on component "%s". Either (1) change this to a simpler value, (2) add the hydrateWith/dehydrateWith options to LiveProp or (3) set "useSerializerForHydration: true" on the LiveProp.', $value::class, $propertyPathForError, $componentClassForError));
+        $hydratedObject = [];
+        foreach ((new \ReflectionClass($classType))->getProperties() as $property) {
```
weaverryan: I think we should use https://symfony.com/doc/current/components/property_info.html#list-information instead of `ReflectionClass::getProperties()`. It'll let us also support private properties with getters & setters (I believe the property info will only return public properties + private properties with a getter).

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -392,7 +394,14 @@ private function dehydrateObjectValue(object $value, string $classType, ?string
             }
         }
 
-        throw new \LogicException(sprintf('Unable to dehydrate value of type "%s" for property "%s" on component "%s". Either (1) change this to a simpler value, (2) add the hydrateWith/dehydrateWith options to LiveProp or (3) set "useSerializerForHydration: true" on the LiveProp.', $value::class, $propertyPathForError, $componentClassForError));
+        $hydratedObject = [];
```
weaverryan: A better name would be `$dehydratedObjectValues`, right?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -59,32 +59,38 @@ public function createPropMetadatas(\ReflectionClass $class): array
...
-                        $collectionValueType = $valueType;
-                        break;
-                    }
+            $metadatas[$property->getName()] = $this->createLivePropMetadata($class->getName(), $property->getName(), $property, $attribute);
+        }
+
+        return array_values($metadatas);
+    }
+
+    public function createLivePropMetadata(string $className, string $propertyName, \ReflectionProperty $property, \ReflectionAttribute $attribute = null): LivePropMetadata
```
weaverryan: Minor: let's make `\ReflectionAttribute` argument a `LiveProp` argument and make it required. The method above can call `$attribute->newInstance()` and `LiveComponentHydrator` can call `new LiveProp()`.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -59,32 +59,38 @@ public function createPropMetadatas(\ReflectionClass $class): array
...
-            );
+        $type = $property->getType();
+        if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
+            throw new \LogicException(sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
         }
 
-        return array_values($metadatas);
+        return new LivePropMetadata(
+            $property->getName(),
+            null !== $attribute ? $attribute->newInstance() : new LiveProp(true),
```
weaverryan: When this is moved to `LiveComponentHydrator`, I'm pretty sure this should become `new LiveProp()` with no args: writable should be `false`. If a DTO is on a `LiveProp`, then we are trying to dehydrate/rehydrate it just so we can get the same DTO back on the next request: the individual properties of that DTO are not being written. The user CAN have `LiveProp(writable: ['somePropOnDto'])`, but that is already handled elsewhere.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -392,7 +394,14 @@ private function dehydrateObjectValue(object $value, string $classType, ?string
...
+        foreach ((new \ReflectionClass($classType))->getProperties() as $property) {
+            $propertyValue = $this->propertyAccessor->getValue($value, $property->getName());
+            $propMetadata = $this->liveComponentMetadataFactory->createLivePropMetadata($classType, $property->getName(), $property);
```
weaverryan: This is cool! When a DTO is being (de)hydrated, it's effectively like each accessible property is given an automatic `#[LiveProp]`. That makes it easy to think about how all of this should work.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -467,6 +476,18 @@ private function hydrateObjectValue(mixed $value, string $className, bool $allow
             }
         }
 
+        if (\is_array($value)) {
+            $object = new $className();
+            foreach ($value as $propertyName => $propertyValue) {
+                $reflexionClass = new \ReflectionClass($className);
+                $property = $reflexionClass->getProperty($propertyName);
+                $propMetadata = $this->liveComponentMetadataFactory->createLivePropMetadata($className, $propertyName, $property);
```
weaverryan: We could probably share this with the dehydration code in a new private method - something like:

```php
/** @return LivePropMetadata[] */
private function getLivePropsMetadataForClass(string $className): array
```

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -467,6 +476,18 @@ private function hydrateObjectValue(mixed $value, string $className, bool $allow
...
+                $this->propertyAccessor->setValue($object, $propertyName, $this->hydrateValue($propertyValue, $propMetadata, $component));
+            }
+
+            return $object;
+        }
+
         throw new HydrationException(sprintf('Unable to hydrate value of type "%s" for property "%s" on component "%s". Change this to a simpler value, add the hydrateWith/dehydrateWith options to LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer..', $className, $propertyPathForError, $componentClassForError));
```
weaverryan: Does this error need to be adjusted now? I'm trying to think of WHEN this would happen. By the time you get to this method, we know that the value is being hydrated onto a property that is typed with a class. If you are not an enum, or one of the other situations and `$value` is not an array... it seems like something went very wrong. Perhaps we should make this error more generic.

src/LiveComponent/tests/Fixtures/Kernel.php
```diff
@@ -90,6 +90,7 @@ protected function configureContainer(ContainerConfigurator $c): void
             'secrets' => false,
             'session' => ['storage_factory_id' => 'session.storage.factory.mock_file'],
             'http_method_override' => false,
+            'property_info' => ['enabled' => true],
```
weaverryan, line 93: This was needed?

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -783,6 +785,109 @@ public function provideDehydrationHydrationTests(): iterable
...
+                    $this->address->city = 'Paris';
+                }
+            })
+                ->mountWith([])
+                ->assertDehydratesTo([
+                    'address' => [
+                        'address' => '1 rue du Bac',
+                        'city' => 'Paris',
+                    ],
+                    '@checksum' => '2P/KES0BbGw1KNZsOn4CSEpN1HJ1s...+bQPY=',
```
weaverryan: Not needed: the code that runs the test adds the correct `@checksum` for you. If you remove this, test still passes :)

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -783,6 +785,109 @@ public function provideDehydrationHydrationTests(): iterable
...
+                ])
+                ->userUpdatesProps(['address' => ['address' => '4 rue des lilas', 'city' => 'Asnieres']])
+                ->assertObjectAfterHydration(function (object $object) {
+                    $this->assertSame($object->address->address, '4 rue des lilas');
+                    $this->assertSame($object->address->city, 'Asnieres');
+                })
+            ;
+        }];
```
weaverryan, line 813: Awesome test!

weaverryan (commented):
WOW. We're down to the last details. Looking at your test cases got me really excited though... :D. Can you also add some documentation? Thank you!

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
WebMamba: Haha thanks!

src/LiveComponent/tests/Fixtures/Kernel.php
WebMamba: Yes otherwise we don't have access to the phpdoc info

src/LiveComponent/src/LiveComponentHydrator.php
WebMamba: Hum this not exactly the same logic I feel like it's gonna introduce more complexity

WebMamba:
Thanks, Ryan for your review! I just gonna need some help for the docs and the errors messages 😁

src/LiveComponent/doc/index.rst
```diff
@@ -242,7 +242,7 @@ LiveProp Data Types
 
 LiveProps must be a value that can be sent to JavaScript. Supported values
 are scalars (int, float, string, bool, null), arrays (of scalar values), enums,
-DateTime objects & Doctrine entity objects.
+DateTime objects, Doctrine entity objects, DTO, or array of DTO.
```
```suggestion
DateTime objects, Doctrine entity objects, DTOs, or array of DTOs.
```

src/LiveComponent/doc/index.rst
```diff
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
 Hydration, DTO's & the Serializer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you try to use a ``LiveProp`` for some unsupported type (e.g.a DTO object),
-it will fail. A best practice is to use simple data.
+.. versionadded:: 2.11
+
+    Dto are now a supported type for LiveComponent
```
```suggestion
    The automatic (de)hydration of DTO objects was introduced in LiveComponents 2.11.
```

src/LiveComponent/doc/index.rst
```diff
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
 Hydration, DTO's & the Serializer
```
weaverryan: Let's rename this section to:

```
Using DTO's on a LiveProp
```

Then, below:

* Change `1) Hydrating with the Serializer` to `Hydrating with the Serializer` and change the `...` below it to be `~~~~` so it is a sibling of this section.
* Do the same for `2) Hydrating with Methods: hydrateWith & dehydrateWith`: remove the `2)` and change `...` to `~~~`

src/LiveComponent/doc/index.rst
```diff
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you try to use a ``LiveProp`` for some unsupported type (e.g.a DTO object),
-it will fail. A best practice is to use simple data.
+.. versionadded:: 2.11
+
+    Dto are now a supported type for LiveComponent
+
+To use a ``LiveProp`` with DTO simply type properly your property
```
weaverryan: You can also use a DTO (i.e. data transfer object / any simple class) with `LiveProp` as long as the property has the correct type::

src/LiveComponent/doc/index.rst
```diff
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
...
+
+.. code-block:: php
```
weaverryan: By using `::` on the line above, you won't need this here

src/LiveComponent/doc/index.rst
```diff
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
...
+
+    class ComponentWithAddressDto
+    {
+        public AddressDto $addressDto;
+    }
+
+And to work with DTO collection, you simply need to specified the collection type in your phpdoc
```
weaverryan: To work with a collection of DTOs, specify the collection type inside PHPDoc::

src/LiveComponent/doc/index.rst
```diff
@@ -625,10 +625,33 @@ persisted entities, which dehydrate to an ``id``).
...
+    {
+        /**
+         * @var AddressDto[]
+        /*
+        public array $addressDtoCollection;
+    }
 
-But there are two options to make this work:
+If this solution doesn't feat your need
+there are two others options to make this work:
```
weaverryan, line 657: Before we finish with these last two lines, we should explain briefly how the DTO system works (please correct any inaccuracies in my statement below)

* It finds all properties on your DTO that are readable and writable and dehydrates each one.
* the PropertyAccess component is used, which means getter and setter methods are supported, in addition to public properties.
* The DTO cannot have any constructor arguments.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -345,38 +348,37 @@ private function dehydrateValue(mixed $value, LivePropMetadata $propMetadata, ob
...
-                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" is an array, but it contains one or more keys that are not scalars: %s', $propMetadata->getName(), $component::class, $badKeysText));
+                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" is an array, but it contains one or more keys that are not scalars: %s', $propMetadata->getName(), $parentObject::class, $badKeysText));
             }
 
             return $value;
         }
 
         if (!\is_object($value)) {
-            throw new \LogicException(sprintf('Unable to dehydrate value of type "%s" for property "%s" on component "%s". Change this to a simpler type of an object that can be dehydrated. Or set the hydrateWith/dehydrateWith options in LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer.', get_debug_type($value), $propMetadata->getName(), $component::class));
+            throw new \LogicException(sprintf('The "%s" object has a hydrateMethod of "%s" but the method does not exist.', $parentObject::class, $propMetadata->hydrateMethod()));
```
weaverryan: This change doesn't look right.

src/LiveComponent/src/LiveComponentHydrator.php
weaverryan: I think this isn't resolved yet. We're using processAccessor to read the values, but not property info to list the properties.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -467,7 +476,19 @@ private function hydrateObjectValue(mixed $value, string $className, bool $allow
             }
         }
 
-        throw new HydrationException(sprintf('Unable to hydrate value of type "%s" for property "%s" on component "%s". Change this to a simpler value, add the hydrateWith/dehydrateWith options to LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer..', $className, $propertyPathForError, $componentClassForError));
+        if (\is_array($value)) {
+            $object = new $className();
+            foreach ($value as $propertyName => $propertyValue) {
+                $reflexionClass = new \ReflectionClass($className);
```
weaverryan:
```suggestion
                $reflectionClass = new \ReflectionClass($className);
```

weaverryan:
THANK YOU Matheo! This works AWESOME - it's a dream. I put some comments about the docs, but they are so minor, I will make them in a different PR.

PR #1093, opened 2023-09-04
[TwigComponent][LiveComponent] Fix DataModelPropsSubscriber for embedded components

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | 
| License       | MIT

### Reproduce

```
{# components/A.html.twig #}

<B>
  {# content doesn't matter, as long as it's an embedded component #}
</B>
```
```
{# components/B.html.twig #}

<C data-model="X"/>
```
(given a component B with a prop X)

With this setup the `DataModelPropsSubscriber::preMount()`  would result in an error "_Can't get a way to read the property "X" in class "A"_".

An even simpler setup

```
{# anyTemplate.html.twig #}

<A>
{# A is already embedded #}
</A>
```
Will result in "_You can only pass "data-model" when rendering a component when you're rendering inside of a parent component._"
Whether the data-model attribute is deep inside template B or already in template A doesn't matter. The thing is that during rendering of A, which is now embedded, there's no component on the stack.

### Problem

Self-closing component are rendered via `ComponentRenderer::createAndRender()`.
Embedded components are rendered via `ComponentRenderer::embeddedContext()` + `Template::display()`.

Both push the component being rendered on the `ComponentStack`, and pop it again at the end of the `ComponentRender` functions. During the rendering of C, B is already been popped from the stack, and therefore the `DataModelPropsSubscriber` doesn't have a way to find the parent component context. And it incorrectly uses component A, which is still on the component stack (while it should be linked to B).

### Solution
The rendering of an embedded component is only really done after the `Template::display()` execution. Therefore the component can only be removed when BOTH are done.

### DIFF
--- a/src/LiveComponent/src/EventListener/DataModelPropsSubscriber.php
+++ b/src/LiveComponent/src/EventListener/DataModelPropsSubscriber.php
@@ -13,9 +13,11 @@ namespace Symfony\UX\LiveComponent\EventListener;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\LiveComponent\Util\ModelBindingParser;
 use Symfony\UX\TwigComponent\ComponentStack;
 use Symfony\UX\TwigComponent\Event\PreMountEvent;
+use Symfony\UX\TwigComponent\MountedComponent;
 
 /**
  * Parses the "data-model" key, which triggers extra props to be passed in.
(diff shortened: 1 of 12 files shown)


sneakyvv:
Triggered a test re-run via force-push.
There were a lot of checks failing with '_Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_7abfb70c-3c2b-4efa-88de-68305c3e9cfc/de1fe6ea-d403-4c00-86a3-50ae693470fa.tar.gz. return code: 2._'

src/LiveComponent/src/EventListener/DataModelPropsSubscriber.php
```diff
@@ -76,4 +78,19 @@ public static function getSubscribedEvents(): array
             PreMountEvent::class => 'onPreMount',
         ];
     }
+
+    private function getCurrentLiveComponent(ComponentStack $componentStack): ?MountedComponent
+    {
+        $iterator = $componentStack->getIterator();
+        while (($current = $iterator->current()) && !$this->isLiveComponent($current->getComponent()::class)) {
```
weaverryan: Because it now implements `IteratorAggregate`, can we just `foreach ($componentStack as $component)`?

Also, can you explain this a little? It's possible that the current component is some *other*, non-live component? And so we need to look further "up"?

weaverryan (commented):
COMPLEX - but it makes sense, and an excellent test to cover it.

src/LiveComponent/src/EventListener/DataModelPropsSubscriber.php
sneakyvv: That's right. If you have for example a, I don't know, Card component as a parent of an Input component, all inside of a Product component, and the input is supposed to control a property of the Product. In HTML this will already be ok, since the product-HTML will have the data-live properties, and the card won't. So, here in PHP, we also need to link it to that Live Component.

I adjusted the comment that was already there.

weaverryan:
Thank you @sneakyvv!

PR #1096, opened 2023-09-06
Don't store ux_live_component URLs in setTargetPath of the security component

### LINKED ISSUE #1095: Symfony security authenticator targetPath saves ux_live_component URL
When expire authentication session and you want to redirect to the previous URL using onAuthenticationSuccess->getTargetPath, if the last request was an ux_live_component route. The browser downloads a file with:

```html
<div  data-controller="live">
...
</div>
```

And keeps the login URL

gbere:
Solves issue #1095

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #1095
| License       | MIT

Prevent to redirect to ux_live_component URL after login

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -2154,6 +2154,7 @@ class RequestBuilder {
         const fetchOptions = {};
         fetchOptions.headers = {
             Accept: 'application/vnd.live-component+html',
+            'X-Requested-With': 'XMLHttpRequest',
         };
         const totalFiles = Object.entries(files).reduce((total, current) => total + current.length, 0);
         const hasFingerprints = Object.keys(children).length > 0;
--- a/src/LiveComponent/assets/src/Backend/RequestBuilder.ts
+++ b/src/LiveComponent/assets/src/Backend/RequestBuilder.ts
@@ -25,6 +25,7 @@ export default class {
         const fetchOptions: RequestInit = {};
         fetchOptions.headers = {
             Accept: 'application/vnd.live-component+html',
+            'X-Requested-With': 'XMLHttpRequest',
         };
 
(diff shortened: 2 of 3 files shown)


src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -2154,6 +2154,7 @@ class RequestBuilder {
         const fetchOptions = {};
         fetchOptions.headers = {
             Accept: 'application/vnd.live-component+html',
+            'X-Requested-With': 'XMLHttpRequest',
```
WebMamba, line 2157: Why the value `XMLHttpRequest'`?

gbere: > Why the value `XMLHttpRequest'`?

Set the request like a AJAX request. It's the only way I found so that it does not store it in the [setTargetPath](https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/Security/Http/Firewall/ExceptionListener.php#L223)

WebMamba: Yes but setTargetPath look only if the XMLHttpRequest is present but doesn't look at his value at all: https://github.com/symfony/symfony/blob/2bac801f8a4cba02e8ecc840fa7bf512034f860a/src/Symfony/Component/HttpFoundation/Request.php#L1714.
The `X-Requested-With` is not an official header and I can't find official documentation about it, but it look like you can set the value you want. So we can have something like:
`'X-Requested-With': 'symfony-ux-live-component',`
What do you think?

gbere: I prefer to keep the one that already exists. This way you don't need to modify any other component.

weaverryan:
Thanks Gerard!

PR #1097, opened 2023-09-06
[LiveComponent] Only consider Live components in InterceptChildComponentRenderSubscriber

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | 
| License       | MIT

This PR is related to the change of #1093.

### Problem
The component stack now also contains embedded Components until they are fully rendered, which is good. But, that also means that with this setup

```
{# someTemplate #}

<twig:aLiveComponent>
    <twig:aTwigComponent>
        <twig:anotherLiveComponent>
            {# ... #}
        </twig:anotherLiveComponent>
    </twig:aTwigComponent>
</twig:aLiveComponent>
```
when re-rendering the `anotherLiveComponent` the `InterceptChildComponentRenderSubscriber` would look at the parent component's `childrenFingerprints` to determine if it can short-circuit the rendering process for `anotherLiveComponent` and return an empty html element causing the FE to not do a callback to re-render it. However since that parent is a non-live component it will never short-circuit.

### Solution

Only take live components into account.


weaverryan:
Thanks Bart!

PR #1099, opened 2023-09-06
[LiveComponent] Dependent Form Fields documentation

smbpunt:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| License       | MIT

Hi, 

I followed the documentation and I had trouble using the same form when editing my object.

Since the third attributes `$data` of `createNamedBuilder()` is `@param mixed $data The initial data`,
should we not give him a Food instead of meal ?
Check my PR.

-----

In order to handle edition using the same form, I also edited the LiveComponent. 
With this, to call the component we need to pass attributes (but it's not on this page of the documentation).
```twig
{{ component('MealPlanner', {
    form: form,
    mealPlan: mealPlan.id ? mealPlan : null,
}) }}
```
Imho, this might be a good idea to add it to the documentation too (I found `mealPlan: mealPlan.id ? mealPlan : null` in an issue...)

Thanks

### DIFF
--- a/ux.symfony.com/src/Form/MealPlannerForm.php
+++ b/ux.symfony.com/src/Form/MealPlannerForm.php
@@ -53,7 +53,7 @@ class MealPlannerForm extends AbstractType
         /** @var ?MealPlan $data */
         $data = $event->getData();
 
-        $this->addFoodField($event->getForm(), $data?->getMeal());
+        $this->addFoodField($event->getForm(), $data?->getMeal(), $data?->getFood());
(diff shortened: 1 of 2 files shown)


weaverryan:
Hi there!

Apologies for the slow reply on your PR :). Actually, it's a nice change - but I had, by coincidence, just 1 day after this PR, refactored that class to use `symfonycasts/dynamic-forms`. Hopefully I've not introduced any new mistakes in that PR ;).

Thanks!

PR #1111, opened 2023-09-13
[Site] Avoid sass files from being exported publicly

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | None
| License       | MIT

This doesn't matter much for an OSS project anyway, but this prevents the `.scss` files from being exported to the `public/assets` directory on deploy. 

Cheers!


Kocal:
Could it be configured in the recipe? Or this is too much app-specific?

weaverryan:
That seems like a good idea - files starting with _ I believe should never be “main” sass files that you build. Can you open a PR?

PR #1113, opened 2023-09-13
fix(live): fix testing events

### LINKED ISSUE #1110: [LiveComponent] 2.11 - Testing an emit fails because of the event name being used for the method
I am implementing the new test helpers in our project and so far all is working well but one thing. Maybe I am misinterpreting the way it should work but the docs state that you can test your component with calling an emit with one of the following two syntaxes:

```php
// call live actions
$testComponent
    ->call('increase')
    ->call('increase', ['amount' => 2]) // call a live action with arguments
;
``` 

I have a component with a `#[LiveListener]` listening to an event, like so:

```php
#[LiveListener('CartUpdated')]
public function update(#[LiveArg] float $balance): void
{
    $this->balance = $balance;
}
```

So in my assumption I can test the above by calling the following in my test:

```php
$testComponent
    ->emit('CartUpdated', ['balance' => $balance])
;
```

However, this results in an error message stating the following: 
**Expected method "CartUpdated" on class "App\Components\CartSummaryMobile"**

Looking in the `TestLiveComponent` class in the UX package it uses the eventName as the method to call on the component it is testing. Otherwise, what would be the point of emitting an event? If it is using the method name I could have used the `->call` method for executing a LiveAction I guess?

```
/**
* @param array<string,mixed> $arguments
*/
public function emit(string $event, array $arguments = []): self
{
    return $this->call($event, $arguments);
}
```

In my assumption those two do not have to be the same right? If I am incorrect, then please let me know.

Thanks in advance.

Robbert

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #1110
| License       | MIT

When implementing #823, I made the wrong assumption that events were identical to actions. This fixes that.

### DIFF
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -15,6 +15,7 @@ use Symfony\Bundle\FrameworkBundle\KernelBrowser;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
 use Symfony\UX\TwigComponent\ComponentFactory;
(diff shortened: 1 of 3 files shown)


src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -89,7 +90,20 @@ public function call(string $action, array $arguments = []): self
...
+            if ($listener['event'] === $event) {
+                $actions[] = ['name' => $listener['action'], 'args' => $arguments];
+            }
+        }
+
+        if (!$actions) {
+            throw new \InvalidArgumentException(sprintf('Event "%s" does not exist on component "%s".', $event, $this->metadata->getName()));
+        }
+
+        return $this->request(['actions' => $actions], '_batch');
```
weaverryan, line 121: Super minor, and it shouldn't make any difference, but if we have just one action, it won't process through batch: it'll just call that one action directly.

kbond: The js makes this distinction?

weaverryan: Exactly. The js always tries to use the normal endpoint (either for calling a `LiveAction` or the "default" action in the case of a re-render). If it finds that it has 2 things queued at once, then it goes through batch. https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/src/Backend/RequestBuilder.ts#L71-L79

kbond: Ok, I'll match in the helper.

kbond: This change has been made.

weaverryan:
Thanks Kevin!

PR #1114, opened 2023-09-14
[LiveComponents] OnUpdate hook for LiveProp

bocharsky-bw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | 
| License       | MIT

Feature allows to add a hook on LiveProp changed, providing the "previous value" for convenience:

```php
    #[LiveProp(writable: true, onUpdated: 'onTitleUpdated')]
    public string $title = 'Test';

    #[LiveProp(writable: ['customerName', 'customerEmail', 'taxRate'], onUpdated: ['customerName' => 'onCustomerNameUpdated'])]
    public Invoice $invoice;

    #[LiveProp(writable: LiveProp::IDENTITY, onUpdated: [LiveProp::IDENTITY => 'onEntireInvoiceChange'])]
    public Invoice $invoice;

    public function onTitleUpdated($oldValue)
    {
        // ...
    }

    public function onCustomerNameUpdated($oldValue)
    {
        // ...
    }

    public function onEntireInvoiceChange($oldValue)
    {
        // ...
    }
```

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## ...
+## 2.12.0
 
 -   Add `onUpdated` hook for `LiveProp`
 
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## ...
+
+-   Add `onUpdated` hook for `LiveProp`
+
 ## 2.11.0
 
 -   Add helper for testing live components.
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3070,9 +3070,9 @@ Add a Hook on LiveProp Update
 
     The ``onUpdated`` option was added in LiveComponents 2.12.
 
-If you want to run a custom code after a specific LiveProp was updated,
-you can do it by adding a ``onUpdated`` option. Specify a method name
-that will be called. The method should be public and created on the Component:
+If you want to run custom code after a specific LiveProp is updated,
+you can do it by adding an ``onUpdated`` option set to a public method name
+on the component:
 
 .. code-block:: diff
 
@@ -3098,8 +3098,8 @@ that will be called. The method should be public and created on the Component:
 As soon as the `query` LiveProp is updated, the ``onQueryUpdated()`` method
 will be called. The previous value is passed there as the first argument.
 
-It also works with non-scalar types, expand ``onUpdated`` configuration
-to an array with format ``['field_name' => 'function_name']``:
+If you're allowing object properties to be writable, you can also listen to
+the change of one specific key:
 
 .. code-block::
 
@@ -3108,34 +3108,12 @@ to an array with format ``['field_name' => 'function_name']``:
     #[AsLiveComponent]
     class EditPost
     {
-        #[LiveProp(writable: [LiveProp::IDENTITY, 'title', 'content'], onUpdated: ['title' => 'onTitleUpdated'])]
+        #[LiveProp(writable: ['title', 'content'], onUpdated: ['title' => 'onTitleUpdated'])]
         public Post $post;
 
         // ...
 
         public function onTitleUpdated($previousValue): void
-        {
-            // ...
-        }
-    }
-
-You can iterate more fields with corresponding function names if you want.
-Also, you can even create a hook for the entire entity identity change - use
-``LiveProp::IDENTITY`` as a key for the function name:
-
-.. code-block::
-
-    // ...
-
-    #[AsLiveComponent]
-    class EditPost
-    {
-        #[LiveProp(writable: [LiveProp::IDENTITY, 'title', 'content'], onUpdated: [LiveProp::IDENTITY => 'onEntirePostChanged'])]
-        public Post $post;
-
-        // ...
-
-        public function onEntirePostChanged($previousValue): void
         {
             // ...
         }
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3101,7 +3101,7 @@ will be called. The previous value is passed there as the first argument.
 It also works with non-scalar types, expand ``onUpdated`` configuration
 to an array with format ``['field_name' => 'function_name']``:
 
-.. code-block:: php
+.. code-block::
 
     // ...
 
@@ -3123,7 +3123,7 @@ You can iterate more fields with corresponding function names if you want.
 Also, you can even create a hook for the entire entity identity change - use
 ``LiveProp::IDENTITY`` as a key for the function name:
 
-.. code-block:: php
+.. code-block::
 
     // ...
 
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3063,6 +3063,84 @@ Then specify this new route on your component:
           use DefaultActionTrait;
       }
 
+Add a Hook on LiveProp Update
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.12
+
+    The ``onUpdated`` option was added in LiveComponents 2.12.
+
+If you want to run a custom code after a specific LiveProp was updated,
+you can do it by adding a ``onUpdated`` option. Specify a method name
+that will be called. The method should be public and created on the Component:
+
+.. code-block:: diff
+
+    // ...
+
+    #[AsLiveComponent]
+    class ProductSearch
+    {
+      - #[LiveProp(writable: true)]
+      + #[LiveProp(writable: true, onUpdated: 'onQueryUpdated')]
+        public string $query = '';
+
+        // ...
+
+        public function onQueryUpdated($previousValue): void
+        {
+            // $this->query already contains a new value
+            // and its previous value is passed as an argument
+        }
+    }
+}
+
+As soon as the `query` LiveProp is updated, the ``onQueryUpdated()`` method
+will be called. The previous value is passed there as the first argument.
+
+It also works with non-scalar types, expand ``onUpdated`` configuration
+to an array with format ``['field_name' => 'function_name']``:
+
+.. code-block:: php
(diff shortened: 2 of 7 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -559,4 +596,13 @@ private function recursiveKeySort(array &$data): void
...
         ksort($data);
     }
+
+    private function ensureOnUpdatedMethodExists(object $component, string $methodName): void
+    {
+        if (method_exists($component, $methodName)) {
+            return;
+        }
+
+        throw new \Exception(sprintf('Method "%s:%s()" specified as LiveProp "onUpdated" hook does not exist', $component::class, $methodName));
```
weaverryan:
```suggestion
        throw new \Exception(sprintf('Method "%s:%s()" specified as LiveProp "onUpdated" hook does not exist.', $component::class, $methodName));
```

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -208,6 +209,42 @@ public function hydrate(object $component, array $props, array $updatedProps, Li
                 // unexpected and can't be set - e.g. a string field for an `int` property.
                 // We ignore this, and allow the original value to remain set.
             }
+
+            // A special hook that will be called if the LiveProp was changed
+            // and $onUpdated argument is set on the attribute
+            if ($propMetadata->onUpdated()) {
+                if (\is_string($propMetadata->onUpdated())) {
```
weaverryan: Let's move everything inside `if ($propMetadata->onUpdated()) {` into a new private method. That will allow you to "exit early" in the code below, instead of needing the else and elseif stuff. For example, if this were in a private method, the next 4 lines could be:

```php
if (!$dehydratedUpdatedProps->hasPropValue($frontendName)) {
    return;
}

$this->ensureOnUpdatedMethodExists($component, $propMetadata->onUpdated());
$propertyOldValue = $dehydratedOriginalProps->getPropValue($frontendName);
$component->{$propMetadata->onUpdated()}($propertyOldValue);

return;
```

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -208,6 +209,42 @@ public function hydrate(object $component, array $props, array $updatedProps, Li
...
+                        $propertyOldValue = $dehydratedOriginalProps->getPropValue($frontendName);
+                        $component->{$propMetadata->onUpdated()}($propertyOldValue);
+                    }
+                } elseif (\is_array($propMetadata->onUpdated())) {
+                    foreach ($propMetadata->onUpdated() as $propName => $funcName) {
+                        $this->ensureOnUpdatedMethodExists($component, $funcName);
+
+                        if (LiveProp::IDENTITY === $propName) {
+                            if ($dehydratedUpdatedProps->hasPropValue($frontendName)
+                                && $dehydratedOriginalProps->getPropValue($frontendName) !== $dehydratedUpdatedProps->getPropValue($frontendName)
```
weaverryan: I'm not sure about the 2nd half of this if statement. If the prop is in the updated props, then I think we should always call the hook.

src/LiveComponent/src/LiveComponentHydrator.php
weaverryan: Also, the following 2 things are equivalent:

```php
#[LiveProp(onUpdate: 'updateFoo')]

#[LiveProp(onUpdate: [LiveProp::IDENTITY => 'updateFoo'])]
```

So, to make the algorithm easier, we could do:

```
$onUpdated = $propMetadata->onUpdated();
if (\is_string($propMetadata->onUpdated())) {
    $onUpdated = [LiveProp::IDENTITY => $onUpdated];
}
```

Then we'll only need the foreach part of the algorithm.

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -144,6 +144,140 @@ public function testCanDehydrateAndHydrateComponentWithTestCases(callable $testF
...
+                        $this->number = $oldValue;
+                    }
+                }
+            })
+                ->mountWith(['number' => 10])
+                ->userUpdatesProps(['number' => 9])
+                ->assertObjectAfterHydration(function (object $object) {
+                    $this->assertSame(10, $object->number);
+                });
+        }];
```
weaverryan: Unless there is something you're specifically worried about, we probably don't need a string and an int test - one for scalars would be fine

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -144,6 +144,140 @@ public function testCanDehydrateAndHydrateComponentWithTestCases(callable $testF
...
+                ->mountWith(['entity1' => $entity1])
+                ->userUpdatesProps(['entity1' => null])
+                ->assertObjectAfterHydration(function (object $object) use ($entity1) {
+                    $this->assertNotNull($object->entity1);
+                    $this->assertSame(
+                        $entity1->id,
+                        $object->entity1->id
+                    );
+                });
+        }];
```
weaverryan: Especially once you do a little refactoring, I think this test won't be needed either (just keep the one basic test for the IDENTITY above)

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -144,6 +144,140 @@ public function testCanDehydrateAndHydrateComponentWithTestCases(callable $testF
...
+                }
+            })
+                ->mountWith(['number' => 10])
+                ->userUpdatesProps(['number' => 9])
+                ->assertObjectAfterHydration(function (object $object) {
+                    $this->assertSame(10, $object->number);
+                });
+        }];
+
+        yield 'onUpdated: with object value' => [function () {
```
weaverryan:
```suggestion
        yield 'onUpdated: set to an array' => [function () {
```

(really, the important part is that you're able to add a hook for specific writable paths - technically the property value could be an array)

weaverryan (commented):
I love this! I've added some technical notes. Could you add some docs for this? Something similar to your PR description would work great. For `LiveProp::IDENTITY`, that only really makes sense when you are making the identity writable AND a few specific properties - e.g. `writable: ['name', 'address', LiveProp::IDENTITY]`, so I'd tweak that example slightly.

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
bocharsky-bw: I thought about getting a properly hydrated value, but mostly in this feature-related code. Anyway, it should work well I suppose, moreover dehydrator should have its own tests too... so will drop this test

src/LiveComponent/src/LiveComponentHydrator.php
bocharsky-bw: Oh cool, seems tests are happy with this change too :)

src/LiveComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## ...
```
weaverryan: You can put 2.12.0 here

src/LiveComponent/doc/index.rst
```diff
@@ -3063,6 +3063,84 @@ Then specify this new route on your component:
...
       }
 
+Add a Hook on LiveProp Update
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.12
+
+    The ``onUpdated`` option was added in LiveComponents 2.12.
+
+If you want to run a custom code after a specific LiveProp was updated,
```
```suggestion
If you want to run custom code after a specific LiveProp is updated,
```

src/LiveComponent/doc/index.rst
```diff
@@ -3063,6 +3063,84 @@ Then specify this new route on your component:
...
+you can do it by adding a ``onUpdated`` option. Specify a method name
+that will be called. The method should be public and created on the Component:
```
weaverryan: How about:

```
you can do it by adding a ``onUpdated`` option set to a public method name on the component:
```

src/LiveComponent/doc/index.rst
```diff
@@ -3063,6 +3063,84 @@ Then specify this new route on your component:
...
+            // $this->query already contains a new value
+            // and its previous value is passed as an argument
+        }
+    }
+}
+
+As soon as the `query` LiveProp is updated, the ``onQueryUpdated()`` method
+will be called. The previous value is passed there as the first argument.
+
+It also works with non-scalar types, expand ``onUpdated`` configuration
```
weaverryan: How about:

```
If you're allowing object properties to be writable, you can also listen to the change of one
specific key:
```

src/LiveComponent/doc/index.rst
```diff
@@ -3063,6 +3063,84 @@ Then specify this new route on your component:
...
+to an array with format ``['field_name' => 'function_name']``:
+
+.. code-block::
+
+    // ...
+
+    #[AsLiveComponent]
+    class EditPost
+    {
+        #[LiveProp(writable: [LiveProp::IDENTITY, 'title', 'content'], onUpdated: ['title' => 'onTitleUpdated'])]
```
```suggestion
        #[LiveProp(writable: ['title', 'content'], onUpdated: ['title' => 'onTitleUpdated'])]
```

`LiveProp::IDENTITY` is just not a common thing, so might as well leave it off.

src/LiveComponent/doc/index.rst
```diff
@@ -3063,6 +3063,84 @@ Then specify this new route on your component:
...
+
+        // ...
+
+        public function onTitleUpdated($previousValue): void
+        {
+            // ...
+        }
+    }
+
+You can iterate more fields with corresponding function names if you want.
```
weaverryan: Let's remove this entire section. This is just *such* an edge-case thing, I don't think it's even worth documenting.

weaverryan (commented):
Yay! Just some minor docs tweaks. Thank you!

kbond (approved):
Nice feature!

weaverryan:
Thanks Victor!

weaverryan:
This is merged

PR #1117, opened 2023-09-14
[TwigComponent] Remove undefined var $name in ComponentExtension

### LINKED ISSUE #1115: [TwigComponent] Undefined var `$name` in ComponentExtension ?
In TwigComponent/src/Twig/[ComponentExtension](https://github.com/symfony/ux/blob/7412bd2384aac5c0f41d5f6321218b54dd603006/src/TwigComponent/src/Twig/ComponentExtension.php#L83
), there is this method 

```php
    public function finishEmbeddedComponentRender(): void
    {
        try {
            $this->container->get(ComponentRenderer::class)->finishEmbeddedComponentRender();
        } catch (\Throwable $e) {
            $this->throwRuntimeError($name, $e);
        }
    }
```

I fear this won't work as `$name` is not defined... or am I missing something ?

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #1115 
| License       | MIT

Implement the solution suggested by @kbond  here:  https://github.com/symfony/ux/issues/1115#issuecomment-1719726445

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentExtension.php
+++ b/src/TwigComponent/src/Twig/ComponentExtension.php
@@ -82,11 +82,7 @@ final class ComponentExtension extends AbstractExtension implements ServiceSubsc
 
     public function finishEmbeddedComponentRender(): void
     {
-        try {
-            $this->container->get(ComponentRenderer::class)->finishEmbeddedComponentRender();
-        } catch (\Throwable $e) {
-            $this->throwRuntimeError($name, $e);
-        }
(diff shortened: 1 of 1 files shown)


kbond (approved):
I'm good with this!

sneakyvv (approved):
This is some copy pasta :\ 

The method could be altered so the name of the component is available and that the regular exception could be thrown, but since this is at the end of rendering it isn't appropriate after all.
TL;DR: removing this is ok 👍

weaverryan:
Thanks Simon!

PR #1119, opened 2023-09-16
[Site] Use `system-ui` font family

smnandre:
### TL;DR;

I suggest we use "system-ui" (native/OS font) instead of "Inter" on the website.

--- 

### What ? 
* System font are the default font used in every OS 
* They can be used instead of webfont .. and are excellent fonts.
* This article is a great overview on system fonts: https://woorkup.com/system-font/ 
* Already used by Github, bootstrap, web.dev, apple, Stripe, and many more

### Why ? 
* better web performance (good for seo, greater for accessibility & UX)
* less bandwidth used (good for 💰, greater for 🌍)
* the system fonts will allow us more things than Inter 
* Allow the user to get the standard font used in the OS

### How does it look ?

The difference is subtile... at least. 
And i'm pretty sure only a very, very low part of the visitors will notice the difference.

### Test it

I made some captures for you to see the "before/after" on our main pages.. 
They are available on a page if you want to test yourself the slider and see the differences
-> https://smnand.re/w/ab4e91/ 

| All demos | Invoice demo |
| - | - |
| all-demos | demo-invoice |

| [Test it yourself (external link)](https://smnand.re/w/ab4e91/) | 
| - | 
| comparaison-tool |

### DIFF
--- a/ux.symfony.com/assets/styles/app/_html.scss
+++ b/ux.symfony.com/assets/styles/app/_html.scss
@@ -1,5 +1,5 @@
 body {
-  font-family: Inter;
+  font-family: var(--font-family-text);
   font-size: 16px;
   font-weight: 400;
   line-height: 26px;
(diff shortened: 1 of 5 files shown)


weaverryan:
I'm convinced! Excellent, detailed arguments. Thanks Simon!

PR #1121, opened 2023-09-16
[Site] Use monospace system font

smnandre:
TL;DR; i suggest to use native system fonts instead of webfont.

Follow https://github.com/symfony/ux/pull/1119, this time for the `monospace` font, with the same reasons.

Again, I put online some before/after captures*  with a slider tool to compare.
(* made on macOS Ventura / Firefox)

Captures here: https://smnand.re/w/b68eda/ 

slider-comparaison

I made two separate PR because developpers tend to be a little more sensible about code fonts :)

### DIFF
--- a/ux.symfony.com/assets/styles/_variables.scss
+++ b/ux.symfony.com/assets/styles/_variables.scss
@@ -16,3 +16,5 @@ $green: #84DE2C;
 $primary: #222 !default;
 
 $size-unit: calc(8px + 1.5625vw);
+
+$font-family-monospace: var(--font-family-code);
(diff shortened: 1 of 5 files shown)


weaverryan:
This time I can see the difference easily on https://smnand.re/w/b68eda/ - but I don't personally mind it.

Anyone have a strong opinion on the before/after of the code blocks in the link above? 

Also, this will need a rebase now - after merging your other PR's ;)

smnandre:
We can play with the multiple font settings / axes.. I find it more readable like that but it's only a personal opinion.

weaverryan:
Let's roll with it - if there are any major objections (which I doubt), then we can talk about those when they happen. Thanks!

weaverryan:
Merged - not sure why GH didn't see that

PR #1134, opened 2023-09-22
Fix Typo in Context / Variables Inside of Blocks doc

rdavaillaud:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | 
| License       | MIT

Some `block content` are missing in the classic syntax.


weaverryan (approved):
You're right. We really need to change the docs exclusively over to the HTML syntax - this is much uglier than that :)

weaverryan:
Thanks Raphaël!

PR #1135, opened 2023-09-22
Add Symfony 7 Support

JoppeDC:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT

Update packages to allow Symfony 7

## Changes done by this PR:

- Update the tests matrix
- - Test php 8.1 with lowest deps
- - Test php 8.2 with highest deps
- Update a small number of classes to comply with updated interfaces (add return types)
- No longer support Symfony 5 and <=6.2 for the Autocomplete Bundle due to BC's in the Security bundle
- Update all composer deps to add Symfony 7.x support


jmsche:
I think all asset mapper requirements can be updated with `^6.3|^7.0` (so -dev is not required anymore)

src/React/composer.json
```diff
@@ -33,11 +33,11 @@
     },
     "require-dev": {
         "symfony/asset-mapper": "6.3.x-dev",
```
weaverryan: This is what @jmsche is referring to :)

weaverryan:
This is looking good, but when I check 2 random jobs - https://github.com/symfony/ux/actions/runs/6295950088/job/17090056239?pr=1135 and https://github.com/symfony/ux/actions/runs/6295950088/job/17090056789?pr=1135 the dependencies are still Symfony 6.4.

What is blocking Symfony 7 from being installed on those jobs? It may be that some other dependency that we can't yet control is blocking us, but we should check and make sure the cause :) 

Thanks!

kbond:
At least some of the dependency issues are caused by zenstruck/foundry & zenstruck/browser. They both rely on upstream packages that do not yet support Symfony 7. I'll be working on this.

weaverryan:
Friendly ping @JoppeDC - there are a few conflicts now (probably easy). And if you can double-check *why* Symfony 7 isn't installed in those cases (as long as we know the reason and it's not "we're doing something wrong"), I'll be happy.

JoppeDC:
I've resolved the conflicts. As for the 6.x installs, i've noticed they are blocked by a few packages. Mostly the zenstruck/* as mentioned above, but also some packages included in this update.


- zenstruck/browser
- zenstruck/foundry
- symfony/ux-twig-component
- symfony/stimulus-bundle

JoppeDC:
Although i notice that the StimulusBundle package still has a failing test. The `AssetDependency` class seems to no longer be available in 6.4 / 7.x. (This is due to removing the assetmapper 6.3 constraint)

kbond:
FYI, zenstruck/browser & foundry should at least be installable with Symfony 7 now.

JoppeDC:
Amazing @kbond , thanks a lot. Just tested locally on some of the packages, and it seems like its indeed resolving to 7.x now :) Will need to fix the StimulusBundle issue first and then i can verify in the new pipeline run that 7.x is installed :)

JoppeDC:
Cool. I've got most issues resolved. There's still an issue related to the SecurityBundle, which i will have to check out further tomorrow.

(We were using `Symfony\Component\Security\Core\Security`, which was deprecated in 6.x and removed in 7. The solution to this is to use `Symfony\Bundle\SecurityBundle\Security`, but this was not yet available in 5.x, and thus breaks the lowest test.)

weaverryan:
Let's drop support for 5.4 in Autocomplete and then use the new class.

weaverryan (approved):

weaverryan:
Thanks for your excellent work @JoppeDC!

PR #1139, opened 2023-09-23
[Site] Quick wins to improve website accessibility

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT

Minor PR to improve the website accessibility via some "quick-wins"
* add missing aria labels
* use pre/code instead of input in package "install command"

### DIFF
--- a/ux.symfony.com/assets/styles/app/_typography.scss
+++ b/ux.symfony.com/assets/styles/app/_typography.scss
@@ -24,8 +24,8 @@ h2.ubuntu {
 h3, a.component-title {
   font-family: var(--font-family-title);
   font-weight: 700;
-  font-size: 28px !important;
-  line-height: 32px !important;
+  font-size: 28px;
+  line-height: 32px;
   letter-spacing: -1px;
   color: #0A0A0A;
   text-decoration: none;
--- a/ux.symfony.com/assets/styles/components/_PackageBox.scss
+++ b/ux.symfony.com/assets/styles/components/_PackageBox.scss
@@ -112,6 +112,14 @@
   flex-direction: row;
   align-items: center;
   gap: .75rem;
+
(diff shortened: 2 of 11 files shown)


ux.symfony.com/templates/base.html.twig
```diff
@@ -28,14 +28,15 @@
...
+        <a href="#main" class="skip-to-content">Skip to main content</a>
         {% block body %}
             <div class="App">
                 {% block header %}
                     {{ include('_header.html.twig') }}
                 {% endblock %}
 
                 {% block main %}
-                    <main>
+                    <main id="main">
```
weaverryan: Playing around, it seems like on most (all?) pages, the `main` element is already at the top of the page. So clicking the link doesn't actually do anything. Are we sure this will be helpful?

smnandre: For me it is just a quick win, we'll have a lot of work to be fully accessible, but it goes in the good direction.
* it is not visible per default
* it is usefull for screen readers / keyboard-only (see https://www.w3.org/TR/WCAG20-TECHS/G1.html)

But we could give it less "contrast" and keep the same background if you think it's too much ? Or see a bit later to handle the accessible question in a more "site-wide" way ?

weaverryan: I'm not worried about the look - as it's invisible unless you hit tab. It just seemed odd that we have this, but the `main` is ALWAYS already on the top.

smnandre: Let's remove it for now then!

The other modifications will have a direct impact so i'll update the PR is that ok ? Or you prefer i close this one and start another ?

weaverryan:
Thanks Simon!

PR #1143, opened 2023-09-25
Add deferred live components

### LINKED ISSUE #994: [LiveComponent] deferred/lazy loading
From https://github.com/symfony/ux/issues/102

> Future G) Potentially add support for "lazy" components (they don't load until they are visible) or at least document how one could use a lazy Turbo Frame nicely with a component URL. For example, on page load, a component has a "loading" animation. Then, it loads via Ajax and the area is updated.

[Livewire's docs for this feature](https://laravel-livewire.com/docs/2.x/defer-loading)

jakubtobiasz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #994
| License       | MIT

Initially this feature was called `lazy loading`, but this name could be misleading, as my solution doesn't provide **real** lazy loading. So, I named it `deferred live components`, I guess this name fits better.

How to use it?

```twig
<twig:MyComponent defer />
```

In such case we'll get an empty div. Once `live:connect` event called, it'll load the component in the background. We can also define a template to be rendered while loading (e.g. a cool spinner or some text).

```twig
<twig:MyComponent defer defer-loading-template="my_cool_spinner.html.twig" />
```

I'm open for any suggestions 🙌🏼!

### DIFF
--- a/src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\UX\LiveComponent\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\UX\TwigComponent\Event\PostMountEvent;
+use Symfony\UX\TwigComponent\Event\PreRenderEvent;
+
+final class DeferLiveComponentSubscriber implements EventSubscriberInterface
+{
+    private const DEFAULT_LOADING_TAG = 'div';
+
+    private const DEFAULT_LOADING_TEMPLATE = null;
+
+    public function onPostMount(PostMountEvent $event): void
+    {
+        $data = $event->getData();
+        if (\array_key_exists('defer', $data)) {
+            $event->addExtraMetadata('defer', true);
+            unset($event->getData()['defer']);
+        }
+
+        if (\array_key_exists('loading-template', $data)) {
+            $event->addExtraMetadata('loading-template', $data['loading-template']);
+            unset($event->getData()['loading-template']);
+        }
+
+        if (\array_key_exists('loading-tag', $data)) {
+            $event->addExtraMetadata('loading-tag', $data['loading-tag']);
+            unset($event->getData()['loading-tag']);
+        }
+
+        $event->setData($data);
+    }
+
+    public function onPreRender(PreRenderEvent $event): void
+    {
+        $mountedComponent = $event->getMountedComponent();
+
+        if (!$mountedComponent->hasExtraMetadata('defer')) {
+            return;
+        }
+
+        $event->setTemplate('@LiveComponent/deferred.html.twig');
+
+        $variables = $event->getVariables();
+        $variables['loadingTemplate'] = self::DEFAULT_LOADING_TEMPLATE;
+        $variables['loadingTag'] = self::DEFAULT_LOADING_TAG;
+
+        if ($mountedComponent->hasExtraMetadata('loading-template')) {
+            $variables['loadingTemplate'] = $mountedComponent->getExtraMetadata('loading-template');
+        }
+
+        if ($mountedComponent->hasExtraMetadata('loading-tag')) {
+            $variables['loadingTag'] = $mountedComponent->getExtraMetadata('loading-tag');
+        }
+
+        $event->setVariables($variables);
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            PostMountEvent::class => ['onPostMount'],
+            PreRenderEvent::class => ['onPreRender'],
+        ];
+    }
+}
--- a/src/LiveComponent/templates/deferred.html.twig
+++ b/src/LiveComponent/templates/deferred.html.twig
@@ -1,0 +1,7 @@
+<{{ loadingTag }} {{ attributes }} data-action="live:connect->live#$render">
+    {% block loadingContent %}
+        {% if loadingTemplate != null %}
+            {{ include(loadingTemplate) }}
+        {% endif %}
+    {% endblock %}
+</{{ loadingTag }}>
--- a/src/TwigComponent/src/MountedComponent.php
+++ b/src/TwigComponent/src/MountedComponent.php
@@ -19,13 +19,6 @@ namespace Symfony\UX\TwigComponent;
 final class MountedComponent
 {
     /**
-     * Any extra metadata that might be useful to set.
-     *
-     * @var array<string, mixed>
-     */
-    private array $extraMetadata = [];
-
-    /**
      * @param array|null $inputProps if the component was just originally created,
      *                               (not hydrated from a request), this is the
      *                               array of initial props used to create the component
@@ -34,7 +27,8 @@ final class MountedComponent
         private string $name,
         private object $component,
         private ComponentAttributes $attributes,
-        private ?array $inputProps = []
+        private ?array $inputProps = [],
+        private array $extraMetadata = [],
     ) {
     }
 
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.13.0
+
+-   Add deferred rendering of Live Components
+
 ## 2.12.0
 
 -   Add `onUpdated` hook for `LiveProp`
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2175,6 +2175,32 @@ To validate only on "change", use the ``on(change)`` modifier:
         class="{{ _errors.has('post.content') ? 'is-invalid' : '' }}"
     >
 
+Deferring the Loading
+---------------------
+
+Certain components might be heavy to load. You can defer the loading of these components
+until after the rest of the page has loaded. To do this, use the ``defer`` attribute:
+
+.. code-block:: twig
+
+    {{ component('SomeHeavyComponent', { defer: true }) }}
+
+Doing so will render an empty "placeholder" tag with the live attributes. Once the ``live:connect`` event is triggered,
+the component will be rendered asynchronously.
+
+By default the rendered tag is a ``div``. You can change this by specifying the ``loading-tag`` attribute:
+
+.. code-block:: twig
+
+    {{ component('SomeHeavyComponent', { defer: true, loading-tag: 'span' }) }}
+
+If you need to signify that the component is loading, use the ``loading-template`` attribute.
+This lets you provide a Twig template that will render inside the "placeholder" tag:
+
+.. code-block:: twig
+
+    {{ component('SomeHeavyComponent', { defer: true, loading-template: 'spinning-wheel.html.twig' }) }}
+
 Polling
 -------
 
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -25,6 +25,7 @@ use Symfony\UX\LiveComponent\ComponentValidatorInterface;
 use Symfony\UX\LiveComponent\Controller\BatchActionController;
 use Symfony\UX\LiveComponent\EventListener\AddLiveAttributesSubscriber;
 use Symfony\UX\LiveComponent\EventListener\DataModelPropsSubscriber;
+use Symfony\UX\LiveComponent\EventListener\DeferLiveComponentSubscriber;
 use Symfony\UX\LiveComponent\EventListener\InterceptChildComponentRenderSubscriber;
 use Symfony\UX\LiveComponent\EventListener\LiveComponentSubscriber;
 use Symfony\UX\LiveComponent\EventListener\ResetDeterministicIdSubscriber;
@@ -212,6 +213,14 @@ final class LiveComponentExtension extends Extension implements PrependExtension
             ])
             ->addTag('kernel.event_subscriber')
             ->addTag('container.service_subscriber', ['key' => LiveControllerAttributesCreator::class, 'id' => 'ux.live_component.live_controller_attributes_creator'])
+        ;
+
+        $container->register('ux.live_component.defer_live_component_subscriber', DeferLiveComponentSubscriber::class)
+            ->setArguments([
+                new Reference('ux.twig_component.component_stack'),
+                new Reference('ux.live_component.live_controller_attributes_creator'),
+            ])
+            ->addTag('kernel.event_subscriber')
         ;
 
         $container->register('ux.live_component.deterministic_id_calculator', DeterministicTwigIdCalculator::class);
--- a/src/LiveComponent/tests/Fixtures/Component/DeferredComponent.php
+++ b/src/LiveComponent/tests/Fixtures/Component/DeferredComponent.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent('deferred_component')]
+final class DeferredComponent
+{
+    use DefaultActionTrait;
+
+    public function getLongAwaitedData(): string
+    {
+        return 'Long awaited data';
+    }
+}
--- a/src/LiveComponent/tests/Fixtures/templates/components/deferred_component.html.twig
+++ b/src/LiveComponent/tests/Fixtures/templates/components/deferred_component.html.twig
@@ -1,0 +1,1 @@
+<div {{ attributes }}>{{ computed.longAwaitedData }}</div>
--- a/src/LiveComponent/tests/Fixtures/templates/dummy/loading.html.twig
+++ b/src/LiveComponent/tests/Fixtures/templates/dummy/loading.html.twig
@@ -1,0 +1,1 @@
+I'm loading a reaaaally slow live component
--- a/src/LiveComponent/tests/Fixtures/templates/render_deferred_component.html.twig
+++ b/src/LiveComponent/tests/Fixtures/templates/render_deferred_component.html.twig
@@ -1,0 +1,1 @@
+<twig:deferred_component defer />
--- a/src/LiveComponent/tests/Fixtures/templates/render_deferred_component_with_li_tag.html.twig
+++ b/src/LiveComponent/tests/Fixtures/templates/render_deferred_component_with_li_tag.html.twig
@@ -1,0 +1,1 @@
+<twig:deferred_component defer loading-tag='li' />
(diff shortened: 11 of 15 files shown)


src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -73,6 +73,21 @@ public function onPreRender(PreRenderEvent $event): void
                     $originalAttributes['data-embedded-template-index'],
                 );
             }
+
+            $hasLiveId = isset($originalAttributes['data-live-id']);
```
weaverryan: Either this isn't quite right, or I'm surprised by this :p. `data-live-id` is optional. So I would expect this to *not* exist at all times for a component where this wasn't specified.

Is looking for `defer` enough? What I mean is: when you originally render a component with `defer`, it will be in `$originalAttributes`. But then if we remove `defer` from here, on a re-render, won't `defer` NOT be present? Or if this isn't quite true yet, perhaps we can tweak something so that it is: so that `defer` is only present during the initial render and is "lost" at this point so that it is naturally *not* present on re-render.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -73,6 +73,21 @@ public function onPreRender(PreRenderEvent $event): void
...
+
+            // if the original attributes has a data-live-id, it means the component
+            // was already rendered
+            if (!$hasLiveId && $isDeferred) {
+                $attributes = $attributes->without('defer', 'defer-loading-template');
+                $event->setTemplate('@LiveComponent/deferred.html.twig');
+
+                if ($loadingTemplate) {
+                    $variables['loadingTemplate'] = $loadingTemplate;
+                }
```
weaverryan: Small simplification: always set a `loadingTemplate` variable, but it will be` null if not passed.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -73,6 +73,21 @@ public function onPreRender(PreRenderEvent $event): void
...
             }
+
+            $hasLiveId = isset($originalAttributes['data-live-id']);
+            $isDeferred = \array_key_exists('defer', $originalAttributes);
+            $loadingTemplate = $originalAttributes['defer-loading-template'] ?? null;
+
+            // if the original attributes has a data-live-id, it means the component
+            // was already rendered
+            if (!$hasLiveId && $isDeferred) {
+                $attributes = $attributes->without('defer', 'defer-loading-template');
```
weaverryan: I'd like the shorter `loading-template` name... or maybe `loadingTemplate`... we don't do any transformations from `foo-bar` in HTML -> to `fooBar` as a property, so we're typically using camel case attribute names... since they don't actually render in HTML.

src/LiveComponent/templates/deferred.html.twig
```diff
@@ -0,0 +1,5 @@
+<div {{ attributes }} data-action="live:connect->live#$render">
+    {% if loadingTemplate is defined and loadingTemplate != null %}
```
weaverryan: Let's surround the entire "inside" of the `div` with a `{% block loadingContent %}{% ednblock %}`. That'll allow a 2nd way to customize the loading.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -73,6 +73,21 @@ public function onPreRender(PreRenderEvent $event): void
...
+                $event->setTemplate('@LiveComponent/deferred.html.twig');
```
weaverryan: It could be outside of the scope of this PR, but it'd be great to have some semantic configuration so you could control the default loading template globally. I have some semantic config introduced in #1140, so it might make sense to wait for that before adding this new option (e.g. `loading_template`)

src/LiveComponent/templates/deferred.html.twig
```diff
@@ -0,0 +1,5 @@
+<div {{ attributes }} data-action="live:connect->live#$render">
```
weaverryan: We should allow an extra `loadingElement` "variable" for the edge-case when the user needs something like an `li` or a `tr`.

weaverryan (commented):
I've left some notes - but this is wonderful! The name "lazy" is trendier... but defer might be slightly more accurate. Thinking ahead: if we, in the future, extended this to "don't load until the element is in the viewport", how would we name that?

src/LiveComponent/templates/deferred.html.twig
jakubtobiasz: Done 👌🏼

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
jakubtobiasz: I decided to use `loading-template` as an attribute name, but in the template it's `loadingTemplate`. `kebab-case` is a common pick for HTML, so I guess it fits perfectly here.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
jakubtobiasz: Done 👌🏼

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
jakubtobiasz: As we talked, I discovered `$event->getMountedComponent()->getInputProps()` returns `lazy: true` for the first time, and on rerender it's empty, so it works now without `data-live-id` 🙌🏼.

I'm only afraid `defer`, `loading-template` and `loading-tag` are going to be visible on the frontend (in the `data-live-props-value`), but it's not a such big deal for now.

src/LiveComponent/templates/deferred.html.twig
jakubtobiasz: Done 🙌🏼! I called it `loading-tag` in terms of an attribute, and `loadingTag` as a Twig's variable in the `deffered.html.twig`.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
```diff
@@ -73,6 +77,17 @@ public function onPreRender(PreRenderEvent $event): void
...
+            $inputProps = $event->getMountedComponent()->getInputProps();
+            $isDeferred = array_key_exists('defer', $inputProps);
+
+            if ($isDeferred) {
+                $event->setTemplate('@LiveComponent/deferred.html.twig');
+
+                $variables['loadingTag'] = $originalAttributes['loading-tag'] ?? self::DEFAULT_LOADING_TAG;
+                $variables['loadingTemplate'] = $originalAttributes['loading-template'] ?? self::DEFAULT_LOADING_TEMPLATE;
+                $attributes = $attributes->without('defer', 'loading-element', 'loading-template');
+            }
```
weaverryan: I still have a preference to have this in its own event listener.

weaverryan (commented):
I like it! Though, I do think we're going to need to remove the `loading-template` from the `data-live-props-value`. Both for being responsible/clean... but also not exposing template paths to the frontend.

jakubtobiasz:
> Thinking ahead: if we, in the future, extended this to "don't load until the element is in the viewport", how would we name that?

I guess we can call it `lazy` then. Or maybe there's a case to keep both, and have the current way (`defer`) and the new one (`lazy`).

weaverryan:
Indeed, I think `defer` better communicates what this does and `lazy` would better describe the viewport idea.

src/LiveComponent/templates/deferred.html.twig
```diff
@@ -0,0 +1,7 @@
+<{{ loadingTag }} {{ attributes }} data-action="live:connect->live#$render">
+    {% block loadingContent %}
```
smnandre, line 2: Would this working ? 

```
<twig:Notifications defer>
    <p>Loading...</p>
</twig:Notifications>

```

If not, i thing this could be a great DX addition... wdyt ?  

If you'd want to use the "default" content block you'd just have to name it 

```
<twig:Notifications defer>
   {# Displayed on page load #}
    <p>Loading...</p>

    {# Displayed on demand #}
    <twig:block name="content">
       No notification
    </twig:block>
</twig:Notifications>
```

kbond:
> Indeed, I think defer better communicates what this does and lazy would better describe the viewport idea.

So, in the future, to be deferred and lazy, you'd need the following?

```html
<twig:MyComponent defer lazy />
```

I think that would work best as we could utilize lazy with polling:

```html
<twig:MyComponent defer lazy data-poll />
```

The above would only load when in the viewport and then only poll when still in the viewport.

jakubtobiasz:
> So, in the future, to be deferred and lazy, you'd need the following?

I guess `deferred` **or** `lazy` should be enough 🤔.

src/LiveComponent/templates/deferred.html.twig
jakubtobiasz: I've checked this trying to do
```twig
<twig:SomeLiveComponent lazy>
    {% block loadingContent %}
        Loading...
    {% endblock %}
</twig:SomeLiveComponent>
```

But it renders `SomeLiveComponent` instantly 🥲. I'll try to debug how it works, maybe some adjustments and it'll work.

kbond:
> I guess deferred or lazy should be enough 🤔.

I think we'd need both to allow the following to be possibly:

```html
<twig:MyComponent lazy data-poll />
```

This would load the component initially, then only poll when in the viewport.

src/LiveComponent/src/EventListener/AddLiveAttributesSubscriber.php
jakubtobiasz: You want it, you have it 🫡!

src/TwigComponent/src/MountedComponent.php
```diff
@@ -53,6 +53,11 @@ public function getAttributes(): ComponentAttributes
         return $this->attributes;
     }
 
+    public function setAttributes(ComponentAttributes $attributes): void
```
smnandre: Do you by any chance have an alternate solution? (don't want to waste your time)

MountedComponent was "almost" immutable and i don't fear your usage right now at all..  
but it may be seen as an open door 

I'm just asking by chance... but not a big deal :)

jakubtobiasz: I was looking for another way, but I didn't find anything :/. This solutions is a result of my talk with Ryan 😅.

smnandre: Well let's ignore my comment then :)

kbond:
Question: should this be `data-defer`? If I understand correctly, since the `defer` attribute isn't rendered in the actual browser dom, it isn't an issue but... consistency? I don't have a strong opinion myself, just wanted to bring this up for discussion. Assuming we add a `lazy` attribute later for viewport/current tab detection, this will need to be `data-lazy` (and we already have `data-poll`).

smnandre:
> Question: should this be `data-defer`? If I understand correctly, since the `defer` attribute isn't rendered in the actual browser dom, it isn't an issue but... consistency? I don't have a strong opinion myself, just wanted to bring this up for discussion. Assuming we add a `lazy` attribute later for viewport/current tab detection, this will need to be `data-lazy` (and we already have `data-poll`).

My (very personal) interpretation:
* `eager` = "now"
* `defer` = "bottom of the todo-list"
* `lazy` = "much later / when it'll be needed"

So "defer" is more to me like the kernel.terminate..  and lazy quite working there..

Two other ideas:
* "`async`" simple and not related to "when" it will be renderer..
* `data-render=live`

jakubtobiasz:
The name `defer` is inspired by attributes available for `script` tag.

> If async is present: The script is downloaded in parallel to parsing the page, and executed as soon as it is available (before parsing completes)
If defer is present (and not async): The script is downloaded in parallel to parsing the page, and executed after the page has finished parsing

The component is "executed" after the page loads (and it's also "downloaded" then in our case 😅).

In terms of `data-` prefix, I guess we get get rid of it in the Twig files, and add this prefix only in the rendered template in all cases 🤔.

weaverryan:
> I think we'd need both to allow the following to be possibly:
 > <twig:MyComponent lazy data-poll />
> This would load the component initially, then only poll when in the viewport.

I was imagining these as 2 different mechanism. The ability to "only poll when in viewport" would be a property of the `data-poll` specifically - so something like;

```html
<twig:MyComponent lazy data-poll="lazy|$render" />
```

Yes, that `|$render` at the end is a little bulky. But the point is, `lazy` is a modifier of the polling.

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    private const PRIORITY = 16;
+
+    private const DEFAULT_LOADING_TAG = 'div';
+
+    private const DEFAULT_LOADING_TEMPLATE = null;
+
+    public function __construct(
+        private ComponentStack $componentStack,
+        private LiveControllerAttributesCreator $attributesCreator,
```
weaverryan: Both of these args are unused

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
```diff
@@ -0,0 +1,66 @@
...
+            \ARRAY_FILTER_USE_KEY,
+        );
+
+        $variables[$attributesKey] = $attributes;
+        $variables['loadingTag'] = $originalAttributes['loading-tag'] ?? self::DEFAULT_LOADING_TAG;
+        $variables['loadingTemplate'] = $originalAttributes['loading-template'] ?? self::DEFAULT_LOADING_TEMPLATE;
+
+        $mountedComponent->setAttributes(
+            $mountedComponent->getAttributes()->without('defer', 'loading-template', 'loading-tag'),
+        );
```
weaverryan: @kbond does this look clunky to you? The fact that we need to... kind of modify the attributes in `$variables[$attributesKey]` AND also on `MountedComponent`?

Should we be doing this in an earlier hook? Something that can intercept the `defer` input prop earlier and remove it? Maybe set that data onto `MountedComponent::extraMetadata()` so it can be read here to set the variables?

weaverryan (commented):
Can you take a shot at some docs also?

Thanks!

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
kbond: Hmm, perhaps, and I agree it is clunky but it _does_ work within the current feature set. Opportunity for a future refactor.. I know I've said this for a few things recently. Maybe _ugly === ok_ until we need to duplicate it elsewhere?

weaverryan: Yea, I definitely have that on my mind too - finding the balance. What about:

A) In `PostMountEvent` https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/ComponentFactory.php#L195 - we add a new `addExtraMetadata()` method to the event.

B) Then, we keep track of that and put it into the `MountedComponent` - https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/ComponentFactory.php#L91-L113 - a little odd, as we need to carry return that "extra metadata" from the private `postMount()`, but I can't think of a better way. The user can intercept and remove "extra keys" from `PostMount`... but they don't have access to "attach" that to the `MountedComponent`.

kbond: Yeah, sounds like that would work. I guess I'd be a little concerned about splitting this logic into two places (assuming I am understanding correctly). It's clunky here but at least it's in one spot (which makes it easier to reason about imo).

weaverryan: Gotcha. I'm leaning towards my solution. It just doesn't feel right to allow `defer` to get ALL the way into attributes... just for us to remove it later, because it is never meant to be an attribute. It is really a special "prop", which should be processed and removed before it ever gets into attributes.

Sorry for the back and forth @jakubtobiasz but does this make sense? It would allow us to remove the new method in `MountedComponent`

kbond: > I'm leaning towards my solution

I'm ok with this - no strong objection.

jakubtobiasz: I'll try to take a look this week 🫡. I'll reach you once I find some difficulties 😂.

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+        if ($mountedComponent->hasExtraMetadata('loading-template')) {
+            $variables['loadingTemplate'] = $mountedComponent->getExtraMetadata('loading-template');
+        }
+
+        if ($mountedComponent->hasExtraMetadata('loading-tag')) {
+            $variables['loadingTag'] = $mountedComponent->getExtraMetadata('loading-tag');
+        }
+
+        $event->setVariables($variables);
+    }
```
weaverryan, line 61: 😻 This feels MUCH better to me.

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $event->setData($data);
+    }
+
+    public function onPreRender(PreRenderEvent $event): void
+    {
+        $mountedComponent = $event->getMountedComponent();
+        $isDeferred = $mountedComponent->hasExtraMetadata('defer');
+
+        if (!$isDeferred) {
```
weaverryan:
```suggestion
        if (!$mountedComponent->hasExtraMetadata('defer')) {
```

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
```diff
@@ -0,0 +1,73 @@
...
+            $variables['loadingTag'] = $mountedComponent->getExtraMetadata('loading-tag');
+        }
+
+        $event->setVariables($variables);
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            PostMountEvent::class => ['onPostMount', self::PRIORITY],
```
weaverryan: I think the priority is not needed on `PostMountEvent`? Wait, is it needed on `PreRenderEvent` either? I was thinking it would need to run before `AddLiveAttributesSubscriber`, but I'm not sure that's true.

src/TwigComponent/src/MountedComponent.php
weaverryan: This can be removed now, right?

weaverryan (commented):
I like this a lot more - THANK you :). Can you add a small section in the docs for this too? I'm thinking a new `-----` section (like Polling) right above Polling https://symfony.com/bundles/ux-live-component/current/index.html#polling called `Lazy-Loaded / Deferred Components`

src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
jakubtobiasz: I believe previously `PreRenderEvent` was requiring the priority, and... I copied the whole line for `onPostMount` 😂. I'll remove it in the upcoming push, as it seems to work without it.

src/TwigComponent/src/MountedComponent.php
jakubtobiasz: Yup, I missed that 😅

weaverryan:
Woohoo! THANK you for this wonderful work and feature Jacob!

PR #1145, opened 2023-09-26
Fix option tag synchronization

### LINKED ISSUE #960: [LiveComponent] When selecting an "empty" choice, it get blanked.
When using a live component, when the choice made is the "empty" choice and i'm using a placeholder as a choice name, it get blanked instead of staying on the placeholder text.

> weaverryan: I believe the problem is that the underlying value is being set to null. And then in JavaScript, LiveComponents is basically doing select.value = null, which causes “nothing” to be selected.

[screencast.webm](https://github.com/symfony/ux/assets/777666/bc1b6ed0-e072-4479-8925-c881b0291c86)

```php
                ->add('paymentDelay', EntityType::class, [
                    'class' => PaymentDelay::class,
                    'label' => 'Délais de paiement',
                    'required' => false,
                    'placeholder' => "Date d'échéance libre",
                    'attr' => [
                        'data-action' => "change->live#action",
                        'data-action-name' => 'refresh()'
                    ]
                ])
```

image (1)

dsoriano:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #960 
| License       | MIT

morphdom need to synchronize all `option` tags of a `select` to set the `selected` attribute on the good tag and then set the right `selectedIndex` on the parent `select` tag. See #960 for an illustration of the bug.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1237,7 +1237,7 @@ function executeMorphdom(rootFromElement, rootToElement, modifiedFieldElements, 
                 if (elementChanges) {
                     elementChanges.applyToElement(toEl);
                 }
-                if (fromEl.isEqualNode(toEl)) {
+                if (fromEl.nodeName.toUpperCase() !== 'OPTION' && fromEl.isEqualNode(toEl)) {
                     const normalizedFromEl = cloneHTMLElement(fromEl);
                     normalizeAttributesForComparison(normalizedFromEl);
                     const normalizedToEl = cloneHTMLElement(toEl);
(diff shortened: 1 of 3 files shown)


weaverryan:
THANK you for hunting this down! Do you think you could add a test for it - maybe in https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/test/controller/render.test.ts you can set up a situation that exposes the bug?

jcrombez:
great job <3

dsoriano:
No problem I will add a test and in the same time fix the failed check

dsoriano:
All done ! ;)

weaverryan:
Absolutely gorgeous - thank you Denis!

PR #1146, opened 2023-09-27
Fix Anonymous component props and Class backed component properties d…

### LINKED ISSUE #1138: [TwigComponent] Anonymous component props and Class backed component properties different behavior
Hi,

It seems that when using anonymous component, with the tag and HTML syntax, any defined property with a default value that is also defined in the parent/caller template is overridden by the parent template value.  
This behavior is not consistent with the function syntax and the class backed one, they should show the component default value.

## How to reproduce

The class backed define a property `message` with a default value 'foo'.
```php
// ClassBacked.php
#[AsTwigComponent]
final class ClassBacked
{
    public string $message = "foo";
}
```

```twig
{# ClassBacked.html.twig #}
{{ message }}
```

The anonymous one define a prop `message` 
```twig
{# Anonymous.html.twig #}
{% props message="foo" %}
{{ message }}
```

The parent template call each one in each available syntax:  

```twig
{# test.html.twig #}
{% set message = 'bar' %}

Function:
{{ component('ClassBacked')  }}
{{ component('Anonymous')  }}

<br>
tag:
{% component ClassBacked  %}
{% endcomponent %}
{% component Anonymous  %}
{% endcomponent %}

<br>
HTML:
<twig:ClassBacked></twig:ClassBacked>
<twig:Anonymous></twig:Anonymous>
```
## Result expected

```text
Function: foo foo
tag: foo foo
HTML: foo foo
```

## Result obtain instead

```text
Function: foo foo
tag: foo bar
HTML: foo bar
```

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #1138 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -120,6 +120,8 @@ final class ComponentRenderer implements ComponentRendererInterface
 
         $component = $mounted->getComponent();
         $metadata = $this->factory->metadataFor($mounted->getName());
+        // expose public properties and properties marked with ExposeInTemplate attribute
+        $props = iterator_to_array($this->exposedVariables($component, $metadata->isPublicPropsExposed()));
         $variables = array_merge(
             // first so values can be overridden
             $context,
@@ -132,9 +134,8 @@ final class ComponentRenderer implements ComponentRendererInterface
 
             // add attributes
             [$metadata->getAttributesVar() => $mounted->getAttributes()],
-
-            // expose public properties and properties marked with ExposeInTemplate attribute
-            iterator_to_array($this->exposedVariables($component, $metadata->isPublicPropsExposed())),
+            $props,
+            ['__props' => $props]
         );
         $event = new PreRenderEvent($mounted, $metadata, $variables);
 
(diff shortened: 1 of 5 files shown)


src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -132,9 +134,8 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
 
             // add attributes
             [$metadata->getAttributesVar() => $mounted->getAttributes()],
-
-            // expose public properties and properties marked with ExposeInTemplate attribute
-            iterator_to_array($this->exposedVariables($component, $metadata->isPublicPropsExposed())),
+            $props,
+            ['props' => $props]
```
weaverryan: So `props` is basically an internal key so that we can decide the correct thing in `PropsNode`. Call it `__props` to make it look more internal?

weaverryan (commented):
This makes sense 👍. Any chance you could hunt down a test for this?

WebMamba:
The failing tests don't look related

weaverryan:
Looks great - thank you Matheo!

PR #1148, opened 2023-09-27
Fix missing double dash in namespaced Stimulus outlets

gregberger:
| Q             | A
| ------------- | ---
| Bug fix?     |  yes
| New feature?  | no 
| Tickets       |  /
| License       | MIT

In `StimulusTwigExtension`, namespaced (ie: in sub directories) stimulus controllers didn't work as the regex in `StimulusAttribute::normalizeKeyName` removes the double-dash needed to mark 
namespace separations. 

As the outlet value is the name of a controller, I suggest to use `normalizeControllerName` from the same class, instead.


weaverryan:
I like it - make sense. Thank you Greg!

PR #1149, opened 2023-09-27
[Site] Fix package "install command" background

smnandre:
It seems i let this ugly visual bug happen in prod... i'm absolutely sorry and I'll make sure it doesn't happen again.

bugbizar


weaverryan:
Thanks Simon!

PR #1151, opened 2023-09-28
[TwigComponent] Support for SubDirectory-named components via namespace config

smnandre:
cc https://github.com/symfony/ux/pull/1140


src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -0,0 +1,335 @@
...
+                $components[] = $componentName;
+            }
+        }
+
+        if ($interactive && \count($components)) {
+            return $io->choice('Select one of the following component to display its information', array_values($components), 0);
+        }
+
+        return null;
+    }
```
weaverryan, line 125: nice method

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -0,0 +1,335 @@
...
+
+        return $components;
+    }
+
+    /**
+     * Return a map of component name => template.
+     *
+     * @return array<string, string>
+     */
+    private function findAnonymousComponents(string $directory = null): array
```
weaverryan: This argument is unused. And I think that makes sense (and I don't think the `dir` option makes sense anymore) now that we have a specific anonymous templates directory.

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -0,0 +1,335 @@
...
+            new TableSeparator(),
+            // ['Attributes Var', $metadata->get('attributes_var')],
+            ['Public Props', $metadata->get('expose_public_props') ? 'Yes' : 'No'],
+            ['Properties', implode("\n", $this->getComponentProperties($metadata))],
+        ]);
+
+        // $properties = $this->getComponentProperties($metadata);
+        // if (\count($properties)) {
+        //    $table->addRow(['Propertires', implode("\n", $properties)]);
+        // }
```
weaverryan: 👀

src/TwigComponent/src/ComponentTemplateFinderInterface.php
```diff
@@ -13,6 +13,8 @@
 
 /**
  * @author Matheo Daninos <matheo.daninos@gmail.com>
+ *
+ * @internal
```
weaverryan: We can't add this after it's already been released

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -30,7 +30,7 @@ final class ComponentFactory
      * @param array<class-string, string> $classMap
      */
     public function __construct(
-        private ComponentTemplateFinderInterface $componentTemplateFinder,
+        private ComponentTemplateFinder $componentTemplateFinder,
```
weaverryan: Should still be the interface, right? As we can't remove the interface

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -67,6 +67,8 @@ public function render(MountedComponent $mounted): string
 
         try {
             if (InstalledVersions::getVersion('twig/twig') < 3) {
+                trigger_deprecation('twig/twig', $version, $message, $args);
+
```
weaverryan: Added by accident?

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -15,33 +15,53 @@
 
 /**
  * @author Matheo Daninos <matheo.daninos@gmail.com>
+ *
+ * @internal
```
weaverryan: Can't add this after the release

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -15,33 +15,53 @@
...
 final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
 {
+    private readonly ?string $directory;
+
     public function __construct(
-        private Environment $environment
+        private Environment $environment,
+        string $directory = null,
     ) {
+        $this->directory = \is_string($directory) ? (rtrim($directory, '/').'/') : null;
```
weaverryan: We should trigger a deprecation is `$directory` is null, right? We want to force that to be passed in the future?

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -15,33 +15,53 @@
...
+        $template = $this->directory.$componentPath.'.html.twig';
+        if ($loader->exists($template)) {
+            return $template;
         }
 
-        if ($loader->exists($componentPath)) {
-            return $componentPath;
+        $template = $this->directory.$componentPath;
+        if ($loader->exists($template)) {
+            return $template;
```
weaverryan, line 58: We could remove this now. Basically, the old, legacy logic is in the if statement above, unchanged. 👍 . So down here, we can do what we want. Does it make sense to have a template name that exactly matches the component name? Would that look like `<twig:foo.html.twig>`? idk...

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
smnandre: I was not sure of the use cases, but i'm inclined to remove it too, and use the "name autocomplete/search" instead.

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
smnandre: Looking at it now, could be simplified (it did more at a moment, using metadata)

src/TwigComponent/src/ComponentFactory.php
smnandre: Hmm.. i'll suggest an other solution then :)

src/TwigComponent/src/ComponentRenderer.php
smnandre: Absolutely (tried some things with deprecations)... sorry

src/TwigComponent/src/ComponentTemplateFinder.php
smnandre: We cannot add an "@internal" on the constructor i guess ? So yes :(

src/TwigComponent/src/ComponentTemplateFinder.php
smnandre: 100%. Shall we restrict the "name" too ? Or is this something we assume is done "upper" ?

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -96,14 +111,78 @@ class_exists(AbstractArgument::class) ? new AbstractArgument(sprintf('Added in %
...
+                            foreach ($v as $namespace => $defaults) {
+                                if (!str_ends_with($namespace, '\\')) {
+                                    throw new InvalidConfigurationException(sprintf('The twig_component.defaults namespace "%s" is invalid: it must end in a "\"', $namespace));
+                                }
+                            }
+
+                            return $v;
+                        })
+                    ->end()
+                    // validate that namespace ends in \
```
weaverryan: We can remove this - probably an old comment I left

src/TwigComponent/tests/Fixtures/Kernel.php
```diff
@@ -52,6 +52,29 @@ protected function configureContainer(ContainerConfigurator $c): void
             'default_path' => '%kernel.project_dir%/tests/Fixtures/templates',
         ]);
 
+        $twigComponentConfig = [];
```
weaverryan, line 55: We should set `anonymous_template_directory` by default to avoid hitting our own deprecations (unless there is a specific test that test for the legacy behavior - that test should get an `@group legacy` and no config)

src/TwigComponent/doc/index.rst
```diff
@@ -56,18 +58,28 @@ Let's get this thing installed! Run:
...
 
-Creating a Basic Component
---------------------------
+```
+# config/packages/twig_component.yaml
+twig_component:
+    defaults:
+        # default component template directory
+        App\Twig\Components\: 'components/'
+```
```
weaverryan, line 73: I used the totally wrong syntax here. And, we need to add the anonymous components config

weaverryan:
Thank you Simon!

PR #1152, opened 2023-09-29
[TwigComponent][WebProfiler] Add profile + StopWatch + WDT

smnandre:
Add a DataCollector to add data in the web debug toolpar and a profiler pane.

Based on the events (`PreRender` / `PostRender`) triggerd during the request.

| Overview | Details | 
| - | - | 
| overview | details | 

| Anonymous components | Stopwatch + performance | 
| - | - | 
| anonymous-timing | perforlmance |

### DIFF
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -11,13 +11,16 @@
 
 namespace Symfony\UX\TwigComponent\DependencyInjection;
 
+use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\AbstractArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\LogicException;
 use Symfony\Component\DependencyInjection\Extension\Extension;
+use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 use Symfony\Component\DependencyInjection\Parameter;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\Stopwatch\Stopwatch;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 use Symfony\UX\TwigComponent\Command\ComponentDebugCommand;
 use Symfony\UX\TwigComponent\ComponentFactory;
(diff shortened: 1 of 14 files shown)


src/TwigComponent/templates/Collector/twig_component.html.twig
```diff
@@ -0,0 +1,305 @@
...
+            font-weight: normal !important;
+        }
+    </style>
+{% endblock %}
+
+{% block toolbar %}
+    {% if collector.renderCount %}
+
+        {% set icon %}
+            {{ source('@TwigComponent/Collector/icon.svg') }}
```
weaverryan, line 124: Not sure if this is ACTUALLY a good idea 😅 , but what about a custom logic that is the Twig logo inside of the `<>`.

So the take the leaf on the left of this and put it inside our `<>` - so kinda short for `<twig/>`.
Screenshot 2023-10-02 at 12 55 28 PM

src/TwigComponent/templates/Collector/twig_component.html.twig
```diff
@@ -0,0 +1,305 @@
...
+        </tr>
+        </thead>
+        <tbody>
+            {% for component in collector.components %}
+                <tr>
+                    <td>{{ component.name }}</td>
+                    <td>
+                        {% if component.class == 'Symfony\\UX\\TwigComponent\\AnonymousComponent' %}
+                            <pre class="sf-dump"><span class="text-muted">AnonymousComponent</span></pre>
+                        {% else %}
```
weaverryan, line 227: Testing this on the homepage of ux.symfony.com, it *seems* like all children of an anonymous component are incorrectly seen as anonymous themselves. Notice here the `Icon` on top are correct, then below, it suddenly looks like `Icon` is anonymous.
Screenshot 2023-10-02 at 12 57 02 PM

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -105,5 +110,19 @@ class_exists(AbstractArgument::class) ? new AbstractArgument(sprintf('Added in %
...
+        }
+
+        $container->register('console.command.stimulus_component_debug', ComponentDebugCommand::class)
+            ->setArguments([
+                new Parameter('twig.default_path'),
+                new Reference('ux.twig_component.component_factory'),
+                new Reference('twig'),
+               tagged_iterator('twig.component'),
+            ])
+            ->addTag('console.command')
```
weaverryan: This is also still registered above

weaverryan (commented):
This is QUITE impressive 👏. I've left some minor notes, but I think we should get this in. I'm sure we will think of other info that will be useful, but wow - was this fun to play with!

src/TwigComponent/templates/Collector/twig_component.html.twig
smnandre: Like the idea really much, tried to do it you'll tell me

weaverryan:
Thank you for this fantastic PR Simon!

PR #1154, opened 2023-10-01
Fix adding the mapped controller asset as dependency

andersonamuller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

After updating to the latest changes from AssetMapper component.

Not sure, this is the correct fix, but I got the error that the `AssetDependency` class was not defined and then changing the line 64 solved for me.

### DIFF
--- a/src/StimulusBundle/src/AssetMapper/StimulusLoaderJavaScriptCompiler.php
+++ b/src/StimulusBundle/src/AssetMapper/StimulusLoaderJavaScriptCompiler.php
@@ -62,11 +62,16 @@ class StimulusLoaderJavaScriptCompiler implements AssetCompilerInterface
              * and mark it as a "content" dependency so that this file's contents
              * will be recalculated when the contents of any controller changes.
              */
-            $asset->addDependency(new AssetDependency(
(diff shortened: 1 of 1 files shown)


weaverryan:
This is the correct change for Symfony 6.4 :). To keep things working with symfony 6.3, let's make this conditional. If the `AssetDependency` class exists, use the old code. Else use the new code. Can you make that change?

andersonamuller:
> This is the correct change for Symfony 6.4 :). To keep things working with symfony 6.3, let's make this conditional. If the `AssetDependency` class exists, use the old code. Else use the new code. Can you make that change?

@weaverryan Is this what you had in mind? Sorry for the missing tests, don't have much time to dig on that now.

weaverryan:
Nailed it! Just an extra, empty line in there now ;). 

There is also one test failing on StimulusBundle - it could be unrelated, I just can't tell - that needs to be debugged to see what the underlying problem is.

weaverryan:
Thanks Anderson!

PR #1155, opened 2023-10-02
Handle array-like objects when working with checkboxes

Lustmored:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | 
| License       | MIT

First of all please take a look at added test cases. I think it shows the purpose of this change clearly.

Background:
Recently when working with forms inside a component I ran into a corner case, where Symfony forms are setting a non-zero-indexed array for the properties of `ChoiceType` with both `multiple` and `expanded`. It breaks live-component model assumption that multiple checkboxes keep their value as an array, which in JavaScript is always zero-indexed.

It happens after a pretty weird set of steps:
- check any option (foo)
- check any other option (bar)
- check the third option (baz)
- uncheck option 'foo' <- Symfony forms store current value as `[1 => 'bar', 2 => 'baz']`
- uncheck option 'bar' <- at this point current implementation is treating checkboxes as a single value and send back `null` instead of `['baz']`.

This simple change fixes this flow (tested on a real life project). I don't see any correlation with failing tests.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -162,6 +162,9 @@ function getValueFromElement(element, valueStore) {
                 if (Array.isArray(modelValue)) {
                     return getMultipleCheckboxValue(element, modelValue);
                 }
+                else if (Object(modelValue) === modelValue) {
+                    return getMultipleCheckboxValue(element, Object.values(modelValue));
+                }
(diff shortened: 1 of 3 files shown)


weaverryan:
Very-well explained and excellent tests - THANK YOU. Check boxes are deceivingly complex

PR #1157, opened 2023-10-03
[LiveComponent] Fix DTO hydration from phpdoc typehints

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | N/A
| License       | MIT

To better understand - see the test. Previously typehints from phpdoc was just ignored. Not sure is this implementation is good enough, but it works.

### DIFF
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -12,6 +12,7 @@
 namespace Symfony\UX\LiveComponent\Metadata;
 
 use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
+use Symfony\Component\PropertyInfo\Type;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
 use Symfony\UX\TwigComponent\ComponentFactory;
 
@@ -54,12 +55,12 @@ class LiveComponentMetadataFactory
                 continue;
             }
 
-            if (isset($metadatas[$property->getName()])) {
+            if (isset($metadatas[$propertyName = $property->getName()])) {
                 // property name was already used
                 continue;
             }
 
-            $metadatas[$property->getName()] = $this->createLivePropMetadata($class->getName(), $property->getName(), $property, $attribute->newInstance());
+            $metadatas[$propertyName] = $this->createLivePropMetadata($class->getName(), $propertyName, $property, $attribute->newInstance());
         }
 
         return array_values($metadatas);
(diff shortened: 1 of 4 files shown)


src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -78,9 +84,17 @@ public function createLivePropMetadata(string $className, string $propertyName,
...
-        $type = $property->getType();
-        if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
-            throw new \LogicException(sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
+        if (null === $type && null === $collectionValueType && isset($infoTypes[0])) {
+            $typeClass = $infoTypes[0]->getClassName();
+
+            return new LivePropMetadata(
+                $property->getName(),
+                $liveProp,
+                $typeClass,
```
weaverryan: If the phpdoc is, for example, an `int`, this argument should be passed `int`, not null. We need to check if the `Type` (i.e.`$infoTypes[0]`) is has a `getBuiltinType()` of `BUILTIN_TYPE_OBJECT`, then either pass `->getClassName()` here or `getBuiltinType()`. 

Also, I don't love choosing the `[0]` index. Does that mean that `@var null|ChildDTO` would fail? Is the `0` index the `null` type in that case? Could we loop over all of them to try to grab the "strongest" (non-null) type we can find?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -78,9 +84,17 @@ public function createLivePropMetadata(string $className, string $propertyName,
             }
         }
 
-        $type = $property->getType();
-        if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
-            throw new \LogicException(sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
+        if (null === $type && null === $collectionValueType && isset($infoTypes[0])) {
```
weaverryan, line 88: Is it possible that `$type` is null, but we DO have a `$collectionValueType`? Maybe gave `/** @var ChildDto[] */` but no `array` type on the field?

weaverryan (commented):
I have some thoughts, but yea, this looks like the right direction.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
norkunas: I've tested the property info with `null|*`case and then it always renders the last parts.
For example:
```
// @var null|int
array:1 [
  0 => Symfony\Component\PropertyInfo\Type {#48476
    -builtinType: "int"
    -nullable: true
    -class: null
    -collection: false
    -collectionKeyType: []
    -collectionValueType: []
  }
]

// @var null|\stdClass
array:1 [
  0 => Symfony\Component\PropertyInfo\Type {#48476
    -builtinType: "object"
    -nullable: true
    -class: "stdClass"
    -collection: false
    -collectionKeyType: []
    -collectionValueType: []
  }
]

// @var null|\stdClass|customobject
array:2 [
  0 => Symfony\Component\PropertyInfo\Type {#36519
    -builtinType: "object"
    -nullable: true
    -class: "stdClass"
    -collection: false
    -collectionKeyType: []
    -collectionValueType: []
  }
  1 => Symfony\Component\PropertyInfo\Type {#48476
    -builtinType: "object"
    -nullable: false
    -class: "customobject"
    -collection: false
    -collectionKeyType: []
    -collectionValueType: []
  }
]
```

just strange thing that in the last example first type is nullable and second is not (and this is always, if you use 3 or more union types then only first one is nullable)

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
norkunas: But in this case we don't care, do we? `/** @var ChildDto[] */` is covered by tests already :)

weaverryan:
Thanks Tomas!

Issue #1129, opened 2023-09-21
Not possible to use options defined by `AutocompleteChoiceTypeExtension` when using a `AsEntityAutocompleteField` field


bendavies:
Hi,

If you define an `AsEntityAutocompleteField` as per [the docs](https://symfony.com/bundles/ux-autocomplete/current/index.html#usage-in-a-form-with-ajax), but add some `tom_select_options` at point of use:

<details>
<summary>FoodAutocompleteField.php</summary>

```php
<?php

// src/Form/FoodAutocompleteField.php
// ...

use Symfony\UX\Autocomplete\Form\AsEntityAutocompleteField;
use Symfony\UX\Autocomplete\Form\ParentEntityAutocompleteType;

#[AsEntityAutocompleteField]
class FoodAutocompleteField extends AbstractType
{
    public function configureOptions(OptionsResolver $resolver)
    {
        $resolver->setDefaults([
            'class' => Food::class,
            'placeholder' => 'What should we eat?',
        ]);
    }

    public function getParent(): string
    {
        return ParentEntityAutocompleteType::class;
    }
}
```
</details>

<details>
<summary>SomeForm.php</summary>

```php
<?php

// src/Form/SomeForm.php
// ...

class SomeForm extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder->add('food', FoodAutocompleteField::class, [
            'label' => 'Food'',
            'tom_select_options' => [
                'plugins' => ['dropdown_input'],
            ],
        ]);
    }

}
```
</details>

You will receive:

```
An error has occurred resolving the options of the form "App\Form\FoodAutocompleteField": The option "tom_select_options" does not exist
```

Thanks,
Ben

weaverryan:
I haven’t looked quickly enough yet to see if this helps, but we have a refactoring of the autocomplete form internals here https://github.com/yceruto/symfony-ux/tree/autocomplete - it needs a champion to check this and create a PR using it. yceruto gave me the code with the understanding that we would take it and integrate it.

if anyone wants to take this on, that would be great. If it wouldn’t solve this issue anyway, sorry for the noise :)

bendavies:
finally looked at this. the above PR doesn't fix the the issue, unfortuantely.

bendavies:
@norkunas do you know where i might start hunting this down?

bendavies:
ah, this issue is fixed, but now we get a different error: https://github.com/symfony/ux/issues/1128

PR #1158, opened 2023-10-03
[Autocomplete] New autocomplete type

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | #1129 
| License       | MIT

Continuation of work from @yceruto.

One thing that bothers me: in the `ParentEntityAutocompleteType` we have `getBlockPrefix(): { return 'ux_entity_autocomplete'; }`. If we'd add this to the new one, it'd break my custom form theme because I access `form.autocomplete.vars..`, should we set something new for block prefix?

### DIFF
--- a/src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
+++ b/src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
@@ -140,6 +140,7 @@ final class AutocompleteExtension extends Extension implements PrependExtensionI
 
         $container
             ->register('ux.autocomplete.entity_type', ParentEntityAutocompleteType::class)
+            ->setDeprecated('symfony/ux-autocomplete', '2.13', 'The "%service_id%" form type is deprecated since 2.13. Use "ux.autocomplete.base_entity_type" instead.')
             ->setArguments([
                 new Reference('router'),
             ])
--- a/src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
+++ b/src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
@@ -27,6 +27,7 @@ use Symfony\UX\Autocomplete\Doctrine\EntityMetadataFactory;
 use Symfony\UX\Autocomplete\Doctrine\EntitySearchUtil;
 use Symfony\UX\Autocomplete\Form\AsEntityAutocompleteField;
 use Symfony\UX\Autocomplete\Form\AutocompleteChoiceTypeExtension;
+use Symfony\UX\Autocomplete\Form\BaseEntityAutocompleteType;
 use Symfony\UX\Autocomplete\Form\ParentEntityAutocompleteType;
 use Symfony\UX\Autocomplete\Form\WrappedEntityTypeAutocompleter;
 use Symfony\UX\Autocomplete\Maker\MakeAutocompleteField;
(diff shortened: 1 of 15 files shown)


weaverryan (commented):
Well... this is incredible. I tested locally - it works fantastically! So, I just think we need to:

A) Deprecate the old stuff - so `ParentEntityAutocompleteType` and `AutocompleteEntityTypeSubscriber`
B) Update the docs to show `BaseEntityAutocompleteType`

> One thing that bothers me: in the ParentEntityAutocompleteType we have getBlockPrefix(): { return 'ux_entity_autocomplete'; }. If we'd add this to the new one, it'd break my custom form theme because I access form.autocomplete.vars.., should we set something new for block prefix?

Ah, I see what you're saying: if we return the same string, and your app has a mixture of old and new style fields, your one form theme will need to work for both situations. Are we sure we need to worry about this? When you start using the new way, you'll need to update your form theme anyway. And when you do, you can update all of your old usages to the new `BaseEntityAutocompleteType` or make your form theme temporarily a bit smarter to check to see if `form.autocomplete` is defineed. I'd vote for returning the same block prefix.

norkunas:
Updated

src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
```diff
@@ -40,14 +41,6 @@ final class AutocompleteExtension extends Extension implements PrependExtensionI
 {
     public function prepend(ContainerBuilder $container)
     {
-        $bundles = $container->getParameter('kernel.bundles');
-
-        if (isset($bundles['TwigBundle'])) {
-            $container->prependExtensionConfig('twig', [
-                'form_themes' => ['@Autocomplete/autocomplete_form_theme.html.twig'],
-            ]);
-        }
```
weaverryan: Ok, I think I misspoke & misunderstood what you were saying about the form theme stuff. We need to keep this (for the "old" system). For the new type, I'm leaning towards not returning ANY special block prefix and just allowing the normal one to be used for the `entity` type. If the user wants to theme something specifically for autocomplete entity types, I think they could probably do that via the `uses_autocomplete` form theme variable.

src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
```diff
@@ -130,6 +123,13 @@ private function registerBasicServices(ContainerBuilder $container): void
...
     {
+        $container
+            ->register('ux.autocomplete.base_entity_type', BaseEntityAutocompleteType::class)
+            ->setArguments([
+                new Reference('router'),
+            ])
+            ->addTag('form.type');
+
         $container
             ->register('ux.autocomplete.entity_type', ParentEntityAutocompleteType::class)
```
weaverryan, line 142: Can we deprecate this service? Or does that trigger a deprecation SIMPLY because it's a valid form type (even if it's unused)?

weaverryan (commented):
Can you also update the docs? Probably updating all of the `ParentEntityAuto...` -> `BaseEntityAuto...` would be enough. Maybe also a `versionadded` to mention that `BaseEntityAuto...` is new in 2.13.

src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
norkunas: I've restored the template but **not to remove `ux_entity_autocomplete`** block prefix, I've added checks to the template if we use new type or the old one and adapt

src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
norkunas: deprecated

weaverryan (approved):
👍 Can you rebase and squash a little? I don't need 1 commit - but 2-3 would be fine. I can normally squash with our merge tool, but not with 2 authors - and we need a squash to remove that merge commit :)

norkunas:
Rebase & squash done

weaverryan:
Thank you Tomas you seriously rock!

PR #1163, opened 2023-10-04
[LiveComponent] Add debug:live-component command

smnandre:
Similar to the TwigComponent one, with additional information:

* LiveProps
* LiveActions
* Pre/Post hydrate & render

### DIFF
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -48,8 +48,10 @@ final class AsLiveComponent extends AsTwigComponent
 
     /**
      * @internal
+     *
+     * @param object|class-string $component
      */
-    public static function isActionAllowed(object $component, string $action): bool
+    public static function isActionAllowed(object|string $component, string $action): bool
     {
         foreach (self::attributeMethodsFor(LiveAction::class, $component) as $method) {
             if ($action === $method->getName()) {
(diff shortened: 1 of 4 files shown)


kbond:
A second command feels a bit confusing as `debug:twig-component` does show live components.

Wonder if we should combine into `debug:twig-component` and have a _descriptor_ system that takes care of the extra stuff live components need to show:
1. `TwigComponentDescriptor` (lives in `ux-twig-component`)
2. `LiveComponentDescriptor extends TwigComponentDescriptor` (lives in `ux-live-component`)

WebMamba:
I agree with @kbond it look like we're going to have some duplication with the TwigComponent debug command see: https://github.com/symfony/ux/pull/1088#discussion_r1318235694 We already looked for LiveProps and LiveAction. We don't have  Pre/Post hydrate and render, and yes It can be interesting to add. But I my opinion is better to keep everything in the TrigComponent command.

smnandre:
I'm not a big fan to bring "LiveComponent" specific code into the TwigComponent bundle to be honest... but that'll be the perfect occasion to test something i wanted to do (multiple factory)

smnandre:
I'll split the PR in two then

* one to try the system @kbond suggested
* one to update the AsLiveComponent static methods signature that i'll need sooner

kbond:
> I'm not a big fan to bring "LiveComponent" specific code into the TwigComponent bundle to be honest...

I think my idea would remove this. Assuming the TwigComponentDescriptor is a service, the live bundle could swap it out for the live-specific one.

smnandre:
Closing this one (i'll use your ideas to improve the twig debug command in a near future @kbond )

PR #1164, opened 2023-10-04
[Site] Use <twig: HTML syntax in demos

smnandre:
* Replace remaining `{{ component(... }}` with `<twig:.. />`
* Fix a route in product demo

### DIFF
--- a/ux.symfony.com/templates/components/InvoiceCreator.html.twig
+++ b/ux.symfony.com/templates/components/InvoiceCreator.html.twig
@@ -44,12 +44,12 @@
                     </thead>
                     <tbody>
                         {% for key, line in lineItems %}
-                            {{ component('InvoiceCreatorLineItem', {
-                                key: key,
-                                productId: line.productId,
-                                quantity: line.quantity,
-                                isEditing: line.isEditing,
-                            }) }}
+                            <twig:InvoiceCreatorLineItem
+                                key="{{ key }}"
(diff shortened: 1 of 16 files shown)


ux.symfony.com/templates/components/InvoiceCreator.html.twig
```diff
@@ -44,12 +44,12 @@
...
-                                key: key,
-                                productId: line.productId,
-                                quantity: line.quantity,
-                                isEditing: line.isEditing,
-                            }) }}
+                            <twig:InvoiceCreatorLineItem
+                                key="{{ key }}"
+                                productId="{{ line.productId }}"
+                                quantity="{{ line.quantity }}"
+                                isEditing="{{ line.isEditing }}"
```
weaverryan, line 51: We should decide whether, in general, we show `key="{{ key }}` or `:key="key"`.

The first has better Twig highlighting. But... I  can't remember if it works for arrays/objects - e.g. `invoice="{{ invoice }}"`. At the very least, that "looks funny" (printing an object... which won't really be printed).

smnandre: No strong position on this... 

I prefer the first one (easier to search / refactor / etc) ... which seems more natural to me... 
I find it differenciate more hard-coded text VS evaluated vars...

weaverryan:
Thanks Simon!

PR #1179, opened 2023-10-11
Password Strength Controller

Spomky:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | none
| License       | MIT

This PR aims at adding a new controller for password strength estimation.
The method is the same as the one used by the Symfony constraint `PasswordStrength`.

- [x] Component
- [x] Tests
- [x] Documentation

```html
<form {{ stimulus_controller('@symfony/ux-password-strength') }}>
    <div>
        <label for="password">Password</label>
        <input id="password" name="password" type="password" {{ stimulus_action('@symfony/ux-password-strength', 'estimatePasswordStrength') }}>
    </div>
    <div class="wrap">
        <div class="item item0" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
        <div class="item item1" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
        <div class="item item2" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
        <div class="item item3" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
        <div class="item item4" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
    </div>
    <span class="text" {{ stimulus_target('@symfony/ux-password-strength', 'message') }}></span>
    <div>
        <span class="scoreBox" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></span>
    </div>
</form>
```

<details>

```css
label {
    display: block;
    position: relative;
    text-transform: uppercase;
    color: #aaa;
}

input[type='password'] {
    width: 500px;
    box-sizing: border-box;
    height: 55px;
    display: inline-block;
    border: 3px solid #2F96EF;
    border-radius: 5px;
    padding: 0 15px;
    margin: 10px 0;
    transition: .2s;
}

input[type='password']:focus {
    outline: none;
}

.wrap {
    width: 300px;
    height: 5px;
    background: #F6F6FA;
    display: inline-block;
    vertical-align: middle;
    overflow: hidden;
    border-radius: 5px;
}

.item {
    display: inline-block;
    height: 100%;
    width: 20%;
    float: left;
    visibility: hidden;
    transition: background-color .2s, visisility .1s;
}

.item0[data-password-strength-estimate] {
    background-color: #FF4B47;
    visibility: visible;
}

.item1[data-password-strength-estimate="1"], .item1[data-password-strength-estimate="2"], .item1[data-password-strength-estimate="3"], .item1[data-password-strength-estimate="4"] {
    background-color: #C98300;
    visibility: visible;
}

.item2[data-password-strength-estimate="2"],.item2[data-password-strength-estimate="3"],.item2[data-password-strength-estimate="4"] {
    background-color: #ffee29;
    visibility: visible;
}

.item3[data-password-strength-estimate="3"],.item3[data-password-strength-estimate="4"] {
    background-color: #00ec4e;
    visibility: visible;
}

.item4[data-password-strength-estimate="4"] {
    background-color: #2DAF7D;
    visibility: visible;
}

.text {
    display: inline-block;
    color: #8f8f8f;
    margin-left: 5px;
    transition: .2s;
}

.scoreBox {
    display: inline-block;
    width: 50px;
    height: 50px;
    border-radius: 50%;
    background: #F6F6FA;
    margin-left: 10px;
    vertical-align: middle;
    text-align: center;
    line-height: 50px;
    font-size: 20px;
    color: #000;
    font-weight: bold;
    transition: .2s;
}

.scoreBox[data-password-strength-estimate="0"] {
    background-color: #fa6f6d;
}
.scoreBox[data-password-strength-estimate="0"]:after {
    content: '😬';
}

.scoreBox[data-password-strength-estimate="1"] {
    background-color: #C98300;
}
.scoreBox[data-password-strength-estimate="1"]:after {
    content: '🤔';
}

.scoreBox[data-password-strength-estimate="2"] {
    background-color: #ffee29;
}
.scoreBox[data-password-strength-estimate="2"]:after {
    content: '😅';
}

.scoreBox[data-password-strength-estimate="3"] {
    background-color: #00ec4e;
}
.scoreBox[data-password-strength-estimate="3"]:after {
    content: '🙂';
}

.scoreBox[data-password-strength-estimate="4"] {
    background-color: #2DAF7D;
}
.scoreBox[data-password-strength-estimate="4"]:after {
    content: '😄';
}
```

</details>

Animation


94noni:
nice addition, do you think we can replace the word password with text or string?
I mean, it could be totally legit to have such logic on a text field to help user to create a strong string (a token or api key, etc etc for example)

Spomky:
Hi @94noni,

I'm not sure this strength estimate is applicable to anything other than a password.
The API keys or tokens you mentioned are either self-contained structures (JWT, Paseto, SAML, etc.) or strings by reference.
* The first type is not concerned here due to the existing means (nonce, token ID, digital signature/hash, timeframe of use, etc.).
* To me, the second type doesn't seem to be concerned either. Most of the time, these are (unique) database identifiers or randomly generated binary strings with a high level of entropy (16 bytes or more).

Spomky:
Hi,

It is almost ready. I will add a section in the documentation for mentioning an event is sent so you can catch it and adapt the UI depending on the score (e.g. disable the form submit button).
In the meantime, let me know if you have any suggestion.

94noni:
indeed, I now see your point @Spomky 
to me the PR looks good :)

weaverryan:
Hey @Spomky!

Before I comment on the PR (it's definitely a nice feature), we need to talk more broadly about UX package that don't contain PHP code. Providing some Stimulus controllers like this to solve a problem is, I think, fantastic. However, there's some overhead each time we add a new UX package: we need a new GitHub repo with sub-tree split, a new Composer package & a new docs page on symfony.com/bundles.

Ideally, we would have another option that's less overhead. That would make life easier, and also allow us to add even more things like this. I don't have a plan in mind exactly, but something like:

A) We have a way to have some pure JS packages that are not published as Composer packages, but are added/published as true npm packages
B) But then how do we install these? Do we just document running `npm add @symfony/ux-foo`/`importmap:require @symfony/ux-foo` and then have the user manually add these controllers to Stimulus? Or is there still some way to hook into `controllers.json` so they can be added automatically?
C) We document these in some aggregate spot - some place to find Stimulus controllers. Honestly, there are many other Stimulus controllers around the internet (just look at https://www.stimulus-components.com/) and it might be nice to at least point to *those* as well.

I'm open to thoughts. As a maintainer, I want these packages to be easy to add and with low maintenance (the PR contains 19 files, but only 5-8 are actually needed). As a user, I want these things to be easy to find and install.

smnandre:
I feel exactly the same: i like it, very much, but i'm not sure i'd want to install a bundle, especially if there is no PHP code coming from it.

I'll still let you some _opiniated_ comments (so take them only for what they are: opinions)

(because even if it's not as a bundle, i'm 100% sure @weaverryan will find a good way to distribute such contribution in the UX initiative)

src/PasswordStrength/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Fabien Potencier
```
smnandre: 2023-present

src/PasswordStrength/README.md
```diff
@@ -0,0 +1,34 @@
+# Symfony UX Password Strength
+
+Symfony UX Password Strength is a Symfony bundle integrating a password strength estimator in
+Symfony applications. It is part of [the Symfony UX initiative](https://symfony.com/ux).
+
+The password strength estimation method is the same as the default Symfony Constraint `PasswordStrength`.
```
smnandre: Link?

src/PasswordStrength/assets/dist/controller.js
```diff
@@ -0,0 +1,111 @@
...
+        const charCount = {};
+        for (const char of password) {
+            const charCode = char.charCodeAt(0);
+            charCount[charCode] = (charCount[charCode] || 0) + 1;
+        }
+        return charCount;
+    }
+}
+default_1.values = {
+    veryWeakMessage: { type: String, default: 'The password strength is too weak!' },
```
smnandre: Isnt "too weak" a bit judgmental ? 

(same thing for all the exclamation marks)

src/PasswordStrength/composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "autoload": {
+        "psr-4": {
+            "Symfony\\UX\\Typed\\": "src/"
```
smnandre: ~Typed~

```json
"Symfony\\UX\\PaswordMeter\\"
```

src/PasswordStrength/doc/Animation.gif
smnandre, line 1: I'd add a table of examples, to illustrate that length does not imply strength.

Something like that (dummy data)

| Password | Length | Strength | 
| - | - | - | 
| 123456 | 🟦⬜️⬜️  | 🟥⬜️⬜️⬜️⬜️
| é±%$§ | 🟦⬜️⬜️   | 🟩🟩🟩🟩⬜️
| Space Crocodile | 🟦🟦⬜️  | 🟥🟥⬜️⬜️⬜️
| Space Crocodi!0 | 🟦🟦⬜️  | 🟧🟧🟧⬜️⬜️
| é±%$§ s@U•? | 🟦🟦⬜️  | 🟩🟩🟩🟩🟩
| Giant Funny Space Crocodile | 🟦🟦🟦  |  🟩🟩🟩🟩⬜️

src/PasswordStrength/doc/index.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. code-block:: html+twig
+
+    <form {{ stimulus_controller('@symfony/ux-password-strength') }}>
+        <label for="password">Password</label>
+        <input id="password" name="password" type="password" {{ stimulus_action('@symfony/ux-password-strength', 'estimatePasswordStrength') }}>
+        <div>The score is: <span {{ stimulus_target('@symfony/ux-password-strength', 'score') }}></span></div>
+        <div>The message is: <span {{ stimulus_target('@symfony/ux-password-strength', 'message') }}></span></div>
+        <div id="jauge" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
+        <div id="progressbar" {{ stimulus_target('@symfony/ux-password-strength', 'meter') }}></div>
```
smnandre, line 53: Why both are targetted ?

src/PasswordStrength/LICENSE
Spomky: 👍🏽

src/PasswordStrength/README.md
Spomky: I added a link to the constraint page

src/PasswordStrength/assets/dist/controller.js
Spomky: I changed the message to `The password strength is very weak.`.
Indeed it may be reviewed again. In any case it can be changed by the users.

src/PasswordStrength/composer.json
Spomky: Comment addressed

src/PasswordStrength/doc/Animation.gif
Spomky: I like the idea. Or it could be added to the demo https://ux.symfony.com/

src/PasswordStrength/doc/index.rst
Spomky: I wanted to show it is possible to target more than one spot.
I changed the example where 2 different ways to show the score are visible.

Spomky:
> we need to talk more broadly about UX package that don't contain PHP code.

Yes indeed, it's a bit weird to have a bundle to just update the AssetMapper paths.

For me, the best would be to use composer and console commands as much as possible. But honestly, I have no idea on how to improve the way components could be shared/detected.
What about a composer plugin to turn a stimulus component into a package?

smnandre:
I think we should do something similar to Notifier / Twig-Extra (simple packages who are not bundle but simple classes with expected interfaces) ... or even just JS packages in some cases...

But that require a bit of time/thinking to not rush into something we'll regret / suffer from later.

I have some ideas i'd like to run by @weaverryan in the next few days (i need to know what we "can" do or not, because some of them would probably require tools or methods the symfony organisation does not currently use, as far as i know)

About the AssetMapper paths, i do think we should inverse the direction ("don't call us, we'll call you").. 
... same thing, this is not related specifically to your proposal but it could be the right time to put all that "on the table" :)

smnandre:
(just keeping that idea there before you mark it resolved in the review comments)


| Password | Length | Strength | 
| - | - | - | 
| 123456 | 🟦⬜️⬜️  | 🟥⬜️⬜️⬜️⬜️
| é±%$§ | 🟦⬜️⬜️   | 🟩🟩🟩🟩⬜️
| Space Crocodile | 🟦🟦⬜️  | 🟥🟥⬜️⬜️⬜️
| Space Crocodi!0 | 🟦🟦⬜️  | 🟧🟧🟧⬜️⬜️
| é±%$§ s@U•? | 🟦🟦⬜️  | 🟩🟩🟩🟩🟩
| Giant Funny Space Crocodile | 🟦🟦🟦  |  🟩🟩🟩🟩⬜️

src/PasswordStrength/doc/Animation.gif
Spomky: Comment addressed.
Note that I removed the column for password length because it could be confusing.
The length is used by the estimation function and has an influence on the score computation.

Edit: I really like the table with the colored squares. It's easy to understand.

seb-jean:
Hi,
Is this feature still in progress?

Spomky:
The feature is done (unless new comments from reviewers).
The problem is about the component distribution (see https://github.com/symfony/ux/pull/1179#issuecomment-1764990007).

Ping @weaverryan @smnandre @WebMamba is there any idea about that. Let's have a discussion to move forward.

src/PasswordStrength/assets/src/controller.ts
```diff
@@ -0,0 +1,127 @@
...
+                        break;
+                    case charCode >= 97 && charCode <= 122:
+                        lower += count * 26;
+                        break;
+                    case charCode >= 128:
+                        other += count * 128;
+                        break;
+                    default:
+                        symbol += count * 33;
+                }
```
cvergne:
```suggestion

                switch (true) {
                    case charCode < 32 || charCode === 127:
                        control = 33;
                        break;
                    case charCode >= 48 && charCode <= 57:
                        digit = 10;
                        break;
                    case charCode >= 65 && charCode <= 90:
                        upper = 26;
                        break;
                    case charCode >= 97 && charCode <= 122:
                        lower = 26;
                        break;
                    case charCode >= 128:
                        other = 128;
                        break;
                    default:
                        symbol = 33;
                }
```

Number of occurrences of each type of character must not be used to entropy calculation.
See :
- https://github.com/symfony/symfony/blob/5876c48757182da02cbea275a40059b5bfc83854/src/Symfony/Component/Validator/Constraints/PasswordStrengthValidator.php#L65-L77
- https://proton.me/blog/what-is-password-entropy

cvergne (changes requested):
Hi,
While searching for a frontend implementation of the password strength validation currently in Symfony, I've notice when testing your method that I had different result in backend side (via Symfony).
Then comparing both methods I've noticed an error in the entropy calculation, so here is the feedback ;)
(thanks for your work by the way)

src/PasswordStrength/assets/src/controller.ts
Spomky: Hello @cvergne,

Indeed the proposal was mot in line with the Symfony constraint.
I will also take the opportunity to rebase this PR.

Regards

Kocal:
Thanks for the contribution!

However, I'm gonna close this PR because this kind of package (a PHP glue and some opinionated JS/Twig/CSS files) is not something we want anymore in UX 3.

We already started to deprecate some packages, and instead they will be part of the Symfony UX Toolkit.

Thanks for your comprehension :)

Spomky:
Hi @Kocal,

Thank you for letting me know.
To be honest, I really believe this is the right decision. I noticed you've also deprecated other JavaScript-only packages like Typed.
There is no real reason to manage such packages here. After all they are just single files that can easily be stored and maintained elsewhere.

PR #1182, opened 2023-10-12
[CI] Fix `jsxBracketSameLine is deprecated`

smnandre:
Replace `jsxBracketSameLine` with `bracketSameLine`

See: https://prettier.io/docs/en/options.html#deprecated-jsx-brackets


weaverryan:
Thanks Simon! I've been seeing this forever, but never took the time to find the fix!

PR #1184, opened 2023-10-13
[CI][Turbo] Fix direct deprecation notice

smnandre:
Fix this deprecation triggered during tests with Doctrine >= 2.8

> Not enabling lazy ghost objects is deprecated and will not be supported in Doctrine ORM 3.0. Ensure Doctrine\ORM\Configuration::setLazyGhostObjectEnabled(true) is called to enable them.

(Failure unrelated)


weaverryan:
Thanks Simon!

PR #1185, opened 2023-10-13
Fix ExtraLazyChoiceLoader

yceruto:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | -
| License       | MIT

Currently, when default data is provided through the autocomplete field, it is not validated against the database or custom query builder. These changes aim to fix that issue and also simplify the code, while the goal remains the same.


weaverryan:
Thanks Yonel!

PR #1187, opened 2023-10-13
[StimulusBundle][CI] Fix Stimulus/AssetMapper tests

smnandre:


src/StimulusBundle/tests/fixtures/assets/controllers.json
```diff
@@ -14,9 +14,7 @@
                 "autoimport": {
                     "in/asset/mapper/controller_second1.css": true,
                     "in/asset/mapper.controller_second2.css": false,
-                    "@fake-vendor/ux-package1/dist/styles.css": true,
-                    "needed-vendor/file.css": true,
-                    "@scoped/needed-vendor/the/file2.css": true
+                    "@fake-vendor/ux-package1/dist/styles.css": true
```
weaverryan, line 17: I think we're avoiding an underlying bug with this change. The problem is that `autoimport` that point to an external package no longer work. Example - https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/package.json#L16

Previously, we would try to generate a CDN URL from the `url` key in the `importmap.php` - https://github.com/symfony/ux/blob/2.x/src/StimulusBundle/src/Twig/UxControllersTwigRuntime.php#L105-L107

But it's not as simple as fixing that. We need to allow (probably by default) users to download these vendor files locally. Probably the flow will need to look like this, but I still haven't thought too deeply about it:

A) When installing the package `symfony/flex`, reads the `autoimport` from `package.json` and adds these as `importmap.php` entries. So, in the case of tom-select, it would run `importmap:require tom-select/dist/css/tom-select.default.css`. Then, our logic would point to that asset.

B) But, we probably need to go a bit further :). With AssetMapper 6.4, the `ux_controller_link_tags()` probably isn't needed anymore and should do nothing. Instead, we can conditionally generate `import` statements for the autoimport CSS files when generating the dynamic `controllers.js` - https://github.com/symfony/ux/blob/2.x/src/StimulusBundle/src/AssetMapper/StimulusLoaderJavaScriptCompiler.php

smnandre: I'll need a bit more details on how this would "cooperate" (stimulus, assets, importmap) when you have time :)

weaverryan:
All fixed now in another PR!

PR #1188, opened 2023-10-14
Add missing LICENCE file

smnandre:


weaverryan:
Thank you! I created that bundle for the MIT license.

PR #1189, opened 2023-10-14
[Site] Use new BaseEntityAutocompleteType

smnandre:
Update the "Autocomplete" package page with the terrific new autocomplete type made by @norkunas


weaverryan:
Good call - thanks Simon!

PR #1192, opened 2023-10-16
[TwigComponent] Allow Index view for Anonymous Components

### LINKED ISSUE #1060: [TwigComponent] Anonymous Index Components
Hi,

With [Anonymous Components](https://github.com/symfony/ux/pull/802), it would be interesting to have an [Anonymous Index Components](https://laravel.com/docs/10.x/blade#anonymous-index-components)

It allows to structure the components well when they have several files.

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #1060 
| License       | MIT

I created this PR using the example on https://github.com/laravel/framework/pull/38847.

Currently we can get the code below but the `Accordion` and `Item` template will not be in the same directory.
```twig
<twig:Accordion>
    <twig:Accordion:Item>...</twig:Accordion:Item>
    <twig:Accordion:Item>...</twig:Accordion:Item>
</twig:Accordion>
```
But we could do it by doing `<twig:Accordion:Accordion>` but that's weird.

With this functionality, it allows you to structure the components well. Here we will then have:
`components/Accordion/Index.html.twig`
`components/Accordion/Item.html.twig`

### DIFF
--- a/src/StimulusBundle/doc/index.rst
+++ b/src/StimulusBundle/doc/index.rst
@@ -576,4 +576,4 @@ it will normalize it:
 .. _`stimulus-components`: https://stimulus-components.netlify.app/
 .. _`TypeScript`: https://www.typescriptlang.org/
 .. _`sensiolabs/typescript-bundle`: https://github.com/sensiolabs/AssetMapperTypeScriptBundle
-.. _`Stimulus plugin`: https://plugins.jetbrains.com/plugin/18940-stimulus
+.. _`Stimulus plugin`: https://plugins.jetbrains.com/plugin/24562-stimulus


src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -50,6 +50,22 @@ public function findAnonymousComponentTemplate(string $name): ?string
                 return $componentPath;
             }
 
+            if ($loader->exists('components/'.$componentPath.'/Index.html.twig')) {
```
smnandre: Not needed here i think (legacy behaviour)

smnandre:
On one hand, i clearly see the DX benefit for front-end development.

On the other hand, this break some things / raise some questions:
* we cannot "just add a class Accordion.php" to add PHP computation.. because the associated template would be "Accordion.html.twig" and that was a big need for the anonymous 
* a component can be called by two names now (twig:Accordion and twig:Accordion:Index) 
* the debug / suggest tools will have more checks to do 

None of those are blocking points, but that deserve a bit of thinking i guess.

Update: that also would create two separate behaviour (depending of class-based or anonymous component)

seb-jean:
Thank you for your answer

seb-jean:
@smnandre Do you know how to move forward with this PR?

WebMamba:
I am sorry but I am not sure about this PR. Because of the reasons @smnandre mention and I don't like the idea of adding new conventions. We want people to feel free to write their code how they wanted without looking at conventions. This PR: https://github.com/symfony/ux/pull/1209 could be a solution for your issue. We just have some fix to do before it get merged 😁

PR #1193, opened 2023-10-16
[StimulusBundle] feat: normalize parameters names given to twig helper 'stimulus_action'

norival:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | NA
| License       | MIT

This allows giving a parameter name as 'myParam', and it will be printed as 'data-controller-name-my-param-param', which will be interpreted as 'myParam' by the Stimulus controller.

### DIFF
--- a/src/StimulusBundle/CHANGELOG.md
+++ b/src/StimulusBundle/CHANGELOG.md
@@ -1,5 +1,14 @@
 # CHANGELOG
 
+## 2.13.0
+
+-   Normalize parameters names given to twig helper 'stimulus_action()'.
+    **BC Break**: previously, parameters given in camelCase (eg.
+    `bigCrocodile`) were incorrectly registered by the controller as
+    flatcase (`event.params.bigcrocodile`). This was fixed, which means
+    they are now correctly registered as camelCase
+    (`event.params.bigCrocodile`).
+
 ## 2.10.0
 
 -   Handle Stimulus outlets
--- a/src/StimulusBundle/src/Dto/StimulusAttributes.php
+++ b/src/StimulusBundle/src/Dto/StimulusAttributes.php
@@ -79,7 +79,9 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
         ];
 
         foreach ($parameters as $name => $value) {
-            $this->attributes['data-'.$controllerName.'-'.$name.'-param'] = $this->getFormattedValue($value);
+            $key = $this->normalizeKeyName($name);
+
+            $this->attributes['data-'.$controllerName.'-'.$key.'-param'] = $this->getFormattedValue($value);
         }
     }
 
--- a/src/StimulusBundle/tests/Twig/StimulusTwigExtensionTest.php
+++ b/src/StimulusBundle/tests/Twig/StimulusTwigExtensionTest.php
@@ -199,6 +199,16 @@ final class StimulusTwigExtensionTest extends TestCase
             'expectedString' => 'data-action="click->symfony--ux-dropzone--dropzone#onClick"',
             'expectedArray' => ['data-action' => 'click->symfony--ux-dropzone--dropzone#onClick'],
         ];
+
+        yield 'normalize-name, with normalized parameters names' => [
+            'controllerName' => 'my-controller',
+            'actionName' => 'onClick',
+            'eventName' => null,
+            'parameters' => ['boolParam' => true, 'intParam' => 4, 'stringParam' => 'test'],
+            'expectedString' => 'data-action="onClick"',
+            'expectedString' => 'data-action="my-controller#onClick" data-my-controller-bool-param-param="true" data-my-controller-int-param-param="4" data-my-controller-string-param-param="test"',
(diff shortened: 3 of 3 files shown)


smnandre:
Nice catch @norival !

@weaverryan  : Is it seen as a bugfix.. or a new feature ? In the later, BC break alert :)

weaverryan:
> This allows giving a parameter name as 'myParam', and it will be printed as 'data-controller-name-my-param-param', which will be interpreted as 'myParam' by the Stimulus controller. 

What would the `myParam` have `data-controller...` attribute looked like before this PR? I'm also interested if there is any possibility that changing from the old string to the new string could potentially break someone's app... or if the old attribute was totally wrong and unusable, and now this fixes it.

smnandre:
Well after playing with this a bit, i'm 100% to consider this a bugfix. 

And it's more a "HTML/DOM/JS" problem than a specific Stimulus one.

```html
<div data-foo-bar="foo-bar" data-fooBar="fooBar">...</div>
<div data-fooBar="fooBar" data-foo-bar="foo-bar">...</div>
```

data-foobar


Data attributes are case-incensitive (and [parsed in lowercase](https://html.spec.whatwg.org/#attribute-name-state)), and the snake to camelCase conversion between HTML and JS is the [expected behaviour](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#javascript_access).


## Big Crocodile

Given a parameter named `bigCrocodile`

### Before 

Before the fix made by @norival , the HTML code rendered by the function was 

```html
<button 
    data-action="foo#bar" 
    data-foo-bigCrocodile-param="green" 
  >...</button>
  ```
  
  But in the Stimulus controller, the parameters passed were in lowercase
  
  ```js
  {bigcrocodile: "green"}
  ```

### After

After the fix, the HTML code will be

```html
<button 
    data-action="foo#bar" 
    data-foo-big-crocodile-param="green" 
  >...</button>
  ```

and the parameters passed to the action
  
  ```js
  {bigCrocodile: "green"}
  ```

That also means that is someone called the stimulusAction with snake-case parameters, nothing changes (they still get the parameter in camelCase)

**Comment by norival** [comment]
Great analysis @smnandre ! :smiley: 

Actually, I think there is a case where it would break things.

If you declare the parameters in the twig function like this:

```twig
{{ stimulus_action('controller', 'action', 'click', {bigCrocodile: "green"}) }}
```
this will render `...data-controller-bigcrocodile-param="green"` and register it as `event.params.bigcrocodile`. After the 'fix', it would render `...data-controller-big-crocodile-param="green"` and register it as `event.params.bigCrocodile`. So this would break the controller code using this parameter.

smnandre:
> this will render `...data-controller-bigcrocodile-param="green"`

I don't find where this transformation is done (bigCrocodile -> bigcrocodile in HTML) ?

norival:
> I don't find where this transformation is done (bigCrocodile -> bigcrocodile in HTML) ?

My bad, the HTML is indeed `data-controller-bigCrocodile-param="green"`, but in the DOM, it is interpreted as `data-controller-bigcrocodile-param="green"`, and, when getting the dataset from JS, it is `{controllerBigcrocodileParam: "green"}` (in Firefox and Chrome, at least).

smnandre:
I checked, and Stimulus handles it (on macOs / Firefox + Chrome) as {bigcrocodile: green} indeed. So there is a difference.


**But...** if a developer passed "bigCrocodile" in HTML and then manually used "bigcrocodile" in the JS code.... they should have made the PR you just wrote instead :))

norival:
> **But...** if a developer passed "bigCrocodile" in HTML and then manually used "bigcrocodile" in the JS code.... they should have made the PR you just wrote instead :))

Haha, I agree, that would be a bit weird :sweat_smile:

weaverryan:
Can you update the CHANGELOG entry to make it a little longer and explain the behavior change if you were relying on the broken behavior?

norival:
> Can you update the CHANGELOG entry to make it a little longer and explain the behavior change if you were relying on the broken behavior?

@weaverryan done! Is it ok like that?

src/StimulusBundle/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.13.0
+
+-   Normalize parameters names given to twig helper 'stimulus_action()'. **BC Break**: previously, parameters given in camelCase (eg. `bigCrocodile`) would be registered by the controller as flatcase (`event.params.bigcrocodile`). Now, they are registered as camelCase (`event.params.bigCrocodile`) as expected.
```
weaverryan: Break onto multiple lines when it gets long (around 72ish chars) and indent the other lines the same as the word `Normalize`. Also, slight rewording:

> camelCase (eg. `bigCrocodile`) were incorrectly registered ...

And change the last sentence to:

> This was fixed, which means they are now correctly registered as camelCase (`event.params.bigCrocodile`).

weaverryan (commented):
One last tweak :)

src/StimulusBundle/CHANGELOG.md
smnandre: And... long live the crocodile  😉

norival: Done :smiley:

weaverryan:
Thank you Xavier!

PR #1194, opened 2023-10-16
[stimulus-bundle] Add support for typescript/react stimulus controllers

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | kinda? (see below)
| New feature?  | no? 
| Tickets       | Fix #... 
| License       | MIT

Previously using the `webpack-encode-bundle` icw `stimulus`, 'custom' controllers were loaded via the `bootstrap.js` file which used a [regex](https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/assets/bootstrap.js#L7) to match the controllers to be loaded. 

In the new `stimulus-bundle`; loading the controllers was simplified, resulting in not loading `typescript` / `react` controllers anymore. This PR adds back the support for those files.

### DIFF
--- a/src/StimulusBundle/.gitignore
+++ b/src/StimulusBundle/.gitignore
@@ -1,5 +1,6 @@
 .php-cs-fixer.cache
 .phpunit.result.cache
 composer.lock
+var/
 vendor/
 tests/fixtures/var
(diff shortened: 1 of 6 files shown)


evertharmeling:
The files needs to be converted to `.js` files first right? So no need for this PR?

edit: [yep](https://symfony.com/doc/current/frontend/asset_mapper.html#can-i-use-with-typescript-jsx-or-vue)

weaverryan:
Yup - there is someone working on a TypeScript bundle that would help easing the ts -> js conversion in the background. When that's here, we can make sure this works - we may in fact need this PR at *that* point - we'll see :)

PR #1195, opened 2023-10-16
[CI] Restrict worflows by paths

smnandre:
Restrict workflows runs with "paths / paths-ignore" options


weaverryan:
Thanks Simon!

PR #1199, opened 2023-10-17
[TwigComponent] Fix twig:lint bug with anonymous component tag

### LINKED ISSUE #1089: [TwigComponent] Twig linter does not recognize anonymous components
When I run the following command

```bash
bin/console lint:twig templates/ --format=github
```

I got this message

```bash
Unknown component "common:atoms:badge:default". And no matching anonymous component template was found.
```

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #1089
| License       | MIT

Sincerely i'm not sure how to handle this properly "today".. 

There is a lot of things i'd like to refactor in a future version, and the fact than auto-closing tags and classic ones are handled by two entirely different processes is at the top of my todo :) 

But right now, i'm not sure we can do better than a quick patch, allowing the "lint:twig" to not crash for anonymous <twig:Foo></twig:Foo> tags.

Timeline of what "bugs":
- the lint:twig command find all templates, and for each one create an empty template loader, then parse & compile the template
- during the parsing, the ComponentParser calls [ComponentFactory::metadataFor](https://github.com/symfony/ux/blob/46d0b6885830b8cf546cd2c2479be2024f154ed3/src/TwigComponent/src/Twig/ComponentTokenParser.php#L50) to find the template matching the component name
- if no class-based component matches, the ComponentFactory calls the the [ComponentTemplateFinder](https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/ComponentTemplateFinder.php) 
- the ComponentTemplateFinder then asks the Loader if a matching template exists
- but as the Loader is temporary "empty".... 🐞 

So we must ether:
- A) stop resolving template during the parser/compiler work
- B) ensure the TemplateFinder always contains the real "full" loader
- C) change the way the lint:twig command works

A is my long-term objective, but clearly not "today". 
C is not so easy, as the corresponding code is hidden between 3 private methods 

So....  first suggestion (dirty but working)

### DIFF
--- a/src/TwigComponent/src/ComponentTemplateFinder.php
+++ b/src/TwigComponent/src/ComponentTemplateFinder.php
@@ -12,16 +12,24 @@
 namespace Symfony\UX\TwigComponent;
 
 use Twig\Environment;
+use Twig\Loader\LoaderInterface;
 
 /**
  * @author Matheo Daninos <matheo.daninos@gmail.com>
  */
 final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
 {
+    private readonly LoaderInterface $loader;
+
     public function __construct(
-        private Environment $environment,
+        Environment|LoaderInterface $loader,
         private readonly ?string $directory = null,
     ) {
+        if ($loader instanceof Environment) {
+            trigger_deprecation('symfony/ux-twig-component', '2.13', 'The "%s()" method will require "%s $loader" as first argument in 3.0. Passing an "Environment" instance is deprecated.', __METHOD__, LoaderInterface::class);
+            $loader = $loader->getLoader();
+        }
+        $this->loader = $loader;
         if (null === $this->directory) {
             trigger_deprecation('symfony/ux-twig-component', '2.13', 'The "%s()" method will require "string $directory" argument in 3.0. Not defining it or passing null is deprecated.', __METHOD__);
         }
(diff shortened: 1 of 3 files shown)


src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -12,12 +12,20 @@
...
  */
 final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
 {
+    /**
+     * Avoid bug when the lint:command changes the Environment Loader.
+     *
+     * @internal
+     */
+    private LoaderInterface $loader;
+
```
weaverryan, line 23: Can we, instead, replace the `$environment` constructor argument with `$loader`? We would just need a BC layer (remove the `Environment` type, then have an if statement that checks the type, and trigger deprecation if it is an `Environmnet`).

smnandre: Is Environment|LoaderInterface wrong ? (real question, i'm not really used to BC break handling :) )

weaverryan: I think that will work - I still forget we can do this now :)

weaverryan (approved):

weaverryan:
Thank you Simon - beautiful PR

PR #1200, opened 2023-10-18
[UX] Moving yarn above npm

ThomasLandauer:
As at https://symfony.com/doc/current/frontend/encore/installation.html

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Tickets       | 
| License       | MIT

### DIFF
--- a/src/TogglePassword/doc/index.rst
+++ b/src/TogglePassword/doc/index.rst
@@ -22,17 +22,19 @@ Then, install this bundle using Composer and Symfony Flex:
 
     $ composer require symfony/ux-toggle-password
 
-If you're using WebpackEncore, install your assets and restart Encore (not
+If you're using Webpack Encore, install your assets and restart Encore (not
 needed if you're using AssetMapper):
 
 .. code-block:: terminal
 
+    # yarn
+    $ yarn install --force
+    $ yarn watch
+
+    # npm
     $ npm install --force
     $ npm run watch
 
-    # or use yarn
-    $ yarn install --force
-    $ yarn watch
 
 Usage with Symfony Forms
 ------------------------


WebMamba:
Haha I think npm was above yarn just because it's more popular, but yes I think yarn is more popular in this community so let's put it above! 😁

weaverryan:
I've been putting `npm` first these days simply because it is something that is already installed for users with node. That lowers the barrier of entry. If we're going to make it consistent, I'd vote for putting npm first.

yarn is likely common in our community because that's what we started recommending when Encore came out. But back then, I'm not sure that npm even had its lock file yet. These days, I don't have much a preference between the two - hence wanting to recommend the one that comes pre-installed with node.

smnandre:
Would it be possible to use a toggler/segmented control like the "configuration bloc" ? Ideally that would remember the choice made by the user ?

That could ease a lot the reading 

config-bloc

Kocal:
Don't forget about pnpm and bun :trollface:

smnandre:
You cannot imagine how the front-end wars does not concern me (especially when you consider the basic install commands we would ask any user to run are exactly the same...)

But **right now,** it would be a blind lie to pretend Yarn is on the rise... and.. whatever "userbase" UX could have today, npm is (sadly or not is not my point) the _de facto_ tool everyone has.

So i'm 100% to write documentation with choices "yarn / npm / pnpm / bread / burger-bread" .... as long as the default displayed one is the default installed one.

And even you, right now, reading this message, i'm pretty sure you did not expect those usage statistics.
* [npm vs yarn](https://npmtrends.com/npm-vs-yarn) .... ("yes but you have to install npm to install yarn".... if we don't see how this is itself a reason..... let's look the next line :p )
* [yarn vs pnpm vs vite](https://npmtrends.com/pnpm-vs-vite-vs-yarn) 

Who can stand and say "yarn is on the rise" right here right now will have my eternal respect.

(please note i have no personal opinion on those tools...  Mootools was the only one <3  ) 

Right now, i strongly beleive we have to use the tool everyone has. And even if i'm not sure it's a good thing...... today this tool is npm.

Capture d’écran 2023-10-28 à 23 47 02

weaverryan:
Forgetting about the order. Is it even common to show `npm` and `yarn` these days? Are there any popular projects showing their install with both? At this moment, I'd prefer to ONLY show `npm`.

Kocal:
Closing in favor of #2263, where `yarn` commands have been removed.

Anyway, thanks for the contribution :)

PR #1202, opened 2023-10-18
Migrate from Jest to Vitest

ChqThomas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Tickets       | #1018 #610 
| License       | MIT

This PR is a draft to test the feasibility of migrating from [jest ](https://jestjs.io/) to [vitest](https://vitest.dev/)

Jest doesn't support ESM out of the box, which makes it difficult to test some libraries like Chart.js and Svelte v4. Vitest supports typescript and ESM by default and is compatible with Jest API, it'll help us test those packages and maybe others in the future

#### Infos:

There is a global config file `vitest.config.js` which contains the default settings for every package, when needed it can be extended with a file in the package root (ex: `src\React\assets\vitest.config.js`)

I replaced the `yarn test` script with `vitest src --run` so it's run globally and the CI doesn't stop at the first error from a package, it still can be run for every/individual packages with `yarn workspaces run vitest --run` / `yarn workspace "@symfony/ux-live-component" run vitest --run`

Libraries like fetch-mock-jest and jest-canvas-mock need to be replaced
The `require_context_poylfill.ts` file used by ux-react, ux-vue and ux-svelte needs to be modified as it uses a dynamic `require`, replacing it with an `import` doesn't work as it is asynchronous

#### Resources:
- https://vitest.dev/guide/migration.html
- https://srivastavaankita080.medium.com/jest-vitest-migration-79f4735dd5d0

### TODO:
-   Packages:
    -   [x] @symfony/ux-autocomplete `0/14`
        -   [x] replace fetch-mock-jest ?
    -   [x] @symfony/ux-chartjs `0/4`
        -   [x] fix failing tests
        -   [x] replace jest-canvas-mock ?
    -   [x] @symfony/ux-cropperjs `1/1`
    -   [x] @symfony/ux-dropzone `3/3`
    -   [x] @symfony/ux-lazy-image `1/1`
    -   [x] @symfony/ux-live-component `217/217` 💯 
        -   [x] `Error: Uncaught [ReferenceError: Node is not defined]` Just a warning but the test suite pass
    -   [x] @symfony/ux-notify `1/1`
    -   [x] @symfony/ux-react `0/4`
        -   [x] install @vitejs/plugin-react
        -   [x] migrate `require_context_poylfill.ts`
    -   [x] @symfony/stimulus-bundle `1/1`
        -   [x]  `Error: Uncaught [ReferenceError: Node is not defined]` Just a warning but the test suite pass
    -   [x] @symfony/ux-svelte `0/4`
        -   [x] replace svelte-jester with @sveltejs/vite-plugin-svelte
        -   [x] migrate `require_context_poylfill.ts`
        -   [x] fix failing tests
    -   [x] @symfony/ux-swup `0/5`
        -   [x] `require() of ES Module node_modules/delegate-it/index.js from node_modules/swup/lib/index.js not supported.`
    -   [x] @symfony/ux-toggle-password `1/1`
    -   [x] @symfony/ux-translator `91/91`
    -   [x] @symfony/ux-turbo `0/2`
        -   [x] `TypeError: Cannot read properties of undefined (reading 'prototype')`
    -   [x] @symfony/ux-typed `1/1`
    -   [x] @symfony/ux-vue `0/5`
        -   [x] replace @vue/vue3-jest with @vitejs/plugin-vue
        -   [x] migrate `require_context_poylfill.ts`
-  [x] Remove jest
-  [x] Cleanup
-  [ ] Benchmark ?

I'll need help to resolve these issues 😁


weaverryan:
Thanks for this 💯! I'm going to try to find time to dig a bit this weekend.

smnandre:
Does it support parralel runs ? something like

```
npm workspace run -p 5 eslint
```

(not working, just to illustrate)

ChqThomas:
@smnandre  Not in the current state as this project uses yarn 1.22 (classic) but with yarn v3+ you can:

```bash
yarn workspaces foreach --parallel run test
```

https://v3.yarnpkg.com/cli/workspaces/foreach

smnandre:
@ChqThomas ok thanks

So it's probably out of the scope of your current work..

I'd like to work the CI to speed things a bit (a lot?), because i must admin having to wait 2 minutes of JS scripts every time i touch a PHP file is not something i love :) 

Capture d’écran 2023-10-21 à 15 14 04


And, as far as i know, it would not be a good idea to condition JS runs to .js files change (as php and js files can impact each other in our packages)

I'm looking at [esbuild](https://esbuild.github.io/) right now, it could help us

Should i create an issue to discuss this ? 

WDYT ? 


(cc @weaverryan )

weaverryan:
> Should i create an issue to discuss this ?

Sure - I'd be interested to see what you have in mind 👍

weaverryan:
Ok, checked into this. Fantastic work - it got me really excited and I think this is going to work! Using your detailed notes, I added a few more commits and resolved many of the remaining issues. So, we're actually quite close now. I also added a little script that will run all the workspace tests and continue going even if some fail. Just running `yarn vitest` from the root directory doesn't work, as it ignores the `vitest.config.js` that's inside some of the packages (it was only using the root). But the new script seems to work and have the desired effect of not stopping if one workspace has a failing test.

A few svelte tests are failing... the markup apparently changed that Svelte is outputting? I'm not sure what's going on there - but there's a good chance we can just change the assertions to match the new markup, though I can't explain the change.

EDIT: I've fixed the rest. I think this might be working! @ChqThomas do you want to remove jest and see if anything else needs to be done? I'm not worried about benchmarking, btw - the tests are fast. If we need to do something faster or run them in parallel, that's a separate thing.

bin/run-vitest-all.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Get all workspace names
+workspaces=$(yarn workspaces --json info | jq -r 'if .type == "log" then .data | fromjson | keys[] else empty end')
```
aqmebrak, line 4: I think `jq` is missing from the dependencies

src/Svelte/assets/package.json
```diff
@@ -5,6 +5,7 @@
     "module": "dist/register_controller.js",
     "version": "1.0.0",
     "license": "MIT",
+    "type": "module",
```
weaverryan, line 8: @stof I'm wondering if I could get your advice. This was needed to get `@sveltejs/vite-plugin-svelte` to play nicely. In reality, we are already using `import` inside all packages and our users are using `import`. So, do you have any concern about this change? If not, we should apply it equally across all packages.

ChqThomas:
@weaverryan Wow that was fast!

> A few svelte tests are failing... the markup apparently changed that Svelte is outputting? I'm not sure what's going on there - but there's a good chance we can just change the assertions to match the new markup, though I can't explain the change.

I can't explain it either, Svelte add HTML comments to identify components in dev mode, seems like they are not stripped out of the output with the Vitest plugin like they were with the Jest one. Updating the assertions is fine 👍 

> EDIT: I've fixed the rest. I think this might be working! @ChqThomas do you want to remove jest and see if anything else needs to be done? I'm not worried about benchmarking, btw - the tests are fast. If we need to do something faster or run them in parallel, that's a separate thing.

Yup I'll take a final look and completely remove jest config

Great work 😄

bin/run-vitest-all.sh
smnandre: I had to install "jq" on my mac .. 

Could we do without ? Is there an alternative ?

ChqThomas: Not as reliable, but it would work:

```bash
workspaces=$(yarn workspaces info | grep -o '@symfony/[^"]*')
```

smnandre: If it works and does not take 5seconds more.. i think it's better not to require tools on a DX pov. WDYT ?

weaverryan: Yup! If it works, then it's reliable enough, as we control (via our packages) the output of `yarn workspaces info`. So it shouldn't suddenly stop working, and it's just a dev/testing tool. So let's change it

weaverryan:
Thank you very very much Thomas!

smnandre:
I'm not sure (never heard of vitest until last week so i'm no expert.. at all)... but it seems that we are not using the vitests workspaces ? Is it not worth it ? Or not a need for this repo ?

weaverryan:
I wasn't aware of them - so maybe something can be improved

Issue #801, opened 2023-04-17
[TwigComponent] Nicer syntax and imports


norkunas:
Now that the html like syntax is supported I'd like to propose to enhance a little bit further to introduce `import_component` or something similar so we can use aliased or direct component names without prefix, for example:
```twig
{% import_component 'App\Component\Alert' %}
{# or #}
{% import_component 'App\Component\Alert' as MyAlert %}
```
to use as:
```twig
<Alert />
or
<MyAlert  />
```

weaverryan:
My 2 cents: I'm not sold that this is needed, but I'm also not "unsold" either :p

norkunas:
For frontenders at least it would be a little similar when going into twig land from js :)

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?

norkunas:
For sure

carsonbot:
Thank you for this suggestion.
There has not been a lot of activity here for a while. Would you still like to see this feature?
[Every feature is developed by the community.](https://symfony.com/doc/current/contributing/community/releases.html#development)
Perhaps someone would like to try?
You can [read how to contribute](https://symfony.com/doc/current/contributing/code/pull_requests.html) to get started.

Kocal:
Hi, I'm not really fan of this kind of feature, in the JavaScript world, you can import and **alias** through the module resolution system, which does not exist here with Twig Components.

Twig Components are globally registered, if you want an alias, then I believe that you can configure an alias in `AsTwigComponent` attribute.

WebMamba:
Yes but what about the anonymous component @Kocal ?

Kocal:
Not a fan either, if I take your example for #1108 with `<twig:Form:Button:PrimaryButton>`, I rather think that it's a design problem on your part, and UX shouldn't try to fix it.

From my point of view, having a `PrimaryButton` component under `Form/Button` folders is totally overkill, and `PrimaryButton` can easily be remplaced by `Button variant="primary"` (by using [`html_cva`](https://twig.symfony.com/html_cva) that you introduced in the ecosystem 😉).

I believe you can write `<twig:Form:Button variant="primary"></twig:Form:Button>` or even simpler `<twig:Button variant="primary"></twig:Button>` instead.

smnandre:
It would only work outside component and not inside

```twig
{% import Foo:Bar as Bar %}

{# would work #}
<twig:Bar />

[...]

PR #1209, opened 2023-10-20
[TwigComponent] Add 'require ** as **' to alias components

### LINKED ISSUE #1108: Import Anonymous or TwigComponent with custom name
Just to get back on @seb-jean comment: https://github.com/symfony/ux/pull/802#issuecomment-1715180965.

I am thinking about something similar. When you start working with anonymous components you can end up with a lot of subfiles and things can get messy easily.

```Twig
 <twig:Form:Button:PrimaryButton/>
```

But instead, we can have something like:

```Twig
   {% require Form:Button:PrimaryButton as PrimaryButton %}

<twig:PrimaryButton/>
```

Tell me what you think about it.
Cheers 😁

maelanleborgne:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #1108
| License       | MIT

As suggested in #1108 and #801, this PR adds a new step in `TwigPreLexer` to replace components aliases by there full name.
This will be useful when reusing many times the same components in your template, especially when making extensive use of components : having a well organized component folder will lead to longer component names.

What do you think ?

TODO:
- [ ] Handle the case when require declaration is commented
- [ ] Handle the case when require declaration is in a verbatim block

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -167,7 +167,7 @@ class TwigPreLexer
             throw new SyntaxError(sprintf('Expected closing tag "</twig:%s>" not found.', $lastComponent), $this->line);
         }
 
-        return $output;
+        return $this->processRequires($output);
     }
 
     private function consumeComponentName(string $customExceptionMessage = null): string
@@ -494,5 +494,34 @@ class TwigPreLexer
         $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock'] = true;
 
         return '{% block content %}';
+    }
+
+    private function processRequires(string $input)
+    {
+        // matches "{% require Some:Component:Name as ComponentAlias %}"
+        $pattern = '/{%\s*require\s+([A-Za-z0-9_:@.\-]+)\s+as\s+([A-Za-z0-9_:@.\-]+)\s*%}\n?/';
+        $matches = [];
+        preg_match_all($pattern, $input, $matches, \PREG_SET_ORDER);
+        // replace all "{{ component('ComponentAlias' ... ) }}" with "{{ component('Some/Component/Name', ...) }}"
+        // and all "{% component 'ComponentAlias' ... %}" with "{% component 'Some/Component/Name' ... %}"
+        foreach ($matches as $match) {
+            $input = str_replace(
+                sprintf("{{ component('%s'", $match[2]),
+                sprintf("{{ component('%s'", $match[1]),
+                $input
+            );
+            $input = str_replace(
+                sprintf("{%% component '%s'", $match[2]),
+                sprintf("{%% component '%s'", $match[1]),
+                $input
+            );
+            $input = str_replace(
+                $match[0],
+                '',
+                $input
+            );
+        }
+
+        return $input;
     }
 }
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -5,6 +5,7 @@
 -   Deprecate calling `ComponentTemplateFinder` constructor without `directory` argument.
 -   Add profiler integration: `TwigComponentDataCollector` and debug toolbar templates
 -   Add search feature in `debug:twig-component` command.
+-   Add `{% require % as % %}` to reference components via alias.
 
 ## 2.12.0
 
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -32,9 +32,39 @@ final class TwigPreLexerTest extends TestCase
             '{{ component(\'foo\') }}',
         ];
 
+        yield 'simple_component_with_require' => [
+            '{% require foo as bar %}<twig:bar />',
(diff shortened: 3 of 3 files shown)


norkunas:
Sadly still have to prefix with `<twig:` 🥲

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -167,7 +167,7 @@ public function preLexComponents(string $input): string
             throw new SyntaxError(sprintf('Expected closing tag "</twig:%s>" not found.', $lastComponent), $this->line);
         }
 
-        return $output;
+        return $this->processRequires($output);
```
WebMamba, line 170: What do you think about dispatching an event `PostTwigPrelexing` and putting the processRequires function's content in a Subscriber?

WebMamba (commented):
Whoooo thanks @maelanleborgne to get into this 🧡

WebMamba:
@norkunas what do you propose then? I feel like people like this syntax

norkunas:
See my issue :)

WebMamba (commented):
@maelanleborgne I was thinking again about your PR. And I don't think to use regex is not the good path here. You have to replicate some logic like verbatim or comment blocks... And if for exemple someone made a typo like so `{% require Foo as Bar }` we not gonna deliver a proper errors. I think we should do the logic from the TwigPrelexer directly. When the TwigPrelexer find `{% require` then he look for componentName and his alias, store this in property. Then when the TwigPrelexer is opening a component he looks via his property if an alias doesn't exist for this componentName. What you think ?

norkunas:
For me it's too much of the prefix when using many components..

smnandre:
I love the idea... but i'd like to solve our problems with the two lexers first if you're OK to wait a couple of days :| 

We may stop resolving the component names at prelex time, so your implementation could be much easier ... or almost impossible depending on what we want to do :) 

Currently with your implementation, i fear the following use case would break, no ? 

```twig
{{ component(a_var) }}
```

```twig
{% verbatim %}
    {{ component('foo') }}
{% endverbatim %}
```

```twig
{% embed %}
    {% block 'inside' %}
        {{ component('foo') }}
    {% endblock %}
    {{ component('foo') }}
{% endembed %}
```


It seems to me that we would be safer with a NodeVisitor (like [trans_domain](https://github.com/symfony/twig-bridge/tree/6.3/NodeVisitor) for example)

smnandre:
Another complex case

```
{% require Foo:Bar as Bar %}
{% require Bar as Foo %}


<twig:Foo:Bar />
<twig:Foo />
<twig:Bar />
```

smnandre:
Another case 


```twig
{#  templates/_blocks.html.twig  #}

{% block sidebar_top %}
    <twig:Crocodile:BigCroco />
{% endblock %}

{% block sidebar_bottom %}
    <twig:Crocodile:LittleCroco />
{% endblock %}
```

```twig
{#  templates/homepage.html.twig  #}

{% require Fake:Crocodile as Crocodile %}
{% require Plastic:BigCroco as Crocodile:BigCroco %}

{% use '_blocks.html.twig' %}

<twig:Crocodile:LittleCroco />

{% bloc sidebar_top %}
    <twig:Crocodile:BigCroco />
    {{ parent() }}
{% endblock %}

{% bloc sidebar_bottom %}
    <twig:Crocodile:LittleCroco />
{% endblock %}
```


(side note: this one is another reminder we should not pre-resolve names before the render @weaverryan ... anything done before the render is a risk some Twig feature stop working)

jungleman12:
Nice feature! I hope it will be merged soon!
It will help me a lot in the structure of my components :).

tibobaldwin:
Ah yes, this kind of functionality is interesting for organizing components in files.
For example, this will save me from making a `<twig:Sidebar:Sidebar>`.

Kocal:
> Ah yes, this kind of functionality is interesting for organizing components in files. For example, this will save me from making a `<twig:Sidebar:Sidebar>`.

You can create a `templates/components/Sidebar.html.twig` file for that

tibobaldwin:
The idea is to structure the `Sidebar` component like this:

- templates/components/Sidebar/Sidebar.html.twig
- templates/components/Sidebar/SidebarHeader.html.twig
- templates/components/Sidebar/SidebarBody.html.twig
- templates/components/Sidebar/SidebarItem.html.twig
- templates/components/Sidebar/SidebarLabel.html.twig
- ...

Kocal:
Why not the following structure? 
- templates/components/Sidebar.html.twig
- templates/components/Sidebar/Header.html.twig
- templates/components/Sidebar/Body.html.twig
- templates/components/Sidebar/Item.html.twig
- templates/components/Sidebar/Label.html.twig

It would allow you to use components like this:
```twig
<twig:Sidebar>
    <twig:Sidebar:Header></twig:Sidebar:Header>
    <twig:Sidebar:Body></twig:Sidebar:Body>
    <twig:Sidebar:Item></twig:Sidebar:Item>
    <twig:Sidebar:Label></twig:Sidebar:Label>
</twig:Sidebar>
```

tibobaldwin:
I understand your logic and that works :). The idea I propose is to group the related components within a single directory.

Kocal:

PR #1218, opened 2023-10-21
Live component force post requests

### LINKED ISSUE #1208: [LiveComponent] Force POST method when model is changed
Hello,

I've built my first LiveComponent and it works well. But I've noticed that backend requests are sent as GET with all props being appended as URL parameters. As my component includes sensitive data, I want to change the request method to POST. According to the code this could be achieved by adding an action (see https://github.com/symfony/ux/blob/d3ce11f105579c95cab72d5967ff7875c9e3da6b/src/LiveComponent/assets/src/Backend/RequestBuilder.ts#L38). But when I tried to add an `data-action="live#update"` attribute to my form element, I got the following error:

Error: Since LiveComponents 2.3, you no longer need data-action="live#update" on form elements. Found on element: `<select data-model="mySelect" data-action="live#update">`

What's the correct way to force POST for backend requests?

```html
    <select data-model="mySelect">
        <!-- options...  should be sent as POST requests on change -->
    </select>
```

hepisec:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tickets       | Fix #1208
| License       | MIT

When your component model may contain sensitive data, you probably don't want your data being transferred via GET requests, because then the data might leak to browser history, server logs etc.

With this PR you can add `forcePost: true` to `#[AsLiveComponent]` and your component will always use POST requests to talk to the backend.

```php
#[AsLiveComponent(forcePost: true)]
class MySensitiveComponent
{
// ...
}
```

### DIFF
--- a/src/LiveComponent/assets/src/Backend/Backend.ts
+++ b/src/LiveComponent/assets/src/Backend/Backend.ts
@@ -25,8 +25,8 @@ export interface BackendAction {
 export default class implements BackendInterface {
     private readonly requestBuilder: RequestBuilder;
 
-    constructor(url: string, csrfToken: string | null = null) {
-        this.requestBuilder = new RequestBuilder(url, csrfToken);
+    constructor(url: string, method: 'get' | 'post' = 'post', csrfToken: string | null = null) {
+        this.requestBuilder = new RequestBuilder(url, method, csrfToken);
     }
 
     makeRequest(
--- a/src/LiveComponent/assets/src/Backend/RequestBuilder.ts
+++ b/src/LiveComponent/assets/src/Backend/RequestBuilder.ts
@@ -2,10 +2,12 @@ import { BackendAction, ChildrenFingerprints } from './Backend';
 
 export default class {
     private url: string;
+    private method: 'get' | 'post';
     private readonly csrfToken: string | null;
 
-    constructor(url: string, csrfToken: string | null = null) {
+    constructor(url: string, method: 'get' | 'post' = 'post', csrfToken: string | null = null) {
         this.url = url;
+        this.method = method;
         this.csrfToken = csrfToken;
     }
 
@@ -37,6 +39,7 @@ export default class {
         if (
             actions.length === 0 &&
             totalFiles === 0 &&
+            this.method === 'get' &&
             this.willDataFitInUrl(JSON.stringify(props), JSON.stringify(updated), params, JSON.stringify(children), JSON.stringify(updatedPropsFromParent))
         ) {
             params.set('props', JSON.stringify(props));
--- a/src/LiveComponent/assets/test/Backend/RequestBuilder.test.ts
+++ b/src/LiveComponent/assets/test/Backend/RequestBuilder.test.ts
@@ -2,7 +2,7 @@ import RequestBuilder from '../../src/Backend/RequestBuilder';
 
 describe('buildRequest', () => {
     it('sets basic data on GET request', () => {
-        const builder = new RequestBuilder('/_components?existing_param=1', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components?existing_param=1', 'get', '_the_csrf_token');
         const { url, fetchOptions } = builder.buildRequest(
             { firstName: 'Ryan' },
             [],
@@ -21,7 +21,7 @@ describe('buildRequest', () => {
     });
 
     it('sets basic data on POST request', () => {
-        const builder = new RequestBuilder('/_components', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components', 'post', '_the_csrf_token');
         const { url, fetchOptions } = builder.buildRequest(
             { firstName: 'Ryan' },
             [{
@@ -52,7 +52,7 @@ describe('buildRequest', () => {
     });
 
     it('sets basic data on POST request with batch actions', () => {
-        const builder = new RequestBuilder('/_components', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components', 'post', '_the_csrf_token');
         const { url, fetchOptions } = builder.buildRequest(
             { firstName: 'Ryan' },
             [{
@@ -87,7 +87,7 @@ describe('buildRequest', () => {
 
     // when data is too long it makes a post request
     it('makes a POST request when data is too long', () => {
-        const builder = new RequestBuilder('/_components', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components', 'get', '_the_csrf_token');
         const { url, fetchOptions } = builder.buildRequest(
             { firstName: 'Ryan'.repeat(1000) },
             [],
@@ -112,8 +112,38 @@ describe('buildRequest', () => {
         }));
     });
 
+    it('makes a POST request when method is post', () => {
+        const builder = new RequestBuilder('/_components', 'post', '_the_csrf_token');
+        const { url, fetchOptions } = builder.buildRequest(
+            {
+                firstName: 'Ryan'
+            },
+            [],
+            { firstName: 'Kevin' },
+            {},
+            {},
+            {}
+        );
+
+        expect(url).toEqual('/_components');
+        expect(fetchOptions.method).toEqual('POST');
+        expect(fetchOptions.headers).toEqual({
+            // no token
+            Accept: 'application/vnd.live-component+html',
+            'X-Requested-With': 'XMLHttpRequest',
+        });
+        const body = <FormData>fetchOptions.body;
+        expect(body).toBeInstanceOf(FormData);
+        expect(body.get('data')).toEqual(JSON.stringify({
+            props: {
+                firstName: 'Ryan'
+            },
+            updated: { firstName: 'Kevin' },
+        }));
+    });
+
     it('sends propsFromParent when specified', () => {
-        const builder = new RequestBuilder('/_components?existing_param=1', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components?existing_param=1', 'get', '_the_csrf_token');
         const { url } = builder.buildRequest(
             { firstName: 'Ryan' },
             [],
@@ -167,7 +197,7 @@ describe('buildRequest', () => {
     };
 
     it('Sends file with request', () => {
-        const builder = new RequestBuilder('/_components', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components', 'post', '_the_csrf_token');
 
         const { url, fetchOptions } = builder.buildRequest(
             { firstName: 'Ryan' },
@@ -192,7 +222,7 @@ describe('buildRequest', () => {
     });
 
     it('Sends multiple files with request', () => {
-        const builder = new RequestBuilder('/_components', '_the_csrf_token');
+        const builder = new RequestBuilder('/_components', 'post', '_the_csrf_token');
 
         const { url, fetchOptions } = builder.buildRequest(
             { firstName: 'Ryan' },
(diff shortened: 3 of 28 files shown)


weaverryan:
Thanks for getting this rolling - that's awesome! Might it make sense to have the setting not as an attribute but on the `#[AsLiveComponent]` attribute? If you have a component that requires POST for some reason, then it will ALWAYS need to be POSTed - it shouldn't need to have a `data-force-post` passed each time.

hepisec:
I tried that first as this was your first suggestion. But I was not able to pass the property to the frontend yet. Would `LiveComponentHydrator` be the right place to do so?

weaverryan:
> But I was not able to pass the property to the frontend yet. Would LiveComponentHydrator be the right place to do so?

Try https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php

If we added a new `forcePost: true` option to `AsLiveComponent`, we should be able to get it in there. Let me know if you need help - I can dig in a bit further. Ultimately, we'll add an attribute that would map to a new "value" in the Live stimulus controller. Then the JS part shouldn't be too much different :)

smnandre:
Should we make that a default ? 

I agree with @hepisec, and i'm even ready to say that any data sent to the server should be in POST...

hepisec:
I've added support for `#[AsLiveComponent(forcePost: true)]` to the PR. I leave the decision for the default value up to you. I'd like to see the security considerations covered in the documentation.

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -28,7 +28,8 @@ public function __construct(
         bool $exposePublicProps = true,
         string $attributesVar = 'attributes',
         public bool $csrf = true,
-        public string $route = 'ux_live_component'
+        public string $route = 'ux_live_component',
+        public bool $forcePost = false
```
smnandre: As it's almost a classic controller, why not use `method: ['GET', 'POST']` ?   as in #[AsController] ?

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
```diff
@@ -98,6 +98,13 @@ public function attributesForRendering(MountedComponent $mounted, ComponentMetad
             $mountedAttributes = $mountedAttributes->defaults(['data-live-id' => $id]);
         }
 
+        $forcePost = AsLiveComponent::forcePost($mounted->getComponent());
```
smnandre: Is this specific case needed ? Is it possible to use `AsLiveComponent::serviceConfig()` like for 'route' ?

src/LiveComponent/src/Attribute/AsLiveComponent.php
hepisec: The `RequestBuilder` in the JS part has a more advanced logic to switch between GET and POST methods. Therefore I think using `method` would be misleading.

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
hepisec: That's totally possible, I'll refactor this part to use `serviceConfig()`.

hepisec: I've added 'force_post' to `serviceConfig()`, but I also kept the `forcePost()` helper method. The `serviceConfig()` is not easily accessible in `attributesForRendering()`.

smnandre:
Is there something checking the request _**is**_ sent in POST in the backend side ? 

I think somehow the LiveComponentSubscriber should be updated no ?

hepisec:
I've added a check for `forcePost` in `LiveComponentSubscriber::onKernelController`.

weaverryan:
After talking with a few people, I think we should reverse your PR. We should make components use POST by default, with an option to opt into GET. The reason is the one given in your description. And, more generally, if you think of a live component re-render as a type of "form submit" (and, if the user is changing a prop value... then they likely *are* filling out a form in one way or another), then form submits should almost always be POST (Symfony forms, for example, default to `POST`).

Could you change the option from `forcePost` to `method` and default it to `post`? We could throw an exception in the attribute's constructor if some value other than `get` or post` were passed.

src/LiveComponent/assets/src/Backend/RequestBuilder.ts
```diff
@@ -33,10 +33,12 @@ export default class {
             0
         );
 
+        const forcePost = props['@attributes']?.['data-force-post'] == true;
```
weaverryan: For this, I think we should pass an argument to the constructor of `RequestBuilder`: `string method`. Then, instead of a `data-force-post`, add a new Stimulus value to the Stimulus controller called `requestMethod` that defaults to `post`. Then you can pass that into `Backend` here https://github.com/symfony/ux/blob/643382b746a3354d441b261588f972414fdd6b2b/src/LiveComponent/assets/src/live_controller.ts#L87 - and into `RequestBuilder` here https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/src/Backend/Backend.ts#L29

smnandre: Polling will still use "GET" right ?

weaverryan: It would not, no. It would be POST all the time. Is there a reason to make polling alone use GET?

hepisec:
I'll update the PR in the next days to have POST by default, as suggested by @weaverryan

src/LiveComponent/assets/src/Backend/Backend.ts
```diff
@@ -25,8 +25,8 @@ export interface BackendAction {
 export default class implements BackendInterface {
     private readonly requestBuilder: RequestBuilder;
 
-    constructor(url: string, csrfToken: string | null = null) {
-        this.requestBuilder = new RequestBuilder(url, csrfToken);
+    constructor(url: string, method: string, csrfToken: string | null = null) {
```
WebMamba: `method` can be typed more stricly: `method: 'get' | 'post' = 'post'`

src/LiveComponent/assets/test/Backend/RequestBuilder.test.ts
```diff
@@ -112,6 +112,36 @@ describe('buildRequest', () => {
         }));
     });
 
+    it('makes a POST request when method is post', () => {
```
WebMamba, line 115: We should also have a test to check for a GET request?

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -116,4 +124,13 @@ public static function liveListeners(object|string $component): array
...
         return $listeners;
     }
+
+    public static function getMethod(object|string $component): string
+    {
+        foreach ((new \ReflectionClass($component))->getAttributes(self::class) as $attr) {
+            return $attr->newInstance()->serviceConfig()['method'];
+        }
+
+        return 'post';
```
WebMamba: Can be shorten like so: ?
```suggestion
       $attributes = (new \ReflectionClass($component))->getAttributes(self::class);
       
       return $attributes[0]->newInstance()->serviceConfig()['method] ?? 'post'
```

WebMamba (commented):
Thanks a lot for your work @hepisec! 🧡

src/LiveComponent/assets/test/Backend/RequestBuilder.test.ts
hepisec: There are cases when requests are sent as POST even if method is set to GET. E.g. if an action is called or the URL length would be too long. Therefore setting method to GET means that GET **and** POST are allowed, while method = 'POST' means **only** POST is allowed.

src/LiveComponent/assets/src/Backend/Backend.ts
hepisec: Done

src/LiveComponent/src/Attribute/AsLiveComponent.php
hepisec: Done

src/LiveComponent/assets/src/Backend/RequestBuilder.ts
hepisec: I've updated the code according to suggestions by @weaverryan.

hepisec:
I've finished the work on this issue. As POST is now the new default method and GET is only allowed when `#[AsLiveComponent(method: 'get')]` is set, I had to update several tests / fixtures to work with the new default method.

Thanks everyone so far, I'm looking forward to your feedback.

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -116,4 +124,11 @@ public static function liveListeners(object|string $component): array
 
         return $listeners;
     }
+
+    public static function getMethod(object|string $component): string
+    {
+        $attributes = (new \ReflectionClass($component))->getAttributes(self::class);
+
+        return $attributes[0]->newInstance()->serviceConfig()['method'] ?? 'post';
+    }
```
weaverryan: I don't think this is needed. The other static methods are for things that need to read specific other attributes - like `LiveListener`. For an argument to `AsLiveComponent` (like `method`), when you need to read this, you can get it from the component metadata - example https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Twig/LiveComponentRuntime.php#L48

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
```diff
@@ -98,6 +98,13 @@ public function attributesForRendering(MountedComponent $mounted, ComponentMetad
             $mountedAttributes = $mountedAttributes->defaults(['data-live-id' => $id]);
         }
 
+        $requestMethod = AsLiveComponent::getMethod($mounted->getComponent());
+
+        if (!isset($mountedAttributes->all()['data-live-request-method-value'])) {
```
weaverryan: I think it's safe to set this in all situations - without the `if` statement.

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
```diff
@@ -98,6 +98,13 @@ public function attributesForRendering(MountedComponent $mounted, ComponentMetad
...
+            $attributesCollection->setRequestMethod($requestMethod);
+            $mountedAttributes = $mountedAttributes->defaults(['data-live-request-method-value' => $requestMethod]);
```
weaverryan: Is this line needed? The `$attributesCollection` should, I think, be handling setting this attribute.

weaverryan (commented):
Minor notes - this is looking good. Thanks for the patience on the slow review - but this is a very important PR!

weaverryan:
Friendly ping to you @hepisec :). Let us know if you have time to do the final steps! Thanks!

hepisec:
I'm starting right now.

src/LiveComponent/src/Attribute/AsLiveComponent.php
hepisec: I've replaced the calls to `AsLiveComponent::getMethod` and removed this method.

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
hepisec: Done

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
hepisec: Line removed

hepisec:
It should be done now. The failed tests seem to be related to TwigComponent and are 8.2 only.

I wish you all happy holidays.

weaverryan:
Hey @hepisec!

Sorry for the delay - this is ready to merge. Can you rebase this please?

Thanks!

hepisec:
Rebase done. Looking forward to my first finished contribution to symfony/ux :-)

weaverryan:
Thank you 1000 for this @hepisec!

PR #1219, opened 2023-10-21
[Autocomplete] Avoid destroying on disconnect

### LINKED ISSUE #1204: [Autocomplete] SortableJS Sorting Issue
## Issue

When an autocomplete control is used inside a sortable list using [SortableJS](https://github.com/SortableJS/Sortable), then the value of the autocomplete is lost when sorting elements.

## Description

This seems to be caused by the `disconnect` method, which...

 1. stops mutation observers
 2. destroys the TomSelect control

But, if the TomSelect control value has been modified then this value is now lost, and when the control is re-initialized after sorting the original value is set.

This issue is not present when using a plain (ie. non-ux-autocomplete) TomSelect control with SortableJS. eg. https://jsfiddle.net/mqwspcf3/

## Solution

I've been able to create a workaround by flushing the current TomSelect value to the underlying control on disconnect like so (not suggesting this as actual solution, just demonstrating problem)...

```javascript
    disconnect() {
        this.stopMutationObserver();

        // Flush current TomSelect value to underlying select control.
        const value = this.tomSelect.getValue();
        this.element.querySelectorAll('option').forEach((option) => {
            option.removeAttribute('selected');

            if (option.value === value) {
                option.setAttribute('selected', 'selected');
            }
        });

        this.tomSelect.destroy();
    }
```

I'm not sure if this is expected behaviour, but it does seem to cause an issue by default which is surprising. Any feedback appreciated, thanks.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #1204
| License       | MIT

The tests pass - so I can't see a problem with NOT destroying the element.

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -62,7 +62,28 @@ class default_1 extends Controller {
     }
     disconnect() {
         this.stopMutationObserver();
+        let currentSelectedValues = [];
+        if (this.selectElement) {
+            if (this.selectElement.multiple) {
+                currentSelectedValues = Array.from(this.selectElement.options)
+                    .filter((option) => option.selected)
+                    .map((option) => option.value);
+            }
+            else {
+                currentSelectedValues = [this.selectElement.value];
+            }
+        }
         this.tomSelect.destroy();
+        if (this.selectElement) {
+            if (this.selectElement.multiple) {
+                Array.from(this.selectElement.options).forEach((option) => {
+                    option.selected = currentSelectedValues.includes(option.value);
+                });
+            }
+            else {
+                this.selectElement.value = currentSelectedValues[0];
+            }
+        }
     }
     getMaxOptions() {
         return this.selectElement ? this.selectElement.options.length : 50;
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -86,7 +86,35 @@ export default class extends Controller {
 
     disconnect() {
         this.stopMutationObserver();
+
+        // TomSelect.destroy() resets the element to its original HTML. This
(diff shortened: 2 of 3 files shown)


smnandre:
> so I can't see a problem with NOT destroying the element.

Is that not best practice in term of memory, Turbo compatibility, etc ? 
* https://www.betterstimulus.com/integrating-libraries/lifecycle
* https://stimulus.hotwired.dev/reference/lifecycle-callbacks
* https://thoughtbot.com/blog/taking-the-most-out-of-stimulus 

If the controller creates the instance, it should handle the deletion too, no ?

Could Sortable mess with the Dom / create duplicates ?

smnandre:
> The tests pass

Is there any test related to the disconnection ? 👼

weaverryan:
Indeed, destroying is a best practice. And looking at tom-select now, it DOES remove some event listeners (for me, removal of event listeners is the REAL important part of `disconnect()`. 

Sigh, in that case, maybe @rodnaph can create a reproducer so I can dive deeper into the underlying cause of why the value is lost?

smnandre:
I'm wondering if the cloning could conflict with stimulus... https://sortablejs.github.io/Sortable/#cloning

smnandre:
So, after [many tests](https://github.com/symfony/ux/issues/1204#issuecomment-1779005663), there is indeed something we need to fix in our implementation... 

...but it seems your solution could be a good fix in the meanwhile (pro/cons 100% favor).

WDYT to merge it and then open another issue to study why our implementation has hidden side-effets ?

weaverryan:
This is ready to go. I kept the `destroy()`. The problem is that `TomSelect.destroy()` reverts the `innerHTML` of the `<select>`. This causes any value the user has selected to be lost. We restore it.

PR #1221, opened 2023-10-22
[CI][LiveComponent] Speed poll test

smnandre:
Divide every polling delay in the test by 2 

cc @ChqThomas @weaverryan 

| Before | After |
| - | - | 
| before | after |


weaverryan:
Ha! Nice one :)

PR #1222, opened 2023-10-22
Update LICENCE 

smnandre:
Update LICENCE files:
* root file
* all packages

Add LICENCE file:
* ux.symfony.com

### DIFF
--- a/ux.symfony.com/LICENSE
+++ b/ux.symfony.com/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2020-present Fabien Potencier
+Copyright (c) 2022-present Fabien Potencier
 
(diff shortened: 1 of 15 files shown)


ux.symfony.com/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2022-present Fabien Potencier
```
weaverryan, line 1: Note: adding this for clarity, but the entire repository has a root LICENSE file. And as ux.symfony.com is never a sub-tree split, this never lives on its own. But adding it just in case.

weaverryan:
Thanks Simon!

PR #1230, opened 2023-10-25
[LiveComponent] Allow binding LiveProp to URL query parameter

squrious:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | N/A
| License       | MIT

Inspired from the Livewire Url attribute, let's bind our LiveProp props to the URL query string!

This is a first proposal, with the minimum feature behavior:
- Add a new `url` constructor parameter to `LiveProp`
- Bound props appear in the URL in a parameter with the same name
- When the component is mounted for the first time (not in a live request), its props are initialized with provided query parameters
- When the component is rendered on client side, the URL is updated according to the mapping (on first render and further live updates)

Currently supported data types: 
- Scalar props
- Array props
- Custom DTO objects with scalar and array properties

```php
#[AsLiveComponent()]
final class MyComponent
{
    #[LiveProp(writable: true, url: true)]
    public ?string $search = null;

    #[LiveProp(writable: ['id', 'name'], url: true)]
    public ?SomeDto $dto = null;
    
    #[LiveProp(url: true)]
    public array $tags = [];
}
```

A bound LiveProp doesn't need to be writable, so it could be updated on server side based on a custom action. 
BUT for DTOs, for the moment I bind the writable props only. Maybe another config option could be a better solution?

What could be next?

- Improve data types support (nested DTOs, enums, specific properties from DTOs...)
- Aliasing a prop name to a query parameter name (`search` => `q`, `dto.id` => `some_id`) 
- Add a `keep` option to force presence in the URL even if the prop is empty
- Listen to external URL modification to synchronize props
- Multiple history strategies (currently the URL is replaced, but we could push new entries instead)
- Enable configuration from a `getQueryString` method in the component instead of attributes

You can also check my [demo app](https://github.com/squrious/symfony-ux-demo-live-url) to try the feature.

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -15,6 +15,7 @@
 -   Fix instantiating LiveComponentMetadata multiple times.
 -   Change JavaScript package to `type: module`.
 -   Throwing an error when setting an invalid model name.
+-   Add support for URL binding in `LiveProp`
 
 ## 2.12.0
 
--- a/src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
@@ -1,0 +1,31 @@
+import Component from '../index';
+import { PluginInterface } from './PluginInterface';
+import { UrlUtils, HistoryStrategy } from '../../url_utils';
+
+interface QueryMapping {
+    /**
+     * URL parameter name
+     */
+    name: string,
+}
+
+export default class implements PluginInterface {
+    constructor(private readonly mapping: {[p: string]: QueryMapping}) {}
+
+    attachToComponent(component: Component): void {
+        component.on('render:finished', (component: Component) => {
+            const urlUtils = new UrlUtils(window.location.href);
+            const currentUrl = urlUtils.toString();
+
+            Object.entries(this.mapping).forEach(([prop, mapping]) => {
+                const value = component.valueStore.get(prop);
+                urlUtils.set(mapping.name, value);
+            });
+
+            // Only update URL if it has changed
+            if (currentUrl !== urlUtils.toString()) {
+                HistoryStrategy.replace(urlUtils);
+            }
+        });
+    }
+}
--- a/src/LiveComponent/assets/src/url_utils.ts
+++ b/src/LiveComponent/assets/src/url_utils.ts
@@ -1,0 +1,169 @@
+/**
+ * Adapted from Livewire's history plugin.
+ *
+ * @see https://github.com/livewire/livewire/blob/d4839e3b2c23fc71e615e68bc29ff4de95751810/js/plugins/history/index.js
+ */
+
+/**
+ * Check if a value is empty.
+ *
+ * Empty values are:
+ * - `null` and `undefined`
+ * - Empty strings
+ * - Empty arrays
+ * - Deeply empty objects
+ */
+function isValueEmpty(value: any): boolean {
+    if (null === value || value === '' || undefined === value || (Array.isArray(value) && value.length === 0)) {
+        return true;
+    }
+
+    if (typeof value !== 'object') {
+        return false;
+    }
+
+    for (const key of Object.keys(value)) {
+        if (!isValueEmpty(value[key])) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+/**
+ * Converts JavaScript data to bracketed query string notation.
+ *
+ * Input: `{ items: [['foo']] }`
+ *
+ * Output: `"items[0][0]=foo"`
+ */
+function toQueryString(data: any) {
+    const buildQueryStringEntries = (data: { [p: string]: any }, entries: any = {}, baseKey = '') => {
+        Object.entries(data).forEach(([iKey, iValue]) => {
+            const key = baseKey === '' ? iKey : `${baseKey}[${iKey}]`;
+
+            if ('' === baseKey && isValueEmpty(iValue)) {
+                // Top level empty parameter
+                entries[key] = '';
+            } else if (null !== iValue) {
+                if (typeof iValue === 'object') {
+                    // Non-empty object/array process
+                    entries = { ...entries, ...buildQueryStringEntries(iValue, entries, key) };
+                } else {
+                    // Scalar value
+                    entries[key] = encodeURIComponent(iValue)
+                        .replace(/%20/g, '+') // Conform to RFC1738
+                        .replace(/%2C/g, ',');
+                }
+            }
+        });
+
+        return entries;
+    };
+
+    const entries = buildQueryStringEntries(data);
+
+    return Object.entries(entries)
+        .map(([key, value]) => `${key}=${value}`)
+        .join('&');
+}
+
+/**
+ * Converts bracketed query string notation to JavaScript data.
+ *
+ * Input: `"items[0][0]=foo"`
+ *
+ * Output: `{ items: [['foo']] }`
+ */
+function fromQueryString(search: string) {
+    search = search.replace('?', '');
+
+    if (search === '') return {};
+
+    const insertDotNotatedValueIntoData = (key: string, value: any, data: any) => {
+        const [first, second, ...rest] = key.split('.');
+
+        // We're at a leaf node, let's make the assigment...
+        if (!second) return (data[key] = value);
+
+        // This is where we fill in empty arrays/objects along the way to the assigment...
+        if (data[first] === undefined) {
+            data[first] = Number.isNaN(Number.parseInt(second)) ? {} : [];
+        }
+
+        // Keep deferring assignment until the full key is built up...
+        insertDotNotatedValueIntoData([second, ...rest].join('.'), value, data[first]);
+    };
+
+    const entries = search.split('&').map((i) => i.split('='));
+
+    const data: any = {};
+
+    entries.forEach(([key, value]) => {
+        value = decodeURIComponent(value.replace(/\+/g, '%20'));
+
+        if (!key.includes('[')) {
+            data[key] = value;
+        } else {
+            // Skip empty nested data
+            if ('' === value) return;
+
+            // Convert to dot notation because it's easier...
+            const dotNotatedKey = key.replace(/\[/g, '.').replace(/]/g, '');
+
+            insertDotNotatedValueIntoData(dotNotatedKey, value, data);
+        }
+    });
+
+    return data;
+}
+
+/**
+ * Wraps a URL to manage search parameters with common map functions.
+ */
+export class UrlUtils extends URL {
+    has(key: string) {
+        const data = this.getData();
+
+        return Object.keys(data).includes(key);
+    }
+
+    set(key: string, value: any) {
+        const data = this.getData();
+
+        data[key] = value;
+
+        this.setData(data);
+    }
+
+    get(key: string): any | undefined {
+        return this.getData()[key];
+    }
+
+    remove(key: string) {
+        const data = this.getData();
+
+        delete data[key];
+
+        this.setData(data);
+    }
+
+    private getData() {
+        if (!this.search) {
+            return {};
+        }
+
+        return fromQueryString(this.search);
+    }
+
+    private setData(data: any) {
+        this.search = toQueryString(data);
+    }
+}
+
+export class HistoryStrategy {
+    static replace(url: URL) {
+        history.replaceState(history.state, '', url);
+    }
+}
--- a/src/LiveComponent/src/Attribute/LiveProp.php
+++ b/src/LiveComponent/src/Attribute/LiveProp.php
@@ -59,6 +59,13 @@ final class LiveProp
     private null|string|array $onUpdated;
 
     /**
+     * @var bool
+     *
+     * Tells if this property should be bound to the URL
+     */
+    private bool $url;
+
+    /**
      * @param bool|array  $writable                  If true, this property can be changed by the frontend.
      *                                               Or set to an array of paths within this object/array
      *                                               that are writable.
@@ -73,6 +80,8 @@ final class LiveProp
      *                                               from the value used when originally rendering
      *                                               this child, the value in the child will be updated
      *                                               to match the new value and the child will be re-rendered
+     * @param bool        $url                       if true, this property will be synchronized with a query parameter
+     *                                               in the URL
      */
     public function __construct(
         bool|array $writable = false,
@@ -84,6 +93,7 @@ final class LiveProp
         string $format = null,
         bool $updateFromParent = false,
         string|array $onUpdated = null,
+        bool $url = false,
     ) {
         $this->writable = $writable;
         $this->hydrateWith = $hydrateWith;
@@ -94,6 +104,7 @@ final class LiveProp
         $this->format = $format;
         $this->acceptUpdatesFromParent = $updateFromParent;
         $this->onUpdated = $onUpdated;
+        $this->url = $url;
 
         if ($this->useSerializerForHydration && ($this->hydrateWith || $this->dehydrateWith)) {
             throw new \InvalidArgumentException('Cannot use useSerializerForHydration with hydrateWith or dehydrateWith.');
@@ -187,5 +198,10 @@ final class LiveProp
     public function onUpdated(): null|string|array
     {
         return $this->onUpdated;
+    }
+
+    public function url(): bool
+    {
+        return $this->url;
     }
 }
--- a/src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
+++ b/src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\LiveComponent\Util\QueryStringPropsExtractor;
+use Symfony\UX\TwigComponent\Event\PreMountEvent;
+
+/**
+ * @author Nicolas Rigaud <squrious@protonmail.com>
+ *
+ * @experimental
+ *
+ * @internal
+ */
+class QueryStringInitializeSubscriber implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly RequestStack $requestStack,
+        private readonly LiveComponentMetadataFactory $metadataFactory,
+        private readonly QueryStringPropsExtractor $queryStringPropsExtractor,
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            PreMountEvent::class => 'onPreMount',
+        ];
+    }
+
+    public function onPreMount(PreMountEvent $event): void
+    {
+        if (!$event->getMetadata()->get('live', false)) {
+            // Not a live component
+            return;
+        }
+
+        $request = $this->requestStack->getMainRequest();
+
+        if (null === $request) {
+            return;
+        }
+
+        $metadata = $this->metadataFactory->getMetadata($event->getMetadata()->getName());
+
+        if (!$metadata->hasQueryStringBindings()) {
+            return;
+        }
+
+        $queryStringData = $this->queryStringPropsExtractor->extract($request, $metadata, $event->getComponent());
+
+        $event->setData(array_merge($event->getData(), $queryStringData));
+    }
+}
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -19,6 +19,7 @@ use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractor;
 use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\Serializer\Exception\ExceptionInterface as SerializerExceptionInterface;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
@@ -226,6 +227,71 @@ final class LiveComponentHydrator
         return $attributes;
     }
 
+    /**
+     * Hydrate a value from a dehydrated value.
+     *
+     * Depending on the prop configuration, the value may be hydrated by a custom method or the Serializer component.
+     *
+     * @internal
+     *
+     * @throws SerializerExceptionInterface
+     */
+    public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, object $parentObject): mixed
+    {
+        if ($propMetadata->hydrateMethod()) {
+            if (!method_exists($parentObject, $propMetadata->hydrateMethod())) {
+                throw new \LogicException(sprintf('The "%s" object has a hydrateMethod of "%s" but the method does not exist.', $parentObject::class, $propMetadata->hydrateMethod()));
+            }
+
+            return $parentObject->{$propMetadata->hydrateMethod()}($value);
+        }
+
+        if ($propMetadata->useSerializerForHydration()) {
+            if (!interface_exists(DenormalizerInterface::class)) {
+                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the Serializer component is not installed. Try running "composer require symfony/serializer".', $propMetadata->getName(), $parentObject::class));
+            }
+            if (null === $this->serializer) {
+                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but no serializer has been set.', $propMetadata->getName(), $parentObject::class));
+            }
+            if (!$this->serializer instanceof DenormalizerInterface) {
+                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the given serializer does not implement DenormalizerInterface.', $propMetadata->getName(), $parentObject::class));
+            }
+
+            if (null === $propMetadata->getType()) {
+                throw new \LogicException(sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
+            }
+
+            return $this->serializer->denormalize($value, $propMetadata->getType(), 'json', $propMetadata->serializationContext());
+        }
+
+        if ($propMetadata->collectionValueType() && Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()) {
+            $collectionClass = $propMetadata->collectionValueType()->getClassName();
+            foreach ($value as $key => $objectItem) {
+                $value[$key] = $this->hydrateObjectValue($objectItem, $collectionClass, true, $propMetadata->getFormat(), $parentObject::class, sprintf('%s.%s', $propMetadata->getName(), $key), $parentObject);
+            }
+        }
+
+        // no type? no hydration
+        if (!$propMetadata->getType()) {
+            return $value;
+        }
+
+        if (null === $value) {
+            return null;
+        }
+
+        if (\is_string($value) && $propMetadata->isBuiltIn() && \in_array($propMetadata->getType(), ['int', 'float', 'bool'], true)) {
+            return self::coerceStringValue($value, $propMetadata->getType(), $propMetadata->allowsNull());
+        }
+
+        // for all other built-ins: int, boolean, array, return as is
+        if ($propMetadata->isBuiltIn()) {
+            return $value;
+        }
+
+        return $this->hydrateObjectValue($value, $propMetadata->getType(), $propMetadata->allowsNull(), $propMetadata->getFormat(), $parentObject::class, $propMetadata->getName(), $parentObject);
+    }
+
     public function addChecksumToData(array $data): array
     {
         $data[self::CHECKSUM_KEY] = $this->calculateChecksum($data);
@@ -428,58 +494,6 @@ final class LiveComponentHydrator
         }
 
         return $dehydratedObjectValues;
-    }
-
-    private function hydrateValue(mixed $value, LivePropMetadata $propMetadata, object $parentObject): mixed
-    {
-        if ($propMetadata->hydrateMethod()) {
-            if (!method_exists($parentObject, $propMetadata->hydrateMethod())) {
-                throw new \LogicException(sprintf('The "%s" object has a hydrateMethod of "%s" but the method does not exist.', $parentObject::class, $propMetadata->hydrateMethod()));
-            }
-
-            return $parentObject->{$propMetadata->hydrateMethod()}($value);
-        }
-
-        if ($propMetadata->useSerializerForHydration()) {
-            if (!interface_exists(DenormalizerInterface::class)) {
-                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the Serializer component is not installed. Try running "composer require symfony/serializer".', $propMetadata->getName(), $parentObject::class));
-            }
-            if (null === $this->serializer) {
-                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but no serializer has been set.', $propMetadata->getName(), $parentObject::class));
-            }
-            if (!$this->serializer instanceof DenormalizerInterface) {
-                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the given serializer does not implement DenormalizerInterface.', $propMetadata->getName(), $parentObject::class));
-            }
-
-            return $this->serializer->denormalize($value, $propMetadata->getType(), 'json', $propMetadata->serializationContext());
-        }
-
-        if ($propMetadata->collectionValueType() && Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()) {
-            $collectionClass = $propMetadata->collectionValueType()->getClassName();
-            foreach ($value as $key => $objectItem) {
-                $value[$key] = $this->hydrateObjectValue($objectItem, $collectionClass, true, $propMetadata->getFormat(), $parentObject::class, sprintf('%s.%s', $propMetadata->getName(), $key), $parentObject);
-            }
-        }
-
-        // no type? no hydration
-        if (!$propMetadata->getType()) {
-            return $value;
-        }
-
-        if (null === $value) {
-            return null;
-        }
-
-        if (\is_string($value) && $propMetadata->isBuiltIn() && \in_array($propMetadata->getType(), ['int', 'float', 'bool'], true)) {
-            return self::coerceStringValue($value, $propMetadata->getType(), $propMetadata->allowsNull());
-        }
-
-        // for all other built-ins: int, boolean, array, return as is
-        if ($propMetadata->isBuiltIn()) {
-            return $value;
-        }
-
-        return $this->hydrateObjectValue($value, $propMetadata->getType(), $propMetadata->allowsNull(), $propMetadata->getFormat(), $parentObject::class, $propMetadata->getName(), $parentObject);
     }
 
     private function hydrateObjectValue(mixed $value, string $className, bool $allowsNull, ?string $dateFormat, string $componentClassForError, string $propertyPathForError, object $component): ?object
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -95,24 +95,22 @@ class LiveComponentMetadataFactory implements ResetInterface
 
         if (null === $type && null === $collectionValueType && isset($infoTypes[0])) {
             $infoType = Type::BUILTIN_TYPE_OBJECT === $infoTypes[0]->getBuiltinType() ? $infoTypes[0]->getClassName() : $infoTypes[0]->getBuiltinType();
-
-            return new LivePropMetadata(
-                $property->getName(),
-                $liveProp,
-                $infoType,
-                null === $infoTypes[0]->getClassName(),
-                $infoTypes[0]->isNullable(),
-                null,
-            );
+            $isTypeBuiltIn = null === $infoTypes[0]->getClassName();
+            $isTypeNullable = $infoTypes[0]->isNullable();
+        } else {
+            $infoType = $type?->getName();
+            $isTypeBuiltIn = $type?->isBuiltin() ?? false;
+            $isTypeNullable = $type?->allowsNull() ?? true;
         }
 
         return new LivePropMetadata(
             $property->getName(),
             $liveProp,
-            $type?->getName(),
-            $type && $type->isBuiltin(),
-            !$type || $type->allowsNull(),
+            $infoType,
+            $isTypeBuiltIn,
+            $isTypeNullable,
             $collectionValueType,
+            $liveProp->url()
         );
     }
 
--- a/src/LiveComponent/src/Metadata/LivePropMetadata.php
+++ b/src/LiveComponent/src/Metadata/LivePropMetadata.php
@@ -30,6 +30,7 @@ final class LivePropMetadata
         private bool $isBuiltIn,
         private bool $allowsNull,
         private ?Type $collectionValueType,
+        private bool $queryStringMapping,
     ) {
     }
 
@@ -51,6 +52,11 @@ final class LivePropMetadata
     public function allowsNull(): bool
     {
         return $this->allowsNull;
+    }
+
+    public function queryStringMapping(): bool
+    {
+        return $this->queryStringMapping;
     }
 
     public function calculateFieldName(object $component, string $fallback): string
--- a/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
+++ b/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
@@ -98,10 +98,23 @@ class LiveControllerAttributesCreator
             $mountedAttributes = $mountedAttributes->defaults(['data-live-id' => $id]);
         }
 
+        $liveMetadata = $this->metadataFactory->getMetadata($mounted->getName());
+
+        if ($liveMetadata->hasQueryStringBindings()) {
+            $queryMapping = [];
+            foreach ($liveMetadata->getAllLivePropsMetadata() as $livePropMetadata) {
+                if ($livePropMetadata->queryStringMapping()) {
+                    $frontendName = $livePropMetadata->calculateFieldName($mounted, $livePropMetadata->getName());
+                    $queryMapping[$frontendName] = ['name' => $frontendName];
+                }
+            }
+            $attributesCollection->setQueryUrlMapping($queryMapping);
+        }
+
         if ($isChildComponent) {
             $fingerprint = $this->fingerprintCalculator->calculateFingerprint(
                 $mounted->getInputProps(),
-                $this->metadataFactory->getMetadata($mounted->getName())
+                $liveMetadata
             );
             if ($fingerprint) {
                 $attributesCollection->setFingerprint($fingerprint);
--- a/src/LiveComponent/src/Util/QueryStringPropsExtractor.php
+++ b/src/LiveComponent/src/Util/QueryStringPropsExtractor.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Util;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\UX\LiveComponent\Exception\HydrationException;
+use Symfony\UX\LiveComponent\LiveComponentHydrator;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadata;
+use Symfony\UX\LiveComponent\Metadata\LivePropMetadata;
+
+/**
+ * @author Nicolas Rigaud <squrious@protonmail.com>
+ *
+ * @experimental
+ *
+ * @internal
+ */
+final class QueryStringPropsExtractor
+{
+    public function __construct(private readonly LiveComponentHydrator $hydrator)
+    {
+    }
+
+    /**
+     * Extracts relevant query parameters from the current URL and hydrates them.
+     */
+    public function extract(Request $request, LiveComponentMetadata $metadata, object $component): array
+    {
+        $query = $request->query->all();
+
+        if (empty($query)) {
+            return [];
+        }
+        $data = [];
+
+        foreach ($metadata->getAllLivePropsMetadata() as $livePropMetadata) {
+            if ($livePropMetadata->queryStringMapping()) {
+                $frontendName = $livePropMetadata->calculateFieldName($component, $livePropMetadata->getName());
+                if (null !== ($value = $query[$frontendName] ?? null)) {
+                    if ('' === $value && null !== $livePropMetadata->getType() && (!$livePropMetadata->isBuiltIn() || 'array' === $livePropMetadata->getType())) {
+                        // Cast empty string to empty array for objects and arrays
+                        $value = [];
+                    }
+
+                    try {
+                        $hydratedValue = $this->hydrator->hydrateValue($value, $livePropMetadata, $component);
+
+                        if ($this->isValueTypeConsistent($hydratedValue, $livePropMetadata)) {
+                            // Only set data if hydrated value type is consistent with prop metadata type
+                            $data[$livePropMetadata->getName()] = $hydratedValue;
+                        }
+                    } catch (HydrationException) {
+                        // Skip hydration errors (e.g. with objects)
+                    }
+                }
+            }
+        }
+
+        return $data;
+    }
+
+    private function isValueTypeConsistent(mixed $value, LivePropMetadata $livePropMetadata): bool
+    {
+        $propType = $livePropMetadata->getType();
+
+        if ($livePropMetadata->allowsNull() && null === $value) {
+            return true;
+        }
+
+        return
+            \in_array($propType, [null, 'mixed'])
+            || $livePropMetadata->isBuiltIn() && ('\is_'.$propType)($value)
+            || !$livePropMetadata->isBuiltIn() && $value instanceof $propType;
+    }
+}
--- a/src/LiveComponent/tests/Fixtures/Component/ComponentWithUrlBoundProps.php
+++ b/src/LiveComponent/tests/Fixtures/Component/ComponentWithUrlBoundProps.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+use Symfony\UX\LiveComponent\Tests\Fixtures\Dto\Address;
+
+#[AsLiveComponent('component_with_url_bound_props')]
+class ComponentWithUrlBoundProps
+{
+    use DefaultActionTrait;
+
+    #[LiveProp(url: true)]
+    public ?string $prop1 = null;
+
+    #[LiveProp(url: true)]
+    public ?int $prop2 = null;
+
+    #[LiveProp(url: true)]
+    public array $prop3 = [];
+
+    #[LiveProp]
+    public ?string $prop4 = null;
+
+    #[LiveProp(url: true)]
+    public ?Address $prop5 = null;
+
+    #[LiveProp(fieldName: 'field6', url: true)]
+    public ?string $prop6 = null;
+}
--- a/src/LiveComponent/tests/Functional/Util/QueryStringPropsExtractorTest.php
+++ b/src/LiveComponent/tests/Functional/Util/QueryStringPropsExtractorTest.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Tests\Functional\Util;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\LiveComponent\Tests\Fixtures\Dto\Address;
+use Symfony\UX\LiveComponent\Tests\LiveComponentTestHelper;
+use Symfony\UX\LiveComponent\Util\QueryStringPropsExtractor;
+
+class QueryStringPropsExtractorTest extends KernelTestCase
+{
+    use LiveComponentTestHelper;
+
+    /**
+     * @dataProvider getQueryStringTests
+     */
+    public function testExtract(string $queryString, array $expected)
+    {
+        $extractor = new QueryStringPropsExtractor($this->hydrator());
+
+        $request = Request::create('/'.!empty($queryString) ? '?'.$queryString : '');
+
+        /** @var LiveComponentMetadataFactory $metadataFactory */
+        $metadataFactory = self::getContainer()->get('ux.live_component.metadata_factory');
+
+        $metadata = $metadataFactory->getMetadata('component_with_url_bound_props');
+        $component = $this->getComponent('component_with_url_bound_props');
+
+        $data = $extractor->extract($request, $metadata, $component);
+
+        $this->assertEquals($expected, $data);
+    }
+
+    public function getQueryStringTests(): iterable
+    {
+        yield from [
+            'no query string' => ['', []],
+            'empty value for nullable string' => ['prop1=', ['prop1' => null]],
+            'string value' => ['prop1=foo', ['prop1' => 'foo']],
+            'empty value for nullable int' => ['prop2=', ['prop2' => null]],
+            'int value' => ['prop2=42', ['prop2' => 42]],
+            'array value' => ['prop3[]=foo&prop3[]=bar', ['prop3' => ['foo', 'bar']]],
+            'array value indexed' => ['prop3[1]=foo&prop3[0]=bar', ['prop3' => [1 => 'foo', 0 => 'bar']]],
+            'not bound prop' => ['prop4=foo', []],
+            'object value' => ['prop5[address]=foo&prop5[city]=bar', ['prop5' => (function () {
+                $address = new Address();
+                $address->address = 'foo';
+                $address->city = 'bar';
+
+                return $address;
+            })(),
+            ]],
+            'invalid scalar value' => ['prop1[]=foo&prop1[]=bar', []],
+            'invalid array value' => ['prop3=foo', []],
+            'invalid object value' => ['prop5=foo', []],
+        ];
+    }
+}
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -75,7 +75,7 @@ final class ComponentFactory
     public function mountFromObject(object $component, array $data, ComponentMetadata $componentMetadata): MountedComponent
     {
         $originalData = $data;
-        $data = $this->preMount($component, $data);
+        $data = $this->preMount($component, $data, $componentMetadata);
 
         $this->mount($component, $data);
 
@@ -88,7 +88,7 @@ final class ComponentFactory
             }
         }
 
-        $postMount = $this->postMount($component, $data);
+        $postMount = $this->postMount($component, $data, $componentMetadata);
         $data = $postMount['data'];
         $extraMetadata = $postMount['extraMetadata'];
 
@@ -174,9 +174,9 @@ final class ComponentFactory
         return $this->components->get($name);
     }
 
-    private function preMount(object $component, array $data): array
+    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): array
     {
-        $event = new PreMountEvent($component, $data);
+        $event = new PreMountEvent($component, $data, $componentMetadata);
         $this->eventDispatcher->dispatch($event);
         $data = $event->getData();
 
@@ -194,9 +194,9 @@ final class ComponentFactory
     /**
      * @return array{data: array<string, mixed>, extraMetadata: array<string, mixed>}
      */
-    private function postMount(object $component, array $data): array
+    private function postMount(object $component, array $data, ComponentMetadata $componentMetadata): array
     {
-        $event = new PostMountEvent($component, $data);
+        $event = new PostMountEvent($component, $data, $componentMetadata);
         $this->eventDispatcher->dispatch($event);
         $data = $event->getData();
         $extraMetadata = $event->getExtraMetadata();
--- a/src/TwigComponent/src/Event/PostMountEvent.php
+++ b/src/TwigComponent/src/Event/PostMountEvent.php
@@ -12,17 +12,38 @@
 namespace Symfony\UX\TwigComponent\Event;
 
 use Symfony\Contracts\EventDispatcher\Event;
+use Symfony\UX\TwigComponent\ComponentMetadata;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
  */
 final class PostMountEvent extends Event
 {
+    private ?ComponentMetadata $metadata;
+    private array $extraMetadata;
+
     public function __construct(
         private object $component,
         private array $data,
-        private array $extraMetadata = [],
+        array|ComponentMetadata $metadata = [],
+        $extraMetadata = []
     ) {
+        if (\is_array($metadata)) {
+            trigger_deprecation('symfony/ux-twig-component', '2.13', 'In TwigComponent 3.0, the third argument of "%s()" will be a "%s" object and the "$extraMetadata" array should be passed as the fourth argument.', __METHOD__, ComponentMetadata::class);
+
+            $this->metadata = null;
+            $this->extraMetadata = $metadata;
+        } else {
+            if (null !== $metadata && !$metadata instanceof ComponentMetadata) {
+                throw new \InvalidArgumentException(sprintf('Expecting "$metadata" to be null or an instance of "%s", given: "%s."', ComponentMetadata::class, get_debug_type($metadata)));
+            }
+            if (!\is_array($extraMetadata)) {
+                throw new \InvalidArgumentException(sprintf('Expecting "$extraMetadata" to be array, given: "%s".', get_debug_type($extraMetadata)));
+            }
+
+            $this->metadata = $metadata;
+            $this->extraMetadata = $extraMetadata;
+        }
     }
 
     public function getComponent(): object
@@ -38,6 +59,11 @@ final class PostMountEvent extends Event
     public function setData(array $data): void
     {
         $this->data = $data;
+    }
+
+    public function getMetadata(): ?ComponentMetadata
+    {
+        return $this->metadata;
     }
 
     public function getExtraMetadata(): array
--- a/src/TwigComponent/src/Event/PreMountEvent.php
+++ b/src/TwigComponent/src/Event/PreMountEvent.php
@@ -12,14 +12,18 @@
 namespace Symfony\UX\TwigComponent\Event;
 
 use Symfony\Contracts\EventDispatcher\Event;
+use Symfony\UX\TwigComponent\ComponentMetadata;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
  */
 final class PreMountEvent extends Event
 {
-    public function __construct(private object $component, private array $data)
+    public function __construct(private object $component, private array $data, private readonly ?ComponentMetadata $metadata = null)
     {
+        if (null === $this->metadata) {
+            trigger_deprecation('symfony/ux-twig-component', '2.13', 'In TwigComponent 3.0, "%s()" method will require a "%s $metadata" argument. Not passing it is deprecated.', __METHOD__, ComponentMetadata::class);
+        }
     }
 
     public function getComponent(): object
@@ -35,5 +39,10 @@ final class PreMountEvent extends Event
     public function setData(array $data): void
     {
         $this->data = $data;
+    }
+
+    public function getMetadata(): ?ComponentMetadata
+    {
+        return $this->metadata;
     }
 }
--- a/src/TwigComponent/tests/Unit/EventListener/TwigComponentLoggerListenerTest.php
+++ b/src/TwigComponent/tests/Unit/EventListener/TwigComponentLoggerListenerTest.php
@@ -38,9 +38,9 @@ class TwigComponentLoggerListenerTest extends TestCase
         $eventB = new PreCreateForRenderEvent('b');
         $logger->onPreCreateForRender($eventB);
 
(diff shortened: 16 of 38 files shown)


src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+    public function onPreMount(PreMountEvent $event): void
+    {
+        $component = $event->getComponent();
+        if (!($metadata = $this->registered[$component::class] ?? null)) {
+            return;
+        }
+
+        $data = $event->getData();
+
+        $request = $this->requestStack->getCurrentRequest();
```
weaverryan: Should this be `getMainRequest()`? It would only make a difference if this component were being rendered inside of a sub-request. And in that case, we want to read the query param from the main, real request.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $queryStringData = $this->queryStringPropsExtractor->extract($request->getQueryString(), $metadata);
```
weaverryan: Does it make sense to pass `$request->query->all()` instead of the string (so we don't need to re-parse the string inside the extractor)?

kbond:
Nice work here! I'm curious how multiple components using this feature would behave. What about multiple of the same component?

weaverryan (commented):
SO excited for this! I'll do a more thorough review tomorrow

squrious:
> Nice work here! I'm curious how multiple components using this feature would behave. What about multiple of the same component?
Thanks!

Hum that's actually a very good question :smile: 

I think this is related to the _alias_ feature in some way; and we should be able to configure/override some part of the mapping in the component attributes. Like, maybe
```html
{% component('MyComponent', { query-mapping: {"prop": "alias1}}) %}
{% component('MyComponent', { query-mapping: {"prop": "alias2}}) %}
```

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: Yes, and I'm about to pass the `Request` object directly, so the strategy to extract the data is fully managed in this class. I think it's better, wdyt?

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: Yes it should, thanks! I change it

squrious:
Actually I'm wondering if the DTO support really worth it. For the moment it's quite limited:
- As we should not use dots in query parameters, we use `_`. This is kind of naming strategy that should be abstracted.
- Nested DTO properties are not allowed. This is mainly because I didn't yet implement any config to _select_ some properties of a DTO prop, so it avoids problems such as self recursion etc.

The DTO support, if shipped with this feature, seems close to the Serializer concepts (max depth, projection, sub context, naming strategy, etc.). On the other hand, handling scalar values is really easy. So handling a custom DTO in a custom way, so the developer can eventually manage the scalar properties he/she wants, should be easy too. 

That's all to say, I suggest to drop the incomplete support of DTOs, and instead provide an extension point (an interface or some tag injection) so :
- The default extractor handles scalars and arrays of scalars
- Additional strategies could be added in custom services by the developer

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
weaverryan: That works for me 👍

weaverryan:
> Actually I'm wondering if the DTO support really worth it. For the moment it's quite limited

My immediate reaction when I saw you had DTO support was "wow" but also "that sounds like it may get super complex" :).

> That's all to say, I suggest to drop the incomplete support of DTOs, and instead provide an extension point (an interface or some tag injection) so :

Agreed! Though I am also not *that* concerned immediately about adding the extension point. If you have something obvious in mind, sure, let's add it. But it is not a need.

src/LiveComponent/src/Metadata/LivePropMetadata.php
```diff
@@ -53,6 +54,11 @@ public function allowsNull(): bool
         return $this->allowsNull;
     }
 
+    public function getQueryStringMapping(): array
```
weaverryan: Let's add some phpdoc above this method to describe the array's shape

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,66 @@
...
+            return [];
+        }
+
+        $query = HeaderUtils::parseQuery($queryString);
+
+        $data = [];
+
+        foreach ($metadata->getAllLivePropsMetadata() as $livePropMetadata) {
+            $queryStringBinding = $livePropMetadata->getQueryStringMapping();
+            foreach ($queryStringBinding['parameters'] ?? [] as $parameterName => $paramConfig) {
```
weaverryan: No need to code defensively here - if `parameters` is missing, it would be a bug in our code I think

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,66 @@
...
+
+        if (Type::BUILTIN_TYPE_ARRAY === $config['type'] && isset($config['collectionType'])) {
+            foreach ($value as &$v) {
+                settype($v, $config['collectionType']);
+            }
+        } else {
+            settype($value, $config['type']);
+        }
+
+        return $value;
```
weaverryan: Yup, I'd definitely like to see this simpler & supporting less complex cases :)

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
```diff
@@ -0,0 +1,52 @@
...
+            this.updateUrl(component);
+        });
+
+        component.on('render:finished', (component: Component)=> {
+            this.updateUrl(component);
+        });
+    }
+
+    private registerBindings(): void {
+        const rawQueryMapping = (this.element as HTMLElement).dataset.liveQueryMapping;
```
weaverryan: Hmm, probably better if we make this a real Stimulus value on the element. Then:

A) Read the value from the stimulus controller
B) Pass the mapping into the constructor of this class

I try to keep the `Component` class and plugins not ultimately dependent on Stimulus or the HTML structure as much as possible.

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
```diff
@@ -0,0 +1,52 @@
...
+export default class implements PluginInterface {
+    private element: Element;
+    private mapping: Map<string,QueryMapping> = new Map;
+
+    attachToComponent(component: Component): void {
+        this.element = component.element;
+        this.registerBindings();
+
+        component.on('connect', (component: Component) => {
+            this.updateUrl(component);
```
weaverryan: My impression from Livewire is that we would only update the URL on page load if they have a (future) `keep: true` option. But, by default, the URL only changes after the value has been changed on the page - https://livewire.laravel.com/docs/url#display-on-page-load

norkunas:
> Actually I'm wondering if the DTO support really worth it. For the moment it's quite limited:
> 
> * As we should not use dots in query parameters, we use `_`. This is kind of naming strategy that should be abstracted.
> * Nested DTO properties are not allowed. This is mainly because I didn't yet implement any config to _select_ some properties of a DTO prop, so it avoids problems such as self recursion etc.
> 
> The DTO support, if shipped with this feature, seems close to the Serializer concepts (max depth, projection, sub context, naming strategy, etc.). On the other hand, handling scalar values is really easy. So handling a custom DTO in a custom way, so the developer can eventually manage the scalar properties he/she wants, should be easy too.
> 
> That's all to say, I suggest to drop the incomplete support of DTOs, and instead provide an extension point (an interface or some tag injection) so :
> 
> * The default extractor handles scalars and arrays of scalars
> * Additional strategies could be added in custom services by the developer

:disappointed: 

Live component could set same urls as like submitting a form with GET method where query string is set to `parent[field][..]` etc. with brackets, I'd expect to work this same way, so I could refactor our live search to REAL live search :)

src/LiveComponent/assets/src/url_utils.ts
```diff
@@ -0,0 +1,57 @@
...
+    delete(name: string) {
+        super.delete(name);
+        const pattern = new RegExp(`^${name}(\\[.*])?$`);
+        for (const key of Array.from(this.keys())) {
+            if (key.match(pattern)) {
+                super.delete(key);
+            }
+        }
+    }
+}
```
weaverryan, line 120: My initial thought is that it looks like you've done a really nice job. If It helps as a reference, I know that Caleb from Livewire worked very hard on the more complex query param creation for Livewire 3 - https://github.com/livewire/livewire/blob/d4839e3b2c23fc71e615e68bc29ff4de95751810/js/plugins/history/index.js#L176

Depending on which more-complex values we handle (e.g. do we handle a `LiveProp` that is an array of strings? What about an array, or array of strings?), we'll need to make sure the query param format works.

weaverryan:
> Live component could set same urls as like submitting a form with GET method where query string is set to parent[field][..] etc. with brackets, I'd expect to work this same way, so I could refactor our live search to REAL live search :)

As I said, I'm interested in getting a rich feature without going crazy on complexities. However, @norkunas may have a point here. We already have a dehydration system that is able to convert a DTO to an array of simple data - e.g. `AddressDto` dehydrates to `['street' => '404 Page Drive', 'city' => 'London', ...]`. So, in theory, if `AddressDto` has `url: true`, could we set its "query string mapping" to be `address[street]=404%20Page%20Drive&address[city]=London`?

squrious:
> > Live component could set same urls as like submitting a form with GET method where query string is set to parent[field][..] etc. with brackets, I'd expect to work this same way, so I could refactor our live search to REAL live search :)
>
> As I said, I'm interested in getting a rich feature without going crazy on complexities. However, @norkunas may have a point here. We already have a dehydration system that is able to convert a DTO to an array of simple data - e.g. `AddressDto` dehydrates to `['street' => '404 Page Drive', 'city' => 'London', ...]`. So, in theory, if `AddressDto` has `url: true`, could we set its "query string mapping" to be `address[street]=404%20Page%20Drive&address[city]=London`?

Seems legit, associative arrays should already be supported. That would have been a better solution that the underscores for the current DTO support. I'll take a look :)

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
squrious: I wasn't sure about this one, Livewire illustrates the `keep` behavior with an initial empty value, and I also considered a non-empty initial value, that could be set on component mount for example, based on some context. But it may be too much for the moment, so let's wait for the `keep` implementation!

WebMamba:
> We already have a dehydration system that is able to convert a DTO to an array of simple data - e.g. AddressDto dehydrates to ['street' => '404 Page Drive', 'city' => 'London', ...]. So, in theory, if AddressDto has url: true, could we set its "query string mapping" to be address[street]=404%20Page%20Drive&address[city]=London

I agree here, but I think we should do some refactoring on the Hydrator class instead of building new hydration logic. We don't want to maintain two hydrations logic since we want this feature to mirror the rest. So we should add some public function to hydrade or dehydrate function.

What do you think about making some refactoring? I have the following implementation in mind:
- we keep the `LiveComponentHydrator`
- this `LiveComponentHydrator` takes as argument a tag iterator of `PropertyHydrator`
a `PropertyHydrator` is an object that implemented the following interface:

````php
  interface PropertyHydratorInterface
{
    public function hydrate(): mixed;

    public function dehydrate(): array;

    public function support(): bool;
}
```
- The `LiveComponentHydrator` foreach properties of the component loop over the tag iterator of `PropertyHydrator` find the one that support the property type.
This is a similar implementation we already did for the DoctrineHydrator. We can have the following `PropertyHydrator`:
DateTime, Enum, Scalar, Dto, Collection, Array.
What do you think?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: I chose to always return an array from `$livePropMetadata->getQueryStringMapping()`, possibly populated with configuration values, such as `parameters`. So all unbound props have an empty array here. Maybe returning `false` instead of the empty array could be better?

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
squrious: Done in [717f726](https://github.com/symfony/ux/pull/1230/commits/717f7260117375ad7362aa13527a5e89efaf39d2)!

Should we update the [polling plugin](https://github.com/squrious/symfony-ux/blob/2f3e9279ecd6ba0be8aa3407da0bc5a8d09f9760/src/LiveComponent/assets/src/Component/plugins/PollingPlugin.ts#L35) too?

src/LiveComponent/assets/src/url_utils.ts
squrious: Thanks for the reference! I'm about to setup more complex use cases to determine the limits of my current implementation. This will help a lot!

squrious:
Hi there! 

I addressed most of your comments, in separated commits:
- Pass query mapping with standard Stimulus value
- Don't update the URL on first load even if some props are initialized on the PHP side
- Use the component hydrator to parse query string data on first render

The latter now allows to initialize complex objects from the URL, with the `prop[foo]=some+string&prop[bar]=42` syntax. It also simplifies a lot the piece of code for the mapping, as we don't need the type guessing logic anymore. I still need to set up more complex use cases to test edge cases :wink:

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: Using the `LiveComponentHydrator` simplified this part a lot!

src/LiveComponent/src/Metadata/LivePropMetadata.php
squrious: Done! Newer version is simpler, but future configuration options will be documented here as well

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
weaverryan: Hmm, I think so yes! But in a different PR :)

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
```diff
@@ -0,0 +1,100 @@
+import Component from '../index';
+import { PluginInterface } from './PluginInterface';
+import {
+    setQueryParam, removeQueryParam,
+} from '../../url_utils';
```
weaverryan: One line - sometimes I'll use multiple lines for a lot, but not this short

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        component.on('render:finished', (component: Component) => {
+            this.initialPropsValues.forEach((initialValue, prop) => {
+                const value = component.valueStore.get(prop);
+
+                // Only update the URL if the prop has changed
+                this.changedProps[prop] ||= JSON.stringify(value) !== JSON.stringify(initialValue);
+                if (this.changedProps) {
+                    this.updateUrlParam(prop, value);
+                }
```
weaverryan: It makes more sense to me to:

A) loop over `mapping` and grab the new value of each item
B) Try to update the URL all at once. 

So, for example, if we have 2 mappings - e.g. `query` and `isDeleted` - instead of changing the URL for `query` and then changing the URL for `isDeleted`, we should grab the data for both fields and update the URL just one time. This will also fit better later with the option to push onto the history (we only want to push 1 time for 1 re-render, not 1 time for each model).

This should, I think, remove also the need to check for "changed props". We would pass all of the "url" field data to some method, it would calculate the new URL and, only if the new URL is different, it would change it. It would also remove the need for `initialPropsValue`.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+            PreMountEvent::class => 'onPreMount',
+        ];
+    }
+
+    public function onPreMount(PreMountEvent $event): void
+    {
+        $component = $event->getComponent();
+        if (!($metadata = $this->registered[$component::class] ?? null)) {
+            return;
+        }
```
weaverryan, line 60: It seems reasonable to me to add `ComponentMetadata` to `PreMountEvent` and `PostMountEvent`.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -410,7 +410,7 @@ private function dehydrateObjectValue(object $value, string $classType, ?string
         return $dehydratedObjectValues;
     }
 
-    private function hydrateValue(mixed $value, LivePropMetadata $propMetadata, object $parentObject): mixed
+    public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, object $parentObject): mixed
```
weaverryan: We should add a phpdoc method description to this and it should be moved up above the private methods

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
weaverryan: I see - I'm ok with it.

weaverryan (commented):
Looking closer and closer

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
squrious: That makes sense to me! I didn't consider the drawbacks of the current implementation for the push history strategy.

But for the last part I'm not sure. If you have a prop that is initialized on the PHP side, but not changed on client side, we should not show it in the URL (unless we have the `keep` option enable). I think we still need to track the changes in the plugin to prevent a such prop to appear in the URL after another prop changed.

For example: 

```php
#[AsLiveComponent]
class MyComponent {
    #[LiveProp(writable: true, url: true)]
    public ?string $search = null;
    
    #[LiveProp(writable: true, url: true)]
    public string $mode = 'fulltext';
}
```

If we update `search` on client side, we should get the `?search=some+string` URL. But if we don't track the initial state of `mode`, as it's not empty, we'll get `?search=some+string&mode=fulltext`. I think that's not expected, wdyt?

squrious:
Hi! I made some updates, especially on the URL utils. I ported the code from Livewire [as suggested](https://github.com/symfony/ux/pull/1230#discussion_r1372978397), which brings some improvements:
- No need to "expand" prop names anymore, we can set complex parameters directly in the URL. I guess it could also better handle dynamically added nested complex objects, as we don't need to know the data structure before updating query params.
- Array values are explicitly indexed in the URL. That's fine but also required additional sort on server side, as `InputBag::all()` doesn't sort the values by itself.
- Parameters order is kept between URL changes (former implementation always put objects and arrays at the end of the query string).

The only main change I made to the original code is to avoid null values in the URL. I found it annoying that setting a single property in a DTO brings all other uninitialized values in the query string. Furthermore, depending on the input type of the prop, nulls may be wrongly casted (got the case with a numeric input filled with `0` because of this).

Also, now the URL is updated once after render. But in Livewire, it seems it is updated on each prop change, because all props may have their own history strategy. Not sure about how we could handle this in our feature?

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
weaverryan: Hmm. I see your point, but I'm not sure we should care. I think (?) the main thing that people will want to avoid is: "when I go to `/foo`, LiveComponents suddenly changes the URL to `/foo?search=bar&mode=fulltext` without the user doing anything". Once the user *does* change one of the `url` props, I don't think it's a big deal to have them all in the URL. For me, it's like submitting a "search" form. Even if I only type into the "search" box (but never change `mode`), I expect both `search` and `mode` to be in the query params.

Also... not tracking really simplifies things :)

weaverryan:
> Also, now the URL is updated once after render. But in Livewire, it seems it is updated on each prop change, because all props may have their own history strategy. Not sure about how we could handle this in our feature?

Hmm, that's a good point. Though, if I have 2 props that *both* will use `history: true`, does Livewire actually add *two* entries into the history (even if both prop changes resulted from a single re-render)? You don't need to actually dig in and answer that, but that looks weird to me. If I change `search` and `mode` and *then* click a button to re-render, that should result in just one history entry. 

But what if `search` is `history: true` and `mode` is `history: false`? Obviously, if I ONLY change `mode`, then no new history entry should be made. And if I ONLY change `search`, then a new history entry *should* be added. But what if I change both at the same time and *then* re-render? I think the correct answer is that one new history entry should be added. So, by this logic (and please tell me if you disagree), we *should* update the URL for all props at the same time. And if even *one* of the changed props has `history: true`, then we would use `pushState()` to add a new history entry.

squrious:
> I think the correct answer is that one new history entry should be added. So, by this logic (and please tell me if you disagree), we should update the URL for all props at the same time. And if even one of the changed props has history: true, then we would use pushState() to add a new history entry.

That seems to me the better way to go. Thanks for your advises! I keep this in mind for the future implementation of the `history` feature.

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
squrious: Alright, I removed tracking from the plugin :wink:

squrious:
> Nice work here! I'm curious how multiple components using this feature would behave. What about multiple of the same component?

Hello! I made some experimentation on my local project to handle the case of multiple components. I think we should handle an optional attribute to prefix all parameter names in the component (so it acts like a "scope"). Thought about something like:

```html
<twig:MyComponent query-param-prefix='prefix_'>
```
So the URL looks like `?prefix_prop1=foo`.

And maybe we could also leverage the special `key` attribute as a shortcut for a such attribute?

weaverryan:
That seems reasonable. Though, if 2 components are on the same page and are using the `url` functionality, as long as their parameter names don't collide, won't it work already? If `ComponentA` is adding a `search` param and `ComponentB` is adding a `page` param, if the URL is currently `?search=hello&page=1` and `ComponentA` re-renders, will the current implementation "keep" the `page` param - e.g. update it to `?search=changed&page=1`?

Also, I'm super busy this week - but this is an important PR. As far as you know, is this ready to go (other than needing a rebase - thanks!) and is it waiting for a detailed review?

squrious:
> That seems reasonable. Though, if 2 components are on the same page and are using the `url` functionality, as long as their parameter names don't collide, won't it work already? If `ComponentA` is adding a `search` param and `ComponentB` is adding a `page` param, if the URL is currently `?search=hello&page=1` and `ComponentA` re-renders, will the current implementation "keep" the `page` param - e.g. update it to `?search=changed&page=1`?

Yes, I confirm that two different components can share the URL if their parameters don't collide. I added an implementation that uses the special `query-param-prefix` attribute to prefix parameters, and used `key` as a shortcut (but it is easily removable).

> Also, I'm super busy this week - but this is an important PR. As far as you know, is this ready to go (other than needing a rebase - thanks!) and is it waiting for a detailed review?

Rebase done :wink:

I think the current code runs well and is tested enough. We already discussed the expected behaviors a lot so it should be fine. But I added two more commits which may need a quick check, if you have the time of course!
- [aff83f2](https://github.com/symfony/ux/pull/1230/commits/aff83f258ce1e0b98a3122e7aceae22900088214) : Simplify the mapping structure on PHP side, and fix some edge cases in URL data management on JS side
- [23d4b7c](https://github.com/symfony/ux/pull/1230/commits/23d4b7c9b124a69e7f046124503fefaddd10df24): The prefix attribute implementation mentioned above

Btw, I also worked on the other features (keep, alias, history) and they work pretty nicely.  I can create a new PR when this one gets merged :)

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
```diff
@@ -0,0 +1,53 @@
...
+    constructor(private readonly mapping: {[p: string]: QueryMapping}) {}
+
+    attachToComponent(component: Component): void {
+        component.on('render:finished', (component: Component) => {
+            const urlUtils = new UrlUtils(window.location.href);
+            const currentUrl = urlUtils.toString();
+
+            Object.entries(this.mapping).forEach(([prop, mapping]) => {
+                const value = component.valueStore.get(prop);
+                if (this.isEmpty(value)) {
```
weaverryan: Can you explain the motivation for taking time to remove the "empty" query params? What if I have a `LiveProp` called `search` which defaults to `foo` in my PHP class. Then, the user changes an input bound to this to `''`. If that removes `?search` from the URL, then if I refresh, won't my `search` property once again be set to `foo` (since there was no query parameter to override it)?

src/LiveComponent/assets/src/url_utils.ts
```diff
@@ -0,0 +1,138 @@
+function isObject(subject: any) {
+    return typeof subject === 'object' && subject !== null;
+}
```
weaverryan, line 32: function is only used once - let's inline the logic below where it's used

src/LiveComponent/assets/src/url_utils.ts
weaverryan: Can you add a little note, perhaps new the top of the file, with a link to this file in Livewire - something like:

> Adapted from https://github.com/livewire/livewire/blob/d4839e3b2c23fc71e615e68bc29ff4de95751810/js/plugins/history/index.js#L176

It's nice to give Livewire the deserved nod and also will help us, in the future, answer "why did we do it this way?"

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+        return [
+            PreMountEvent::class => 'onPreMount',
+        ];
+    }
+
+    public function onPreMount(PreMountEvent $event): void
+    {
+        $metadata = new LiveComponentMetadata(
+            $event->getMetadata(),
+            $this->metadataFactory->createPropMetadatas(new \ReflectionClass($event->getComponent()::class))
```
weaverryan: Could it just be this?

```php
$metadata = $this->metadataFactory->getMetadata($event->getMetadata()->getName());
```

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $component = $event->getComponent();
+
+        $data = $event->getData();
+
+        $request = $this->requestStack->getMainRequest();
+
+        $prefix = $data[LiveControllerAttributesCreator::URL_PREFIX_PROP_NAME]
+            ?? $data[LiveControllerAttributesCreator::KEY_PROP_NAME]
+            ?? '';
```
weaverryan: I'm not sold on this automatic collision avoidance. My gut is that I'd rather keep things simple. Then (later) give users the ability to control the name of the query parameter in the URL. Then, if they have a collision, they can decide which of the 2 colliding props to change the "parameter" for.

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @internal
+ */
+final class QueryStringPropsExtractor
+{
+    public function __construct(private readonly LiveComponentHydrator $hydrator)
+    {
+    }
+
+    public function extract(Request $request, LiveComponentMetadata $metadata, object $component, string $prefix = ''): array
```
weaverryan: Add phpdoc:

> Extracts relevant query parameters from the current URL & hydrates them.

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -194,9 +194,9 @@ private function preMount(object $component, array $data): array
     /**
      * @return array{data: array<string, mixed>, extraMetadata: array<string, mixed>}
      */
-    private function postMount(object $component, array $data): array
+    private function postMount(object $component, array $data, ComponentMetadata $componentMetadata): array
     {
-        $event = new PostMountEvent($component, $data);
+        $event = new PostMountEvent($component, $data, $componentMetadata);
```
weaverryan, line 199: Add a CHANGELOG entry for TwigComponents for this please!

src/TwigComponent/src/Event/PostMountEvent.php
```diff
@@ -21,6 +22,7 @@ final class PostMountEvent extends Event
     public function __construct(
         private object $component,
         private array $data,
+        private ComponentMetadata $metadata,
         private array $extraMetadata = [],
     ) {
```
weaverryan, line 30: This is a BC break, though almost definitely minor. But we should handle it. General steps:

A) Remove the types from the 3rd and 4th argument
B) If the 3rd argument is an array, trigger a deprecation - something like:

> In TwigComponent 3.0, the 3rd argument of `PostMountEvent` will be a `ComponentMetadata` object.

And, in this case, set the 3rd arg `$metadata` will be set onto the `$extraMetadata` property (so, also, we will need to remove the `private` from args 3 and 4 and set them as normal properties).

Also in this situation, set the `$metadata` property to null. We'll need to make the return type on `getMetadata()` nullable for now.

src/TwigComponent/src/Event/PreMountEvent.php
```diff
@@ -12,13 +12,14 @@
...
 use Symfony\Contracts\EventDispatcher\Event;
+use Symfony\UX\TwigComponent\ComponentMetadata;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
  */
 final class PreMountEvent extends Event
 {
-    public function __construct(private object $component, private array $data)
+    public function __construct(private object $component, private array $data, private ComponentMetadata $metadata)
```
weaverryan: Make this argument optional defaulting to null. If it's null, trigger a deprecation. Make the getter below have a nullable return type.

weaverryan (commented):
Down to the small stuff. I played with this and it works FANTASTICALLY - super excited to have this!

src/LiveComponent/assets/src/Component/plugins/QueryStringPlugin.ts
squrious: I didn't consider this use case. You're right, providing an empty value in the URL should override values defined in PHP. The initial motivation for this check was to avoid empty values for objects and arrays. But I'll find a better way to manage these edge cases!

src/TwigComponent/src/Event/PostMountEvent.php
squrious: If both `$metadata` and `$extraMetadata` are provided, should I enforce their types in the constructor body so the error comes from the constructor and not the assignment?

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: In think the prefix will still help in solving collisions if we have the same component rendered multiple times. Even if users can control the mapping from the template in the future, they will still need to override the full mapping, while a prefix easily solves the collision with a single attribute. And without the ability to override the mapping in attributes, it's a way to avoid collisions in other use cases.  

However, I agree that adds more complexity to this first implementation, and maybe it's not worth it right now. If you prefer I can drop this attribute support until we have a better solution.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: I hoped that would have worked, but it breaks the hydrator tests because the tested component is mounted manually and not registered in the class map:

```
InvalidArgumentException: Unknown component "__testing". And no matching anonymous component template was found.
```

Is there something to fix here?

src/TwigComponent/src/Event/PostMountEvent.php
weaverryan: Yup. So, use an if statement and trigger a deprecation if they pass the "old" way of doing things. But then, if they just pass some bad value - e.g. they pass a string to the 3rd arg - which was NEVER legal - check that and throw an exception.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
weaverryan: I'm worried about people getting confused when they're query params suddenly have a different name in the URL. Heck, it's also possible that people will purposely want to have a collision e.g. 2 separate search forms that both have `query` (though, they would need to do extra work to make it so that when the `query` changes in one component, it also updates the other - but this is totally possible)

squrious: Alright, I dropped the commit with the prefix management

squrious:
> Down to the small stuff. I played with this and it works FANTASTICALLY - super excited to have this!

I'm glad to hear that! 

Latest commit handles bc break for TwigComponent and fix empty params removal :)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+            $this->metadataFactory->createPropMetadatas(new \ReflectionClass($event->getComponent()::class))
+        );
+
+        if (!$metadata->hasQueryStringBindings()) {
+            return;
+        }
+
+        $queryStringData = $this->queryStringPropsExtractor->extract($request, $metadata, $event->getComponent());
+
+        $event->setData(array_merge($event->getData(), $queryStringData));
```
weaverryan, line 64: We have a tricky problem here. Suppose I have:

```php
    #[LiveProp(writable: true, url: true)]
    public array $other = [];
```

Then I purposely go to `/my-page?other=a-string`. Obviously, this `a-string` is an invalid value. And so this shouldn't happen in practice. However, it currently triggers a 500 error:

Screenshot 2023-11-16 at 9 07 11 AM

As you can see the 500 comes from the mounting process. Normally, it is totally ok to fail with a 500 during the mounting process, as the developer is providing the mounting props directly (so if they mess something up, we want to tell them). But in this case, the mounted data is coming from the user. So if it's invalid, it needs to be ignored. To make things trickier, a could be mounted directly on a property, or via a setter or even by passing it as an argument to the `mount()` method.

Here's my idea, which I think will be solid enough. In `QueryStringsPropsExtractor`, after hydrating the data, we do a type-check on the property. If the type of the hydrated data is incompatible with the property type (if there is one), we do NOT add the hydrates value to `$data`. WDYT?

weaverryan (commented):
Played with it again - I thing we're there, except for this last item I found :)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: That makes sense for me! I updated the code with your suggestion. I also ignored issues from the hydrator itself to prevent errors when hydrating objects from invalid values. I ignored `HydrationException` errors, but I wonder if ignoring any error would be better.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+        private readonly RequestStack $requestStack,
+        private readonly LiveComponentMetadataFactory $metadataFactory,
+        private readonly QueryStringPropsExtractor $queryStringPropsExtractor,
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            PreMountEvent::class => 'onPreMount',
```
smnandre, line 39: If i read it correctly you will listen to all components (Live or not) with this Subscriber no ?

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -226,6 +227,55 @@ public function hydrate(object $component, array $props, array $updatedProps, Li
...
+        if ($propMetadata->hydrateMethod()) {
+            if (!method_exists($parentObject, $propMetadata->hydrateMethod())) {
+                throw new \LogicException(sprintf('The "%s" object has a hydrateMethod of "%s" but the method does not exist.', $parentObject::class, $propMetadata->hydrateMethod()));
+            }
+
+            return $parentObject->{$propMetadata->hydrateMethod()}($value);
+        }
+
+        if ($propMetadata->useSerializerForHydration()) {
+            return $this->normalizer->denormalize($value, $propMetadata->getType(), 'json', $propMetadata->serializationContext());
```
smnandre: $propMetadata->getType is nullable

src/LiveComponent/src/Attribute/LiveProp.php
```diff
@@ -58,6 +58,13 @@ final class LiveProp
      */
     private null|string|array $onUpdated;
 
+    /**
+     * @var bool
+     *
+     * Tells if this property should be bound to the URL
+     */
+    private bool $url;
```
smnandre, line 66: If i did not read the code i'd believe you expected a URL string. Maybe something more explicit like liveParam ou boundToUrl ?

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -226,6 +227,55 @@ public function hydrate(object $component, array $props, array $updatedProps, Li
         return $attributes;
     }
 
+    /**
+     * Hydrate a value from a dehydrated value.
+     *
+     * Depending on the prop configuration, the value may be hydrated by a custom method or the Serializer component.
+     *
+     * @throws SerializerExceptionInterface
```
smnandre, line 237: @internal ?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -95,24 +95,24 @@ public function createLivePropMetadata(string $className, string $propertyName,
...
-            );
+            $isTypeBuiltIn = null === $infoTypes[0]->getClassName();
+            $isTypeNullable = $infoTypes[0]->isNullable();
+        } else {
+            $infoType = $type?->getName();
+            $isTypeBuiltIn = $type?->isBuiltin() ?? false;
+            $isTypeNullable = $type?->allowsNull() ?? true;
         }
 
+        $queryStringBinding = $this->createQueryStringMapping($propertyName, $liveProp);
```
smnandre: I'd test here if it's necessary

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -130,6 +130,17 @@ private static function propertiesFor(\ReflectionClass $class): iterable
         }
     }
 
+    private function createQueryStringMapping(string $propertyName, LiveProp $liveProp): array
```
smnandre: Is it used elsewhere ?

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
```diff
@@ -108,6 +108,18 @@ public function attributesForRendering(MountedComponent $mounted, ComponentMetad
             }
         }
 
+        $liveMetadata = $this->metadataFactory->getMetadata($mounted->getName());
```
smnandre, line 101: Can a LiveComponent be a child component AND use urlquery string ?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+        if (empty($query)) {
+            return [];
+        }
+        $data = [];
+
+        foreach ($metadata->getAllLivePropsMetadata() as $livePropMetadata) {
+            if ($queryStringMapping = $livePropMetadata->getQueryStringMapping()) {
+                if (null !== ($value = $query[$queryStringMapping['name']] ?? null)) {
+                    if (\is_array($value) && $this->isNumericIndexedArray($value)) {
+                        // Sort numeric array
```
smnandre: Why ?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+                        $value = [];
+                    }
+
+                    try {
+                        $hydratedValue = $this->hydrator->hydrateValue($value, $livePropMetadata, $component);
+
+                        if ($this->isValueTypeConsistent($hydratedValue, $livePropMetadata)) {
+                            // Only set data if hydrated value type is consistent with prop metadata type
+                            $data[$livePropMetadata->getName()] = $hydratedValue;
+                        }
```
smnandre, line 60: Would that not silent problems one could want to know about ?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+        }
+
+        return $data;
+    }
+
+    private function isNumericIndexedArray(array $array): bool
+    {
+        return 0 === \count(array_filter(array_keys($array), 'is_string'));
```
smnandre: `array_filter($foo, is_string(...),  ARRAY_FILTER_USE_KEY)` ?

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
smnandre, line 64: I'm not sure this should be done there. Or at leasst without additional safeguards.

Maybe i missread the code, but if the live component uses ComponentValidator for example,.... it will not be called right now ?

Because the data set there is the data which will be used to render the component, to write in HTML, etc..

I may be wrong... but i'm pretty sure there is a big security risk there.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -130,6 +130,17 @@ private static function propertiesFor(\ReflectionClass $class): iterable
...
+    {
+        if (false === $liveProp->url()) {
+            return [];
+        }
+
+        return [
+            'name' => $propertyName,
```
smnandre: If there is only a string name, could we start with a unique string property ?

src/LiveComponent/src/Metadata/LivePropMetadata.php
```diff
@@ -53,6 +54,14 @@ public function allowsNull(): bool
         return $this->allowsNull;
     }
 
+    /**
+     * @return array{'name': string}
```
smnandre: `array{'name'?: string} `

or 

`array{'name': string}|null `

I'd rather have a nullable here cause at least if this is not null we will be certain to have a "name" key.

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+use Symfony\UX\LiveComponent\Metadata\LivePropMetadata;
+
+/**
+ * @author Nicolas Rigaud <squrious@protonmail.com>
+ *
+ * @experimental
+ *
+ * @internal
+ */
+final class QueryStringPropsExtractor
```
smnandre, line 27: Maybe find a name which more explicitely says that it extracts, but also _**hydrates**_ into the component !

src/TwigComponent/tests/Unit/EventListener/TwigComponentLoggerListenerTest.php
```diff
@@ -38,9 +38,9 @@ public function testLoggerStoreEvents(): void
         $eventB = new PreCreateForRenderEvent('b');
         $logger->onPreCreateForRender($eventB);
 
-        $eventC = new PreMountEvent(new \stdClass(), []);
+        $eventC = new PreMountEvent(new \stdClass(), [], new ComponentMetadata([]));
```
smnandre, line 41: Maybe one event with "null" ? As it's a nullable argument

src/TwigComponent/src/Event/PostMountEvent.php
smnandre: That also would require to check in LiveComponent if the installed TwigComponent has the good version (as they are two different packages)

src/LiveComponent/assets/src/url_utils.ts
```diff
@@ -0,0 +1,169 @@
...
+        }
+    });
+
+    return data;
+}
+
+/**
+ * Wraps a URL to manage search parameters with common map functions.
+ */
+export class UrlUtils extends URL {
```
smnandre, line 125: If i understand correctly, the URL is only changed after a replace (from the backend)..

Do we really need to duplicate all that URL / query logic in the front ? 

Is it not possible to send the URL from the back-end ? I must miss something here I think :)

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
```diff
@@ -0,0 +1,87 @@
...
+            return [];
+        }
+        $data = [];
+
+        foreach ($metadata->getAllLivePropsMetadata() as $livePropMetadata) {
+            if ($queryStringMapping = $livePropMetadata->getQueryStringMapping()) {
+                if (null !== ($value = $query[$queryStringMapping['name']] ?? null)) {
+                    if (\is_array($value) && $this->isNumericIndexedArray($value)) {
+                        // Sort numeric array
+                        ksort($value);
```
smnandre: I made a few tests, and i don't think we should.

Or there will be unnecessary changes (if per example i had [2 => 2, 1 => 1] in the liveprop value)

src/LiveComponent/tests/Functional/Util/QueryStringPropsExtractorTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+        $this->assertEquals($expected, $data);
+    }
+
+    public function getQueryStringTests(): iterable
+    {
+        yield from [
+            ['', []],
+            ['prop1=foo', ['prop1' => 'foo']],
+            ['prop2=42', ['prop2' => 42]],
```
smnandre: There are edge cases i'm not sure how we should handle. 

But if you add

```php
    #[LiveProp(writable: true, url: true)]
    public int $prop6 = 123;

    #[LiveProp(writable: true, url: true)]
    public ?int $prop7 = 123;
```

Then in the tests, i got those cases i'm not sure of

(when i write `--> []` it is the result obtained)

```php
            ['prop6=122', ['prop6' => 122]],
            ['prop6=', []], // NOK --> ['prop6' => 0]
            ['prop6=0', ['prop6' => 0]],
            ['prop6=-1', ['prop6' => -1]],
            ['prop6=00', []], // or exception NOK --> ['prop6' => 0]
            ['prop6=01', []], // or exception NOK --> ['prop6' => 1]
            ['prop6=abc', []], // or exception NOK --> ['prop6' => 0]

            ['prop7=', ['prop7' => null]], // NOK --> []
```

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
smnandre: Do we want to use "fieldName" there ?

src/LiveComponent/tests/Fixtures/Component/ComponentWithUrlBoundProps.php
```diff
@@ -0,0 +1,29 @@
...
+#[AsLiveComponent('component_with_url_bound_props')]
+class ComponentWithUrlBoundProps
+{
+    #[LiveProp(writable: true, url: true)]
+    public ?string $prop1 = null;
+
+    #[LiveProp(writable: true, url: true)]
+    public ?int $prop2 = null;
+
+    #[LiveProp(writable: true, url: true)]
```
smnandre: Remove writable:true as it is not used ?

src/LiveComponent/tests/Fixtures/Component/ComponentWithUrlBoundProps.php
```diff
@@ -0,0 +1,29 @@
...
+    public array $prop3 = [];
+
+    #[LiveProp(writable: true)]
+    public ?string $prop4 = null;
+
+    #[LiveProp(writable: ['address', 'city'], url: true)]
```
smnandre: Remove writable:['address', 'city'] as it is not used ?

src/TwigComponent/src/Event/PostMountEvent.php
```diff
@@ -12,17 +12,38 @@
...
+            $this->extraMetadata = $metadata;
+        } else {
+            if (null !== $metadata && !$metadata instanceof ComponentMetadata) {
+                throw new \InvalidArgumentException(sprintf('Expecting "$metadata" to be null or an instance of "%s", given: "%s."', ComponentMetadata::class, get_debug_type($metadata)));
+            }
+            if (!\is_array($extraMetadata)) {
+                throw new \InvalidArgumentException(sprintf('Expecting "$extraMetadata" to be array, given: "%s".', get_debug_type($extraMetadata)));
+            }
+
+            $this->metadata = $metadata;
```
smnandre, line 44: There is a regression here.

Before it was possible to do 

new PostMountEvent($component, ['foo'])


I think we can do the following constructor to keep things working as before, and still a bit typed

```
    public function __construct(
        private object $component,
        private array $data,
        array|ComponentMetadata $metadata = [],
        array $extraMetadata = [],
    ) {
```

smnandre:
Sorry for the flood i did not start thinking i'd write many comments :)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: Yes, and I think see your point here. Something like this?

```php
    public function onPreMount(PreMountEvent $event): void
    {
        if (!$event->getMetadata()->get('live', false)) {
            // Not a live component
            return;
        }
```

src/LiveComponent/src/Attribute/LiveProp.php
squrious: Seems legit. I also consider future url options that could be passed to `LiveProp` constructor. I see two ways: 

Allow an object to configure options:

Then `urlMapping` will be better imo. We could still allow a boolean, so `true` leads to basic defaults, and `false` disables the binding. Depending on when these new options get available, it could also improve BC.

Pass each option separately: 

Then `boundToUrl` seems good to me. But maybe passing each option separately will complexify the constructor too much.

wdyt?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
squrious: Nope, good catch, I think we could remove this function.

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
squrious: Yes it can, what are the implications you'd have in mind here?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: We discussed this point [here](https://github.com/symfony/ux/pull/1230#discussion_r1395754272). @weaverryan suggested that inconsistencies in data are not relevant to be propagated from there, because they come from a user input. Do you have other use cases in mind?

src/LiveComponent/src/Attribute/LiveProp.php
smnandre: No particular preference... i wonder is something "query" would not be more accurate / self-explanatory ..

If it was a new attribute (may be an idea if we need many attributes) ...  i think i'd call it LiveQueryParam so maybe .. something like queryParam ? 
mapQuery ? 
queryName ? 

What do you have in mind for the object / multiple options ? That could help me better see the vision you have for this :)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
smnandre: Yes perfect !

 .. but it's another proof we should do something sooner to differenciate the two.. I'm starting to think that an EventFilter/Gateway could help

smnandre: (in general I mean between Twig&Live Components, not related to this PR specifically)

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
smnandre: This could be any attempt that should otherwise raise a security alert 

If someone add `&year=DROP TABLE;`  i think i'd like to know it :)

I'm particulary attentive there because the LiveComponent actions will soon be POST-only, protected by CSRF, etc.. 

But the moment we allow anyone to have a direct access to the data via a queryString can share this URL to make others the last "actor", that requires for me an extra layer of monitoring / checks.

src/LiveComponent/src/Attribute/LiveProp.php
squrious: Potential new options are `alias`, `keep` and `history` (at least if we want to stick with the [Livewire feature](https://livewire.laravel.com/docs/url)). So I thought about either a `QueryMapping` object:
```php
class QueryMapping {
    public function __construct(string $alias, bool $keep = false, bool $history = false) { ... }
}
```

Or params directly in the `LiveProp` constructor: `queryAlias`, `queryKeep`, etc. (or `urlAlias`...).

Even if these features are out of this PR scope, I think that keeping them in mind to choose the appropriate naming may worth it . But maybe I'm overthinking this ^^

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
smnandre: Is this only a tests problem ? If yes let's find a way to resolve it :)

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
smnandre: Just wanted to see if we could early exit somewhere.

But otherwise, we can move thus line before L101 and reuse $liveMetadata in the calculateFingerprint call ?

src/LiveComponent/src/Attribute/LiveProp.php
smnandre: What would happen if two LiveProps had different values for keep or history ?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
smnandre: I have missread the code there apparently...

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: Originally, it was because `InputBag::all()` (used to get query params) doesn't sort array values even if the array contains only numeric keys. The order of elements is the same as their order in the URL. So it depends on how the URL is modified on client side. 

Now the JS code writes array elements in the correct order, so it's not strictly required anymore.

But I'd still expect a such array to be ordered by default, to not rely on the JS implementation. Also, there is no native `kstort` filter in Twig that could be used as a workaround.

smnandre: But why would you sort [2 => 2, 1 => 1] for example ?

squrious: Hum because I tend to expect an array numerically indexed to be ordered. At least from a data source like an URL. But I may miss something here?

src/LiveComponent/src/Attribute/LiveProp.php
squrious: For the `keep` option, it'd actually be used only on first load to tell if the param should appear in the URL, even if it's not specified. So each prop handle its own parameter, they work independently.

For the `history` option, we challenged this in [this comment](https://github.com/symfony/ux/pull/1230#issuecomment-1796349698). Basically, we'd add a new history entry if there is at least one param involved in the URL modification that has the option set.

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
smnandre: I can understand why but ... there is nothing saying that in the [URI RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.1). 

(and that is probably the reason why Symfony\...\Request doesn't do such modification)

And, in fact, the [ ] have no more meaning than "abc" ou "--" ... it's to the backend engine to decide the semantic given to the special chars and how they combine.

The URI (as described in the RFC) must be parsed from left to right. 

PHP found a way to represent array easily, but in the same time lost the `foo=bar&foo=baz` format, that is also a 100% valid URL
(format used by the Web / JS [URL](https://developer.mozilla.org/fr/docs/Web/API/URL) (that you extends from in your uri_utils i think) with its searchUrlParam - almost like how POST data are handled in php) 

-- 

In a pratical point of view, the following arrays are similar, but not equal.

```
$foo = [2 => 2, 0 => 0];
$bar = [0 => 0, 2 => 2];

dump($foo == $bar);    // true
dump($foo === $bar);  // false
dump(serialize($foo) === serialize($bar));  // false
dump(json_encode($foo) === json_encode($bar)); // false
```

Imagine in database i have the array $foo

If i use it as a LiveProp-url-mapped, you will consider it as a different URL / data, that will try to replace the URL... but somehow in the backed i keep my order, and we have an infinite loop

src/LiveComponent/src/Attribute/LiveProp.php
smnandre: Ok that seems logical, and i don't see how we could do "smarter" if a developper uses contradictory values.

👍

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: Alright I got it, thanks for the explanations! I'll remove this part

src/LiveComponent/tests/Functional/Util/QueryStringPropsExtractorTest.php
squrious: The results you got in your tests with `prop6` are because of the string coercing [used in the hydrator](https://github.com/symfony/ux/blob/1f326f06e36a52c2df42d176e388f1378019d484/src/LiveComponent/src/LiveComponentHydrator.php#L245). This also occurs with any live prop type hinted `int` and not nullable (regardless of the url binding). I'd consider it out of this PR scope, but maybe we caught something here...

For `prop7` it's something else, I forgot to check nullability [here](https://github.com/squrious/symfony-ux/blob/live-component/live-prop-url-binding/src/LiveComponent/src/Util/QueryStringPropsExtractor.php#L78), so `null` is wrongly considered invalid. I'll fix that!

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
squrious: Not sure, this is the parameter name used in the URL (currently defaults to the property name). `fieldName` would rather relate to the property name, no?

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
smnandre: Oh I'm 100% open to any other views ..  i was just explaining my question / doubts :)

src/LiveComponent/tests/Functional/Util/QueryStringPropsExtractorTest.php
smnandre: But should we not stop here before the type cohercion ? (genuine question)
If not you're right, that is outside this scope :)

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
smnandre: fieldName is the "name of the property  exposed in the front, if different from the propertyName"..

But your right we should let the possibility to have a field name and a different queryParam name. We should just keep that in mind if a component is submitted.

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: I think the RFC argument wins the debate ;) 
My initial point of view was quite arbitrary, soooo let's stick to something solid!

src/LiveComponent/tests/Functional/Util/QueryStringPropsExtractorTest.php
squrious: > But should we not stop here before the type cohercion ? (genuine question)

You mean with an additional check before we call the hydrator, to ensure the coercion will be consistent?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
squrious: Ow I think I misunderstood your initial remark, you meant that we should prefer the `fieldName` value of the prop instead of the `propertyName`?

src/LiveComponent/src/LiveComponentHydrator.php
squrious: Should we throw an exception if `$propMetadata->getType()` is `null`? Like

```php
if (null === $propMetadata->getType()) {
    throw new \LogicException(sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
}
```

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: Yup, from what I see the `ComponentValidator` is only used after hydration, so only when the component is refreshed in live.

IMO, the initializations performed in `QueryStringInitializerSubscriber` are similar to the arbitrary values you can pass to the component from the template. Those could be set from user data too, it's up to the developer. If they worry about the data consistency, they can hook into the `PostMountEvent` to trigger validation, but that's not required by design (`ComponentValidator` is not even part of the base `TwigComponent`). 

So I don't see the security risk here, or at least the one introduced by this feature. Wdyt?

src/TwigComponent/tests/Unit/EventListener/TwigComponentLoggerListenerTest.php
squrious: The argument is nullable only for BC purpose, I'm not sure that checking this BC fits the goal of this test. Maybe in another test?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
smnandre: Yeah that was my (badly expressed 😅 ) idea

src/TwigComponent/tests/Unit/EventListener/TwigComponentLoggerListenerTest.php
smnandre: Yep you're right, let's focus :)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
smnandre: Anyone can set a value by writing something in the URL... that seems enough to validate, above all if the user added some Assert on that prop

But maybe it's on the ComponentValidator there is something to do indeed

src/LiveComponent/src/LiveComponentHydrator.php
smnandre: Maybe yes ! I don't have a good vision on the impacts, but i don't see many cases where we'd be happy with un undefined type

smnandre:
> You can also check my [demo app](https://github.com/squrious/symfony-ux-demo-live-url) to try the feature.

Finally took the time to install it and play a bit and .... this demo is 🔥 !!  (this PR too)

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
squrious: Ok so yes I agree it should default to the field name

src/LiveComponent/assets/src/url_utils.ts
squrious: Well, it _might_ be possible, but I'd prefer to keep the implementation on frontend side.

I don't think the URL / query logic is really duplicated. I'd rather say it's split by responsibility. The frontend is responsible for the current URL manipulation. I don't know why we would like to delegate the computing to the server, since there is all the required material to do it on client side. It also avoids to pass the current URL as an attribute in each Live Component request.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: Consider the following:

```php
// ...

class MyController extends AbstractController {
    
    #[Route('/{title}')]
    public function index(Environment $twig, string $title): Response
    {
        $template = '<twig:MyComponent title="{{ title }}"></twig:MyComponent>';

        $view = $twig->createTemplate($template)->render(['title' => $title]);
        
        return new Response($view);
    } 
}
```

This is, imo, something valid. If `MyComponent` is a regular TwigComponent, it can not even use the validator from LiveComponent. If the developers want to validate data, they have to implement their own logic in a PostMount hook.

I'm not found of setting up automatic validation in the specific case of a LiveComponent whom props have been set from the query string:
- we'll need to track these props from the `PreMountEvent`
- on `PostMountEvent`, we'll have to get these props back, ensure the current values are the ones set on PreMount, use reflection to determine if the component has validation mechanisms, then trigger validation

I'd rather add a documentation entry in this feature doc, suggesting the use of validation in a PostMount hook.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
squrious: Hum just tried to replace it, but.. The field name calculation depends on the component object. I'll keep property name as a default for the moment ;)

src/LiveComponent/src/Util/QueryStringPropsExtractor.php
squrious: How is it different from providing `year=DROP TABLE;` in an regular input field? The LiveHydrator itself will not raise anything if there is hydration errors (see [here](https://github.com/symfony/ux/blob/85231b9696a20e4d36d3902bd51db059b22156da/src/LiveComponent/src/LiveComponentHydrator.php#L194) and [here](https://github.com/symfony/ux/blob/85231b9696a20e4d36d3902bd51db059b22156da/src/LiveComponent/src/LiveComponentHydrator.php#L212)).

smnandre: No but the validation is triggered if Asserts are defined and the trait is used :)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
smnandre: The developper here made an explicit error himself there. We indeed cannot do anything.

But the second you propose a single on/off attribute in a class, people will use it without thinking there may be consequences. 

For me there is a real danger here (similar to what happened with Autocomplete). We'll see and i hope you are right and that will never happen.

src/LiveComponent/assets/src/url_utils.ts
smnandre: Just a tentative to avoid adding weight to the JS code but i understand your point.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
smnandre: Could you add props wich has a public fieldName in one of the tests, just to be certain all works well ?

squrious: That was a good suggestion, this use case doesn't work properly. I'll do some refactoring here, because the mapping is not consistent with the field name.

squrious:
@smnandre Updates are coming, I compile all fixes in a single commit ;)

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
```
smnandre, line 39: Can you add some tests to see how the different listener interfere (or not) with this listener ? 

I think IntreceptChild / Defer / DataModel  but maybe more, maybe less

And i think we should add priorities to ensure a determined order.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+            return;
+        }
+
+        $request = $this->requestStack->getMainRequest();
+
+        if (null === $request) {
+            return;
+        }
+
+        $metadata = new LiveComponentMetadata(
```
smnandre: I'm less and less convinced with this one... Could we use a single factory everywhere.. This `new` make me fear future problems if we add cache for example in a place and not the other one.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
smnandre: What is blocking precisely ? Can i help you with something ?

smnandre:
Not sure yet about those things (is there something or ... not at all)

But could you explain me what happens if:
* someone adds any query parameters to the _component endpoint ? should they be used ?
* someone has a component with a non writable field BUT queryMapping = true  ? 
* someone has a component with a writable field AND queryMapping = true ?

My intuition would be "it must be writable if a query can change that value" but i'm not sure if some use cases contredict that 

On the opposite, i think the "standard" live hydration starts by setting values, before hydrating the "new ones"... Would that not interfere in a way with the query data ?

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: Sorry I missed that one, I take a closer look

squrious:
>     someone adds any query parameters to the _component endpoint ? should they be used ?

You mean the route used by the LiveComponent bundle to render components? I'd say it should not be processed at all.   

> My intuition would be "it must be writable if a query can change that value" but i'm not sure if some use cases contredict that

You can consider a prop that is not writable directly from the component but can be updated on server side. For example, an array whom items can be added through an input field and deleted with a button, but with a LiveAction. It is _possible_, even if I'd tend to avoid in favor of a stimulus controller.    

> On the opposite, i think the "standard" live hydration starts by setting values, before hydrating the "new ones"... Would that not interfere in a way with the query data ?

I don't think there could be interference here, because the LiveComponentHydrator is used only on live updates. The query data are used on the first load, and at this point the component is mounted with the regular TwigComponent's factory.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: I just tried again with the factory and now it works, because we check if there is a request to process (wasn't checked last time I tried). Not sure if there is still something to fix now.

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
squrious: I don't think it would be useful to test with InterceptChild and Defer, as they don't subscribe to the PreMount event. 

For DataModel, I'll have to setup some advanced use cases with parent and child components in order to test it efficiently. 

My first thought is the query string initialization should come before the DataModel listener. The latter purpose is to set the child prop value from the parent bound prop, and I think replacing this value with the one extracted from the query string will lead the child component in an inconsistent state regarding its parent. Wdyt?

src/LiveComponent/src/Attribute/LiveProp.php
```diff
@@ -58,6 +58,13 @@ final class LiveProp
      */
     private null|string|array $onUpdated;
 
+    /**
+     * @var bool
+     *
+     * Tells if this property should be bound to the URL
+     */
+    private bool $queryMapping;
```
weaverryan: I know you and @smnandre discussed this change... but I did like `url` more. A `url: true` that, potentially later, could be changed to a `bool|UrlMapping` type with some new `UrlMapping` object. Then the user could have `url: new UrlMapping(..)`.

Though, I am ok with keeping the "query mapping" or "url mapping" terminology deeper in the code.

weaverryan (commented):
Thanks for keeping this up to date! For me, just one comment on naming. Then let's get this thing merged :)

src/LiveComponent/src/Attribute/LiveProp.php
squrious: Fixed!

weaverryan:
Sorry - one last change :p. Can you rebase? We have a conflict, I believe, from #1341. Then we'll get this merged!

squrious:
> Sorry - one last change :p. Can you rebase? We have a conflict, I believe, from #1341. Then we'll get this merged!

Should be ok now ;)

weaverryan:
Thank you 1 million for this HUGE effort and your constant responsiveness. I can't wait to use this! Cheers @squrious!

weaverryan:
Btw, this does still needs some docs. @squrious do you want to make a PR for that? It doesn't need to be perfect - just getting something is 95% of the important part.

smnandre:
I cannot say it better ! I have been really attentive to some details but this PR is really amazing and i’m so glad you worked as much to deliver. Congrats and thank you !

squrious:
Thanks to you @weaverryan @smnandre  for all the time you invested in the reviews! That was so cool to work on a such feature. I'll make the PR for the documentation ;)

src/LiveComponent/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 -   Fix instantiating LiveComponentMetadata multiple times.
 -   Change JavaScript package to `type: module`.
 -   Throwing an error when setting an invalid model name.
+-   Add support for URL binding in `LiveProp`
```
lyrixx, line 18: I think this should be updated. It looks like it has not been released in v2.13.3: https://github.com/symfony/ux/blob/v2.13.3/src/LiveComponent/CHANGELOG.md

PR #1232, opened 2023-10-25
[LiveComponent] Trigger model:set if a prop changes on the server

### LINKED ISSUE #1212: [LiveComponent] Execute model:set hook even for props changed on the server
Currently, the [model:set JS hook](https://symfony.com/bundles/ux-live-component/current/index.html#javascript-component-hooks) is only triggered if a model is changed by something on the client-side - e.g. a user types into an `input` or the user has a `<button data-model="foo" data-value="bar" data-action="live#update">`.

But, if the user triggers a `LiveAction`, and in that code, we change the value of a prop in PHP, the `model:set` hook is NOT called. This was done originally... only slightly on purpose: it was more code to add this feature, so I waited until it came up. Now it has, via a user on Slack.

I think it makes sense to trigger this whenever a prop changes for any reason. We DO have a spot where the props from the server are used to update the client-side props -  https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/src/Component/ValueStore.ts#L107 - it might require some sort of `diff` to detect any changes, then trigger the hook.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1212
| License       | MIT

Hi!

If you use the `model:set` hook, and a prop changes on the server during a re-render, when the Ajax call finishes, `model:set` will now be called. Details:

A) The code was a bit complex, but for writable sub-properties (`#[LiveProp(writable: 'firstName, 'lastName']` above a `user` property), `model:set` is called for each sub-key that changes - e.g. `user.firstName` and `user.lastName`. That is consistent with the behavior if those were changed by the user (e.g. by typing into a `user.firstName` input).

B) If the user changes a prop *during* the Ajax request (e.g. they type into `user.firstName`), then `model:set` will not be triggered for this one model. That's because the "frontend value" takes precedence over any server updates (this has always been the case). And so, the end result is that, even if the server changed `user.firstName`, after the Ajax call finishes, the actual value of that prop on the frontend will be whatever the user changed it to during the re-render.

Cheers!

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -257,15 +257,15 @@ final class LiveComponentHydrator
         return base64_encode(hash_hmac('sha256', json_encode($dehydratedPropsData), $this->secret, true));
     }
 
-    private function verifyChecksum(array $identifierPops, string $error = 'Invalid checksum sent when updating the live component.'): void
+    private function verifyChecksum(array $identifierProps, string $error = 'Invalid checksum sent when updating the live component.'): void
     {
-        if (!\array_key_exists(self::CHECKSUM_KEY, $identifierPops)) {
+        if (!\array_key_exists(self::CHECKSUM_KEY, $identifierProps)) {
             throw new HydrationException(sprintf('Missing %s. key', self::CHECKSUM_KEY));
         }
-        $sentChecksum = $identifierPops[self::CHECKSUM_KEY];
-        unset($identifierPops[self::CHECKSUM_KEY]);
+        $sentChecksum = $identifierProps[self::CHECKSUM_KEY];
+        unset($identifierProps[self::CHECKSUM_KEY]);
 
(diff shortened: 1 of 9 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -257,15 +257,15 @@ private function calculateChecksum(array $dehydratedPropsData): ?string
         return base64_encode(hash_hmac('sha256', json_encode($dehydratedPropsData), $this->secret, true));
     }
 
-    private function verifyChecksum(array $identifierPops, string $error = 'Invalid checksum sent when updating the live component.'): void
+    private function verifyChecksum(array $identifierProps, string $error = 'Invalid checksum sent when updating the live component.'): void
```
weaverryan, line 260: Unrelated - just identified this typo.

Kocal:

Thanks.

PR #1234, opened 2023-10-27
[Router] Add Symfony UX Router package

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Issues        | Fix #... 
| License       | MIT

As _promised_ in https://github.com/symfony/ux/pull/616#issuecomment-1480001911, it's time for a Symfony UX Router implementation. :tada:


smnandre:
Nice! 

Do you already have some thought about the security / splits questions or do you prefer we address those later in the process ?

Kocal:
The sooner the better I guess :p

kbond:
I think, at least by default, only routes flagged with an `expose: true` option should be dumped. This matches FOSJsRoutingBundle and makes migration easier.

tacman:
I _think_ FOSJsRoutingBundle may already work, by dumping not the JSON but a JS file.   

```bash
bin/console fos:js-routing:dump  --target=assets/js/fos_js_routes.js --pretty-print
```

Ryan had some comments here: https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/issues/466 that would automatically publish the routes (I think during the CompilerPass).  That'd be really cool, but as a start a simple working example of FOSJsRoutingBundle with importmap('app') and loading the routes via a module would be great. We could document that here, and perhaps we wouldn't need a UX Router package, at least not right away.

weaverryan:
Hi!

Thanks for getting this starts! But... since FOSJsRoutingBundle already exists and has for a long time, I would love to just add AssetMapper support for *that*. I think it would be fairly easy. Then the community doesn't need to migrate to a new package and we don't kill a community-supported package.

Kocal:
Hey, 

That's fair enough, I will just push my latest work and close the PR. We could re-open it later if needed :)

tacman:
@Kocal would you be able to work on adding the AssetMapCompiler that Ryan described to the FOS bundle?  I think that'd be a huge contribution and a nice solution to the problem addressed in this issue title.

I've added a comment to https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/issues/466 that shows a working version in a module, and Ryan added his idea to greatly simplify the DX experience.

tacman:
If you create your own version, though, I'm happy to test it.  I'm pretty much just looking for a Routing.generate() in javascript.

PR #1235, opened 2023-10-27
[Site] Remove highlight.php

### LINKED ISSUE #683: [Site] Change to use highlight.js directly
Hi!

We currently use https://github.com/scrivo/highlight.php for code highlighting. Unfortunately, this has fallen behind the real package - https://github.com/highlightjs/highlight.js - and so we're lacking things like proper highlighting for attributes (PHP 8 attributes are currently seen as comments and aren't even selectable by users!)

We should remove highlight.php and, instead, just call out to highlight.js via JavaScript for it to add the highlighting classes.

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        | Fix [#683](https://github.com/symfony/ux/issues/683)
| License       | MIT

Remove highlight.php and use highlight.js directly.

(highlight.php is in fact abandonned)

### DIFF
--- a/ux.symfony.com/assets/controllers/code-highlighter-controller.js
+++ b/ux.symfony.com/assets/controllers/code-highlighter-controller.js
@@ -1,0 +1,27 @@
+import { Controller } from '@hotwired/stimulus';
+import hljs from 'highlight.js/lib/core';
+import hljs_javascript from 'highlight.js/lib/languages/javascript';
+import hljs_php from 'highlight.js/lib/languages/php';
+import hljs_xml from 'highlight.js/lib/languages/xml';
+import hljs_twig from 'highlight.js/lib/languages/twig';
+
+hljs.registerLanguage('javascript', hljs_javascript);
+hljs.registerLanguage('php', hljs_php);
+hljs.registerLanguage('twig', hljs_twig);
+// xml is the language used for HTML
+hljs.registerLanguage('xml', hljs_xml);
+
+export default class extends Controller {
+    static targets = ['codeBlock'];
+
+    codeBlockTargetConnected() {
+        this.codeBlockTargets.forEach(this.#highlightCodeBlock)
+    }
+
+    #highlightCodeBlock(codeBlock) {
+        if (codeBlock.dataset.highlighted) {
+            return;
+        }
+        hljs.highlightElement(codeBlock);
+    }
+}
--- a/ux.symfony.com/src/Twig/CodeBlock.php
+++ b/ux.symfony.com/src/Twig/CodeBlock.php
@@ -13,7 +13,6 @@ namespace App\Twig;
 
 use App\Util\FilenameHelper;
 use App\Util\SourceCleaner;
-use Highlight\Highlighter;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 
@@ -40,30 +39,28 @@ class CodeBlock
     public bool $stripExcessHtml = false;
 
     public function __construct(
-        private Highlighter $highlighter,
         #[Autowire('%kernel.project_dir%')] private string $rootDir,
     ) {
     }
 
-    public function highlightSource(): string
+    /**
+     * Returns a list of source code pieces, extracted from the filename
+     * argument, ready to be renderer in the template.
+     *
+     * Every piece is composed as an array {content, highlight} with
+     *      content: the raw source code (after cleaning)
+     *      highlight: whether they must be syntax-highlighted or not
+     *
+     * @return list<array{content: string, highlight: ?bool}>
+     */
+    public function prepareSource(): array
     {
         $content = $this->getRawSource();
         if ('php' === $this->getLanguage()) {
             $content = SourceCleaner::cleanupPhpFile($content);
         }
 
-        $pieces = $this->splitAndProcessSource($content);
-
-        $highlighted = [];
-        foreach ($pieces as $piece) {
-            if ($piece['highlight']) {
-                $highlighted[] = $this->highlighter->highlight($this->getLanguage(), $piece['content'])->value;
-            } else {
-                $highlighted[] = $piece['content'];
-            }
-        }
-
-        return implode('', $highlighted);
+        return $this->splitAndProcessSource($content);
     }
 
     public function getRawSource(): string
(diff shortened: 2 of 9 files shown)


smnandre:
Just a first step to remove the PHP package easily... i keep in mind the @kbond [suggestion](https://github.com/symfony/ux/issues/683#issuecomment-1736079237) to use shiki

ux.symfony.com/assets/controllers/code-highlighter-controller.js
```diff
@@ -0,0 +1,24 @@
+import { Controller } from '@hotwired/stimulus';
+import hljs from 'highlight.js/lib/core';
+import hljs_javascript from 'highlight.js/lib/languages/javascript';
+import hljs_php from 'highlight.js/lib/languages/php';
+// import hljs_twig from 'highlight.js/lib/languages/twig';
```
weaverryan: This is commented-out on purpose?

ux.symfony.com/templates/components/CodeBlock.html.twig
```diff
@@ -1,23 +1,38 @@
+
+
```
weaverryan: Are these line breaks on purpose?

ux.symfony.com/src/Twig/CodeBlock.php
```diff
@@ -96,15 +93,17 @@ public function getGithubLink(): string
         return sprintf('https://github.com/symfony/ux/blob/2.x/ux.symfony.com/%s', $this->filename);
     }
 
-    private function getLanguage(): string
+    public function getLanguage(): string
```
weaverryan, line 96: Does this need to be public - I couldn't spot what is using it

weaverryan:
The issue I have with shiki is that it's a Node package that you run on the server. And I think you'd even need Node on production (not just locally) so it can be called at runtime. I'm not real excited by that.

smnandre:
> The issue I have with shiki is that it's a Node package that you run on the server. And I think you'd even need Node on production (not just locally) so it can be called at runtime. I'm not real excited by that.

I agree and still looking for a more integrated solution. In the meanwhile i think the JS version is OK and provide good results.

ux.symfony.com/assets/controllers/code-highlighter-controller.js
nvdbeek: Why add something as a comment if you don't need it?

ux.symfony.com/assets/controllers/code-highlighter-controller.js
```diff
@@ -0,0 +1,24 @@
...
+
+hljs.registerLanguage('javascript', hljs_javascript);
+hljs.registerLanguage('php', hljs_php);
+// hljs.registerLanguage('twig', hljs_twig);
```
nvdbeek: Same as above

ux.symfony.com/templates/components/CodeBlock.html.twig
nvdbeek: This blank lines necessary?

nvdbeek (changes requested):

ux.symfony.com/templates/components/CodeBlock.html.twig
smnandre: No it was not necessary :)

ux.symfony.com/assets/controllers/code-highlighter-controller.js
smnandre: Because i will need it soon..

.. but you're 100% right, as "soon" is not "now", and it could add confusion to anyone reading this code.

smnandre:
Rebased and ready

wouterj:
Mostly for my own interest (as we're using highlight.php in the Symfony docs as well, and I'm working integrating it in our new rst-parser): is there something truly wrong about highlight.php? We can override the PHP grammar with a customized grammar (we do this in the Symfony docs), and we can even try loading the latest highlight.js grammar. Or is there a BC break in the new grammars that causes highlight.php to no longer support them?

_Btw, I've been side-tracked in creating a PHP version of shiki - i.e. using TextMate grammars (used by VScode and Jetbrains) to highlight code using a PHP library. But I'm facing some challenges with supporting TextMate grammars. Maybe you'll see a new highlighter some day in my GitHub 😉_

smnandre:
> Mostly for my own interest (as we're using highlight.php in the Symfony docs as well, and I'm working integrating it in our new rst-parser): is there something truly wrong about highlight.php? We can override the PHP grammar with a customized grammar (we do this in the Symfony docs), and we can even try loading the latest highlight.js grammar. Or is there a BC break in the new grammars that causes highlight.php to no longer support them?
> 

For what it's worth, I'm not too fond of using abandoned packages / librairies... and it seems this is the case for this one (as the author explained [multiple](https://github.com/scrivo/highlight.php/issues/73) [times](https://github.com/scrivo/highlight.php/pull/109) he's working in another direction)

I don't know if those are "BC" or just easy fix, but currently on the ux.symfony.com pages (using hightlight php) the PHP attributes cannot be selected, for example.

Also, i'd really want to demo some things with CSS 4 stuff in the next weeks/months, and i'd like to avoid spending too much time checking everytime if this is working in highlight.php or if i need to find some time to fix it. But maybe i'm too pessimist and everything is ok.

So i genuinely don't know if we can easily "fix what need to be fixed" or if this is deeper than that. And i'd love to ear more about your experience with it, and your view on what/how we should use in the next monthes!

smnandre:
> Btw, I've been side-tracked in creating a PHP version of shiki - i.e. using TextMate grammars (used by VScode and Jetbrains) to highlight code using a PHP library. But I'm facing some challenges with supporting TextMate grammars. Maybe you'll see a new highlighter some day in my GitHub 😉

Shiki was the advice @kbond gave me [at first ](https://github.com/symfony/ux/issues/683#issuecomment-1735619160), but a full PHP version would be even better! 

I'd love to beta-test your solution when/if you want to share it !

smnandre:
Do you see something blocking here @weaverryan  ?

nvdbeek (approved):

ux.symfony.com/src/Twig/CodeBlock.php
smnandre: Time to understand again... in the template (and that was not done, thank you)

Before we gave this value to the Highlight PHP lib... now we need to pass it to the twig template and set a "language-XXX" class.

weaverryan:
Thanks Simon!

PR #1236, opened 2023-10-27
[TwigComponent] Add tests ComponentParser should pass

smnandre:
As the name resolution must be delayed from "parse-time" to "render-time", those tests should then pass.

As we know, they currently fail for
* HTML full tags: `<twig:XYZ></twig:XYZ>`
* Twig tag: `{% component XYZ %}`


smnandre:
Ready here @weaverryan ... and the tests are green thanks to you :)

weaverryan:
Thanks Simon!

PR #1237, opened 2023-10-27
[TwigComponent] Add tests ComponentRenderer should pass

smnandre:
As written in the documentation, it is possible to modify the template during the PreRenderEvent, so those tests should pass.

### DIFF
--- a/src/TwigComponent/tests/Fixtures/EventListener/FooBarComponentTemplateListener.php
+++ b/src/TwigComponent/tests/Fixtures/EventListener/FooBarComponentTemplateListener.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Tests\Fixtures\EventListener;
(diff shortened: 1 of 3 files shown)


src/TwigComponent/tests/Fixtures/EventListener/FooBarComponentTemplateListener.php
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Change the template of "FooBarBaz:" components during PrePrenderEvent.
+ * Before:
+ *      FooBar:Baz -> FooBar/Baz.html.twig
+ * After:
+ *      FooBar:Baz -> FooBar/Baz.foo_bar.html.twig
+ *
+ * @see PreRenderEvent
+ */
+class FooBarComponentTemplateListener
```
weaverryan, line 26: Is this registered in the test kernel and being used?

smnandre: It is registered (the test kernel has autowire+autoconfigure for the entire namespace) and it is used 

There are 4 data provided / tested in src/TwigComponent/tests/Integration/ComponentEventTest.php and only two fail because the FooBarComponentTemplateListener is called

smnandre:
Ready here @weaverryan ... and the tests are green thanks to you :)

weaverryan:
Thanks Simon!

Issue #1244, opened 2023-11-01
[Live] Using hash in the URL breaks updating the page 


1ed:
Hello,

if I navigate to a page with turbo, than I click on link calling a live action and redirect from the action to an URL containing a hash part (`#data`) the page does not update. When I refresh the page and click on the link again it works. Also works when the hash part is removed from the URL.

I have a reproducer here: https://github.com/1ed/live-turbo-refresh

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly ping? Should this still be open? I will close if I don't hear anything.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Hello? This issue is about to be closed if nobody replies.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly ping? Should this still be open? I will close if I don't hear anything.

PR #1245, opened 2023-11-01
[Live] Fix refresh when redirecting to the same page without Turbo

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        | Partially #1244
| License       | MIT


1ed:
Actually this scrolls up so it's useless.

PR #1246, opened 2023-11-02
[doc] adding ->createView() in normal symfony controller

bdujon:
Sorry if i did something wrong since it s my first PR there.

If you don't add ->createView() in the render you ll get an error Cannot assign Symfony\Component\Form\Form to property xxxComponent::$formView of type ?Symfony\Component\Form\FormView

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -1482,7 +1482,7 @@ this, create your controller like normal, including the submit logic::
 
             return $this->render('post/edit.html.twig', [
                 'post' => $post,
-                'form' => $form,
+                'form' => $form, // use $form->createView() in Symfony <6.2
             ]);
         }
     }


src/LiveComponent/doc/index.rst
```diff
@@ -1482,7 +1482,7 @@ this, create your controller like normal, including the submit logic::
 
             return $this->render('post/edit.html.twig', [
                 'post' => $post,
-                'form' => $form,
+                'form' => $form->createView(),
```
```suggestion
                'form' => $form, // use $form->createView() in Symfony <6.2
```

Not calling `createView()` in [Symfony 6.2+ is valid](https://symfony.com/blog/new-in-symfony-6-2-dx-improvements#simpler-form-rendering), but in <6.2, it is not.

kbond (changes requested):

src/LiveComponent/doc/index.rst
smnandre: Will LiveComponent support 5.4 when it is marked as non experimental ?

bdujon: > Not calling `createView()` in [Symfony 6.2+ is valid](https://symfony.com/blog/new-in-symfony-6-2-dx-improvements#simpler-form-rendering), but in <6.2, it is not.

Thanks I am indeed using symfony 5.4, i didn't know ->createView() is not needed anymore

kbond (approved):

weaverryan:
Thanks @bdujon!

PR #1249, opened 2023-11-04
[LiveComponent] Normalize "true" & "false" model values

### LINKED ISSUE #1211: [LiveComponent] Normalize true/false model values?
Via a question from Slack:

```html
<button
    data-model="isDeleting"
    data-value="false"
    data-action="live#update"
/>
```

This sends the string `false` for the field. I think we should be careful trying to "guess" types and coerce them. But perhaps we should for `true` and `false`? We could do this in JavaScript (ALWAYS change `true` to a boolean), or in PHP (i.e. look at the type of the property, and if it is `bool`, add some code to coerce `true`/`false`. Heck that might be pretty easy - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/LiveComponentHydrator.php#L448-L450

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes?
| New feature?  | no? 
| Issues        | Fix #1211
| License       | MIT

Coerce strings "true" & "false" to their boolean equivalent (same for 0 and 1)

```php
  private static function coerceStringToBoolean(string $value): bool
    {
        $booleanMap = [
            '0' => false,
            '1' => true,
            'false' => false,
            'true' => true,
        ];

        return $booleanMap[$value] ?? (bool) $value;
    }
```

Is that what you had in mind @weaverryan ?

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -244,9 +244,21 @@ final class LiveComponentHydrator
         return match ($type) {
             'int' => (int) $value,
             'float' => (float) $value,
-            'bool' => (bool) $value,
+            'bool' => self::coerceStringToBoolean($value),
             default => throw new \LogicException(sprintf('Cannot coerce value "%s" to type "%s"', $value, $type)),
         };
+    }
+
+    private static function coerceStringToBoolean(string $value): bool
+    {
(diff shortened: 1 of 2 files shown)


weaverryan (approved):
Yea, I like this 👍

weaverryan:
Thanks Simon for working on this feature, this is much appreciated.

PR #1251, opened 2023-11-06
Fix instantiating LiveComponentMetadata multiple times

TheDutchScorpion:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | N/A
| License       | MIT

This pull request fixes an issue with instantiating LiveComponentMetadata multiple times, when loading the same component multiple times.

### DIFF
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -25,6 +25,9 @@ use Symfony\UX\TwigComponent\ComponentFactory;
  */
 class LiveComponentMetadataFactory
 {
+    /** @var LiveComponentMetadata[] */
+    private array $liveComponentMetadata = [];
+
     public function __construct(
         private ComponentFactory $componentFactory,
         private PropertyTypeExtractorInterface $propertyTypeExtractor,
@@ -33,12 +36,16 @@ class LiveComponentMetadataFactory
 
     public function getMetadata(string $name): LiveComponentMetadata
     {
+        if (isset($this->liveComponentMetadata[$name])) {
+            return $this->liveComponentMetadata[$name];
+        }
+
         $componentMetadata = $this->componentFactory->metadataFor($name);
 
         $reflectionClass = new \ReflectionClass($componentMetadata->getClass());
         $livePropsMetadata = $this->createPropMetadatas($reflectionClass);
 
-        return new LiveComponentMetadata($componentMetadata, $livePropsMetadata);
+        return $this->liveComponentMetadata[$name] = new LiveComponentMetadata($componentMetadata, $livePropsMetadata);
     }
 
     /**


src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -33,12 +33,17 @@ public function __construct(
 
     public function getMetadata(string $name): LiveComponentMetadata
     {
+        static $liveComponentMetadata;
```
smnandre: Maybe a private typed property ?

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -33,12 +33,17 @@ public function __construct(
...
+            return $liveComponentMetadata[$name];
+        }
+
         $componentMetadata = $this->componentFactory->metadataFor($name);
 
         $reflectionClass = new \ReflectionClass($componentMetadata->getClass());
         $livePropsMetadata = $this->createPropMetadatas($reflectionClass);
 
-        return new LiveComponentMetadata($componentMetadata, $livePropsMetadata);
+        return $liveComponentMetadata[$name] = new LiveComponentMetadata($componentMetadata, $livePropsMetadata);
```
smnandre: Are we sure we want to share a single metadata instance for all the components of the same name ?

TheDutchScorpion: The ComponentFactory metadataFor method also returns ComponentMetadata with configurations based on name, so it won't be a problem.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
TheDutchScorpion: I could change it to

    /** @var LiveComponentMetadata[] */
    private static array $liveComponentMetadata = [];

    ...

    public function getMetadata(string $name): LiveComponentMetadata
    {
        if (isset(self::$liveComponentMetadata[$name])) {
            return self::$liveComponentMetadata[$name];
        }
        
        $componentMetadata = $this->componentFactory->metadataFor($name);

        $reflectionClass = new \ReflectionClass($componentMetadata->getClass());
        $livePropsMetadata = $this->createPropMetadatas($reflectionClass);

        return self::$liveComponentMetadata[$name] = new LiveComponentMetadata($componentMetadata, $livePropsMetadata);
    }

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
smnandre: But is the instance modified elsewhere in the code ?

kevinmade: The metadata DTO `LiveComponentMetadata` that will be stored only has private propeties and getters. All data within the DTO is static and related to property attribute `LiveProp` and static component configurations.

TheDutchScorpion: The metadata DTO `LiveComponentMetadata` that will be stored only has private propeties and getters. All data within the DTO is static and related to property attribute `LiveProp` and static component configurations.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
weaverryan: Yup - do this. But don't make the property static. In practice, there will only be one `LiveComponentMetadataFactory` service, so static vs non-static won't make any real difference. But we should not *enforce* that there can only be one instance of this object by using static.

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
weaverryan: Yup, this looks good to me too. `LiveComponentMetadata` is meant to be the metadata for the component with the name `$name`. If we ever modify any of that metadata on component-by-component basis, that should be considered a bug (and I don't see any of that at the moment).

weaverryan (commented):
Good PR 👍

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -25,6 +25,9 @@
  */
 class LiveComponentMetadataFactory
 {
+    /** @var LiveComponentMetadata[] */
+    private array $liveComponentMetadata = [];
+
```
weaverryan: This empty line has some extra whitespace (you can see it in the php-cs-fixer job)

weaverryan (approved):
just a php-cs thing - then good to go!

weaverryan:
Fantastic! Thanks @TheDutchScorpion!

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
TheDutchScorpion: This is what happens when using committing via github.com 😄

norkunas:
Maybe this should implement Symfony's ResetInterface in case of long-running webserver?

weaverryan:
I'm not sure. Are Doctrine or validation metadata, for example, reset in this way? Or, perhaps those use proper caching vs storing on a property. So, you might have a point

TheDutchScorpion:
Should I open a PR for this or do you make this change? @weaverryan

smnandre:
Thanks @TheDutchScorpion  👍

norkunas:
> I'm not sure. Are Doctrine or validation metadata, for example, reset in this way? Or, perhaps those use proper caching vs storing on a property. So, you might have a point

I mean that if we don't reset it could eat more memory than it should to without releasing the cache, so I think it's fine to have in memory cache for fetching metadata of same component many times, but sometimes it needs to be cleared up :)

PR #1258, opened 2023-11-08
[Doc] Fix preConnect/connect inversion in "disconnect()" doc examples

smnandre:
Following https://github.com/symfony/ux/pull/1227 i found other similar cases of callback inversion in the documentation.


weaverryan:
Thanks Simon!

PR #1263, opened 2023-11-09
[ChartJs] Replace `chart.js/auto` import with `Chart.register` calls

### LINKED ISSUE #1262: [Chart.js] Build fails with latest 2.13.0 release
Hi!
I tried to update `symfony/ux-chartjs` to latest 2.13.0 release but then the webpack failed to compile...

```shell
yarn run v1.22.11
$ xxxxxxxxxxxxxxxxx/node_modules/.bin/encore dev
Running webpack ...

 ERROR  Failed to compile with 1 errors

Module build failed: Module not found:
"./node_modules/@symfony/ux-chartjs/dist/controller.js" contains a reference to the file "chart.js/auto".
This file can not be found, please check it for typos or update it if the file got moved.
```

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1262
| License       | MIT

Ongoing attempt to fix ChartJS problems

### DIFF
--- a/src/Chartjs/assets/dist/controller.js
+++ b/src/Chartjs/assets/dist/controller.js
@@ -1,6 +1,9 @@
 import { Controller } from '@hotwired/stimulus';
-import Chart from 'chart.js/auto';
+import { registerables, Chart } from 'chart.js';
 
+if (registerables != undefined) {
+    Chart.register(...registerables);
+}
 let isChartInitialized = false;
 class default_1 extends Controller {
     constructor() {
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -3,6 +3,7 @@
     "description": "Chart.js integration for Symfony",
     "license": "MIT",
     "version": "1.1.0",
+    "type": "module",
     "main": "dist/controller.js",
     "types": "dist/controller.d.ts",
     "symfony": {
@@ -16,7 +17,7 @@
         },
         "importmap": {
             "@hotwired/stimulus": "^3.0.0",
-            "chart.js/auto": "^3.4.1 || ^4.0"
+            "chart.js": "^3.4.1 || ^4.0"
         }
     },
     "peerDependencies": {
@@ -25,7 +26,6 @@
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "@types/chart.js": "^2.9.34",
         "chart.js": "^3.4.1 || ^4.0",
         "resize-observer-polyfill": "^1.5.1",
         "vitest-canvas-mock": "^0.3.3"
--- a/src/Chartjs/assets/src/controller.ts
+++ b/src/Chartjs/assets/src/controller.ts
@@ -10,7 +10,12 @@
 'use strict';
 
 import { Controller } from '@hotwired/stimulus';
-import Chart from 'chart.js/auto';
+import { Chart, registerables } from 'chart.js';
+
+// ChartJs 3.x
+if (registerables != undefined) {
+    Chart.register(...registerables);
+}
 
 let isChartInitialized = false;
 
--- a/src/Chartjs/CHANGELOG.md
+++ b/src/Chartjs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.18.0
+
+-   Replace `chartjs/auto` import with explicit `Chart.register()` call #1263
+
 ## 2.17.0
 
 -   Add `chartjs:view-value-change` event #1605
--- a/yarn.lock
+++ b/yarn.lock
@@ -2127,7 +2127,7 @@
   resolved "https://registry.npmjs.org/@hotwired/stimulus/-/stimulus-3.2.1.tgz"
   integrity sha512-HGlzDcf9vv/EQrMJ5ZG6VWNs8Z/xMN+1o2OhV1gKiSG6CqZt5MCBB1gRg5ILiN3U0jEAxuDTNPRfBcnZBDmupQ==
 
-"@hotwired/turbo@^8.0":
+"@hotwired/turbo@^7.1.0 || ^8.0":
   version "8.0.4"
   resolved "https://registry.yarnpkg.com/@hotwired/turbo/-/turbo-8.0.4.tgz#5c5361c06a37cdf10dcba4223f1afd0ca1c75091"
   integrity sha512-mlZEFUZrJnpfj+g/XeCWWuokvQyN68WvM78JM+0jfSFc98wegm259vCbC1zSllcspRwbgXK31ibehCy5PA78/Q==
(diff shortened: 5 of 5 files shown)


src/Chartjs/assets/package.json
```diff
@@ -17,17 +17,17 @@
...
         "importmap": {
             "@hotwired/stimulus": "^3.0.0",
-            "chart.js/auto": "^3.4.1"
+            "chart.js": "^3.9"
         }
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1 <3.9"
+        "chart.js": "^3.9"
```
weaverryan: For others looking at this, we can absolutely allow v4 right now. But we should do it in a minor release, not a patch release.

weaverryan:
Background info: Chart.js changed some things starting in 3.9, which is why we previously restricted that version (they were arguably BC breaks).

In chart.js 2.13.0, we changed the JS package shipped to `type: module`. That had an unexpected side effect of no longer loading the `chart.js/auto` import (even though it loads fine in our test). Tbh, the reasons are a bit of a mystery to me.

Anyway, here is the breakdown of the fix:

A) For Encore users, you will need to upgrade `chart.js` to `^3.9`. You may currently have, for example, 3.8.0, so upgrading to 3.9.* is fine - just a minor update.
B) For AssetMapper users, you will need to `importmap:require 'chart.js'` (and you no longer need `chart.js/auto`). Flex may actually do this automatically - I think it will, but we'll need to try it :)

weaverryan (approved):

smnandre:
About the typescript error : 
Capture d’écran 2023-11-09 à 15 11 08


> I think 4.0 was release the other day and since Chart.js v3 it's been rewritten with Typescript, so you don't need these typedefinitions anymore.

https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/63641#discussioncomment-4419010


Just removed the file... let's see

weaverryan:
I'm checking into this failure. What's being imported in the tests is different than what's imported in Encore... which is why our original tests didn't catch the bug.

smnandre:
So we could release this fix, and check all this module/ES/node/2015/6/beta/tango/charlie later ? Or is there a risk ?

weaverryan:
Fixed in #1264. However, we can make this PR the proper PR for the NEXT version, which would include:

* Re-adding `type: module`
* Upgrading chart.js to v4

smnandre:
I did not realize Chart.js 4.0 was released 12 months ago. 

That could itself be a good argument to push gently users towards an upgrade :)

smnandre:
I rebased and tried to clean this PR i opened a long time ago, but i always end with test failures somewhere...

Could you guys @evertharmeling @simondaigre  maybe help me there ? 

(cc issue #1695 )

src/Chartjs/assets/dist/controller.js
```diff
@@ -1,6 +1,7 @@
 import { Controller } from '@hotwired/stimulus';
-import Chart from 'chart.js/auto';
+import { Chart, registerables } from 'chart.js';
 
+Chart.register(...registerables);
```
simondaigre:
```suggestion
registerables.forEach(registerable => Chart.register(registerable));
```
https://stackoverflow.com/questions/75449457/typeerror-chart-registerables-is-not-iterable-cannot-read-property-undefined

src/Chartjs/assets/src/controller.ts
```diff
@@ -10,7 +10,9 @@
 'use strict';
 
 import { Controller } from '@hotwired/stimulus';
-import Chart from 'chart.js/auto';
+import { Chart, registerables } from 'chart.js';
+
+Chart.register(...registerables);
```
evertharmeling:
```suggestion
// ChartJs 3.x
if (registerables != undefined) {
    Chart.register(...registerables);
}
```

It's kinda a too simple fix, but it makes the test-suite pass (for `3.x` and `4.x`). In the `3.x` version the `registerables` variable is `undefined` where as in `4.x` it consists of an list of 'registerables' :)

```js
[
  [Object: null prototype] {
    BarController: [class BarController extends DatasetController] {
      id: 'bar',
      defaults: [Object],
      overrides: [Object]
    },
    BubbleController: [class BubbleController extends DatasetController] {
      id: 'bubble',
      defaults: [Object],
      overrides: [Object]
    },
    DoughnutController: [class DoughnutController extends DatasetController] {
      id: 'doughnut',
      defaults: [Object],
      descriptors: [Object],
      overrides: [Object]
    }
    ...
```

I currently don't have a project where I use a plugin, so that should be tested/verified. Maybe we should add a simple test to the suite, where we test using a plugin?

An other solution I came across also seems to work, but then I'd prefer the previous solution:
```js
import { Chart } from 'chart.js';
import * as Chartjs from 'chart.js';

const controllers = Object.values(Chartjs).filter(
  (chart) => chart.id !== undefined
);

Chart.register(...controllers);
```

smnandre: That's a good lead thank you. We'll need to check if everything works the same ... (and i still don't understand how it can be undefined as it is exported in the 3.9.x source code https://cdn.jsdelivr.net/npm/chart.js@3.9.0/dist/chart.esm.js 

And i have the impression all the problem is indeed that the esm version is not the one used (maybe only in test)

smnandre: If you look at "chartjs/auto" https://cdn.jsdelivr.net/npm/chart.js@3.9.0/auto/auto.mjs

evertharmeling: Yeah that is pretty strange, it could specifically be testsuite related. I'll try to test it in a real project.

src/Chartjs/assets/dist/controller.js
evertharmeling: I came across this one too, however we still need to verfify `registerables != undefined` as with this code we would still get the following error on `3.x`:

```js
TypeError: Cannot read properties of undefined (reading 'forEach')
```

smnandre: And i still think it should not be "undefined" :))

evertharmeling: https://github.com/evertharmeling/ux-1263 

This seems to backup that it is our testsuite that's messing with us. I've added a console log on the `registerables` ([symfony/ux](https://github.com/evertharmeling/ux/commit/419f553fd985a8491ceb93ee5f5c09088a2c0dfb)). And when using `chart.js@3.9.1` the `registerables` variable is NOT `undefined`.

smnandre: Ok so its a « module » thing we need to force on the tests… that narrows the problem !

thanks a lot for taking this time it really helps a lot :)

simondaigre:
@smnandre Do you need some help to finish the PR ?

smnandre:
> @smnandre Do you need some help to finish the PR ?

That'd be really nice, yes :)

simondaigre:
I also checked and : 
- chart.js file (non-ESM) for 3.x versions doesn't export `registerables` : https://cdn.jsdelivr.net/npm/chart.js@3.9.0/dist/chart.js
- Vitest doesn't select the correct version (see https://github.com/valor-software/ng2-charts/issues/1421 + https://github.com/vitest-dev/vitest/discussions/4233 for more explanations)
- I tried to use `resolve.alias` in `vitest.config.js` but filenames are not the same between chart.js 3.x and 4.x.

So there is two solutions:

1. We stick chart.js to 3.x OR 4.x (BC break ?)
2. Solution provided by @evertharmeling works and as the issue is only related to our Vitest test suite, I think we can use it

smnandre:
@simondaigre thank you so much for taking this time, really 🙇 

Let's say we take the solution 2, is there any way to still check 3.x on the CI with no "false failures" ?

simondaigre:
I have no failures on CI with 2/ (for both 3.x and 4.x).

smnandre:
> I have no failures on CI with 2/ (for both 3.x and 4.x).

So you'd merge the PR ? 

(i updated to take the @evertharmeling suggestion)

simondaigre:
I checked on a real project with both chart.js 3.x and 4.x, it's OK for me: no breaks and it fixes my issue.
(changes to lockfiles seems unrelated to this PR I think).

smnandre:
Well let's merge it and test one last time on the dev branch before the next release ! :)  

(@kbond if you can ?)

A massive thank to you guys @simondaigre @evertharmeling !!  🙇

kbond:
Thanks all!

PR #1268, opened 2023-11-10
Remove "type": "module" from all package.json files

smnandre:
Since #1264, another person could not build its assets with Encore and a babel plugin.

So for now we decided to revert the "type: module" declaration.

### Revert the "type: module" 

Remove this "type" line from all package.json files and revert replacing "main:" with "module:"

Before
```json
   "type": "module",
   "module": "dist/foo.js",
```

After
```json
   "main": "dist/foo.js",
```

### Fix Svelte ESM bug with vite/vitest

```
We recommend converting your config to ESM by either:

    adding "type": "module" to the nearest package.json
    renaming vite.config.js/vite.config.ts to vite.config.mjs/vite.config.mts
```
(https://vitejs.dev/guide/troubleshooting.html#config) 

So i used the second one.... and it seems to work


weaverryan:
Thank you Simon!

weaverryan:
PR was merged... just didn't close

PR #1270, opened 2023-11-10
[LiveComponent][LiveComponentBundle] Add groups to components validation

lanfisis:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | no
| License       | MIT

I've had the possibility to use validation groups with `ValidatableComponentTrait` in Live Components.

Example:

```php
use App\Entity\User;
use Symfony\Component\Validator\Constraints as Assert;
use Symfony\UX\LiveComponent\ValidatableComponentTrait;

#[AsLiveComponent]
class EditUser
{

    #[LiveProp(writable: ['email', 'plainPassword'])]
    #[Assert\Valid(groups: ['my_group'])]
    public User $user;

     #[LiveProp]
     #[Assert\IsTrue(groups: ['my_group'])]
    public bool $agreeToTerms = false;
    
    // Use your own groups
    private function getValidationGroups(): array
    {
        return ['my_group'];
    }
}
```

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 2.13.2
 
 -   Revert "Change JavaScript package to `type: module`"
+-   Add capability to use validation groups in Live Components validation trait.
+
 
 ## 2.13.0
 
--- a/src/LiveComponent/src/ComponentValidator.php
+++ b/src/LiveComponent/src/ComponentValidator.php
@@ -12,6 +12,7 @@
 namespace Symfony\UX\LiveComponent;
 
 use Psr\Container\ContainerInterface;
+use Symfony\Component\Validator\Constraints\GroupSequence;
 use Symfony\Component\Validator\ConstraintViolation;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use Symfony\Contracts\Service\ServiceSubscriberInterface;
@@ -32,9 +33,9 @@ class ComponentValidator implements ComponentValidatorInterface, ServiceSubscrib
     /**
      * @return ConstraintViolation[][]
      */
-    public function validate(object $component): array
+    public function validate(object $component, string|GroupSequence|array $groups = null): array
     {
-        $errors = $this->getValidator()->validate($component);
+        $errors = $this->getValidator()->validate($component, groups: $groups);
 
         $validationErrors = [];
         foreach ($errors as $error) {
@@ -59,12 +60,12 @@ class ComponentValidator implements ComponentValidatorInterface, ServiceSubscrib
      *
      * @return ConstraintViolation[]
      */
-    public function validateField(object $component, string $propertyPath): array
+    public function validateField(object $component, string $propertyPath, string|GroupSequence|array $groups = null): array
     {
         $propertyParts = explode('.', $propertyPath);
         $propertyName = $propertyParts[0];
 
-        $errors = $this->getValidator()->validateProperty($component, $propertyName);
+        $errors = $this->getValidator()->validateProperty($component, $propertyName, groups: $groups);
 
         $errorsForPath = [];
         foreach ($errors as $error) {
--- a/src/LiveComponent/src/ComponentValidatorInterface.php
+++ b/src/LiveComponent/src/ComponentValidatorInterface.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\LiveComponent;
 
+use Symfony\Component\Validator\Constraints\GroupSequence;
 use Symfony\Component\Validator\ConstraintViolation;
 
 /**
@@ -32,12 +33,12 @@ interface ComponentValidatorInterface
      *
      * @return ConstraintViolation[][]
      */
-    public function validate(object $component): array;
+    public function validate(object $component, string|GroupSequence|array $groups = null): array;
 
     /**
      * Returns an array of violations for this one specific property.
      *
      * @return ConstraintViolation[]
      */
-    public function validateField(object $component, string $propertyName): array;
+    public function validateField(object $component, string $propertyName, string|GroupSequence|array $groups = null): array;
 }
--- a/src/LiveComponent/src/ValidatableComponentTrait.php
+++ b/src/LiveComponent/src/ValidatableComponentTrait.php
@@ -12,6 +12,7 @@
 namespace Symfony\UX\LiveComponent;
 
 use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
+use Symfony\Component\Validator\Constraints\GroupSequence;
 use Symfony\Component\Validator\ConstraintViolation;
 use Symfony\Contracts\Service\Attribute\Required;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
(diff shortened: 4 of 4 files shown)


smnandre:
I may have missunderstood the behaviour.. do you have an other example ?

It seems to me that, in the one you provided, the method is superflous ad its content and the attributes are both statically written in the class code.

lanfisis:
Hi @smnandre , thank you for your comment. I want to give you more informations.

### Context

I'm currently working on a [Sylius](https://sylius.com/fr/) based project. I try to replace JQuery scripts in the checkout by Live Components. For example, when a customer add his email in the checkout address form, if Sylius detects an account with this email, a password field and a login button appear. It currently based and a JQuery custom plugin and an ajax request that I try to switch with a Live Component.

### My problem

In Sylius, all entities validations are based on at least a `sylius` group.  For example, this is [the Customer constraints file](https://github.com/Sylius/Sylius/blob/1.13/src/Sylius/Bundle/CustomerBundle/Resources/config/validation/CustomerGroup.xml). As you can see, email attribute constraints are also grouped with multiple groups:
```xml
        <property name="email">
            <constraint name="NotBlank">
                <option name="message">sylius.customer.email.not_blank</option>
                <option name="groups">
                    <value>sylius</value>
                    <value>sylius_customer_guest</value>
                    <value>sylius_customer_checkout_guest</value>
                </option>
            </constraint>
            <constraint name="Length">
                <option name="max">254</option>
                <option name="maxMessage">sylius.customer.email.max</option>
                <option name="groups">
                    <value>sylius</value>
                    <value>sylius_customer_guest</value>
                    <value>sylius_customer_checkout_guest</value>
                </option>
            </constraint>
            <constraint name="Email">
                <option name="message">sylius.customer.email.invalid</option>
                <option name="mode">strict</option>
                <option name="groups">
                    <value>sylius</value>
                    <value>sylius_customer_guest</value>
                    <value>sylius_customer_checkout_guest</value>
                </option>
            </constraint>
        </property>
```

I've create a LiveComponent called `EmailLogin` with a `customer` live props write like this:
```php
<?php

declare(strict_types=1);

namespace App\Components\Shop;

use App\Entity\Customer\Customer;
use App\Entity\Customer\CustomerInterface;
use Sylius\Component\Core\Repository\CustomerRepositoryInterface;
use Sylius\Component\Resource\Factory\FactoryInterface;
use Symfony\UX\LiveComponent\ValidatableComponentTrait;
use Symfony\Component\Validator\Constraints as Assert;


#[AsLiveComponent(name: 'Shop:EmailLogin')]
final class EmailLogin extends AbstractController
{
    use DefaultActionTrait;

    #[LiveProp(writable: [LiveProp::IDENTITY, 'email'])]
    #[Assert\Valid]
    public ?Customer $customer = null;

    #[LiveProp(writable: true)]
    public string $password = '';

    #[LiveProp(writable: true)]
    public bool $hasLogin = false;

    public function __construct(
        private readonly CustomerRepositoryInterface $customerRepository,
        private readonly FactoryInterface $customerFactory,
    ) {
    }

    public function mount(?CustomerInterface $customer = null): void
    {
        $this->customer = $customer ?? $this->customerFactory->createNew();
    }

    public function __invoke()
    {
        $this->validate();
    }
}
```

and the associated component:
```twig
<div {{ attributes }} class="relative">
    <twig:Loader/>

        <div class="mb-6">
        <div class="flex justify-between">
            <label class="block text-sm font-medium leading-6 text-gray-900 required"
                   for="email_login_email">{{ 'sylius.ui.email'|trans }}</label>
        </div>
        <div class="mt-2">
            <input data-model="on(change)|customer.email" type="text" id="email_login_email"
                   name="sylius_address[firstName]" required="required"
                   class="block w-full rounded-md border-0 py-1.5 ring-1 ring-inset focus:ring-2 focus:ring-inset sm:text-sm sm:leading-6">
        </div>
        {% if _errors.has('customer.email') %}
            <span class="mt-2 text-sm text-red-600">
                {{ _errors.get('customer.email') }}
            </span>
        {%- endif -%}
    </div>

    {% if hasLogin %}
        {# ... #}
    {% endif %}

</div>
```

I was disapointed because any error appears when I'm tipping a wrong formatted email. After some search, I remember myself this `sylius` group thing.

### My solution

My goal was to add the capability, from my Live Component, to give groups to the Symfony validator. So I've climb up the call stack and add the `groups` parameter to the `ComponentValidatorInterface`  methods. Those methods are called in the `ValidatableComponentTrait` so I also add an overwritable private method `getValidationGroups()` to the trait. I could also use and abstract trait method but it would be a BC and groups are not alway used on validation and it would more be a constraint for users.

 If you don't overwrite this method in your component who using the `ValidatableComponentTrait` trait, the `$this->getValidator()->validate($this, $this->getValidationGroups())` line code use `null` as default groups. If you overwrite it in  your component, `$this->getValidator()->validate($this, $this->getValidationGroups())` will use the groups return by your overwrite. 

### How it works

Great I hope 😛. After added this PR changes in my project `vendor` folder, I just overwrite the `getValidationGroups()` like this:
```php

    // ...

    #[LiveProp(writable: [LiveProp::IDENTITY, 'email'])]
    #[Assert\Valid(groups: ['sylius'])]
    public ?Customer $customer = null;

     // ...

    private function getValidationGroups(): array
    {
        return ['sylius'];
    }
```
and it works like a charm:
image

### Conclusion

I hope I have respond to your interrogation with this shitload of comments (sorry). If not, feel free to asks me other questions or add more comments.

Thanks !

lanfisis:
Hi @smnandre ,

After few discussion with my colleagues I've another proposal that would be easier to understand and to use: instead of have to overwrite  `getValidationGroups()` method I had a new optional `$groups` parameter to the validate function. So my previous example became:

```php
    // ...

    #[LiveProp(writable: [LiveProp::IDENTITY, 'email'])]
    #[Assert\Valid(groups: ['sylius'])]
    public ?Customer $customer = null;

     // ...

    public function __invoke()
    {
        $this->validate(groups: ['sylius']);
    }
```

Is it feel better to you ?

Thanks

WebMamba (commented):
Hey thanks @lanfisis! This changed looks completly legit to me! Maybe you can add a test in the ValidatableComponentTraitTest?

lanfisis:
> WebMamba



> Hey thanks @lanfisis! This changed looks completly legit to me! Maybe you can add a test in the ValidatableComponentTraitTest?

Hi @WebMamba , thanks for your reply. I'm not a big test expert but I will make a try.

Thanks again

smnandre:
Hi @lanfisis, thank for this really detailed explanation! 

I better understand the real-life need your PR would adress and i like it :) 

I'm still wondering the impacts on the rest of the validation usage. I mean, how would this work with
* [partial validation](https://symfony.com/bundles/ux-live-component/current/index.html#validation-without-a-form) (is it a field validated, a group, both ? then is it statically defined of for every call... in the later, we'd change a lot of code to ensure the component is valid)
* [validation reset](https://symfony.com/bundles/ux-live-component/current/index.html#resetting-validation-errors) (same thing, should it reset all fields, only those of the groups...)

Does it work with form [validation groups](https://symfony.com/doc/current/form/validation_groups.html) too ?

But i insist, i like the idea very much :)

weaverryan (commented):
This looks good to me. Can you update the docs? I think that one thing to explain is that, if you put a constraint into a group other than `Default`, it will not automatically be validated via the hook in the trait - https://github.com/symfony/ux/blob/8fe391d96f1173de03e05f8d2007830476be4881/src/LiveComponent/src/ValidatableComponentTrait.php#L130-L144

So, if your constraint is in a group, you would only be able to trigger that validation via a `LiveAction`, I believe. Or perhaps we also need a way to dynamically change the validation groups that `ValidatableComponentTrait` will use? For example, the trait has a `getValidationGroups()` method that returns `null` and you could override in your class to return whatever you want? This is your use-case, so let that guide you. If this is not something you need, that's ok too. I'm just trying to think of HOW this is used.

Kocal:

PR #1277, opened 2023-11-14
[Docs] Add caution alert when bundle require StimulusBundle

ker0x:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | To complement  #1276 
| License       | MIT

This PR add a `caution` alert at the beginning of `Installation` section for all bundle requiring StimulusBundle

### DIFF
--- a/src/Autocomplete/doc/index.rst
+++ b/src/Autocomplete/doc/index.rst
@@ -12,9 +12,11 @@ into an Ajax-powered autocomplete smart UI control (leveraging `Tom Select`_):
 Installation
 ------------
 
-Before you start, make sure you have `StimulusBundle configured in your app`_.
+.. caution::
 
-Then install the bundle using Composer and Symfony Flex:
+    Before you start, make sure you have `StimulusBundle configured in your app`_.
+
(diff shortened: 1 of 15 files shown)


smnandre:
Imho, the "caution" may be a bit _too strong_ (red block)... maybe a "info" should do the job ?

I made a quick check (i was not sure) so, for information, those packages require `symfony/stimulus-bundle` in their composer file, so the risk of installing them without Stimulus bundle is null

* Chartjs
* Notify
* React
* Svelte
* Turbo
* Vue

(not saying that does not deserve a message in the doc)


And those ones do not 

* Autocomplete
* CropperJS
* Dropzone
* LazyImage
* LiveComponent
* Swup
* TogglePassword
* Translator
* Typed

Finally (to cover all the packages), TwigComponent does not depend / use / require StimulusBundle.

weaverryan:
This can definitely be a gotcha. We could require stimulus-bundle in the rest of the packages. It is not technically required, but it is *practically* required.

weaverryan:
Thanks Romain!

PR #1282, opened 2023-11-16
append backslash to namespace in twig_component.yaml

boedah:
This is the way the [flex recipe](https://github.com/symfony/recipes/blob/flex/main/symfony.ux-twig-component.2.13.json) creates this file and it throws an error without the final backslash.


smnandre (approved):

weaverryan:
Thanks Peter!

PR #1288, opened 2023-11-20
[TwigComponent] Merge data-action in ComponentAttributes

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | None that I know of
| License       | MIT

If you pass `data-action` into a component & the root `attributes` has a `data-action` already thanks to `attributes.defaults`, they should be merged together. Same behavior as `data-controller`. I found this while adding a `data-action` to a component and suddenly losing my existing `data-action` from inside.

Cheers!

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -80,7 +80,7 @@ final class ComponentAttributes
         }
 
         foreach ($this->attributes as $key => $value) {
-            if (\in_array($key, ['class', 'data-controller'], true) && isset($attributes[$key])) {
+            if (\in_array($key, ['class', 'data-controller', 'data-action'], true) && isset($attributes[$key])) {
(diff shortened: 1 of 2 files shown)


norkunas:
What if I want to override the `data-action` totally, will it be possible? :)

weaverryan:
> What if I want to override the data-action totally, will it be possible? :)

Nope :). It's a valid question. Right now, `data-controller` and `class` are already "merge". This is hardcoded in. I've talked with @kbond a bit about adding some sort of API to replace vs merge. That's something we need to add. In the mean time, this is the "closer to proper" behavior.

So, we can't have it both ways right now. Currently, it's not possible to merge a `data-action` from the inside with a `data-action` from the outside. However, at the very least, you could introduce some other "prop" flag and use that to turn on / off an internal `data-action` after this PR. Since `data-controller` merges, this is closer to the correct behavior.

Issue #1241, opened 2023-10-30
[LiveCollection] Live collection breaks autocomplete input


luchidalgo:
When using LiveCollectionTrait breaks autocomplete when item in collection is removed.

The main form :

```
// PlaceForm.php

        $builder->add(
            'availabilities',
            LiveCollectionType::class,
            [
                'label' => 'Disponibilités',
                'entry_type' => AvailabilityForm::class,
                'allow_add' => true,
                'allow_delete' => true,
                'entry_options' => [
                    'label' => false,
                ],
                'by_reference' => false,
            ]
        );
```

The embeded form in collection :

```
//  AvailabilityForm.php

        $builder->add('endDate', DateType::class, [
            'label' => 'Date de fin',
            'widget' => 'single_text',
            'required' => false,
        ]);

        $builder->add(
            'weekDays',
            ChoiceType::class,
            [
                'label' => 'Jours de la semaine',
                'choices' => [
                    'Lundi' => 1,
                    'Mardi' => 2,
                    'Mercredi' => 3,
                    'Jeudi' => 4,
                    'Vendredi' => 5,
                    'Samedi' => 6,
                    'Dimanche' => 7
                ],
                'multiple' => true,
                'autocomplete' => true,
            ]
        );
```

The Twig component : 

```
// PlaceFormComponent.php

#[AsLiveComponent()]
class PlaceFormComponent extends AbstractController
{
    use DefaultActionTrait;
    use LiveCollectionTrait;

    #[LiveProp]
    public ?Place $place;
    
    protected function instantiateForm(): FormInterface
    {
        return $this->createForm(
            PlaceForm::class,
            $this->place
        );
    }
}
```

The template

```
        {{ component('PlaceFormComponent', {
            form: form,
            place: place
        }) }}
```

The rendered form : 

```
<input type="time" id="place_form_availabilities_0_endTime" name="place_form[availabilities][0][endTime]" required="required" class="form-control" value="18:00">

<select id="place_form_availabilities_0_weekDays" name="place_form[availabilities][0][weekDays][]" required="required" 
(... BLA BLA BLA SYMFONY UX STUFF AND AUTOCOMPLETE STUFF AND TOM SELECT STUFF ... )

<button type="button" id="place_form_availabilities_0_delete" name="place_form[availabilities][0][delete]" data-action="live#action" data-action-name="removeCollectionItem(name=place_form[availabilities], index=0)" class="btn btn-danger btn-sm btn">Delete entry</button>

[...]

Issue #909, opened 2023-05-30
[LiveComponent] v2.9 Autocomplete field not refreshed


bastien70:
I am currently on version 2.9 for Live-component, Twig-component, and Autocomplete.
We updated from 2.7, and we ran `yarn install --force` and `yarn run dev`. We also updated Autocomplete and Live-component recipes.

In version 2.7, when we had an AutocompleteType field (from Symfony UX), and an Ajax request goes, the Autocomplete field kept its "TomSelect" well.

Oddly since version 2.9, this one loses its "TomSelect". However, when we retrieve the TomSelect instance of the Autocomplete field, it seems well initialized and has the right parameters.

https://github.com/symfony/ux/assets/53140475/c5e21a21-57c9-41e2-afc2-5b5a3a527307

Before first Ajax request :
![1](https://github.com/symfony/ux/assets/53140475/76ec4b83-d335-451b-bcbe-036f9869422c)

After first Ajax request : 
![2](https://github.com/symfony/ux/assets/53140475/228973b1-39ad-4ee0-a06a-d4a74335dd60)

And TomSelect attributes at this moment : 

![3](https://github.com/symfony/ux/assets/53140475/b0100031-779a-4094-beff-35e7f58a5f31)

In our form, we have : 

```php
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder
            ->add('course', ListingSchoolClassCourseAutocompleteField::class, [
                'label' => 'Course',
                'required' => true,
            ])
        ;
        
        // ...
}
```

And this is our ListingSchoolClassCourseAutocompleteField : 

```php
#[AsEntityAutocompleteField(alias: 'listing_school_class_course', route: 'ux_entity_autocomplete')]
class ListingSchoolClassCourseAutocompleteField extends AbstractType
{
    public function __construct(
        private readonly Security $security,
        private readonly SessionService $sessionService,
        private readonly ScopeService $scopeService
    ) {
    }

    public function configureOptions(OptionsResolver $resolver)
    {
        /** @var User $user */
        $user = $this->security->getUser();
        $collegeYear = $this->sessionService->getCollegeYear();

        $resolver->setDefaults([
            'class' => SchoolClassCourse::class,
            'placeholder' => '',
            'choice_label' => 'createShortLabel',
            'no_more_results_text' => false,
            'query_builder' => function () use ($user, $collegeYear) {
[...]

PR #1290, opened 2023-11-20
[Autocomplete] Attempting a simpler "reset" of items

### LINKED ISSUE #1261: [LiveComponet] [Autocomplete] Bug when combining live with autocomplete select field
Hi there,

I'don't find the reason but in some cases when a field has the 'autocomplete' => true option in a live component the real value selected in the form is not the value sent when submitting.

I've created a repository to reproduce the bug:
 [https://github.com/tdumalin/bug-ux-live-plus-autocomplete](https://github.com/tdumalin/bug-ux-live-plus-autocomplete)
All the installation informations are setted in the README file.

I've notice the bug multiple times but always on select input, noticed on:
ChoiceType
EnumType
EntityType

Adding a random data-live-id on the row_attr seems to resolve the bug but it's not convenient to set globally
```twig
{{ form_row(form.type, {row_attr: {
        'data-live-id': random()
}}) }}
```
If needed here is my work around:

```php
<?php

namespace App\Form\Extension;

use Symfony\Bridge\Doctrine\Form\Type\EntityType;
use Symfony\Component\Form\AbstractTypeExtension;
use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
use Symfony\Component\Form\Extension\Core\Type\EnumType;
use Symfony\Component\OptionsResolver\Options;

class AddAutocompleteExtension extends AbstractTypeExtension
{

    {
        $resolver->setDefaults([
            'autocomplete' => true,
        ]);


        $resolver->setNormalizer('row_attr', function (Options $options, $value) {
            if ($options['autocomplete']) {
                $value['data-live-id'] = uniqid('autocomplete-');
            }

            return $value;
        });
    }

    public static function getExtendedTypes(): iterable
    {
        return [
            ChoiceType::class,
            EntityType::class,
            EnumType::class,
        ];
    }
}

```

NB: This work around making you lost the focus of the input

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Possibly #1290 (need to check and remove data-live-ignore), fixes #1261 possibly #1241 possibly #909
| License       | MIT

Tomselect is a real pain in the butt. When you select an option, it re-orders the `option` elements and uses the elements themselves internally to remember which one is selected. It plays poorly with LiveComponents, which is why we added the MutationObserver logic. However, at least in one straightforward case that I had today, when you selected a new item, the new value was being lost. This will only get worse when Turbo 8 adds morphing.

This PR attempts to simplify: when needed, completely destroy TomSelect and recreate it. It seems to work well, but 2 tests are failing. And I'm pulling my hair out - so it may not even be fixable. No matter what I try, when I reinitialize TomSelect on a set of HTML that has been reordered, it gets confused. It's as if it has a static cache somewhere that I can't find...

### DIFF
--- a/src/Autocomplete/assets/dist/controller.d.ts
+++ b/src/Autocomplete/assets/dist/controller.d.ts
@@ -33,8 +33,10 @@ export default class extends Controller {
     private mutationObserver;
     private isObserving;
     private hasLoadedChoicesPreviously;
+    private originalOptions;
     initialize(): void;
     connect(): void;
+    initializeTomSelect(): void;
     disconnect(): void;
     private getMaxOptions;
     get selectElement(): HTMLSelectElement | null;
@@ -43,9 +45,9 @@ export default class extends Controller {
     get preload(): string | boolean;
     private resetTomSelect;
     private changeTomSelectDisabledState;
-    private updateTomSelectPlaceholder;
     private startMutationObserver;
     private stopMutationObserver;
     private onMutations;
-    private requiresLiveIgnore;
+    private createOptionsDataStructure;
+    private areOptionsEquivalent;
 }
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -29,26 +29,25 @@ class default_1 extends Controller {
         _default_1_instances.add(this);
         this.isObserving = false;
         this.hasLoadedChoicesPreviously = false;
+        this.originalOptions = [];
     }
     initialize() {
-        if (this.requiresLiveIgnore()) {
-            this.element.setAttribute('data-live-ignore', '');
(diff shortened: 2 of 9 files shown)


weaverryan:
Related: https://github.com/orchidjs/tom-select/issues/666

I simply can't figure out how to make TomSelect keep the selected option value.

weaverryan:
SO happy to say that this is ready 😅.

tl;dr TomSelect rearranges the `<option>` elements. If something "morphes" the `<option>` element like LiveComponents (or Turbo 8) does, TomSelect can't handle this. This PR removes morphing for the`<option>` elements for Autocomplete and fixes & improves how we reinitialize TomSelect if the option elements DO change.

PR #1293, opened 2023-11-22
Simplify code: use variable instead of method call

bocharsky-bw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | no
| License       | MIT

We already set that on a var, so we can just reuse that var for simplicity


weaverryan:
Thanks Victor!

PR #1299, opened 2023-11-28
[Site] Fix highlight.js version

smnandre:
We downgraded by accident the Highlight.js version

(made the translation page bug)


weaverryan:
Thanks for the fast fix Simon!

PR #1308, opened 2023-11-29
check not to access a character longer than the string

### LINKED ISSUE #1306: lint:twig error, possibly with twig component
I've opened an issue at https://github.com/symfony/symfony/issues/52778, but posting here because I think it may have to do with symfony/ux-twig-component.

I wonder if I should actually post this to twigphp/twig?

I'm pretty sure it's a bug that was recently introduced, as I think the template referenced used to work.  @kbond , can you take a quick peek and at least tell me where to post this issue?

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1306 
| License       | MIT

Makes sure that $this->input[$this->position] won't fail if position is greater than the input length.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -263,6 +263,10 @@ class TwigPreLexer
      */
     private function consume(string $string): bool
     {
+        if (strlen($this->input) >= $this->position) {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


smnandre:
I tried to understand "why / how" that particular case happened and.. there is maybe a deeper fix to do. 

I'll update you as soon as possible.

smnandre:
Fixed in #1312

PR #1311, opened 2023-11-29
[Site] Make package box titles visible in dark mode

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | desc below
| License       | MIT

Right now package box titles are really dark in the dark mode:
Zrzut ekranu 2023-11-29 o 16 54 15

With that change it will look like:
Zrzut ekranu 2023-11-29 o 16 54 08

### DIFF
--- a/ux.symfony.com/assets/styles/components/_PackageBox.scss
+++ b/ux.symfony.com/assets/styles/components/_PackageBox.scss
@@ -120,6 +120,10 @@
   letter-spacing: -1px;
   color: #0A0A0A;
   text-decoration: none;
+  
+  [data-bs-theme="dark"] & {
+    filter: invert(1);
+  }
 }
 
 .PackageBox_link {


ux.symfony.com/assets/styles/components/_PackageBox.scss
```diff
@@ -120,6 +120,10 @@
   letter-spacing: -1px;
   color: #0A0A0A;
   text-decoration: none;
+  
+  [data-bs-theme="dark"] & {
+    filter: invert(1);
+  }
 }
```
smnandre, line 127:
```
color: var(--bs-body-color);
text-decoration: none;
```

Simpler no ?

stloyd: "Depends" ;) If you mean to change it on line 121, then it will be changed to a light theme.

I followed already used code from: https://github.com/symfony/ux/blob/2.x/ux.symfony.com/assets/styles/sections/_header.scss#L39

smnandre: Because i wanted to invert all the colors, but if you look on other CSS properties, i tried to use CSS vars as much as possible because it's easier to read/maintain.. 

But i won't go against your suggestion :)) Iit's a good catch and a really nice attention to open this PR, and i'm really glad you did :)

weaverryan:
Thank you Joseph!

PR #1312, opened 2023-11-29
[TwigComponent] Fix TwigPreLexer edge case

### LINKED ISSUE #1306: lint:twig error, possibly with twig component

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1306
| License       | MIT

Quick fix to handle commented blocks in sublexer 

@weaverryan  @WebMamba  it may need a second look as i find weird the changes made do not have any other impacts...

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -263,10 +263,6 @@ class TwigPreLexer
      */
     private function consume(string $string): bool
     {
-        if ($string[0] !== $this->input[$this->position]) {
-            return false;
-        }
-
         $stringLength = \strlen($string);
         if (substr($this->input, $this->position, $stringLength) === $string) {
             $this->position += $stringLength;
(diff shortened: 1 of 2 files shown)


WebMamba:
Hum, I think if we have a problem with comments we also have an issue with the verbatim tag? And I think an easier solution could be if the pointer consumes `{#` then moves the pointer position to the next `#}` as we do at the top. With this method, you can remove all your condition

smnandre:
I just wanted to fix the specific issue there.. and i guess a lot more people uses block in comments (in a dev phase it is a frequent thing) than verbatim :)  

But it's an issue indeed. 

I do believe we can fix _and improve_ some things here... but i'd suggest to do it a bit differently and that seems to me another (bigger) discussion. 

WDYT ? 
We can open an issue if you want to discuss this next move ?

smnandre:
> Hum, I think if we have a problem with comments we also have an issue with the verbatim tag? And I think an easier solution could be if the pointer consumes `{#` then moves the pointer position to the next `#}` as we do at the top. With this method, you can remove all your condition

Sadly not if we want to handle properly verbatim + comments + strings +  ....

weaverryan:
This moves things a small step forward - thanks Simon!

PR #1314, opened 2023-11-29
[TwigComponent] Allow Bundle configuration with PHP ConfigBuilder

### LINKED ISSUE #1304: [TwigComponent] Configuration with PHP ConfigBuilders fails
I use configuration with PHP ConfigBuilders and tried to add configuration for symfony/ux. That failed, ``TwigComponentConfig`` could not be loaded.

So I was digging around (not knowing anything about it), comparing how others do it and finally found a working solution:

I created

``ux/src/TwigComponent/src/DependencyInjection/Configuration.php``

with content like this

```
class Configuration implements ConfigurationInterface
{
}
```

and then moved the [``getConfigTreeBuilder()``](https://github.com/symfony/ux/blob/1f326f06e36a52c2df42d176e388f1378019d484/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php#L132-L190) method from

``ux/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php``

into that class.

I don’t know if this the correct way, if that would break anything else, but at least now ``TwigComponentConfig`` gets built and configuration works.

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| Issues        | Fix #1304
| License       | MIT

Allow TwigComponent configuration via configbuilder.

### DIFF
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -53,7 +53,8 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
             throw new LogicException('The TwigBundle is not registered in your application. Try running "composer require symfony/twig-bundle".');
         }
 
-        $config = $this->processConfiguration($this, $configs);
+        $configuration = $this->getConfiguration($configs, $container);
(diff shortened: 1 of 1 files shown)


kbond (approved):

kbond:
Thanks Simon.

PR #1319, opened 2023-11-30
[StimulusBundle] Remove stimulus.asset_mapper.loader_javascript_compiler when no asset-mapper

### LINKED ISSUE #1313: [StimulusBundle] Asset mapper dependency
Hi there, 

I had a strange issue running composer update, and having an upgrade for stimulus-bundle from 2.12 to 2.13.2

```
Upgrading symfony/stimulus-bundle (v2.12.0 => v2.13.2)
```

After upgrading, on the clear:cache, had the following issue : 

image

After looked at the stacktrace, I inspected the compiler pass, and it seems to browse all definitions to check each one is a command, but nevermind. 

I don't have _yet_ installed asset-mapper (but I plan to :wink:)

The `StimulusLoaderJavaScriptCompiler`, which implement an interface from asset-mapper, is now loaded in a service, whereas in 2.12 it wasn't, probably thanks to the following lines : 

image

So I'm wondering if now, the asset-mapper should be a direct dependency of stimulus-bundle, or should we find a workaround to not load this service when the asset-mapper component is not installed. 

I tried adding 
```php
if (interface_exists(AssetMapperInterface::class)) {
        $container->services()
            ->set('stimulus.asset_mapper.loader_javascript_compiler', StimulusLoaderJavaScriptCompiler::class)
            ->args([
                service('stimulus.asset_mapper.controllers_map_generator'),
                param('kernel.debug'),
            ])
            ->tag('asset_mapper.compiler', ['priority' => 100])
        ;
    }
```
in `config/services.php` and it seems working - _for my case_, and I can provide a fix PR. 

WDYT ?

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1313 
| License       | MIT

The service name was incorrect here. 

All credits to @smnandre who found this gem

### DIFF
--- a/src/StimulusBundle/src/DependencyInjection/Compiler/RemoveAssetMapperServicesCompiler.php
+++ b/src/StimulusBundle/src/DependencyInjection/Compiler/RemoveAssetMapperServicesCompiler.php
@@ -26,7 +26,7 @@ class RemoveAssetMapperServicesCompiler implements CompilerPassInterface
         if (!$container->hasDefinition('asset_mapper')) {
             $container->removeDefinition('stimulus.ux_controllers_twig_runtime');
             $container->removeDefinition('stimulus.asset_mapper.controllers_map_generator');
-            $container->removeDefinition('stimulus.asset_mapper.stimulus_loader_javascript_compiler');
+            $container->removeDefinition('stimulus.asset_mapper.loader_javascript_compiler');
         }
     }
 }


smnandre:
All credits to you for signaling the bug and doing the PR :)) 

@weaverryan should we send a bugfix release soon for this one, as it will happen to every user using StimulusBundle without AssetMapper installed no ?

weaverryan:
Thank you for this - good find/fix & PR Jibé!

Issue #1320, opened 2023-12-01
[TwigComponent] Deprecations not setting defaults and anonymous template directory


rarila:
With v2.13.0 you get a deprecation warning for not setting for not setting `twig_component.defaults` and `twig_component.anonymous_template_directory`.

I’m not familiar with the internals or future plans but at least for now it’s no problem to ignore those two settings, if you don’t have any anonymous components and explicitly name your components. So being forced to set them creates a bit of boilerplate.

With setting `twig_component.anonymous_template_directory` (and later being it `components` by default) there’s even another little problem. Until now I had a folder `/templates/components` where my named components live. Setting the anonymous template directory to `components` makes at least the `console debug:twig-component` command fail:

>Unknown component "foobar.twig". And no matching anonymous component template was found.

I don’t know, if that’s just a minor problem of that command or if that can also be problematic in normal use, though. Anyway, I had to move my components to an `ux` subfolder and add a useless empty `anonymous` folder.

Anyway, declaring _I have no anonymous components_ by not setting a directory would make sense for me and perhaps would even save a few cycles searching for them. Probably the same with default naming.

What do you think?

PS: I wonder if the configuration `defaults` should get a better, more self explanatory name like `anonymous_template_directory`, where you know what it does by just looking at it.

smnandre:
Those config values should be set via the flex recipe so it should not be too much troubles for most. 

There is something with the command i’ll check that in a couple of hours.

smnandre:
I dont get your PS: how would you rename the option ?

rarila:
@smnandre Something like `namespace_directory_mapping`.

smnandre:
The configuration file copied by Flex should be [this one](https://github.com/symfony/recipes/blob/main/symfony/ux-twig-component/2.13/config/packages/twig_component.yaml). 

```yaml
twig_component:
    anonymous_template_directory: 'components/'
    defaults:
        # Namespace & directory for components
        App\Twig\Components\: 'components/'
 ```
 
 I find it pretty good but there is always room for improvement, so feel free to open an issue/PR to suggest changes :)

**Comment by smnandre** [comment]
> t least the `console debug:twig-component` command fail:
> 
> > Unknown component "foobar.twig". And no matching anonymous component template was found.

What are the arguments given to the command ? 

This message should only happen if you type something like 

```
php bin/console debug:twig-component foobar.twig
```

But maybe there is something not working as expected there

rarila:
@smnandre 

[...]

PR #1321, opened 2023-12-03
[TwigComponent] Fix `debug:twig-component` command (anonymous vs non-anonymous components)

rarila:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| License       | MIT

The `debug:twig-component` command had problems when anonymous and non-anonymous components were in the same directory.

- Templates are now also taken into account when they have just a `.twig` extension instead of a `.html.twig` extension.
- Templates belonging to non-anonymous components now are not additionally listed as  anonymous component.

_Parts of #1320 includes some discussion about this bug._

### DIFF
--- a/src/TwigComponent/src/ComponentTemplateFinder.php
+++ b/src/TwigComponent/src/ComponentTemplateFinder.php
@@ -42,28 +42,27 @@ final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
 
         // Legacy auto-naming rules < 2.13
         if (null === $this->directory) {
-            if ($loader->exists('components/'.$componentPath.'.html.twig')) {
-                return 'components/'.$componentPath.'.html.twig';
-            }
-
-            if ($loader->exists($componentPath.'.html.twig')) {
-                return $componentPath.'.html.twig';
-            }
-
-            if ($loader->exists('components/'.$componentPath)) {
-                return 'components/'.$componentPath;
-            }
-
-            if ($loader->exists($componentPath)) {
-                return $componentPath;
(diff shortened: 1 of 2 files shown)


src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -42,28 +42,27 @@ public function findAnonymousComponentTemplate(string $name): ?string
...
-
-            if ($loader->exists('components/'.$componentPath)) {
-                return 'components/'.$componentPath;
-            }
-
-            if ($loader->exists($componentPath)) {
-                return $componentPath;
+            foreach ([
+                'components/'.$componentPath.'.html.twig',
+                'components/'.$componentPath.'.twig',
```
smnandre, line 47: That would be a BC break in a BC layer, we cannot do that :/

src/TwigComponent/src/ComponentTemplateFinder.php
```diff
@@ -42,28 +42,27 @@ public function findAnonymousComponentTemplate(string $name): ?string
...
+                }
             }
 
             return null;
         }
 
-        $template = rtrim($this->directory, '/').'/'.$componentPath.'.html.twig';
-        if ($loader->exists($template)) {
-            return $template;
+        foreach (['.html.twig', '.twig'] as $extension) {
```
smnandre, line 61: I dont think we want to accept anonymous Component with multiple extensions. The idea was to have a 1:1 relation betwee  an anonymous template and its name..

Kocal:

PR #1322, opened 2023-12-03
[Autocomplete] Allow passing extra options to the autocomplete fields

jakubtobiasz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

This PR still needs some tweaks and tests, but I want to validate my idea. Currently, if we pass anything as a third argument in the `->add()` method, it is only used on the form render, and is fully ignored during the AJAX call.
CleanShot 2023-12-03 at 12 36 59

While implementing UX's Autocomplete in Sylius, I wanted to complete the following user story
```
Given I want to edit a taxon
When I want to assign a parent taxon to it
And I check a list of available taxa
Then I should see all taxa except the edited one
```

So basically, I have to pass a **current** taxon's ID to the autocomplete's query. As we know, currently it's not possible. So after contacting @weaverryan, I decided to implement a mechanism similar to the one from Live Components.

When you pass an array of options as a `3rd` argument, you can use a special `extra_options` key, which is an array consisting `scalars`/`arrays`/`nulls`.

Next, when the form is rendered, I get these values, calculate a checksum for them, then pass them through `json_encode` and `base64_encode` functions. In the end, I glue them to the `url` values in the `\Symfony\UX\Autocomplete\Form\AutocompleteChoiceTypeExtension::finishView` method.

So, basically with the following configuration:
CleanShot 2023-12-03 at 12 48 51

we end up with the following HTML code:
CleanShot 2023-12-03 at 12 49 36

I decided to "glue" the `extra_options` to the URL, as I didn't have to deal with JS. Of course, I do not exclude a chance to refactor it, as a whole method should be refactored anyway.

Finally, the controller decodes the data, checks the checksum and passes the values to the autocomplete's form.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -11,6 +11,8 @@
 -   Added `tom-select/dist/css/tom-select.bootstrap4.css` to `autoimport` - this
     will cause this to appear in your `controllers.json` file by default, but disabled
     see.
+-   Allow passing `extra_options` key in an array passed as a `3rd` argument of the `->add()` method.
+    It will be used during the Ajax call to fetch results.
 
 ## 2.13.2
 
--- a/src/Autocomplete/src/Checksum/ChecksumCalculator.php
+++ b/src/Autocomplete/src/Checksum/ChecksumCalculator.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Autocomplete\Checksum;
+
+/** @internal */
+class ChecksumCalculator
+{
+    public function __construct(private readonly string $secret)
+    {
+    }
+
+    public function calculateForArray(array $data): string
+    {
+        $this->sortKeysRecursively($data);
+
+        return base64_encode(hash_hmac('sha256', json_encode($data), $this->secret, true));
+    }
+
+    private function sortKeysRecursively(array &$data): void
+    {
+        foreach ($data as &$value) {
+            if (\is_array($value)) {
+                $this->sortKeysRecursively($value);
+            }
+        }
+        ksort($data);
+    }
+}
--- a/src/Autocomplete/src/Controller/EntityAutocompleteController.php
+++ b/src/Autocomplete/src/Controller/EntityAutocompleteController.php
@@ -14,20 +14,27 @@ namespace Symfony\UX\Autocomplete\Controller;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\Autocomplete\AutocompleteResultsExecutor;
 use Symfony\UX\Autocomplete\AutocompleterRegistry;
+use Symfony\UX\Autocomplete\Checksum\ChecksumCalculator;
+use Symfony\UX\Autocomplete\Form\AutocompleteChoiceTypeExtension;
+use Symfony\UX\Autocomplete\OptionsAwareEntityAutocompleterInterface;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
  */
 final class EntityAutocompleteController
 {
+    public const EXTRA_OPTIONS = 'extra_options';
+
     public function __construct(
         private AutocompleterRegistry $autocompleteFieldRegistry,
         private AutocompleteResultsExecutor $autocompleteResultsExecutor,
         private UrlGeneratorInterface $urlGenerator,
+        private ChecksumCalculator $checksumCalculator,
     ) {
     }
 
@@ -36,6 +43,11 @@ final class EntityAutocompleteController
         $autocompleter = $this->autocompleteFieldRegistry->getAutocompleter($alias);
         if (!$autocompleter) {
             throw new NotFoundHttpException(sprintf('No autocompleter found for "%s". Available autocompleters are: (%s)', $alias, implode(', ', $this->autocompleteFieldRegistry->getAutocompleterNames())));
+        }
+
+        if ($autocompleter instanceof OptionsAwareEntityAutocompleterInterface) {
+            $extraOptions = $this->getExtraOptions($request);
+            $autocompleter->setOptions([self::EXTRA_OPTIONS => $extraOptions]);
         }
 
         $page = $request->query->getInt('page', 1);
@@ -53,5 +65,49 @@ final class EntityAutocompleteController
             'results' => ($data->optgroups) ? ['options' => $data->results, 'optgroups' => $data->optgroups] : $data->results,
             'next_page' => $nextPage,
         ]);
+    }
+
+    /**
+     * @return array<string, scalar|array|null>
+     */
+    private function getExtraOptions(Request $request): array
+    {
+        if (!$request->query->has(self::EXTRA_OPTIONS)) {
+            return [];
+        }
+
+        $extraOptions = $this->getDecodedExtraOptions($request->query->getString(self::EXTRA_OPTIONS));
+
+        if (!\array_key_exists(AutocompleteChoiceTypeExtension::CHECKSUM_KEY, $extraOptions)) {
+            throw new BadRequestHttpException('The extra options are missing the checksum.');
+        }
+
+        $this->validateChecksum($extraOptions[AutocompleteChoiceTypeExtension::CHECKSUM_KEY], $extraOptions);
+
+        return $extraOptions;
+    }
+
+    /**
+     * @return array<string, scalar>
+     */
+    private function getDecodedExtraOptions(string $extraOptions): array
+    {
+        return json_decode(base64_decode($extraOptions), true, flags: \JSON_THROW_ON_ERROR);
+    }
+
+    /**
+     * @param array<string, scalar> $extraOptions
+     */
+    private function validateChecksum(string $checksum, array $extraOptions): void
+    {
+        $extraOptionsWithoutChecksum = array_filter(
+            $extraOptions,
+            fn (string $key) => AutocompleteChoiceTypeExtension::CHECKSUM_KEY !== $key,
+            \ARRAY_FILTER_USE_KEY,
+        );
+
+        if ($checksum !== $this->checksumCalculator->calculateForArray($extraOptionsWithoutChecksum)) {
+            throw new BadRequestHttpException('The extra options have been tampered with.');
+        }
     }
 }
--- a/src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
+++ b/src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
@@ -21,6 +21,7 @@ use Symfony\Component\Form\Form;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\UX\Autocomplete\AutocompleteResultsExecutor;
 use Symfony\UX\Autocomplete\AutocompleterRegistry;
+use Symfony\UX\Autocomplete\Checksum\ChecksumCalculator;
 use Symfony\UX\Autocomplete\Controller\EntityAutocompleteController;
 use Symfony\UX\Autocomplete\Doctrine\DoctrineRegistryWrapper;
 use Symfony\UX\Autocomplete\Doctrine\EntityMetadataFactory;
@@ -116,6 +117,7 @@ final class AutocompleteExtension extends Extension implements PrependExtensionI
                 new Reference('ux.autocomplete.autocompleter_registry'),
                 new Reference('ux.autocomplete.results_executor'),
                 new Reference('router'),
+                new Reference('ux.autocomplete.checksum_calculator'),
             ])
             ->addTag('controller.service_arguments')
         ;
@@ -126,6 +128,13 @@ final class AutocompleteExtension extends Extension implements PrependExtensionI
                 new Reference('maker.doctrine_helper', ContainerInterface::IGNORE_ON_INVALID_REFERENCE),
             ])
             ->addTag('maker.command')
+        ;
+
+        $container
+            ->register('ux.autocomplete.checksum_calculator', ChecksumCalculator::class)
+            ->setArguments([
+                '%kernel.secret%',
+            ])
         ;
     }
 
@@ -149,6 +158,7 @@ final class AutocompleteExtension extends Extension implements PrependExtensionI
         $container
             ->register('ux.autocomplete.choice_type_extension', AutocompleteChoiceTypeExtension::class)
             ->setArguments([
+                new Reference('ux.autocomplete.checksum_calculator'),
                 new Reference('translator', ContainerInterface::IGNORE_ON_INVALID_REFERENCE),
             ])
             ->addTag('form.type_extension');
--- a/src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
+++ b/src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
@@ -19,6 +19,7 @@ use Symfony\Component\Form\FormView;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Contracts\Translation\TranslatorInterface;
+use Symfony\UX\Autocomplete\Checksum\ChecksumCalculator;
 
 /**
  * Initializes the autocomplete Stimulus controller for any fields with the "autocomplete" option.
@@ -27,8 +28,12 @@ use Symfony\Contracts\Translation\TranslatorInterface;
  */
 final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
 {
-    public function __construct(private ?TranslatorInterface $translator = null)
-    {
+    public const CHECKSUM_KEY = '@checksum';
+
+    public function __construct(
+        private readonly ChecksumCalculator $checksumCalculator,
+        private readonly ?TranslatorInterface $translator = null,
+    ) {
     }
 
     public static function getExtendedTypes(): iterable
@@ -79,6 +84,10 @@ final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
             $values['min-characters'] = $options['min_characters'];
         }
 
+        if ($options['extra_options']) {
+            $values['url'] = $this->getUrlWithExtraOptions($values['url'], $options['extra_options']);
+        }
+
         $values['loading-more-text'] = $this->trans($options['loading_more_text']);
         $values['no-results-found-text'] = $this->trans($options['no_results_found_text']);
         $values['no-more-results-text'] = $this->trans($options['no_more_results_text']);
@@ -90,6 +99,41 @@ final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
 
         $view->vars['uses_autocomplete'] = true;
         $view->vars['attr'] = $attr;
+    }
+
+    private function getUrlWithExtraOptions(string $url, array $extraOptions): string
+    {
+        $this->validateExtraOptions($extraOptions);
+
+        $extraOptions[self::CHECKSUM_KEY] = $this->checksumCalculator->calculateForArray($extraOptions);
+        $extraOptions = base64_encode(json_encode($extraOptions));
+
+        return sprintf(
+            '%s%s%s',
+            $url,
+            $this->hasUrlParameters($url) ? '&' : '?',
+            http_build_query(['extra_options' => $extraOptions]),
+        );
+    }
+
+    private function hasUrlParameters(string $url): bool
+    {
+        $parsedUrl = parse_url($url);
+
+        return isset($parsedUrl['query']);
+    }
+
+    private function validateExtraOptions(array $extraOptions): void
+    {
+        foreach ($extraOptions as $optionKey => $option) {
+            if (!\is_scalar($option) && !\is_array($option) && null !== $option) {
+                throw new \InvalidArgumentException(sprintf('Extra option with key "%s" must be a scalar value, an array or null. Got "%s".', $optionKey, get_debug_type($option)));
+            }
+
+            if (\is_array($option)) {
+                $this->validateExtraOptions($option);
+            }
+        }
     }
 
     public function configureOptions(OptionsResolver $resolver): void
@@ -106,6 +150,7 @@ final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
             'min_characters' => null,
             'max_results' => 10,
             'preload' => 'focus',
+            'extra_options' => [],
         ]);
 
         // if autocomplete_url is passed, then HTML options are already supported
--- a/src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
+++ b/src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
@@ -22,17 +22,18 @@ use Symfony\Component\PropertyAccess\PropertyPathInterface;
 use Symfony\UX\Autocomplete\Doctrine\EntityMetadata;
 use Symfony\UX\Autocomplete\Doctrine\EntityMetadataFactory;
 use Symfony\UX\Autocomplete\Doctrine\EntitySearchUtil;
-use Symfony\UX\Autocomplete\EntityAutocompleterInterface;
+use Symfony\UX\Autocomplete\OptionsAwareEntityAutocompleterInterface;
 
 /**
  * An entity auto-completer that wraps a form type to get its information.
  *
  * @internal
  */
-final class WrappedEntityTypeAutocompleter implements EntityAutocompleterInterface
+final class WrappedEntityTypeAutocompleter implements OptionsAwareEntityAutocompleterInterface
 {
     private ?FormInterface $form = null;
     private ?EntityMetadata $entityMetadata = null;
+    private array $options = [];
 
     public function __construct(
         private string $formType,
@@ -139,7 +140,7 @@ final class WrappedEntityTypeAutocompleter implements EntityAutocompleterInterfa
     private function getForm(): FormInterface
     {
         if (null === $this->form) {
-            $this->form = $this->formFactory->create($this->formType);
+            $this->form = $this->formFactory->create($this->formType, options: $this->options);
         }
 
         return $this->form;
@@ -167,5 +168,14 @@ final class WrappedEntityTypeAutocompleter implements EntityAutocompleterInterfa
         }
 
         return $this->entityMetadata;
+    }
+
+    public function setOptions(array $options): void
+    {
+        if (null !== $this->form) {
+            throw new \LogicException('The options can only be set before the form is created.');
+        }
+
+        $this->options = $options;
     }
 }
--- a/src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
+++ b/src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Autocomplete;
+
+/**
+ * Interface for classes that will have an "autocomplete" endpoint exposed with a possibility to pass additional form options.
+ */
+interface OptionsAwareEntityAutocompleterInterface extends EntityAutocompleterInterface
+{
+    public function setOptions(array $options): void;
+}
--- a/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
+++ b/src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
@@ -31,7 +31,7 @@ class AutocompleteFormRenderingTest extends KernelTestCase
             ->throwExceptions()
             ->get('/test-form')
             ->assertElementAttributeContains('#product_category', 'data-controller', 'custom-autocomplete symfony--ux-autocomplete--autocomplete')
-            ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type')
+            ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type?extra_options=')
             ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-min-characters-value', '2')
             ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-max-results-value', '25')
 
@@ -122,5 +122,59 @@ class AutocompleteFormRenderingTest extends KernelTestCase
             ])
             ->assertElementCount('#product_ingredients option', 0)
         ;
+    }
+
+    public function testItUsesPassedExtraOptions()
+    {
+        $ingredient1 = IngredientFactory::createOne(['name' => 'Flour']);
+        $ingredient2 = IngredientFactory::createOne(['name' => 'Sugar']);
+        $ingredient3 = IngredientFactory::createOne(['name' => 'Modified Flour']);
+
+        $this->browser()
+            ->throwExceptions()
+            ->get('/test-form')
+            ->assertElementCount('#product_ingredients option', 0)
+            ->assertNotContains('Flour')
+            ->assertNotContains('Sugar')
+            ->assertNotContains('Modified Flour')
+            // request all three ingredients
+            ->post('/test-form', [
+                'body' => [
+                    'product' => [
+                        'ingredients' => [
+                            (string) $ingredient1->getId(),
+                            (string) $ingredient2->getId(),
+                            (string) $ingredient3->getId(),
+                        ],
+                    ],
+                ],
+            ])
+            // assert that "Modified Flour" is not included
+            ->assertElementCount('#product_ingredients option', 2)
+            ->assertContains('Flour')
+            ->assertContains('Sugar')
+            ->assertNotContains('Modified Flour')
+        ;
+    }
+
+    public function testItReturnsErrorWhenSendingMalformedExtraOptions(): void
+    {
+        $extraOptionsWithoutChecksum = $this->encodeData(['foo' => 'bar']);
+        $extraOptionsWithInvalidChecksum = $this->encodeData(['foo' => 'bar', '@checksum' => 'invalid']);
+        $extraOptionsWithValidChecksum = $this->encodeData(['foo' => 'bar', '@checksum' => 'O2nYjcGr/l8GmUuYUSfE52hoyEL0NtDhBzUbn17KVHQ=']);
+
+        $this->browser()
+            ->post(sprintf('/test/autocomplete/category_autocomplete_type?extra_options=%s', $extraOptionsWithoutChecksum))
+            ->assertStatus(400)
+            ->post(sprintf('/test/autocomplete/category_autocomplete_type?extra_options=%s', $extraOptionsWithInvalidChecksum))
+            ->assertStatus(400)
+            ->post(sprintf('/test/autocomplete/category_autocomplete_type?extra_options=%s', $extraOptionsWithValidChecksum))
+            ->assertStatus(200)
+        ;
+    }
+
+    private function encodeData(array $data): string
+    {
+        return base64_encode(json_encode($data));
     }
 }
--- a/src/Autocomplete/tests/Unit/Calculator/ChecksumCalculatorTest.php
+++ b/src/Autocomplete/tests/Unit/Calculator/ChecksumCalculatorTest.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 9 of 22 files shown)


src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
```diff
@@ -149,6 +158,7 @@ private function registerFormServices(ContainerBuilder $container): void
         $container
             ->register('ux.autocomplete.choice_type_extension', AutocompleteChoiceTypeExtension::class)
             ->setArguments([
+                new Reference('ux.autocomplete.checksum_calculator'),
```
smnandre, line 161: Maybe second  argument to avoid a BC problem ?

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -49,6 +49,8 @@ public function getValue(object $entity): mixed;
      */
     public function isGranted(Security $security): bool;
 
+    public function withOptions(array $options): self;
```
smnandre: Any precision about $options type ? string => scalar ?/ mixed  ?

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -91,6 +107,19 @@ public function finishView(FormView $view, FormInterface $form, array $options):
         $view->vars['attr'] = $attr;
     }
 
+    private function validateExtraOptions(array $extraOptions): void
+    {
+        foreach ($extraOptions as $option) {
+            if (!\is_scalar($option) && !\is_array($option) && null !== $option) {
+                throw new \InvalidArgumentException(sprintf('Extra option "%s" must be a scalar value, an array or null.', $option));
```
smnandre:
```
, received "%s", $option, get_debug_type($option)) ? 
```

src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -18,16 +18,20 @@
...
  */
 final class EntityAutocompleteController
 {
+    public const EXTRA_OPTIONS = 'extra_options';
+
     public function __construct(
         private AutocompleterRegistry $autocompleteFieldRegistry,
         private AutocompleteResultsExecutor $autocompleteResultsExecutor,
         private UrlGeneratorInterface $urlGenerator,
+        private ChecksumCalculatorInterface $checksumCalculator,
```
smnandre: Do we 100% need an interface here ?

src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -38,6 +42,11 @@ public function __invoke(string $alias, Request $request): Response
             throw new NotFoundHttpException(sprintf('No autocompleter found for "%s". Available autocompleters are: (%s)', $alias, implode(', ', $this->autocompleteFieldRegistry->getAutocompleterNames())));
         }
 
+        $extraOptions = $this->getExtraOptions($request);
+        $autocompleter = $autocompleter->withOptions([
```
smnandre: I have this vision where "withOptions" means immutable and a new object.. If this is not the goal here, could we use a more usual "setOptions" or "addOptions" ?

src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
jakubtobiasz: In such a case, we will have to make it an optional argument, as translator is nullable and is null by default.

src/Autocomplete/src/Controller/EntityAutocompleteController.php
jakubtobiasz: It's final and non-internal, so IMO yes. We might consider making it internal, but at the same time, someone might want to change how the checksum is calculated 🤔.

src/Autocomplete/src/Controller/EntityAutocompleteController.php
jakubtobiasz: `setOptions` sounds cool 🤔. Yeah, `with` might create some confusion here.

src/Autocomplete/src/DependencyInjection/AutocompleteExtension.php
smnandre: Yep my bad, sorry

alexander-schranz:
As I'm currently facing a similar issue in recreating some more complex Selection Type which is doing similar things like autocomplete. I wanted to share it here also here. https://gist.github.com/alexander-schranz/b1cb8843bc54dc368001375fe6998e61#file-mycontactchoicetype-php-L39 In my case I did also go with providing a list of options, so the loading part will have possibility todo such things. But instead of duplicating the options I just provide a list of which options should be forwarded (string[], instead of array<string, mixed>`. So I don't have to duplicate the value of the option inside the extra_options, item_loader_options in my case.

alexander-schranz:
I try currently some magic to find out automatically my `extra_options` without have the need to define them in my custom Selection Type wanted to share it here, not sure if its too much magic:

```php
    public function buildView(FormView $view, FormInterface $form, array $options): void
    {
        $extraOptions = [];
        $formType = $form->getConfig()->getType();

        $innerType = $formType->getInnerType();
        $defaultOptionsResolver = new OptionsResolver();
        $innerType->configureOptions($defaultOptionsResolver);
        $definedOptions = $defaultOptionsResolver->getDefinedOptions();
        foreach ($defaultOptionsResolver->getMissingOptions() as $missingOption) {
            $defaultOptionsResolver->remove($missingOption);
        }
        // Form, ChoiceType options which we keep always in mind and are important for form widget rendering e.g. required, disabled, ...
        foreach ($options['extra_options'] as $extra_option) {
            $definedOptions[] = $extra_option;
        }

        $defaultOptions = $defaultOptionsResolver->resolve([]);
        foreach ($options as $key => $value) {
            if (!\is_scalar($value)) {
                // ignore all options which are non-scalar values
                continue;
            }

            if (!\in_array($key, $definedOptions)) {
                // we ignore all options which were not defined specifically for the inner type
                continue;
            }

            if (isset($defaultOptions[$key]) && $defaultOptions[$key] === $value) {
                // we ignore all options which have the same value as the default value
                continue;
            }

            $extraOptions[$key] = $value;
        }

        $view->vars['extra_options'] = $extraOptions;
    }
    {
        $resolver->setDefaults([
            'extra_options' => ['required', 'disabled'],
        ]);
    }
```

-----------

What could also be used to avoid a seperate extra_options is the `$resolver->getRequiredOptions();` used as `extra_options`, and so via `$resolver->setRequired('some_value');` can be defined what is required:


```php
    public function buildView(FormView $view, FormInterface $form, array $options): void
    {
        $extraOptions = [];
        $formType = $form->getConfig()->getType();

        $innerType = $formType->getInnerType();
        $defaultOptionsResolver = new OptionsResolver();
        $requiredOptions = $defaultOptionsResolver->getRequiredOptions();
        foreach ($defaultOptionsResolver->getMissingOptions() as $missingOption) {
            $defaultOptionsResolver->remove($missingOption);
        }

        $defaultOptions = $defaultOptionsResolver->resolve([]);
        foreach ($options as $key => $value) {
            if (!\is_scalar($value)) {
                // ignore all options which are non-scalar values
                continue;
            }

            if (!\in_array($key, $requiredOptions)) {
                // we ignore all options which are not required
                continue;
            }

            if (isset($defaultOptions[$key]) && $defaultOptions[$key] === $value) {
                // we ignore all options which have the same value as the default value
                continue;
            }

            $extraOptions[$key] = $value;
        }

        $view->vars['extra_options'] = $extraOptions;
    }
```

src/Autocomplete/src/Calculator/ChecksumCalculator.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\UX\Autocomplete\Calculator;
+
+final class ChecksumCalculator implements ChecksumCalculatorInterface
```
alexander-schranz: In my own selection type I did instead of using a checksum go with the `UriSigner` but ignore all query parameters which I don't know, so maybe we could do here the avoid own checksum calculator:

```php
if (!$this->uriSigner->check(
    // ignore unrelated query parameters as they can be search filters which are allowed
    $request->getSchemeAndHttpHost() . $request->getPathInfo() . '?_config=' . \rawurlencode($request->query->get('_config')) . '&_hash=' . \rawurlencode($request->query->get('_hash')),
)) {
    throw new AccessDeniedHttpException('Invalid signature.');
}

/**
 * @var array{
 *     options: array<string, mixed>,
 *     name: string,
 *     theme: string,
 * }|false
 */
$config = \json_decode(\base64_decode($request->query->get('_config'), true), true, 512, \JSON_THROW_ON_ERROR);
```

src/Autocomplete/src/Calculator/ChecksumCalculator.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\UX\Autocomplete\Calculator;
```
smnandre: Minor detail, but should we not have a Checksum namespace ?

smnandre:
Do you want some help/feedback/other on this ?

jakubtobiasz:
Hi @smnandre,
I've refactored one of two places I marked as "needs to be refactored". I have to write some functional tests for it, and... basically it will be ready to be fully reviewed.

I'll let you know once ready to be reviewed :D. It should be ready today :D.

smnandre:
Great! Good luck then :)

jakubtobiasz:
I guess I haven't missed anything, and it's ready to be reviewed 👍🏻. AFAIR LiveComponent build failures are not related to my PR.

src/Autocomplete/src/Checksum/ChecksumCalculator.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function __construct(private readonly string $secret)
+    {
+    }
+
+    public function calculateForArray(array $data): string
+    {
+        $this->sortKeysRecursively($data);
+
+        return base64_encode(hash_hmac('sha256', json_encode($data), $this->secret, true));
```
stloyd, line 27: Maybe use some lighter hash algo? 
```suggestion
        return base64_encode(hash_hmac('xxh128', json_encode($data), $this->secret, true));
```
Ref: https://php.watch/articles/php-hash-benchmark

src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -38,6 +45,11 @@ public function __invoke(string $alias, Request $request): Response
             throw new NotFoundHttpException(sprintf('No autocompleter found for "%s". Available autocompleters are: (%s)', $alias, implode(', ', $this->autocompleteFieldRegistry->getAutocompleterNames())));
         }
 
+        if ($autocompleter instanceof OptionsAwareEntityAutocompleterInterface) {
+            $extraOptions = $this->getExtraOptions($request);
+            $autocompleter->setOptions(['extra_options' => $extraOptions]);
```
stloyd: Should be newly added const too, no?

src/Autocomplete/src/Checksum/ChecksumCalculator.php
smnandre: Would the base64 still be necessary ?

smnandre:
Somewhere in the future documentation, should we add a small warning about the visibility of extra_options ? Something like "the given extra options can be retrieved and therefore should not contain sensible or personal information"

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -79,6 +84,10 @@ public function finishView(FormView $view, FormInterface $form, array $options):
             $values['min-characters'] = $options['min_characters'];
         }
 
+        if ($options['extra_options']) {
```
smnandre, line 87: I have some doubts here, as AutocompleteChoiceTypeExtension support every TextType / ChoiceType (so i'd say 80% of the types used in forms)

Doing so, don't we raise the risk of conflict with userland extensions or other extensions ? 

I did **not** find any other "extra_options" currently in the current Symfony codebase, so it's more a "what if" / "when if" comment. 

This is a problem related to Symfony Form more than this specific PR, but as we extends other types, i'm thinking about having a namespace-ish conventions.

But, as we already define other options without namespace, i would not add confusion with two parralels systems.

So probably "too late" there, but i'd love to have your (and others) opinion on this :)

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -91,6 +100,41 @@ public function finishView(FormView $view, FormInterface $form, array $options):
...
 
+    private function getUrlWithExtraOptions(string $url, array $extraOptions): string
+    {
+        $this->validateExtraOptions($extraOptions);
+
+        $extraOptions[self::CHECKSUM_KEY] = $this->checksumCalculator->calculateForArray($extraOptions);
+        $extraOptions = base64_encode(json_encode($extraOptions));
+
+        return sprintf(
+            '%s%s%s',
```
smnandre, line 112: Suprised to see sprintf "just" to append string... but I must admit it improves code readability

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -104,6 +148,7 @@ public function configureOptions(OptionsResolver $resolver): void
             'min_characters' => null,
             'max_results' => 10,
             'preload' => 'focus',
+            'extra_options' => [],
```
smnandre, line 153: Commenting there but it's more general... Don't you force the extra_options existence everywhere by setting it to [] instead of null ?

src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
```diff
@@ -168,4 +169,9 @@ private function getEntityMetadata(): EntityMetadata
 
         return $this->entityMetadata;
     }
+
+    public function setOptions(array $options): void
```
smnandre, line 173: Maybe add some miniaml phpdoc type signature ? array<string, mixed> ? Or, as it must be query encoded maybe more specific ?

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
```diff
@@ -31,7 +31,7 @@ public function testFieldsRenderWithStimulusController()
             ->throwExceptions()
             ->get('/test-form')
             ->assertElementAttributeContains('#product_category', 'data-controller', 'custom-autocomplete symfony--ux-autocomplete--autocomplete')
-            ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type')
+            ->assertElementAttributeContains('#product_category', 'data-symfony--ux-autocomplete--autocomplete-url-value', '/test/autocomplete/category_autocomplete_type?extra_options=')
```
smnandre, line 34: That illustrate my fear, that is something i'd like to avoid... or we just add a massive BC break, changing the URL for all existing autompleters ? 

Or i missread something ?

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -27,8 +28,12 @@
  */
 final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
 {
-    public function __construct(private ?TranslatorInterface $translator = null)
-    {
+    public const CHECKSUM_KEY = '@checksum';
```
smnandre, line 31: I'm not 100% of this key beeing hard-coded here and statically accessed in the Controller... but i do not have immediate better solution.

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
```diff
@@ -123,4 +123,55 @@ public function testMultipleDoesNotFailWithoutSelectedChoices()
...
+            ])
+            // assert that "Modified Flour" is not included
+            ->assertElementCount('#product_ingredients option', 2)
+            ->assertContains('Flour')
+            ->assertContains('Sugar')
+            ->assertNotContains('Modified Flour')
+        ;
+    }
+
+    public function testItReturnsErrorWhenSendingMalformedExtraOptions(): void
```
smnandre, line 160: This test does not provide much for me, as there is working comparaison.. 

I mean instead of the "$this->encodeData" you could pass "foo" to get the same result no ?

src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Autocomplete;
+
+/**
+ * Interface for classes that will have an "autocomplete" endpoint exposed with a possibility to pass additional options.
+ */
+interface OptionsAwareEntityAutocompleterInterface extends EntityAutocompleterInterface
```
smnandre, line 17: I'd rather have separate Interfaces, what do you think ?

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -91,6 +100,41 @@ public function finishView(FormView $view, FormInterface $form, array $options):
...
+        $parsedUrl = parse_url($url);
+
+        return isset($parsedUrl['query']);
+    }
+
+    private function validateExtraOptions(array $extraOptions): void
+    {
+        foreach ($extraOptions as $option) {
+            if (!\is_scalar($option) && !\is_array($option) && null !== $option) {
+                throw new \InvalidArgumentException(sprintf('Extra option "%s" must be a scalar value, an array or null. Got "%s".', $option, get_debug_type($option)));
```
smnandre: Would that not create an error if $option is a resource or an non stringable object ? 

I'd use keys there i think

src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -54,4 +66,48 @@ public function __invoke(string $alias, Request $request): Response
...
         ]);
     }
+
+    /**
+     * @return array<string, scalar|array|null>
+     */
+    private function getExtraOptions(Request $request): array
+    {
+        if (!$request->query->has(self::EXTRA_OPTIONS)) {
+            return [];
```
smnandre, line 76: Do we still detect if anyone removes the extra options ?

src/Autocomplete/src/Checksum/ChecksumCalculator.php
jakubtobiasz: Right, I guess the `base64_encode` is kind of a leftover before I started using `hash_hmac` (or I used it to get a shorter string) 🤔. We can remove it and increase the string length from 44 to 64. 

In terms of the algorithm:
CleanShot 2023-12-25 at 13 42 35@2x

src/Autocomplete/src/Controller/EntityAutocompleteController.php
jakubtobiasz: Definitely!

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: We can add a configuration entry and make this configurable (the name of the property will be passed via a constructor).

In case of any conflict, a developer will be able to change it. But from my experience, it might be an overkill.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: I don't remember when I used the `.` for the last time to concatenate strings :D.

src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
jakubtobiasz: The doc is present on the method in the interface. 
CleanShot 2023-12-25 at 13 58 06@2x

At the end of day:
1. Interface is used (usually) as an argument type
2. IDE correctly detects the type of the argument

:P

src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
jakubtobiasz: I do like creating "subtype" interfaces like this, but it's not a thing I'd fight for :D. I'm open for changing it.

src/Autocomplete/src/Checksum/ChecksumCalculator.php
stloyd: The question here would be how we treat that data, for me it's checksum which is not a password but should be kinda of cryptographic, but it doesn't need to be perfect as a password and it doesn't need to be as safe as it is.

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
jakubtobiasz: No, I wanted to check two cases here:
1. `@checksum` is required
2. `@checksum` must be valid
I could add also 3rd case `$extraOptionsWithValidChecksum` and check if the status is 200 🤔. I'll add it, and then the test should be complete and sensible.

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
jakubtobiasz: The default value for the `extra_options` is an empty array. In the Form Extension there's
```php
 if ($options['extra_options']) {
```

so it doesn't do anything with already existing URLs if they have no `extra_options` set.

I had to add
```diff
->add('category', CategoryAutocompleteType::class, [
+               'extra_options' => [
+                   'some' => 'cool_option',
+               ],
            ])
```

in the `ProductType` in the `tests/Fixtures` to make Autocomplete add the `extra_options` parameter.

src/Autocomplete/src/Checksum/ChecksumCalculator.php
jakubtobiasz: Ah, you're right. I thought (idk why :D) that we encrypt data here, but the data is not encrypted at all. 

So the answer here is simple, I used `sha256`, because that's exactly the same way as we calculate the checksum in the LiveComponent. I'm open to changing it, I just copied the code from the LiveComponent and didn't bother to think whether `sha256` is the correct one here :D.

jakubtobiasz: Now I'm thinking about extracting checksum calculating inside the LiveComponent component, and using it in the autocomplete component. Autocomplete depends on the LiveComponent so it might make sense.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: AFAIR it'll work exactly the same way.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: I got your point, but in that case I treat the `AutocompleteChoiceTypeExtension` as an information owner. If we want to change the checksum key, we'll have to change only one place. Putting it anywhere else, or creating separate constants won't keep this benefit :/.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: The requirement was to allow setting only scalar values for the extra options, and these lines validate this requirement 🤔.

src/Autocomplete/src/Controller/EntityAutocompleteController.php
jakubtobiasz: It simplifies the code. The method always returns an array, so no additional checks are required method "above".

src/Autocomplete/src/Checksum/ChecksumCalculator.php
smnandre: That would make perfect sense to me. 

Feel free to do it in a second/later PR if you want to release this one first... in all maner this should be "internal" so there is no BC involved there :)

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
smnandre: Yes i do agree.. and as i said it's a broader problem that concern more how the form extensions work than this PR in particular.

src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
smnandre: Oops... that's the problem of reviewing code in the gh interface ... sorry for that!

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
smnandre: Ok I was 100% sure, thank you for thinking of that :)

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
smnandre: Perfect

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
smnandre: What i meant is the sprintf will throw an error if the first %s is non stringable

src/Autocomplete/src/Controller/EntityAutocompleteController.php
smnandre: My scenario was: if someone manually removes the extra_options paramters in the query string, and we expect some to be defined, what would happen ?

src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
smnandre: No strong opinion on this.

i'm often more concerned about "what could happen later" than i probably should... i'm working on it but habits are hard to change :) 

That's why i ask / raise questions but would never "require" changes :))

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
smnandre: Ok with that reasoning, thanks!

smnandre:
Do you have a demo in mind (once its merged) ? 

I was thinking about it, and i think a "basic" demo would be great (with the kind of scenario you describe in the PR introduction).

But even greater, and showing how all the UX component can together create some smart interface, i'd suggest something that combine LiveComponent and Autocomplete+option.

Something like: 

Frame 8

What do you think ? Have you any other idea ?

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: > Scalar variables are those containing an int, float, string or bool. Types array, object, resource and null are not scalar.

This `if` statement won't pass anything else then `int`, `float`, `string`, `bool` nor `array`.

smnandre: https://3v4l.org/Wf0ke

smnandre: This line will throw an error if $option is NOT scalar and NOT array and NOT null ;) 

```php
sprintf('Extra option "%s" must[...]. Got "%s".', $option, get_debug_type($option));
```

src/Autocomplete/src/Checksum/ChecksumCalculator.php
jakubtobiasz: I was 100% sure `LiveComponent` is required by the `Autocomplete`, but it is not.

So, for now, I leave the `Calculator` where it lives. I can mark the class as `internal` but I'll leave the interface as it is, to allow people to create their own implementation of checksum calculator. If anyone finds they need a lighter version of hash algorithm, it'll make sense to create their own.

src/Autocomplete/src/Controller/EntityAutocompleteController.php
jakubtobiasz: 🤷🏼‍♂️

I believe it's not a big deal to not check whether an option was set (and I guess it might be a little bit problematic to check that), but maybe I'm too lax on this topic.

src/Autocomplete/tests/Functional/AutocompleteFormRenderingTest.php
jakubtobiasz: Done, it should be better now :D.

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
jakubtobiasz: Damn :D I was blind :D. Thanks!

Edit: Done

src/Autocomplete/src/Controller/EntityAutocompleteController.php
smnandre: Ok !

src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
smnandre: Don't be :) I should have tested earlier... and suggested a fix...  that would have been more constructive :)

jakubtobiasz:
FYI my solution seems to work. I set up a new Sylius branch locally with our new admin panel, and here're results :).

```php
final class TaxonTypeExtension extends AbstractTypeExtension
{
    /**
     * @param array<string, mixed> $options
     */
    {
        $taxonId = $builder->getData()->getId();

        $builder
            ->add('parent', TaxonAutocompleteChoiceType::class, [
                'label' => 'sylius.form.taxon.parent',
                'extra_options' => [
                    'excluded_taxons' => null !== $taxonId ? [$taxonId] : [],
                ],
            ])
        ;
    }

    public static function getExtendedTypes(): iterable
    {
        return [TaxonType::class];
    }
}
```  

```php
// ...
    {
        $resolver->setDefaults([
            'class' => $this->entityClass,
            'searchable_fields' => ['translations.name'],
            'choice_label' => 'fullname',
            'query_builder' => function (Options $options) {
                return function (EntityRepository $er) use ($options) {
                    $qb = $er->createQueryBuilder('o');
                    $qb->andWhere('o.enabled = true');

                    $taxonsToBeExcluded = $options['extra_options']['excluded_taxons'] ?? [];
                    if ([] !== $taxonsToBeExcluded) {
                        $qb->andWhere($qb->expr()->notIn('o.id', $taxonsToBeExcluded));
                    }

                    return $qb;
                };
            }
        ]);
    }
// ...
```

CleanShot 2023-12-27 at 15 30 58@2x


@smnandre about the demo... I wasn't thinking about it yet :D.

smnandre:
> @smnandre about the demo... I wasn't thinking about it yet :D.

Oh it's not something required or asked in any way haha :)  Just to know if you had something in mind :))

src/Autocomplete/src/Controller/EntityAutocompleteController.php
```diff
@@ -54,4 +66,48 @@ public function __invoke(string $alias, Request $request): Response
...
+
+        return $extraOptions;
+    }
+
+    /**
+     * @return array<string, scalar>
+     */
+    private function getDecodedExtraOptions(string $extraOptions): array
+    {
+        return json_decode(base64_decode($extraOptions), true);
```
smnandre: Sorry for the late comment, but i think we should: 

* add  `JSON_THROW_ON_ERROR` flag here to throw if anything goes wrong (or we will have a type error, as array_key_exists will throw a TypeError if null is passed as second argument)
* probably replace `$request->query->get(self::EXTRA_OPTIONS)` with `$request->query->getString(..)` to avoid the same problem

src/Autocomplete/src/Checksum/ChecksumCalculatorInterface.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
```
smnandre: Not usually set in symfony components AFAIK

src/Autocomplete/tests/Unit/Calculator/ChecksumCalculatorTest.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
```
smnandre, line 3: Not usually set in symfony components AFAIK

src/Autocomplete/src/Checksum/ChecksumCalculator.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
```
smnandre, line 3: Not usually set in symfony components AFAIK

(sorry for the c/c)

src/Autocomplete/src/Checksum/ChecksumCalculator.php
weaverryan: > but I'll leave the interface as it is,

I'd prefer to remove the interface. I can't see a legit reason to need to write your own. If there is a problem with our class, then we want them to tell us so we can fix it :)

src/Autocomplete/src/Controller/EntityAutocompleteController.php
weaverryan: What if we used Symfony's `UriSigner` instead of the checksum? If extra options exist, we sign the URL when generating it and passing it to the frontend. Here, we validate the signed URL.

src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
```diff
@@ -168,4 +169,9 @@ private function getEntityMetadata(): EntityMetadata
...
+    {
+        $this->options = $options;
```
weaverryan, line 179: To be extra safe, if `$this->form` is set, throw an exception. It means the user called this too late.

src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * @param array<string, scalar> $options
```
weaverryan: In practice, we'll pass an array with scalar values. But as this is passed as the `options` argument when creating a form, there is no technical requirement to pass a scalar, right? I'd remove the php doc entirely. Or remove the `@param` and add a description that these options will be passed when creating the form.

src/Autocomplete/src/OptionsAwareEntityAutocompleterInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Autocomplete;
+
+/**
+ * Interface for classes that will have an "autocomplete" endpoint exposed with a possibility to pass additional options.
```
weaverryan: > ... pass additional form options.

weaverryan (commented):
A few comments, but this is really great! Could you also bootstrap docs for this? Thanks!

weaverryan:
Friendly ping :) - I'd love to get this across the finish line. Thanks!

jakubtobiasz:
Hi @weaverryan,
I'll try to get back to it this weekend 😅.

src/Autocomplete/src/Checksum/ChecksumCalculator.php
jakubtobiasz: @weaverryan should I remove it? I've added it as I have it in my PHP class template in the IDE (as I consider it as a good practice) 🤔.

src/Autocomplete/src/Controller/EntityAutocompleteController.php
jakubtobiasz: I've started reimplementing this feature, and... I noticed it computes the hash in the same way as we calculate the checksum :D.

```
private function computeHash(string $uri): string
    {
        return base64_encode(hash_hmac('sha256', $uri, $this->secret, true));
    }
```

I have troubles with testing `UriSigner` as (at least from what I know) it requires the full URL, so it's harder to simulate passing the `_hash` in the tests. If you really want to switch to `UriSigner` I'll rewrite it after the confirmation :P.

src/Autocomplete/src/Checksum/ChecksumCalculator.php
jakubtobiasz: @weaverryan I've removed it!

src/Autocomplete/src/Checksum/ChecksumCalculator.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Autocomplete\Checksum;
+
+/** @internal */
```
smnandre, line 16:
```suggestion
/**
 * @author .... ? 
 * 
 * @internal 
 */
```

(same thing on OptionsAware..Interface ?)

src/Autocomplete/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 -   Added `tom-select/dist/css/tom-select.bootstrap4.css` to `autoimport` - this
     will cause this to appear in your `controllers.json` file by default, but disabled
     see.
+-   Allow passing `extra_options` key in an array passed as a `3rd` argument of the `->add()` method.
```
smnandre, line 14: Compared to [symfony/form changelog](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/Form/CHANGELOG.md), i'd write it maybe more like 

- Add `extra_options` to XXX and YYY types (those options are .... to ...)

smnandre (approved):

weaverryan:
Thanks for your hard work on this Jacob!

arlenreyb:
Not sure if this is bad etiquette, but if anyone came to this thread looking for how to use this new feature, an example of how to use it is in this thread, it's just *hidden* by GitHub. It's [right here](https://github.com/symfony/ux/pull/1322#issuecomment-1870359541).

weaverryan:
Oh yea, it looks like we missed creating a docs PR. @jakubtobiasz are you able to create a quick docs PR for this feature?

PR #1330, opened 2023-12-13
Fix twig deprecation: twig_escape_filter

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/Util/LiveAttributesCollection.php
+++ b/src/LiveComponent/src/Util/LiveAttributesCollection.php
@@ -12,6 +12,7 @@
 namespace Symfony\UX\LiveComponent\Util;
 
 use Twig\Environment;
+use Twig\Extension\EscaperExtension;
 
 /**
  * An array of attributes that can eventually be returned as an escaped array.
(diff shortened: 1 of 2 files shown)


WebMamba:
Oups that was working locally! I get back on it soon

src/LiveComponent/src/Util/LiveAttributesCollection.php
```diff
@@ -99,6 +100,8 @@ public function setBrowserEventsToDispatch(array $browserEventsToDispatch): void
 
     private function escapeAttribute(string $value): string
     {
-        return twig_escape_filter($this->twig, $value, 'html_attr');
+        $escapeExtension = new EscaperExtension();
+
+        return (string) $escapeExtension->escape($this->twig, $value, 'html_attr');
```
Kocal: What about this?
```suggestion
        return (string) EscaperExtension::escape($this->twig, $value, 'html_attr');
```

Since `EscaperExtension::escape` is static https://github.com/twigphp/Twig/blob/3.x/src/Extension/EscaperExtension.php#L182

WebMamba: Good catch! I miss that the escape function was static

kbond (approved):

weaverryan:
Thanks Matheo!

PR #1335, opened 2023-12-13
[StimulusBundle] Add support for typescript controllers

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | n/a
| License       | MIT

To be able to use `typescript`-controllers we need to be able to load those controllers. This PR adds support for loading `*.ts` files. 

Full support for `typescript`-controllers needs the [sensiolabs/typescript-bundle](https://github.com/sensiolabs/AssetMapperTypeScriptBundle) to compile the controllers.

A reproducer project can be found [here](https://github.com/evertharmeling/typescript-sass-stimulus). It needs this PR to run.

Also see this [slack conversation](https://symfony-devs.slack.com/archives/C01FN4EQNLX/p1702461840204599).

### DIFF
--- a/src/StimulusBundle/.gitignore
+++ b/src/StimulusBundle/.gitignore
@@ -1,5 +1,6 @@
 .php-cs-fixer.cache
 .phpunit.result.cache
 composer.lock
+var/
 vendor/
 tests/fixtures/var
--- a/src/StimulusBundle/CHANGELOG.md
+++ b/src/StimulusBundle/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.x
+
+-   Added Typescript controllers support
+
 ## 2.13.2
 
 -   Revert "Change JavaScript package to `type: module`"
--- a/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
+++ b/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
@@ -28,6 +28,8 @@ use Symfony\UX\StimulusBundle\Ux\UxPackageReader;
  */
 class ControllersMapGenerator
 {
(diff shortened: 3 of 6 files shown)


weaverryan:
Good stuff - thank you @evertharmeling!

src/StimulusBundle/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.x
+
+-   Added Typescript controllers support
```
smnandre, line 5: I think there should be doc somewhere explaining a bit more this change.. as it could be disappointing for some :)

src/StimulusBundle/.gitignore
```diff
@@ -1,5 +1,6 @@
 .php-cs-fixer.cache
 .phpunit.result.cache
 composer.lock
+var/
```
smnandre, line 4: If not necessary, could this be reverted ?

src/StimulusBundle/CHANGELOG.md
weaverryan: Good point - just a bit more clarity on how `.ts` controller files are now located (if you use TypeScript bundle to process in the first place)

src/StimulusBundle/.gitignore
evertharmeling: I'd added it because when running the testsuite it created a failure html-file in `var/browser/source` and to not accidentally add it git. I'll create a PR.

src/StimulusBundle/CHANGELOG.md
evertharmeling: Yeah fair enough, I'll try to add some documentation

EDIT: #1345

PR #1336, opened 2023-12-13
Add a note about using raw data attributes as a prefered way

bocharsky-bw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | no 
| License       | MIT

Now with the Stimulus plugin autocomplete in PhpStorm it's better to use raw data attributes instead of built-in functions/filters. Normal Stimulus code is more readable and cleaner IMO.

### DIFF
--- a/src/StimulusBundle/doc/index.rst
+++ b/src/StimulusBundle/doc/index.rst
@@ -57,12 +57,25 @@ directory. In fact, you should have an example controller there already: ``hello
         }
     }
 
-Use the Twig functions from this bundle to activate your controllers:
+Then, activate the controller in your HTML:
+
+.. code-block:: html+twig
+
+    <div data-controller="hello">
+       ...
+    </div>
+
+Optionally, this bundle has a Twig function to render the attribute:
 
 .. code-block:: html+twig
 
     <div {{ stimulus_controller('hello') }}>
         ...
+    </div>
+
+    <!-- would render -->
+    <div data-controller="hello">
+       ...
     </div>
 
 That's it! Whenever this element appears on the page, the ``hello`` controller
@@ -159,6 +172,17 @@ Stimulus Twig Helpers
 This bundle adds 3 Twig functions/filters to help add Stimulus controllers,
 actions & targets in your templates.
 
+.. note::
+
+    Though this bundle provides these helpful Twig functions/filters, it's
+    recommended to use raw data attributes instead, as they're straightforward.
+
+.. tip::
+
+    If you use PhpStorm IDE - you may want to install
+    [Stimulus plugin](https://plugins.jetbrains.com/plugin/18940-stimulus)
+    to get nice auto-completion for the attributes.
+
 stimulus_controller
 ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


weaverryan:
I need to check the PR still, but I’m also preferring the direct attributes these days.

src/StimulusBundle/doc/index.rst
```diff
@@ -65,6 +65,11 @@ Use the Twig functions from this bundle to activate your controllers:
         ...
     </div>
 
+    <!-- would render -->
+    <div data-controller="hello">
+       ...
+    </div>
```
weaverryan, line 79: We should update a bit more, starting with the "Use the Twig functions..." line. Something like:

> Then, activate the controller in your HTML:

(show the raw `data-controller` example).

Then after, something like:

Optionally, this bundle has a Twig function to render the attribute:

(show the `stimulus_controller()`)

src/StimulusBundle/doc/index.rst
```diff
@@ -159,6 +164,13 @@ Stimulus Twig Helpers
 This bundle adds 3 Twig functions/filters to help add Stimulus controllers,
 actions & targets in your templates.
 
+.. note::
+
+    Though this bundle provides these helpful functions/filters, it's recommended to
+    use raw data attributes instead. If you use PhpStorm - you may want to install
```
weaverryan: attributes instead, as they're simple enough. If you use...

src/StimulusBundle/doc/index.rst
```diff
@@ -159,6 +164,13 @@ Stimulus Twig Helpers
...
+    [Stimulus plugin](https://plugins.jetbrains.com/plugin/18940-stimulus) that will
+    bring you nice autocompletion for them.
```
weaverryan: Stimulus plugin to get nice auto-completion for the attributes

bocharsky-bw:
@weaverryan done! Thanks for the review!

src/StimulusBundle/doc/index.rst
```diff
@@ -159,6 +172,14 @@ Stimulus Twig Helpers
 This bundle adds 3 Twig functions/filters to help add Stimulus controllers,
 actions & targets in your templates.
 
+.. note::
+
+    Though this bundle provides these helpful Twig functions/filters, it's
+    recommended to use raw data attributes instead, as they're simple enough.
+    If you use PhpStorm IDE - you may want to install
```
94noni, line 182: I would add this phrase as a tip bloc :)

src/StimulusBundle/doc/index.rst
```diff
@@ -159,6 +172,14 @@ Stimulus Twig Helpers
 This bundle adds 3 Twig functions/filters to help add Stimulus controllers,
 actions & targets in your templates.
 
+.. note::
+
+    Though this bundle provides these helpful Twig functions/filters, it's
+    recommended to use raw data attributes instead, as they're simple enough.
```
94noni: Perhaps replace « simple enough » with « straight forward » ?

94noni (approved):
I second this, it should be the recommended way on doc :)

bocharsky-bw:
@94noni thank you for your review, done!

weaverryan:
Thanks Victor!

PR #1337, opened 2023-12-14
Fix twig_to_array is deprecated

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentNode.php
+++ b/src/TwigComponent/src/Twig/ComponentNode.php
@@ -13,6 +13,7 @@ namespace Symfony\UX\TwigComponent\Twig;
 
 use Symfony\UX\TwigComponent\BlockStack;
 use Twig\Compiler;
+use Twig\Extension\CoreExtension;
 use Twig\Node\Expression\AbstractExpression;
 use Twig\Node\Node;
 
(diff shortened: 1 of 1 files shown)


smnandre:
It's the same bug for all the pacakges failing in the CI ?

WebMamba:
With this PR in complement #1330 the CI should be green again! 😁

src/TwigComponent/src/Twig/ComponentNode.php
```diff
@@ -43,6 +43,16 @@ public function compile(Compiler $compiler): void
     {
         $compiler->addDebugInfo($this);
 
+        // since twig/twig 3.9.0: Using the internal "twig_to_array" function is deprecated.
+        $compiler
+            ->write('$toArray = function ($data) {')
+            ->write('if (method_exists(Twig\Extension\CoreExtension::class, \'toArray\')) {')
```
smnandre: I think this test should be done on compile time and not on render time... Or we will call method_exists a billion time per render..

So i'd make something like 

if (method_exists(...))  {
    $compiler->write('return Twig\Extension\CoreExtension::toArray($data);')
} else {
    $compiler->write('return twig_to_array'...)
}

weaverryan: That's a nice idea 👍

evertharmeling (approved):

smnandre:
Could we merge this one @weaverryan @kbond  ?  

( if it's all good for you @WebMamba )

I'd like to bring the green back on the CI, having a better vision of current PR checks

kbond:
Thanks Matheo.

PR #1339, opened 2023-12-15
[Site] Stop the ReactSvelteVue Mix

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no
| License       | MIT

Remove a loooot of computation, script, events, by not calling React / Svelte / Vue files except on their demo.

### DIFF
--- a/ux.symfony.com/assets/app-vue.js
+++ b/ux.symfony.com/assets/app-vue.js
@@ -1,0 +1,5 @@
+import {registerVueControllerComponents} from '@symfony/ux-vue';
+
+import './app.js';
+
+registerVueControllerComponents();
--- a/ux.symfony.com/importmap.php
+++ b/ux.symfony.com/importmap.php
@@ -18,6 +18,18 @@ return [
         'path' => './assets/app.js',
         'entrypoint' => true,
     ],
+    'app-react' => [
+        'path' => './assets/app-react.js',
+        'entrypoint' => true,
+    ],
+    'app-svelte' => [
+        'path' => './assets/app-svelte.js',
+        'entrypoint' => true,
+    ],
+    'app-vue' => [
+        'path' => './assets/app-vue.js',
+        'entrypoint' => true,
+    ],
     '@symfony/stimulus-bundle' => [
         'path' => '@symfony/stimulus-bundle/loader.js',
     ],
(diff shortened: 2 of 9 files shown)


ux.symfony.com/assets/app-vue.js
```diff
@@ -0,0 +1,5 @@
+import {registerVueControllerComponents} from "@symfony/ux-vue";
```
evertharmeling: `import {registerVueControllerComponents} from '@symfony/ux-vue';`

Single quotes to keep it in line with the rest :)

smnandre: Good catch, thank you

ux.symfony.com/importmap.php
```diff
@@ -156,4 +168,20 @@
...
+    '@swup/theme' => [
+        'version' => '2.1.0',
+    ],
+    '@kurkle/color' => [
+        'version' => '0.3.2',
+    ],
+    'bootstrap/dist/css/bootstrap.min.css' => [
+        'version' => '5.3.2',
+        'type' => 'css',
+    ],
```
weaverryan: Where did these new imports come from? Were they missing before? Or needed after upgrades?

weaverryan (commented):
That’s a very interesting idea! Can you say more about the computation, events, etc that this saves? All of these controllers are lazy, so none of (for example) the react stuff should be loaded on a normal page. What am I not thinking about?

smnandre:
⚠️  ~do not merge for now... i may have a problem with the loading order...~

Update: good

ux.symfony.com/importmap.php
smnandre: Good catch i'll check...

smnandre:
This seems to work now with dedicated entrypoint "extending" app.js 

```js
// app-svelte.js 
import {registerSvelteControllerComponents} from '@symfony/ux-svelte';

import './app.js';

registerSvelteControllerComponents();
```

-- 

Info/note: 

We could also compose things, like that:

```twig
{% extends 'packageBase.html.twig' %}

{% block importmap %}
    {{ importmap(['app', 'app-react']) }}
{% endblock %}
```

```js
// app-react.js

import {registerReactControllerComponents} from '@symfony/ux-react';

registerReactControllerComponents();
```

ux.symfony.com/importmap.php
smnandre: Fixed (i must have made something i should not have 😅 )

weaverryan:
Thanks Simon!

PR #1327, opened 2023-12-10
[LiveComponent] make serializer optional

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        |
| License       | MIT

PR #1341, opened 2023-12-17
[LiveComponent] Make LiveComponentHydrator work without Serializer

### LINKED ISSUE #1340: [LiveComponent] Serializer still required
Capture d’écran 2023-12-17 à 19 08 54

We still need to fix the DependencyInjection (i don't want to "i-told-you-so" but.... 👼  )

(i'm on it)

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| License       | MIT

Allow nullable "serializer" argument in LiveComponentHydrator constructor.

(fix #1340  and complete #1327)

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -50,8 +50,8 @@ final class LiveComponentHydrator
         private iterable $hydrationExtensions,
         private PropertyAccessorInterface $propertyAccessor,
         private LiveComponentMetadataFactory $liveComponentMetadataFactory,
-        private NormalizerInterface|DenormalizerInterface $normalizer,
-        private string $secret
+        private NormalizerInterface|DenormalizerInterface|null $serializer,
+        private string $secret,
     ) {
     }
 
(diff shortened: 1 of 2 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -357,8 +357,14 @@ private function dehydrateValue(mixed $value, LivePropMetadata $propMetadata, ob
             if (!interface_exists(NormalizerInterface::class)) {
                 throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the Serializer component is not installed. Try running "composer require symfony/serializer".', $propMetadata->getName(), $parentObject::class));
             }
+            if (null === $this->serializer) {
+                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but no serializer has been set.', $propMetadata->getName(), $parentObject::class));
+            }
+            if (!$this->serializer instanceof NormalizerInterface) {
+                throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the given serializer does not implement NormalizerInterface.', $propMetadata->getName(), $parentObject::class));
+            }
```
WebMamba, line 365: Do we need those two errors? How those two errors could happen?

WebMamba (commented):
Oups super sorry about that! Thanks a lot @smnandre

src/LiveComponent/src/LiveComponentHydrator.php
smnandre: The null test is required (it's not because symfony/serializer is installed that it is configured in Symfony).

The instance test is necessary because the constructor argument is NormalizerInterface|DenormalizerInterface .. so one could be normalizer without beeing denormalizer :|

weaverryan:
Thanks Simon!

PR #1344, opened 2023-12-19
[Site] Minor fixes

smnandre:
Remove 40kb glitchy background under hero
(it was visible only on the homepage, in dark mode... and see pictures)

Before light: 
before-light

Before dark:
before

After dark
after

Also:
* fix package box titles 
* improve copy button colors


ux.symfony.com/assets/styles/sections/_hero.scss
```diff
@@ -10,20 +10,7 @@
 }
 
 .hero-background {
-  background: url(../images/background_homepage.png);
```
weaverryan, line 13: That file should be deleted also

weaverryan:
Thanks Simon!

PR #1345, opened 2023-12-19
[StimulusBundle] Added docs on TypeScript support

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | #1335 
| License       | MIT

Follow-up on #1335, added more info on usage of TypeScript controllers support.


94noni (approved):
I think it is good to add the link, but perhaps just as a tip bloc and devs would go read the package doc to install/configure

weaverryan:
Thanks Evert!

PR #1346, opened 2023-12-20
Add doc for URL binding feature

squrious:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | NA
| License       | MIT

Add documentation for https://github.com/symfony/ux/pull/1230

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2081,6 +2081,8 @@ then render it manually after:
 
     {{ form_widget(form.todoItems.vars.button_add, { label: '+ Add Item', attr: { class: 'btn btn-outline-primary' } }) }}
 
+.. _validation:
+
 Validation (without a Form)
 ---------------------------
 
@@ -2301,6 +2303,130 @@ You can also trigger a specific "action" instead of a normal re-render:
         data-poll="delay(2000)|save"
         #}
     >
+
+Changing the URL when a LiveProp changes
+----------------------------------------
+
+.. versionadded:: 2.14
+
+    The ``url`` option was introduced in Live Components 2.14.
+
+If you want the URL to update when a ``LiveProp`` changes, you can do that with the ``url`` option::
+
+    // src/Components/SearchModule.php
+    namespace App\Components;
+
+    use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+    use Symfony\UX\LiveComponent\Attribute\LiveProp;
+    use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+    #[AsLiveComponent]
+    class SearchModule
+    {
+        use DefaultActionTrait;
+
+        #[LiveProp(writable: true, url: true)]
+        public string $query = '';
+    }
+
+Now, when the user changes the value of the ``query`` prop, a query parameter in the URL will be updated to reflect the
+new state of your component, for example: ``https://my.domain/search?query=my+search+string``.
+
+If you load this URL in your browser, the ``LiveProp`` value will be initialized using the query string
+(e.g. ``my search string``).
+
+.. note::
+
+    The URL is changed via ``history.replaceState()``. So no new entry is added.
+
+.. warning::
+
+    You can use multiple components with URL bindings in the same page, as long as bound field names don't collide.
+    Otherwise, you will observe unexpected behaviors.
+
+Supported Data Types
+~~~~~~~~~~~~~~~~~~~~
+
+You can use scalars, arrays and objects in your URL bindings:
+
+============================================  =================================================
+JavaScript ``prop`` value                     URL representation
+============================================  =================================================
+``'some search string'``                      ``prop=some+search+string``
+``42``                                        ``prop=42``
+``['foo', 'bar']``                            ``prop[0]=foo&prop[1]=bar``
+``{ foo: 'bar', baz: 42 }``                   ``prop[foo]=bar&prop[baz]=42``
+
+
+When a page is loaded with a query parameter that's bound to a ``LiveProp`` (e.g. ``/search?query=my+search+string``),
+the value - ``my search string`` - goes through the hydration system before it's set onto the property. If a value can't
+be hydrated, it will be ignored.
+
+Multiple Query Parameter Bindings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can use as many URL bindings as you want in your component. To ensure the state is fully represented in the URL,
+all bound props will be set as query parameters, even if their values didn't change.
+
+For example, if you declare the following bindings::
+
+    // ...
+    #[AsLiveComponent]
+    class SearchModule
+    {
+        #[LiveProp(writable: true, url: true)]
+        public string $query = '';
+
+        #[LiveProp(writable: true, url: true)]
+        public string $mode = 'fulltext';
+
+        // ...
+    }
+
+
+And you only set the ``query`` value, then your URL will be updated to
+``https://my.domain/search?query=my+query+string&mode=fulltext``.
+
+Validating the Query Parameter Values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Like any writable ``LiveProp``, because the user can modify this value, you should consider adding
+:ref:`validation <validation>`. When you bind a ``LiveProp`` to the URL, the initial value is not automatically
+validated. To validate it, you have to set up a `PostMount hook`_::
+
+    // ...
+    use Symfony\Component\Validator\Constraints as Assert;
+    use Symfony\UX\LiveComponent\ValidatableComponentTrait;
+    use Symfony\UX\TwigComponent\Attribute\PostMount;
+
+    #[AsLiveComponent]
+    class SearchModule
+    {
+        use ValidatableComponentTrait;
+
+        #[LiveProp(writable: true, url: true)]
+        public string $query = '';
+
+        #[LiveProp(writable: true, url: true)]
+        #[Assert\NotBlank]
+        public string $mode = 'fulltext';
+
+        #[PostMount]
+        public function postMount(): void
+        {
+            // Validate 'mode' field without throwing an exception, so the component can be mounted anyway and a
+            // validation error can be shown to the user
+            if (!$this->validateField('mode', false)) {
+                // Do something when validation fails
+            }
+        }
+
+        // ...
+    }
+
+.. note::
+
+    You can use `validation groups`_ if you want to use specific validation rules only in the PostMount hook.
 
 .. _emit:
 
@@ -3315,3 +3441,5 @@ bound to Symfony's BC policy for the moment.
 .. _`Symfony's built-in form theming techniques`: https://symfony.com/doc/current/form/form_themes.html
 .. _`pass content to Twig Components`: https://symfony.com/bundles/ux-twig-component/current/index.html#passing-blocks
 .. _`Twig Component debug command`: https://symfony.com/bundles/ux-twig-component/current/index.html#debugging-components
+.. _`PostMount hook`: https://symfony.com/bundles/ux-twig-component/current/index.html#postmount-hook
+.. _`validation groups`: https://symfony.com/doc/current/form/validation_groups.html


kbond (commented):
Looks good! Here's a couple questions I had after reading the doc:
- Is the browser's history updated?
- What happens if multiple components on the same page use this feature? Should this be avoided?

squrious:
> Is the browser's history updated?

No, the current behavior is to update the URL in place. There could be a feature to push new history entries in the future.

> What happens if multiple components on the same page use this feature? Should this be avoided?

If multiple components use this feature on the same page, things should work fine as long as they don't use the same field names. Otherwise parameters will collide, so the last rendered component will take priority on URL parameters, overriding ones set by the other components. 

Should this be part of the documentation too?

kbond:
> Should this be part of the documentation too?

Maybe just a warning

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,114 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+============================================  =================================================
+``'some search string'``                      ``prop=some+search+string``
+``42``                                        ``prop=42``
+``['foo', 'bar']``                            ``prop[0]=foo&prop[1]=bar``
+``{ foo: 'bar', baz: 42 }``                   ``prop[foo]=bar&prop[baz]=42``
+
+
+.. note::
+
+    Type consistency is enforced by the LiveComponent internal hydrator when the component is initialized with query
```
smnandre: > Type consistency is enforced by the LiveComponent internal hydrator. When the component is initialized, If a query parameter could not be hydrated properly into a LiveProp, it will be ignored to avoid unfriendly errors for the end user.

Just a suggestion.. feel free to ignore if you prefer the original ;)

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,114 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+    parameters. If a value could not be hydrated properly, it will be ignored to avoid unfriendly errors for the end
+    user.
+
+Multiple bindings
+~~~~~~~~~~~~~~~~~
+
+You can use as many URL bindings as you want in your component. To ensure the state is fully represented in the URL,
+all bound props will be added to query parameters, even if their values didn't change.
```
smnandre: "added" may be missleading there i think...    "set as" ?

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,114 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+        // ...
+    }
+
+
+And you only set the ``query`` value, then your URL will be updated to ``https://my.domain/search?query=my+query+string&mode=fulltext``.
+
+Validation
+~~~~~~~~~~
+
+To validate your component state when it is rendered for the first time, you have to setup a `PostMount hook`_::
```
smnandre: I'd use a warning or caution block there..  wdyt ?

src/LiveComponent/doc/index.rst
squrious: I prefer the original, I find it more accurate as this type consistency check is only used on component initialization. But thanks for the suggestion :)

src/LiveComponent/doc/index.rst
squrious: I added a caution block at the top of the section

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
         #}
     >
 
+Binding LiveProp to URL query parameters
+----------------------------------------
+
+.. versionadded:: 2.13
+
+    The ``url`` option was introduced in Live Components 2.13.
```
weaverryan: 2.14

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
         #}
     >
 
+Binding LiveProp to URL query parameters
```
weaverryan: How about:

> Changing the URL when a LiveProp changes

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+----------------------------------------
+
+.. versionadded:: 2.13
+
+    The ``url`` option was introduced in Live Components 2.13.
+
+You can bind your component LiveProp's values to URL query parameters thanks to the ``url`` option::
```
weaverryan: How about:

> If you want the URL to update when a ``LiveProp`` changes, you can do that with the ``url`` option:

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+
+    // src/Components/SearchModule.php
```
weaverryan, line 2316:
```
    // src/Twig/Components/SearchModule.php
```

right?

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+    #[AsLiveComponent]
+    class SearchModule
+    {
+        use DefaultActionTrait;
+
+        #[LiveProp(writable: true, url: true)]
+        public string $query = '';
+    }
+
+Now if you change the value of the ``query`` prop, the url will be updated to reflect the new state of your
```
weaverryan: > Now, when the user changes the...

> ... prop, a query parameter in the URL will be updated...

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+component, for example: ``https://my.domain/search?query=my+search+string``.
+
+Then, if you load this URL in your browser, your component will be initialized with the values provided in the query
```
```suggestion
If you load this URL in your browser, the ``LiveProp`` value will be initialized using the query string (e.g. ``my search string``).
```

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+string.
+
+.. note::
+
+    URL is updated in place, no new entry is added to the browser's history.
```
weaverryan: The URL is changed via ``history.replaceState()``. So no new entry is added...

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+
+.. warning::
+
+    You can use multiple components with URL bindings in the same page, as long as bound field names don't collide.
+    Otherwise, you will observe unexpected behaviors.
```
weaverryan, line 2345: Someone may have suggested adding this, I say remove it. I think this is self-evident (e.g. if you have 2 variables defined in the same function with the same name, unexpected things will happen).

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+
+Supported data types
```
```suggestion
Supported Data Types
```

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+~~~~~~~~~~~~~~~~~~~~
+
+You can use scalars, arrays and objects in your URL bindings:
```
weaverryan, line 2350: You say objects but then don't show an example below.

src/LiveComponent/doc/index.rst
weaverryan: I think we should make this not in a note, rather a normal sentence, as it's really important. And, from a user's point of view, the important thing is that it goes through the hydration system. Something like:

> When a page is loaded with a query parameter that's bound to a ``LiveProp`` (e.g. ``/search?query=my+search+string``), the value - ``my search string`` - goes through the hydration system before it's set onto the property. If a value can't be hydrated, it will be ignored.

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+``{ foo: 'bar', baz: 42 }``                   ``prop[foo]=bar&prop[baz]=42``
+
+
+.. note::
+
+    Type consistency is enforced by the LiveComponent internal hydrator when the component is initialized with query
+    parameters. If a value could not be hydrated properly, it will be ignored to avoid unfriendly errors for the end
+    user.
+
+Multiple bindings
```
```suggestion
Multiple Query Parameter Bindings
```

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+        #[LiveProp(writable: true, url: true)]
+        public string $mode = 'fulltext';
+
+        // ...
+    }
+
+
+And you only set the ``query`` value, then your URL will be updated to ``https://my.domain/search?query=my+query+string&mode=fulltext``.
+
+Validation
```
```suggestion
Validating the Query Parameter Values
```

src/LiveComponent/doc/index.rst
weaverryan: `setup` -> `set up`

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+
+        #[LiveProp(writable: true, url: true)]
+        #[Assert\NotBlank]
+        public string $mode = 'fulltext';
+
+        #[PostMount]
+        public function postMount(): void
+        {
+            // Validate 'mode' field without throwing an exception
+            $this->validateField('mode', false);
```
weaverryan: It's not clear to me what happens when this fails.

src/LiveComponent/doc/index.rst
```diff
@@ -2302,6 +2302,131 @@ You can also trigger a specific "action" instead of a normal re-render:
...
+
+And you only set the ``query`` value, then your URL will be updated to ``https://my.domain/search?query=my+query+string&mode=fulltext``.
+
+Validation
+~~~~~~~~~~
+
+.. caution::
+
+    It is recommended to validate data provided through query string parameters, so you can prevent malicious inputs
+    from being processed and rendered.
```
weaverryan: This is interesting. If a `LiveProp` is writable, then the fact that it is bound to a query parameter does not open a new malicious entrypoint: the value can always be changed to "anything". Two things:

A) If we want to talk about validating writable properties in general more clearly, then we can talk about that in another PR
B) This section may still make sense, but I think it's only important for the (probably rare) situation where you have `url: true` and `writable: false` (e.g. when the user calls a `LiveAction`, you change the prop inside that method, but the user doesn't change it directly).

weaverryan (commented):
Thank you for this! What a great way to start 2024, being reminded that we can to release this feature soon.

src/LiveComponent/doc/index.rst
squrious: Do you suggest removing the full warning block or just the last line?

src/LiveComponent/doc/index.rst
squrious: That was the purpose of the last example: `{ foo: 'bar', baz: 42 }`. Those values are the "serialized' ones, maybe I should change the table header to something like `JavaScript prop values`?

src/LiveComponent/doc/index.rst
squrious: The goal of this section was to be explicit about the fact that bound props are not validated when initialized from the URL, even if the component uses validation. But yes, the security concern only applies when the validation logic itself involves some security process (sanitizing or whatever). Maybe I can change the caution block to a note like:

`If your component uses validation, it is recommended to also validate data provided through query string parameters.`

For the B) part, I could add another example in this section to illustrate this use case, wdyt?

weaverryan: Hmm, ok, yes, it's more clear to me. So the messaging would be something like:

> Like any writable `LiveProp`, because the user can modify this value, you should consider adding validation (<-- link to that section). When you bind a `LiveProp` to the URL, the initial value is *not* automatically validated. To validate it: (then the code).

Also, it's beyond the scope of the doc, but should the URL-binding feature validate automatically? And reject the value (skip it) if it's invalid?

src/LiveComponent/doc/index.rst
squrious: Other examples in the doc live in `src/Components`, not sure why this one should be different?

weaverryan:
Thank you @squrious!

Issue #1279, opened 2023-11-15
Get rid of Webpack for app manifest and service worker


Spomky:
Hi,

I'm really excited about the direction this frontend integration is taking. It covers almost all we need.

There is one piece that is missing and still covered by Webpack: the service worker.
At the moment, I don't know how to manage that, but I will be happy to contribute to it.

Hereafter, an extract of the `webpack.config.js` and the `assets/sw.js` files

```js
const {InjectManifest} = require('workbox-webpack-plugin');
const WebpackPwaManifest = require('webpack-pwa-manifest');

Encore
    // Usual Webpack config goes here

    // Plugin for compiling the service worker
    .addPlugin(new InjectManifest({
        swSrc: './assets/sw.js',
        swDest: '../sw.js',
        maximumFileSizeToCacheInBytes: 50000000,
        exclude: [
            /\.map$/,
            /manifest$/,
            /\.htaccess$/,
            /service-worker\.js$/,
            /sw\.js$/,
        ],
    }))

    // Plugin for writing the Progressive Web App manifest
    .addPlugin(new WebpackPwaManifest({
        name: 'My App',
        short_name: 'my-app',
        description: 'This is my application',
        filename: "application.json",
        orientation: "any",
        display: "standalone",
        id: "/",
        start_url: "/app",
        inject: false,
        fingerprints: true,
        ios: true,
        publicPath: null,
        includeDirectory: true,
        background_color: '#ffffff',
        crossorigin: 'use-credentials',
        icons: [
            {
                src: path.resolve('assets/images/logo.png'),
                sizes: [48, 57, 60, 72, 76, 96, 114, 128, 144, 152, 180, 192, 256, 384, 512, 1024], // multiple sizes,
            },
            {
                purpose: "maskable",
                src: path.resolve('assets/images/maskable_logo.png'),
                sizes: [48, 57, 60, 72, 76, 96, 114, 128, 144, 152, 180, 192, 256, 384, 512, 1024],
            }
        ],
        protocol_handlers: [
            {
                protocol: "web+app-report",
                url: "/dashboard/report/%s"
            },
            {
                protocol: "web+app-customer",
                url: "/dashboard/customer/%s"
            }
        ]
    }))
;
//... Export
```

```js
import {BackgroundSyncPlugin} from 'workbox-background-sync';
import {
    pageCache,
    imageCache,
    staticResourceCache,
    googleFontsCache,
    offlineFallback,
    warmStrategyCache,
} from 'workbox-recipes';
[...]

PR #1350, opened 2023-12-22
From WA to PWA

Spomky:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Nothing, but for reference #1279
| License       | MIT

The goal of this PR is to add Progressive Web Application capabilities.
I enabled lots of features including:
* The ability to install the application and have a few number of shortcuts
* The application can run offline for certain pages.
* Fine grained cache strategy for a fast application and less server loads.

2024-01-18_12h23_17
2024-02-06_23h20_53


ux.symfony.com/config/packages/pwa.yaml
```diff
@@ -0,0 +1,50 @@
...
+          description: "An Interactive UI entirely built in Twig"
+          url: "app_live_component"
+          icons:
+              - src: "%kernel.project_dir%/assets/images/pwa/component.svg"
+                format: 'png'
+                sizes: [48, 96, 256]
+    serviceworker:
+        generate: true
+        filepath: "%kernel.project_dir%/public/sw.js"
+        src: "/sw.js"
```
weaverryan: Does the bundle generate this file (I see it committed in this PR - but was it generated)?

In a real app, would you want this file - like other .js files - to have a version hash in the filename to bust browser cache on updates?

ux.symfony.com/public/site.webmanifest
```diff
@@ -1 +1,350 @@
-{"name":"Symfony UX","short_name":"Symfony UX","icons":[{"src":"/android-chrome-192x192.png","sizes":"192x192","type":"image/png"},{"src":"/android-chrome-512x512.png","sizes":"512x512","type":"image/png"}],"theme_color":"#ffffff","background_color":"#ffffff","display":"standalone"}
+{
```
weaverryan: Was this file generated or written by hand?

ux.symfony.com/src/Controller/UxPackage/PasswordStrengthController.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace App\Controller\UxPackage;
+
+use App\Form\PasswordStrengthForm;
+use App\Service\UxPackageRepository;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+class PasswordStrengthController extends AbstractController
```
weaverryan: This extra stuff snuck in ;)

ux.symfony.com/templates/base.html.twig
```diff
@@ -14,6 +14,7 @@
         <link rel="icon" href="/favicon.svg" sizes="any" type="image/svg+xml">
         <link rel="apple-touch-icon" href="/apple-touch-icon.png">
         <link rel="manifest" href="/site.webmanifest">
+        <meta name="theme-color" content="#212529">
```
weaverryan: It feels like we should, perhaps, by sharing some values (somehow) between `base.html.twig` and the config file. The `meta` description is another, and also the `site.webmanifest`

weaverryan (commented):
Wow! What a great treat for me to see in the new year. I'd like to focus a lot on PWA's in 2024 - Rails is doing a similar push. So having ux.symfony.com as a PWA playground that we can iterate on is a great idea.

How did you create all of the screenshots? And how many are needed in reality? You included a lot - is one needed for every page? Or were you just being very "robust"? :)

ux.symfony.com/src/Controller/UxPackage/PasswordStrengthController.php
Spomky: 😅

ux.symfony.com/public/site.webmanifest
Spomky: It is generated by the command provided by the bundle.

ux.symfony.com/config/packages/pwa.yaml
Spomky: I think I will split the serviceworker and manifest generation.
The serviceworker is supposed to be created once and modified depending on the application needs.
The manifest could be re-generated as the application evolves.

Spomky:
> How did you create all of the screenshots?

By hand 😐. With the day 30 of the LAST series and this comment, an idea came to my mind💡: what about auto-screenshots generated just by setting a path/route and an image size?

I tried with panther and it just works as expected.
I added a few options and a config like as follows is correctly understood

```php
[
    'path' => 'https://ux.symfony.com',
    'form_factor' => 'narrow',
    'platform' => 'ios',
    'height' =>  2_532,
    'width' =>  1_170,
    'format' => 'png',
],
[
    'path' => 'https://ux.symfony.com',
    'form_factor' => 'wide',
    'platform' => 'ipados',
    'height' =>  1_536,
    'width' =>  2_048,
    'format' => 'png',
],
```


> And how many are needed in reality? You included a lot - is one needed for every page? Or were you just being very "robust"? :)

The number of screenshots is not really clear. What I understand is that only 2 or 3 are really used, but could be more e.i. on Microsoft Store. You can tag a screenshot for a particular platform, but I think it should be limited to specific features you want to highlight.
In this example, I went too far. Some pictures could be removed.

ux.symfony.com/config/packages/pwa.yaml
Spomky: OK it is now a dedicated command `pwa:sw`

Spomky:
@weaverryan I removed the screenshots in this example and now they are [all generated](https://github.com/symfony/ux/pull/1350/files#diff-b2a4507ed8296c012fa725f7d1363c4e39fa7c878cb32558b793980e2b98b157R23-R42) using Panther.

ux.symfony.com/public/site.webmanifest
weaverryan: Should it be committed or running that command be part of the deploy process?

Could it be integrated (in theory) with AssetMapper so that this file is generated dynamically through that? The URL would be something like`/assets/site-{HASH}.webmainfest`, I think. I don't think that's a problem - but I'm not positive. If we did this, the file would be available in dev automatically and would be naturally dumped during deploy with `asset-map:compile`.

weaverryan:
> I tried with panther and it just works as expected.
> I added a few options and a config like as follows is correctly understood

WOW!!!

kbond:
Forgive these ignorant questions but I don't know anything about PWAs:
- with this PR, does it mean we can register ux.symfony.com on the various app stores (Microsoft, android, apple)?
- does it download to your device as a normal app?
- what is the worker you described and how does it add offline support?

Spomky:
Hi @kbond,

> with this PR, does it mean we can register ux.symfony.com on the various app stores (Microsoft, android, apple)?

If you had asked me this question a month ago I probably would have said "no" or "I don't know".
But I recently discovered that it is possible to submit a PWA to an app store by packaging it. I don't know how exactly it works, but tools like https://www.pwabuilder.com/ help create the packages that are submitted.

> does it download to your device as a normal app?

Not exactly. This is still a web application loaded by a browser so there is no full download of the application (although this is still possible). In general, you select the most important pages/assets that are pre-cached. Other pages may be cached as the user visit uses the application.

> what is the worker you described and how does it add offline support?

The Service Worker is a JS script that acts as a proxy and allows you to react to events or interact with the web API. It is saved by the browser when you "install" the application.
Where it's very clever is that `fetch` can be totally controlled. I generally use Workbox because it offers [caching strategies](https://developer.chrome.com/docs/workbox/caching-strategies-overview#stale-while-revalidate) to simplify this (NetworkOnly, CacheOnly, CacheFirst, NetworkFirst, or StaleWhileRevalidate). Offline mode is possible by telling the Service Worker to keep certain responses so your application will always show a real page, including assets. Thanks to the web API like BackgroundSync, it is even possible to keep POST/PUT/DELETE requests in a queue and re-execute them when the network is available again.

ux.symfony.com/public/site.webmanifest
Spomky: Until now I saw the web manifest as a static file generated by the bundle. But indeed it is interesting to leverage on Asset Mapper for having it during dev and ease the deployment.
This could be a way to answer your question on the meta tag `theme-color`, because this value must be present in the manifest and `head` to be used.

The main question to me is: how to manage the assets? Like asset Mapper, the bundle generates the links to the icons and screenshots, but also manipulates them to have them all in the specified format and sizes.
For example with an "app.svg" icon, the bundle may have to create 3 different files depending on the requested sizes and in another format (96x96, 256x256 and 512x512 in PNG format). On this point, I don't know how Asset Mapper can work.

But it is definitively a very good idea to leverage on it => https://github.com/Spomky-Labs/phpwa/issues/28

tacman:
https://store.app/ attempts to be an app store for PWAs.  It's not very good just yet, but the idea is there.  I've submitted some support issues and never received an answer.

Spomky:
That's great. The offline functionality now works wonderfully for reading pages.
Not a single line of JS and yet everything is there.

2024-02-06_23h20_53

ux.symfony.com/.gitignore
```diff
@@ -16,3 +16,6 @@
 /public/assets/
 /assets/vendor/
 ###< symfony/asset-mapper ###
+
+/public/site.manifest
+/public/sw.js
```
maxhelias: These lines can be generated by a Flex recipe for your bundle. And maybe even the files as well unless the generation command is required ?

Spomky: Hi,

My mistake. I put those 2 lines in the early phase of the bundle development, but they are  not require anymore.
The files are not generated unless asked (`symfony console asset-map:compile`).

Spomky: Removed in the last PR update

Spomky:
Hi all,

Initially, this PR was just a playground for the development of a bundle.
Now the bundle is marked as stable and I really think this PR could be considered stable as well.
Let me know what you think about it and if it should be updated, closed or could be merged.

Many thanks.
Regards.

tacman:
@Spomky How do I run this as a PWA?  I've cloned the repo and am in the 2.x branch, but I don't see the pwa-bundle in composer.json.  The features/pwa branch is gone.

PR #1352, opened 2023-12-22
[LiveComponent] remove final

### LINKED ISSUE #1351: [LiveComponent] make AsLiveComponent non-final
Request to remove final from class declaration.

Use-case:
Re-usable components that all use the same template, but have a different name.

Example
```php
abstract class Table
{
    // declare re-usable live props/actions/...
}
```

As-Is:
```php
#[AsLiveComponent('UserTable', 'components/table.html.twig')]
class UserTable extends Table
{
    // specific functions for this table
}
```

Desired:
```php
#[AsTable('UserTable')]
class UserTable extends Table
{
    // specific functions for this table
}

#[\Attribute(\Attribute::TARGET_CLASS)]
class AsTable extends AsLiveComponent
{
    public function __construct(
        string $name = null,
        string $template = 'components/table.html.twig',
    ) {
        parent::__construct($name, $template);
    }
```

https://github.com/symfony/ux-live-component/blob/d0897b7dea722ef9253bd9c07410d6fb80dd4f0a/src/Attribute/AsLiveComponent.php#L22

Are there any consequences to this? What is the reason for AsLiveComponent to be marked final?

jannes-io:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #1351 
| License       | MIT

Remove final from `AsLiveComponent` so implementors can create abstractions. See #1351 for use-case examples.

### DIFF
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -19,7 +19,7 @@ use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
  * @experimental
  */
 #[\Attribute(\Attribute::TARGET_CLASS)]
-final class AsLiveComponent extends AsTwigComponent
+class AsLiveComponent extends AsTwigComponent
 {
     public function __construct(
         string $name = null,


smnandre:
I'm not sure we should remove the "final", as the "trend" in symfony/symfony is more to add "final" where it's possible. 

If you search "final" in the main symfony repository issues, you'll find several similar suggestions. In the answers you'll often find Symfony core team members giving the reasons why.

nicolas-grekas:
It all depends, there's no dogma on the topic.
But here, removing final won't help: using it will still require a dedicated call to registerAttributeForAutoconfiguration to make any child class work.
I would even go one step further and recommend that AsLiveComponent shouldn't extend AsTwigComponent (because the inheritance doesn't mean anything from an abstraction pov).

jannes-io:
> In the answers you'll often find Symfony core team members giving the reasons why.

If there's a solid reason for something to be marked final, then I agree. If it's just because "the trend is x", then that's not good enough...

> using it will still require a dedicated call to registerAttributeForAutoconfiguration to make any child class work.

Can't the abstract class declare the required tags using attributes as well?

smnandre:
If I mentionned the multiple similar topic it’s because I dont want to explain « my » vision, and would not do it well haha 

If i used the word trend it’s to say « and its more going in the opposite direction right now, so don’t give too much hope in a change in the near future ».

jannes-io:
Ok gents, thanks for the quick responses! I’ll go a different route and register the custom attribute for auto wiring in my bundle the same way live component is auto registered. I thought perhaps because Live component extends TwigComponent that this was the way to go but it’s not a big deal to register it manually 😊

Thanks again.

smnandre:
Not sure if that would work.. (as you do not a matching template at first) but you can look at the PreRenderEvent and see if you can change the template there ?

https://symfony.com/bundles/ux-twig-component/current/index.html#prerenderevent

jannes-io:
> Not sure if that would work.. (as you do not a matching template at first) but you can look at the PreRenderEvent and see if you can change the template there ?
> 
> https://symfony.com/bundles/ux-twig-component/current/index.html#prerenderevent

Thanks for the pointer @smnandre !

smnandre:
Another "hacky but probably working" solution would be to create a "AliasLoader"  and pass it to Twig, that would resolve some templates to the one you'd want.

I would not advise you to manually register your components as things can still change a bit on this part of the component :|

smnandre:
> It all depends, there's no dogma on the topic. But here, removing final won't help: using it will still require a dedicated call to registerAttributeForAutoconfiguration to make any child class work. I would even go one step further and recommend that AsLiveComponent shouldn't extend AsTwigComponent (because the inheritance doesn't mean anything from an abstraction pov).

I'm sorry i would not have answered if i saw that you already answered... i just got a notification and saw your comment oO It did not appear before.. even when i re-re-answered...  Did you use some dark github magic ? 🧙‍♂️

PR #1354, opened 2023-12-26
[LiveComponent] Add support for Doctrine ArrayCollections Hydration

jpvdw86:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        |  None
| License       | MIT

The existing DoctrineEntityHydration could not handle Doctrine ArrayCollections. When using the EntityType with the option 'multiple' => true, it returns an ArrayCollection, which result in an error. Additionally, setting useSerializerForHydration to true, in some cases, leads to a circular error.

This PR introduces an ArrayCollectionHydration class to enable support for ArrayCollections.

Example form:
```php
{
        $builder->add('items', EntityType::class, [
            'class' => Item::class,
            'label' => 'Your items',
            'autocomplete' => true,
            'multiple' => true,
            'query_builder' => function (ItemRepository $itemRepository): QueryBuilder {
                return $itemRepository->createQueryBuilder('i'));
            },
        ]);
}

{
        $resolver->setDefaults([
            'data_class' => ItemsDTO::class,
          
        ]);
}
```

Example DTO:

```php
class ItemsDTO
{
    private ArrayCollection $items;

    public function __construct()
    {
        $this->items = new ArrayCollection();
    }

    public function getItems(): ArrayCollection
    {
        return $this->items;
    }

    public function setItems(ArrayCollection $items): void
    {
        $this->items = $items;
    }
}
```

### DIFF
--- a/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
+++ b/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
@@ -31,14 +31,13 @@ final class OptionalDependencyPass implements CompilerPassInterface
     public function process(ContainerBuilder $container): void
     {
         if ($container->hasDefinition('doctrine')) {
-            $doctrineArguments = [new IteratorArgument([new Reference('doctrine')])]; // TODO: add support for multiple entity managers
             $container->register('ux.live_component.doctrine_entity_hydration_extension', DoctrineEntityHydrationExtension::class)
-                ->setArguments($doctrineArguments)
+                ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
 
             $container->register('ux.live_component.array_collection_hydration_extension', ArrayCollectionHydrationExtension::class)
-                ->setArguments($doctrineArguments)
+                ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
         }
--- a/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
+++ b/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
@@ -15,6 +15,7 @@ use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\UX\LiveComponent\Hydration\ArrayCollectionHydrationExtension;
 use Symfony\UX\LiveComponent\Hydration\DoctrineEntityHydrationExtension;
 use Symfony\UX\LiveComponent\LiveComponentBundle;
 
@@ -30,8 +31,14 @@ final class OptionalDependencyPass implements CompilerPassInterface
     public function process(ContainerBuilder $container): void
     {
         if ($container->hasDefinition('doctrine')) {
+            $doctrineArguments = [new IteratorArgument([new Reference('doctrine')])]; // TODO: add support for multiple entity managers
             $container->register('ux.live_component.doctrine_entity_hydration_extension', DoctrineEntityHydrationExtension::class)
-                ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
+                ->setArguments($doctrineArguments)
+                ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
+            ;
+
+            $container->register('ux.live_component.array_collection_hydration_extension', ArrayCollectionHydrationExtension::class)
+                ->setArguments($doctrineArguments)
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
         }
--- a/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
+++ b/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
@@ -15,7 +15,7 @@ use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use Symfony\UX\LiveComponent\Hydration\ArrayCollectionHydrationExtension;
+use Symfony\UX\LiveComponent\Hydration\DoctrineArrayCollectionHydrationExtension;
 use Symfony\UX\LiveComponent\Hydration\DoctrineEntityHydrationExtension;
 use Symfony\UX\LiveComponent\LiveComponentBundle;
 
@@ -36,7 +36,7 @@ final class OptionalDependencyPass implements CompilerPassInterface
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
 
-            $container->register('ux.live_component.array_collection_hydration_extension', ArrayCollectionHydrationExtension::class)
+            $container->register('ux.live_component.doctrine_array_collection_hydration_extension', DoctrineArrayCollectionHydrationExtension::class)
                 ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
--- a/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
+++ b/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
@@ -15,8 +15,8 @@ use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use Symfony\UX\LiveComponent\Hydration\DoctrineArrayCollectionHydrationExtension;
-use Symfony\UX\LiveComponent\Hydration\DoctrineEntityHydrationExtension;
+use Symfony\UX\LiveComponent\Hydration\DoctrineArrayCollectionHydrationTrait;
+use Symfony\UX\LiveComponent\Hydration\DoctrineEntityHydrationTrait;
 use Symfony\UX\LiveComponent\LiveComponentBundle;
 
 /**
@@ -31,12 +31,12 @@ final class OptionalDependencyPass implements CompilerPassInterface
     public function process(ContainerBuilder $container): void
     {
         if ($container->hasDefinition('doctrine')) {
-            $container->register('ux.live_component.doctrine_entity_hydration_extension', DoctrineEntityHydrationExtension::class)
+            $container->register('ux.live_component.doctrine_entity_hydration_extension', DoctrineEntityHydrationTrait::class)
                 ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
 
-            $container->register('ux.live_component.doctrine_array_collection_hydration_extension', DoctrineArrayCollectionHydrationExtension::class)
+            $container->register('ux.live_component.doctrine_array_collection_hydration_extension', DoctrineArrayCollectionHydrationTrait::class)
                 ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
--- a/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
+++ b/src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
@@ -15,8 +15,8 @@ use Symfony\Component\DependencyInjection\Argument\IteratorArgument;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use Symfony\UX\LiveComponent\Hydration\DoctrineArrayCollectionHydrationTrait;
-use Symfony\UX\LiveComponent\Hydration\DoctrineEntityHydrationTrait;
+use Symfony\UX\LiveComponent\Hydration\DoctrineArrayCollectionHydrationExtension;
+use Symfony\UX\LiveComponent\Hydration\DoctrineEntityHydrationExtension;
 use Symfony\UX\LiveComponent\LiveComponentBundle;
 
 /**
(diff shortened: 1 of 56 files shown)


src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
```diff
@@ -30,8 +31,14 @@ final class OptionalDependencyPass implements CompilerPassInterface
     public function process(ContainerBuilder $container): void
     {
         if ($container->hasDefinition('doctrine')) {
+            $doctrineArguments = [new IteratorArgument([new Reference('doctrine')])]; // TODO: add support for multiple entity managers
```
smnandre: I'd personally prefer keeping definitions separated, it's simpler to read/maintain

src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
```diff
@@ -30,8 +31,14 @@ final class OptionalDependencyPass implements CompilerPassInterface
...
             $container->register('ux.live_component.doctrine_entity_hydration_extension', DoctrineEntityHydrationExtension::class)
-                ->setArguments([new IteratorArgument([new Reference('doctrine')])]) // TODO: add support for multiple entity managers
+                ->setArguments($doctrineArguments)
+                ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
+            ;
+
+            $container->register('ux.live_component.array_collection_hydration_extension', ArrayCollectionHydrationExtension::class)
```
smnandre: `ux.live_component.doctrine_array..... `  feels a bit more consistent with the other one

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @experimental
+ *
+ * @internal
+ */
+class ArrayCollectionHydrationExtension extends AbstractDoctrineHydrationExtension implements HydrationExtensionInterface
+{
+    public function supports(string $className): bool
+    {
+        return ArrayCollection::class === $className;
```
smnandre: So this class supports every ArrayCollection, whatever the entity inside ?

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+
+        return $output;
+    }
+
+    public function dehydrate(object $object): mixed
+    {
+        $output = [];
+        foreach ($object as $class) {
```
smnandre: Can we find another name than `class`there ? It feels a bit confusing.

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
```diff
@@ -0,0 +1,56 @@
...
+    public function supports(string $className): bool
+    {
+        return ArrayCollection::class === $className;
+    }
+
+    public function hydrate(mixed $value, string $className): ?object
+    {
+        $output = new ArrayCollection();
+        foreach ($value as $item) {
+            $object = $this->findObject($item['class'], $item['identifierValue']);
```
smnandre: No control on the $item keys ?

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
```diff
@@ -0,0 +1,103 @@
...
+{
+    /**
+     * @param ManagerRegistry[] $managerRegistries
+     */
+    public function __construct(
+        private iterable $managerRegistries,
+    ) {
+    }
+
+    protected function objectManagerFor(string $class): ?ObjectManager
```
smnandre: Could this abstract class be a trait ? Those methods were private, and i'm not sure we should change their visibility

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+namespace Symfony\UX\LiveComponent\Hydration;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Persistence\ManagerRegistry;
+use Doctrine\Persistence\ObjectManager;
+
+/**
+ * @author Jean-Paul van der Wegen <info@jpvdw.nl>
```
smnandre: Maybe the original author of the methods ?

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * @psalm-param class-string<T> $className
+     *
+     * @return object|null
+     *
+     * @psalm-return T|null
+     *
+     * @template T of object
+     */
+    protected function findObject(string $className, mixed $id)
```
smnandre: Add return type ?

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if ($metadata->isEmbeddedClass || $metadata->isMappedSuperclass) {
+            return null;
+        }
+
+        return $om;
+    }
+
+    /**
+     * @psalm-param class-string<T> $className
```
smnandre: Is `@param class-string<T>` possible ?

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @return object|null
+     *
+     * @psalm-return T|null
+     *
+     * @template T of object
```
smnandre: `@template` first in docblock ?

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
```diff
@@ -0,0 +1,56 @@
...
+use Doctrine\Common\Collections\ArrayCollection;
+
+/**
+ * @author Jean-Paul van der Wegen <info@jpvdw.nl>
+ *
+ * @experimental
+ *
+ * @internal
+ */
+class ArrayCollectionHydrationExtension extends AbstractDoctrineHydrationExtension implements HydrationExtensionInterface
```
smnandre: DoctrineArrayCollection... would be more explicit i think

smnandre (commented):
Can you add tests ?

src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
jpvdw86: I have reverted the change and separated it foreach register

src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
jpvdw86: Done !

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
jpvdw86: It should be persistent entities. I have converted abstract class into a trait. Within the trait, the function findObject has been changed to getObject, and if it is not a persistent entity, exception "EntityNotFoundException" will be given

For non-persistent entities, I believe the ArrayCollection is not the right solution. The LiveCollectionType should be used for this usecase ?

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
jpvdw86: There was more mixings of these names. I changed to $className on other places to make it more clear.
In this usecase i used $object as $entityObject

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
jpvdw86: There was no original author. i remove this comment.

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
jpvdw86: Done

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
jpvdw86: Done

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
jpvdw86: Done

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
jpvdw86: In the current setup there is not. Off-course the (de)hydration process will be in the same order. 

I believe that for this use case this approach should be enough. But adding index key is not that big of change. You preference to do this?

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
jpvdw86: Yes it is, i changed to non-psalm version.

jpvdw86:
Integration test is added.

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
smnandre: Sorry, what i meant was: what if "class" or "identifierValue" is missing from $item ?

src/LiveComponent/src/Hydration/AbstractDoctrineHydrationExtension.php
smnandre: Oh sorry it's because you extract a bit of code from the DoctrineHydrationExtension and the way github presented me induced me in error

smnandre:
Thanks for all those changes :) 

Now is time for the "core" questions:
- what are the security risks ? (cc https://symfony.com/bundles/ux-live-component/current/index.html#allowing-an-entity-to-be-changed-to-another )
- how does it work when combined with "writable: true" ?

src/LiveComponent/src/Hydration/ArrayCollectionHydrationExtension.php
jpvdw86: just add some checks and throw exception when is not present.

jpvdw86:
> Thanks for all those changes :)
> 
> Now is time for the "core" questions:
> 
> * what are the security risks ? (cc https://symfony.com/bundles/ux-live-component/current/index.html#allowing-an-entity-to-be-changed-to-another )
> * how does it work when combined with "writable: true" ?

The behavior of this functionality is the same as the existing Entity version. You can't edit the contents of the EntityObject in the ArrayCollection. Its only allowed to add or remove a EntityObject in the ArrayCollection by the className and identifierValue. 

Also just added an test with ArrayCollection as Liveprop with writable: true

jpvdw86:
@smnandre  I've addressed the feedback. Is there anything else needed from my side, or any further questions? Looking forward to your update.

src/LiveComponent/src/Hydration/DoctrineHydrationTrait.php
```diff
@@ -0,0 +1,91 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Hydration;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityNotFoundException;
+use Doctrine\Persistence\ObjectManager;
+
+trait DoctrineHydrationTrait
```
weaverryan, line 21: `@internal`

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -324,6 +326,111 @@ public function onEntireEntityUpdated($oldValue)
...
+            $firstProduct = create(ProductFixtureEntity::class, ['id' => 1, 'name' => 'first'])->object();
+            \assert($firstProduct instanceof ProductFixtureEntity);
+
+            $secondProduct = create(ProductFixtureEntity::class, ['id' => 2, 'name' => 'second'])->object();
+            \assert($secondProduct instanceof ProductFixtureEntity);
+
+            $productList = new ArrayCollection([$firstProduct, $secondProduct]);
+            \assert($productList instanceof ArrayCollection);
+
+            $doctrineEntityDto = create(HoldsArrayCollectionAndEntity::class)->object();
```
weaverryan: Just use:

```php
$doctrineEntityDto = new HoldsArrayCollectionAndEntity();
```

`create()` is a nice way to create entities. But because this is a simple object, for clarity, we can instantiate directly.

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -324,6 +326,111 @@ public function onEntireEntityUpdated($oldValue)
...
+                #[LiveProp()]
+                public HoldsArrayCollectionAndEntity $doctrineEntityDto;
+            })
+                ->mountWith(['doctrineEntityDto' => $doctrineEntityDto])
+                ->assertDehydratesTo([
+                    'doctrineEntityDto' => [
+                        'product' => $firstProduct->id,
+                        'productList' => [
+                            [
+                                'className' => $firstProduct::class,
```
weaverryan, line 354: Thus far, we've avoided sending PHP class names to the frontend. For example, an `array($product1, $product2)` is already supported, as long as the `LiveProp` that this is attached to properly documents that this is `@var Product[]`. That logic, iirc, for hydration is this: https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/LiveComponentHydrator.php#L267-L272 - and here is the fixture class that we use in the tests - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php#L1643-L1648

Not exposing PHP class info to the frontend might be paranoid. But until we get a big need for it, I'd like to continue avoiding it.

But we should definitely support your situation. Does it already work if you add documentation above the property to say that this is a collection - e.g. something like:

```php
class ItemsDTO
{
    /** @var Item[] */
    private ArrayCollection $items;
}
```

If this DOES already work, then what we need is a better error: if we detect that a property can't be hydrated, but it is some sort of iterable of objects, then we can tell them extra documentation is needed. If this does *not* currently work, then we should make it work - but the change might be inside `LiveComponentHydrator`.

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
jpvdw86: Changed!

src/LiveComponent/src/Hydration/DoctrineHydrationTrait.php
jpvdw86: Added

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
jpvdw86: @weaverryan thanks for your reply 

I just tested it and indeed with the annotation and this work like a charm
```php
class ItemsDTO
{
    /** @var Item[] */
    private array $items;
}
```

The ArrayCollection version does not work.
```php
class ItemsDTO
{
    /** @var Item[] */
    private ArrayCollection $items;
}
```

This generates the following error. "Error rendering "form_name" component: Can't get a way to read the property "elements" in class "Doctrine\Common\Collections\ArrayCollection"."

It would be nice to have a clearer error message in this situation.

This PR is not needed anymore :)

weaverryan: Yay!

> It would be nice to have a clearer error message in this situation.

I agree! It feels like, a bit earlier, we should check to see if the property is an iterable. And, if so, throw a better exception. I'd love a PR if you can make one - you obviously understand the technical details quite well.

PR #1365, opened 2023-12-31
[Site] Update README (importmap:install)

smnandre:
- ~~add importmap:install~~
- use sections

Update composer
- add importmap:install in "auto-scripts"
- add PHP8.1 polyfill in "replace"

### DIFF
--- a/ux.symfony.com/README.md
+++ b/ux.symfony.com/README.md
@@ -2,48 +2,63 @@
 
 Source code for [ux.symfony.com](https://ux.symfony.com).
 
-## Contributing
+## Installation
 
-### Local Development
+### Source code
 
-1. Install the project:
-    ```bash
-    git clone git@github.com:symfony/ux
-    cd ux/ux.symfony.com/
-    ```
+Install the project:
+```bash
+git clone git@github.com:symfony/ux
+cd ux/ux.symfony.com/
+```
 
-2. Install the dependencies:
-    ```bash
-    symfony composer install
-    ```
+Install the dependencies:
+```bash
+symfony composer install
+```
 
-3. (optional) Configure Docker:
-    ```bash
-    docker compose up -d
-    ```
+### Services
 
-5. Run database migrations:
-    ```bash
-    symfony console doctrine:migration:migrate
-    ```
+(optional) Configure Docker to launch Mercure
+```bash
+docker compose up -d
+```
 
-6. Populate the database:
-    ```bash
-    symfony console app:load-data
-    ```
+### Database
 
-7. Compile the Sass files:
-    ```bash
-    symfony console sass:build --watch
-    ```
+Run database migrations:
+```bash
+symfony console doctrine:migration:migrate
+```
 
-8. Start the web server:
-    ```bash
-    symfony server:start -d
-    ```
+Populate the database:
+```bash
+symfony console app:load-data
+```
 
+### Assets
 
-### Running the Test Suite
+Download the importmap packages locally:
+```bash
+symfony console importmap:install
+```
+
+Compile the Sass files:
+```bash
+symfony console sass:build
+
+# (optional) Add the --watch flag to automatically recompile the Sass files on change.
+symfony console sass:build --watch
+```
+
+### Local server
+
+Start the local web server (in background):
+```bash
+symfony server:start -d
+```
+
+## Testing
 
 ```bash
 symfony php bin/phpunit
--- a/ux.symfony.com/.symfony.local.yaml
+++ b/ux.symfony.com/.symfony.local.yaml
@@ -1,0 +1,4 @@
+workers:
+    docker_compose: ~
+    sass:
+        cmd: ['symfony', 'console', 'sass:build', '--watch']
--- a/ux.symfony.com/composer.json
+++ b/ux.symfony.com/composer.json
@@ -87,7 +87,8 @@
         "symfony/polyfill-php72": "*",
         "symfony/polyfill-php73": "*",
         "symfony/polyfill-php74": "*",
-        "symfony/polyfill-php80": "*"
+        "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*"
     },
     "scripts": {
         "auto-scripts": {
--- a/ux.symfony.com/composer.lock
+++ b/ux.symfony.com/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "700edc2b087bedd361b1b04b7840c701",
+    "content-hash": "4cdf0a644759acdd08bd5c7f17c23206",
     "packages": [
         {
             "name": "babdev/pagerfanta-bundle",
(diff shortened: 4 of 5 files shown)


ux.symfony.com/README.md
```diff
@@ -2,48 +2,63 @@
...
+
+# (optional) Add the --watch flag to automatically recompile the Sass files on change.
+symfony console sass:build --watch
+```
+
+### Local server
+
+Start the local web server (in background):
+```bash
+symfony server:start -d
```
maxhelias, line 58: Since Symfony CLI 5.7.0 we are able to auto start the docker compose services when starting server with the file `.symfony.local.yaml` and the key `workers.docker_compose: ~`.
Doc : https://symfony.com/doc/current/setup/symfony_server.html#configuring-workers

ux.symfony.com/README.md
```diff
@@ -2,48 +2,63 @@
...
-8. Start the web server:
-    ```bash
-    symfony server:start -d
-    ```
+### Assets
 
+Download the importmap packages locally:
+```bash
+symfony console importmap:install
+```
```
maxhelias, line 44: We can also omit this step if we update the recipes so that it takes place during the composer installation process: https://github.com/symfony/recipes/blob/main/symfony/asset-mapper/6.4/manifest.json#L13C31-L13C31

ux.symfony.com/README.md
smnandre: Not sure to understand… you’d like it to be mentionned ?

ux.symfony.com/README.md
smnandre: Yes great idea, i’ll update it then

ux.symfony.com/README.md
maxhelias: The `symfony server:start` command is able to run the `docker compose up -d` at the same time, this will remove the service step

ux.symfony.com/README.md
smnandre: Done, thank you!

ux.symfony.com/README.md
smnandre: Ok i understand.. but that would need to cmmit the .symfony.local.yaml no ? As Docker is only required for Mercure and that concern only a part of the site, i think i'd rather keep it optional

ux.symfony.com/README.md
```diff
@@ -2,48 +2,63 @@
 
 Source code for [ux.symfony.com](https://ux.symfony.com).
 
-## Contributing
+## Installation
```
94noni, line 5: as a dev who wants to contribute/install locally, why not create a bare raw makefile with all such instructions in it?
i dont see the value of adding them in the readme :s

smnandre: I did not "add" them you know ? :) 

For this README i'm not sure to see what's "wrong" to have a bit of documentation / getting started. 

But If you want to add / use a Makefile i'm sure no one will block you :)

94noni: I mean to help a guys who wants to contribute in this sub project, it surely know symfony and related logic of install
so the added value to describe in the readme is low (to me of course ^^)
perhaps a bare raw makefile and tell the dev to just go make install may be better ;)

ux.symfony.com/README.md
kbond: This file can also run the sass watch/build step: https://symfony.com/bundles/TailwindBundle/current/index.html#symfony-cli (tailwind-bundle docs but same applies to sass-bundle)

smnandre: What is your vision on the ".local" thing. I install many binaries locally to "not depend on" Docker images, but i'm not sure to represent most of users.

Maybe suggest symfony-cli team to handle .symfony.dev.yaml Vs .symfony.local.yaml 

(hmm.. it may already be possible i'll check)

kbond: I'm not sure here as the (optional) docker stuff adds a bit of a wrench in the idea.

maxhelias: My vision was more to reduce the number of steps to launch the project, if we can do it in one command, we should to it.

What do you mean by `handle .symfony.dev.yaml vs .symfony.local.yaml` ? To me the dev env is the integration with platform.sh : https://docs.platform.sh/guides/symfony/local.html which also supported with the symfony-cli

smnandre: II think we should not commit the .symfony.local.yaml, as it's a ".local" file, dedicated to machine settings, and that could be annoying to edit this file everytime i want to work on the website, because IP port is hardcoded in it, or because it requires Docker :)

kbond: What happens if `docker-compose: ~` is added but not using docker for anything? I'm trying to think of a contributor that isn't using docker and isn't contributing to a mercure feature.

smnandre: It seems "nothing", but there is a worker, and i don't find its logs

smnandre: You think "Docker" but if the sass watch tries to run natively in background that's not good either

(here, it's a special case i don't care that much.. but as this file is the one saving the proxy TLD, the local host... it should not be encouraged to share/commit it imho)

smnandre: But i 100% agree with you that it would be ideal to just run "composer install" + "symfony serve" and have a working environment !

smnandre: So let's do how you prefer... initially i just wanted to add "importmap:install" hahaha

ux.symfony.com/README.md
smnandre: Again, feel free to add a Makefile, i may even use it :)

ux.symfony.com/README.md
kbond: Why don't we try adding `.symfony.local.yaml` and see if we have anyone with issues. @smnandre, you could test w/o docker and @maxhelias, you could test with.

I'm thinking the `.symfony.local.yaml` file would look like the following:

```yaml
workers:
    docker_compose: ~
    sass:
        cmd: ['symfony', 'console', 'sass:build', '--watch']
```

smnandre: Well while you do so, i'm gonna work a script to ignore this one and use mine, while not overriding this one 😅

smnandre: (and this is not to say "don't do it", i think it's a good idea... not just for me :p Let's see if i'm the only one :)

smnandre: > ```yaml
> workers:
>     docker_compose: ~
>     sass:
>         cmd: ['symfony', 'console', 'sass:build', '--watch']
> ```

You can commit in this PR ? Or i can update it with this test if you want ?

kbond: > i'm gonna work a script to ignore this one and use mine

Well as soon as one person wants to ignore, it is probably going to be a problem. Perhaps this is a good reason to not add it.

smnandre: No seriously, Let's try. 
And I need to decorate symfony cli in all manners :)

kbond: > Or i can update it with this test if you want ?

Ok, would you mind adding and testing then?

smnandre: Good

kbond (approved):

Issue #687, opened 2023-02-14
Using Stimulus controller with live-component


fajino17:
I've read discussion on this issue https://github.com/symfony/ux/issues/527, but I do not clearly understand what means 'put controller around root element'.
Something like this?
```
<div data-controller="custom-controller">
<div{{attributes}}></div> -- live component root
</div>
```
also
```
div {{ attributes.add(stimulus_controller('custom-controller', { someValue: 'foo' })) }}> -- live component root
```

#https://symfony.com/bundles/ux-live-component/current/index.html#working-with-the-component-in-javascript
I couldn't make none of these working.
Can somebody explain this for me. Thanks

nicodemuz:
@fajino17 did you figure this out?

IvoPereira:
If you have created a controller named `custom_controller`, you would only use `custom` when referencing it.

Basically, you create your component (let's say `NavBar`), and in its Twig definition you would have something in the lines of:
```twig
<div {{ attributes.defaults(stimulus_controller('custom')) }}>
    <!-- your component here -->
</div>
```

PR #1367, opened 2024-01-01
[live-component docs] Rearrange information about using custom Stimul…

srich387:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

In the section 'Working with the Component in JavaScript', the preamble mentions that you can create a custom Stimulus controller, and attach it to or put it around the root component element, in order to control its behaviour.

How to attach a custom controller is, however, only explained further down in the subsection 'Adding a Stimulus Controller to your Component Root Element'. This subsection is currently buried beneath an intervening subsection on a different topic, 'JavaScript Component Hooks', making it easy for a reader to overlook.

I have therefore swapped these two subsections around to improve their logical flow.

It is especially important that this information is presented clearly because it has caused confusion in the past; see e.g. #687.

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -842,6 +842,21 @@ of the change:
 
     input.dispatchEvent(new Event('change', { bubbles: true }));
 
+Adding a Stimulus Controller to your Component Root Element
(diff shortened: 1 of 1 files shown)


smnandre (approved):

kbond:
Thank you @srich387.

PR #1369, opened 2024-01-01
[LiveComponent] Fix test Kernel deprecations

smnandre:
Fix some Kernel deprecation.

But the CS remains, because (and it's kinda weird) PhpCSFixer says

```diff
 --- /home/runner/work/ux/ux/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
+++ /home/runner/work/ux/ux/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
@@ -752,7 +752,7 @@
 
             return HydrationTest::create(new class() {
                 #[LiveProp()]
-                /** @var \Symfony\UX\LiveComponent\Tests\Fixtures\Entity\ProductFixtureEntity[] */
+                /** @var ProductFixtureEntity[] */
                 public array $products = [];
             })
                 ->mountWith(['products' => [$prod1, $prod2, $prod3]])
```                 

But if we do that ....

```
1) Symfony\UX\LiveComponent\Tests\Integration\LiveComponentHydratorTest::testCanDehydrateAndHydrateComponentWithTestCases with data set "Array with objects: (de)hydrates correctly" (Closure Object (...))
LogicException: The LiveProp "products" on component "class@anonymous[...]src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php:753$4c" is an array. We determined the array is full of \ProductFixtureEntity objects, but at least on key had a different value of Symfony\UX\LiveComponent\Tests\Fixtures\Entity\ProductFixtureEntity
```

### DIFF
--- a/src/LiveComponent/tests/Fixtures/Kernel.php
+++ b/src/LiveComponent/tests/Fixtures/Kernel.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\LiveComponent\Tests\Fixtures;
 
+use Composer\InstalledVersions;
 use Doctrine\Bundle\DoctrineBundle\DoctrineBundle;
 use Psr\Log\NullLogger;
 use Symfony\Bundle\FrameworkBundle\FrameworkBundle;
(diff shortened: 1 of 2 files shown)


smnandre:
Any idea @kbond @weaverryan ?

smnandre:
(we should merge this one "now" if we need time to figure out the phpcs bug, as this PR fixes the other CI fails)

kbond:
Thanks Simon.

kbond:
Regarding the cs issue - looks the anon class must be the culprit. Bug in symfony/property-access maybe?

smnandre:
> Regarding the cs issue - looks the anon class must be the culprit. Bug in symfony/property-access maybe?

Hmm I feared that answer haha… well, time to discover the internals of a new component :)

kbond:
I think the easy fix for now would be to use a concrete class instead of anon in this test.

smnandre:
Ok, will do tonight

kbond:
Fixed in #1371

smnandre:
Thank you @kbond !

PR #1378, opened 2024-01-04
[LiveComponent] Add docblocks on attributes

smnandre:
That was on my todo for too long :)

### DIFF
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -14,6 +14,10 @@ namespace Symfony\UX\LiveComponent\Attribute;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 
 /**
+ * An attribute to register a LiveComponent.
+ *
+ * @see https://symfony.com/bundles/ux-live-component
+ *
  * @author Kevin Bond <kevinbond@gmail.com>
  *
  * @experimental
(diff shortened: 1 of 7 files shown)


src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -14,13 +14,26 @@
...
 final class AsLiveComponent extends AsTwigComponent
 {
+    /**
+     * @param string|null $name              the component name (ie: TodoList)
+     * @param string|null $template          The template path of the component (ie: components/TodoList.html.twig).
+     * @param string|null $defaultAction     The default action to call when the component is mounted (ie: __invoke)
+     * @param bool        $exposePublicProps whether to expose every public property as a Twig variable
+     * @param string      $attributesVar     the name of the special "attributes" variable in the template
+     * @param bool        $csrf              Whether to enable CSRF protection (default: true)
+     * @param string      $route             The route used to render the component & handle actions (default: ux_live_component)
```
kbond:
```suggestion
     * @param string|null $name              The component name (ie: TodoList)
     * @param string|null $template          The template path of the component (ie: components/TodoList.html.twig).
     * @param string|null $defaultAction     The default action to call when the component is mounted (ie: __invoke)
     * @param bool        $exposePublicProps Whether to expose every public property as a Twig variable
     * @param string      $attributesVar     The name of the special "attributes" variable in the template
     * @param bool        $csrf              Whether to enable CSRF protection (default: true)
     * @param string      $route             The route used to render the component & handle actions (default: ux_live_component)
```

Let's standardize the casing.

smnandre: Great idea

kbond:
Thank you Simon.

Issue #1358, opened 2023-12-28
[LiveComponent] ElementChanges fails with 'DOMException: DOMTokenList.remove: The empty string is not a valid token' if HTML class attribute contains newlines


manuelkiessling:
I'm using UX Live Components 2.13.2 in a Symfony 5.4.33 project on top of PHP 8.2.14.

Admittingly, I'm kind of holding it wrong, so no bad feelings if this is not going to be addressed — but here we go:

Within the Twig template of a UX Live Component, if I distribute the list of CSS classes in the class attribute of an HTML element over multiple lines, like this:

    <button
        class="
            tw-p-4
            tw-m-8
        "
    >Foo</button>

instead of putting everything on the same line as the opening and closing quotations marks (as in `class="tw-p-4 tw-m-8"`), then every LiveAction call will result in the client-side JavaScript error `Uncaught (in promise) DOMException: DOMTokenList.remove: The token can not contain whitespace.`.

According to the Firefox Developer Tools Debugger, the error is thrown by line `element.classList.remove(className);` of this code:

    key: "applyToElement",
    value: function applyToElement(element) {
      this.addedClasses.forEach(function (className) {
        element.classList.add(className);
      });
      this.removedClasses.forEach(function (className) {
        element.classList.remove(className);
      });

which is part of the `_createClass(ElementChanges, ...` call.

It seems as if the issues is limited to those HTML elements that are "part of the LiveAction" — I'm running into the problem with <button> elements, but not with e.g. <div> elements in other parts of the UX Live Component template.

Looks like line breaks within HTML attributes are not really officially supported, so this is not necessarily a bug, but just me abusing HTML — however, it really helps with readability if there a lots of classes...

smnandre:
It seems to be indeed a syntax violation : https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/add#exceptions 

if you hard-replace in the LiveComponent JS code 
` .add(className)` and `.remove(className)  `
 
 with
 
` .add(className.trim())` and `.remove(className.trim()) `
  
Does it work as expected ?

smnandre:
I wanted to try this, but i'm not able to reproduce your bug... could you share a bit more of your template code ? Or ideally create a little reproducer ?

manuelkiessling:
@smnandre I'll try to find the time, I will also test the behaviour once a release which includes https://github.com/symfony/ux/pull/1380 is out.

smnandre:
[...]

PR #1380, opened 2024-01-05
[LiveComponent] Fix trim classes in live_controller.js

smnandre:
Fix trim classes in live_controller.js .. . and should solve #1358 

Add various CS / minor fixes while i looked at the code... you may refuse as it's way over the original scope :)

### DIFF
--- a/src/LiveComponent/assets/src/Component/ValueStore.ts
+++ b/src/LiveComponent/assets/src/Component/ValueStore.ts
@@ -69,10 +69,7 @@ export default class {
      */
     set(name: string, value: any): boolean {
         const normalizedName = normalizeModelName(name);
-
-        const currentValue = this.get(normalizedName);
-
-        if (currentValue === value) {
+        if (this.get(normalizedName) === value) {
             return false;
         }
 
--- a/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
@@ -3,7 +3,7 @@ import {
     DirectiveModifier,
     parseDirectives
 } from '../../Directive/directives_parser';
-import { combineSpacedArray}  from '../../string_utils';
+import { combineSpacedArray } from '../../string_utils';
 import BackendRequest from '../../Backend/BackendRequest';
 import Component from '../../Component';
 import { PluginInterface } from './PluginInterface';
@@ -64,11 +64,9 @@ export default class implements PluginInterface {
 
         const validModifiers: Map<string, (modifier: DirectiveModifier) => void> = new Map();
         validModifiers.set('delay', (modifier: DirectiveModifier) => {
-            // if loading has *stopped*, the delay modifier has no effect
             if (!isLoading) {
                 return;
             }
-
             delay = modifier.value ? parseInt(modifier.value) : 200;
         });
         validModifiers.set('action', (modifier: DirectiveModifier) => {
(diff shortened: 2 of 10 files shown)


src/LiveComponent/assets/src/Component/ValueStore.ts
```diff
@@ -41,19 +41,10 @@ export default class {
...
-
-        if (this.props[normalizedName] !== undefined) {
-            return this.props[normalizedName];
-        }
-
-        return getDeepData(this.props, normalizedName);
+       return this.dirtyProps[normalizedName] ||
+               this.pendingProps[normalizedName] ||
+               this.props[normalizedName] ||
+               getDeepData(this.props, normalizedName);
```
weaverryan: Is this form of if statement precisely the same as before? For example, if a value is defined, but is null, previously, it would return null. Will it now?

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
```diff
@@ -57,32 +57,34 @@ export default class implements PluginInterface {
 
     private handleLoadingDirective(element: HTMLElement|SVGElement, isLoading: boolean, directive: Directive, backendRequest: BackendRequest|null) {
         const finalAction = parseLoadingAction(directive.action, isLoading);
+        if (!finalAction) {
+            return;
+        }
```
weaverryan: What is this checking for?

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
```diff
@@ -154,13 +137,11 @@ export default class implements PluginInterface {
...
         const loadingDirectives: ElementLoadingDirectives[] = [];
-
-        let matchingElements = element.querySelectorAll('[data-loading]');
+        const matchingElements: Array<HTMLElement|SVGElement> = Array.from(element.querySelectorAll('[data-loading]'));
 
         // querySelectorAll doesn't include the element itself
         if (element.hasAttribute('data-loading')) {
-            // add element at the beginning of matchingElements
-            matchingElements = [element, ...matchingElements];
+            matchingElements.unshift(element);
```
weaverryan: This looks less clear to me. Is there some advantage?

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
```diff
@@ -215,21 +195,18 @@ export default class implements PluginInterface {
...
-            return isLoading ? 'removeAttribute' : 'addAttribute';
+    const actions = new Map(Object.entries({
+        'show': isLoading ? 'show' : 'hide',
+        'hide': isLoading ? 'hide' : 'show',
+        'addClass': isLoading ? 'addClass' : 'removeClass',
+        'removeClass': isLoading ? 'removeClass' : 'addClass',
+        'addAttribute': isLoading ? 'addAttribute' : 'removeAttribute',
+        'removeAttribute': isLoading ? 'removeAttribute' : 'addAttribute',
+    }));
+    if (!actions.has(action)) {
```
weaverryan: Why create a Map just to fetch from it one time? The switch case seems at least as good

src/LiveComponent/assets/src/HookManager.ts
```diff
@@ -13,20 +9,15 @@ export default class {
...
         const index = hooks.indexOf(callback);
-        if (index === -1) {
-            return;
+        if (index !== -1) {
+            hooks.splice(index, 1);
+            this.hooks.set(hookName, hooks);
         }
-
-        hooks.splice(index, 1);
-        this.hooks.set(hookName, hooks);
```
weaverryan: Unless I’m missing something, it was better before: it was using an if to “exit earlier” -> less nested code

weaverryan:
Thank you Simon!

PR #1381, opened 2024-01-07
[Site] Load importmap polyfill conditionnaly

smnandre:
With importmap support now over 90% of total usage.. i think we can try something here :)

### DIFF
--- a/ux.symfony.com/config/packages/asset_mapper.yaml
+++ b/ux.symfony.com/config/packages/asset_mapper.yaml
@@ -6,6 +6,7 @@ framework:
         excluded_patterns:
             - '*/assets/styles/_*.scss'
             - '*/assets/styles/**/_*.scss'
+        importmap_polyfill: false
 
 react:
     controllers_path: '%kernel.project_dir%/assets/build/react/controllers'
--- a/ux.symfony.com/templates/base.html.twig
+++ b/ux.symfony.com/templates/base.html.twig
@@ -26,6 +26,14 @@
                 })();
             </script>
             {% block importmap %}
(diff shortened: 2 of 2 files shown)


smnandre:
Poke @javiereguiluz  https://github.com/symfony/symfony/issues/52939

weaverryan:
@smnandre do you need anything from me on this? Like a deploy of a dev environment to test? Ultimately, I'd like to move this to a PR on Symfony :)

smnandre:
We need something to test things with non-compatible browsers... 

Cannot find it anymore on Slack, but I suggested to get a BrowserStack / similar account.

(on this website... to be honest, i'd skip polyfill entirely with overall support over 92% and this website beeing browsed almost only by tech working visitors haha)

javiereguiluz (approved):
The code used here is minimal and won't easily change ... so I'd say we don't need to test it in a real legacy browser.

smnandre:
> The code used here is minimal and won't easily change ... so I'd say we don't need to test it in a real legacy browser.

But i'd like to be sure it's working ... and my first attempt did not (because `HTMLScriptElement.supports` was not available on old browser).

We also (and i'm almost not kidding) can merge this, test live on ux.symfony.com and wait for people experiencing bug, and then fix/built on that.

The last experiments i've done with AssetMapper showed me that the polyfill performance cost is far from minor (at least for someone so ~~crazy~~focused on performances).

So believe me when i say i'm the first one wanting to remove it for the vast majority of users that do not need it :)

weaverryan:
Thanks! Let's deploy and try it out

PR #1385, opened 2024-01-09
[LiveComponents] Fix 1384

### LINKED ISSUE #1384: [LiveComponent] addCollectionItem don't work when using DTO as source
When using the LiveCollectionType with a DTO as a data object, you now encounter the following error during the addCollectionItem live action. 

Scherm­afbeelding 2024-01-08 om 23 42 48

This occurs because the new item is still null and, according to the $propMetadata, an object is expected.

Example code:

Component:
```php
#[AsLiveComponent]
class VehicleFinancialFormComponent extends AbstractController
{
    use DefaultActionTrait;
    use LiveCollectionTrait;

    #[LiveProp]
    public ?FinancialsModel $financials = null;


    {
        return $this->createForm(FinancialsFormType::class, $this->financials);
    }
}
```
FinancialsDTO:
```php
class FinancialsModel
{
    #[Assert\Valid]
    private array $invoiceRecords;

    #[Assert\Valid]
    private array $estimatedRecords;
}
```

FormType
```php
class FinancialsFormType extends AbstractType
{
    /**
     * {@inheritDoc}
     */
    {

        $builder->add('invoiceRecords', LiveCollectionType::class, [
            'entry_type' => FinancialRecordForm::class,
            'entry_options' => [
                'label' => false,
                'isEstimation' => false,
            ],
            'label' => false,
            'required' => false,
            'allow_add' => true,
            'allow_delete' => true,
            'by_reference' => false,
        ]);
        $builder->add('estimatedRecords', LiveCollectionType::class, [
            'entry_type' => FinancialRecordForm::class,
            'entry_options' => [
                'label' => false,
                'isEstimation' => true,
            ],
            'label' => false,
            'required' => false,
            'allow_add' => true,
            'allow_delete' => true,
            'by_reference' => false,
        ]);
    }


    {
        $resolver->setDefaults([
            'data_class' => FinancialsModel::class,
        ]);
    }
}
```

RecordFormType
```php
class FinancialRecordForm extends AbstractType
{
    {
        $builder
            ->add('description', ChoiceType::class, [
                'label' => false,
                'required' => false,
                'choices' => FinancialRecordType::getChoices(),
                'empty_data' => '',
            ])
            ->add('amountExVat', MoneyType::class, [
                'label' => false,
                'required' => false,
                'empty_data' => '',
            ])
            ->add('vatAmount', MoneyType::class, [
                'label' => false,
                'required' => false,
                'empty_data' => '',
            ])
            ->add('vatable', HiddenType::class, [
                'label' => false,
                'required' => false,
            ]);
        }
    }

    {
        $resolver->setDefaults([
            'data_class' => FinancialRecordModel::class,
            'isEstimation' => false,
        ]);
        $resolver->setAllowedTypes('isEstimation', 'bool');
    }
}
```
RecordDto:
```php
class FinancialRecordModel
{
    #[Assert\NotBlank(message: 'Categorie is verplicht')]
    private ?string $description = null;

    #[Assert\NotBlank(message: 'Bedrag dient minimaal 0 of hoger te zijn')]
    #[Assert\PositiveOrZero(message: 'Bedrag mag niet negatief zijn')]
    private ?float $amountExVat = null;

    #[Assert\NotBlank(message: 'BTW dient minimaal 0 of hoger te zijn')]
    #[Assert\PositiveOrZero(message: 'Bedrag mag niet negatief zijn')]
    private ?float $vatAmount = null;

    private ?bool $vatable = null;


    public function getDescription(): ?string
    {
        return $this->description;
    }

    public function setDescription(?string $description): void
    {
        $this->description = $description;
    }

    public function getAmountExVat(): ?float
    {
        return $this->amountExVat;
    }

    public function setAmountExVat(?float $amountExVat): void
    {
        $this->amountExVat = $amountExVat;
    }

    public function getVatAmount(): ?float
    {
        return $this->vatAmount;
    }

    public function setVatAmount(?float $vatAmount): void
    {
        $this->vatAmount = $vatAmount;
    }

    public function getVatable(): ?bool
    {
        return $this->vatable;
    }

    public function setVatable(?bool $vatable): void
    {
        $this->vatable = $vatable;
    }
}

```

jpvdw86:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1384
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -441,6 +441,10 @@ final class LiveComponentHydrator
             if ($propMetadata->collectionValueType() && Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()) {
                 $collectionClass = $propMetadata->collectionValueType()->getClassName();
                 foreach ($value as $key => $objectItem) {
+                    if (null !== $collectionClass && null === $objectItem) {
+                        $objectItem = new $collectionClass();
+                    }
+
                     if (!$objectItem instanceof $collectionClass) {
                         throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" is an array. We determined the array is full of %s objects, but at least on key had a different value of %s', $propMetadata->getName(), $parentObject::class, $collectionClass, get_debug_type($objectItem)));
                     }
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.14.0
 
 -   Add support for URL binding in `LiveProp`
+-   Fix initialization objectItem when new is added.
 
 ## 2.13.2
 
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.14.0
 
 -   Add support for URL binding in `LiveProp`
--   Fix initialization objectItem when new is added. 
+-   Fix initialization objectItem when new is added.
 
 ## 2.13.2
 
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.14.0
 
 -   Add support for URL binding in `LiveProp`
--   Fix initialization objectItem when new is added.
+-   Fix initialization objectItem when new is added. 
 
 ## 2.13.2
 


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -441,6 +441,10 @@ private function dehydrateValue(mixed $value, LivePropMetadata $propMetadata, ob
             if ($propMetadata->collectionValueType() && Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()) {
                 $collectionClass = $propMetadata->collectionValueType()->getClassName();
                 foreach ($value as $key => $objectItem) {
+                    if (null !== $collectionClass && null === $objectItem) {
+                        $objectItem = new $collectionClass();
```
norkunas, line 445: This will break as soon as constructor will have arguments. (yes we have some DTO's with constructor args) :)

jpvdw86: Having DTO or Entity with constructor is already not allowed in this package:

Look at the follow code in DoctrineEntityHydrationExtension.php; line 38

```php
public function hydrate(mixed $value, string $className): ?object
    {
        // an empty array means a non-persisted entity
        // we support instantiating with no constructor args
        if (\is_array($value) && 0 === \count($value)) {
            return new $className();
        }
```

Also in the hydrateObjectValue function in LiveComponentHydrator.php; line: 498

```php
if (\is_array($value)) {
  $object = new $className();
  foreach ($value as $propertyName => $propertyValue) {
      $reflectionClass = new \ReflectionClass($className);
      $property = $reflectionClass->getProperty($propertyName);
      $propMetadata = $this->liveComponentMetadataFactory->createLivePropMetadata($className, $propertyName, $property, new LiveProp());
      $this->propertyAccessor->setValue($object, $propertyName, $this->hydrateValue($propertyValue, $propMetadata, $component));
  }
  
  return $object;
}
```

norkunas: So that is for the Entity construction, but not the DTO.

Anyway, probably initializers won't be implemented so I just wanted to point it and that's all

WebMamba (commented):
Hey @jpvdw86, thanks for your PR! But I am not sure this is the right way to fix it? And I am not sure there are actually a bug here. The exception you have looks expected, LiveComponent can handle collection of DTO, but only if every item have a same type. The thing is if not every DTO have the same in the collection we can dehydrated, but we don't know how to hydrated back. The fix work in your case but this is quiet like you edit collection during the hydratation and this not what we wan't. I thing the best solution for you is to remove every nullable before you prop get hydrated, or just implement your own hydratation logic like so: https://symfony.com/bundles/ux-live-component/current/index.html#hydrating-with-methods-hydratewith-dehydratewith

src/LiveComponent/src/LiveComponentHydrator.php
jpvdw86: The second example, LiveComponentHydrator.php, is not an Entity example but rather the default hydration function for the DTO.

$object = new $className() is also without arguments. If we allow this in hydration, it should not be a problem in the dehydration function and its resolved this issue.

jpvdw86:
Hi @WebMamba, thanks for your response, and I have also looked into the custom hydration function. However, the issue arises from using the built-in LiveCollectionType and LiveCollectionTrait.

As soon as the built-in Add function it used, it create a index => null record in the existing collection. Would be odd to expect everyone to create a custom hydration function when using LiveCollectionType.  My opinion is that this should be a basic functionality of the package or needs documented.

We could also choose to add more logic in the LiveCollectionTrait; function addCollectionItem() to create an instance here instead of adding empty value.

WebMamba:
> We could also choose to add more logic in the LiveCollectionTrait; function addCollectionItem() to create an instance here instead of adding empty value.

It sounds like a better way to go IMHO

src/LiveComponent/src/LiveComponentHydrator.php
weaverryan: I think we can have a different solution. If your array is truly an array of DTO objects OR null values, then during dehydration, when we see a null value, we can keep that value as null. Then during hydration, when we see the null value in the collection, once again, we keep it as null. This means you get back the exact same array after dehydration & hydration as you had originally.

weaverryan:
Ping! This PR in its current state, I think, is not the correct solution. But we have 2 proposed alternatives:

> We could also choose to add more logic in the LiveCollectionTrait; function addCollectionItem() to create an instance here instead of adding empty value.

And 

https://github.com/symfony/ux/pull/1385#discussion_r1455917789

What's the best path?

PR #1389, opened 2024-01-12
[ChartJs] Add support for chart.js:^4.0

### LINKED ISSUE #607: [ChartJs] Add support for chart.js:^4.0

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #607
| License       | MIT

Since #1202 has been merged, which solved the blockage of #610, the support for ChartJs v4 should be as simple as this PR :) and would replace #610.

🤞🏼 on the CI-flow

### DIFF
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -16,17 +16,17 @@
         },
         "importmap": {
             "@hotwired/stimulus": "^3.0.0",
-            "chart.js/auto": "^3.4.1"
+            "chart.js/auto": "^3.4.1 || ^4.0"
         }
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1 <3.9"
+        "chart.js": "^3.4.1 <3.9 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
         "@types/chart.js": "^2.9.34",
-        "chart.js": "^3.4.1 <3.9",
+        "chart.js": "^3.4.1 <3.9 || ^4.0",
         "resize-observer-polyfill": "^1.5.1",
         "vitest-canvas-mock": "^0.3.3"
     }
--- a/src/Chartjs/CHANGELOG.md
+++ b/src/Chartjs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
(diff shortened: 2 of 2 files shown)


smnandre:
Do we add the "module" thing back now.. or in a distinct release to play safe ? cc @weaverryan

src/Chartjs/assets/package.json
```diff
@@ -21,12 +21,12 @@
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1 <3.9"
+        "chart.js": "^3.4.1 <3.9 || ^4.0"
```
weaverryan, line 24: There's another one above under `importmap` (this is rare, weird case where the node deps vs importmap are different).

weaverryan (commented):
Thanks for getting this! Are the tests using chart.js 4? Irrc, they use what's in `yarn.lock`

src/Chartjs/assets/package.json
evertharmeling: Shoot, fixed!

evertharmeling:
> Thanks for getting this! Are the tests using chart.js 4? Irrc, they use what's in `yarn.lock`

Not sure what's used in the [github flow](https://github.com/symfony/ux/actions/runs/7504456953/job/20431533425?pr=1389#step:5:13), but locally I tested it installing `v3.4.1` and `v4.0` manually and the suite passed.

~~Or should I include [this 'hack'](https://github.com/symfony/ux/pull/610/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R238-R262) again to be sure? But probably we need a more robust solution for supporting multiple versions in the testsuite, as [Svelte](https://github.com/symfony/ux/blob/a2418ef146f09a858bd12ed0116c3f8058d0a3a8/src/Svelte/assets/package.json#L23) is atm supporting 2 versions and probably when Turbo 8 is released we need to continue supporting Turbo 7 at the same time. So it's something we'll come across more often.~~ 
See #1417

weaverryan:
Thanks Evert!

PR #1392, opened 2024-01-13
[LiveComponent] Fix loadingDirectives match elements from nested components

### LINKED ISSUE #1382: [LiveComponent] data-loading behaviour not working correctly in child components
I'm trying to add the loading state behaviour documented at https://symfony.com/bundles/ux-live-component/current/index.html#targeting-loading-for-a-specific-action to my application.

I think I'm running into a bug that is not specific for my setup (PHP 8.2.14, Symfony 5.4.33, symfony/ux-live-component 2.13.3).

I have a component that looks like this:

Class:
```php
#[AsLiveComponent(
    'test_child',
    '@test_child.component.html.twig'
)]
class TestChildComponent extends AbstractController
{
    use DefaultActionTrait;

    #[LiveAction]
    public function slowOperation(): void
    {
        sleep(2);
    }
}
```

Template:
```twig
<div {{ attributes }}>
    <a
            data-action="live#action"
            data-action-name="slowOperation"
    >
        Click me...
    </a>

    <div data-loading="action(slowOperation)|show">Loading...</div>
</div>
```

This works as expected when embedding the component into a template which is served through a "normal" Controller action:

Adobe Express 2024-01-08 09 58 01

(Please see https://www.dropbox.com/scl/fo/cifcwzq66p5fkxmw0lsh6/h?rlkey=h2o2z1753sdoqeqhis17vqf3n&dl=0 for hi-resolution versions of the screen recordings)

The "loading" message briefly flashes while the page and its component is being loaded, and is then only shown while the request for the "slowOperation" action is running, being removed when this has finished — as expected.

This behaviour becomes broken when I put the TestChildComponent into a TestParentComponent, like this:

Parent class:
```php
#[AsLiveComponent(
    'test_parent',
    '@test_parent.component.html.twig'
)]
class TestParentComponent extends AbstractController
{
    use DefaultActionTrait;
}
```

Parent template:
```twig
<div {{ attributes }}>
    <twig:test_child />
</div>
```

My assumption is that this should not change the behaviour in any way, but it does:
Adobe Express 2024-01-08 09 59 35

As before, there is only a brief flash of the loading message upon full-page-load. However, when clicking the button and thus triggering the action, the loading message is shown but *remains* shown, even after the action/request has finished.

Only subsequent clicks then result in the correct behaviour.

I'm seeing even wilder misbehaviour in my actual code, where the element with the data-loading attribute is shown (and remains shown) even when I'm only leaving a form field and no action is triggered at all.

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1382
| License       | MIT

The elements with loading directives were used in all LiveController they were in.

This PR fixes that and restrict loadingDirective to the scope of the LiveContoller (and filter those belonging to nested / child components)

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -2370,27 +2370,27 @@ class StandardElementDriver {
 class LoadingPlugin {
     attachToComponent(component) {
         component.on('loading.state:started', (element, request) => {
-            this.startLoading(element, request);
+            this.startLoading(component, element, request);
         });
         component.on('loading.state:finished', (element) => {
-            this.finishLoading(element);
+            this.finishLoading(component, element);
         });
-        this.finishLoading(component.element);
+        this.finishLoading(component, component.element);
     }
-    startLoading(targetElement, backendRequest) {
-        this.handleLoadingToggle(true, targetElement, backendRequest);
+    startLoading(component, targetElement, backendRequest) {
+        this.handleLoadingToggle(component, true, targetElement, backendRequest);
     }
-    finishLoading(targetElement) {
-        this.handleLoadingToggle(false, targetElement, null);
+    finishLoading(component, targetElement) {
+        this.handleLoadingToggle(component, false, targetElement, null);
     }
-    handleLoadingToggle(isLoading, targetElement, backendRequest) {
+    handleLoadingToggle(component, isLoading, targetElement, backendRequest) {
         if (isLoading) {
             this.addAttributes(targetElement, ['busy']);
         }
         else {
             this.removeAttributes(targetElement, ['busy']);
         }
-        this.getLoadingDirectives(targetElement).forEach(({ element, directives }) => {
+        this.getLoadingDirectives(component, targetElement).forEach(({ element, directives }) => {
             if (isLoading) {
                 this.addAttributes(element, ['data-live-is-loading']);
             }
@@ -2474,9 +2474,10 @@ class LoadingPlugin {
         }
         loadingDirective();
     }
-    getLoadingDirectives(element) {
+    getLoadingDirectives(component, element) {
         const loadingDirectives = [];
-        let matchingElements = element.querySelectorAll('[data-loading]');
+        let matchingElements = [...element.querySelectorAll('[data-loading]')];
+        matchingElements = matchingElements.filter((elt) => elementBelongsToThisComponent(elt, component));
         if (element.hasAttribute('data-loading')) {
             matchingElements = [element, ...matchingElements];
         }
--- a/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
@@ -3,6 +3,7 @@ import {
     DirectiveModifier,
     parseDirectives
 } from '../../Directive/directives_parser';
+import { elementBelongsToThisComponent } from '../../dom_utils';
 import { combineSpacedArray } from '../../string_utils';
 import BackendRequest from '../../Backend/BackendRequest';
 import Component from '../../Component';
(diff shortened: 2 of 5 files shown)


src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -2393,7 +2393,8 @@ class LoadingPlugin {
     }
     getLoadingDirectives(element) {
         const loadingDirectives = [];
-        let matchingElements = element.querySelectorAll('[data-loading]');
+        let matchingElements = [...element.querySelectorAll('[data-loading]')];
+        matchingElements = matchingElements.filter((elt) => elt.closest('[data-controller="live"]') === element);
```
norkunas: This won't work when using multiple controllers, you should use `data-controller~="live"`

smnandre: Oh thank you! Well spotted

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
```diff
@@ -150,7 +150,10 @@ export default class implements PluginInterface {
 
     getLoadingDirectives(element: HTMLElement|SVGElement) {
         const loadingDirectives: ElementLoadingDirectives[] = [];
-        let matchingElements = element.querySelectorAll('[data-loading]');
+        let matchingElements = [...element.querySelectorAll('[data-loading]')];
+
+        // ignore elements which are inside a nested "live" component
+        matchingElements = matchingElements.filter((elt) => elt.closest('[data-controller~="live"]') === element);
```
weaverryan: We have a built-in mechanism for this we should use - it's `elementBelongsToThisComponent` from `dom_utils.ts`. We can pass the `component` instance from `LoadingPlugin.attachToComponent` into the callbacks, then into here so we can call that.

weaverryan (commented):
Thanks for getting on this. Possible to add a test too - should be able to combine something from `child.test.js` into `loading.test.js`.

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
smnandre: I think using the DOM will be better in term of performance, as we just manipulate Node element.. but that could be challenged. Do you want to use it for now ?  

We could probably use only a CSS selector but i think we need to define this browser compatibility list before :)

smnandre: Test added @weaverryan

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -2393,7 +2393,8 @@ class LoadingPlugin {
     }
     getLoadingDirectives(element) {
         const loadingDirectives = [];
-        let matchingElements = element.querySelectorAll('[data-loading]');
+        let matchingElements = [...element.querySelectorAll('[data-loading]')];
+        matchingElements = matchingElements.filter((elt) => elt.closest('[data-controller~="live"]') === element);
```
WebMamba: I am not sure about the use of the `closest` function. `closest` finds only one element, but in a live component you can have multiple elements with a loading attribute in nested element

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
weaverryan: Thanks for adding the test! I still think we should use `elementBelongsToThisComponent()`. By not using it, we're adding "dangling code": we have just one part of the code that works differently than everywhere else. We would need a real, verified performance reason to do such a thing. And, except for cases where there actually *are* children (which is not most live components), the `elementBelongsToThisComponent()` is basically just a call to `element.contains()` - https://github.com/symfony/ux/blob/2.x/src/LiveComponent/assets/src/dom_utils.ts#L204

weaverryan (changes requested):

src/LiveComponent/assets/dist/live_controller.js
smnandre: closest is an ascending selector, it look the first parent "live controller"

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
smnandre: Made changes, i cannot test locally for now, is this how you imagined it ?

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
```diff
@@ -15,33 +16,33 @@ interface ElementLoadingDirectives {
 
 export default class implements PluginInterface {
     attachToComponent(component: Component): void {
-        component.on('loading.state:started', (element: HTMLElement, request: BackendRequest) => {
-            this.startLoading(element, request);
+        component.on('loading.state:started', (element: HTMLElement, request: BackendRequest, component: Component) => {
+            this.startLoading(component, element, request);
```
weaverryan, line 20: We already have the `component` as the argument to `attachToComponent` and we're attaching the `loading.state.started` to that specific component. So o need for the new argument to the callback: we can use the `component` variable already

weaverryan (changes requested):
One minor tweak needed - thanks!

src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
smnandre: As i could not test, i did it in a defensive way :) 

I'll remove it

smnandre: Fix / rebased / rebuilded (style cannot test locally -- will fix this next week)

smnandre: Think i did what you meant but not sure

weaverryan:
Thanks Simon!

smnandre:
Oh damn... i was pushing to rebase .... hope i did not mess on the go with things

weaverryan:
All good, I merged before you rebased. I did the rebasing ;)

PR #1396, opened 2024-01-16
[LiveComponent] Alias URL bound props

squrious:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | N/A
| License       | MIT

Following https://github.com/symfony/ux/pull/1230.

Allow custom parameter names for URL bound props, and mapping specification from Twig templates.

## Usage

From PHP definition: 
```php
#[AsLiveComponent()]
final class MyComponent
{
    // ...

    #[LiveProp(writable: true, url: new QueryMapping(alias: 'q'))
    public ?string $search = null;
}
```

From templates:

```twig
{{ component('MyComponent', {
     'data-live-url-mapping-search': {
       'alias': 'q'
      }
   }) 
}}
{{ component('MyComponent', { 'data-live-url-mapping-search-alias': 'q' }) }}
```

HTML syntax also works:
```twig
<twig:MyComponent :data-live-url-mapping-search="{ alias: 'q'}" />
<twig:MyComponent data-live-url-mapping-search-alias="q" />
```

## Result

Changing the value of `search` will update the url to `https://my.domain?q=my+search+string`. 

Mappings provided in Twig templates are merged into those provided in PHP. Thus, query mappings in PHP act as defaults, and we can override them in templates (e.g. for specific page requirements). So a page with:

```twig
<twig:MyComponent/>
<twig:MyComponent data-live-url-mapping-search-alias="q" />
```

will update its URL to `http://my.domain?search=foo&q=bar`.

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2489,11 +2489,6 @@ If you load this URL in your browser, the ``LiveProp`` value will be initialized
 
     The URL is changed via ``history.replaceState()``. So no new entry is added.
 
-.. warning::
-
-    You can use multiple components with URL bindings in the same page, as long as bound field names don't collide.
-    Otherwise, you will observe unexpected behaviors.
-
 Supported Data Types
 ~~~~~~~~~~~~~~~~~~~~
 
@@ -2537,6 +2532,65 @@ For example, if you declare the following bindings::
 And you only set the ``query`` value, then your URL will be updated to
 ``https://my.domain/search?query=my+query+string&mode=fulltext``.
 
+Controlling the Query Parameter Name
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.17
+
+    The ``as`` option was added in LiveComponents 2.17.
+
+
+Instead of using the prop's field name as the query parameter name, you can use the ``as`` option in your ``LiveProp``
+definition::
+
+    // ...
+    use Symfony\UX\LiveComponent\Metadata\UrlMapping;
+
+    #[AsLiveComponent]
+    class SearchModule
+    {
+        #[LiveProp(writable: true, url: new UrlMapping(as: 'q')]
+        public string $query = '';
+
+        // ...
+    }
+
+Then the ``query`` value will appear in the URL like ``https://my.domain/search?q=my+query+string``.
+
+If you need to change the parameter name on a specific page, you can leverage the :ref:`modifier <modifier>` option::
+
+    // ...
+    use Symfony\UX\LiveComponent\Metadata\UrlMapping;
+
+    #[AsLiveComponent]
+    class SearchModule
+    {
+        #[LiveProp(writable: true, url: true, modifier: 'modifyQueryProp')]
+        public string $query = '';
+
+        #[LiveProp]
+        public ?string $alias = null;
+
+        public function modifyQueryProp(LiveProp $liveProp): LiveProp
+        {
+            if ($this->alias) {
+                $liveProp = $liveProp->withUrl(new UrlMapping(as: $this->alias));
+            }
+            return $liveProp;
+        }
+    }
+
+.. code-block:: html+twig
+
+    <twig:SearchModule alias="q" />
+
+This way you can also use the component multiple times in the same page and avoid collisions in parameter names:
+
+.. code-block:: html+twig
+
+    <twig:SearchModule alias="q1" />
+    <twig:SearchModule alias="q2" />
+
 Validating the Query Parameter Values
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -2564,8 +2618,8 @@ validated. To validate it, you have to set up a `PostMount hook`_::
         #[PostMount]
         public function postMount(): void
         {
-            // Validate 'mode' field without throwing an exception, so the component can be mounted anyway and a
-            // validation error can be shown to the user
+            // Validate 'mode' field without throwing an exception, so the component can
+            // be mounted anyway and a validation error can be shown to the user
             if (!$this->validateField('mode', false)) {
                 // Do something when validation fails
             }
@@ -3500,6 +3554,8 @@ the change of one specific key::
             // ...
         }
     }
+
+.. _modifier:
 
 Set LiveProp Options Dynamically
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/src/LiveComponent/src/Attribute/LiveProp.php
+++ b/src/LiveComponent/src/Attribute/LiveProp.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\UX\LiveComponent\Attribute;
 
+use Symfony\UX\LiveComponent\Metadata\UrlMapping;
+
 /**
  * An attribute to mark a property as a "LiveProp".
  *
@@ -97,10 +99,11 @@ final class LiveProp
         private string|array|null $onUpdated = null,
 
         /**
-         * If true, this property will be synchronized with a query parameter
-         * in the URL.
+         * Whether to synchronize this property with a query parameter
+         * in the URL. Pass true to configure the mapping automatically, or a
+         * {@see UrlMapping} instance to configure the mapping.
          */
-        private bool $url = false,
+        private bool|UrlMapping $url = false,
 
         /**
          * A hook that will be called when this LiveProp is used.
@@ -114,6 +117,10 @@ final class LiveProp
         private ?string $modifier = null,
     ) {
         self::validateHydrationStrategy($this);
+
+        if (true === $url) {
+            $this->url = new UrlMapping();
+        }
     }
 
     /**
@@ -277,15 +284,15 @@ final class LiveProp
         return $clone;
     }
 
-    public function url(): bool
+    public function url(): UrlMapping|false
     {
         return $this->url;
     }
 
-    public function withUrl(bool $url): self
+    public function withUrl(bool|UrlMapping $url): self
     {
         $clone = clone $this;
-        $clone->url = $url;
+        $clone->url = (true === $url) ? new UrlMapping() : $url;
 
         return $clone;
     }
--- a/src/LiveComponent/src/Metadata/UrlMapping.php
+++ b/src/LiveComponent/src/Metadata/UrlMapping.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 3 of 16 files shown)


smnandre:
Is it accessible in PHP easily ?

squrious:
> Is it accessible in PHP easily ?

It's in the `LiveProp` metadata, so yes

src/LiveComponent/src/Attribute/LiveProp.php
```diff
@@ -99,14 +101,19 @@ public function __construct(
...
 
         /**
-         * If true, this property will be synchronized with a query parameter
-         * in the URL.
+         * Whether to synchronize this property with a query parameter
+         * in the URL. Pass true to configure the mapping automatically, or a
+         * {@see QueryMapping} instance to configure the mapping.
          */
-        private bool $url = false,
+        private bool|QueryMapping $url = false,
```
WebMamba: Instead of having the mapping in the url argument, is it not more readable to have a separate argument alias?

weaverryan: I think I like it inside of `url` as a `QueryMapping`. There will be other options coming in the future too. But we could talk about the naming: maybe `UrlMapping` to match the `url` config?

weaverryan:
Thanks for this! 

What's the use-case for needing to do:

> data-live-url-mapping-search-alias="q" 

This doesn't look like something that Livewire has (I check there as a quick way to validate the value of features).

Also: `QueryMapping(alias: 'q')` or `QueryMapping(as: 'q')`? `as` stolen from Livewire :p

smnandre:
> > Is it accessible in PHP easily ?
> 
> It's in the `LiveProp` metadata, so yes

But... it's on a call per call basis... so a call to metadataFactory would not bring this data right ? So it's not usable in events, cache, etc ? 

(just want to clarify when/where this data will be exposed, accessible, etc)

smnandre:
I'm not sure to understand how the live prop mapping is "kept" during live action / event.. but it's probably just me needing coffee..

squrious:
@weaverryan
> What's the use-case for needing to do:
>
> >   data-live-url-mapping-search-alias="q"

If the users just want to override this mapping property they can do it with this shortened syntax instead of passing an object. It may also be useful in the future when the mapping object contain other options: it will be easier to conditionally render specific attributes in this way, than setting up an object, IMO.

> This doesn't look like something that Livewire has (I check there as a quick way to validate the value of features).

That's true, but I tried the Livewire feature, and honestly I missed it. As far I know, you can't avoid parameter name collision in Livewire, but I find it pretty useful.  

> Also: QueryMapping(alias: 'q') or QueryMapping(as: 'q')? as stolen from Livewire :p

I'm totally open for any renaming :) `as` looks good to me!

@smandre

> But... it's on a call per call basis... so a call to metadataFactory would not bring this data right ? So it's not usable in events, cache, etc ?

The metadata defined in your PHP component is available through the metadata factory. If you override the mapping in a template, the overridden values will not be available from metadata. Though there is a [utility class](https://github.com/symfony/ux/blob/977df51e379021ab4ca4aac1fc0e87e240673d5b/src/LiveComponent/src/Util/QueryMappingAttributeUtils.php) that helps in getting those values from mounted attributes.

> I'm not sure to understand how the live prop mapping is "kept" during live action / event.. but it's probably just me needing coffee..

The mapping is render as a Stimulus value [here](https://github.com/symfony/ux/blob/977df51e379021ab4ca4aac1fc0e87e240673d5b/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php#L116).

smnandre:
I'm still not sure if i'm paranoiac ... but if you're telling me that the "mapping configuration" is read and used from request data ... 

I'm 90% sure i still miss something so i stop here :)

squrious:
I renamed QueryMapping to UrlMapping, alias to as,  and rebased to add some doc on this feature :)

src/LiveComponent/doc/index.rst
```diff
@@ -2389,6 +2384,50 @@ For example, if you declare the following bindings::
...
+
+If you need to use a custom parameter name on a specific page, you can override the mapping on a per-template basis:
+
+.. code-block:: html+twig
+
+    {# Pass a mapping object #}
+    <twig:SearchModule :data-live-url-mapping-query="{ alias: 'q'}" />
+
+    {# Or configure the mapping option directly #}
+    <twig:SearchModule data-live-url-mapping-query-alias="q" />
```
weaverryan: Let's talk about this approach :). What I don't love about it is that users need to know about more `data` attribute options. And we need to make sure we keep this value on the frontend so we can reference it on re-renders.

The Livewire solution is to allow the mapping to be done in a method if you have something custom: https://livewire.laravel.com/docs/url#using-the-querystring-method

We could even go more generic: adding some way of configure a `LiveProp` entirely:

```php
#[AsLiveComponent]
class SearchModule
{
    #[LiveProp(writable: true, 'modifier' => 'modifyQueryProp')]
    public string $query = '';

    #[LiveProp]
    public string $alias = null;

    public function modifyQueryProp(LiveProp $prop): void
    {
        if (!$this->alias) {
             return;
        }

        $prop->url = new UrlMapping(as: $this->alias);
    }
}
```

Usage:

```php
<twig:SearchModule alias="q" />
```

A tiny bit more work for the user, but I think it's more understandable (and will also be more flexible). WDYT?

squrious: Oh I like it! It looks even easier to set up. I have a few questions in mind: 
- Does the `alias` prop need to be a `LiveProp`?
- We need to make `LiveProp` writable in some way, is there any drawback?
- The `modifier` method will need the component to be mounted before it's called, so we can use data passed from templates. Are we sure all `LiveProp` options are safe to be overridden here?

weaverryan: Awesome!

> Does the `alias` prop need to be a `LiveProp`?

I think so, yes - because we need some way to persist this setting across re-renders. You were doing that before by making sure to always render the attribute. This, at least, uses a known mechanism.

> We need to make `LiveProp` writable in some way, is there any drawback?

There *could* be, yes - you're correct to worry about this. Currently, you can grab a `LiveProp[]` from a component in a static way: that metadata is always the same for a given component class. I've just started looking at the component. The `LiveComponentMetadata` (which holds the `LiveProp` info) is nicely created only by the `LiveComponentMetadataFactory`. This would be the ideal situation: if we could continue to rely on `LiveComponentMetadataFactory::getMetadata()` everywhere and it would handle this new `modifier` config).

 However, the signature currently is:

```php
public function getMetadata(string $name): LiveComponentMetadata
```

It would need to become:

```php
public function getMetadata(string $name, object $component): LiveComponentMetadata
```

Looking at the code that calls this, that's a problem in at least one spot: `ChildComponentPartialRenderer`. When a parent component re-renders, this component is responsible for figuring out whether or not the child components need to re-render (based on if the input props changed to the child). To do this, it needs the `LiveComponentMetadata` so it can read an `updateFromParent` option on each `LiveProp`. If we do not need to re-render the child, then we even avoid creating it. See the problem? With my idea, we would need to mount/create the child component in order to see if it needs to be re-rendered.

But, this is the ONLY place where that's a problem. A `git grep '\->getMetadata(\$' src/LiveComponent` shows a few other places where we call this method, but all of the have access to the `$component` object. In other words adding a 2nd argument would not be a problem. The `LiveProp` would need to become mutable... though if we want, we could actually keep it immutable and use `->with()` methods to modify it - e.g.

```php
- $prop->url = new UrlMapping(as: $this->alias);
- return $prop->withUrlMapping(new UrlMapping(as: $this->alias));
```

Where `->withUrlMapping()` creates a new instance of itself. A bit more work for the user, but this is not a main use-case.

If we like this, we could work around the `ChildComponentPartialRenderer`, I think. This could be ONE spot where do something different to find these `LiveProp` with `updateFromParent` (and this property would NOT have a `with()` method to change it dynamically). We could, for example, add an internal `LiveComponentMetadataFactory::getMetadataWithoutModifiers(string $name)` that would do what it does now: return the metadata but *without* calling the `modifier`). It would be internal and only used in the one special case of `ChildComponentPartialRenderer`.

> The modifier method will need the component to be mounted before it's called, so we can use data passed from templates. Are we sure all LiveProp options are safe to be overridden here?

Based on the above research and because all metadata is created from `LiveComponentMetadataFactory`, I think yes, they are all safe to be changed *except* for `updateFromParent`.

squrious: Thanks for the investigations! So now this "side feature" of the alias option is growing up. I would remove it from this PR and work on it in a dedicated one, where we could specifically address the `ChildComponentPartialRenderer` modifications you mentioned above. Wdyt?

weaverryan: Ha! Indeed it is - it's exposed a bigger thing!

>  I would remove it from this PR and work on it in a dedicated one, where we could specifically address the ChildComponentPartialRenderer modifications you mentioned above. Wdyt?

It'd be great to focus on that part in a smaller PR. But as LiveComponents will become stable at the end of Feb, I don't want to merge this now, if it'll change later.

If you'd like, there could be a compromise: perhaps you create a PR now with this refactoring of the `getMetadata()` method and `ChildComponentPartialRenderer`? Then we get that merged and *then* update and merge this PR.

Thanks for your patient work - you PR's have proved to be complex in a good way: handling important things in the correct way.

squrious: Alright, let's do this! 

> Thanks for your patient work - you PR's have proved to be complex in a good way: handling important things in the correct way.

Thank you!

squrious:
Hi there! 

Just rebased to get the LiveProp modifier feature. 

I also rewrote the test component for url bindings. We often add use cases and I found it becomes hard to understand things like "prop1", "prop2", etc in assertions. This is in a separate commit, so if it's not the best place do do this it can be removed easily ;) 

Doc section is also updated!

smnandre:
@squrious where are we here ? Do you need something / is it "ready" for you ?

squrious:
> @squrious where are we here ? Do you need something / is it "ready" for you ?

Hi @smnandre! It's ready for me

src/LiveComponent/src/Metadata/UrlMapping.php
```diff
@@ -0,0 +1,33 @@
...
+         * The name of the prop that appears in the URL. If null, the LiveProp's field name is used.
+         */
+        private ?string $as = null,
+    ) {
+    }
+
+    public function getAs(): ?string
+    {
+        return $this->as;
+    }
```
kbond:
```suggestion
        public readonly ?string $as = null,
    ) {
    }
```

kbond (changes requested):
This is great @squrious! One small nitpick.

kbond:
Thanks Nicolas for your time and patience on this!

PR #1397, opened 2024-01-16
[TwigComponent] Fix usage of {% embed %} with {% block %} in <twig:> components, close #952

### LINKED ISSUE #952: [TwigComponent] Block in embed generate error inside `<twig:MyComponent>`
I'm currently switching our old legacy components to Symfony UX components in the form. Exemple : 
`<twig:Separator>...</twig:Separator>`. 

I'm having trouble with a UX component that contains an embed with blocks. If I remove the blocks, the component is displayed correctly. But if I put them back in, I get the following error:
`Expected closing tag "</twig:Separator>" not found.`

Exemple : 
### Component 
```
<div {{ attributes.defaults({ class: 'tw-mb-3' }) }}>
    {% block content '' %}
</div>
```

### Call component : 
```
<twig:Separator>
    <a href="#">
        {% embed "page.html.twig" %}
            {% block extra_classes 'my-extra-classes' %}
        {% endembed %}
    </a>
</twig:Separator>
```

image

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #952 
| License       | MIT

Hi everyone, this PR is an attempt to fix #952.
It fix supports for `{% block X %}foo{% endblock %}` and `{% block X "foo"%}`, from `{% embed %}`, in `<twig:>` syntax.

The following code:
```twig
<twig:Alert>
    <p>
        {% embed "my_embed.html.twig" %}
            {% block my_embed_block_1 "foo" %}
            {% block my_embed_block_2 %}bar{% endblock %}
        {% endembed %}
    </p>
</twig:Alert>
```

is now properly converted to:
```twig
{% component 'Alert' %}
    {% block content %}<p>
        {% embed "my_embed.html.twig" %}
            {% block my_embed_block_1 "foo" %}
            {% block my_embed_block_2 %}bar{% endblock %}
        {% endembed %}
    </p>
{% endblock %}{% endcomponent %}
```

Since I'm not familiar with the `PreLexer`, let me know if a better solution/implementation exists.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -12,6 +12,7 @@
 namespace Symfony\UX\TwigComponent\Twig;
 
 use Twig\Error\SyntaxError;
+use Twig\Lexer;
 
 /**
  * Rewrites <twig:component> syntaxes to {% component %} syntaxes.
@@ -42,6 +43,8 @@ class TwigPreLexer
         $this->length = \strlen($input);
         $output = '';
 
+        $inTwigEmbed = false;
+
         while ($this->position < $this->length) {
             // ignore content inside verbatim block #947
             if ($this->consume('{% verbatim %}')) {
@@ -67,14 +70,30 @@ class TwigPreLexer
                 }
             }
 
+            if ($this->consume('{% embed')) {
+                $inTwigEmbed = true;
+                $output .= '{% embed';
+                $output .= $this->consumeUntil('%}');
+
+                continue;
+            }
+
+            if ($this->consume('{% endembed %}')) {
+                $inTwigEmbed = false;
+                $output .= '{% endembed %}';
+
+                continue;
+            }
+
             $isTwigHtmlOpening = $this->consume('<twig:');
             $isTraditionalBlockOpening = false;
+
             if ($isTwigHtmlOpening || (0 !== \count($this->currentComponents) && $isTraditionalBlockOpening = $this->consume('{% block'))) {
                 $componentName = $isTraditionalBlockOpening ? 'block' : $this->consumeComponentName();
 
                 if ('block' === $componentName) {
                     // if we're already inside the "default" block, let's close it
-                    if (!empty($this->currentComponents) && $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock']) {
+                    if (!empty($this->currentComponents) && $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock'] && !$inTwigEmbed) {
                         $output .= '{% endblock %}';
 
                         $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock'] = false;
@@ -83,8 +102,16 @@ class TwigPreLexer
                     if ($isTraditionalBlockOpening) {
                         // add what we've consumed so far
                         $output .= '{% block';
-                        $output .= $this->consumeUntil('%}');
-                        $output .= $this->consumeUntilEndBlock();
+                        $output .= $stringUntilClosingTag = $this->consumeUntil('%}');
+
+                        // If the last-consumed string does not match the Twig's block name regex, we assume the block is self-closing
+                        $isBlockSelfClosing = '' !== preg_replace(Lexer::REGEX_NAME, '', trim($stringUntilClosingTag));
+
+                        if ($isBlockSelfClosing && $this->consume('%}')) {
+                            $output .= '%}';
+                        } else {
+                            $output .= $this->consumeUntilEndBlock();
+                        }
 
                         continue;
                     }
@@ -420,8 +447,8 @@ class TwigPreLexer
 
             if (!$inComment && '{% endblock %}' === substr($this->input, $this->position, 14)) {
                 if (1 === $depth) {
-                    // in this case, we want to advanced ALL the way beyond the endblock
-                    $this->position += 14;
+                    // in this case, we want to advance ALL the way beyond the endblock
+                    $this->position += 14 /* strlen('{% endblock %}') */;
                     break;
                 } else {
                     --$depth;
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -183,6 +183,104 @@ final class TwigPreLexerTest extends TestCase
             EOF
         ];
 
+        yield 'component_where_entire_default_block_is_twig_embed' => [
+            <<<EOF
+            <twig:Alert>
+                <p>
+                    {% embed "my_embed.html.twig" with { foo: 'bar' } %}{% endembed %}
+                </p>
+            </twig:Alert>
+            EOF,
+            <<<EOF
+            {% component 'Alert' %}
+                {% block content %}<p>
+                    {% embed "my_embed.html.twig" with { foo: 'bar' } %}{% endembed %}
+                </p>
+            {% endblock %}{% endcomponent %}
+            EOF,
+        ];
+        yield 'component_where_entire_default_block_is_twig_embed_with_block_string' => [
+            <<<EOF
+            <twig:Alert>
+                <p>
+                    {% embed "my_embed.html.twig" %}
+                        {% block my_embed_block "foo" %}
+                    {% endembed %}
+                </p>
+            </twig:Alert>
+            EOF,
+            <<<EOF
+            {% component 'Alert' %}
+                {% block content %}<p>
+                    {% embed "my_embed.html.twig" %}
+                        {% block my_embed_block "foo" %}
+                    {% endembed %}
+                </p>
+            {% endblock %}{% endcomponent %}
+            EOF,
+        ];
+        yield 'component_where_entire_default_block_is_twig_embed_with_block_variable' => [
+            <<<EOF
+            <twig:Alert>
+                <p>
+                    {% embed "my_embed.html.twig" %}
+                        {% block my_embed_block fooVar %}
+                    {% endembed %}
+                </p>
+            </twig:Alert>
+            EOF,
+            <<<EOF
+            {% component 'Alert' %}
+                {% block content %}<p>
+                    {% embed "my_embed.html.twig" %}
+                        {% block my_embed_block fooVar %}
+                    {% endembed %}
+                </p>
+            {% endblock %}{% endcomponent %}
+            EOF,
+        ];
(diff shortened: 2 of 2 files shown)


smnandre:
I'm **personnaly**  👎 

Main reasons:
* I'm really not in favor of mixing syntaxes... 
* I don't want twig:blocks to be used outside components
* TwigComponent should not be a trojan horse to change all the Twig syntax

When we'll rewrite the HTML prelexer, or the component rendering/ code generation...  we don't want to deal with all Twig. As we already have not all the time to solve or improve TwigComponent / LiveComponent things.

I won't fight this in anyway of course, but really i'd rather go the other way around and forbid <twig:block /> outside TwigComponent "today"...

Kocal:
So what's your alternative for smoothly migrating big code-base that uses `{% embed %}` with `{% block %}` inside, without adding a lot of works?

To me, blocks inside `<twig:>`' component should accept any valid Twig syntax.

smnandre:
> To me, blocks inside <twig:>' component should accept any valid Twig syntax.

On this, i 100% agree.  Even if "philosophically" i'm having doubt about this embed-in-template situation.
I really agree (or we should announce this "non-compatibility").

But this may/should not be done by writing more code about specific template tags (again, not criticising your work on this, far from it..) but i lost so many hours trying to debug things in Twig/Live components, that i'm 100% certain if we start to code specific things about every tag construct of twig, there is a big maintanability problem... 

So i really do not thing we should have any code concerning "embed"  in this repository. Or we should also have to handle "macro", "with"..... and "use". 

That beeing said, what are the alternatives right now ? I don't know :(  Can this by fixing exisiting block handling ?

Kocal:
> But this may/should not be done by writing more code about specific template tags (again, not criticising your work on this, far from it..) but i lost so many hours trying to debug things in Twig/Live components, that i'm 100% certain if we start to code specific things about every tag construct of twig, there is a big maintanability problem...
> 
> So i really do not thing we should have any code concerning "embed" in this repository. Or we should also have to handle "macro", "with"..... and "use".

Yeah I understand your point too, maybe my approach is too naive and can be optimized for others Twig tokens like `macro`, `with`...
 
> That beeing said, what are the alternatives right now ? I don't know :( Can this by fixing exisiting block handling ?

TBH I don't know, all I can say during my debug phase is that the following code (without blocks in embed): 
```twig
<twig:Alert>
    <p>
        {% embed "my_embed.html.twig" with { foo: 'bar' } %}{% endembed %}
    </p>
</twig:Alert>
```

is nicely converted to
```twig
{% component 'Alert' %}
    {% block content %}<p>
        {% embed "my_embed.html.twig" with { foo: 'bar' } %}{% endembed %}
    </p>
{% endblock %}{% endcomponent %}
```

But this code:
```
<twig:Alert>
    <p>
        {% embed "my_embed.html.twig" %}
            {% block my_embed_block_1 "foo" %}
            {% block my_embed_block_2 %}bar{% endblock %}
        {% endembed %}
    </p>
</twig:Alert>
```

raises error `Expected closing tag "</twig:Alert>" not found at line 8`, and generates the following output (before raising the error): 
```twig
{% component 'Alert' %}
    {% block content %}<p>
        {% embed "my_embed.html.twig" %}
            {% endblock %}{% block my_embed_block_1 "foo" %}
            {% block my_embed_block_2 %}bar{% endblock %}
        {% endembed %}
    </p>
</twig:Alert>
```

You can see a `{% endblock %}` between `{% embed "my_embed.html.twig" %}` and `{% block my_embed_block_1 "foo" %}`, it has been added by the line 
```php
if (!empty($this->currentComponents) && $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock']) {
    $output .= '{% endblock %}';

    $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock'] = false;
}
```

weaverryan:
Yo @Kocal!

We should definitely fix this. But, to be clear, I think there are 2 separate issues. I think you fixed them both at once, but I want to clarity.

**Issue 1**: blocks without `endblock`:

```
<twig:Terminal bottomPadding="145">
    composer require symfony/asset-mapper symfony/stimulus-bundle
    {% block my_embed_block_1 "foo" %}
</twig:Terminal>
```

This errors - we're simply not processing the `{% block anything "content" %}` properly.

**Issue 2**: the embed issue where we think blocks that are inside of the `embed` are blocks for the component.

smnandre:
I think fixing the block thing in the Prelexer would fix both, no ? https://github.com/symfony/ux/blob/a2418ef146f09a858bd12ed0116c3f8058d0a3a8/src/TwigComponent/src/Twig/TwigPreLexer.php#L86

In all manner (and not related to this PR) we have to refacto/improve the PreLexer soon, as we do not handle a lot of minor / edge cases

smnandre:
@Kocal  .. do you think you could extract this part in your code and try without any "embed" handling at first ? Or you do think both need to be fixed at once ?

Kocal:
Hi everyone 👋, thanks for your suggestions, I'll try them ASAP!

Cheers from the mountains 😛

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -67,24 +70,54 @@ public function preLexComponents(string $input): string
...
                         $output .= '{% block';
                         $output .= $this->consumeUntil('%}');
-                        $output .= $this->consumeUntilEndBlock();
+
+                        // If the last consumed string does not match the Twig's block name regex, we assume the block is self-closing
+                        $isBlockSelfClosing = '' !== preg_replace(
+                            Lexer::REGEX_NAME,
+                            '',
+                            trim(mb_substr($this->input, $prevPosition, $this->position - $prevPosition)))
+                        ;
```
weaverryan: Instead of keeping calculating the content between the positions, can we use the return value of `$this->consumeUntil('%}');`?

Also, in the test, you have

```twig
{% block my_embed_block_1 "foo" %}
```

I can understand that, after the `preg_replace()`, the `""` would remain. And so `isBlockSelfClosing` would be true. But isn't this a valid syntax?

```twig
{% block my_embed_block_1 fooVariable %}
```

In this case, everything would be replaced, the result would be an empty string, and `isBlockSelfClosing` would be false, right? Could we, instead, (A) grab the content between `{% block` and `%}`, `trim()` the whitespace, then set `isBlockSelfClosing` to true if there are ANY non-whitespace characters left?

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
```diff
@@ -183,6 +183,45 @@ public static function getLexTests(): iterable
...
+                </p>
+            {% endblock %}{% endcomponent %}
+            EOF,
+        ];
+        yield 'component_where_entire_default_block_is_twig_embed_with_blocks' => [
+            <<<EOF
+            <twig:Alert>
+                <p>
+                    {% embed "my_embed.html.twig" %}
+                        {% block my_embed_block_1 "foo" %}
```
weaverryan: I'd love to see this tested independently, both with a string - like '"foo"` - and also a variable like `{% block my_block fooVar %}`

weaverryan (commented):
I'm ok with the 2 things in on PR as long as we can get test cases to cover both individually :).

View from the mountain looks crazy to me. I ski, but on hills - never have been on a real mountain in winter!

src/TwigComponent/src/Twig/TwigPreLexer.php
smnandre: > But isn't this a valid syntax?
> 
> ```
> {% block my_embed_block_1 fooVariable %}
> ```

Indeed.  (https://twig.symfony.com/doc/3.x/tags/extends.html#block-shortcuts)

```twig
{% block title page_title|title %}
```




(but pretty sure this is not the [hardest problem](https://twig.symfony.com/doc/3.x/tags/use.html) we may _theoretically_ have while mixing twig/html  :)

Kocal: > Instead of keeping calculating the content between the positions, can we use the return value of `$this->consumeUntil('%}');`?

Hum you're right, I will give it a try.


> Also, in the test, you have
> 
> ```
> {% block my_embed_block_1 "foo" %}
> ```
> 
> I can understand that, after the `preg_replace()`, the `""` would remain. And so `isBlockSelfClosing` would be true. But isn't this a valid syntax?

This is not true, when running the regex [`/[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/A`](https://github.com/twigphp/Twig/blob/3.x/src/Lexer.php#L46) on `my_embed_block "foo"`, only the `my_embed_block` will be removed from the string, ` "foo"` will still be there.

See https://regex101.com/r/LVIP6p/1.

> ```
> {% block my_embed_block_1 fooVariable %}
> ```
> 
> In this case, everything would be replaced, the result would be an empty string, and `isBlockSelfClosing` would be false, right? Could we, instead, (A) grab the content between `{% block` and `%}`, `trim()` the whitespace, then set `isBlockSelfClosing` to true if there are ANY non-whitespace characters left?

Nope, ` fooVariable` will still be there after `preg_replace`, see https://regex101.com/r/FU3VmO/1.

Kocal:
I've applied your suggestions :)

> do you think you could extract this part in your code and try without any "embed" handling at first ? Or you do think both need to be fixed at once ?

@smnandre unfortunately removing the `$inTwigEmbed` logic makes the 4 tests "embeds + blocks" fail :/ But maybe I'm missing something

smnandre:
I looked a bit and it's almost perfect. But i think @weaverryan was right, and we are solving two problems at once that deserve independant solutions.

Here is a new test that fails (and probably not because of your code, it was just not tested before i guess)

```php
yield 'component_where_entire_default_block_is_twig_embed_with_endblock_name' => [
            <<<EOF
            <twig:Alert>
                <p>
                    {% embed "my_embed.html.twig" %}
                        {% block my_embed_block %}{% endblock my_embed_block %}
                    {% endembed %}
                </p>
            </twig:Alert>
            EOF,
            <<<EOF
            {% component 'Alert' %}
                {% block content %}<p>
                    {% embed "my_embed.html.twig" %}
                        {% block my_embed_block %}{% endblock my_embed_block %}
                    {% endembed %}
                </p>
            {% endblock %}{% endcomponent %}
            EOF,
        ];
```

Trying to solve it, and thinking about what i suggested about the embed isolation, i realized.... we should never be in that situation in the first place :)

A current component cannot have any impact on children embed content... So i think when we detect  an embed we should call the prelexer (as it's done in the block parsing)

But it was already there, so let's merge your fix and adress all this after :)

weaverryan:
Thanks for your work on this Hugo!

Kocal:
You're welcome :)

PR #1398, opened 2024-01-16
[Live] Control the type of generated URLs

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

Add a way to generate absolute URLs for live components.

My use case is that I embed a page with live components 
into a site on a different host, so I need absolute URLs 
for the components to work properly.

### DIFF
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\LiveComponent\Attribute;
 
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 
 /**
@@ -33,6 +34,7 @@ final class AsLiveComponent extends AsTwigComponent
      * @param string      $attributesVar     The name of the special "attributes" variable in the template
      * @param bool        $csrf              Whether to enable CSRF protection (default: true)
      * @param string      $route             The route used to render the component & handle actions (default: ux_live_component)
+     * @param int         $urlReferenceType  Which type of URL should be generated for the given route. Use the constants from UrlGeneratorInterface (default: absolute path, e.g. "/dir/file").
      */
     public function __construct(
         string $name = null,
@@ -43,6 +45,7 @@ final class AsLiveComponent extends AsTwigComponent
         public bool $csrf = true,
         public string $route = 'ux_live_component',
         public string $method = 'post',
+        public int $urlReferenceType = UrlGeneratorInterface::ABSOLUTE_PATH,
     ) {
         parent::__construct($name, $template, $exposePublicProps, $attributesVar);
 
@@ -64,6 +67,7 @@ final class AsLiveComponent extends AsTwigComponent
             'csrf' => $this->csrf,
             'route' => $this->route,
             'method' => $this->method,
+            'url_reference_type' => $this->urlReferenceType,
         ]);
     }
 
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -89,7 +89,7 @@ final class LiveComponentExtension extends Extension implements PrependExtension
             AsLiveComponent::class,
             function (ChildDefinition $definition, AsLiveComponent $attribute) {
                 $definition
-                    ->addTag('twig.component', array_filter($attribute->serviceConfig()))
+                    ->addTag('twig.component', array_filter($attribute->serviceConfig(), static fn ($v) => null !== $v && '' !== $v))
                     ->addTag('controller.service_arguments')
                 ;
             }
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -6,6 +6,8 @@
 -   Allow multiple `LiveListener` attributes on a single method.
 -   Requests to LiveComponent are sent as POST by default
 -   Add method prop to AsLiveComponent to still allow GET requests, usage: `#[AsLiveComponent(method: 'get')]`
+-   Add a new `urlReferenceType` parameter to `AsLiveComponent`, which allows to
+    generate different type URL (e.g. absolute) for the component Ajax calls.
 
 ## 2.13.2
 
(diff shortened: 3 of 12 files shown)


smnandre:
Seems to me a very specific use case.. could it be handled with router configuration ?

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\LiveComponent\Attribute;
 
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
```
smnandre, line 14: Not sure we should couple UrlGeneratorInterface and this attribute.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -89,7 +89,7 @@ public function load(array $configs, ContainerBuilder $container): void
             AsLiveComponent::class,
             function (ChildDefinition $definition, AsLiveComponent $attribute) {
                 $definition
-                    ->addTag('twig.component', array_filter($attribute->serviceConfig()))
+                    ->addTag('twig.component', array_filter($attribute->serviceConfig(), static fn ($v) => null !== $v))
```
smnandre: Is this related to this PR ?

1ed: Without this, it does not work. It's a separate commit, so it can be cherry-picked as a bugfix or it can be moved to its own PR if needed.

src/LiveComponent/src/Attribute/AsLiveComponent.php
1ed: We can invent or own constants and map them over, but the component already bound with the routing component.

1ed (commented):
> Seems to me a very specific use case.. could it be handled with router configuration ?

IDK maybe.. what do you have in mind?

If you suggesting to use the `host` parameter it doesn't work, because from Symfony's point of view the host is the same as the request comes in to.

src/LiveComponent/src/Attribute/AsLiveComponent.php
weaverryan: I agree - we're already bound to routing - so I'm ok with it

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
weaverryan: Is there an error without it? I actually can't remember why we have that `array_filter()` at all

weaverryan (commented):
This looks sensible to me!

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
1ed: If I remove the `array_filter` it breaks the compiler pass in TwigComponent...

https://github.com/symfony/ux/blob/a2418ef146f09a858bd12ed0116c3f8058d0a3a8/src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php#L48

and other places where `null` is not checked.

src/LiveComponent/src/Attribute/AsLiveComponent.php
smnandre: Allright then

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
smnandre: Yes but would that break things now that you allow empty strings ?

smnandre:
Wait, so you'll POST from domainA.com to domainB.com ? 

Should this be on a component - per - component basis... or a config wide one (that seem more logical to me) ?

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
1ed: I think it's not, all tests are green, but I can add `'' !== $v` too, if needed.

1ed:
> Wait, so you'll POST from domainA.com to domainB.com ?
> 
> Should this be on a component - per - component basis... or a config wide one (that seem more logical to me) ?

As the route can be configured on the component level, for me it is logical to do the same with the URL generated for that route.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
1ed: OK, I added back filtering out empty strings too.

smnandre:
> As the route can be configured on the component level, for me it is logical to do the same with the URL generated for that route.

Yes i mean, should this be possible to config it on an app-level too, as the route can be configured there for all components ? 

Last thing, is there any implication/effect with the "liveQuery" thing ?

1ed:
> > As the route can be configured on the component level, for me it is logical to do the same with the URL generated for that route.
> 
> Yes i mean, should this be possible to config it on an app-level too, as the route can be configured there for all components ?

Ah, I see. That would mean we need to introduce configuration for the live component and add a config entry for it, e.g. something like

```yaml
live_component:
    default_url_reference_type: 'absolute_path' # support strings and UrlGeneratorInterface consts too -> map strings back to consts
```

read the configuration and use this as the default value for the metadata (the default in `AsLiveComponent` should be null) e.g. in `LiveComponentMetadataFactory`. That can be done, maybe in a different PR.
 
> Last thing, is there any implication/effect with the "liveQuery" thing ?

I'm not sure, I'm not familiar with that, but I took a look at the PR and I think using the browser's history api and getting query string values from the request should not be affected by this.

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
smnandre: Man it took me so much time to understand i'm sorry .. 

I suggest that we handle that in another PR, as it's clearly not in LiveComponent that we should handle this.

https://github.com/symfony/ux/pull/1402

smnandre:
Could you add a test with an action (just to be sure everything work with in JS and in the request handling ? 

(and that way we'd have a regression test later on this)

1ed:
> Could you add a test with an action (just to be sure everything work with in JS and in the request handling ?
> 
> (and that way we'd have a regression test later on this)

I just added a test for that (but without js, for that we would need to setup panther, I think).

1ed:
I think this is ready. Can we merge it in this form or it needs more work?

weaverryan (approved):

weaverryan:
Thanks Gábor!

PR #1402, opened 2024-01-18
[TwigComponent] Filter service config in AsTwigComponent

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | almost
| New feature?  | no
| License       | MIT

As this attribute is currently extended by AsLiveComponent, it allow us to also call parent::serviceConfig in LiveComponent

(will require a conflict in composer to force Twig&Live update together)

### DIFF
--- a/src/TwigComponent/src/Attribute/AsTwigComponent.php
+++ b/src/TwigComponent/src/Attribute/AsTwigComponent.php
@@ -67,12 +67,12 @@ class AsTwigComponent
      */
     public function serviceConfig(): array
     {
-        return [
+        return array_filter([
             'key' => $this->name,
             'template' => $this->template,
             'expose_public_props' => $this->exposePublicProps,
             'attributes_var' => $this->attributesVar,
-        ];
+        ]);
     }
 
     /**
--- a/src/TwigComponent/tests/Unit/Attribute/AsTwigComponentTest.php
+++ b/src/TwigComponent/tests/Unit/Attribute/AsTwigComponentTest.php
@@ -21,6 +21,41 @@ use Symfony\UX\TwigComponent\Attribute\PreMount;
  */
 final class AsTwigComponentTest extends TestCase
 {
+    /**
+     * @dataProvider provideServiceConfigData
+     */
+    public function testServiceConfigValues(AsTwigComponent $attribute, array $expectedConfig): void
+    {
+        $this->assertSame($expectedConfig, $attribute->serviceConfig());
+    }
+
+    public static function provideServiceConfigData(): iterable
+    {
+        yield 'No values' => [
+            new AsTwigComponent(),
+            [
+                'expose_public_props' => true,
+                'attributes_var' => 'attributes',
+            ],
+        ];
+        yield 'Default values' => [
+             new AsTwigComponent(null, null, true, 'attributes'),
+             [
+                 'expose_public_props' => true,
+                 'attributes_var' => 'attributes',
+             ],
+        ];
+        yield 'Name and template set' => [
+             new AsTwigComponent('foo', 'template'),
+             [
+                'key' => 'foo',
+                'template' => 'template',
+                 'expose_public_props' => true,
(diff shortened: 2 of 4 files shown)


src/TwigComponent/tests/Unit/Attribute/AsTwigComponentTest.php
```diff
@@ -21,6 +21,41 @@
...
+                 'attributes_var' => 'attributes',
+             ],
+        ];
+        yield 'Name and template set' => [
+             new AsTwigComponent('foo', 'template'),
+             [
+                'key' => 'foo',
+                'template' => 'template',
+                 'expose_public_props' => true,
+                 'attributes_var' => 'attributes',
```
norkunas, line 54:
```suggestion
                'expose_public_props' => true,
                'attributes_var' => 'attributes',
```

src/TwigComponent/tests/Unit/Attribute/AsTwigComponentTest.php
```diff
@@ -21,6 +21,41 @@
...
+             [
+                 'expose_public_props' => true,
+                 'attributes_var' => 'attributes',
+             ],
+        ];
+        yield 'Name and template set' => [
+             new AsTwigComponent('foo', 'template'),
+             [
+                'key' => 'foo',
+                'template' => 'template',
```
norkunas, line 52:
```suggestion
                 'key' => 'foo',
                 'template' => 'template',
```

smnandre: Thanks.. i'll see why the CI did not get those..

smnandre: Well `array_indententation` is not in the Symfony rules https://cs.symfony.com/doc/rules/whitespace/array_indentation.html

src/TwigComponent/src/Attribute/AsTwigComponent.php
```diff
@@ -67,12 +67,12 @@ public function __construct(
...
     public function serviceConfig(): array
     {
-        return [
+        return array_filter([
             'key' => $this->name,
             'template' => $this->template,
             'expose_public_props' => $this->exposePublicProps,
             'attributes_var' => $this->attributesVar,
-        ];
+        ]);
```
1ed, line 75: Nitpicking, but I think the intent here should be to remove `null` and `''` values, but keep `0` as a valid value, so being explicit and doing `fn ($v) => null !== $v && '' !== $v)` would be better and less confusing.

smnandre: I (it took some time^^) understand why for LiveComponent... but we don't want to keep 0 for any of those 4 vars do we ?

1ed: It took some time me too to figure out why the stuff in my PR does not work. I'm just saying it could have been avoided if valid values like `0` not have been filtered out. Anyway, it doesn't really matter at the moment, with these values.

smnandre: To be honest, i'd like to remove all the array_filters. We needed it for a BC thing, maybe we can handle it differently.

And it's because you lost time with this (and me too after haha) that i suggest to put the array_filter with the vars it filters :) 

The second thing is... i'm more and more uncomfortable with the way LiveComponent uses TwigComponent... because none of it is covered by any contract or so...

smnandre: Honestly i did that to avoid that situation to happen again but if you think (and i see why) that would not, let's close this and focus back on yours ;)

weaverryan: Should this PR be closed then?

smnandre:
Let's close for now and adress this later

Issue #1047, opened 2023-08-09
[TwigComponent] Passing attributes to component blocks


sadikoff:
Hello,

What if we can pass attributes to blocks, for example I'd like to create complex Dropdown component with stimulus, so without Twig component I'll have following markup

```html
<div {{ stimulus_controller('dropdown', {'placement':'bottom'})>
  <button {{ stimulus_action('dropdown', 'toggle') }} class="button-classes">Open/Close</button>
  <div {{ stimulus_target('dropdown', 'block') class="some-classes">
      <!-- some menu items -->
  </div>
</div>
```

so it will be cool to have this component markup:

```
<twig:Dropdown placement="bottom">
  <twig:block name="toggler" class="button-classes" content="Open/Close"/>
  <twig:block name="dropdownBlock" class="some-classes">
     <!-- some menu items -->
  </twig:block>
</twig:Dropdown>
```

WDYT?

norkunas:
`<twig:block>` isn't an element by itself and would be replaced to anything that you pass to it, so you will need to provide `<div class="some-classes">` anyway :thinking:

sadikoff:
The main idea behind it is to allow to pass some attributes inside block so I can have predefined markup like dropdown which can be reused in multiple places and be pretty configurable

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Could I get an answer? If I do not hear anything I will assume this issue is resolved or abandoned. Please get back to me <3

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Just a quick reminder to make a comment on this. If I don't hear anything I'll close this.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #1405, opened 2024-01-18
[Twig] Allow nested attributes

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Ref #1047 (possible solution)
| License       | MIT

This PR adds a syntax for nested attributes: attributes that aren't meant to be used on the _root_ element but one of its _descendants_.

Here's an example:

```twig
{# MyDialog #}
<div{{ attributes }}>
    <div{{ attributes.nested('title') }}>
        {% block title %}Default Title{% endblock %}
    </div>
    <div{{ attributes.nested('body') }}>
    </div>
    <div{{ attributes.nested('footer') }}>
        {% block title %}Default Footer{% endblock %}
    </div>
</div>

{# render #}
<twig:MyDialog class="foo" body:class="bar" footer:class="baz">
   Some content
</twig:MyDialog>

{# output #}
<div class="foo">
    <div>
        Default Title
    </div>
    <div class="bar">
        Some content
    </div>
    <div class="baz">
        Default Footer
    </div>
</div>
```

The nesting is recursive so you could potentially do something like this:

```twig
<twig:Form
    :form="form"
    class="ui-form"
    row:class="ui-form-row"
    row:label:class="ui-form-label"
    row:widget:class="ui-form-widget"
/>
```

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.17.0
+
+-   Add nested attribute support #1405
+
 ## 2.16.0
 
 -   Introduce CVA to style TwigComponent #1416
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -1058,6 +1058,63 @@ Exclude specific attributes:
       My Component!
     </div>
 
+Nested Attributes
+~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.17
+
+    The Nested Attributes feature was added in TwigComponents 2.17.
+
+You can have attributes that aren't meant to be used on the *root* element
+but one of its *descendants*. This is useful for, say, a dialog component where
+you want to allow customizing the attributes of the dialog's content, title,
+and footer. Here's an example of this:
+
+.. code-block:: html+twig
+
+    {# templates/components/Dialog.html.twig #}
+    <div{{ attributes }}>
+        <div{{ attributes.nested('title') }}>
+            {% block title %}Default Title{% endblock %}
+        </div>
+        <div{{ attributes.nested('body') }}>
+            {% block content %}{% endblock %}
+        </div>
+        <div{{ attributes.nested('footer') }}>
+            {% block footer %}Default Footer{% endblock %}
+        </div>
+    </div>
+
+    {# render #}
+    <twig:Dialog class="foo" title:class="bar" body:class="baz" footer:class="qux">
+        Some content
+    </twig:MyDialog>
+
+    {# output #}
+    <div class="foo">
+        <div class="bar">
+            Default Title
+        </div>
+        <div class="baz">
+            Some content
+        </div>
+        <div class="qux">
+            Default Footer
+        </div>
+    </div>
+
+The nesting is recursive so you could potentially do something like this:
+
+.. code-block:: html+twig
+
+    <twig:Form
+        :form="form"
+        class="ui-form"
+        row:class="ui-form-row"
+        row:label:class="ui-form-label"
+        row:widget:class="ui-form-widget"
+    />
+
 Component with Complex Variants (CVA)
 -------------------------------------
 
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -19,8 +19,10 @@ use Symfony\WebpackEncoreBundle\Dto\AbstractStimulusDto;
  *
  * @immutable
  */
-final class ComponentAttributes implements \IteratorAggregate, \Countable
+final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Countable
 {
+    private const NESTED_REGEX = '#^([\w-]+):(.+)$#';
+
     /** @var array<string,true> */
     private array $rendered = [];
 
@@ -39,6 +41,10 @@ final class ComponentAttributes implements \IteratorAggregate, \Countable
                 fn (string $key) => !isset($this->rendered[$key])
             ),
             function (string $carry, string $key) {
+                if (preg_match(self::NESTED_REGEX, $key)) {
+                    return $carry;
+                }
+
                 $value = $this->attributes[$key];
 
                 if ($value instanceof \Stringable) {
@@ -192,6 +198,19 @@ final class ComponentAttributes implements \IteratorAggregate, \Countable
         $attributes = $this->attributes;
 
         unset($attributes[$key]);
+
+        return new self($attributes);
+    }
+
+    public function nested(string $namespace): self
+    {
+        $attributes = [];
+
+        foreach ($this->attributes as $key => $value) {
+            if (preg_match(self::NESTED_REGEX, $key, $matches) && $namespace === $matches[1]) {
+                $attributes[$matches[2]] = $value;
+            }
+        }
 
         return new self($attributes);
     }
--- a/src/TwigComponent/tests/Fixtures/templates/components/JustAttributes.html.twig
+++ b/src/TwigComponent/tests/Fixtures/templates/components/JustAttributes.html.twig
@@ -1,0 +1,1 @@
+<div{{ attributes }}/>
--- a/src/TwigComponent/tests/Fixtures/templates/components/NestedAttributes.html.twig
+++ b/src/TwigComponent/tests/Fixtures/templates/components/NestedAttributes.html.twig
@@ -1,0 +1,7 @@
+<main{{ attributes }}>
+    <div{{ attributes.nested('title') }}>
+        <span{{ attributes.nested('title').nested('span') }}>
+            {{ component('JustAttributes', [...attributes.nested('inner')]) }}
+        </span>
+    </div>
+</main>
--- a/src/TwigComponent/tests/Integration/ComponentExtensionTest.php
+++ b/src/TwigComponent/tests/Integration/ComponentExtensionTest.php
@@ -263,6 +263,84 @@ final class ComponentExtensionTest extends KernelTestCase
         $this->assertStringContainsString('class="alert alert-red alert-lg font-semibold rounded-md dark:bg-gray-600 flex p-4"', $output);
     }
 
+    public function testRenderingComponentWithNestedAttributes(): void
+    {
+        $output = $this->renderComponent('NestedAttributes');
+
+        $this->assertSame(<<<HTML
+            <main>
+                <div>
+                    <span>
+                        <div/>
+
+                    </span>
+                </div>
+            </main>
+            HTML,
+            trim($output)
+        );
+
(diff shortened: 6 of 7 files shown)


kbond:
I'm still thinking about how we could pass nested attributes down to nested components:

Something like:

```twig
{# Form #}
<form{{ attributes }}>
    <twig:Input :_attributes="attributes.nested('input')" />

    <!-- or -->

    <twig:Input ...attributes.nested('input') />
</form>
```

smnandre:
Oh that's a really nice feature ! 👏 

Does it work with live components ? (please say no hahaha)

kbond:
> Does it work with live components ?

I think it would actually, only the rendering of the attributes takes the "nested" syntax into account.

kbond:
> I'm still thinking about how we could pass nested attributes down to nested components:

Solved! 

```twig
<twig:Input {{ ...attributes.nested('input') }} />
```

smnandre:
Little question just to be curious, nothing about "right now" ;)

Would this be a lot of work to handle (same thing but in "json" notation instead of "html") ? 

```
<twig:Form
    :form="form"
    class="ui-form"
    :row:="{class: "ui-form-row", label: .., widget: .{class: ......}"
/>
```

kbond:
> Little question just to be curious, nothing about "right now" ;)

I think the current syntax looks _cleaner_ (just standard html attributes).

kbond:
Regarding the CI failure, it's because I'm using the twig spread operator in a test. I either need to bump the min twig version or skip this test if not supported. Thoughts?

smnandre:
Skip for now ? We will probably drop Twig < 3.something when Twig 4.0 is released right ?

weaverryan:
I've wondered about how to do this a bunch of times 😅. I had imagined it working via blocks someday:

```twig
<twig:MyDialog class="foo">
    <twig:block name="footer" class="baz">Footer</twig:block>
</twig:MyDialog>
```

That's a bit inspired by Blade, because our blocks == Blade slots, and they have slot attributes - https://laravel.com/docs/10.x/blade#slot-attributes

However, 3 issues I see are:

A) What if I wanted to define a `class` on the footer but not override the content? `<twig:block name="footer" class="baz" />`? And is that enough to differentiate between "I do not want to override this block content" vs other cases where "I want to override this block content to an empty string?"'.

B) Do we want to require a block in order to render an element with "nested" attributes?

C) The more obvious issue is: blocks are a core concept... so this may not even be possible unless we modify Twig itself. Though, I had imagined that we might cheat by doing a translation:


```twig
<twig:MyDialog class="foo">
    <twig:block name="footer" class="baz">Footer</twig:block>
</twig:MyDialog>

becomes (line breaks added for clarity)

{% component 'MyDialog' with { class: 'foo' } %}
    {% block footer %}
        {% do attributes.nested('footer', { class: 'baz' }) %}
        Footer
    {% endblock %}
{% endblock %}
```

The usage would be:

```twig
{# Modal.html.twig #}
<div{{ attributes }}>
    {% block footer %}
    <div{{ attributes.nested('footer') }}>
        {% block footer %}Default Footer{% endblock %}
    </div>
</div>
```

Not sure this is better, tbh. But I wanted to throw out this alternate universe before we decide on a syntax.

smnandre:
We could totally use a intermediary tag like "slot"

kbond:
> We could totally use a intermediary tag like "slot"

Wasn't the issue with this: we won't have the benefits of blocks? I think this was the reason we ended up abandoning a new slot type.

> But I wanted to throw out this alternate universe before we decide on a syntax.

The block attributes system is interesting but it wouldn't solve the initial reason for me creating this PR: [passing nested attributes to subcomponents](https://github.com/symfony/ux/pull/1405#issuecomment-1899328396).

What I like about this implementation is it doesn't require any new twig fanciness.

smnandre:
> Wasn't the issue with this: we won't have the benefits of blocks? I think this was the reason we ended up abandoning a new slot type.

Was'nt this before the HTML syntax ? I feel we have learned a lot about "hacking" things into Twig since, maybe there is more things "possible" than before ? 

Did you remember what specifically in the "benefits of blocks" was laking ? Or list some of those you count more on ?

kbond:
> Was'nt this before the HTML syntax ?

Possibly, I can't quite remember.

> Did you remember what specifically in the "benefits of blocks" was laking ? Or list some of those you count more on ?

I just remember Ryan and I realizing how powerful the out of the box block system was and all the benefits we'd be missing by doing our own thing.

weaverryan:
> I just remember Ryan and I realizing how powerful the out of the box block system was and all the benefits we'd be missing by doing our own thing.

Yup, the block system is massively powerful and we should keep using it.

I'm leaning towards "let's merge this & we could always deprecate it later). But `<twig:Foo body:class="bar"` isn't valid HTML... and so it sucks to land on a solution that's going to look wrong to editors (my editor also yells about `<twig:Foo`, but that is just because it's unaware of this `twig` namespace: it IS valid HTML).

> The block attributes system is interesting but it wouldn't solve the initial reason for me creating this PR: https://github.com/symfony/ux/pull/1405#issuecomment-1899328396.

```twig
<twig:Input {{ ...attributes.nested('input') }} />
```

That problem occurred to me also:

> A) What if I wanted to define a class on the footer but not override the content? <twig:block name="footer" class="baz" />? And is that enough to differentiate between "I do not want to override this block content" vs other cases where "I want to override this block content to an empty string?"'.

It could *possibly* work via:

```twig
<twig:Form>
    <twig:block name="widget" class="ui-form-widget" />
</twig:Form>
```

It would require defining a self-closing `block` tag as meaning "Do not actually override this block" vs "Override with null". The above would result in:

```twig
{% component 'Form' %}
    {% do attributes.nested('widget', { class: 'ui-form-widget' }) %}
{% endcomponent %}
```

Inside `Form.html.twig`, you'd have:

```
{% block widget %}<twig:Input {{ ...attributes.nested('widget') }} />{% endblock %}
```

On the one hand, this feels much crazier than your solution. On the other hand, this is a hard problem because it's not solved in Twig. And that's what we're trying to do here - ideally with the best API possible.

kbond:
> But <twig:Foo body:class="bar" isn't valid HTML...

Do you mean `body:class="bar"` here? Isn't `:class="bar"` just as invalid?

weaverryan:
> Do you mean `body:class="bar"` here? Isn't `:class="bar"` just as invalid?

Fair point, but at least in the eyes of PhpStorm, `:class` is ok. This may be the Symfony plugin at work (I have it disabled on this project, but still possible), but `:class` is at least a known syntax from other languages.

Screenshot 2024-01-30 at 8 39 51 AM

smnandre:
:class="bar" may be invalid.... But on custom elements, class:="bar" is not ;) 

```html
<!DOCTYPE html>
<html>
<head>
<title>Hello</title>
</head>
<body>
<twig-bar class:="foo"></twig-bar>
</body>
</html>
```

This is perfectly valid HTML code ;)

kbond:
For me, it looks like PhpStorm marks attributes with `:` in their name as invalid for standard html elements but not "custom components"

image

`class` autocompletion only appears to work for `class` (not `:class` or `widget:class`) which is a bit of a bummer.

Edit:

The official tailwindcss plugin allows adding custom class names (I can add "widget:class" to the "classAttributes" config and it works). There is an experimental regex option that I think we could make work for `*:class` but I haven't figured that out.

image

smnandre:
I think all this can be handled with IDE, syntax rules, Symfony plugin... and i personnaly prefer a syntax non final (different from the HTML produced) 

The only question should be: what is at risk if something not intended to is displayed in HTML. 
And for that we could also "just" add some Listener  (or a ux:component:lint) to check final output.


-- 

In a (not-so) distant future, i'm more and more convinced by using "custom web components" syntax instead of our current "grammar-free custom XML namespace") .... 

That would allow us reserved attribute, custom syntax rules, templates,... and would offer incredible possibilities in term of lazy load too ....  while beeing entirely supported in HTML/Twig editors

src/TwigComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.17.0
+
+-   Added nested attribute support.
```
smnandre, line 5: Add

```suggestion
-   Add nested attribute support. #1405
```

smnandre (approved):
I approve for you to merge :) But still nitpicking doc things we can send later

src/TwigComponent/doc/index.rst
```diff
@@ -1058,6 +1058,63 @@ Exclude specific attributes:
...
+
+.. versionadded:: 2.17
+
+    The Nested Attributes feature was added in TwigComponents 2.17.
+
+You can have attributes that aren't meant to be used on the *root* element
+but one of its *descendants*. This is useful for, say, a dialog component where
+you want to allow customizing the attributes of the dialog's content, title,
+and footer. Here's an example of this:
+
```
smnandre, line 1072: You can pass attributes from the outside directly to the descendants of a component. This approach is especially useful for components like dialogs, to configure the rendering of inner parts like content, title, and footer.  Here's an example:

src/TwigComponent/doc/index.rst
```diff
@@ -1058,6 +1058,63 @@ Exclude specific attributes:
...
+        </div>
+        <div class="baz">
+            Some content
+        </div>
+        <div class="qux">
+            Default Footer
+        </div>
+    </div>
+
+The nesting is recursive so you could potentially do something like this:
```
smnandre, line 1106: I'm on very small details but this is not really recursive and more a tree here (in the meaning a parent call child, but never call itself back)

This paragraph and your following example make me want to do two small TwigComponent demos 
* one litterally recursive where you pass to the child a level max and something a bit dynamic (oh i think i have n idea, poke me if your'e around) and the recursion goes until the level is reached
* a form demo with a cool theme, just like you did.

WDYT ? 


(and we need to start the TwigComponent demo)

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -19,8 +19,10 @@
  *
  * @immutable
  */
-final class ComponentAttributes implements \IteratorAggregate, \Countable
+final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Countable
```
smnandre, line 22: <3

smnandre (approved):
(Later on, that'be nice if we could add some more tests on this.. to ensure non-regressive later PR, and maybe one or two on "non-happy-path". I'd be glad to help if you want)

:)

WebMamba (approved):
LGTM! Thanks @kbond!

kbond:
Thanks! @smnandre, when time permits, mind making a followup PR with some of the doc changes/tests you'd like?

smnandre:
Will do in April... but won't mind a bit of help on this :))

Issue #940, opened 2023-06-14
[TwigComponent] A `HtmlTag` meta-component to easily renders dynamic HTML tag / attributes


Kocal:
Hi!

Sometimes, inside a Twig template, you need to render HTML with a dynamic tag or attributes, depending if variables/props are presents, depending of their values, etc... 

Given the following template:

```twig 
{% set tag = href ? 'a' : 'p' %}

<{{ tag }}
    {% if href %} href="{{ href }}"{% endif %}
    {% if title %} title="title"{% endif %}
    {% if newTab %} target="_blank"{% endif %}
>
    ...
</{{ tag }}>
```

We can see that it is not easy to render a dynamic tag (you must think about the closing tag too), and you can't re-use `attributes` nor use a new `ComponentAttributes` instance (because you are not inside a Twig Component) to easily render tags.

Currently you need to do it yourself and it can leads to inconsistency / bad readability.

With a new meta-component (which has no logic but to renders HTML), we can imagine passing a tag and HTML attributes as an object (think as a [simplified `h` function](https://vuejs.org/guide/extras/render-function.html#creating-vnodes)):
```twig
{% component HtmlTag with {
    _tag: href ? 'a' : 'p',
    href,
    title,
    ...(newTab ? { target: '_blank' } : {})
} %}
    ...
{% endcomponent %}
```
(_The `_` prefix in `_tag` is here to prevent a potential conflict with a `tag` HTML attribute, but maybe there is a better solution?_)

The template associated to this components would looks like this:
```twig
<{{_tag}} {{ attributes }}>{% block content %}{% endblock %}</{{_tag}}>
```

WDYT?
Thanks

WebMamba:
Yes, I like the idea! I remember that @norkunas was doing something similar I ping you here in case you are interest by the topics.

Kocal:
Or maybe we can simply use the HTML tag name as meta-component, but IDK how it can works if it's dynamic.

norkunas:
For simplicity in components I define as `string|false` for example `$href` and in template:
```twig
<{{ element }}{{ attributes.defaults({ href })>
  {% block content %}{% endblock %}
</{{ element }}>
```
So if href is not set it won't be included in output

kbond:
I think this will be pretty simple once #802 is available.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
[...]

PR #1413, opened 2024-01-22
[Twig] Make `ComponentAttributes` traversable/countable

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

This will help with #1405 and #1414.

PR #1414, opened 2024-01-23
[Twig] Add `attributes()` twig function

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

This allows passing attributes to elements easier.

```twig
<div{{ attributes({foo: 'bar'})>
```

When/if #1413 is merged, the following will also be possible:

```twig
<twig:Some:Component {{ ...attributes({foo: 'bar'}) }} />
```

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+-   Add `attributes()` twig function.
+
 ## 2.13.0
 
 -   [BC BREAK] Add component metadata to `PreMountEvent` and `PostMountEvent`
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -110,13 +110,27 @@ final class ComponentRenderer implements ComponentRendererInterface
         $this->dispatcher->dispatch($event);
     }
 
+    public function createAttributes(array $attributes = []): ComponentAttributes
+    {
+        $this->registerSafeClasses();
+
+        return new ComponentAttributes($attributes);
+    }
+
+    private function registerSafeClasses(): void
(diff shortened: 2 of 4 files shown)


kbond:
Similar twig core issues/PRs: https://github.com/twigphp/Twig/pull/3930, https://github.com/twigphp/Twig/issues/3907, https://github.com/twigphp/Twig/pull/3760

kbond:
The function name `attributes` is unfortunate as it can be confused with the attributes parameter in twig/live components:

```twig
<div{{ attributes }}> {# correct #}

<div{{ attributes() }}> {# wrong #}
```

`html_attr()` maybe?

Kocal:
That's a great idea, it would also close #940 since the main idea was to easily renders attributes

WebMamba (approved):
Love that 🧡

WebMamba:
> The function name attributes is unfortunate as it can be confused with the attributes parameter in twig/live components:

Yes, I agree here. I like `html_attr` but also `attr` as it's proposed in the Twig issue.

How it's gonna work with Live components for example?

```twig
<div attributes({role: 'alert', class='text-blue-800 px-4',  ...attributes})>
...
</div>
```

norkunas:
Would love to have this in Twig itself :)

leevigraham:
Implementation in CraftCMS / Yii2:

* https://craftcms.com/docs/5.x/reference/twig/filters.html#attr
* https://www.yiiframework.com/doc/api/2.0/yii-helpers-basehtml#renderTagAttributes()-detail

mpdude:
https://github.com/twigphp/Twig/pull/3930 is taking shape, so I'd like to hear if it would fit your needs as well. 🙏🏻

Kocal:
Replaced by https://github.com/twigphp/Twig/pull/3930

PR #1415, opened 2024-01-23
[Twig] Allow `string[]` for `ComponentAttribute` values

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

This allows the following:

```twig
{% set errors = false %}
{% set valid = true %}

<div{{ attributes.defaults({
    class: [
        'form-row',
        errors ? 'has-errors',
        valid ? 'is-valid',
    ],
}) }})

{# renders as: #}
<div class="form-row is-valid">
```

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -39,6 +39,11 @@ final class ComponentAttributes
             function (string $carry, string $key) {
                 $value = $this->attributes[$key];
 
+                if (null === $value) {
+                    trigger_deprecation('symfony/ux-twig-component', '2.8.0', 'Passing "null" as an attribute value is deprecated and will throw an exception in 3.0.');
+                    $value = true;
+                }
+
                 return match ($value) {
                     true => "{$carry} {$key}",
                     false => $carry,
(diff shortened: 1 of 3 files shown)


src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -21,29 +21,24 @@
  */
 final class ComponentAttributes
 {
+    private bool $normalized = false;
```
kbond, line 24: This normalization tracking is required because of how #1041 was implemented. (it first sets all passed values as attributes, then removes the ones that are defined)

norkunas:
I'd use [html_classes](https://twig.symfony.com/doc/3.x/functions/html_classes.html) for this case :)

kbond:
> I'd use [html_classes](https://twig.symfony.com/doc/3.x/functions/html_classes.html) for this case :)

Forgot about that function! We can close this if others agree using that function is best.

weaverryan:
Yup! I think the goal should be to make using `html_classes()` as easy as possible, and possibly document it, as this is a common need.

PR #1417, opened 2024-01-24
Run tests for all supported versions defined for peerDependencies per component

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Relates to #1389
| License       | MIT

Refactored the `run-vitest-all.sh` script to loop over every component/workspace and check if there are `peerDependencies` who defined support for multiple versions (like https://github.com/symfony/ux/blob/2.x/src/Svelte/assets/package.json#L23). The script then runs the tests for every defined version for that library in that workspace.

The greatest benefit of this change is that this needs no extra administration in keeping a `matrix` value up-to-date in the GitHub workflow for every library in a workspace that supports more than 1 version as it checks every `package.json` automatically.

Tell me what you think about it! Cheers!

- https://github.com/symfony/ux/actions/runs/7646573943/job/20835661388?pr=1417#step:6:200 -> `svelte@3.59.2` is installed
- https://github.com/symfony/ux/actions/runs/7646573943/job/20835661388?pr=1417#step:6:259 -> `svelte@4.2.9` is installed

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -143,5 +143,5 @@ jobs:
                   key: ${{ runner.os }}-yarn-${{ hashFiles('**/package.json') }}
                   restore-keys: |
                       ${{ runner.os }}-yarn-
-            - run: yarn --frozen-lockfile
+            - run: yarn --immutable
             - run: yarn test
--- a/bin/run-vitest-all.sh
+++ b/bin/run-vitest-all.sh
@@ -1,23 +1,72 @@
 #!/bin/bash
 
-# Get all workspace names
-workspaces=$(yarn workspaces info | grep -o '@symfony/[^"]*')
-
 # Flag to track if any test fails
 all_tests_passed=true
 
-for workspace in $workspaces; do
-  echo "Running tests in $workspace..."
+# Check if jq is installed
+if ! command -v jq &> /dev/null; then
+    echo "jq is required but not installed. Aborting."
+    exit 1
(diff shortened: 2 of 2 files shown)


weaverryan:
Fantastic! The shell script is now a bit complex, but this is very pragmatic, so let's roll with it. Thank you Evert!

evertharmeling:
Great! Yeah I thought it was the most pragmatic way to fix it and since there's lots of documentation on shell scripting, it should be doable to keep working on it. Let's see what the future brings :)

Only thing to keep in mind that the script currently checks on `||` (double pipe) for finding out if multiple versions are defined. So the format `^3.41 || ^4.0` is kinda mandatory within the repo or we should add support for alternative syntax (like single pipe).

Cheers!

smnandre:
Just a little thing @evertharmeling  ... it does not work for me on macOS :/ 

I'll try to fix what i can... but i may need to change/adapt minor things (will poke you)

evertharmeling:
Alright, what error do you get? I've also opened up #1433 where the boundary range notation caused problems. Let me know if I can help

EDIT: we could also add a flag to run the 'full version'-suite and by default stick to the old way, so in CI we could leverage the 'full version'.

smnandre:
No your changes are perfect... i just need to fix something (and some errors may come from my own zsh scripts so... ) 

Just poking you if someone else had the problem ;)

PR #1418, opened 2024-01-26
[Live] Rename data-action-name to use standard Stimulus features

### LINKED ISSUE #1065: [LiveComponent] Use keyboard events
Hello,
I actually develop for fun a little RPG game, and I need to move a character in a map using keyboard arrows. My challenge is to make it work with live component of course :D 

But I do not understand how to do it, as keypress events have to be attached to document, and not directly to the element, it do not find the data-action-name attached. Here the example : 

in my component
`<div id="map" {{attributes}} data-action="keydown.up@document->live#action" data-action-name="move(direction=N)">`

The event is "working", but when I press arrow up, I get this error : 
```
Error invoking action "keydown.up@document->live#action"
 TypeError: event.currentTarget.dataset is undefined
```
do you have any idea to make it work ? thanks for your help :-)
And if you could make it easier to use keyboard events, it could be very useful to make some cool interactions :)

### LINKED ISSUE #1283: [LiveComponent] LiveAction argument do not allow a closing parenthesis
Hi! I have a component where I list some strings and a user selects one of them.

I found that if I have a Live Action argument with a closing parenthesis like this
```html
<button type="button" data-action="live#action" data-action-name="myAction(foo=bar(123))">Action</button>
```
it throws the error: [`Missing space after myAction`](https://github.com/symfony/ux-live-component/blob/67372d192aeb8a4216bbc6f64ba34e438afe7cec/assets/src/Directive/directives_parser.ts#L224) due to [this](https://github.com/symfony/ux-live-component/blob/67372d192aeb8a4216bbc6f64ba34e438afe7cec/assets/src/Directive/directives_parser.ts#L183).

I was expecting to enclose the argument values with single or double quotes but checking the directive parser it seems this is not supported (_yet?_):
```html
<button type="button" data-action="live#action" data-action-name="myAction(foo='bar(123)')">Action</button>
```

To workaround this I'm encoding the strings in hex (with a custom Twig filter) and decoding in the component function.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1283 Fix #1065
| License       | MIT

Hi!

The `data-action-name` was flawed because the parser inside couldn't handle things like `,` or `)` - e.g. `data-action-name="save(I am a string with ) inside of it)"`

One solution was to write a giant parser to parse these arguments correctly. Instead, in a live stream, some of us decided to "just use Stimulus":

```diff
    <button
        data-action="live#action"
    -    data-action-name="debounce(300)|save"
    +    data-live-action-param="debounce(300)|save"
    >Save</button>
```

This uses [Stimulus action parameters](https://stimulus.hotwired.dev/reference/actions#action-parameters). The new syntax is basically the same length as the previous one, but it's pure Stimulus. Passing arguments is also slightly different, as they are a query string (this allows us to have special characters in the argument values without inventing our own syntax). For single arguments, things look like before (other than the new attribute name):

```diff
<button
    data-action="live#action"
-     data-action-name="addItem(id={{ item.id }}, itemName=CustomItem)"
+     data-live-action-param="addItem"
+     data-live-id-param="{{ item.id }}"
>Add Item</button>
```

This also removes the `|prevent` modifier... because... again ❗  Stimulus already does this:

```diff
<button
-    data-action="live#action
+    data-action="live#action:prevent"
-    data-action-name="prevent|save"
+    data-live-action-param="save"
 >Save</button>
```

Cheers!

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,51 @@
 # CHANGELOG
 
+## 2.16.0
+
+-   [BC BREAK] The `data-action-name` attribute behavior was removed in favor of
+    using Stimulus "action parameters" and `data-live-action-param`. This is a
+    breaking change if you were using the `data-action-name` attribute directly
+    in your templates.
+
+    To upgrade your application, follow these changes:
+
+    ```diff
+    <button
+        data-action="live#action"
+    -    data-action-name="debounce(300)|save"
+    +    data-live-action-param="debounce(300)|save"
+    >Save</button>
+    ```
+
+    To pass arguments to an action, also use the Stimulus "action parameters" syntax:
+
+    ```diff
+    <button
+        data-action="live#action"
+    -     data-action-name="addItem(id={{ item.id }}, itemName=CustomItem)"
+    +     data-live-action-param="addItem"
+    +     data-live-id-param="{{ item.id }}"
+    +     data-live-item-name-param="CustomItem"
+    >Add Item</button>
+    ```
+
+    Additionally, the `prevent` modifier (e.g. `prevent|save`) was removed. Replace
+    this with the standard Stimulus `:prevent` action option:
+
+    ```diff
+    <button
+    -    data-action="live#action
+    +    data-action="live#action:prevent"
+    -    data-action-name="prevent|save"
+    +    data-live-action-param="save"
+     >Save</button>
+    ```
+
+-   [BC BREAK] The `data-event` attribute was removed in favor of using Stimulus
+    "action parameters": rename `data-event` to `data-live-event-param`. Additionally,
+    if you were passing arguments to the event name, use action parameter attributes
+    for those as well - e.g. `data-live-foo-param="bar"`.
+
 ## 2.15.0
 
 -   [BC BREAK] The `data-live-id` attribute was changed to `id` #1484
--- a/src/LiveComponent/assets/dist/Directive/directives_parser.d.ts
+++ b/src/LiveComponent/assets/dist/Directive/directives_parser.d.ts
@@ -5,7 +5,6 @@ export interface DirectiveModifier {
 export interface Directive {
     action: string;
     args: string[];
-    named: any;
     modifiers: DirectiveModifier[];
     getString: {
         (): string;
--- a/src/LiveComponent/assets/dist/live_controller.d.ts
+++ b/src/LiveComponent/assets/dist/live_controller.d.ts
@@ -95,10 +95,10 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
     disconnect(): void;
     update(event: any): void;
     action(event: any): void;
-    emit(event: Event): void;
-    emitUp(event: Event): void;
-    emitSelf(event: Event): void;
     $render(): Promise<import("./Backend/BackendResponse").default>;
+    emit(event: any): void;
+    emitUp(event: any): void;
+    emitSelf(event: any): void;
     $updateModel(model: string, value: any, shouldRender?: boolean, debounce?: number | boolean): Promise<import("./Backend/BackendResponse").default>;
     propsUpdatedFromParentValueChanged(): void;
     fingerprintValueChanged(): void;
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -6,10 +6,8 @@ function parseDirectives(content) {
         return directives;
     }
     let currentActionName = '';
-    let currentArgumentName = '';
     let currentArgumentValue = '';
     let currentArguments = [];
-    let currentNamedArguments = {};
     let currentModifiers = [];
     let state = 'action';
     const getLastActionName = function () {
@@ -25,52 +23,30 @@ function parseDirectives(content) {
         directives.push({
             action: currentActionName,
             args: currentArguments,
-            named: currentNamedArguments,
             modifiers: currentModifiers,
             getString: () => {
                 return content;
             }
         });
         currentActionName = '';
-        currentArgumentName = '';
         currentArgumentValue = '';
         currentArguments = [];
-        currentNamedArguments = {};
         currentModifiers = [];
         state = 'action';
     };
     const pushArgument = function () {
-        const mixedArgTypesError = () => {
-            throw new Error(`Normal and named arguments cannot be mixed inside "${currentActionName}()"`);
-        };
-        if (currentArgumentName) {
-            if (currentArguments.length > 0) {
-                mixedArgTypesError();
-            }
-            currentNamedArguments[currentArgumentName.trim()] = currentArgumentValue;
-        }
-        else {
-            if (Object.keys(currentNamedArguments).length > 0) {
-                mixedArgTypesError();
-            }
-            currentArguments.push(currentArgumentValue.trim());
-        }
-        currentArgumentName = '';
+        currentArguments.push(currentArgumentValue.trim());
         currentArgumentValue = '';
     };
     const pushModifier = function () {
         if (currentArguments.length > 1) {
             throw new Error(`The modifier "${currentActionName}()" does not support multiple arguments.`);
         }
-        if (Object.keys(currentNamedArguments).length > 0) {
-            throw new Error(`The modifier "${currentActionName}()" does not support named arguments.`);
-        }
         currentModifiers.push({
             name: currentActionName,
             value: currentArguments.length > 0 ? currentArguments[0] : null,
         });
         currentActionName = '';
-        currentArgumentName = '';
         currentArguments = [];
         state = 'action';
     };
@@ -102,11 +78,6 @@ function parseDirectives(content) {
                 }
                 if (char === ',') {
                     pushArgument();
-                    break;
-                }
-                if (char === '=') {
-                    currentArgumentName = currentArgumentValue;
-                    currentArgumentValue = '';
                     break;
                 }
                 currentArgumentValue += char;
@@ -325,7 +296,7 @@ function getAllModelDirectiveFromElements(element) {
     }
     const directives = parseDirectives(element.dataset.model);
     directives.forEach((directive) => {
-        if (directive.args.length > 0 || directive.named.length > 0) {
+        if (directive.args.length > 0) {
             throw new Error(`The data-model="${element.dataset.model}" format is invalid: it does not support passing arguments to the model.`);
         }
         directive.action = normalizeModelName(directive.action);
@@ -342,7 +313,7 @@ function getModelDirectiveFromElement(element, throwOnMissing = true) {
         if (formElement && 'model' in formElement.dataset) {
             const directives = parseDirectives(formElement.dataset.model || '*');
             const directive = directives[0];
-            if (directive.args.length > 0 || directive.named.length > 0) {
+            if (directive.args.length > 0) {
                 throw new Error(`The data-model="${formElement.dataset.model}" format is invalid: it does not support passing arguments to the model.`);
             }
             directive.action = normalizeModelName(element.getAttribute('name'));
@@ -2942,15 +2913,18 @@ class LiveControllerDefault extends Controller {
         this.updateModelFromElementEvent(event.currentTarget, null);
     }
     action(event) {
-        const rawAction = event.currentTarget.dataset.actionName;
+        const params = event.params;
+        if (!params.action) {
+            throw new Error(`No action name provided on element: ${getElementAsTagText(event.currentTarget)}. Did you forget to add the "data-live-action-param" attribute?`);
+        }
+        const rawAction = params.action;
+        const actionArgs = Object.assign({}, params);
+        delete actionArgs.action;
         const directives = parseDirectives(rawAction);
         let debounce = false;
         directives.forEach((directive) => {
             let pendingFiles = {};
             const validModifiers = new Map();
-            validModifiers.set('prevent', () => {
-                event.preventDefault();
-            });
             validModifiers.set('stop', () => {
                 event.stopPropagation();
             });
@@ -2985,11 +2959,14 @@ class LiveControllerDefault extends Controller {
                 }
                 delete this.pendingFiles[key];
             }
-            this.component.action(directive.action, directive.named, debounce);
+            this.component.action(directive.action, actionArgs, debounce);
             if (getModelDirectiveFromElement(event.currentTarget, false)) {
                 this.pendingActionTriggerModelElement = event.currentTarget;
             }
         });
+    }
+    $render() {
+        return this.component.render();
     }
     emit(event) {
         this.getEmitDirectives(event).forEach(({ name, data, nameMatch }) => {
@@ -3006,9 +2983,6 @@ class LiveControllerDefault extends Controller {
             this.component.emitSelf(name, data);
         });
     }
-    $render() {
-        return this.component.render();
-    }
     $updateModel(model, value, shouldRender = true, debounce = true) {
         return this.component.set(model, value, shouldRender, debounce);
     }
@@ -3019,11 +2993,13 @@ class LiveControllerDefault extends Controller {
         this.component.fingerprint = this.fingerprintValue;
     }
     getEmitDirectives(event) {
-        const element = event.currentTarget;
-        if (!element.dataset.event) {
-            throw new Error(`No data-event attribute found on element: ${getElementAsTagText(element)}`);
+        const params = event.params;
+        if (!params.event) {
+            throw new Error(`No event name provided on element: ${getElementAsTagText(event.currentTarget)}. Did you forget to add the "data-live-event-param" attribute?`);
         }
-        const eventInfo = element.dataset.event;
+        const eventInfo = params.event;
+        const eventArgs = Object.assign({}, params);
+        delete eventArgs.event;
         const directives = parseDirectives(eventInfo);
         const emits = [];
         directives.forEach((directive) => {
@@ -3039,7 +3015,7 @@ class LiveControllerDefault extends Controller {
             });
             emits.push({
                 name: directive.action,
-                data: directive.named,
+                data: eventArgs,
                 nameMatch,
             });
         });
--- a/src/LiveComponent/assets/src/Directive/directives_parser.ts
+++ b/src/LiveComponent/assets/src/Directive/directives_parser.ts
@@ -26,10 +26,6 @@ export interface Directive {
      */
     args: string[];
     /**
-     * An object of named arguments
-     */
-    named: any;
-    /**
      * Any modifiers applied to the action
      */
     modifiers: DirectiveModifier[];
@@ -41,17 +37,8 @@ export interface Directive {
  * into an array of directives, with this format:
  *
  *      [
- *          { action: 'addClass', args: ['foo'], named: {}, modifiers: [] },
- *          { action: 'removeAttribute', args: ['bar'], named: {}, modifiers: [] }
- *      ]
- *
- * This also handles named arguments
- *
- *      save(foo=bar, baz=bazzles)
- *
- * Which would return:
- *      [
- *          { action: 'save', args: [], named: { foo: 'bar', baz: 'bazzles }, modifiers: [] }
+ *          { action: 'addClass', args: ['foo'], modifiers: [] },
+ *          { action: 'removeAttribute', args: ['bar'], modifiers: [] }
  *      ]
  *
  * @param {string} content The value of the attribute
@@ -64,10 +51,8 @@ export function parseDirectives(content: string|null): Directive[] {
     }
 
     let currentActionName = '';
-    let currentArgumentName = '';
     let currentArgumentValue = '';
     let currentArguments: string[] = [];
-    let currentNamedArguments: any = {};
     let currentModifiers: { name: string, value: string | null }[] = [];
     let state = 'action';
 
@@ -86,7 +71,6 @@ export function parseDirectives(content: string|null): Directive[] {
         directives.push({
             action: currentActionName,
             args: currentArguments,
-            named: currentNamedArguments,
             modifiers: currentModifiers,
             getString: () => {
                 // TODO - make a string representation of JUST this directive
@@ -95,36 +79,15 @@ export function parseDirectives(content: string|null): Directive[] {
             }
         });
         currentActionName = '';
-        currentArgumentName = '';
         currentArgumentValue = '';
         currentArguments = [];
-        currentNamedArguments = {};
         currentModifiers = [];
         state = 'action';
     }
     const pushArgument = function() {
-        const mixedArgTypesError = () => {
-            throw new Error(`Normal and named arguments cannot be mixed inside "${currentActionName}()"`)
-        }
-
-        if (currentArgumentName) {
-            if (currentArguments.length > 0) {
-                mixedArgTypesError();
-            }
-
-            // argument names are also trimmed to avoid space after ","
-            // "foo=bar, baz=bazzles"
-            currentNamedArguments[currentArgumentName.trim()] = currentArgumentValue;
-        } else {
-            if (Object.keys(currentNamedArguments).length > 0) {
-                mixedArgTypesError();
-            }
-
-            // value is trimmed to avoid space after ","
-            // "foo, bar"
-            currentArguments.push(currentArgumentValue.trim());
-        }
-        currentArgumentName = '';
+        // value is trimmed to avoid space after ","
+        // "foo, bar"
+        currentArguments.push(currentArgumentValue.trim());
         currentArgumentValue = '';
     }
 
@@ -133,16 +96,11 @@ export function parseDirectives(content: string|null): Directive[] {
             throw new Error(`The modifier "${currentActionName}()" does not support multiple arguments.`)
         }
 
-        if (Object.keys(currentNamedArguments).length > 0) {
-            throw new Error(`The modifier "${currentActionName}()" does not support named arguments.`)
-        }
-
         currentModifiers.push({
             name: currentActionName,
             value: currentArguments.length > 0 ? currentArguments[0] : null,
         });
         currentActionName = '';
-        currentArgumentName = '';
         currentArguments = [];
         state = 'action';
     }
@@ -192,14 +150,6 @@ export function parseDirectives(content: string|null): Directive[] {
                 if (char === ',') {
                     // end of current argument
                     pushArgument();
-
-                    break;
-                }
-
-                if (char === '=') {
-                    // this is a named argument!
-                    currentArgumentName = currentArgumentValue;
-                    currentArgumentValue = '';
 
                     break;
                 }
--- a/src/LiveComponent/assets/src/dom_utils.ts
+++ b/src/LiveComponent/assets/src/dom_utils.ts
@@ -141,7 +141,7 @@ export function getAllModelDirectiveFromElements(element: HTMLElement): Directiv
     const directives = parseDirectives(element.dataset.model);
 
     directives.forEach((directive) => {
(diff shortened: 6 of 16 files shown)


sukei:
The weird thing here is that the action name to invoke is treated equally as its own arguments preventing the developer to use any argument named "action".

I was thinking of something like that:

```
data-live-action-param="addItem(id, itemName)"
data-live-args-param="escaped JSON string with id and itemName values"
```

The rationale here is to allow a rich action syntax and use an other parameter to prevent "injections" in the action DSL. Just like SQL or GraphQL prevents the same issue.

Then we could create a really cool Twig helper to handle this for us.

```
{{ live_action("addItem(id, itemName)", { id, itemName }) }}
```

On top of that, I would keep the "prevent" into the action DSL to not having to mix native Stimulus syntax and Symfony Live syntax (e.g. prevent + debounce).

weaverryan:
> The weird thing here is that the action name to invoke is treated equally as its own arguments preventing the developer to use any argument named "action".

That's true. But I'm not sure we should care. Having some reserved words is a pretty normal thing. And in this case, there is just one reserved word: `action`.

> data-live-action-param="addItem(id, itemName)"
> data-live-args-param="escaped JSON string with id and itemName values"

I'm not against allowing this as an option - it seems perfectly acceptable to me. But I don't want to require people to write JSON or use a helper to write JSON. So if people want this, someone can make a PR :). It would even be a way for you to have an argument named `action`.

> {{ live_action("addItem(id, itemName)", { id, itemName }) }}

Yea! Why not? It'd be optional, but it looks nice. Or even:

```twig
{{ live_action('addItem', {
    id: theId,
    itemName: 'Foo'
}) }}
```

Someone should make a PR for that :). And, btw, if we did this, it could render using the `data-live-id-param` format - it wouldn't need the JSON format.

> On top of that, I would keep the "prevent" into the action DSL to not having to mix native Stimulus syntax and Symfony Live syntax (e.g. prevent + debounce).

Nah. I mean, I understand why this might make sense. But we should lean on Stimulus as much as possible so that LiveComponent is as much Stimulus and a little "custom LiveComponents stuff" as possible.

Cheers!

sukei:
Hello! I made some tests on my side and what I found raised some more concerns. I'm not on my phone anymore, so I will explain that thing a little bit.

# 1. The inflection of the parameters names

The parameters names are inflected from `dash-case` to `lowerCamelCase` so you can't enforce the final form of the parameters:

```
<button type="button"
        data-action="live#action"
        data-live-action-param="addItem"
        data-live-item-id-param="42"
        data-live-item-name-param="foo">add to cart</button>
```

Stimulus will produce the parameters `itemId` and `itemName`. My PHP project may use `$item_id` and `$item_name` as parameters or ; in my own case ; `$itemID`. The mapping between the two worlds seems impossible using the Stimulus way.

# 2. Parameters types

Stimulus uses the static `values` property to guess the types of passed values (e.g. `static values { data: Number }`) and map the values accordingly. This does not exists for the action parameters. Every type is guessed from the form of the value:

```
<button type="button"
        data-action="live#action"
        data-live-action-param="dumpValues"
        data-live-one-param="42"
        data-live-two-param="foo"
        data-live-three-param="true"
        data-live-four-param="[1, 2]"
        data-live-five-param='{"foo": "bar"}'></button>
```

Stimulus will interpret this as `one = number, two = string, three = boolean, four = JSON, five = JSON`. The types of the values may be obvious but what if the parameter `one` or `three` is a string on the PHP side? I think we lose the opportunity here to keep some types doing it that way.

# 3. The Stimulus philosophy

I think that Stimulus really shine when used as a thin bridge between heavy (or not that heavy) JS libraries and HTML. I may not use it as a "framework" to develop things on top of it. Once we said that, the `LiveComponent` should not be built as a "Stimulus Controller" but should be a different proposition THAT USE Stimulus as a bridge to make it works.

# 4. Parameter name collision

As I said previously, adding reserved or magic words is not ideal. This was already discussed.

# Proposal war

Here is what you propose:

```
<button type="button"
        data-action="live#action:prevent"
        data-live-action-param="debounce(300)|addItem"
        data-live-item-id-param="42"
        data-live-item-name-param="foo"
>add to cart</button>
```

1. The parameters names convention on the PHP side is enforced by Stimulus.
2. The parameters types are what Stimulus think they are.
3. It separate modifiers across different philosopies (prevent and debounce are declared separately)
4. The `action` parameter name is unusable since its now a reserved word.

On the other side, my original proposal fix all the concerns I raised above. I would consider the `action` parameter as a LiveComponent DSL (like a SQL prepared statement) and the `args` parameter separately (like a SQL statement bound parameters):

```
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID, itemName)"
        data-live-args-param='{ "itemID": 42, "itemName": "foo" }'
        >add to cart</button>
```

This allow the DSL to be rich, powerful and open for extension (think about allowing the developer to create its own modifiers).
This prevent the DSL injections we are now sensitive to (it may even be considered as a security issue at this point).

1. The parameters names are preserved.
2. The parameters types are preserved (while they are serializable: int, float, string, bool, array, stdClass).
3. It do not try to blend the two worlds and keep things collocated.
5. No more parameter names collisions.

## DX

Disclaimer: I'm just talking about my own experience here.

I found Stimulus attributes horrible to write and maintain, especially when using folders to organize many controllers. The Twig helpers really change my mind and the way I use it daily. I really do like writing:

```
<div {{ stimulus_controller('shopping/cart') }}>
  <button type="button" {{ stimulus_action('shopping/cart', 'addItem', 'click', { itemId: 42, itemName: 'foo' }) }}>add to cart</button>
</div>
```

over:

```
<div data-controller="shopping--cart">
  <button type="button"
          data-action="click->shopping--cart#addItem"
          data-shopping--cart-item-id-param="42"
          data-shopping--cart-item-name-param="foo">add to cart</button>
</div>
```

The `--` dash, the `really-long-names`, and the inflection rules really grind my gears and I'm glad that Symfony handle this for me. Think about the controller named `ui/form/input_autosize_controller.ts` that became  the identifier `ui--form--input-autosize` and blend in the other attributes conventions (`data-ui--form--input-autosize-max-size-value`). 

You were concerned about the need to write JSON as attribute parameters. It may be harder without the Twig helpers, but if you want to send arrays or objects, you'll not have choice. Stimulus has some examples on its documentation and use single quote to ease the writing of the JSON attributes. The helper I may want to use will looks like:

```
<button type="button" {{ live_action("prevent|debounce(300)|addItem(itemID, itemName)", { itemID: 42, itemName: 'foo' }) }}>add to cart</button>
```

May produce:

```
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID, itemName)"
        data-live-args-param="&#x7B;&quot;itemID&quot;&#x3A;42,&quot;itemName&quot;&#x3A;&quot;foo&quot;&#x7D;"
        >add to cart</button>
```

## PS

If this proposal is approved, I would be glad to work on it.

weaverryan:
You've made some very compelling points :). Especially about the inflection of the parameter names.

So let's explore your idea a bit more (I'm ignoring the Twig `{{ live_action() }}` helper just to keep things simple):

```
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID, itemName)"
        data-live-args-param="&#x7B;&quot;itemID&quot;&#x3A;42,&quot;itemName&quot;&#x3A;&quot;foo&quot;&#x7D;"
>add to cart</button>
```

You're separating the param names - `itemId` and `itemName` from their values to follow the idea of a prepared statement in SQL. But, is that necessary? In SQL, I might have:

> # DO NOT DO AT HOME ;) 
> $sql = "SELECT * FROM product WHERE id = $id";

The `$id` becomes part of the query string before it's executed. Suppose we combined the params and their values:

```
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID=7, itemName='banana')"
>add to cart</button>
```

I don't think we have that same problem. The string `banana` is not a variable that will be evaluated and become part of the attribute value. Even if we decided to add some sort of dynamic variable ability later - e.g. `data-live-action-param="addItem($itemName)` (where `$itemName` fetches the `itemName` prop/model value) - I still think it's not a problem, right?

My point is:

* I was using action parameters to avoid needing a parser for the argument values. But, you've highlighted some problems with this approach.
* If we don't use that approach then (unless I'm missing the "injections" problem), then we might as well write a parser and have the nice `save(itemId=5, itemName='banana')` syntax.

Or, maybe we live with the limitation of not using quotes and then not allowing `'` and `'(` in values. Or, we could use a syntax like `itemId: 5, itemName: 'banana'` and parse that as JavaScript (add `{` at the front and `}` at the end and use some version of `eval()` in JS). Unfortunately, that wouldn't be CSP-compliant.

I'm not entirely sure of the best path forward. Making the syntax use JSON - `save("itemId": 5, "itemName": "banana")` would be possible, but ugly... especially since you're probably already using double quotes for your attributes.

sukei:
We can't ignore the fact that some of the parameters are first extrapolated from a variable on the PHP side. Let me explain it further. Given:

```twig
{# Variables are passed to the template and their origins are unsure (DB, Form, User, etc.).  #}
{% set itemID = 7 %}
{% set itemName = 'banana' %}

<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID={{ itemID }}, itemName='{{ itemName }}')"
>add to cart</button>
```

Then, it'll be compiled to:

```twig
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID=7, itemName='banana')"
>add to cart</button>
```

But what about a variable with special characters?

```twig
{# Variables are passed to the template and their origins are unsure (DB, Form, User, etc.).  #}
{% set itemID = 7 %}
{% set itemName = 'cherry\')" data-ignore="' %}

<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID={{ itemID }}, itemName='{{ itemName }}')"
>add to cart</button>
```

Then, it'll be compiled to:

```twig
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID=7, itemName='cherry')" data-ignore="')"
>add to cart</button>
```

Note that it is just a POC and that Twig may mitigate the issue by escaping values someway into the process, but still, it can break the expression.

Trying to remove this from the API is just a way to have a easier implementation that prevent us to work on the double escaping of quotes (inside the expression) and double quotes (of the surrounding attribute). If I ignore the eventual helper, I would write:

```twig
{# Variables are passed to the template and their origins are unsure (DB, Form, User, etc.).  #}
{% set itemID = 7 %}
{% set itemName = 'cherry\')" data-ignore="' %}

<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID, itemName)"
        data-live-args-param="{{ { itemID, itemName }|json_encode|e('html_attr') }}"
>add to cart</button>
```

As a bonus, I should probably go deeper in the reasoning and escaping every var to completely disallow the usage of literals:

```twig
{# Variables are passed to the template and their origins are unsure (DB, Form, User, etc.).  #}
{% set itemID = 7 %}
{% set itemName = 'cherry\')" data-ignore="' %}

<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(t)|addItem(itemID, itemName)"
        data-live-args-param="{{ { itemID, itemName, t: 300 }|json_encode|e('html_attr') }}"
>add to cart</button>
```

src/LiveComponent/CHANGELOG.md
```diff
@@ -2,6 +2,49 @@
...
+
+    ```diff
+    <button
+        data-action="live#action"
+    -     data-action-name="addItem(id={{ item.id }}, itemName=CustomItem)"
+    +     data-live-action-param="addItem"
+    +     data-live-id-param="{{ item.id }}"
+    +     data-live-item-name-param="CustomItem"
+    >Add Item</button>
+    ```
```
WebMamba, line 30: Maybe I am missing something but it looks like by doing this change we losing de debounce. Maybe we should also show a good way to keep the debounce?

weaverryan: ~~Thanks - that was accidental. The diff should still include the `data-live-action-param="debounce|(300)addItem". I'll fix that when I figure out the direction for the PR :)~~

Edit: actually, it's fine! In the first example, I show `debounce` before and after. The second example doesn't have `debounce` before or after.

weaverryan:
> We can't ignore the fact that some of the parameters are first extrapolated from a variable on the PHP side. Let me explain it further.

Ah, I see - good point. However, I think it's not strictly a security concern. Twig escapes your example, and it outputs as:

```html
<button type="button"
        data-action="live#action"
        data-live-action-param="prevent|debounce(300)|addItem(itemID=7, itemName='cherry&#039;)&quot; data-ignore=&quot;')"
>add to cart</button>
```

I'm not against having `data-live-args-param="{{ { itemID, itemName }|json_encode|e('html_attr') }}"` type of option, but I'd like a version of this feature that works in raw HTML without needing to write JSON by hand (then we can add fancy helper functions after, which I think is a good idea).

Unfortunately, using a JS syntax - e.g. `data-live-action-param="save(id: 5, name: 'Ryan')"` *would* work... except we'd need to use `eval()` and that'll cause problems with csp. 

So what about a simple query string?

```html
<button type="button"
    data-action="live#action"
    data-live-action-param="debounce(300)|addItem(itemId=7&itemName=cherry)"
>add to cart</button>
```

The values would be url-encoded. In most simple cases, that wouldn't mean any change (like the above). But if you DO have a special character, it's supported - e.g. a complex value of `i like = (cherries) & bananas`

```html
<button type="button"
    data-action="live#action"
    data-live-action-param="debounce(300)|addItem(itemId=7&itemName=i%20like%20%3D%20%28cherries%29%20%26%20bananas)"
>add to cart</button>
```

Of course, some optional `live_action()` helper could hide this urlencoding.

Query params give us a (1) readable value (2) value that supports all characters (3) a recognizable format and (4) zero complex / extra code to parse it.

sukei:
> Ah, I see - good point. However, I think it's not strictly a security concern. Twig escapes your example, and it outputs as:

Yeah, you are right. I pointed it out too.

> we'd need to use eval() and that'll cause problems with csp

Yep, I don't want that too.

> The values would be url-encoded.

That's an interesting idea. I just fear that complex values such as objects and arrays may be hard to write. But it can be abstracted by the Twig helper so it may not be a big deal.

```twig
<button type="button"
    data-action="live#action"
    data-live-action-param="debounce(300)|addItem(command[itemID]=7&command[itemName]=cherry)"
>add to cart</button>
```

If we go with this, I suggest to go further with something like this:

```twig
<button type="button"
    data-action="live#action"
    data-live-modifiers-param="prevent|debounce(300)"
    data-live-action-param="addItem?command[itemID]=7&command[itemName]=cherry"
>add to cart</button>
```

WDYT?

weaverryan:
PR & description updated. I like where we've ended up - thanks for the push @sukei.

> That's an interesting idea. I just fear that complex values such as objects and arrays may be hard to write. But it can be abstracted by the Twig helper so it may not be a big deal.

💯 

> If we go with this, I suggest to go further with something like this:

I kept the `addItem(id=5)` syntax instead of `addItem?id=5`. Subjective, but I do like how `addItem(id=5)` looks like a method call. And also, for 90% of the use-cases (a single argument with no weird characters), `addItem(id=5)` is the same syntax people are already using today.

smnandre:
Made a table on how we used Directiveparser and why... 

How are we after this PR ? If i'm not mistaken, it's used only in loading attributes and events ? 


```
addItem(id=5)
```

Still think we should use a "json-php-namedparameter" syntax there :|  That can be done later i know i know :) 

```
addItem(id: 5)
```

smnandre:
Do you have a reason to not use stimulus syntax for args ?

Unfortunately, using a JS syntax - e.g. data-live-action-param="save(id: 5, name: 'Ryan')" would work... except we'd need to use eval() and that'll cause problems with csp.

```
data-live-action-NAME-param=VALUE
```

Here:

```
data-live-action-id-param="5"
data-live-action-name-param="Ryan"
```

weaverryan:
> Do you have a reason to not use stimulus syntax for args ?

That was the original proposal. @sukei made some good arguments against it - https://github.com/symfony/ux/pull/1418#issuecomment-1913571042 - especially item 1.

> Unfortunately, using a JS syntax - e.g. data-live-action-param="save(id: 5, name: 'Ryan')" would work... except we'd need to use eval() and that'll cause problems with csp.

Precisely. That's what pushed me to realize that a query string format is a great middle ground. For 95% of the use-cases (single argument and no weird characters) it looks dead-simple: `save(id=5)`

smnandre:
> That was the original proposal. @sukei made some good arguments against it - [#1418 (comment)](https://github.com/symfony/ux/pull/1418#issuecomment-1913571042) - especially item 1.

* lowerCamelCase is the Symfony official coding convention (and PER/PSR12 too)
* dashed-case is the data-attribute one, the HTML one, and the Stimulus one

We have LiveArg mapping done for this exact reason.

For me we should stick to standards (that would allow other controllers to get this value easily, to observe it, etc ...)

smnandre:
> For 95% of the use-cases (single argument and no weird characters) it looks dead-simple: save(id=5)

For 95% of use cases, data-xxx works perfectly (too). 

For 95% of the remaining use cases, (foo: bar, blip: blop) is perfect too

But query strings has this big problem of its handling of "array" data, that is different in JS, in HTML and in PHP... :|  

They are harder to secure too, validate, update, too than isolated fields.

sukei:
> We have LiveArg mapping done for this exact reason.

You are right, this can do the trick for my use case.

> But query strings has this big problem of its handling of "array" data, that is different in JS, in HTML and in PHP.

I know, that's why I proposed JSON first. Its easier to implement IMO.

> Do you have a reason to not use stimulus syntax for args ?

Just have a look at my `2.` observation. I fear that the magic interpretation of types by Stimulus may hurt. What if I need the string `"true"` and it ends casted as a `true` boolean value? That's why I wouldn't go that way (note that query string may have similar issues and we should refer to the Form component to handle this).

> dashed-case is the data-attribute one, the HTML one, and the Stimulus one

`dashed-case` is the HTML one. The Stimulus one is the consequence of this: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#javascript_access. It just follows the standard inflection.

smnandre:
I pushed for json too, but there is one thing it does not allow: reusing those values easily in other JS / DOM parts (well... query string has the same problem)

And finally Ryan convinced me about the DX aspect of it. Some people will want to just write some values, and we should not force them to use any special helper.

In a perfect world, we would have two formats, but at first i fear it's too much work to maintain/document/test/etc... 


Concerning your "2" .. 

I do understand your point and i agree we should not let an external library implementation decide how our values are typed (casted?).
But we _could_ type the values in the controller/component to inform stimulus what type we'd like, or expose some metadata in the props...  or re-typing after stimulus...  

What i mean is: there are solutions to all of this, and i'd rather we decide based on DX or maintenance..

(but really i agree/understand with most of your points)

sukei:
> but there is one thing it does not allow: reusing those values easily in other JS / DOM parts

I do not understand the use-case, can you elaborate this point? For what I see, those values will be passed to the Stimulus Live Controller when calling the action method. Those values are "closed" to this specific use case. For sure, you can select the node and read the dataset from outside, but you will get raw values (everything is a string) and some noise in the parameters names (`data-live-item-id-param -> liveItemIdParam`). Isn't it easier to bind the values you need in a specific attribute when you really need to (`data-item-id -> itemId`)?

smnandre:
Yes this is better ... and this is why i don't think query strings are the best solution :)) But i think we do agree on 99% here, i must miss-express myself .. again :))

anatholius:
Hi all

Inspired by today's Ryan's stream and reading this discussion I prepared some of my thoughts about, `all-in-one-code`

```Twig

        <button
            data-action="live#action"

            {# example: #}
            {# data-action-name="addItem(id={{ item.id }}, itemName=CustomItem)" #}
            {# variation: #}
            {# data-live-action-param="addItem(id={{ item.id }}&itemName=custom%20with%20space)" #}

            {# proposal refactoring thorough values API #}
            data-action-add-item-thread-name-value="addItem"
            data-action-add-item-thread-id-param-value="{{ item.id }}"
            data-action-add-item-thread-item-name-param-value="CustomItem"
            {# or #}
            data-action-add-item-thread-item-name-param-value="Custom with space"

            {# or with `live-` prefix #}
            {# NOTE: is that prefix necesarily needed? at all 'live-action' part #}
            {#       could be known from `data-action` attr #}

            {# example: #}
            {# data-live-action-param="prevent|debounce(300)|addItem(itemID, itemName)" #}
            {# proposal refactoring  thorough values API #}
            data-live-action-prevent-thread-name-value="prevent"
            data-live-action-debounce-thread-name-value="debounce"
            data-live-action-debounce-thread-{time?}-param-value="300"
            data-live-action-add-item-thread-name-value="addItem"
            data-live-action-add-item-thread-id-param-value="{{ itemId }}"
            data-live-action-add-item-thread-item-name-param-value="{{ itemName }}"

            {# NOTE: In Stimulus controller we can easy get values object:
                ```Stimulus
                ... extends Controller {
                    values = {
                        liveActionPreventThreadName: this.liveActionPreventThreadNameValue,
                        ...
                        liveActionAddItemThreadItemName: this.liveActionAddItemThreadItemNameValue,
                    }
                }
                ```
                // just as example but it could be prepared automatically by digging deeper in `this`
                parsing those name strings should be much easier than parsing some regex tries
                - live (is it needed here? probably not)
                - action (is it needed here? probably not)
                - add-item
                - thread (separator)
                - name (this.___Value is going to be a `name` of `add-item` thread)
                ...
                - item-name
                - param (separator)
                  dashToCamel(previous part - 'item-name')=itemName is going to be a `param` name with value=this.___Value
                _____
                ...parsing to:...
                #threads = [
                    {
                        name: addItem
                        params: {
                            id: ...
                            itemName: ...
                        }
                    },
                    ...
                ]
                having that kind of object should be easier to deal with in context of PROPS,
                or some local stimulus controller STATE
                and is unambiguous without special parsing
            #}
        >Add Item</button>
```

in Stimulus lifecycle: `initialize` before `connect` could be used to prepare this first

I hope this can be helpful, WDYT?

weaverryan:
We're down to debating between:

**A) query parameter arguments**

```html
<button
    data-action="live#action"
    data-live-action-param="addItem(id={{ item.id }}&foo=bar)"
>Add Item</button>
```

vs

**B) individual "action params"**

```html
<button
    data-action="live#action"
    data-live-action-param="addItem(id={{ item.id }})"
    data-live-id-param="{{ item.id }}"
    data-live-foo-param="bar"
>Add Item</button>
```

(A) It sounds like query params suffers from there being different array syntaxes in php vs JavaScript. Also this was mentioned as a con:

> They are harder to secure too, validate, update, too than isolated fields.

Though, I'm not really sure what this really means.

(B) has the cons of needing to map some `LiveParam` arg names in edge cases & edge-case situations where we don't want the type-casting from Stimulus.

Honestly, it's edge-cases vs edge-cases on 2 workable solutions. I vote for keeping things how they are: **query parameters**. We *would* have a way to deprecate this later and switch from (A) to (B) (by deprecating if you have `(` after your action name to detect you're passing args that way).

smnandre:
So how would you use spaces (that were the main demand before this) ?

adeys:
Hi everyone,

I'd like to contribute to the discussion about choosing between `A) query parameters arguments` and `B) stimulus action args parameter`. In my opinion, option B seems more favorable.

**A) query parameters arguments:**
From a Developer Experience (DX) standpoint, it appears less readable to me. Additionally, as some have pointed out, dealing with issues like spaces, data escaping, and array values may prove challenging.

**B) stimulus action args parameter:**
Given that LiveComponent is built on Stimulus, I believe it's beneficial to fully leverage its capabilities and encourage developers to explore its potential (although not mandatory). Despite the cons related to edge cases (such as non-flexible `LiveParam` arg name mapping and stimulus type casting), adopting a JSON format for parameters like `data-live-args-param='{ "first": "value" }'` could be a better approach.

This approach may require developers to adhere to a strict JSON format, but I find it more readable, explicit, and capable of addressing concerns like custom `LiveParam` names mapping and argument types. It also simplifies dealing with complex data types, such as arrays or objects.

To illustrate, developers can choose between two options:

**a) for purists:**
```html
data-action="live#action:prevent"
data-live-action-name="debounce(300)|addItem"
data-live-args-param='{"id": {{ item.id }}, "options": {"first": {{ item.value }}, "second": "true"}}'
```

or

**b) for helper users:**
```html
data-action="live#action:prevent"
data-live-action-name="debounce(300)|addItem"
data-live-args-param="{{ {id: item.id, options: {first: item.value, second: "true"}|json_encode }}"
```

I appreciate your consideration and look forward to your feedback.

smnandre:
My feeling is that we try to use more and more "standards" (Idiomorph as last example), and i'd find logical and safer to not make an exception here.

> Honestly, it's edge-cases vs edge-cases on 2 workable solutions

I think it is not "edge-cases vs edge-cases" ... at all :)

Letting the user handle query string is imho a really bad DX, for the following reasons:
- for anything but [a-zA-Z0-9] string, it requires an aditionnal function/helper from Twig 
- once in the queryString, it's harder to get, update, ...
- it become really hard to have conditions in it

```twig
<button
    data-action="live#action"
    data-live-action-name="createUser"
    data-live-first-name-param="{{ user.firstName }}"
    data-live-last-name-param="{{ user.lastName }}"
    {% if user.city is defined %}
    data-live-city-param="{{ user.city }}"
    {% endif  %}
>
Create User
</button>
```

How would look like the same thing with the "query string" ?  😵‍💫 

I know there are drawbacks in both options, but i really don't think it's fair to compare specific type casting problems (that i do not negate) to ... requiring additionnal methods to handle _**every**_ string, accents, spaces.. almost all but identifiers in fact :|

anatholius:
Thank you all for your replies

I took a few steps back to get to the root of the problem

These two linked issues at the beginning have basically the same approach: specifying the method statement to be executed in the "data-live-action-name" attribute.

Beyond any consideration, why?
Shouldn't it just be a `string` representing the name of the function, and the parameters should be passed appropriately as indicated by Stimulus?
Otherwise I understand it like this:
> The `data` attribute of an action with the name provided as the `value of this attribute` for the `live` controller contains a string representing a call to another function.
> This looks like a proxy to a function other than the action function.

If this action is intended to be a proxy for another action whose statement is passed as the name parameter of the current action, this is where parsing problems arise.

Maybe it's better to do it then
```
data-action="life#myAction"
data-life-my-action-proxy="theRealAction"
data-life-my-action-proxy-a-param="..."
data-life-my-action-proxy-b-param="..."
```
> or with the json representation as mentioned earlier few times

Sorry, these are purely theoretical considerations, because all of that could be made right away
```
data-action="life#theRealAction"
data-live-params="..."
```
> note that here is no indication of the action name other than in the `data-action` attribute

---

On the other hand

If a single "life#action" wants to use some filters (like in Twig): `action|filter1(...)|filter2`
wouldn't it be better to indicate this in the convention used by the stimulus, e.g.
```
data-life-action-filters="encode|raw"
```
or in the case of multiple filters with possible parameters
```
data-life-action-filter1-filter="..."
data-life-action-filter1-filter-x-param="..."
data-life-action-filter1-filter-y-param="..."
...
```

---

Another consideration is if we wanted to provide some parameters for the live action that would trigger some query/request to the backend.

Then (probably) it would be better to also provide it in the appropriate parameters

---

Since this problem is part of #1419, I think it's better not to do anything weird now, and if the current implementation is sufficient, I would focus on that, and the rest could be improved later if needed.

weaverryan:
This is ready to go again - back to the original attributes version of the feature. Test failures are unrelated, but we need to get on top of those :)

PR #1406, opened 2024-01-18
[LiveComponent] Document ``getComponent`` install with AssetMapper

smnandre:
Add document suggested by @weaverryan about [getComponent / AssetMapper ](https://github.com/symfony/ux/issues/1401)

PR #1419, opened 2024-01-28
[Live][Stimulus] Prepping the LiveComponent Stable Release 🚀 

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | None
| License       | MIT

Hi!

LiveComponents has, really, been quite stable for a long time, but it's kept its experimental status. Removing that is really about deciding that we'll protect backwards-compatibility. It's time to do that :).

This is planned as the 2.15.0 release near the end of Feb (assuming we get the items below done before then).

TODOs:

* #1418 
* Possibly remove Twig 2.x compat
* #1428
* #1392
* Moving `Idiomorph` to a peer dependency would be great, but blocked by https://github.com/bigskysoftware/idiomorph/pull/35 - **still need a tag** for the PR merge
* #1426 as it may include some edge-case BC breaks.

If there's anything else on your mind before stable, now is the time to mention it :).

Cheers!

### DIFF
--- a/src/LiveComponent/README.md
+++ b/src/LiveComponent/README.md
@@ -1,15 +1,12 @@
 # Live Components
 
-**EXPERIMENTAL** This component is currently experimental and is
-likely to change, or even change drastically.
-
 Live components work with the [TwigComponent](https://symfony.com/bundles/ux-twig-component/current/index.html)
 library to give you the power to automatically update your
 Twig components on the frontend as the user interacts with them.
 Inspired by [Livewire](https://laravel-livewire.com/) and
 [Phoenix LiveView](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html).
 
-Want a demo? Check out https://github.com/weaverryan/live-demo.
+Want a demo? Check out https://ux.symfony.com/live-component.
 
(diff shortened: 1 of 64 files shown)


src/LiveComponent/README.md
```diff
@@ -1,15 +1,12 @@
 # Live Components
 
-**EXPERIMENTAL** This component is currently experimental and is
```
smnandre, line 3: I'd put an info box to explicitely say "it's stable since .." .. wdyt ?

src/LiveComponent/doc/index.rst
```diff
@@ -1,11 +1,6 @@
 Live Components
 ===============
 
-.. caution::
```
smnandre, line 4: Same here :)

weaverryan: We don't add it to symfony/symfony, so I'm inclined to follow that.

smnandre:
3/ #1406 closed and replaced by  #1428

smnandre:
2/ #1436 opened (drop twig support)

PR #1421, opened 2024-01-28
[CI] Upgrade PHP highest version -> 8.3

smnandre:


.github/workflows/test-turbo.yml
```diff
@@ -38,11 +38,11 @@ jobs:
         strategy:
            fail-fast: false
            matrix:
-               php-version: ['8.1', '8.2']
+               php-version: ['8.1', '8.3']
```
WebMamba, line 41: Maybe we should also keep 8.2 in the matrix?

weaverryan: Our matrix is already huge, so doing 8.1 and 8.3 isn't ideal, but it feels like a fair compromise

weaverryan:
Thanks Simon!

PR #1428, opened 2024-01-30
[LiveComponent] Update importmap conf in package.json

smnandre:
Replaces #1406 

Allow using getComponent with no aditional configuration

```js
```


weaverryan:
Thanks Simon!

PR #1431, opened 2024-01-31
[LiveComponent] Add better error message when hydrating dates

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

Instead of 

```
Return value must be of type ?object, bool returned"
```
return a proper error message
```
The model path "createdAt" was sent invalid date data "0" or in an invalid format. Make sure it's a valid date and it matches the expected format "Y. m. d.".
```

Not sure if it's a bugfix or feature. Maybe the wording can be tweaked.


weaverryan:
Love this! Thank you Gábor!

Issue #1013, opened 2023-07-18
symfony/ux-chartjs impossible to use adapters


k-ecotrade:
Hello,
I've seen a few posts about this on stackoverflow, but no answer exists yet, so I wanted to know if it is possible to implement an adapter (example date-fns) to Symfony Ux's implementation of chart.js.
I tried for a long time but I can't find a solution.

Symfony seems to be using '@symfony/ux-chartjs' while the adapters are modifying 'chart.js' directly, from what I understand.

I didn't find a way to pass this problem, so if anybody can help me about this, it would be very appreciated.

May be if anyone can bring an example of working code for this.

Thank you for your time.

weaverryan:
Hi there!

This "should" just work - so we will need to dig into the problem a bit. Both `symfony/ux-chartjs` and `chartjs-adapter-date-fns` import code from the `chart.js` library. Because `chart.js` is a `peerDependency` of both, npm/yarn should only be installing 1 copy of `chart.js` into `node_modules/`. And so, when these 2 libraries import something from `chart.js`, they should be importing from the same package. The flow should be:

* `chartjs-adapter-date-fns` is imported and it modifies `chart.js`
* `chart.js` is imported by `symfony/ux-chartjs`, which uses the modified version.

So there could be 2 things going wrong:

A) Ordering: if `chart.js` is imported by `symfony/ux-chartjs` first... it's possible that it will use an unmodified version (though, since the chart isn't actually rendered until slightly later, I would still imagine that `chartjs-adapter-date-fns` would have a chance to modify things before the actual chart is built).

B) Or, perhaps the 2 libraries are, in the end, importing different files from the `chart.js` package for some reason - like one is importing `dist/chart.js` and another is importing `dist/chart.cjs` for some reason. 

Btw, where are you importing `chartjs-adapter-date-fns`? I would do it right in `assets/app.js` at the very top (at least, that's both the simplest and earliest place to do it).

k-ecotrade:
Thank you for answering and for your time.

I'm not 100% sure if I understand correctly how `@symfony/ux-chartjs` works, but when I check the files, it installs a '`/vendor/symfony/ux-chartjs/assets/node_modules`' folder, and it seems to use a chart.js 'copy' inside this folder. 
While '`chartjs-adapter-date-fns`' seems to modify what is inside the '`/node_modules/chart.js`' folder.

[...]

PR #1433, opened 2024-01-31
[ChartJs] Remove <3.9 boundary and fix testsuite

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  |no 
| Issues        | Relates to #1013 and #1417
| License       | MIT

In #1013 the version of `chartjs` was restricted to `<3.9` and since adjusting the test script in #1417 (and later merging #1389) it seemed that version constraint notation `^3.4.1 <3.9` wasn't supported ([see](https://github.com/symfony/ux/actions/runs/7723726257/job/21054412107?pr=1431#step:6:32)) by it. I tried supporting the boundary condition but it turned into quite the rabbit hole correctly parsing it and keeping it simple/pragmatic.

If I understand correctly, issue #1013, introduced the boundary condition because it was related to our testsuite, since that has been fixed in #1202. It looks like we therefor can drop the upper boundary again. I'm not sure if we cover the issue in our testsuite, but our testsuite passes on `chartjs@3.9.1`.

So the removal of the boundary introduces:
- Fully support `chartjs` versions `^3.4.1` and `^4.0`
- Testscript is fully ran again ([see](https://github.com/symfony/ux/actions/runs/7724645865/job/21057199292?pr=1433#step:6:29)) with no complex adjustments supporting exotic notations
  - When such boundary conditions are necessary in the future we could take a look at it again

Cheers!

### DIFF
--- a/src/Chartjs/assets/package.json
+++ b/src/Chartjs/assets/package.json
@@ -21,12 +21,12 @@
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0",
-        "chart.js": "^3.4.1 <3.9 || ^4.0"
+        "chart.js": "^3.4.1 || ^4.0"
     },
     "devDependencies": {
         "@hotwired/stimulus": "^3.0.0",
         "@types/chart.js": "^2.9.34",
-        "chart.js": "^3.4.1 <3.9 || ^4.0",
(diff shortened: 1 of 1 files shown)


evertharmeling:
Snugged in with #1518

PR #1434, opened 2024-01-31
fix: use method from metadata for live component test helper

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

After #1218 has been released as part of 2.14.0 my tests fail. This change passes the actual method to the requests made by the test helper.

### DIFF
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -51,13 +51,25 @@ final class TestLiveComponent
             $this->metadataFactory->getMetadata($mounted->getName())
         );
 
-        $this->client->request('GET', $this->router->generate(
-            $this->metadata->get('route'),
-            [
-                '_live_component' => $this->metadata->getName(),
-                'props' => json_encode($props->getProps(), flags: \JSON_THROW_ON_ERROR),
-            ]
-        ));
+        if ('POST' === strtoupper($this->metadata->get('method'))) {
+            $this->client->request(
+                'POST',
+                $this->router->generate($this->metadata->get('route'), [
+                    '_live_component' => $this->metadata->getName(),
+                ]),
+                [
+                    'data' => json_encode(['props' => $props->getProps()], flags: \JSON_THROW_ON_ERROR),
+                ],
+            );
+        } else {
+            $this->client->request('GET', $this->router->generate(
+                $this->metadata->get('route'),
+                [
+                    '_live_component' => $this->metadata->getName(),
+                    'props' => json_encode($props->getProps(), flags: \JSON_THROW_ON_ERROR),
+                ]
+            ));
+        }
     }
 
     public function render(): RenderedComponent


weaverryan (approved):

weaverryan (changes requested):
The change looks good, but I see that there are now failures inside the LiveComponent package itself.

smnandre:
> After #1218 has been released as part of 2.14.0 my tests fail. This change passes the actual method to the requests made by the test helper.

If you set "method: 'get'" on some of the tests that now fail, does it fix the problem or not ? (without the changes from this PR)

(just want to know if everything works in that case)

daFish:
@smnandre Setting `method: 'get'` let's my test pass. Otherwise an `Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException` is thrown.

smnandre:
From your PR, could you try to, when in POST, send "props" in JSON and not in the query string ? 


Someything "like that" (did not test at all)

```php
  $this->client->request('POST', $this->router->generate(
            $this->metadata->get('route'),
            [
                '_live_component' => $this->metadata->getName(),
            ]
        ), [
            'json' => ['props' => $props->getProps()]
        ]);
```

daFish:
This did the trick: 
```$this->client->request(
            'POST',
            $this->router->generate(
                $this->metadata->get('route'),
                [
                    '_live_component' => $this->metadata->getName(),
                ]
            ),
            [
                'data' => json_encode(['props' => $props->getProps()]),
            ],
        );
```

smnandre:
@daFish good news! Could you update your PR to include this change, and we'll see then if tests are OK or if there is something else to fix :)

daFish:
Looks like there is more work todo. I'll see if I can do something about it. Any hints are appreciated.

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -51,13 +51,16 @@ public function __construct(
...
         );
 
-        $this->client->request('GET', $this->router->generate(
-            $this->metadata->get('route'),
-            [
-                '_live_component' => $this->metadata->getName(),
-                'props' => json_encode($props->getProps(), flags: \JSON_THROW_ON_ERROR),
-            ]
-        ));
+        $this->client->request($this->metadata->get('method'),
```
smnandre:
```php
if ('POST' === $method = strtoupper($this->metadata->get('method'))) {
    $this->client->request(
        $method,
        $this->router->generate($this->metadata->get('route'), [
            '_live_component' => $this->metadata->getName(),
        ]),
        [
            'json' => ['props' => $props->getProps()],
        ],
    );
} else {
    $this->client->request($method, $this->router->generate(
        $this->metadata->get('route'),
        [
            '_live_component' => $this->metadata->getName(),
            'props' => json_encode($props->getProps(), flags: \JSON_THROW_ON_ERROR),
        ]
    ));
}
```
        
We need to handle both situations there

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -141,7 +144,7 @@ private function request(array $content = [], string $action = null): self
         $csrfToken = $this->csrfToken();
 
         $this->client->request(
-            'POST',
+            $this->metadata->get('method'),
```
smnandre: I'm not sure this one need any change... do your tests fail there ?

src/LiveComponent/src/Test/TestLiveComponent.php
daFish: Of course - adjusted.

src/LiveComponent/src/Test/TestLiveComponent.php
daFish: Reverted.

smnandre:
Do all your tests pass with this change? 😄

daFish:
Yes, they do. And the checks here are also passing. But some more feedback from others would also be appreciated.

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -51,13 +51,25 @@ public function __construct(
...
-        $this->client->request('GET', $this->router->generate(
-            $this->metadata->get('route'),
-            [
-                '_live_component' => $this->metadata->getName(),
-                'props' => json_encode($props->getProps(), flags: \JSON_THROW_ON_ERROR),
-            ]
-        ));
+        if ('POST' === $method = strtoupper($this->metadata->get('method'))) {
+            $this->client->request(
+                $method,
```
weaverryan: We can avoid the `$method` entirely and put `POST` here and `GET` below. Those are the only 2 valid methods.

daFish: Done.

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -51,13 +51,25 @@ public function __construct(
...
-            ]
-        ));
+        if ('POST' === strtoupper($this->metadata->get('method'))) {
+            $this->client->request(
+                'POST',
+                $this->router->generate($this->metadata->get('route'), [
+                    '_live_component' => $this->metadata->getName(),
+                ]),
+                [
+                    'data' => json_encode(['props' => $props->getProps()]),
```
smnandre: Oops, we lost the flag :) 

```php
'data' => json_encode(['props' => $props->getProps()], flags: \JSON_THROW_ON_ERROR),
```

daFish:
Failure unrelated.

weaverryan:
Thanks Marcus!

PR #1443, opened 2024-02-01
[Translator] Update intl-messageformat range version to ^10.5.11, close #1437

### LINKED ISSUE #1437: [Translator] Replace `intl-messageformat-parser` by `@formatjs/icu-messageformat-parser`
As mentioned in https://www.npmjs.com/package/intl-messageformat-parser, we should migrate to [`@formatjs/icu-messageformat-parser`](https://www.npmjs.com/package/@formatjs/icu-messageformat-parser) which is a much faster and backward compatible implementation.

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #1437 
| License       | MIT

This way, we can ensure users will use a version of `intl-messageformat` that use `@formatjs/icu-messageformat-parser`, a 6x faster and backward compatible implementation of `intl-messageformat-parser`.

### DIFF
--- a/src/Translator/CHANGELOG.md
+++ b/src/Translator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## UNRELEASED
+
+-   Increase version range of `intl-messageformat` to `^10.5.11`, in order tou se a faster implementation of ICU messages parsing.
+
 ## 2.13.2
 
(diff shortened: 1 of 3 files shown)


weaverryan:
Thanks Hugo!

PR #1444, opened 2024-02-01
[StimulusBundle] Handles Windows directory separator when normalizing controller names

### LINKED ISSUE #1422: [StimulusBundle] AssetMapper compiler doesn't handle controllers in subdirectory correctly
I am probably referring to this code segement (this is my first trial on AssetMapper):

https://github.com/symfony/ux/blob/b07f8642231d7d6b1c7c95c43335b62c6b99907a/src/StimulusBundle/src/AssetMapper/StimulusLoaderJavaScriptCompiler.php#L52-L132

Where the `$name` variable in the foreach loop is the controller name. But when the Stimulus controller javascript file is placed under a subdirectory (e.g. in `assets/controllers/admin/user_controller.js`, when `assets/controllers` is the root dir) this code doesn't replace the directory separater with a double-dash (`--`) according to the [naming convention](https://stimulus.hotwired.dev/handbook/installing#controller-filenames-map-to-identifiers), which causes the resultant controller array to look like this (I am using Windows, thus the backslash): 

```js
export const eagerControllers = {"admin\users": controller_0};
```

The result is that calling `{{ stimulus_controller('admin/users') }}` fails to work.

tamcy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1422
| License       | MIT

This fixes the issue that controller name normalization doesn't work as intended in Windows which uses `\` as the directory separator. In fact, without the fix, PHPUnit fails when running on Windows platform with the following error:

```
1) Symfony\UX\StimulusBundle\Tests\AssetMapper\ControllerMapGeneratorTest::testGetControllersMap
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
     4 => 'hello-with-dashes'
     5 => 'hello-with-underscores'
     6 => 'other'
-    7 => 'subdir--deeper'
-    8 => 'subdir--deeper-with-dashes'
-    9 => 'subdir--deeper-with-underscores'
+    7 => 'subdir\deeper'
+    8 => 'subdir\deeper-with-dashes'
+    9 => 'subdir\deeper-with-underscores'
     10 => 'typescript'
 )
```

### DIFF
--- a/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
+++ b/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
@@ -76,7 +76,7 @@ class ControllersMapGenerator
             // use regex to extract 'controller'-postfix including extension
             preg_match(self::FILENAME_REGEX, $name, $matches);
             $name = str_replace(['_'.$matches[1], '-'.$matches[1]], '', $name);
-            $name = str_replace(['_', '/'], ['-', '--'], $name);
+            $name = str_replace(['_', '/', '\\'], ['-', '--', '--'], $name);
 
             $asset = $this->assetMapper->getAssetFromSourcePath($file->getRealPath());
             $content = $asset->content ?: file_get_contents($asset->sourcePath);


weaverryan:
Thanks for this excellent PR @tamcy!

PR #1450, opened 2024-02-01
[Icons] Introduce `symfony/ux-icons`

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

[**_Documentation_**](https://github.com/kbond/ux/blob/feat/ux-icons-draft-2/src/Icons/README.md)

## TODO

- [x] Remove `symfony/dom-crawler` dependency
- [x] `<twig:UX:Icon />` component
- [x] More tests
- [x] `ux:icons:import` command
- [x] _Icons on demand_ system
- [x] Warmup only icons found in twig templates
- [x] _Symfony style_ docs

## Future Scope/Ideas

- Improved cache system (similar to how say, validation metadata is cached)
- Deferred icons (to avoid inline rendering many of the exact same icons)
- Icon _packs_: instead of adding icons one by one, include an entire set (ie fontawesome)
- Suggest similar icon names on `IconNotFoundException`
- `lint:icons` command:
  - show licenses of all iconify icons
  - warn when using a dynamic icon name (as it cannot be warmed)
  - error for icons that cannot be found
- Ability to disable iconify in test environment (to avoid a bunch of http calls when running tests)


src/Icons/src/Registry/LocalSvgIconRegistry.php
```diff
@@ -0,0 +1,64 @@
...
+        $svg = file_get_contents($filename) ?: throw new \RuntimeException(sprintf('The icon file "%s" could not be read.', $filename));
+        $crawler = (new Crawler($svg))->filter('svg');
+        $node = $crawler->getNode(0) ?? throw new \RuntimeException(sprintf('The icon file "%s" is not a valid SVG.', $filename));
+        $attributes = [];
+
+        if ($viewBox = $node->attributes?->getNamedItem('viewbox')?->nodeValue) {
+            $attributes['viewBox'] = $viewBox;
+        }
+
+        return [$crawler->html(), $attributes];
```
kbond: Can anyone with php-dom experience help me find the equivalent code that doesn't require `symfony/dom-crawer` & `symfony/css-selector`?

src/Icons/src/Registry/LocalSvgIconRegistry.php
```diff
@@ -0,0 +1,64 @@
...
+            throw new IconNotFoundException(sprintf('The icon "%s" (%s) does not exist.', $name, $filename));
+        }
+
+        $svg = file_get_contents($filename) ?: throw new \RuntimeException(sprintf('The icon file "%s" could not be read.', $filename));
+        $crawler = (new Crawler($svg))->filter('svg');
+        $node = $crawler->getNode(0) ?? throw new \RuntimeException(sprintf('The icon file "%s" is not a valid SVG.', $filename));
+        $attributes = [];
+
+        if ($viewBox = $node->attributes?->getNamedItem('viewbox')?->nodeValue) {
+            $attributes['viewBox'] = $viewBox;
```
smnandre: Only comment for tonight (brace yourself for tomorrow 😅)

We need a solution to keep some (or add some) or some svg packages won't be usable without adding a lot of manual things in every call :| 

I'm thinking height and width in particular

src/Icons/tests/Integration/Twig/UXIconExtensionTest.php
```diff
@@ -0,0 +1,42 @@
...
+final class UXIconExtensionTest extends KernelTestCase
+{
+    public function testRenderIcons(): void
+    {
+        $output = self::getContainer()->get(Environment::class)->createTemplate(<<<TWIG
+            <ul class="svg">
+                <li id="first">{{ ux_icon('user', {class: 'h-6 w-6'}) }}</li>
+                <li id="second">{{ ux_icon('user') }}</li>
+                <li id="third">{{ ux_icon('sub:check') }}</li>
+                <li id="forth">{{ ux_icon('sub/check') }}</li>
```
smnandre: I say let's pick one syntax and forbid other.

(or we'll have to deal with conversion/normalizations in many places) 

wdyt ?

src/Icons/src/Registry/LocalSvgIconRegistry.php
kbond: Doesn't viewbox cover height and width? I've tested with several different icons in separate collections and didn't see any issues.

src/Icons/tests/Integration/Twig/UXIconExtensionTest.php
kbond: I want to choose `:` as the standard separator but in this case, I can't really forbid `/` unless I want to explicitly throw an exception if the name contains it. I don't really want to do that. I think this would be a documentation thing.

The cache key remains the same whether `:` or `/` is used.

src/Icons/src/Registry/LocalSvgIconRegistry.php
smnandre: Height and view are set to zero if not given, so the icon won't be displayed without additionnal CSS

And it's a performance front-end best practices.

That's an example but my point is: i don't think we should, i'll do more tests tomorrow and give you other examples (lucide with their path strok / round instructions is another example)

src/Icons/tests/Integration/Twig/UXIconExtensionTest.php
smnandre: I see no problem to have a regexp basic check (that will be necessary to avoid some caracters, weird pathes, etc..)

smnandre: If name is handled as an "identifier" i do believe there should be only one syntax, but open to discussion.

src/Icons/src/Registry/LocalSvgIconRegistry.php
smnandre: Minor exception and checks, that is the idea:

```
$doc = new \DOMDocument();
$doc->loadXML($svg);
$svgTag = $doc->firstElementChild;
$html = $doc->saveHTML($svgTag);
$attributes = array_map(fn(\DOMAttr $a) => $a->value, [...$svgTag->attributes]);
```

There may be quicker

src/Icons/src/IconRenderer.php
```diff
@@ -0,0 +1,62 @@
...
+            '<svg%s>%s</svg>',
+            self::normalizeAttributes([...$iconAttr, ...$attributes]),
+            $content,
+        );
+    }
+
+    /**
+     * @param array<string,string|bool> $attributes
+     */
+    private static function normalizeAttributes(array $attributes): string
```
smnandre: We'll need some html escape there

Minor: i'd rename this renderAttributes() as it returns HTML

src/Icons/src/Registry/LocalSvgIconRegistry.php
kbond: > `$html = $doc->saveHTML($svgTag);`

This gives me the "outer html" (the entire svg tag), not the guts.

kbond: Solved but needs some tests.

src/Icons/src/IconRenderer.php
kbond: Can you give an example so I can make a test

src/Icons/tests/Integration/Twig/UXIconExtensionTest.php
kbond: Okay, I guess do we even want `:` has a directory separator? I chose this because that's what we do for twig components so it's familiar.

src/Icons/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### HTML Syntax
+
+> [!NOTE]
+> `symfony/ux-twig-component` is required to use the HTML syntax.
+
+```html
+<twig:UX:Icon name="user-profile" class="w-4 h-4" /> <!-- renders "user-profile.svg" -->
+
+<twig:UX:Icon name="sub-dir:user-profile" class="w-4 h-4" /> <!-- renders "sub-dir/user-profile.svg" (sub-directory) -->
```
weaverryan: Could we allow this name to be configured? I'd personally want it to just be `<twig:Icon`

src/Icons/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Caching
+
+To avoid having to parse icon files on every request, icons are cached (`app.cache` by
```
weaverryan: Does the "system" cache make more sense? Then, ideally, all of these caching details happen entirely in the background. There would be no reason to *not* warm up the cache in that case, as it's just saving on the filesystem.

src/Icons/README.md
```diff
@@ -0,0 +1,73 @@
...
+```html
+<twig:UX:Icon name="user-profile" class="w-4 h-4" /> <!-- renders "user-profile.svg" -->
+
+<twig:UX:Icon name="sub-dir:user-profile" class="w-4 h-4" /> <!-- renders "sub-dir/user-profile.svg" (sub-directory) -->
+```
+
+## Caching
+
+To avoid having to parse icon files on every request, icons are cached (`cache.system` by
+default but can be [configured](#full-default-configuration)).
```
weaverryan: Do we need to allow this to be configured? It feels like allowing the "validation metadata cache" to be configured.

src/Icons/README.md
```diff
@@ -0,0 +1,73 @@
...
+```yaml
+ux_icons:
+    # The local directory where icons are stored.
+    icon_dir: '%kernel.project_dir%/templates/icons'
+
+    # The cache pool to use for icons.
+    cache: cache.system
+
+    # Whether to warm the icon cache when the container is warmed up.
+    cache_on_container_warmup: true
```
weaverryan: Do we need this config? It seems like you always want this (again, like validation metadata cache)

src/Icons/composer.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Symfony\\UX\\Icons\\Tests\\": "tests/"
+        }
+    },
+    "require": {
+        "php": ">=8.1",
+        "symfony/framework-bundle": "^6.0|^7.0",
+        "symfony/twig-bundle": "^6.0|7.0"
```
weaverryan: I'd bump all of these to 6.4 min

src/Icons/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Installation
+
+```bash
+composer require symfony/ux-icons
+```
+
+## Add Icons
+
+No icons are provided by this package. Add your svg icons to the `templates/icons/` directory.
```
weaverryan:
```suggestion
No icons are provided by this package. Add your svg icons to the `templates/icons/` directory and commit them.
```

I do think we need a mechanism in v1 to fetch icons, whether that's an entire pack or one-by-one.

src/Icons/src/IconRenderer.php
weaverryan: If the user calls `$renderer->renderIcon('star', ['data-foo' => 'a "string" with a quote'])`, the end-result should be escaped for the HTML attributes, since the return value from this function is meant to be rendered raw. Apparently this is how we do it in StimulusBundle https://github.com/symfony/stimulus-bundle/blob/2.x/src/Dto/StimulusAttributes.php#L123

src/Icons/README.md
kbond: Yes, we can I think, great idea!

src/Icons/README.md
kbond: The reason I made it configurable is: if you have a ton of icons (1000's), the container warmup can be slow (ie 2-3 seconds). If we're ok with that, I can remove this config and the warmup command.

src/Icons/src/IconRenderer.php
kbond: Oh man, this strategy escapes valid attribute values (like spaces):

```diff
-viewBox="0 0 24 24" 
+viewBox="0&#x20;0&#x20;24&#x20;24"
```

Is this what we want?

kbond:
I've made the following cache changes:
1. Always use `cache.system` (no longer configurable)
2. Always warmup on `cache:warmup`/`cache:clear` (no longer configurable)

As Ryan pointed out, there's likely a better way to handle the cache. I'll dig into this but now, it shouldn't hold up a merge/initial release.

src/Icons/src/IconRenderer.php
smnandre: If you use Escaper, you need to use the "html_attr" strategy

src/Icons/tests/Integration/Twig/UXIconExtensionTest.php
smnandre: For ID / classes we'll need a way to distinct all icons from a packages, and i like the "--" method, as it's Stimulus one too.

I send you a PR in that sense.

For me the "path separator" is an implementation details, what we need is something that says "this icon in this package/set/..;"

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -0,0 +1,81 @@
...
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $builder = new TreeBuilder('ux_icons');
+        $rootNode = $builder->getRootNode();
+
+        $rootNode
+            ->children()
+                ->scalarNode('icon_dir')
+                    ->info('The local directory where icons are stored.')
+                    ->defaultValue('%kernel.project_dir%/templates/icons')
```
smnandre: I'm starting to think it's not a good idea.

This make impossible to expose icons, and i'm pretty sure, if we have icon packs, people would want to use them as assets. 

This also brings a lot of assets in a directory made for ".twig" files. 

What about using assets/icons and adding an ignore rule in the assetmapper conf ?

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -0,0 +1,81 @@
...
+                ->end()
+                ->scalarNode('twig_component_name')
+                    ->info('The name of the Twig component to use for rendering icons.')
+                    ->defaultValue('UX:Icon')
```
smnandre: We should not do this. 

If someone want to use a custom component, he shall create its own. But he won't benefit from all the optimisations we will do.

But making this configurable will be a hell to handle / maintain in the TwigComponent bundle, and will add even more "if cases" in places where they should not be handled.

Not speaking about all the unavoidable "why can't I create a <foo:bar component" questions that will create immediatly :|

src/Icons/src/Registry/CacheIconRegistry.php
```diff
@@ -0,0 +1,83 @@
...
+final class CacheIconRegistry implements IconRegistryInterface, CacheWarmerInterface
+{
+    /**
+     * @param IconRegistryInterface[] $registries
+     */
+    public function __construct(private \Traversable $registries, private CacheInterface $cache)
+    {
+    }
+
+    public function get(string $name, bool $refresh = false): array
```
smnandre: Not sure of this "$refresh" ...  what would be the usage ?

src/Icons/src/Registry/CacheIconRegistry.php
```diff
@@ -0,0 +1,83 @@
...
+            yield from $registry;
+        }
+    }
+
+    public function count(): int
+    {
+        $count = 0;
+
+        foreach ($this->registries as $registry) {
+            $count += \count($registry);
```
smnandre: What is the goal / usage ?

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -0,0 +1,81 @@
...
+            ->setArguments([
+                $mergedConfig['icon_dir'],
+            ])
+        ;
+
+        $container->getDefinition('.ux_icons.icon_renderer')
+            ->setArgument(1, $mergedConfig['default_icon_attributes'])
+        ;
+
+        $container->getDefinition('.ux_icons.twig_component.icon')
```
smnandre: Only if TwigComponent will be available ? Can this be done easily ?

(Will push more and more for an UxBundle that could coordinate all that, as FrameworkBundle does)

src/Icons/src/IconRegistryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ * @internal
+ */
+interface IconRegistryInterface extends \IteratorAggregate, \Countable
+{
+    /**
+     * @return array{0: string, 1: array<string, string|bool>}
+     *
+     * @throws IconNotFoundException
+     */
+    public function get(string $name): array;
```
smnandre: Made the PR against your other branch, so i'll write here too: if we expose an Interface, we definitively should use a POO instead of an array

src/Icons/src/Registry/CacheIconRegistry.php
```diff
@@ -0,0 +1,83 @@
...
+
+    public function get(string $name, bool $refresh = false): array
+    {
+        return $this->cache->get(
+            sprintf('ux-icon-%s', str_replace([':', '/'], ['-', '-'], $name)),
+            function () use ($name) {
+                foreach ($this->registries as $registry) {
+                    try {
+                        return $registry->get($name);
+                    } catch (IconNotFoundException) {
```
smnandre: In my mind
- we have only on registry there, and thus we should "just" decorate 
- we have multiple registries there, and then i think we should have something like "has" or "support", allowing them to be registered for specific paths / storage / etc

src/Icons/src/Twig/UXIconExtension.php
```diff
@@ -0,0 +1,31 @@
...
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @internal
+ */
+final class UXIconExtension extends AbstractExtension
+{
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('ux_icon', [IconRenderer::class, 'renderIcon'], ['is_safe' => ['html']]),
```
smnandre, line 28: Here or in the renderer, i think we should improve the DX a bit either suggestive correct names when user made a typo, maybe how to import icons, ..

src/Icons/src/IconRenderer.php
kbond: I copied the code from the stimulus bundle (uses `html_attr`) and that was the result.

smnandre:
I found the problem with Iconify... it removes attributes from the SVG tag, and insert them in a <g> in the innerSVG. 

### Lucide icon (downloaded from [website](https://lucide.dev/icons/activity-square))

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-activity-square"><rect width="18" height="18" x="3" y="3" rx="2"/><path d="M17 12h-2l-2 5-2-10-2 5H7"/></svg>
```

### Lucide icon (from [repo](https://github.com/lucide-icons/lucide/blob/main/icons/activity-square.svg?short_path=3edf692))

```svg
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <rect width="18" height="18" x="3" y="3" rx="2" />
  <path d="M17 12h-2l-2 5-2-10-2 5H7" />
</svg>
```

### Lucide icon (as SVG + CSS)

The idea beeing obviously to not repeat those attributes everywhere and set them globally, leading to small simple SVG files (more or less what Github does on this website)

Doc: https://lucide.dev/guide/packages/lucide-static

```css
.lucide-icon {
  width: 24px;
  height: 24px;
  stroke: currentColor;
  fill: none;
  stroke-width: 2;
  stroke-linecap: round;
  stroke-linejoin: round;
}
```

```svg
<svg  class="lucide-icon">
<rect width="18" height="18" x="3" y="3" rx="2" /><path d="M17 12h-2l-2 5-2-10-2 5H7"/>
</svg>
```

Much cooler, right ? 


### Iconify Icon (downloaded from [website](https://icon-sets.iconify.design/lucide/activity-square/))

There is where the problem starts. Iconify added a <g> tag where it moved root attributes to.

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 24 24"><g fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><rect width="18" height="18" x="3" y="3" rx="2"/><path d="M17 12h-2l-2 5l-2-10l-2 5H7"/></g></svg>
```

Much harder then to optimise things.

And the result is the same when provided from the API.



## Problems

I see there multiple problems 

* it generate twice the code needed
* it makes harder to set values globally (not every icon will come from Iconify)
* it hard-codes things in the innerSVG --something that need to be accessible easily via CSS, ideally on the root component
* it creates a version of the icon that will be different (in content and possibilities) of the original one

I'm pretty sure there are solutions to all those points, i am NOT suggesting we don't use this amazing service.. but we should address those

src/Icons/src/DependencyInjection/UXIconsExtension.php
kbond: Yeah, I was on the fence on what to do here. Here's my thinking for the current default:
1. It'd be easy enough for someone to change this to `%kernel.project_dir%/assets/icons` if they want to expose them.
2. I think we don't want them exposed by default so it'd add a bit more complexity (adding the ignore rule).
3. The latest twig recipe excludes non-twig files from being looked at

src/Icons/src/Registry/CacheIconRegistry.php
kbond: This forces the cache to be rebuilt (even if the icon is cached). See the warmUp method below for its usage.

src/Icons/src/Registry/CacheIconRegistry.php
kbond: Originally, it was to provide a total for the `ux:icons:warm-cache`'s progress bar: so  "a way to get a count of all currently registered icons"

src/Icons/src/DependencyInjection/UXIconsExtension.php
kbond: I believe if TwigComponent is not available, this service will just be removed from the container but I'll confirm.

src/Icons/src/IconRegistryInterface.php
kbond: > if we expose an Interface

Technically, this interface isn't exposed (it's internal)... But I get your point.

smnandre:
> * Always use `cache.system` (no longer configurable)
> 
> * Always warmup on `cache:warmup`/`cache:clear` (no longer configurable)

TwigBundle uses a dedicated directory

```php
    private function addTwigOptions(ArrayNodeDefinition $rootNode): void
    {
        $rootNode
            ->fixXmlConfig('path')
            ->children()
                ->scalarNode('autoescape_service')->defaultNull()->end()
                ->scalarNode('autoescape_service_method')->defaultNull()->end()
                ->scalarNode('base_template_class')->example('Twig\Template')->cannotBeEmpty()->end()
                ->scalarNode('cache')->defaultValue('%kernel.cache_dir%/twig')->end()
                ->scalarNode('charset')->defaultValue('%kernel.charset%')->end()
                ->booleanNode('debug')->defaultValue('%kernel.debug%')->end()
                ->booleanNode('strict_variables')->defaultValue('%kernel.debug%')->end()
                
```

If the doc is right, the system cache does not seem the ideal place to store 3000 icons, especially if we generate one cache file per icon (two with tags?)

> There are two pools that are always enabled by default. They are cache.app and cache.system. **The system cache is used for things like annotations, serializer, and validation.** The cache.app can be used in your code. You can configure which adapter (template) they use by using the app and system key like:

https://symfony.com/doc/current/cache.html#cache-app-system

kbond:
> If the doc is right, the system cache does not seem the ideal place to store 3000 icons, especially if we generate one cache file per icon (two with tags?)

Yes, I think this is the conclusion Ryan came to as well. I'm going to dig deeper into this.

src/Icons/src/Registry/LocalSvgIconRegistry.php
smnandre: Documentation (if you want to digg) https://html.spec.whatwg.org/multipage/embedded-content-other.html#dimension-attributes

src/Icons/src/IconRenderer.php
smnandre: I used a more direct method in my Icon object PR

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: I do agree, as we already talked about, to not send all this into assetmapper.

The problem i see there is that it creates an weird split between assets and ... assets. 
For a standard website, that'd mean icons needed to be exposed would be in a place (the logo, partner logos, idk) would be in assets/logo (preventing to use it with the icon function) but the github icon, or the partners ones would be in templates/icons ?

In a way, that make perfect sense, but with the favicons in a place, that start to look like an annoying DX (Designer Experience) :p 

(not having a perfect solution to present) 

In the end, if people download TablerIcon twice (once in assets, once in templates) we will need to reconsider this :)) But let's see, the "experimental" period can be used for that :)

src/Icons/src/Registry/CacheIconRegistry.php
smnandre: Is the cache cleared before ? If "yes", no need for this attribute... if "not"... well i think it should :) 

Let's see when we have sorted the "pool" question, we may have access to a "clearer" then that would remove any need for those things.

src/Icons/src/DependencyInjection/UXIconsExtension.php
kbond: > The problem i see there is that it creates an weird split between assets and ... assets.

Yes, exactly! This is why I was on the fence. `assets/icons/*` feels like a much better home than `templates/icons/*`.

src/Icons/src/Registry/CacheIconRegistry.php
kbond: Yes, this is true. I guess it's a a relic from when I had a configurable cache pool.

src/Icons/src/Registry/CacheIconRegistry.php
smnandre: Ok, i see. Love it for the command DX, but as it's not using cached data, it may be a bit "dangerous" to expose i think.

```
if (!empty($registry)) {
```

Would that not call "count" somehow ?

kbond: > but as it's not using cached data

This is the case for `->getIterator()` as well. As it's not currently used, we can remove `count()` but need to keep `getIterator()`.

src/Icons/src/DependencyInjection/UXIconsExtension.php
kbond: I've confirmed this.

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: The "only" problem with assets/icons/* is for the massive packs (which are probably already in this folder today in many projects).

So why not start with "assets/icons" for now, and find a solution for the packs while we develop the "pack" features ? 

I see no problem having packs in specific folders, that's really the main "header / footer" icons that i'd find strange to separate from the CSS / image files.

src/Icons/src/Registry/CacheIconRegistry.php
smnandre: I genuinely don't know if it's cleared or not, i see no "clear" in the base CacheInterface :/

src/Icons/src/Registry/CacheIconRegistry.php
smnandre: I think TwigBundle uses specific code for that in commands, but on the other hand i'm not a big fan of having that kind of logic hardcoded in a command, requiring to duplicate it for the "lint" command (that we will need)

src/Icons/.gitattributes
```diff
@@ -0,0 +1,5 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.symfony.bundle.yaml export-ignore
+/assets/src export-ignore
+/assets/test export-ignore
```
smnandre:
```suggestion
/tests export-ignore
```

src/Icons/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2020-present Fabien Potencier
```
smnandre:
```suggestion
Copyright (c) 2024-present Fabien Potencier
```

src/Icons/composer.json
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "symfony/ux-icons",
+    "type": "symfony-bundle",
+    "description": "Twig component to use svg icons in your Symfony app",
+    "keywords": [
+        "symfony-ux",
+        "twig",
+        "icons"
```
smnandre:
```suggestion
        "icons",
        "svg"
```

src/Icons/composer.json
```diff
@@ -0,0 +1,52 @@
...
+            "Symfony\\UX\\Icons\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Symfony\\UX\\Icons\\Tests\\": "tests/"
+        }
+    },
+    "require": {
+        "php": ">=8.1",
```
smnandre, line 30:
```suggestion
        "php": ">=8.2",
```

src/Icons/composer.json
```diff
@@ -0,0 +1,52 @@
...
+        "symfony/framework-bundle": "^6.4|^7.0",
+        "symfony/twig-bundle": "^6.4|7.0"
```
smnandre:
```suggestion
        "symfony/twig-bundle": "^6.4|^7.0"
```

src/Icons/composer.json
kbond: Is this what we want?

src/Icons/src/Twig/UXIconExtension.php
kbond: I've added to the future scope list.

src/Icons/src/Registry/LocalSvgIconRegistry.php
kbond: For now, I've changed to not remove any attributes

kbond:
~Anyone know what's up with the [CI failure](https://github.com/symfony/ux/actions/runs/7768770773/job/21187083273?pr=1450#step:7:23) on prefer-lowest? I'm seeing the same thing locally but I've never encountered this issue before.~ Fixed, I had to allow phpunit-bridge 6.3

src/Icons/src/Registry/LocalSvgIconRegistry.php
smnandre: How much time do you let me to suggest an implementation that would solve this question ?

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: Ok, i realize you "just" wanted to allow `<twig:Foo:Bar />` here ... 

As i understood it, `<UX:Icon />` was something a user could chose.

So sorry... still think we should not let this configurable, but i'm less afraid :))

src/Icons/composer.json
smnandre: Symfony 7.x requires >= 8.2, but 6.4 still allows 8.1 

We can accept 8.1, but active support ended 2 months ago, and security updates will at the end of the year. 
During the experimental phase, not a problem of course :) 

So is there something in 8.2 we might need (readonly classes? not a real _need_ i think) ?

src/Icons/src/Registry/LocalSvgIconRegistry.php
kbond: I guess I just want svg icons downloaded from _some icon registry site_ to _just work_ without any extra requirements. I thought there was some issues with height/width attributes taking precedence over css classes but as you've explained, that's not the case.

src/Icons/composer.json
kbond: I think requiring 8.1 makes this package consistent with the other packages. If we make 8.2, I think we'll need to make some considerations in the CI system that automatically runs tests for all components (I'd prefer not to do this).

src/Icons/src/Registry/CacheIconRegistry.php
kbond: I've refactored this and removed the refresh arg.

src/Icons/src/Registry/CacheIconRegistry.php
kbond: I've refactored to remove countable as it isn't used.

src/Icons/src/DependencyInjection/UXIconsExtension.php
weaverryan: I'm leaning towards `assets/icons` also. Not sure, in practice, if having 1000 icons served through AssetMapper is an actual problem or not. And, as you said, perhaps the "packs" solution could fix this anyway - e.g. maybe the packs are downloaded into the `var/` directory (as they don't need be committed).

src/Icons/src/DependencyInjection/UXIconsExtension.php
weaverryan: I'm on the fence about between:

A) Keeping this
B) Jumping in and eating out own dog food by adding a proper way to "import" 3rd party components and leveraging it here. I just really hate having the YAML config to do this :p

src/Icons/src/Twig/UXIconComponentListener.php
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        $attributes = $event->getInputProps();
+        $name = (string) $attributes['name'];
+        unset($attributes['name']);
+
+        $svg = $this->iconRenderer->renderIcon($name, $attributes);
+        $event->setRenderedString($svg);
+        $event->stopPropagation();
+    }
```
weaverryan, line 42: Is this an optimization for rendering the Twig component? Is `Icon.html.twig` in `templates/` ever used?

weaverryan (commented):
Pretty happy with this :). That being said, it feels feature complete without a way for users to grab icons. Telling them to download them manually, I think, isn't good enough.

src/Icons/src/DependencyInjection/UXIconsExtension.php
kbond: I've made this change (it's still configurable).

src/Icons/src/Twig/UXIconComponentListener.php
kbond: Good catch, I was able to remove the template entirely!

src/Icons/src/DependencyInjection/UXIconsExtension.php
kbond: I've removed the ability to customize for now.

kbond:
I've added an [`ux:icons:import` command](https://github.com/kbond/ux/blob/feat/ux-icons-draft-2/src/Icons/README.md#import-command). This allows importing individual icons from https://icon-sets.iconify.design/ (which is a huge searchable repository of icons).

src/Icons/README.md
```diff
@@ -0,0 +1,77 @@
+# Symfony UX Icons
```
smnandre: I do like the README (and sometimes think the symfony ones are a bit.... rough)

As the other packages uses this "standard" (not sure it's the good term), should we writing this content in the same RST file: `doc/index.rst` ?

kbond: I used the readme to document this feature as it's easy to review on github (and easier for me to write). Before release, we'll switch to rst.

smnandre: This answer makes so much sense I feel stupid haha

src/Icons/README.md
```diff
@@ -57,8 +57,6 @@ site.
 
 To avoid having to parse icon files on every request, icons are cached.
 
-During container warmup (`cache:warmup` and `cache:clear`), the icon cache is warmed.
-
 > [!NOTE]
 > During development, if you change an icon, you will need to clear the cache (`bin/console cache:clear`)
```
smnandre: To remove too ?

src/Icons/src/Registry/CacheIconRegistry.php
```diff
@@ -33,25 +32,6 @@ public function get(string $name): Icon
...
-    public function warmUp(string $cacheDir, ?string $buildDir = null): array
-    {
-        foreach ($this as $name) {
-            $this->fetchIcon($name, refresh: true);
-        }
-
-        return [];
-    }
-
     private function fetchIcon(string $name, bool $refresh = false): Icon
```
smnandre: Was the $refresh made only for the cache warmer ?

src/Icons/src/Twig/IconFinder.php
```diff
@@ -0,0 +1,81 @@
...
+     * @return string[]
+     */
+    public function icons(): array
+    {
+        $found = [];
+
+        foreach ($this->files($this->twig->getLoader()) as $file) {
+            $contents = file_get_contents($file);
+
+            if (preg_match_all('#ux_icon\(["\']([\w:-]+)["\']#', $contents, $matches)) {
```
kbond, line 41: @smnandre, I'm hoping you can help refactor and improve this logic. 🙏🏻

kbond:
**NEW FEATURE: Iconify Icons on Demand!**

You can now just use _any icon_ from https://icon-sets.iconify.design/ without importing/downloading or anything!

src/Icons/src/Twig/IconFinder.php
smnandre: NodeVisitor :)) 

I'll look tuesday! 

We can find ideas in Translation extraction too i think

src/Icons/src/Twig/UXIconComponentListener.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function onPreCreateForRender(PreCreateForRenderEvent $event): void
+    {
+        if ('UX:Icon' !== $event->getName()) {
+            return;
+        }
+
+        $attributes = $event->getInputProps();
+        $name = (string) $attributes['name'];
```
smnandre, line 36: Reminder i need to add better tests there

src/Icons/src/Registry/LocalSvgIconRegistry.php
```diff
@@ -0,0 +1,86 @@
...
+
+    public function get(string $name): Icon
+    {
+        if (!file_exists($filename = sprintf('%s/%s.svg', $this->iconDir, str_replace(':', '/', $name)))) {
+            throw new IconNotFoundException(sprintf('The icon "%s" (%s) does not exist.', $name, $filename));
+        }
+
+        $svg = file_get_contents($filename) ?: throw new \RuntimeException(sprintf('The icon file "%s" could not be read.', $filename));
+
+        $doc = new \DOMDocument();
```
smnandre: Ok if i move all the parsing stuff in a dedicated class ?

kbond: Sure, what about a named constructor on `Icon`?

src/Icons/.gitattributes
```diff
@@ -0,0 +1,4 @@
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.symfony.bundle.yaml export-ignore
+/tests export-ignore
```
smnandre:
```suggestion
/.gitattributes export-ignore
/.gitignore export-ignore
/.symfony.bundle.yaml export-ignore
/phpunit.xml.dist export-ignore
/tests export-ignore
```

src/Icons/composer.json
```diff
@@ -0,0 +1,54 @@
+{
+    "name": "symfony/ux-icons",
+    "type": "symfony-bundle",
+    "description": "Twig component to use svg icons in your Symfony app",
```
smnandre: I think almost all Symfony component have been reworded in a not-so-distant past, and the "rule" seems to me to say what it "does" more than what it "is"..

https://packagist.org/search.json?q=symfony/

I wrote that for mine: 

> Renders and transforms SVG icons, manages local or remote Icon Packs and generates SVG sprites.

 but as we have trimmed down, we should ... trim the description too :))

src/Icons/src/Command/WarmCacheCommand.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\UX\Icons\Exception\IconNotFoundException;
+use Symfony\UX\Icons\Registry\CacheIconRegistry;
+use Symfony\UX\Icons\Twig\IconFinder;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @internal
```
smnandre, line 24: Hard to justify as it's not registered as cachewarmer for now, right ?

src/Icons/README.md
```diff
@@ -0,0 +1,103 @@
...
+Add your svg icons to the `assets/icons/` directory and commit them.
+The name of the file is used as the _name_ of the icon (`name.svg` will be named `name`).
+If located in a subdirectory, the _name_ will be `sub-dir:name`.
+
+### Iconify Icons
+
+[Iconify Design](https://iconify.design/) is a huge searchable repository of icons from many different icon sets.
+This package provides a way to include any icon found on this site _on-demand_.
+
+1. Visit [Iconify Design](https://icon-sets.iconify.design/) and search for an icon
```
smnandre: ux.symfony.com ? 👼

src/Icons/README.md
```diff
@@ -0,0 +1,103 @@
...
+You can import any icon from Iconify to your local directory using the `ux:icons:import` command:
+
+ ```bash
+ bin/console ux:icons:import flowbite:user-solid # saved as `flowbite/user-solid.svg` and name is `flowbite:user-solid`
+
+ # import several at a time
+ bin/console ux:icons:import flowbite:user-solid flowbite:home-solid
+ ```
+
+## Usage
```
javiereguiluz: Kevin, thanks for this contribution. I like this a lot.

I have a question about how this would work in practice. When the docs say _"it renders the XXX icon"_ does it mean:

* It embeds the `<svg ...> ... </svg>` contents where you called the `ux_icon()` function
* It makes an Ajax request to fetch the contents and embeds it?
* It just links to some SVG asset URL?

I think it'd be more clear to me if you showed the full real output of these functions. This would also be useful to know where the CSS classes are applied, etc. Thanks.

kbond: Thanks Javier, indeed it renders the svg tag in your html. I'll update to make that clear.

src/Icons/src/Command/WarmCacheCommand.php
kbond: My intention here was to mark the command code as internal - but running the command not internal.

src/Icons/README.md
kbond: I've updated the readme to explain with an example at the very top - let me know if you think this is sufficient.

src/Icons/src/Command/WarmCacheCommand.php
smnandre: Oh i see ! Maybe `@final`  then ?

src/Icons/README.md
javiereguiluz: Yes, it's very clear now. Thanks!

src/Icons/src/Command/WarmCacheCommand.php
kbond: Well the class is final. Maybe I go a little internal-happy sometimes but the idea here is if someone is using intellisense to autocomplete a class name, I want this class "excluded" from that search (by excluded, I mean stikethough in phpstorm anyway)

smnandre: Smart!

src/Icons/composer.json
kbond: "Renders local and remote SVG icons in your Twig templates." ?

src/Icons/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Icons _On-Demand_
+
+[ux.symfony.com/icons](https://ux.symfony.com/icons) has a huge searchable repository of icons
+from many different sets. This package provides a way to include any icon found on this site _on-demand_.
+
+1. Visit [ux.symfony.com/icons](https://ux.symfony.com/icons) and search for an icon
+   you'd like to use. Once you find one you like, copy one of the code snippets provided.
+2. Paste the snippet into your twig template and the icon will be automatically fetched (and cached).
+3. That's it!
```
weaverryan: Maybe 1-2 sentences (brief) explaining how it works might be useful for some

src/Icons/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+You can import any icon from [ux.symfony.com/icons](https://ux.symfony.com/icons) to your local
+directory using the `ux:icons:import` command:
+
+ ```bash
+ bin/console ux:icons:import flowbite:user-solid # saved as `flowbite/user-solid.svg` and name is `flowbite:user-solid`
+
+ # import several at a time
+ bin/console ux:icons:import flowbite:user-solid flowbite:home-solid
+ ```
```
weaverryan: Hmm, but why would I want to do this versus on-demand? This second option, unless qualified with being the best option for a specific use-case, clouds things.

src/Icons/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+## Caching
+
+To avoid having to parse icon files on every request, icons are cached.
+
+In production, you can pre-warm the cache by running the following command:
+
+```bash
+bin/console ux:icons:warm-cache
```
weaverryan: And this is still an independent command? It doesn't hook into `cache:warmup` or `asset-map:compile` to happen automatically? I'm mixed, tbh, on if it should. Part of me feels like this is a detail that I would love users not to need to worry about.

src/Icons/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+This command looks in all your twig templates for `ux_icon` calls and caches the icons it finds.
+
+> [!NOTE]
+> During development, if you change an icon, you will need to clear the cache (`bin/console cache:clear`)
```
weaverryan:
```suggestion
> During development, if you modify an icon, you will need to clear the cache (`bin/console cache:clear`)
```

(to make it more clear that we're talking about modifying the `.svg` file vs changing from one icon to another in a Twig template)

src/Icons/README.md
```diff
@@ -0,0 +1,112 @@
...
+ux_icons:
+    # The local directory where icons are stored.
+    icon_dir: '%kernel.project_dir%/assets/icons'
+
+    # Default attributes to add to all icons.
+    default_icon_attributes:
+        # Default:
+        fill: currentColor
+
+    # Configuration for the Iconify.design functionality.
```
weaverryan:
```suggestion
    # Configuration for the "on demand" icons powered by Iconify.design
```

src/Icons/composer.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "require": {
+        "php": ">=8.1",
+        "symfony/framework-bundle": "^6.4|^7.0",
+        "symfony/twig-bundle": "^6.4|^7.0"
+    },
+    "require-dev": {
+        "symfony/console": "^6.4|^7.0",
+        "symfony/http-client": "6.4|^7.0",
+        "symfony/phpunit-bridge": "^6.3|^7.0",
```
weaverryan, line 38: minor - but bump to 6.4

src/Icons/config/services.php
```diff
@@ -0,0 +1,78 @@
...
+            ->tag('twig.runtime')
+
+        ->set('.ux_icons.twig_component_listener', UXIconComponentListener::class)
+            ->args([
+                service('.ux_icons.icon_renderer'),
+            ])
+            ->tag('kernel.event_listener', [
+                'event' => 'Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent',
+                'method' => 'onPreCreateForRender',
+            ])
```
weaverryan: Minor, but should this service be registered conditionally? Else, it's in the event dispatcher even with an app that doesn't use twig component.

src/Icons/src/IconRenderer.php
```diff
@@ -0,0 +1,43 @@
...
+    ) {
+    }
+
+    /**
+     * @param array<string,string|bool> $attributes
+     */
+    public function renderIcon(string $name, array $attributes = []): string
+    {
+        // TODO generate class name(s)
+        // TODO add role/aria
```
weaverryan: We've got some TODOs here :)

src/Icons/src/Twig/IconFinder.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * @return string[]
+     */
+    public function icons(): array
+    {
+        $found = [];
+
+        foreach ($this->files($this->twig->getLoader()) as $file) {
```
weaverryan, line 38: Should we use https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/TwigBundle/TemplateIterator.php for this? Idea taken from https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/TwigBundle/CacheWarmer/TemplateCacheWarmer.php#L44

src/Icons/src/Twig/UXIconComponentListener.php
```diff
@@ -0,0 +1,43 @@
...
+final class UXIconComponentListener
+{
+    public function __construct(
+        private IconRenderer $iconRenderer,
+    ) {
+    }
+
+    public function onPreCreateForRender(PreCreateForRenderEvent $event): void
+    {
+        if ('UX:Icon' !== $event->getName()) {
```
weaverryan, line 31: There's not an instance check we can do here? Or even a check for a template? The point of this listener, if I understand correctly, is performance: to short-circuit the rendering process by avoiding the template render and, instead rendering right here, early. Would it make sense to include a nonsense template name and we look for that here? The philosophical idea is that this listener is doing the work of a template - almost standing in for a template. So maybe a template name is what we should use as the key?

weaverryan (changes requested):
Love this so much ❤️

src/Icons/README.md
smnandre: Doc is work in progress, i started to rewrite things a bit, but as i had difficulties i paused it a bit. Will rewrite with your feedbacks later tonight.

src/Icons/src/Twig/IconFinder.php
smnandre: Ongoing PR about this, could use help: https://github.com/smnandre/ux/tree/icon-extractor

(not used to work via PR on PR and honestly i'm still not confortable with it)

src/Icons/src/Twig/UXIconComponentListener.php
smnandre: No it replaces a TwigComponent rendering, so the name is the name of the component :) 

We bypass the TwigComponent instanciation/mount/etc entirely

src/Icons/README.md
kbond: I'm concerned about possibly 100's of http calls being made whenever the container cache is warmed if we hook into it directly.

Maybe `PreAssetsCompileEvent` is the better place? But this does remove the possibility of someone using w/o asset-mapper.

src/Icons/src/Twig/IconFinder.php
kbond: Yeah, Simon and I have discussed this - my preference is to improve this with a followup PR.

src/Icons/README.md
kbond: > why would I want to do this versus on-demand?

We've talked a lot about how best to "lock" icons (to completely ensure you never get a different version). I really think this is the best and simplest way.

During heavy development, the on-demand system is great but I'm thinking some user's would want to lock their choices down. Simon and I talked about a `ux:icons:lock` command to look through all your templates, find icons, and import them.

Maybe we can adjust the docs to move this section lower? Perhaps use the term _lock_?

src/Icons/composer.json
kbond: 6.3 is [actually required](https://github.com/symfony/ux/pull/1450#issuecomment-1925428075) (but I'm not entirely sure why).

src/Icons/README.md
weaverryan: Much clearer! I'm ok with the location anywhere. But let's talk specifically about the reason you might do this: to lock in a specific version of an icon. And that you will commit this to your repository.

src/Icons/README.md
weaverryan: Yea, let's do `PreAssetsCompileEvent`, but keep the command, but as an encore-only thing.

src/Icons/README.md
kbond: > But let's talk specifically about the reason you might do this: to lock in a specific version of an icon. And that you will commit this to your repository.

Here's some reasons but maybe @smnandre has more? I think Simon's a bit hesitant on the icons-on-demand system.

1. Like I said above, ensure the svg version is exactly what you expect (in case the version on iconify.design has changed)
2. Increase the speed of cache warming.

src/Icons/src/Command/WarmCacheCommand.php
```diff
@@ -41,16 +39,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        foreach ($this->icons->icons() as $icon) {
-            try {
-                $this->registry->get($icon, refresh: true);
-
-                if ($output->isVerbose()) {
-                    $io->writeln(sprintf(' Warmed icon <comment>%s</comment>.', $icon));
+        $this->warmer->warm(
+            onSuccess: function (string $name) use ($io) {
+                if ($io->isVerbose()) {
+                    $io->writeln(sprintf(' Warmed icon <comment>%s</comment>.', $name));
```
javiereguiluz, line 45: Minor comment: could we replace _"Warmed icon ..."_ by something that describes with more precision what's going on? E.g. _"Downloaded icon ..."_ or _"Extracted icon ..."_

src/Icons/src/Command/WarmCacheCommand.php
```diff
@@ -41,16 +39,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $io = new SymfonyStyle($input, $output);
         $io->comment('Warming the icon cache...');
```
javiereguiluz, line 40: In WarmIconCacheOnAssetCompileListener we already display this same comment ... maybe this is oputput twice then?

src/Icons/src/Command/WarmCacheCommand.php
kbond: Hmm, with the currently architecture, in this spot, I can't tell exactly what happened (if it was downloaded or just extracted from a local svg). Adding verbosity to the command I think could get the http-client logs.

Perhaps a future feature could be to add logging/events.

src/Icons/src/Command/WarmCacheCommand.php
kbond: This shouldn't be an issue as they are two different things:
1. `ux:icons:warm-cache` is a standalone cli command
2. `WarmIconCacheOnAssetCompileListener` is only used with `asset-map:compile`

src/Icons/README.md
weaverryan: Let's lock those reasons then into this section. We can iterate on it later. The important thing is that if we have 2 ways to accomplish the same thing, we at least give our users the quick pros/cons to help them decide.

weaverryan:
Congrats! And a huge, huge thanks to both you Kevin and Simon - can't wait to use this

tacman:
So this is semi-live now, e.g. https://symfony.com/bundles/ux-icons/current/index.html

But the all-important link to get the icons codes is still broken: https://ux.symfony.com/icons

Any chance at least that link could work?

smnandre:
This will be live soon :) 

Until then, you can find the icons on iconify, as we use their search engine / api.

seb-jean:
Hi,
Why is SVG on https://icon-sets.iconify.design/lucide/activity-square/ different from https://lucide.dev/icons/square-activity ?

https://icon-sets.iconify.design/lucide/activity-square/:
```html
<svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 24 24"><g fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><rect width="18" height="18" x="3" y="3" rx="2"/><path d="M17 12h-2l-2 5l-2-10l-2 5H7"/></g></svg>
```

https://lucide.dev/icons/square-activity:
```html
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-square-activity"><rect width="18" height="18" x="3" y="3" rx="2"/><path d="M17 12h-2l-2 5-2-10-2 5H7"/></svg>
```

There is the `<g>` tag which is added on https://icon-sets.iconify.design/

javiereguiluz:
@seb-jean thanks for reporting this ... but it's better that you create a new issue in the repo. This PR was already merged and the discussion is so long, that your comment will probably get lost. Thank you!

smnandre:
Hello @seb-jean, could you open a new issue ? It's easier afterwards to follow the issues / what's open or not, etc :) 

Update: what a timing @javiereguiluz 😸

seb-jean:
Ok 😄 : https://github.com/symfony/ux/issues/1677

PR #1459, opened 2024-02-05
[Autocomplete] Add conflict on doctrine/orm:^3.0

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | see below
| License       | MIT

Since [doctrine/orm:3.0](https://github.com/doctrine/orm/releases/tag/3.0.0) is released and there is no direct dependency defined in the `Autocomplete`-component on `doctrine/orm`, you're able to install it in your project. However when doing so it will result in an error when using the autocomplete:

Capture-2024-02-05-150938

A hacky workaround fix in `\Symfony\UX\Autocomplete\Doctrine\EntityMetadata:47` to keep BC, could be:

```php
if (\array_key_exists($propertyName, $this->metadata->fieldMappings)) {
    return (array) $this->metadata->fieldMappings[$propertyName];
}
```

But also since `zenstruck/foundry` doesn't support `doctrine/orm:^3.0` yet, we would not be able to run the testsuite and see any other incompatibilities in the component. So we need to take a closer look for all implications.

So for now it's probably wise to add a conflict on `doctrine/orm:^3.0` until we are able to confidently add support for it. Alternatively we could also add a direct dependency on `doctrine/orm:^2.9.4` (instead of only a dev dep)?

- [LiveComponent](https://github.com/symfony/ux/blob/2.x/src/LiveComponent/composer.json#L38) doesn't seem to be impacted atm, as it requires `doctrine/annotations:^1.0` which requires `doctrine/lexer:^1 || ^2` which does not allow `doctrine/orm:^3.0`.
- [Turbo](https://github.com/symfony/ux/blob/2.x/src/Turbo/composer.json#L42) already seems to have support for it.

### DIFF
--- a/src/Autocomplete/composer.json
+++ b/src/Autocomplete/composer.json
@@ -52,7 +52,7 @@
         "zenstruck/foundry": "^1.33"
     },
     "conflict": {
-        "doctrine/orm": "2.9.0 || 2.9.1"
+        "doctrine/orm": "2.9.0 || 2.9.1 || ^3.0"
     },
     "config": {
         "sort-packages": true


smnandre:
I don't think we should add a composer conflict there, as the Autocomplete does not require doctrine-orm. Probably more a runtime message (if really it cannot be used with ORM 3.0)

We should neither restrict any dependencies based on the internal toolkit used to test/develop it, as it would bring limitations in userland based on something they probably do not use or need.

With the change you suggest, would this component be usable in production in an app using Doctrine ORM 3 ? If yes, let's create a PR on that, wdyt ? 

What do you guys think about that ? @weaverryan  @kbond  is there already a position about Doctrine 3 in the symfony/symfony repo ?

evertharmeling:
> I don't think we should add a composer conflict there, as the Autocomplete does not require doctrine-orm. Probably more a runtime message (if really it cannot be used with ORM 3.0)
> 
> We should neither restrict any dependencies based on the internal toolkit used to test/develop it, as it would bring limitations in userland based on something they probably do not use or need.

Sorry if was unclear, the error occurs when you select an autocomplete field and start filtering, so it's not that it's happing in the testsuite (only). Therefor it would currently break applications with an autocomplete field implemented and upgrading to Doctrine ORM 3... That's why I went for the 'easy' fix of adding the conflict which would then prevent this problem and we would then have time to add proper support for it (and then remove the conflict again).

> With the change you suggest, would this component be usable in production in an app using Doctrine ORM 3 ? If yes, let's create a PR on that, wdyt ?

Sure, only thing is that our testsuite (`zenstruck/foundry` for 1 as I saw) currently isn't capable of testing Doctrine ORM 3, so I couldn't be sure if it was the only changed needed to support it. We could take action on that to start :)

weaverryan:
Thanks for this! I'm also leaning toward "let's just fix this now" if reasonable. For the EntityMetadata, it looks like it might make sense to change the return value from an array to FieldMapping, no? If we ignore BC on that method for a moment, would that fix things?

If so, I might be ok with introducing that small BC break. The thing is: if you're calling this method directly (I really should have made it internal) and you're on orm 3, then us returning an array doesn't make sense and will likely break your app. What the method was intended to do was "return the metadata from Doctrine, whatever that format".

smnandre:
> the error occurs when you select an autocomplete field and start filtering, so it's not that it's happing in the testsuite (only)

Sorry if i missread it first :) 

Doctrine ORM 3.0 released just a couple of days ago, right ? So you're right we must do something _now_ : either adding a conflict or fixing it, before anyone accuses us to "hide" some incompatibilities.

Is your "hacky" workaround really "hacky" ... or is it something we'd have to do anyway to support ORM 3 ?

wouterj:
A conflict will not work: If you add the conflict in 2.15.0, Composer will always fallback to installing 2.14.0 as it didn't declare the conflict (if it's within the allowed version range)

So fixing it now or throwing a runtime exception is the only possible way forward.

evertharmeling:
Good call @wouterj, I wasn’t aware of that…

Then the easiest way will be to introduce the ‘small’ BC break by changing the return type from ‘array’ to ‘FieldMapping’ as @weaverryan suggested.

Hopefully its the only change needed. Will take a look at it tomorrow, but feel free to beat me to it 😉

evertharmeling:
closed in favor of #1468

Issue #1124, opened 2023-09-18
[Twig Component] Test helper traits don't play nicely with CSRF, sessions and authentication


benr77:
Thank you @kbond for the awesome test helper traits for Live and Twig components. 

However, in some of my new tests that use these traits, I'm running into some issues. All this is PHP 8.2 on Symfony 6.3, with Live/Twig components on the latest `2.11.2` version.

All these tests extend Symfony's `KernelTestCase` via my own `KernelTestCase` wrapper, where I add a `setUp` method to run before all tests cases:

```php
abstract class KernelTestCase extends \Symfony\Bundle\FrameworkBundle\Test\KernelTestCase
{
    /** @var TestContainer */
    public Container|ContainerInterface $container;

    protected function setUp(): void
    {
        self::bootKernel();
        $this->container = static::getContainer();
    }
}
```

I think this is quite a normal configuration.

However, then with some tests I'm running into the following issues:

### 1) Session not found

Any Twig or Live Component that touches the session, either in PHP or in the template, will throw `SessionNotFoundException: There is currently no session available.` when the test is run. 

For me this is usually because I have e.g. a Date Range filter stored in the session, and this is then used inside the component to filter e.g. a repository query.

The solution to this is to pass in the value from the session when placing the component in the parent template. i.e. 

`{{ component('report_income_by_client', { dateRange: app.session.get('dateRange') }) }}`

This is actually more elegant in a way, as it keeps the concerns of the component itself to a bare minimum.

Then, in the test, we just pass in the "dateRange" as a parameter that we control from inside the test:

```php
$rendered = $this->renderTwigComponent(
    name: ReportIncomeByClientComponent::class,
    data: [
        'dateRange' => $dateRange, // define this elsewhere in the test
    ],
);
```

So this is a working solution that I'm more posting here for future reference than anything else. However, it leads nicely on to the second problem:

### 2) CSRF token not found

If any of the components under test are rending a form, then I get failures with the following exception:

```
Symfony\Component\HttpFoundation\Exception\SessionNotFoundException: There is currently no session available.

/var/www/app/vendor/symfony/http-foundation/RequestStack.php:107
/var/www/app/vendor/symfony/security-csrf/TokenStorage/SessionTokenStorage.php:110
/var/www/app/vendor/symfony/security-csrf/TokenStorage/SessionTokenStorage.php:74
/var/www/app/vendor/symfony/security-csrf/CsrfTokenManager.php:69
[...]

PR #1460, opened 2024-02-05
minor(live): add security tests

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Could help solve/replicate #1124
| License       | MIT

This PR adds some security-related tests:

- Ensure `IsGranted` at the class level is respected.
- Ensure `#[CurrentUser]` can be injected into actions/listeners.

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -45,7 +45,7 @@
         "symfony/options-resolver": "^5.4|^6.0|^7.0",
         "symfony/phpunit-bridge": "^6.0|^7.0",
         "symfony/property-info": "^5.4|^6.0|^7.0",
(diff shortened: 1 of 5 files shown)


weaverryan:
Thanks Kevin!

smnandre:
They fails on PHP 8.1 with old version of security: https://github.com/symfony/ux/actions/runs/7849800864/job/21423912123 

Should we fix... or increase min version of security bundle ? 

@kbond  ?

kbond:
I think I'm find bumping the dev dependency. It saves a bunch of extra compat code in the test fixtures and it should all still work in the old security system. @weaverryan, wdyt?

PR #1461, opened 2024-02-05
feat(live): improve `TestLiveComponent::actingAs()`

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

_(depends on #1460)_

Currently, if you have a component that requires authentication for all actions (ie `#[IsGranted]` on the component class), in your tests, as soon as you call `$this->createLiveComponent(...)`, you'll get an access denied exception. You never get a chance to call `->actingAs()`.

This PR makes the initial request for the live component _lazy_. Only when the initial request is required (for other methods), is it made. This gives you the chance to call `->actingAs()` before the initial render.

### DIFF
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -28,12 +28,14 @@ use Symfony\UX\TwigComponent\Test\RenderedComponent;
  */
 final class TestLiveComponent
 {
+    private bool $performedInitialRequest = false;
+
     /**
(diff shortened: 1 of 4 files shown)


daFish:
@kbond Using your changes works in my case. I no longer get the CSRF-exception.

kbond:
Rebased and fixed the failing test I introduced in #1460.

weaverryan:
Thanks Kevin!

PR #1468, opened 2024-02-07
[Autocomplete] Add support for doctrine/orm:^3.0

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes-ish?
| Issues        | Replaces #1459
| License       | MIT

Follow up on #1459 as it seemed a bit more complicated :)

Because in the current `getPropertyMetadata` function, [see](https://github.com/symfony/ux/blob/2.x/src/Autocomplete/src/Doctrine/EntityMetadata.php#L44) and below:

```php
public function getPropertyMetadata(string $propertyName): array
{
    if (\array_key_exists($propertyName, $this->metadata->fieldMappings)) {
        return $this->metadata->fieldMappings[$propertyName];
    }

    if (\array_key_exists($propertyName, $this->metadata->associationMappings)) {
        return $this->metadata->associationMappings[$propertyName];
    }

    throw new \InvalidArgumentException(sprintf('The "%s" field does not exist in the "%s" entity.', $propertyName, $this->metadata->getName()));
}
```

The function combines getting metadata from `fieldMappings` and `associationMappings`, that both return an `array` in **Doctrine ORM 2**. 

This however changes in **Doctrine ORM 3** in `fieldMappings` returning a `FieldMapping` object and `associationMappings` in an `AssociationMapping` object.

To support both Doctrine ORM 2 and 3, I didn't changed the return type of the function for now, as we would otherwise need to bump the major version for the `Autocomplete` component.

To ease the transition to **Doctrine ORM 3** in the future, I've splitted `getPropertyMetadata` in `getFieldMetadata` and `getAssociationMetadata`.

## TODO
- [x] Should we keep the `getPropertyMetadata()` function as a proxy to the new methods? (I removed it for now as the method shouldn't be used directly as [mentioned](https://github.com/symfony/ux/pull/1459#issuecomment-1928120377).
- [x] Should we retroactively add the `@internal` the `src/Autocomplete/src/Doctrine/EntityMetadata.php` class also (as [mentioned](https://github.com/symfony/ux/pull/1459#issuecomment-1928120377))?
- [ ] Test on Doctrine ORM 3, currently blocked by https://github.com/zenstruck/foundry/pull/556

### DIFF
--- a/src/Autocomplete/composer.json
+++ b/src/Autocomplete/composer.json
@@ -26,6 +26,7 @@
     "require": {
         "php": ">=8.1",
         "symfony/dependency-injection": "^6.3|^7.0",
+        "symfony/deprecation-contracts": "^2.5|^3",
         "symfony/http-foundation": "^6.3|^7.0",
         "symfony/http-kernel": "^6.3|^7.0",
         "symfony/property-access": "^6.3|^7.0",
@@ -34,7 +35,7 @@
     "require-dev": {
         "doctrine/collections": "^1.6.8|^2.0",
         "doctrine/doctrine-bundle": "^2.4.3",
-        "doctrine/orm": "^2.9.4",
+        "doctrine/orm": "^2.9.4|^3.0",
         "fakerphp/faker": "^1.22",
         "mtdowling/jmespath.php": "^2.6",
         "symfony/form": "^6.3|^7.0",
--- a/src/Autocomplete/src/Doctrine/EntityMetadata.php
+++ b/src/Autocomplete/src/Doctrine/EntityMetadata.php
@@ -43,12 +43,40 @@ class EntityMetadata
 
     public function getPropertyMetadata(string $propertyName): array
     {
+        trigger_deprecation('symfony/ux-autocomplete', '2.15.0', 'Calling EntityMetadata::getPropertyMetadata() is deprecated. You should stop using it, as it will be removed in the future.');
+
+        try {
+            return $this->getFieldMetadata($propertyName);
+        } catch (\InvalidArgumentException $e) {
+            return $this->getAssociationMetadata($propertyName);
+        }
+    }
+
+    /**
+     * @internal
+     *
+     * @return array<string, mixed>
+     */
+    public function getFieldMetadata(string $propertyName): array
+    {
         if (\array_key_exists($propertyName, $this->metadata->fieldMappings)) {
-            return $this->metadata->fieldMappings[$propertyName];
+            // Cast to array, because in doctrine/orm:^3.0; $metadata will be a FieldMapping object
+            return (array) $this->metadata->fieldMappings[$propertyName];
         }
 
+        throw new \InvalidArgumentException(sprintf('The "%s" field does not exist in the "%s" entity.', $propertyName, $this->metadata->getName()));
+    }
+
+    /**
+     * @internal
+     *
+     * @return array<string, mixed>
+     */
+    public function getAssociationMetadata(string $propertyName): array
+    {
         if (\array_key_exists($propertyName, $this->metadata->associationMappings)) {
-            return $this->metadata->associationMappings[$propertyName];
+            // Cast to array, because in doctrine/orm:^3.0; $metadata will be an AssociationMapping object
+            return (array) $this->metadata->associationMappings[$propertyName];
         }
 
         throw new \InvalidArgumentException(sprintf('The "%s" field does not exist in the "%s" entity.', $propertyName, $this->metadata->getName()));
@@ -56,7 +84,11 @@ class EntityMetadata
 
     public function getPropertyDataType(string $propertyName): string
     {
-        return $this->getPropertyMetadata($propertyName)['type'];
+        if (\array_key_exists($propertyName, $this->metadata->fieldMappings)) {
+            return $this->getFieldMetadata($propertyName)['type'];
+        }
+
+        return $this->getAssociationMetadata($propertyName)['type'];
     }
 
     public function getIdValue(object $entity): string
--- a/src/Autocomplete/src/Doctrine/EntitySearchUtil.php
+++ b/src/Autocomplete/src/Doctrine/EntitySearchUtil.php
@@ -60,7 +60,7 @@ class EntitySearchUtil
                 }
 
                 $originalPropertyName = $associatedProperties[0];
-                $originalPropertyMetadata = $entityMetadata->getPropertyMetadata($originalPropertyName);
+                $originalPropertyMetadata = $entityMetadata->getAssociationMetadata($originalPropertyName);
                 $associatedEntityDto = $this->metadataFactory->create($originalPropertyMetadata['targetEntity']);
 
                 for ($i = 0; $i < $numAssociatedProperties - 1; ++$i) {
@@ -75,9 +75,8 @@ class EntitySearchUtil
                     }
 
                     if ($i < $numAssociatedProperties - 2) {
-                        $propertyMetadata = $associatedEntityDto->getPropertyMetadata($associatedPropertyName);
-                        $targetEntity = $propertyMetadata['targetEntity'];
-                        $associatedEntityDto = $this->metadataFactory->create($targetEntity);
+                        $propertyMetadata = $associatedEntityDto->getAssociationMetadata($associatedPropertyName);
+                        $associatedEntityDto = $this->metadataFactory->create($propertyMetadata['targetEntity']);
                     }
                 }
 
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+-   Add doctrine/orm 3 support.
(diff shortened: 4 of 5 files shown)


src/Autocomplete/src/Doctrine/EntitySearchUtil.php
```diff
@@ -75,9 +75,8 @@ public function addSearchClause(QueryBuilder $queryBuilder, string $query, strin
                     }
 
                     if ($i < $numAssociatedProperties - 2) {
-                        $propertyMetadata = $associatedEntityDto->getPropertyMetadata($associatedPropertyName);
-                        $targetEntity = $propertyMetadata['targetEntity'];
-                        $associatedEntityDto = $this->metadataFactory->create($targetEntity);
+                        $propertyMetadata = $associatedEntityDto->getAssociationMetadata($associatedPropertyName);
```
evertharmeling, line 78: Not sure if I have to add a `$associatedEntityDto->isAssociation($propertyName)` check here and use the corresponding (introduced) method.

src/Autocomplete/src/Doctrine/EntityMetadata.php
```diff
@@ -41,22 +41,45 @@ public function isEmbeddedClassProperty(string $propertyName): bool
         return \array_key_exists($propertyNameParts[0], $this->metadata->embeddedClasses);
     }
 
-    public function getPropertyMetadata(string $propertyName): array
+    public function getFieldMetadata(string $propertyName): array
```
weaverryan, line 60: Can you add phpdoc with `@return` for what this is an array of - e.g. `FieldMapping[]` if i'm correct.

src/Autocomplete/src/Doctrine/EntityMetadata.php
```diff
@@ -41,22 +41,45 @@ public function isEmbeddedClassProperty(string $propertyName): bool
...
+                return (array) $metadata;
+            }
+
+            return $metadata;
         }
 
+        throw new \InvalidArgumentException(sprintf('The "%s" field does not exist in the "%s" entity.', $propertyName, $this->metadata->getName()));
+    }
+
+    public function getAssociationMetadata(string $propertyName): array
```
weaverryan, line 75: Nice idea splitting the 2 methods. Let's add `@internal` to these 2 methods at least. Then yes, leave the old method - we just won't call it anymore directly. And trigger a deprecation inside of it. We'll need to require the deprecations contract package, but that's no big deal.

src/Autocomplete/src/Doctrine/EntitySearchUtil.php
weaverryan: Not sure - this area is complex. If I understand correctly, we already know we're dealing with an association, so it should be safe as it is.

weaverryan (commented):
Thanks for getting this going - I think I answered the pending questions

src/Autocomplete/src/Doctrine/EntityMetadata.php
evertharmeling: Added `@return` doc, however the function still returns a plain array as before ([see](https://github.com/symfony/ux/pull/1468/files#diff-c35f51a8ce40a6d4ef620df4b1482777a35d9bd5f0b758ee9429d2422054dc81R57)), I've kept it BC, because the `FieldMapping` and `AssociationMapping` classes are introduced in Doctrine ORM 3 and not available in Doctrine ORM 2.

So when we would drop support for Doctrine ORM 2, we could also refactor the function(s) to something like:

```php
public function getFieldMetadata(string $propertyName): FieldMapping
{
    try {
        return $this->metadata->getFieldMapping($propertyName);
    } catch (MappingException $e) {
        throw new \InvalidArgumentException(sprintf('The "%s" field does not exist in the "%s" entity.', $propertyName, $this->metadata->getName()));
    }
}
```

src/Autocomplete/src/Doctrine/EntitySearchUtil.php
evertharmeling: That was my thought too, thanks!

src/Autocomplete/composer.json
```diff
@@ -26,6 +26,7 @@
     "require": {
         "php": ">=8.1",
         "symfony/dependency-injection": "^6.3|^7.0",
+        "symfony/deprecation-contracts": "^2.5|^3",
         "symfony/http-foundation": "^6.3|^7.0",
         "symfony/http-kernel": "^6.3|^7.0",
         "symfony/property-access": "^6.3|^7.0",
```
weaverryan, line 32: Below, in `require-dev`, we should allow for `doctrine/orm` 2 AND 3 now, so both are tested.

src/Autocomplete/src/Doctrine/EntityMetadata.php
```diff
@@ -42,21 +42,65 @@ public function isEmbeddedClassProperty(string $propertyName): bool
...
+    public function getFieldMetadata(string $propertyName): array
     {
         if (\array_key_exists($propertyName, $this->metadata->fieldMappings)) {
-            return $this->metadata->fieldMappings[$propertyName];
+            $metadata = $this->metadata->fieldMappings[$propertyName];
+
+            // In doctrine/orm:^3.0; $metadata will be a FieldMapping object
+            if (!\is_array($metadata)) {
+                return (array) $metadata;
+            }
```
weaverryan: For ORM 3, won't this return `[0 => FieldMapping]`? But in 2.9, it would be an associative array containing the metadata? Or am I misunderstanding?

src/Autocomplete/src/Doctrine/EntityMetadata.php
```diff
@@ -42,21 +42,65 @@ public function isEmbeddedClassProperty(string $propertyName): bool
...
         if (\array_key_exists($propertyName, $this->metadata->associationMappings)) {
-            return $this->metadata->associationMappings[$propertyName];
+            $metadata = $this->metadata->associationMappings[$propertyName];
+
+            // In doctrine/orm:^3.0; $metadata will be an AssociationMapping object
+            if (!\is_array($metadata)) {
+                return (array) $metadata;
+            }
+
+            return $metadata;
```
smnandre:
```suggestion
            return (array) $this->metadata->associationMappings[$propertyName];
```

src/Autocomplete/src/Doctrine/EntityMetadata.php
```diff
@@ -42,21 +42,65 @@ public function isEmbeddedClassProperty(string $propertyName): bool
...
 
         throw new \InvalidArgumentException(sprintf('The "%s" field does not exist in the "%s" entity.', $propertyName, $this->metadata->getName()));
     }
 
     public function getPropertyDataType(string $propertyName): string
     {
-        return $this->getPropertyMetadata($propertyName)['type'];
+        try {
+            return $this->getFieldMetadata($propertyName)['type'];
+        } catch (\InvalidArgumentException $e) {
```
smnandre: Could we avoid handling exception ? Is there a way to detect which method to use ?

src/Autocomplete/src/Doctrine/EntityMetadata.php
evertharmeling: Yes in ORM 3, the `ClassMetaData::$fieldMappings` is of type `array<string, FieldMapping>` ([see](https://github.com/doctrine/orm/blob/3.0.x/src/Mapping/ClassMetadata.php#L334-L336)) and in ORM 2 it's just an `array` (`mixed` in docs, [see](https://github.com/doctrine/orm/blob/2.10.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L449-L452)), it's been defined as a `psalm-type` ([see](https://github.com/doctrine/orm/blob/2.10.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L73-L92)), that may cause the confusion?

Since we access the `ClassMetaData::$fieldMappings` with a 'key' it will return a `FieldMapping` object directly and casting it to an array works for the `FieldMapping` object as it is defined using public properties.

src/Autocomplete/src/Doctrine/EntityMetadata.php
evertharmeling: Good one Simon, thanks!

src/Autocomplete/src/Doctrine/EntityMetadata.php
evertharmeling: I refactored it to use `array_key_exists` like in the `getFieldMapping` function ;)

evertharmeling:
Thanks for reviewing Ryan and Simon! Sorry for the late response, but the flu hit me and my little one got the chickenpox, it was a messy week 😅

weaverryan:
Thanks Evert! We'll test on ORM 3 when Foundry adds support

smnandre:
> Thanks for reviewing Ryan and Simon! Sorry for the late response, but the flu hit me and my little one got the chickenpox, it was a messy week 😅

Thank _you_ for this, and i hope everything is better now for you & familly 😃

PR #1471, opened 2024-02-07
[Turbo] Fixing support for not using old ClassUtils

### LINKED ISSUE #1469: Upgrade to Doctrine ORM 3.0 breaks Turbo
I tested upgrading to Doctrine ORM 3.0/DBAL 4.0 and ran into a problem where those packages no longer require `doctrine/common` but `ux-turbo` relies on the `ClassUtils` class from `common`.

I was able to fix the problem by directly requiring `doctrine/common` in my project, but I think `ux-turbo` should require `doctrine/common` unless there are plans to move away from using the `ClassUtils` class (I may have missed discussion about this somewhere).

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1469
| License       | MIT

### DIFF
--- a/src/Turbo/src/Doctrine/ClassUtil.php
+++ b/src/Turbo/src/Doctrine/ClassUtil.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo\Doctrine;
+
+use Symfony\Component\VarExporter\LazyObjectInterface;
+
+/**
+ * @internal
+ */
+final class ClassUtil
+{
+    public static function getEntityClass(object $entity): string
+    {
+        if ($entity instanceof LazyObjectInterface) {
+            $entityClass = get_parent_class($entity);
+            if (false === $entityClass) {
+                throw new \LogicException('Parent class missing');
+            }
+
+            return $entityClass;
+        }
+
+        // @legacy for old versions of Doctrine
+        if (class_exists(ClassUtils::class)) {
+            return ClassUtils::getClass($entity);
+        }
+
+        return $entity::class;
+    }
+}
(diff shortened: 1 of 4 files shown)


weaverryan:
This, indeed, fixes the reported problem

src/Turbo/src/Doctrine/ClassUtil.php
```diff
@@ -0,0 +1,39 @@
...
+            $entityClass = get_parent_class($entity);
+            if (false === $entityClass) {
+                throw new \LogicException('Parent class missing');
+            }
+
+            return $entityClass;
+        }
+
+        // @legacy for old versions of Doctrine
+        if (class_exists(ClassUtils::class)) {
```
stloyd, line 33: This class is not imported so it will always return false.

jrushlow: https://github.com/symfony/ux/pull/1475

jrushlow: Good catch!

src/Turbo/src/Doctrine/ClassUtil.php
```diff
@@ -0,0 +1,39 @@
...
+ * @internal
+ */
+final class ClassUtil
+{
+    public static function getEntityClass(object $entity): string
+    {
+        if ($entity instanceof LazyObjectInterface) {
+            $entityClass = get_parent_class($entity);
+            if (false === $entityClass) {
+                throw new \LogicException('Parent class missing');
```
nicolas-grekas, line 26: no need for this exception IMHO, using get_class should be the way if there's no parent

src/Turbo/src/Doctrine/ClassUtil.php
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * @internal
+ */
+final class ClassUtil
+{
+    public static function getEntityClass(object $entity): string
+    {
+        if ($entity instanceof LazyObjectInterface) {
+            $entityClass = get_parent_class($entity);
```
nicolas-grekas, line 24: `return get_parent_class($entity) ?: \get_class($entity);`

PR #1477, opened 2024-02-08
[Turbo] simplify returning entity class name

jrushlow:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes/no 
| Issues        | Fix #
| License       | MIT

credit goes to nicolas, but alternate way to return the class name

### DIFF
--- a/src/Turbo/src/Doctrine/ClassUtil.php
+++ b/src/Turbo/src/Doctrine/ClassUtil.php
@@ -22,12 +22,7 @@ final class ClassUtil
     public static function getEntityClass(object $entity): string
     {
         if ($entity instanceof LazyObjectInterface) {
-            $entityClass = get_parent_class($entity);
-            if (false === $entityClass) {
-                throw new \LogicException('Parent class missing');
-            }
-
-            return $entityClass;
+            return get_parent_class($entity) ?: $entity::class;
         }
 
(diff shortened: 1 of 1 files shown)


src/Turbo/src/Doctrine/ClassUtil.php
```diff
@@ -22,12 +22,7 @@ final class ClassUtil
     public static function getEntityClass(object $entity): string
     {
         if ($entity instanceof LazyObjectInterface) {
-            $entityClass = get_parent_class($entity);
-            if (false === $entityClass) {
-                throw new \LogicException('Parent class missing');
-            }
-
-            return $entityClass;
+            return get_parent_class($entity) ?: $entity::class;
```
weaverryan, line 25: If something implements `LazyObjectInterface` but it doesn't have a parent class, does returning `$entity::class` make sense? Or is that a misconfiguration of some sort (when *would* this really occur?)?

nicolas-grekas: it does yes, the interface doesn't have to be implemented by a proxy
this is mostly theory in the context of doctrine of course :)

weaverryan:
Thanks Jesse!

PR #1484, opened 2024-02-09
[Live] Fixed idiomorph id child handling & JavaScript cleanup

### LINKED ISSUE #1470: [LiveComponent] Sharing templates between child components broken since 2.14
Since live component 2.14 we can no longer use child components inside a parent component with the same template. The child components will generate the same live id.

I've created a reproducer here: https://github.com/tijnema/livebug
In the current state it shows how it should work (live components version 2.13), pressing the button will render child A or B. 
Upgrading live components to 2.14 breaks it, and will either render the wrong component, or render nothing at all.

### LINKED ISSUE #1472: [LiveComponent] Setting custom live-id in template not working anymore (since 2.14)
Hi,

We've been generating a custom live-id inside components for a while now, which fixes a lot of reloading issues (including #1470 ). Since release 2.14 however, this no longer works as expected.

I've created a reproducer project: https://github.com/tijnema/livebug2
Currently it is running with live components 2.13, pressing the button will reload the parent (render time will update), and childs are not being updated.
When upgrading the reproducer to 2.14, clicking the rerender button will reload all 100 child components (see the 102 network request in debug bar).

We set the live-id in the template like this:
```twig
<div {{ attributes.without('data-live-id').defaults({'data-live-id': this.liveId}) }}>
```
I haven't found a better way to set the live-id inside a component (docs mention only setting it as argument when creating the component).
Maybe live components can have a method for generating a live id?

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        |  Fix #1470 Fix #1472
| License       | MIT

Hi!

TODO:

* [x] Finish fixing minor js tests

Sorry, quite a lot going on here to get morphing & child rendering *just* right after switching to Idiomorph. The referenced tickets exposed a few cases that were not handled by the tests:

## IMPORTANT CHANGES

* A) `data-live-id` is now `id`. That's because Idiomorph uses `id` as its special key. So it's time to embrace that.
* B) Fixed a bug if `ChildComponentA` were changed to `ChildComponentB` in the same location. Under some situations, we would NOT swap for the new HTML, and would instead "refresh" `ChildComponentA`.
* C) In the above situation, even when we DID swap to `ChildComponentB` correctly, the `LiveController` would still be attached to a `Component` object representing `ChildComponentA`. That would cause problems later on. We now watch for the `id` attribute to change, and then re-create a new `Component` if changed.
* D) Removed special "morph" handling for children components (which are always empty). Instead added a new `data-live-preserve` attribute for elements that want to keep their current element. Improved handling of this.

## OTHER CHANGES

* Various internal attributes like `data-live-emit` were changed to Stimulus values, like `data-events-to-emit-value`. These are not attributes the user ever sets directly.
* Moved the remaining JS child handling into a plugin
* Added `morph:started` and `morph:finished` 
* Simplified "child" handling. Previously we really tried to keep track of our children as they were attached / detached. Now, if we need to know our children, we loop over all components to find children.

Cheers!

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.15.0
+
+-   [BC BREAK] The `data-live-id` attribute was changed to `id`.
+
 ## 2.14.1
 
 -   Fixed a regression in the testing tools related to the default HTTP
@@ -7,8 +11,10 @@
 
 ## 2.14.0
 
+-   [BC BREAK] DOM morphing changed from `morphdom` to `idiomorph`. As this is
+    a different morphing library, there may be some edge cases where the
+    morphing behavior is different.
 -   Add support for URL binding in `LiveProp`
--   DOM morphing changed from `morphdom` to `idiomorph`
 -   Allow multiple `LiveListener` attributes on a single method
 -   Requests to LiveComponent are sent as POST by default
 -   Add method prop to AsLiveComponent to still allow GET requests, usage: `#[AsLiveComponent(method: 'get')]`
--- a/src/LiveComponent/assets/dist/Component/ElementDriver.d.ts
+++ b/src/LiveComponent/assets/dist/Component/ElementDriver.d.ts
@@ -1,31 +1,30 @@
+import LiveControllerDefault from '../live_controller';
 export interface ElementDriver {
     getModelName(element: HTMLElement): string | null;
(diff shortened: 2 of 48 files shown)


smnandre:
Wow that's some major changes :) 

Every one of them make perfect sense and, as we said a _few times_ already, "better change while still experimental" :) 

Is this the perfect time to create our missing UPGRADE.md file ?

Issue #1390, opened 2024-01-12
[TwigComponent][LiveComponent] Compatibility with Twig 2.9 / use_yield  


smnandre:
See https://github.com/twigphp/Twig/pull/3950#issuecomment-1889971913 

For now i test with a local clone of the fabpot branch, and use_yield hard coded / changed in Environment class

### Work in progress

- [x] Adapt TwigComponent code [90% OK ](https://github.com/smnandre/ux/tree/sa/yield-template)
- [ ] Adapt LiveComponent code

### Todo/Later

- [ ] Use PrintNode/CaptureNode/etc instead of "echo" 

### PHPUnit log

<details>
<summary>TwigComponent failures</summary>

```

Time: 00:00.312, Memory: 34.00 MB

There was 1 error:

1) Symfony\UX\TwigComponent\Tests\Integration\EmbeddedComponentTest::testANonEmbeddedComponentCanRenderParentBlocksAsFallback

--

There were 20 failures:

1) Symfony\UX\TwigComponent\Tests\Integration\ComponentEventTest::testTemplateIsUpdatedByEventListener with data set "TWIG component tag" ('{% component "FooBar:Baz" %}{...ent %}')

src/TwigComponent/tests/Integration/ComponentEventTest.php:40

2) Symfony\UX\TwigComponent\Tests\Integration\ComponentEventTest::testTemplateIsUpdatedByEventListener with data set "HTML open-close tag" ('<twig:FooBar:Baz></twig:FooBar:Baz>')

src/TwigComponent/tests/Integration/ComponentEventTest.php:40

3) Symfony\UX\TwigComponent\Tests\Integration\ComponentExtensionTest::testCanRenderComponentWithMoreAdvancedTwigExpressions

src/TwigComponent/tests/Integration/ComponentExtensionTest.php:57

4) Symfony\UX\TwigComponent\Tests\Integration\ComponentExtensionTest::testCanRenderEmbeddedComponent

src/TwigComponent/tests/Integration/ComponentExtensionTest.php:156

5) Symfony\UX\TwigComponent\Tests\Integration\ComponentExtensionTest::testComponentPropsOverwriteContextValue

src/TwigComponent/tests/Integration/ComponentExtensionTest.php:208

6) Symfony\UX\TwigComponent\Tests\Integration\ComponentLexerTest::testComponentSyntaxOpenTags

src/TwigComponent/tests/Integration/ComponentLexerTest.php:28

[...]

PR #1478, opened 2024-02-08
[StimulusBundle/TwigComponent] Fix "use_yield" to "true" deprecation notice in CI

evertharmeling:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | 
| License       | MIT

Only this one remains (in StimulusBundle), but I can't find a direct implementation related to the `Environment` in that test. Maybe also related to https://github.com/symfony/symfony/pull/53805 or thoughts anyone?

```
1x: Since twig/twig 3.9.0: Not setting "use_yield" to "true" is deprecated.
    1x in StimulusControllerLoaderFunctionalTest::testFullApplicationLoad from Symfony\UX\StimulusBundle\Tests\AssetMapper
```

PR #1486, opened 2024-02-10
[Twig][Live] Skip Twig 3.9 🚒  (do not set `use_yield = true` for now)

smnandre:
As Symfony Twig Bridge & Twig Bundle have decided to skip Twig 3.9 for now, we should do the same, as testing and/or maintaining a compatible version would be really hard without the bundle support.

https://github.com/symfony/twig-bridge/commit/2abddb106523c53aebfe075c04a9d01bc8cdc5b4 
https://github.com/symfony/twig-bundle/commit/23a02ff1693c2c76a4fd55eae83704048d1e93e7

Update: 

Twig 3.9 introduce a major change in the way templates are rendered, and to fully test/fix our compatibility, we need to wait Bridge and Bundle to allow Twig 3.9

In the meantime, please do not set "use_yield = true" as we are not ready yet.

This is something that has a major impact on TwigComponent and LiveComponent, and we'll probably need some changes on the CI / test suite to handle both modes.

---

Pull Request (compatibility - wip) : https://github.com/symfony/ux/pull/1487 (help welcomed)
Issue: https://github.com/symfony/ux/issues/1390 
Twig PR : https://github.com/twigphp/Twig/pull/3950

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -29,7 +29,7 @@
         "php": ">=8.1",
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
         "symfony/ux-twig-component": "^2.8",
-        "twig/twig": "^2.14.7|^3.0.4"
+        "twig/twig": "^2.14.7|~3.8.0"
     },
     "require-dev": {
         "doctrine/annotations": "^1.0",
--- a/src/StimulusBundle/composer.json
+++ b/src/StimulusBundle/composer.json
@@ -18,7 +18,7 @@
         "symfony/dependency-injection": "^5.4|^6.0|^7.0",
         "symfony/finder": "^5.4|^6.0|^7.0",
         "symfony/http-kernel": "^5.4|^6.0|^7.0",
-        "twig/twig": "^2.15.3|^3.4.3",
+        "twig/twig": "^2.15.3|~3.8.0",
         "symfony/deprecation-contracts": "^2.0|^3.0"
     },
     "require-dev": {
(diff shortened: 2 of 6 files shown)


smnandre:
I "revert" https://github.com/symfony/ux/pull/1478 in this PR

Please do **_not_** set "use_yield = true" while our compatibility is not OK.

This is something that has a major impact on TwigComponent and LiveComponent, and we'll probably need some changes on the CI to handle both modes.

cc @weaverryan @kbond

src/LiveComponent/composer.json
```diff
@@ -29,7 +29,7 @@
         "php": ">=8.1",
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
         "symfony/ux-twig-component": "^2.8",
-        "twig/twig": "^2.14.7|^3.0.4"
+        "twig/twig": "^2.14.7|~3.8.0"
```
weaverryan, line 32: Minor - but technically we are still ok if you’re using 3.0.4. We just want to not allow 3.9. Perhaps we should write it to still allow the lower twig 3 versions?

weaverryan (commented):
Thanks for tracking in this

src/LiveComponent/composer.json
smnandre: How would you write this ? If we use `<=` it will allow twig 1.0 no ?

weaverryan: hmm, I see you've just matched what Symfony core is doing. Ok for me then!

weaverryan:
Thanks Simon!

PR #1494, opened 2024-02-14
Bootstrapping an UPGRADE log

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | None
| License       | MIT

As suggested by @smnandre, a CHANGELOG. This would be for things that user should actually change from one version to another. symfony/symfony maintains something similar.


smnandre:
Did you want to write "a UPGRADE " ? :p

PR #1501, opened 2024-02-15
[Live] Fixing bug where the active input would maintain its value, but lose its cursor position

### LINKED ISSUE #1490: [InlineEdit] Unable to edit text field as a prepend
Hello,

At https://ux.symfony.com/demos/live-component/inline-edit when positioning your cursors anywhere other than the end of the input field after your first keystroke the cursors jumps at the end. 

I wanted to type `Yellow Banana 🍌` and this is how it looks like on first key-stroke. To achieve a prepend, one has to move the cursor per letter.

Screenshot 2024-02-12 at 11 29 37

Unfortunately I cannot help with this since I am looking at the demos to get back into trying JS after a decade of not working with it, but I wanted to report the issue at least.

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1490
| License       | MIT

This only affects "model" elements, as unmapped element changes are tracked and the "to" element's value updated in the beforeNodeMorphed() callback.

Also, this setting matches Turbo 8. This also updates Idiomorph to the absolute latest version, which includes a related bug fix.

Cheers!

### DIFF
--- a/src/LiveComponent/assets/package.json
+++ b/src/LiveComponent/assets/package.json
@@ -27,7 +27,7 @@
         }
     },
     "dependencies": {
-        "idiomorph": "^0.3.0"
+        "idiomorph": "https://github.com/bigskysoftware/idiomorph.git"
     },
     "peerDependencies": {
         "@hotwired/stimulus": "^3.0.0"
--- a/src/LiveComponent/assets/src/morphdom.ts
+++ b/src/LiveComponent/assets/src/morphdom.ts
@@ -53,6 +53,12 @@ export function executeMorphdom(
     });
 
     Idiomorph.morph(rootFromElement, rootToElement, {
+        // We handle updating the value of fields that have been changed
+        // since the HTML was requested. However, the active element is
+        // a special case: replacing the value isn't enough. We need to
+        // prevent the value from being changed in the first place so the
+        // user's cursor position is maintained.
+        ignoreActiveValue: true,
         callbacks: {
             beforeNodeMorphed: (fromEl: Element, toEl: Element) => {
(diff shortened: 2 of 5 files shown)


src/LiveComponent/assets/package.json
```diff
@@ -27,7 +27,7 @@
         }
     },
     "dependencies": {
-        "idiomorph": "^0.3.0"
+        "idiomorph": "https://github.com/bigskysoftware/idiomorph.git"
```
smnandre, line 30: Sure about this ?

src/LiveComponent/assets/src/morphdom.ts
```diff
@@ -53,6 +53,12 @@ export function executeMorphdom(
     });
 
     Idiomorph.morph(rootFromElement, rootToElement, {
+        // We handle updating the value of fields that have been changed
+        // since the HTML was requested. However, the active element is
+        // a special case: replacing the value isn't enough. We need to
+        // prevent the value from being changed in the first place so the
+        // user's cursor position is maintained.
+        ignoreActiveValue: true,
```
smnandre, line 61: \o/ As we suspected  🕵️ 🕵️

smnandre:
Checked on the demos: 👍

src/LiveComponent/assets/package.json
weaverryan: Yup, but good question. Copied from Turbo - it allows us to get the absolute latest version. Will change after they tag next. We embed this library (for now), so this isn't exposed to the user.

Issue #1500, opened 2024-02-15
[Autocomplete][LiveComponent] Select box doesn't reset when used in a live component


pierredup:
When using the Autocomplete component in a form as part of a LiveComponent, the field is not reset when re-rendering the component (E.G when submitting the form). The previous selected values are still selected.

Here is an example using a multi-select dropdown, with a custom query builder that removes the selected items from the list:

https://github.com/symfony/ux/assets/144858/d0add8fa-9547-401d-99db-c9dabf357a3e

When changing the form to an Autocomplete type, and selecting items, it doesn't get removed from the select

https://github.com/symfony/ux/assets/144858/cb710386-64c8-4444-b480-e4ff1c6933a6

Currently in order to reset the autocomplete dropdown, I'm emitting a custom browser event with a custom controller on the select with the following which seems to work fine:

```js
import { Controller } from '@hotwired/stimulus';
import TomSelect from 'tom-select';
import { TomInput, TomSettings } from 'tom-select/dist/types/types';

/* stimulusFetch: 'lazy' */
export default class extends Controller {
    connect(): void {
        document.addEventListener('select:clear', (): void => {
            if ((this.element as TomInput).tomselect) {
                const settings: TomSettings|undefined = (this.element as TomInput).tomselect?.settings;
                (this.element as TomInput).tomselect?.destroy();
                new TomSelect(<TomInput>this.element, (settings as TomSettings));
            }
        });
    }
}
```

weaverryan:
Hey @pierredup!

Ok, let's see if we can fix this :). The underlying problem is that TomSelect does NOT like to reset itself when things change on the underlying `select` and `option` elements. So, ignoring TomSelect, if I understand things correctly, this is the situation:

A) A `<select multiple>` appears on the page with 10 options.
B) The user selects 2 of these: `Pan Jan` and `John Roberts` and the submits the form / triggers the `LiveAction`.
C) When then LiveComponent re-renders, the `<select multiple>` still on the page. But now it only has 8 options. Is that correct? And, perhaps also important, even if it did still have the original 10 options, `Pan Jan` and `John Roberts` would no longer be selected (if you try this iteration, I expect the problem will remain).

The autocomplete live controller has logic to try to determine when TomSelect needs to be reset - https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/src/controller.ts#L398-L401

[...]

PR #1502, opened 2024-02-15
[Autocomplete] Fix 2 bugs where TomSelect would reset when not necessary

### LINKED ISSUE #1499: [Autocomplete] + [LiveComponent] Infinite loop of requests with data-model on autocomplete.
Hi,

I've got a `LiveComponent` with two selects whose have `data-model` on them. First one to choose a "groupType", second one to choose a "companyGroup". "companyGroup" is the one with `autocomplete`. 

So when I change my groupType, my companyGroups list has to change too. **But** when I change the groupType, it actually launch an infinite loop of requests on my component... 

Here is my code : 

Component :

```
#[LiveProp(writable: true)]
public ?CompanyGroupType $companyGroupType = null;

#[LiveProp(writable: true)]
public ?CompanyGroup $companyGroup = null;

#[LiveAction]
public function onCompanyGroupTypeUpdated(): void
{
      $this->companyGroups = $this->groupRepository->findByType($this->companyGroupType);
      $this->companyGroup = $this->companyGroups[0];
}
```

Twig : 

```
<select class="form-control" data-model="companyGroupType" data-action="live#action" data-action-name="onCompanyGroupTypeUpdated">
    {% for type in groupTypes %}
        <option value="{{ type.id }}">
            {{ type.name }}
        </option>
    {% endfor %}
</select>
<select class="form-control my-1" data-model="companyGroup" {{ stimulus_controller('symfony/ux-autocomplete/autocomplete') }}>
    {% for group in companyGroups %}
        <option id="{{ group.id }}" value="{{ group.id }}">
            {{ group.name }}
        </option>
    {% endfor %}
</select>
```

Screen of the console after changed the first select : 

image

I tried with a `data-action` instead of `data-model="companyGroup"` to manually change companyGroup value, same problem. 

If I remove the `{{ stimulus_controller('symfony/ux-autocomplete/autocomplete') }}` everything work well (but no autocomplete...)

If i downgrade `symfony/ux-autocomplete` to 2.13.3 it works well too. 

I know autocomplete and liveComponent not work very well together, but maybe is there something I miss to make this work ? 

Thanks, 

Pierre

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1499
| License       | MIT

Hi!

The Autocomplete component "resets" the TomSelect instance under certain situations:

* A) The underlying `<option>` elements have changed
* B) The `<select>` changed its `multiple` attribute

Both of these situations had a bug:

* A) If there was no empty `<option value"">` element, then each time a value was selected, it incorrectly looked like the options had changed, triggering a reset.
* B) If the `<select multiple>` attribute was used without the `multiple="multiple"`, it would incorrectly look like the multiple value was changing, when in fact it was just varying between these 2 valid formats.

In both cases, when combined with LiveComponents, an infinite loop was triggered. That's because, as part of the reset process, when we recreate the TomSelect instance, we "select" the original value. This triggers a `change` event, causing LiveComponents to re-render. On its own , that's fine. But the re-render would trigger one of the bugs above, which would trigger another reset and another re-render.

Cheers!

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -116,6 +116,7 @@ class default_1 extends Controller {
     }
     resetTomSelect() {
         if (this.tomSelect) {
+            this.dispatchEvent('before-reset', { tomSelect: this.tomSelect });
             this.stopMutationObserver();
             const currentHtml = this.element.innerHTML;
             const currentValue = this.tomSelect.getValue();
@@ -143,6 +144,7 @@ class default_1 extends Controller {
                 subtree: true,
                 attributes: true,
                 characterData: true,
+                attributeOldValue: true,
             });
             this.isObserving = true;
         }
@@ -164,7 +166,11 @@ class default_1 extends Controller {
                         break;
                     }
                     if (mutation.target === this.element && mutation.attributeName === 'multiple') {
-                        requireReset = true;
+                        const isNowMultiple = this.element.hasAttribute('multiple');
+                        const wasMultiple = mutation.oldValue === 'multiple';
+                        if (isNowMultiple !== wasMultiple) {
+                            requireReset = true;
+                        }
                         break;
                     }
                     break;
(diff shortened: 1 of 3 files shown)


yceruto:
Hey Ryan! do you think this PR will also fix this issue https://github.com/symfony/ux/issues/1500?

weaverryan:
> Hey Ryan! do you think this PR will also fix this issue https://github.com/symfony/ux/issues/1500?

Checking that now. I don't think it will - this PR is about resetting *less*. That issue is about needing MORE resetting 🙃

PR #1503, opened 2024-02-15
[Twig][Live] Migrate PHPUnit configuration

smnandre:
```
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!
```

(will batch the other ones if there are no bad surprises in the CI with those ones)

### DIFF
--- a/src/LiveComponent/phpunit.xml.dist
+++ b/src/LiveComponent/phpunit.xml.dist
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


src/LiveComponent/phpunit.xml.dist
```diff
@@ -1,17 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="vendor/bin/.phpunit/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.6/phpunit.xsd"
```
norkunas: Why switching to hardcoded version? :)

smnandre: I'm glad you're here today 😅

weaverryan:
Thanks Simon!

PR #1505, opened 2024-02-15
[Autocomplete] Reset TomSelect when updating controller attributes

### LINKED ISSUE #1500: [Autocomplete][LiveComponent] Select box doesn't reset when used in a live component

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1500 
| License       | MIT

When using remote data for Autocomplete, the mutation observer is not started, so the select cannot be reset when some options change (E.G when limiting the options returned after selecting other options using extra_options from #1322).

Instead, we start the mutation observer when using remote data. When the observer runs and checks if the options have changes, we explicitly set the `areOptionsEquivalent` to `true`, since the page will just render an empty `select`. We also check if any of the `data-autocomplete` attributes changes and forces a reset (E.G using extra_options, the `data-autocomplete-url` value can be different when using extra options, so we reset the select when this attribute changes).

This fixes only one of the issues from #1500, I'll investigate the other issues and create separate PRs.

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -91,6 +91,9 @@ class default_1 extends Controller {
             }
         }
     }
+    urlValueChanged() {
+        this.resetTomSelect();
+    }
     getMaxOptions() {
         return this.selectElement ? this.selectElement.options.length : 50;
     }
@@ -131,7 +134,6 @@ class default_1 extends Controller {
             this.element.innerHTML = currentHtml;
             this.initializeTomSelect();
             this.tomSelect.setValue(currentValue);
-            this.startMutationObserver();
         }
     }
     changeTomSelectDisabledState(isDisabled) {
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -128,6 +128,10 @@ export default class extends Controller {
         }
     }
 
+    urlValueChanged() {
+        this.resetTomSelect();
+    }
+
     #getCommonConfig(): Partial<TomSettings> {
         const plugins: TPluginHash = {};
 
@@ -391,8 +395,6 @@ export default class extends Controller {
             this.element.innerHTML = currentHtml;
             this.initializeTomSelect();
             this.tomSelect.setValue(currentValue);
-
-            this.startMutationObserver();
         }
     }
 
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23.0
+
+-   Reset TomSelect when updating url attribute #1505
+
 ## 2.22.0
 
 -   Take `labelField` TomSelect option into account #2382
--- a/src/Autocomplete/assets/dist/controller.d.ts
+++ b/src/Autocomplete/assets/dist/controller.d.ts
@@ -40,6 +40,7 @@ export default class extends Controller {
     connect(): void;
     initializeTomSelect(): void;
     disconnect(): void;
+    urlValueChanged(): void;
     private getMaxOptions;
     get selectElement(): HTMLSelectElement | null;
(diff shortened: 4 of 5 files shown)


src/Autocomplete/assets/src/controller.ts
```diff
@@ -389,12 +393,21 @@ export default class extends Controller {
...
                     }
 
+                    if (
+                        mutation.target === this.element &&
+                        mutation.attributeName.match(/data-(symfony--ux-)?autocomplete/)
+                    ) {
+                        requireReset = true;
+
+                        break;
+                    }
```
weaverryan: There may be a simpler path: a stimulus value changed callback (or, really, one for each value we care about). The mutation observer is needed because we don’t have a way watch for option changes or things like the multiple attribute. But stimulus does give us a way to watch for value changes.

The only tricky part might be avoiding a “double” reset. For example, if some options are changed AND some values are changed at the same time, we only need to reset 1 time. If we use the value callbacks for one thing and the mutation observer for another, it may reset twice. Or if 3 values change, we don’t want to reset 3 times. Honestly, my idea sounds like more trouble. But since value changed callbacks are a standard solution, we should check into using them.

pierredup: That would work much better. I have added a `urlValueChanged` callback. Currently there is nothing else that triggers a reset when using remote data, so I don't think it's necessary to add a check to ensure we don't reset multiple times. When re-rendering and the url-value changes, the dropdown will only reset once.

src/Autocomplete/assets/dist/controller.js
```diff
@@ -123,7 +126,6 @@ class default_1 extends Controller {
             this.element.innerHTML = currentHtml;
             this.initializeTomSelect();
             this.tomSelect.setValue(currentValue);
-            this.startMutationObserver();
```
pierredup, line 134: The `this.initializeTomSelect()` call 2 lines above already starts the mutation observer (only under certain conditions, E.G when not using remote data). So this call incorrectly starts the mutation observer in the cases when it shouldn't run, which is why I removed the call here.

momocode-de:
I also need this fix. What is the status here?

arrabiata-asanz:
thank you to everyone that participated in the bug fix!
@weaverryan  Are you merging the branch soon?

smnandre:
Hey @pierredup, i'm reviving this PR, hoping it's still something you'd like to merge!

How "ready-to-merge" are these changes? Is there anything i need to check in particular? 

Thank you!

pierredup:
Hi @smnandre , there's nothing outstanding on this from my side, so I think it's ready to be merged

smnandre:
@pierredup great, LGTM!

Could you rebase and add an entry in the src/Autocomplete/CHANGELOG.md (2.23) ?

smnandre:
Thanks for your patience! :)

PR #1507, opened 2024-02-16
[LiveComponent] Add modifier option to LiveProp

squrious:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | N/A
| License       | MIT

Hi!

Following discussions in [#1396](https://github.com/symfony/ux/pull/1396#discussion_r1471304150), here is a proposal to allow a `LiveProp` to be modified at runtime.

The feature adds a `modifier` option to `LiveProp`, so users can use a custom method to modify the `LiveProp`'s options
depending on the context. 

This context could be a service, or even another prop within the component. 

### Usage

Here is the example given in the doc: 

```php
#[AsLiveComponent]
class ProductSearch
{
    #[LiveProp(writable: true, modifier: 'modifyAddedDate')]
    public ?\DateTimeImmutable $addedDate;

    #[LiveProp]
    public string $dateFormat = 'Y-m-d';

    // ...

    public function modifyAddedDate(LiveProp $prop): LiveProp
    {
        return $prop->withFormat($this->dateFormat);
    }
}
```

```twig
{{ component('ProductSearch', {
    dateFormat: 'd/m/Y'
}) }}
```

### Changes overview

First, we have to make the `LiveProp` attribute editable in some way. I opted out for immutable methods, so caching is preserved when applicable.

Then add a `LivePropMetadata::withModifier(object $component)` method that returns a new instance modified by the user-defined method. This is called in a few places, similarly to `LiveComponentMetadata::calculateFieldName()`.  

The tricky part is component hydration. If we use props in modifiers to change other `LiveProp` options, they have to be hydrated before the modifier is applied. This is only useful for options related to hydration, but it concerns most of them... The trick is to sort metadata, so modified ones are hydrated at the end.

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.17.0
+
+-   Add `modifier` option in `LiveProp` so options can be modified at runtime.
+
 ## 2.16.0
 
 -   LiveComponents is now stable and no longer experimental 🥳
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3450,6 +3450,53 @@ the change of one specific key::
         }
     }
 
+Set LiveProp Options Dynamically
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.17
+
+    The ``modifier`` option was added in LiveComponents 2.17.
+
+
+If you need to configure a LiveProp's options dynamically, you can use the ``modifier`` option to use a custom
+method in your component that returns a modified version of your LiveProp::
+
+
+    #[AsLiveComponent]
+    class ProductSearch
+    {
+        #[LiveProp(writable: true, modifier: 'modifyAddedDate')]
+        public ?\DateTimeImmutable $addedDate = null;
+
+        #[LiveProp]
+        public string $dateFormat = 'Y-m-d';
+
+        // ...
+
+        public function modifyAddedDate(LiveProp $prop): LiveProp
+        {
+            return $prop->withFormat($this->dateFormat);
+        }
+    }
+
+Then, when using your component in a template, you can change the date format used for ``$addedDate``:
+
+.. code-block:: twig
+
+    {{ component('ProductSearch', {
+        dateFormat: 'd/m/Y'
+    }) }}
+
+
+All ``LiveProp::with*`` methods are immutable, so you need to use their return value as your new LiveProp.
+
+.. caution::
+
+    Avoid relying on props that also use a modifier in other modifiers methods. For example, if the ``$dateFormat``
+    property above also had a ``modifier`` option, then it wouldn't be safe to reference it from the ``modifyAddedDate``
+    modifier method. This is because the ``$dateFormat`` property may not have been hydrated by this point.
+
+
 Debugging Components
 --------------------
 
--- a/src/LiveComponent/src/Attribute/LiveProp.php
+++ b/src/LiveComponent/src/Attribute/LiveProp.php
@@ -101,10 +101,19 @@ final class LiveProp
          * in the URL.
          */
         private bool $url = false,
+
+        /**
+         * A hook that will be called when this LiveProp is used.
+         *
+         * Allows to modify the LiveProp options depending on the context. The
+         * original LiveProp attribute instance will be passed as an argument to
+         * it.
+         *
+         * @var string|null
+         */
+        private string|null $modifier = null,
     ) {
-        if ($this->useSerializerForHydration && ($this->hydrateWith || $this->dehydrateWith)) {
-            throw new \InvalidArgumentException('Cannot use useSerializerForHydration with hydrateWith or dehydrateWith.');
-        }
+        self::validateHydrationStrategy($this);
     }
 
     /**
@@ -117,6 +126,14 @@ final class LiveProp
         }
 
         return \in_array(self::IDENTITY, $this->writable, true);
+    }
+
+    public function withWritable(bool|array $writable): self
+    {
+        $clone = clone $this;
+        $clone->writable = $writable;
+
+        return $clone;
     }
 
     /**
@@ -141,12 +158,32 @@ final class LiveProp
         return $this->hydrateWith ? trim($this->hydrateWith, '()') : null;
     }
 
+    public function withHydrateWith(?string $hydrateWith): self
+    {
+        $clone = clone $this;
+        $clone->hydrateWith = $hydrateWith;
+
+        self::validateHydrationStrategy($clone);
+
+        return $clone;
+    }
+
     /**
      * @internal
      */
     public function dehydrateMethod(): ?string
     {
         return $this->dehydrateWith ? trim($this->dehydrateWith, '()') : null;
+    }
+
+    public function withDehydrateWith(?string $dehydrateWith): self
+    {
+        $clone = clone $this;
+        $clone->dehydrateWith = $dehydrateWith;
+
+        self::validateHydrationStrategy($clone);
+
+        return $clone;
     }
 
     /**
@@ -157,12 +194,32 @@ final class LiveProp
         return $this->useSerializerForHydration;
     }
 
+    public function withUseSerializerForHydration(bool $userSerializerForHydration): self
+    {
+        $clone = clone $this;
+        $clone->useSerializerForHydration = $userSerializerForHydration;
+
+        self::validateHydrationStrategy($clone);
+
+        return $clone;
+    }
+
     /**
      * @internal
      */
     public function serializationContext(): array
     {
         return $this->serializationContext;
+    }
+
+    public function withSerializationContext(array $serializationContext): self
+    {
+        $clone = clone $this;
+        $clone->serializationContext = $serializationContext;
+
+        self::validateHydrationStrategy($clone);
+
+        return $clone;
     }
 
     /**
@@ -181,9 +238,25 @@ final class LiveProp
         return $this->fieldName;
     }
 
+    public function withFieldName(?string $fieldName): self
+    {
+        $clone = clone $this;
+        $clone->fieldName = $fieldName;
+
+        return $clone;
+    }
+
     public function format(): ?string
     {
         return $this->format;
+    }
+
+    public function withFormat(?string $format): self
+    {
+        $clone = clone $this;
+        $clone->format = $format;
+
+        return $clone;
     }
 
     public function acceptUpdatesFromParent(): bool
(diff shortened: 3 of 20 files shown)


src/LiveComponent/src/Attribute/LiveProp.php
```diff
@@ -198,8 +271,36 @@ public function onUpdated(): string|array|null
...
         return $this->url;
     }
+
+    public function withUrl(bool $url): self
+    {
+        $clone = clone $this;
+        $clone->url = $url;
+
+        return $clone;
+    }
```
WebMamba, line 291: Do we need all these methods? I don't see where we use them.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -145,7 +146,20 @@ public function hydrate(object $component, array $props, array $updatedProps, Li
...
+        $allMetadata = $componentMetadata->getAllLivePropsMetadata();
+        uasort($allMetadata, static function (LivePropMetadata $a, LivePropMetadata $b) {
+            if ($a->hasModifier() && !$b->hasModifier()) {
+                return 1;
+            } elseif (!$a->hasModifier() && $b->hasModifier()) {
+                return -1;
+            } else {
+                return 0;
+            }
+        });
```
WebMamba: Why do we need to put the method with a modifier in priority of others?

WebMamba (commented):
Ok, just let me summarize the approch here:
When a method have a modifier attribute, we execute the modifier function, get the modified value, we create a clone of the PropMetada with the new value. All of that happen during the hydratation. 
What about the dehydratation?

kbond:
I actually had a need recently to only enable live props that update the URL under certain conditions. I believe this PR would allow that, correct?

src/LiveComponent/src/Attribute/LiveProp.php
squrious: They are intended to be used by users in their modifiers, so I'd say yes!

src/LiveComponent/src/LiveComponentHydrator.php
squrious: Metadata with a modifier are considered "greater" than others, and the array is sorted in asc direction. I'd say this code prioritize metadata without modifier?

squrious:
> I actually had a need recently to only enable live props that update the URL under certain conditions. I believe this PR would allow that, correct?

Yes totally!

> When a method have a modifier attribute, we execute the modifier function, get the modified value, we create a clone of the PropMetada with the new value. All of that happen during the hydratation.
> What about the dehydratation?

Modifiers are not only used on hydration. Actually this is used a bit like `calculateFieldName`: we need it applied on hydration and dehydration, but also in other places like when processing/rendering URL mappings.

src/LiveComponent/doc/index.rst
```diff
@@ -3348,6 +3348,52 @@ the change of one specific key::
...
+        dateFormat: 'd/m/Y'
+    }) }}
+
+
+All ``LiveProp::with*`` methods are immutable, so you need to use their return value as your new LiveProp.
+
+.. caution::
+
+    You must not rely on props that also use a modifier in other modifiers methods, as their modifiers will not
+    necessarily be applied in the correct order when hydrating components.
```
```
You must not rely on properties that also use a modifier in other modifiers methods. This is because
those properties may not have been hydrated by this point.
```

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -66,6 +66,7 @@ public function dehydrate(object $component, ComponentAttributes $attributes, Li
 
         $dehydratedProps = new DehydratedProps();
         foreach ($componentMetadata->getAllLivePropsMetadata() as $propMetadata) {
+            $propMetadata = $propMetadata->withModifier($component);
```
weaverryan: This is the riskiest part of this PR. Currently, if you want the `LivePropMetadata[]` for a component, there is one source of truth: `getAllLivePropsMetadata()`. But now, someone needs to know to loop over these `LivePropMetadata` and call `withModifier($component)` to get the final version. That bothers me :/. I understand the "chicken and egg" problem, though: we need to hydrate the component (or as much of it as possible) before we can execute the modifier callbacks. I wish there was a way around that, but I can't think of anything more clever than your sorting algo.

However, to minimize the problem I mentioned above, would it be possible to add an `object $component` argument to `getAllLivePropsMetadata()` and make it return an iterable? Internally, it would do the sorting, then, one-by-one, call `->withModifier()` and yield the final `LiveProp`. Would that work? Then everywhere in the code would call `$componentMetadata->getAllLivePropsMetadata($component)`.

weaverryan (commented):
Very tricky PR - I'm hoping we can make this work - excellent job 🤞

src/LiveComponent/src/LiveComponentHydrator.php
squrious: Just tested this, and it works perfectly! 
But  I suggest that we sort these `LivePropMetadata` in the `LiveComponentMetadata` constructor, so they are sorted once.  As the `modifier` option is not editable, this should be fine, WDYT?

squrious:
Modifiers are now applied each time `getAllLivePropsMetadata` is called. Maybe we could add an internal cache for this, based on `spl_object_id`?

I also had to move the listener for query string initialization to `PostMount` instead of `PreMount`, as it needs the component to be "pre-mounted" to know if some props are bound with modifiers. The test for this was not right in the first place... I set the priority to 256 to ensure it is executed before other custom listeners.

Some tests failed after rebasing, but I don't think this is related to this PR.

src/LiveComponent/doc/index.rst
```diff
@@ -3348,6 +3348,52 @@ the change of one specific key::
         }
     }
 
+Set LiveProp Options Dynamically
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.15
```
weaverryan: 2.17 - I'm going to release 2.16 now without this, but then I think we can merge this very shortly so it's ready for 2.17.

src/LiveComponent/doc/index.rst
```diff
@@ -3348,6 +3348,52 @@ the change of one specific key::
...
+    {{ component('ProductSearch', {
+        dateFormat: 'd/m/Y'
+    }) }}
+
+
+All ``LiveProp::with*`` methods are immutable, so you need to use their return value as your new LiveProp.
+
+.. caution::
+
+    You must not rely on props that also use a modifier in other modifiers methods. This is because
```
```suggestion
    Avoid relying on props that also use a modifier in other modifiers methods. For example, if
    the `$dateFormat` property above also had a `modifier` option, then it wouldn't be safe
    to reference it from the `modifyAddedDate` modifier method. This is because the
    `$dateFormat` property may not have been hydrated by this point.
```

Often, I try to keep these notes short. But this is a tricky one. It will rarely be a problem... but I think the shorter version wasn't clear enough and could make some users worry they are doing something wrong when they are not.

src/LiveComponent/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
     new morphing library in 2.14.0 #1484
 -   Fix bug where the active input would maintain its value, but lose its cursor position #1501
 -   Restrict Twig 3.9 for now #1486
+-   Add `modifier` option in `LiveProp` so options can be modified at runtime.
```
weaverryan: You can move this up under a 2.17 header

src/LiveComponent/src/EventListener/QueryStringInitializeSubscriber.php
```diff
@@ -55,12 +58,20 @@ public function onPreMount(PreMountEvent $event): void
...
-        $event->setData(array_merge($event->getData(), $queryStringData));
+        $component = $event->getComponent();
+
+        foreach ($queryStringData as $name => $value) {
+            try {
+                $this->propertyAccessor->setValue($component, $name, $value);
+            } catch (PropertyAccessExceptionInterface $exception) {
+                // Ignore errors
+            }
+        }
```
weaverryan, line 73: It's unfortunate to need to do this by hand, but I think it's the only way 👍

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
```diff
@@ -105,11 +105,11 @@ public function attributesForRendering(MountedComponent $mounted, ComponentMetad
...
         }
 
-        if ($liveMetadata->hasQueryStringBindings()) {
+        if ($liveMetadata->hasQueryStringBindings($mounted->getComponent())) {
             $queryMapping = [];
-            foreach ($liveMetadata->getAllLivePropsMetadata() as $livePropMetadata) {
+            foreach ($liveMetadata->getAllLivePropsMetadata($mounted->getComponent()) as $livePropMetadata) {
                 if ($livePropMetadata->queryStringMapping()) {
-                    $frontendName = $livePropMetadata->calculateFieldName($mounted, $livePropMetadata->getName());
+                    $frontendName = $livePropMetadata->calculateFieldName($mounted->getComponent(), $livePropMetadata->getName());
```
weaverryan, line 110: I'm confused by this change. Rather, it looks right now, but the old code looks totally wrong. Was this an unrelated bug that snuck in? And is it not tested?

weaverryan (changes requested):

src/LiveComponent/src/LiveComponentHydrator.php
weaverryan: Yup, I think that's ok 👍

weaverryan:
> Modifiers are now applied each time getAllLivePropsMetadata is called. Maybe we could add an internal cache for this, based on spl_object_id?

It sounds nice... but I don't know if it will make enough difference. Most components will not have a modifier, so the extra work will be a `foreach` over all the LiveProps & a call to the `withModifier()` method that exits immediately. That might (?) be something we should cache away... I just don't know.

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
squrious: I was confused by the (my) old code too. It's a bug that wasn't covered because I used a string field name in tests, and the component object is only used when the field name is computed by a method... Not directly related to this PR though, but I thought fixing the typo was worth it.

squrious:
> It sounds nice... but I don't know if it will make enough difference. Most components will not have a modifier, so the extra work will be a `foreach` over all the LiveProps & a call to the `withModifier()` method that exits immediately. That might (?) be something we should cache away... I just don't know.

Yup, that's probably too much for the moment. I was wondering about modifiers that could use services, maybe heavy, to get configuration etc. But that could be cached in the component class directly. In the end, no need to generalize it :smile:

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
squrious: I added another prop in test component to test with the method for field name

weaverryan:
Thanks Nicolas for your hard and careful work on this - well done!

PR #1512, opened 2024-02-17
[WIP][Autocomplete] Adding morph support for Turbo & overhauling process

### LINKED ISSUE #1500: [Autocomplete][LiveComponent] Select box doesn't reset when used in a live component

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        | Fix #1500
| License       | MIT

This is WIP. The goal is to, during morphing, do both of the following:

1) Prevent morphing from removing any manual DOM changes made by TomSelect (which we want to keep)
2) Notice any changes (e.g. the underlying `<option>` elements change) that would require a TomSelect "reset" and doing that reset.

There are lot of permutations to get right, especially with (2), like making sure that both single & multiple `<select>` work as well as TomSelect used on an `<input>`.

Cheers!

### DIFF
--- a/src/Autocomplete/assets/dist/controller.d.ts
+++ b/src/Autocomplete/assets/dist/controller.d.ts
@@ -50,4 +50,6 @@ export default class extends Controller {
     private onMutations;
     private createOptionsDataStructure;
     private areOptionsEquivalent;
+    private beforeMorphElement;
+    private beforeMorphAttribute;
 }
--- a/src/Autocomplete/assets/dist/controller.d.ts
+++ b/src/Autocomplete/assets/dist/controller.d.ts
@@ -34,6 +34,7 @@ export default class extends Controller {
     private isObserving;
     private hasLoadedChoicesPreviously;
     private originalOptions;
+    private parentElement;
     initialize(): void;
     connect(): void;
     initializeTomSelect(): void;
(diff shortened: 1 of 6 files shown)


Kocal:

Thanks.

PR #1517, opened 2024-02-18
[Site] Replace stimulus_* helpers with data-attributes

### LINKED ISSUE #1514: [Site] Using raw data attributes
raw data attributes is a preferred way. The `stimulus_*` functions will then have to be replaced on https://ux.symfony.com/.

Reference: https://github.com/symfony/ux/pull/1336

What do you think ?

smnandre:
Fixes #1514 - cc @seb-jean 

@weaverryan  : I had to adapt a bit the SourceCleaner to keep data-* attributes.

### DIFF
--- a/ux.symfony.com/src/Model/RecipeFileTree.php
+++ b/ux.symfony.com/src/Model/RecipeFileTree.php
@@ -23,7 +23,7 @@ class RecipeFileTree
             ->addFile('assets/app.js', 'Your main JavaScript file. It\'s job is to import and load all other files.')
             ->addFile('assets/controllers.json', 'Configures 3rd-party Stimulus controllers. This file is automatically updated when you install a UX package.')
             ->addDirectory('assets/controllers', 'The home of your custom Stimulus controllers!')
-            ->addFile('assets/controllers/hello_controller.js', 'An example controller. Add it to any element with <code class="text-nowrap">{{ stimulus_controller(\'hello\') }}</code>')
+            ->addFile('assets/controllers/hello_controller.js', 'An example controller. Add it to any element with <code class="text-nowrap">data-controller="hello"</code>')
             ->addDirectory('assets/styles')
             ->addFile('assets/styles/app.css', 'Your main CSS file')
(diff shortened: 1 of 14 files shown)


seb-jean:
Great 💯

weaverryan (approved):

weaverryan:
Thanks Simon!

CMH-Benny:
May I ask what was the motivation of removing those helpers? One could use both notations but one is now just gone?

I used those helpers heavily, because especially the value-api is really verbose. You have to know how to convert the casing, need to put even multiple dashes here and there... It just makes the whole value API now harder to use imho. Not only will I have to refactor many places now, it will raise cognitive load for less afine collegues to parse for the values in the twig code to do adjustements.

Any chance we will get something else back that simplifies it as these helpers did before? @weaverryan @smnandre

weaverryan:
> I used those helpers heavily, because especially the value-api is really verbose

The helpers are still there in StimulusBundle and won't go away. So if you like them, keep using them. For most cases, I just now prefer to use the raw values attributes and I want to encourage people to give them a try. But sometimes they are a bit complex / ugly. Do whatever you like :)

CMH-Benny:
Oh, thanks for clarifying that! I misread that this is only about the usage of the helpers, but not the helpers themself :D 

I mean the helpers are also having their weaknesses, for example I had to merge two controllers with their values applied together, that is also kinda ugly, those cases are easier with the attributes, tho. But especially the values API is really suboptimal to write in an attribute way. Was there alternative ideas about those in the past? 

It probably doesn't fit here perfectly, but when we are on that topic - Would it be possible to have the twig parser take care of that and have somthing like this:
```twig
<div data-controller="a_controller"
     a_controller#aValueFromThatController="foo" 
     :a_controller#anotherValue="{{ something.number }}"
>
....
```

Or is this way too complicated?

smnandre:
Adding new syntax things can really increase the difficulty (not even talking about maintenance, doc, ..) so i'd like to avoid this.

But i do think some view helper can and should be used in some cases.

PR #1416, opened 2024-01-23
Introduce CVA to style TwigComponent

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | 
| License       | MIT

This PR introduces a new concept CVA (Class Variance Authority), by adding a1 twig function, to help you manage your class in your component.

Let's take an example an Alert component. In your app, an alert can have a lot of different styles one for success, one for alert, one for warning, and different sizes, with icons or not... You need something that lets you completely change the style of your component without creating a new component, and without creating too much complexity in your template.

Here is the reason came CVA.

Your Alert component can now look like this:

```twig
{% props color = 'blue', size = 'md' %}

{% set alert =  cva({
     base: 'alert rounded-lg',
     variants: {
        color: {
            blue: 'text-blue-800 bg-blue-50 dark:bg-gray-800 dark:text-blue-400',
            red: 'text-red-800 bg-red-50 dark:bg-gray-800 dark:text-red-400',
            green: 'text-green-800 bg-green-50 dark:bg-gray-800 dark:text-green-400',
            yellow: 'text-yellow-800 bg-yellow-50 dark:bg-gray-800 dark:text-yellow-400',
        },
        size: {
            sm: 'px-4 py-3 text-sm',
            md: 'px-6 py-4 text-base',
            lg: 'px-8 py-5 text-lg',
        }
    },
    compoundVariants: [{
           color: ['red'],
           size: ['lg'],
          class: 'font-semibold'
    }],
    defaultVariants: {
           rounded: 'md'
    }
}) %}

<div class="{{ cva.apply({color, size}, attribute.render('class'), 'flex p-4') }}">
    ...
</div>
```

So here you have a `cva` function that lets you define different variants of your component.

You can now use your component like this:
```twig
<twig:Alert color="red" size="md"/>

<twig:Alert color="green" size="sm"/>

<twig:Alert color="yellow" size="lg"/>

<twig:Alert color="red" size="md" class="dark:bg-gray-800"/>
```

And then you get the following result:

Capture d’écran 2024-01-24 à 00 52 33

If you want to know more about the concept I implement here you can look at:
- CVA (js version): https://cva.style/docs
- tailwind merge: https://github.com/gehrisandro/tailwind-merge-php, https://github.com/dcastil/tailwind-merge
- this implementation by using tailwind-merge and cva is inspired a lot by: https://ui.shadcn.com/ (shadcn is the most starred library on github in 2023) 
[...]

PR #1531, opened 2024-02-22
[Twig] `ComponentAttributes::append()/prepend()`

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

This is a followup to #1442.

When reviewing #1416, I noticed how using `~` to build attribute values with just `render()` was a bit annoying. This PR adds the following:

1. Default value to `render()` (`render('class', 'default')`) to avoid needing to use `render('class') ?? 'default'`
2. `prepend()`/`append()` to avoid needing to to use `class="{{ attributes.render('class') }} some defaults"`

Example:

```twig
{# templates/components/MyComponent.html.twig #}
<div
  class="{{ attributes.render('class', 'default') }}" {# second parameter is the default value #}
  style="{{ attributes.append('style', 'display:block;') }}" {# second parameter value to append #}
  data-controller="{{ attributes.append('data-controller', 'hello-controller') }}" {# second parameter value to prepend #}
  {{ attributes }} {# be sure to always render the remaining attributes! #}
>
  My Component!
</div>

{# render component #}
{{ component('MyComponent', { style: 'color:red;' }) }}

{# renders as: #}
<div class="default" style="color:red; display:block;" data-controller="hello-controller">
  My Component!
</div>
```

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -65,10 +65,10 @@ final class ComponentAttributes implements \IteratorAggregate, \Countable
         $this->rendered = [];
     }
 
-    public function render(string $attribute): ?string
+    public function render(string $attribute, ?string $default = null): ?string
     {
         if (null === $value = $this->attributes[$attribute] ?? null) {
-            return null;
+            return $default;
         }
 
         if (!\is_string($value)) {
@@ -78,6 +78,16 @@ final class ComponentAttributes implements \IteratorAggregate, \Countable
         $this->rendered[$attribute] = true;
 
         return $value;
+    }
+
+    public function prepend(string $attribute, string $value, string $separator = ' '): string
+    {
+        return trim(sprintf('%s%s%s', $value, $separator, $this->render($attribute)));
+    }
+
+    public function append(string $attribute, string $value, string $separator = ' '): string
+    {
+        return trim(sprintf('%s%s%s', $this->render($attribute), $separator, $value));
     }
 
     /**
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 2.16.0
+
+-   Add default value to `ComponentAttributes::render()` #1531
+-   Add `ComponentAttributes::append()/prepend()` #1531
+
 ## 2.15.0
 
 -   Add the ability to render specific attributes from the `attributes` variable #1442
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -952,14 +952,21 @@ Render
 
     The ability to *render* attributes was added in TwigComponents 2.15.
 
+.. versionadded:: 2.16
+
+    The ability to *render* attributes with a default, *prepend*, and *append* were
+    added in TwigComponents 2.16.
+
 You can take full control over the attributes that are rendered by using the
-``render()`` method.
+``render()``, ``prepend()``, and ``append()`` methods.
 
 .. code-block:: html+twig
 
     {# templates/components/MyComponent.html.twig #}
     <div
-      style="{{ attributes.render('style') }} display:block;"
+      class="{{ attributes.render('class', 'default') }}" {# second parameter is the default value #}
+      style="{{ attributes.append('style', 'display:block;') }}" {# second parameter value to append #}
+      data-controller="{{ attributes.append('data-controller', 'hello-controller') }}" {# second parameter value to prepend #}
       {{ attributes }} {# be sure to always render the remaining attributes! #}
     >
       My Component!
@@ -969,15 +976,15 @@ You can take full control over the attributes that are rendered by using the
     {{ component('MyComponent', { style: 'color:red;' }) }}
 
     {# renders as: #}
-    <div style="color:red; display:block;">
+    <div class="default" style="color:red; display:block;" data-controller="hello-controller">
       My Component!
     </div>
 
 .. caution::
 
-    There are a few important things to know about using ``render()``:
+    There are a few important things to know about using ``render()``/``prepend()``/``append()``:
 
-    1. You need to be sure to call your ``render()`` methods before calling ``{{ attributes }}`` or some
+    1. You need to be sure to call these methods before calling ``{{ attributes }}`` or some
        attributes could be rendered twice. For instance:
 
             .. code-block:: html+twig
@@ -998,7 +1005,7 @@ You can take full control over the attributes that are rendered by using the
                     My Component!
                 </div>
 
-    2. If you add an attribute without calling ``render()``, it will be rendered twice. For instance:
+    2. If you add an attribute without calling these methods, it will be rendered twice. For instance:
 
          .. code-block:: html+twig
 
--- a/src/TwigComponent/tests/Fixtures/templates/components/RenderAttributes.html.twig
+++ b/src/TwigComponent/tests/Fixtures/templates/components/RenderAttributes.html.twig
@@ -1,7 +1,10 @@
 <div
-    foo="{{ attributes.render('foo') }}"
-    bar="{{ attributes.render('bar')|default('default') }}"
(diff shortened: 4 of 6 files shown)


kbond:
When working on this I kept getting confused by append/prepend (what's prepended/appended, the real value or the default?).

I think in most cases the order here doesn't matter, or am I wrong? Could we reduce to a single method that just _joins_ a default value with the actual value? If so, what should this method be called?

smnandre:
prefix ?

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -80,6 +80,16 @@ public function render(string $attribute): ?string
         return $value;
     }
 
+    public function prepend(string $attribute, string $value, string $separator = ' '): string
+    {
+        return trim(sprintf('%s%s%s', $value, $separator, $this->render($attribute)));
```
smnandre, line 85: Genuine question, are the "trim" necessary ?

kbond: Nope not at all, I just dislike the attribute value being suffixed/prefixed by an empty space.

No problem for me to remove if desired.

smnandre: No it was really by curiosity

WebMamba (commented):
Haha thanks @kbond! I had the same idea while working on my PR! You are too fast! 😁

kbond:
> I think in most cases the order here doesn't matter, or am I wrong?

The order does indeed matter when using `tailwind_merge`.

smnandre:
So `.after('alert')` and `.before('activated')` ?

WebMamba:
I know this already merged but what do you think about the following syntax:
```twig
   <div
    class="{{ attributes.render('class', 'default') }}" {# second parameter is the default value #}
    style="{{ attributes.render.before('display:block;') }}" {# second parameter value to append #}
    data-controller="{{ attributes.render.after('hello-controller') }}" {# second parameter value to prepend #}
    {{ attributes }} {# be sure to always render the remaining attributes! #}
>
  My Component!
</div>
```

kbond:
> So .after('alert') and .before('activated') ?

Yes, that seems more explicit - maybe `.renderAfter()/renderBefore()`?

> I know this already merged but what do you think about the following syntax:

I originally had render return an object where we could chain these types of methods. I think it reads nicer too. We ended up dropping as it removed the possibility to do:

```
{{ attributes.render('class') ?? 'default' }}
```

Since render would always return the object `??` would never work.

Seeing this is a super new feature, we could probably get away with sneaking in this obscure BC break?

WebMamba:
I am voting to sneak a bit here 😈

kbond:
> I am voting to sneak a bit here

Me to, @weaverryan, @smnandre, any objections?

I'm thinking the following:

```twig
{# templates/components/MyComponent.html.twig #}
<div
  class="{{ attributes.render('class').default('default') }}"
  style="{{ attributes.render('style').before('display:block;') }}" 
  data-controller="{{ attributes.render('data-controller').after('hello-controller') }}"
  {{ attributes }}
>
  My Component!
</div>
```

smnandre:
> ```
>   class="{{ attributes.render('class').default('default') }}"
> ```

```
 class="{{ attributes.render('class') ?? 'default' }}"
```
 
 ?

smnandre:
I'd feel easier to get something like renderAfter / renderBefore but as you want

kbond:
> class="{{ attributes.render('class') ?? 'default' }}"
> ?

This would no longer work (and be the BC break here). render() would always return the new object.

> I'd feel easier to get something like renderAfter / renderBefore but as you want

I don't have a strong opinion but yes, renderAfter/renderBefore would be a bit simpler and not create a BC break.

smnandre:
Hmm... if you make this object countable that'd be helpful :)

WebMamba:
What do you think if the function renders instead of returning null return a new object called `ComponentAttribute`? This new ComponentAttribute object is Stringable. So we should not have a BC break. And we can put the function before and after on this class directly. I thinking about that because if I use the function render('class'), I expect something that represents the class attribute and nothing more, but if I understand well your implementation we get a clone of the full ComponentAttributes object. 
So with my proposal we can do:
```twig
// this work since ComponentAttribute is Stringable
class="{{ attributes.render('class') ?? 'default' }}"

// but also all of this
<div
  class="{{ attributes.render('class').default('default') }}"
  style="{{ attributes.render('style').before('display:block;') }}" 
  data-controller="{{ attributes.render('data-controller').after('hello-controller') }}"
  {{ attributes }}
>
  My Component!
</div>
```

so the function default, before, and after will be on the ComponentAttribute object

kbond:
Wouldn't it need to be changed to the following?

```diff
- {{ attributes.render('class') ?? 'default' }}
+ {{ attributes.render('class') ?: 'default' }}
```

Also, and we'd need to double check, but at the time `??` or `?:` is evaluated, it would be an object still, wouldn't it?

WebMamba:
Ha yes sorry, so the function render returns null or the new ComponentAttribute object.

smnandre:
If you implement countable the ?: would work

kbond:
> so the function render returns null or the new ComponentAttribute object.

I don't want it to return null as that would require an extra check when chaining .before()/after().

> If you implement countable the ?: would work

Why would it work for countable but not a stringable object that returns an empty string? I'm not a huge fan of making it countable just for this purpose.

smnandre:
Oh it was just to give a fun fact :)) I'm 100% ok with your proposal :) 

Count is called before empty, i discoverd that after a loooong afternoon of debugging (when Jane Automapper started to use ArrayObject with no public properties in a minor version 😅.... making every |default working in the opposite way than excpected :p )

smnandre:
After a while i'm starting to thing.... why passing to a method something that can be direct in the HTML ? 

```twig
<div style="{{ attributes.render('style').before('display:block;') }}" >
  ```
is much easier to read as 

```twig
<div style="display:block; {{ attributes.render('style') }}">
  ```  
  
 Same for "after"... no ?

**Comment by smnandre** [comment]
And what could improve DX would be magic getter i think

```twig
<div style="display:block; {{ attributes.style }}" class="{{ attributes.class }}">
```

kbond:
> Same for "after"... no ?

Totally valid and maybe preferred by some. The primary reason I added was to make it _prettier_ when using something like tailwind_merge.

> And what could improve DX would be magic getter i think

One problem with using the magic methods to avoid calling ->render() is there's no longer an error for a bad method call (ie a typo in defaults()): {{ attributes.dfaults({class: 'block'}) }} === null

smnandre:
Ok lets Forget then :)

weaverryan:
Hi!

A) The 2nd arg of `render()` (a default value) was motivated by the `~` usage in #1416 - `{% set alert = cva('alert ' ~ attributes.render('class'), {`. But I have an alternate idea https://github.com/symfony/ux/pull/1416#discussion_r1504966227

B) The prepend/append idea seems to be entirely wanted / needed JUST for `tailwind_merge`. In all other situations, something like `style="display:block; {{ attributes.style }}"` I think is totally fine. For `tailwind_merge`, the current syntax would require:

```twig
class="{{ (attributes.render('class')~' bg-red-500')|tailwind_merge }}"
```

That *is* a bit ugly :). Could `tailwind_merge` perhaps be offered also as a function with unlimited args (note, this is how the underlying php library is meant to be used).

```twig
class="{{ tailwind_merge(attributes.render('class'), 'bg-red-500') }}"
```

Or this?

```twig
class="{{ [attributes.render('class'), 'bg-red-500']|tailwind_merge }}"
```

`tailwind_merge` already supports receiving an array... and the underlying library supports even those arrays being string with spaces in them - e.g. `$tw->merge(['text-red-500 text-sm'], 'text-blue-500');`.

kbond:
I agree, I like the idea of unlimited arguments in both cva and tailwind_merge. It really seems like the most explicit way to prepend/append when compared to:

> When working on this I kept getting confused by append/prepend (what's prepended/appended, the real value or the default?).

kbond:
Closing as there are, I think, better solutions.

PR #1536, opened 2024-02-22
[TwigComponent] Render in PostRenderEvent

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| License       | MIT

Pass the rendered content in the Event
(not available for the `embedded` components)

Similar to what's done in HttpFramework with the ResponseEvent


kbond:
Can you show a use-case for this? Would this help with what we do in #1450?

smnandre:
For a single icon i don't think so... as we bypass the component system to call a function directly.

For the website [yes](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/Attribute/Cache.php) :)

PR #1538, opened 2024-02-22
[Doc] Add more examples of adding attributes in Live Components

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

Note: I tried this in a Symfony app and it didn't work ... so I'm opening this PR to ask you if this should work or if `addAttribute()` is only for empty attributes that don't define a value.

Thanks!

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -929,15 +929,21 @@ a class:
     <div data-loading="removeClass(opacity-50)">...</div>
 
     <!-- add multiple classes when loading -->
-    <div data-loading="addClass(opacity-50 disabled)">...</div>
+    <div data-loading="addClass(opacity-50 text-muted)">...</div>
 
-Sometimes you may want to add or remove an attribute when loading. That
+Sometimes you may want to add or remove an HTML attribute when loading. That
 can be accomplished with ``addAttribute`` or ``removeAttribute``:
 
 .. code-block:: html+twig
 
     <!-- add the "disabled" attribute when loading -->
     <div data-loading="addAttribute(disabled)">...</div>
+
+.. caution::
+
+    The ``addAttribute()`` and ``removeAttribute()`` functions only work with
+    empty HTML attributes (``disabled``, ``readonly``, ``required``, etc.) and
+    not with attributes that define their values (e.g. this won't work: ``addAttribute(style='color: red')``).
 
 You can also combine any number of directives by separating them with a
 space:


smnandre:
It only works with attributes in a toggle way .. you may want to use addClass there i think

javiereguiluz:
Thanks for the review. I updated the PR to show an example that won't work and explain that explicitly. Thanks.

src/LiveComponent/doc/index.rst
```diff
@@ -931,14 +931,18 @@ a class:
     <!-- add multiple classes when loading -->
     <div data-loading="addClass(opacity-50 disabled)">...</div>
```
smnandre: Realizing after your modification we should not use "disabled" both in addClass and addAttribute, it does not help.. Could 
```suggestion
    <div data-loading="addClass(opacity-50 text-warning)">...</div>
```

smnandre (approved):

src/LiveComponent/doc/index.rst
```diff
@@ -931,14 +931,18 @@ a class:
...
-Sometimes you may want to add or remove an attribute when loading. That
+Sometimes you may want to add or remove an HTML attribute when loading. That
 can be accomplished with ``addAttribute`` or ``removeAttribute``:
 
 .. code-block:: html+twig
 
     <!-- add the "disabled" attribute when loading -->
     <div data-loading="addAttribute(disabled)">...</div>
 
+    <!-- this feature only works for empty attributes that don't define any value;
```
kbond: What about wrapping this in a caution annotation? I'm worried readers might skip over _the following example won't work_ and think it's valid

weaverryan: I agree - excellent detail to point out

weaverryan (changes requested):

javiereguiluz:
Sorry I forgot about this PR. I did all the changes that you suggested. Thanks for the reviews!

kbond:
> I did all the changes that you suggested.

@javiereguiluz, did you miss pushing these changes?

javiereguiluz:
@kbond yes, sorry 😊

kbond (approved):

src/LiveComponent/doc/index.rst
smnandre: Missed a bit of message there, thank you @javiereguiluz for guessing the end of my sentence and fixing it ❤️

weaverryan:
Thanks Javier!

PR #1539, opened 2024-02-22
[Doc] Reword the section about the locale in Live Components

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

Context: we have a localized project that doesn't add the locale in the URL. When refreshing the page after any UX change, the local was lost.

I read the docs but I couldn't fully understand it. I thought this was only for projects that include the locale in the URLs. In any case, after asking to Kevin and Ryan, they told me that this solution also applied to my case. They were right ... so I propose a minor reword to make it more clear that this solution is for everybody that works on localized projects.

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -96,9 +96,9 @@ needed if you're using AssetMapper):
     $ yarn install --force
     $ yarn watch
 
-In case your project `localizes its URLs`_ by adding the special
-``{_locale}`` parameter to the paths of its route definitions,
-you need to do the same with the UX Live Components route definition:
+If your project is localized in different languages (either via the `_locale route parameter`_
+or by `setting the locale in the request`_) add the ``{ _locale}`` attribute to
+the UX Live Components route definition to keep the locale between re-renders:
 
 .. code-block:: diff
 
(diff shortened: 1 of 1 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -96,9 +96,9 @@ needed if you're using AssetMapper):
     $ yarn install --force
     $ yarn watch
 
-In case your project `localizes its URLs`_ by adding the special
-``{_locale}`` parameter to the paths of its route definitions,
-you need to do the same with the UX Live Components route definition:
+If your project is localized in different languages (either via the `_locale route parameter`_
+or by `setting the locale in the request`_) add the ``{ _locale}`` attribute to
+the UX Live Components route definition to keep the locale between refreshes:
```
```suggestion
the UX Live Components route definition to keep the locale between re-renders:
```

kbond (changes requested):
Looks good! Can you remove the following line to make the CI happy?

```
.. _`localizes its URLs`: https://symfony.com/doc/current/translation/locale.html#translation-locale-url
```

smnandre (approved):

src/LiveComponent/doc/index.rst
javiereguiluz: Fixed! Thanks Kevin.

kbond (approved):

kbond:
Thank you Javier.

PR #1549, opened 2024-02-26
[LiveComponent] Add doc about the placeholder macro

smnandre:
Add documentation about  #1532 (placeholder macro)

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2242,8 +2242,12 @@ Deferring / Lazy Loading Components
 If a component is heavy to render, you can defer rendering it until after
 the page has loaded. To do this, add the ``defer`` option:
 
-.. code-block:: twig
+.. code-block:: html+twig
 
+    {# With the HTML syntax #}
+    <twig:SomeHeavyComponent defer />
+
+    {# With the component function #}
     {{ component('SomeHeavyComponent', { defer: true }) }}
 
 This renders an empty ``<div>`` tag, but triggers an Ajax call to render the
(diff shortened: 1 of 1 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -2255,26 +2259,88 @@ real component once the page has loaded.
...
+
+    {# With the component tag #}
     {% component SomeHeavyComponent with { defer: true } %}
         {% block loadingContent %}Loading...{% endblock %}
     {% endcomponent %}
 
-To change the initial tag from a ``div`` to something else, use the ``loading-tag`` option:
+Both ``loading-template`` and ``loadingContent`` can be used in the template
+_calling_ the component. But if want to set this content only once, you can add a
+``placeholder`` macro directly in the component template.
```
weaverryan: macro called placeholder

that makes it more clear that placeholder with name vs placeholder is an adjective

src/LiveComponent/doc/index.rst
```diff
@@ -2255,26 +2259,88 @@ real component once the page has loaded.
...
+        {% for i in 1..props.size %}
+            <div class="loading-product">
+                ...
+            </div>
+        {% endfor %}
+    {% endmacro %}
+
+If you style the placeholder content (giving the rows a specific height),
+you can then avoid layout shifts when the component is fully loaded (and
+the loading content replaced by the real content).
```
weaverryan: I think it’s a bit too long - the first example is probably enough. 

Also, I’m wondering if this should license the FIRST option of customizing loading content. I’d imagine that I would more often embed the loading content inside the template vs in the calling template.

so logic would be:

- to customize loading, you can add this macro
- Else, if you want to control loading when calling the component, use the block trick.
- And, btw, you can control the loading tag for edge cases, like when your element is a tr

src/LiveComponent/doc/index.rst
smnandre: "named" placeholder ?

src/LiveComponent/doc/index.rst
smnandre: I've reordered all .. tell me if better

weaverryan:
Very nice - thank you Simon!

PR #1551, opened 2024-02-27
[CI] Configure 'test' workflow to ignore doc updates

smnandre:
No need to run full tests when a PR update only doc


weaverryan:
Thanks Simon!

PR #1552, opened 2024-02-27
[TwigComponent] Fix full context is stored in profiler

### LINKED ISSUE #1540: [TwigComponent] Profiler, memory consumption
Hi,

https://github.com/symfony/ux/blob/b81fe2cbbb0bf750141875f53165f3464b2115b7/src/TwigComponent/src/DataCollector/TwigComponentDataCollector.php#L168

I noticed that this line causes consumption of a lot of memory when the profiler is enabled, so we have to increase the php memory_limit. 

Thanks

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix [#1540](https://github.com/symfony/ux/issues/1540)
| License       | MIT

As `$event->getVariables()` get the full context for non-embed components, this led to huge data stored for every component called during the request. (see [#1540](https://github.com/symfony/ux/issues/1540))

As this is internal data, and the "variables" was not used in the profiler, I believe we should consider this a bug and simply remove it.


weaverryan:
Thanks Simon!

alinceDev:
Thanks Simon !

PR #1562, opened 2024-03-01
[Interactive form validation] New interactive form validation package

andersmateusz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

Draft pull request for a new package. I am open for suggestions.
Funcionality: 
 - Automatic focus on invalid input/form and scroll into the invalid input/form. 
 - Alert that notifies user that the form is invalid and why. Two strategies are available. Just emit event with message (then you must implement alert yourself with usage of many alert js libraries available) or show browser native alert.
 
This should be working well with full page reloads as well with turbo bundle streams and frames.
Global configuration (for each form) and for each form separately is possible.
It is recommended to set form attr `novalidate`  on a form to fully test this functionality. Wondering if novalidate attr should not be set by default on each form with this feature enabled.


Kocal:
Hi, and thanks for the contribution!
I think this feature is already covered by LiveComponent (which support Form & Form validation), or even with Turbo.

I'm closing, but feel free to re-open if necessary.
Thanks!

PR #1571, opened 2024-03-01
[TwigComponent] attributes.has() method

barbieswimcrew:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Issues        | -
| License       | MIT

Provide a `ComponentAttributes::has()` function to be able to check via `{{ attributes.has('foo') }}` whether the attribute has been passed. 

I would like to show the use case in the following example. Using `attributes.has('href')`, I check whether the TwigComponent has set an 'href' attribute when it is called. If so, an `<a>` tag should be rendered, otherwise a `<button>` tag.

```twig
{% set element = attributes.has('href') ? 'a' : 'button' %}

<{{ element }} {{ attributes }}>
</{{ element }}>
```

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -193,6 +193,11 @@ final class ComponentAttributes implements \IteratorAggregate, \Countable
         return new \ArrayIterator($this->attributes);
     }
 
+    public function has(string $attribute): bool
+    {
+        return array_key_exists($attribute, $this->attributes);
+    }
+
     public function count(): int
     {
         return \count($this->attributes);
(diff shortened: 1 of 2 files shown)


smnandre:
Could we implement ArrayAccess to simplify all this ?

kbond:
Solid use case. Trying to think if there's any downsides to @smnandre's suggestion.

barbieswimcrew:
> Solid use case. Trying to think if there's any downsides to @smnandre's suggestion.

Hmm... let's say we would implement ArrayAccess... would the add() and offsetSet() methods become one? And if so... what would someone expect when using offsetSet(), for example? That you can add any attribute from inside? And what happens then with the custom logic from the current add() method? 

And remove() would therefore become offsetUnset(), wouldn't it?

smnandre:
I'd implement only offsetExists and offsetGet, and throw a runtime exception in the two others.. 

it would "just" be syntaxic sugar to ease usage in templates :)

weaverryan (approved):

weaverryan:
Because ComponentAttributes is a bit complex and we've been working a lot with subtle additions to it lately, I think we should keep it simple and just add has(). Thanks @barbieswimcrew!

PR #1572, opened 2024-03-01
[Live] Fixing lost behavior of handling element with an action and a model on it

### LINKED ISSUE #1557: [LiveComponent] Updated values not submitted to the server properly
I've got an update problem, with 2.14.2 version of twig+live components. The use case is: a Sale Order has many items. I'm using a Live Component + Form + LiveCollectionType.
Every item has a "product", a "price", a "quantity" and an "untaxedTotal". Ideally when you select a product, the price is updated with the product salePrice, and when you change quantity or price on the SaleOrderItem, the untaxedTotal is also updated.

OrderComponent
```
#[AsLiveComponent]
final class OrderComponent extends AbstractController
{
    use ComponentWithFormTrait;
    use DefaultActionTrait;
    use LiveCollectionTrait;

    #[LiveProp(writable: true)]
    public ?SaleOrderInput $initialFormData = null;

    {
        return $this->createForm(OrderType::class, $this->initialFormData);
    }

    #[LiveAction]
    public function updateUnitPrice(): void
    {
        foreach ($this->formValues['items'] as &$item) {
            $item['unitPrice'] = rand(0, 40);
            if (empty($item['quantity'])) {
                $item['quantity'] = 1;
            }
            $this->setUntaxedTotal($item);
        }
    }

    #[LiveAction]
    public function updateUntaxedTotal(): void
    {
        foreach ($this->formValues['items'] as &$item) {
            $this->setUntaxedTotal($item);
        }
    }

    private function setUntaxedTotal(array &$item): void
    {
        $item['untaxedTotal'] = $item['unitPrice'] * $item['quantity'] - $item['discount'];
    }
}
```

and the relative form (simplified):

```
class OrderType extends AbstractType
{
    {
        $builder
            ->add('items', LiveCollectionType::class, [
                'entry_type' => OrderItemType::class,
                'entry_options' => ['label' => false],
                'label' => false,
                'allow_add' => true,
                'allow_delete' => true,
                'by_reference' => false,
            ])
        ;
    }

    {
        $resolver->setDefaults([
            'data_class' => SaleOrderInput::class
        ]);
    }
}
```

also the OrderItemType:

```
class OrderItemType extends AbstractType
{
    /** @param array<string, mixed> $options */
    {
        $builder = new DynamicFormBuilder($builder);

        $builder
            ->add('product', EntityType::class, [
                'class' => Product::class,
                'required' => false,
                'label' => 'wd.field.product',
                'autocomplete' => true,
                'attr' => [
                    'placeholder' => 'wd.field.product',
                    'data-action' => 'live#action',
                    'data-action-name' => 'updateUnitPrice',
                ],
            ])
            ->add('quantity', NumberType::class, [
                'label' => 'wd.field.quantity',
                'scale' => 1,
                'required' => true,
                'html5' => true,
                'attr' => [
                    'data-action' => 'live#action',
                    'data-action-name' => 'updateUntaxedTotal',
                ],
            ])
            ->add('unitPrice', NumberType::class, [
                'label' => 'wd.field.unitPrice',
                'scale' => 1,
                'required' => true,
                'html5' => true,
                'attr' => [
                    'data-action' => 'live#action',
                    'data-action-name' => 'updateUntaxedTotal',
                ],
            ])
            ->add('untaxedTotal', NumberType::class, [
                'label' => 'wd.field.total',
                'attr' => [
                    'readonly' => true,
                ],
            ])
        ;

        [..]
    }

    {
        $resolver->setDefaults([
            'data_class' => SaleOrderItemInput::class,
        ]);
    }
}
```

When i select a product all the form fields are updated correctly. The problem is that when i try to change manually quantity or unitPrice, the untaxedTotal stays always "one step back". 
Let's say that unitPrice is 1 and quantity is 1, if i click to increase the NumberType or if i manually edit the number to "2" **a UX ajax call is made, with an empty "updated" object**. On the second change (to "3") another call is made with an updated object containing "2" as the new value. This is true with an "html5"-> false setting, or even with a TextType as form field.

[Something funny](https://stackoverflow.com/questions/28955330/input-type-number-value-does-not-update-in-html-is-this-normal) could be going on between HTML/DOM and UX, because on the first change the value attribute in the html "rightfully" remains "one step behind" (so on the first change, the value remains "1" in the HTML). So maybe i'm wrong using a LiveAction to update the form data for this use case, or some additional javascript is required to force the value of the HTML to match the one displayed, or UX is not catching the proper updated value correctly. In the most basic test "add a collection element"->"select a product"->"increase the unitPrice by 1" it seems to me that UX should send the updated value inside the "updated" object of the ajax call. Any ideas about why this is happening?

EDIT: i actually built a simple prototype, and the behaviour is the same. On further testing, if i keep the focus on the fields and as long as i edit them by typing, all the ajax calls are made, but the value is never changed, even after N edits. If i edit the field, lose the focus, and edit it again, the values start to change (but always "one step back"). I also tried a different syntax to bind to different events (eg. "keyup->live#action") but with no change, i doubt that the problem is event binding, because when i edit the field the ajax calls are always launched correctly, even though with "not updated data". This is also true with twig/live components 2.15.0

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1557
| License       | MIT

UPDATE

I think this fix is not needed at all - I think #1557 happened just due to the "gotcha" listed below. If you fix that, things DO work... and we even have a test for this :). I've reverted my fix... this is now just a docs PR.

------------------

This is old behavior that we once had, but got lost in a reshuffle of code a long time ago. It's an edge case. Imagine:

```html
<input data-model="count" data-action="live#action" data-live-action-param="doSomething">
```

This element is both bound to a `count` model *and* should trigger an action. Due to how the events are registered, here is the natural order:

1. The `action()` is triggered
2. THEN the model is updated

The result is that the `doSomething` `LiveAction` is triggered... but *without* the updated `count` value. Our fix is to detect this situation and add a slightly delay (waiting for the potential model update) before triggering the action.

There is one gotcha that will need to be documented: if you (1) trigger the model update on `change` but trigger the action on `input`, then the action fires quite a bit before the model is updated on `change`. This is a misconfiguration, but easy to do with the form system, since `change` is the default, but any custom `data-action` on an input will default to the `input` event.

### TODO 

* [x] Add a spot in the docs to describe this & the gotcha
* [ ] ~~Add tests~~

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.d.ts
+++ b/src/LiveComponent/assets/dist/live_controller.d.ts
@@ -87,6 +87,8 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
     private mutationObserver;
     component: Component;
     pendingActionTriggerModelElement: HTMLElement | null;
+    pendingActionCallable: (() => void) | null;
+    pendingActionTimeout: number;
     private elementEventListeners;
     private pendingFiles;
     static backendFactory: (controller: LiveControllerDefault) => BackendInterface;
--- a/src/LiveComponent/assets/dist/live_controller.d.ts
+++ b/src/LiveComponent/assets/dist/live_controller.d.ts
@@ -87,8 +87,6 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
     private mutationObserver;
     component: Component;
     pendingActionTriggerModelElement: HTMLElement | null;
-    pendingActionCallable: (() => void) | null;
-    pendingActionTimeout: number;
     private elementEventListeners;
     private pendingFiles;
     static backendFactory: (controller: LiveControllerDefault) => BackendInterface;
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -2918,6 +2918,8 @@ class LiveControllerDefault extends Controller {
     constructor() {
         super(...arguments);
         this.pendingActionTriggerModelElement = null;
+        this.pendingActionCallable = null;
+        this.pendingActionTimeout = 0;
         this.elementEventListeners = [
             { event: 'input', callback: (event) => this.handleInputEvent(event) },
             { event: 'change', callback: (event) => this.handleChangeEvent(event) },
@@ -2991,10 +2993,23 @@ class LiveControllerDefault extends Controller {
                 }
                 delete this.pendingFiles[key];
             }
-            this.component.action(directive.action, actionArgs, debounce);
-            if (getModelDirectiveFromElement(event.currentTarget, false)) {
-                this.pendingActionTriggerModelElement = event.currentTarget;
+            this.pendingActionCallable = () => {
+                clearTimeout(this.pendingActionTimeout);
+                this.pendingActionCallable = null;
+                this.pendingActionTriggerModelElement = null;
+                this.component.action(directive.action, actionArgs, debounce);
+            };
+            if (!getModelDirectiveFromElement(event.currentTarget, false)) {
+                this.pendingActionCallable();
+                return;
             }
+            this.pendingActionTriggerModelElement = event.currentTarget;
+            this.pendingActionTimeout = window.setTimeout(() => {
+                if (!this.pendingActionCallable) {
+                    return;
+                }
+                this.pendingActionCallable();
+            }, 10);
         });
     }
     $render() {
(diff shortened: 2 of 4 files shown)


smnandre:
https://github.com/symfony/ux/issues/1648

Kocal:

Thanks.

PR #1574, opened 2024-03-01
[Live] Adding an example of a norender model with a button to trigger the render

### LINKED ISSUE #1569: Explicit re-render of a field when clicking a button
I have a complex component with many form elements (it's not a Symfony form).

When I change `<select>` and `<input>` elements, the page re-renders as expected. But then, I have this other element:

image

The input text is a live prop and I only want the page to re-render when the user clicks on the related button.

I can't do that. I read the docs and followed this: https://symfony.com/bundles/ux-live-component/current/index.html#lazy-updating-on-change-of-a-field

So, I applied the `<input data-model="on(change)|coupon">` syntax. It almost works:

* When you type the coupon code, the page doesn't re-render
* When you click on the button, the page re-renders
* BUT ... there's an edge-case: you change the contents of the coupon input and don't click on the button ... you change other input and then the coupon is applied to that re-render. It's strange because two different things are changing at the same time

I'd need something like this:

`<input data-model="on(click=couponButton)|coupon">`

Thanks!

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #1569
| License       | MIT

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -370,6 +370,29 @@ it won't, yet, make an Ajax call to re-render the component. Whenever
 the next re-render *does* happen, the updated ``max`` value will be
 used.
 
+This can be useful along with a button that triggers a render on click:
+
+.. code-block:: html+twig
+
+    <input data-model="norender|coupon">
+    <button data-action="live#$render">Apply</button>
+
+Forcing a Re-Render Explicitly
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some cases, you might want to force a component re-render explicitly. For
+example, consider a checkout component that provides a coupon input that must
+only be used when clicking on the associated "Apply coupon" button:
+
+.. code-block:: html+twig
+
+    <input data-model="norender|coupon">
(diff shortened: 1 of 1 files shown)


kbond (approved):

javiereguiluz:
Thanks Ryan!

I like that the new doc addition is concise ... but as a newcomer, I might need something more detailed to understand the reasoning for this and the special action used to trigger the re-render.

If you agree, I'd add the following section after the existing `Deferring a Re-Render Until Later` section:

```rst
Forcing a Re-Render Explictly
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

In some cases, you might want to force a component re-render explicitly. For
example, consider a checkout component that provides a coupon input that must
only be used when clicking on the associated "Apply coupon" button:

.. code-block:: html+twig

    <input data-model="norender|coupon">
    <button data-action="live#$render">Apply coupon</button>

The ``norender`` option on the input ensures that the component won't re-render
when this input changes. The ``live#$render`` action is a special built-in action
that triggers a re-render.
```

weaverryan:
I like it - thanks Javier!

PR #1575, opened 2024-03-01
CVA apply function accept null arguments

### LINKED ISSUE #1570: [TwigComponent] Examples in the CVA documentation do not work
Hey guys, curious as I am, I just wanted to try out CVA. However, I'm running into problems:

```
An exception has been thrown during the rendering of a template ("Symfony\UX\TwigComponent\CVA::apply(): Argument #2 must be of type string, null given, called in /var/www/html/vendor/twig/twig/src/Extension/CoreExtension.php on line 1635").
```

That error happens when trying out CVA as documented here: 
[https://symfony.com/bundles/ux-twig-component/current/index.html#component-with-complex-variants-cva](https://symfony.com/bundles/ux-twig-component/current/index.html#component-with-complex-variants-cva)

The error happens when I use <twig:Alert /> and do not define any CSS classes. So, to me it looks like `attributes.render('class')` does not return an empty string but null. Am I doing something wrong or is the documentation incorrect? 

`{{ alert.apply({color, size}, attributes.render('class')|format) }}` fixes the issue temporarily by casting the null to an empty string.

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1570 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/CVA.php
+++ b/src/TwigComponent/src/CVA.php
@@ -39,9 +39,9 @@ final class CVA
     ) {
     }
 
-    public function apply(array $recipes, string ...$classes): string
+    public function apply(array $recipes, ?string ...$classes): string
     {
-        return trim($this->resolve($recipes).' '.implode(' ', $classes));
+        return trim($this->resolve($recipes).' '.implode(' ', array_filter($classes)));
     }
 
     public function resolve(array $recipes): string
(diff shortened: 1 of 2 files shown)


kbond (approved):

weaverryan:
Thanks Matheo!

PR #1578, opened 2024-03-03
[LiveMemory] Regroup files & fix performances issues

smnandre:
Two main things:

* Started to regroup & isolate files as promised
* Fixed the performances on webkit / chrome


weaverryan:
Thanks Simon!

PR #1583, opened 2024-03-05
[LiveComponent] Fix collections hydration with serializer

squrious:
| Q            | A   |
|--------------|-----|
| Bug fix?     | yes |
| New feature? | no  |
| Issues       | N/A |
| License      | MIT |

Hello!

I encountered issues when trying to hydrate a `LiveProp` as a collection of objects, using the Symfony Serializer.

Given `phpdocumentor/reflection-dockblock` is installed in the project, and I have the following component:

```php

#[AsLiveComponent]
class MyComponent
{
    /**
     * @var MyDto[] 
     */
    #[LiveProp(useSerializerForHydration: true)]
    public array $prop = [];
}
```

Then I should be able to hydrate `$prop` as a collection of `MyDto`. 

However, the `LiveComponentHydrator` doesn't consider the collection type when `useSerializerForHydration` is enabled, and tries to deserialize to `array` type, throwing the following exception:
```
Symfony\Component\Serializer\Exception\NotNormalizableValueException : Could not denormalize object of type "array", no supporting normalizer found.
```

So here is a proposal to fix the issue.

Cheers!

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -256,11 +256,22 @@ final class LiveComponentHydrator
                 throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the given serializer does not implement DenormalizerInterface.', $propMetadata->getName(), $parentObject::class));
             }
 
-            if (null === $propMetadata->getType()) {
+            if ($propMetadata->collectionValueType()) {
+                $builtInType = $propMetadata->collectionValueType()->getBuiltinType();
+                if (Type::BUILTIN_TYPE_OBJECT === $builtInType) {
+                    $type = $propMetadata->collectionValueType()->getClassName().'[]';
+                } else {
+                    $type = $builtInType.'[]';
+                }
+            } else {
+                $type = $propMetadata->getType();
+            }
+
+            if (null === $type) {
                 throw new \LogicException(sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
             }
 
-            return $this->serializer->denormalize($value, $propMetadata->getType(), 'json', $propMetadata->serializationContext());
+            return $this->serializer->denormalize($value, $type, 'json', $propMetadata->serializationContext());
         }
 
         if ($propMetadata->collectionValueType() && Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()) {
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.17.0
 
 -   Add `modifier` option in `LiveProp` so options can be modified at runtime.
+-   Fix collections hydration with serializer in LiveComponents
 
 ## 2.16.0
 
--- a/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
+++ b/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
@@ -1031,6 +1031,42 @@ final class LiveComponentHydratorTest extends KernelTestCase
             ;
         }];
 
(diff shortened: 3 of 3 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -259,7 +259,9 @@ public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, objec
                 throw new \LogicException(sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
             }
 
-            return $this->serializer->denormalize($value, $propMetadata->getType(), 'json', $propMetadata->serializationContext());
+            $type = $propMetadata->collectionValueType() ? $propMetadata->collectionValueType()->getClassName().'[]' : $propMetadata->getType();
```
weaverryan: Do we need to also check for `Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()` like we do on line 267? Basically, what if this is typed as `string[]`?

weaverryan (changes requested):

src/LiveComponent/src/LiveComponentHydrator.php
squrious: Yes absolutely, I missed those cases. Next commit fixes that, and also ensures the built-in collection type is scalar (if used). Otherwise the "not guessable type" exception is thrown

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -255,11 +255,27 @@ public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, objec
                 throw new \LogicException(sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the given serializer does not implement DenormalizerInterface.', $propMetadata->getName(), $parentObject::class));
             }
 
-            if (null === $propMetadata->getType()) {
+            $type = null;
+            if ($propMetadata->collectionValueType()) {
+                $builtInType = $propMetadata->collectionValueType()->getBuiltinType();
+                if (Type::BUILTIN_TYPE_OBJECT === $builtInType) {
+                    $type = $propMetadata->collectionValueType()->getClassName();
+                } elseif (\in_array($builtInType, ['string', 'int', 'float', 'bool'])) {
```
weaverryan: Why this line? Are there other possible values for `$builtInType` that we don't want to send to the serializer?

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -255,11 +255,27 @@ public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, objec
...
+            }
+
+            if (null === $type) {
                 throw new \LogicException(sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
             }
 
-            return $this->serializer->denormalize($value, $propMetadata->getType(), 'json', $propMetadata->serializationContext());
+            if ($propMetadata->collectionValueType()) {
+                $type .= '[]';
+            }
```
weaverryan: Minor - but I think it's more clear if we move this up to the if statement above. Sure, we'll need it on both sides of the if statement, but I was confused at first when I didn't see this.

weaverryan (changes requested):

src/LiveComponent/src/LiveComponentHydrator.php
squrious: Values like `array`, `iterable`, `object`, `mixed` will cause the serializer to throw an exception, as no normalizer exists for these types. 
I thought we could filter them here, so these types are not "recognized" and we throw the exception for not guessable types.
Would you prefer letting the serializer handle this? Or just add a comment to explain?

weaverryan: I see! Let's let the serializer handle them.

weaverryan:
Thanks Nicolas!

smnandre:
Oh i missed this one ! So cool 👏

PR #1593, opened 2024-03-08
[LiveComponent] Throw exception for typed LiveProps as interfaces

### LINKED ISSUE #1590: [LiveComponent] `LiveComponentHydrator` doesn't handle interfaces
My model is a `Product` which has a typed `discount` property of type `DiscountInterface`:

```php
namespace App\Model;

class Product
{
    public ?DiscountInterface $discount = null;
}
```

I'musing `Product` as a `LiveProp` (it doesn't matter how I use it or where, keep reading). 

During the de/hydration process, a warning is raised:

> Warning: foreach() argument must be of type array|object, null given.
> in [vendor/symfony/ux-live-component/src/LiveComponentHydrator.php ](file:///var/www/html/vendor/symfony/ux-live-component/src/LiveComponentHydrator.php#L490)(line 490)

Investigating more, around the [`dehydrateObjectValue`](https://github.com/symfony/ux-live-component/blob/7bbc05afc0c8ed8795389327e848ff2ddcd9d7a5/src/LiveComponentHydrator.php#L500) of `LiveComponentHydrator`: 

- The passed `$value` is the instance of `App\Model\StackableDiscount`
- The `$classType` is `DiscountInterface`
- The `getProperties` on `PropertyInfoExtractor` returns `null` (hence the `foreach` warning)

image

I think this is a bug but I don't know how to fix it. Logically, we can't get properties from an interface. `$classType` is cleary wrong here.

Maybe:

```
if (is_subclass_of($value, $classType, true)) {
    $classType = get_class($value);
}
```

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1590
| License       | MIT

Check is done just after the hydration extensions are called, to allow userland implementations.

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -496,6 +496,10 @@ final class LiveComponentHydrator
             }
         }
 
+        if (interface_exists($classType)) {
+            throw new \LogicException(sprintf('Cannot dehydrate value typed as interface "%s" on component "%s". Change this to a concrete type that can be dehydrated. Or set the hydrateWith/dehydrateWith options in LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer.', get_debug_type($value), $parentObject::class));
+        }
+
         $dehydratedObjectValues = [];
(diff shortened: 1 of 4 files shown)


WebMamba (approved):

kbond:
Thanks Simon.

PR #1594, opened 2024-03-09
[Icons] Improve cache warmup by looking at every string

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

Improve working with dynamic icon names (inspired by https://v2.tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html)

```twig
{# This will be cached #}
{{ ux_icon('flag-fr') }}

{# This will NOT be cached #}
{{ ux_icon('flag-' ~ locale) }}

{% set flags = {fr: 'flag-fr', de: 'flag-de'} %} {# both "flag-fr" and "flag-de" will be cached #}
{{ ux_icon(flags[locale]) }}
```

Closes #1601

### DIFF
--- a/src/Icons/src/Iconify.php
+++ b/src/Icons/src/Iconify.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\Icons;
 use Symfony\Component\HttpClient\Exception\JsonException;
 use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Component\HttpClient\ScopingHttpClient;
+use Symfony\Contracts\Cache\CacheInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\UX\Icons\Exception\IconNotFoundException;
 use Symfony\UX\Icons\Svg\Icon;
@@ -26,8 +27,10 @@ use Symfony\UX\Icons\Svg\Icon;
 final class Iconify
 {
     private HttpClientInterface $http;
+    private \ArrayObject $sets;
 
     public function __construct(
+        private CacheInterface $cache,
         string $endpoint = 'https://api.iconify.design',
         ?HttpClientInterface $http = null,
     ) {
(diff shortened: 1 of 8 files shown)


kbond:
@smnandre made some fair points about performance/security.

Regarding security: I want to clarify that only strings that match `'foo:bar'` will be used to make http requests to iconify. `'foo:bar'` will create an http request with this value in the request path but `'something-sensitive'` will not. However, `'something:sensitive'` would create an http request with this value in the request path.

I'd like to find a compromise that we're all ok with, some thoughts:
- opt-in to this much looser string check
- configure specific paths/globs (like tailwind does)

smnandre:
We can add a failsafe: store locally the list of available icon prefixes. 

Then we reverse the logic: 
* collect all the icons as you do now
* list the prefixes
* download the list of icons for each prefix

--> finally download the icons.

It can be even faster in the end, because we can group parallelize + group icon downloads (up to 500 per request)

smnandre:
I'm reading the doc 

> During development, if you modify an icon, you will need to clear the cache (``bin/console cache:clear``) to see the changes.
    
Could we add a isFresh / mtime check about that ? What do you think ?

kbond:
I believe the security concerns have been alleviated. I now only make the http request if the prefix is valid. `'something:sensitive'` will no longer create a request.

> Could we add a isFresh / mtime check about that ? What do you think ?

I'd prefer to avoid this complexity until this is determined to be a problem. I feel this situation is be rare.

src/Icons/src/Iconify.php
```diff
@@ -40,13 +43,15 @@ public function __construct(
...
+        return $this->sets()[$prefix] ?? throw new \RuntimeException(sprintf('The icon prefix "%s" does not exist on iconify.design.', $prefix));
     }
 
     public function fetchIcon(string $prefix, string $name): Icon
     {
+        if (!isset($this->sets()[$prefix])) {
+            throw new IconNotFoundException(sprintf('The icon "%s:%s" does not exist on iconify.design.', $prefix, $name));
+        }
+
         $response = $this->http->request('GET', sprintf('/%s.json?icons=%s', $prefix, $name));
```
smnandre, line 55: If (i insist on _if_) this become the bottleneck, we have a card in hand: grouping icons per set   https://iconify.design/docs/api/icon-data.html#query

src/Icons/src/Twig/IconFinder.php
```diff
@@ -38,11 +38,7 @@ public function icons(): array
         foreach ($this->files($this->twig->getLoader()) as $file) {
```
smnandre, line 38: Should we exclude the Bundle ?

smnandre:
I love this!

src/Icons/src/Twig/IconFinder.php
kbond: What bundle?

smnandre: At this point, i think Twig loader has been registered with all the bundles of the app. So it contains all the files in "vendor/acme/*-bundle/Resources/templates" and we should (imho) not go there to extract things. 

Imagine a twig file there with a list of icons (think any template using JS icons)... the probability to have 'bi:check' string is far from null. 

We even could see this as a vector "attack", a single file in twig from a 3rd party bundle would make you downlaod icons locally / generate a lot of download. 

Oh and i'll check something about that we may have a problem there.


So here i'd like us to only look in userland files :)

kbond: What if you had a bundle of that uses icons that you wanted to use? I prefer to keep it simple and adjust when/if we come across these type of problems.

PR #1595, opened 2024-03-09
[Icons] Allow disabling the "on demand" system

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        | n/a
| License       | MIT

This config gives the option to import icons iconify.design but not use them "on demand".

/cc @smnandre


smnandre:
🙇

PR #1596, opened 2024-03-09
[Icons] allow configuring multiple local directories and optionally prefixing them

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        | n/a
| License       | MIT

Configure multiple paths:

```yaml
ux_icons:
    icon_dir:
        - %kernel.project_dir%/assets/icons
        - %kernel.project_dir%/vendor/your-vendor/your-bundle/assets/icons
```

Additionally, you can suffix the path with `@<alias>` to add a prefix for icons in this directory:

```yaml
ux_icons:
    icon_dir:
        # ...
        - %kernel.project_dir%/vendor/fortawesome/font-awesome/svgs@fa
```

Now you can render icons in this directory by prefixing the name with `fa:`:

```twig
{{ ux_icon('fa:solid:adjust', {class: 'w-4 h-4'}) }} <!-- renders "vendor/fortawesome/font-awesome/svgs/solid/adjust.svg" -->
```

### DIFF
--- a/src/Icons/src/DependencyInjection/UXIconsExtension.php
+++ b/src/Icons/src/DependencyInjection/UXIconsExtension.php
@@ -37,12 +37,30 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
                 ->arrayNode('icon_dir')
                     ->beforeNormalization()
                         ->ifString()
-                        ->then(static function ($v) { return [$v]; })
+                        ->then(static fn ($v) => [$v])
+                    ->end()
+                    ->beforeNormalization()
+                        ->always(function (array $values) {
+                            $ret = [];
+
+                            foreach ($values as $value) {
+                                if (2 === \count($parts = explode('@', $value, 2))) {
+                                    $ret[$parts[1]] = $parts[0];
+
+                                    continue;
+                                }
+
+                                $ret[] = $value;
+                            }
+
+                            return $ret;
+                        })
                     ->end()
                     ->info(<<<EOF
                         The local directory('s) where icons are stored.
                         Order matters as the first directory to contain the icon will be used.
                         The first directory will be used to store imported icons.
+                        Suffix with "@<alias>" to use a prefix.
                         EOF)
                     ->scalarPrototype()->end()
                     ->defaultValue(['%kernel.project_dir%/assets/icons'])
--- a/src/Icons/src/DependencyInjection/UXIconsExtension.php
+++ b/src/Icons/src/DependencyInjection/UXIconsExtension.php
@@ -34,9 +34,18 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
 
         $rootNode
             ->children()
-                ->scalarNode('icon_dir')
-                    ->info('The local directory where icons are stored.')
-                    ->defaultValue('%kernel.project_dir%/assets/icons')
+                ->arrayNode('icon_dir')
+                    ->beforeNormalization()
+                        ->ifString()
+                        ->then(static function ($v) { return [$v]; })
+                    ->end()
+                    ->info(<<<EOF
+                        The local directory('s) where icons are stored.
+                        Order matters as the first directory to contain the icon will be used.
+                        The first directory will be used to store imported icons.
+                        EOF)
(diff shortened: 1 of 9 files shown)


smnandre:
I need something similar for the rendering options... in the end we maybe will need to merge the syntax into something a bit more complex, lire the transports, or the doctrine connections.

I guess thoses are "read-only" ? (as we won't be able to write in "vendors" with the import command)

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -34,9 +34,36 @@ public function getConfigTreeBuilder(): TreeBuilder
 
         $rootNode
             ->children()
-                ->scalarNode('icon_dir')
-                    ->info('The local directory where icons are stored.')
-                    ->defaultValue('%kernel.project_dir%/assets/icons')
+                ->arrayNode('icon_dir')
```
kbond, line 37: Should I rename this to `icons_dirs`?

kbond:
> I need something similar for the rendering options... in the end we maybe will need to merge the syntax into something a bit more complex, lire the transports, or the doctrine connections.

Something like?

```yaml
ux_icons:
  icon_dir:
    - path: some/path
      prefix: fa
      readonly: true
      more: ...
```

> I guess thoses are "read-only" ?

All but the first one, yeah. We could make configurable as shown above.

src/Icons/src/DependencyInjection/UXIconsExtension.php
yceruto: maybe `paths` if they can be relative?

yceruto:
```yaml
ux_icons:
    paths:
        vendor/fortawesome/font-awesome/svgs: fa
```
? ... then normalize to `path`, `prefix`?

yceruto:
Also, it seems the concept is pretty close to Twig namespaces. It could be good if we reuse same concept that we already know, so "prefix" could be named "namespace" and the usage could `ux_icon('@fa/solid/adjust.svg')`, it's a path to a resource in the end, right? Less new magic/convention, more easy to understand.

smnandre:
> ```yaml
> ux_icons:
>     paths:
>         vendor/fortawesome/font-awesome/svgs: fa
> ```

I'd use the "prefix" as key, like a namespace. As it's the only thing we need to garantee the unicity.

So maybe

```yaml

    dfgdfsdfd:    # word i don't have; set ? packs ? groups ? ns ? 
        bi:
            path: assets/bootstrap
       ld: 
           alias: lucide
      bi-reg:
            dir: vendor/fortawesome/font-awesome/svgs:
```

Do you agree to make it  the "primary key" ?

kbond:
Yeah, prefixing with a key seems best - this will allow choosing the path to import into when using the import command. And @yceruto, namespacing is an interesting idea. Maybe something like:

```yaml
ux_icons:
  paths:
    default: some/path

    fa:
        path: vendor/some/path
        readonly: true
        namespace: fa
```

About the `:` separator. Twig switched from `:` to `/` eventually, I wonder if we should do the same here initially (have no concept of `:`). Like @yceruto says, keep to twig's conventions.

smnandre:
I'd say...

* prefix/key:  no "directory" allowed (no ":", no "/")
* path: free as a french mobile phone carrier


Regarding "name/identifier":
* HTML syntax uses the ":" notation, i'd feel maybe more logic to keep it...  but if you want to use "/" i'm not gonna fight for this :)) 
* Twig component (in component function) names: i'll check, but i think we have 0 constraints there, so no hot take there either.

smnandre:
How can we move on this ? What do you need @kbond ? Or can maybe start in parralel the attributes configuration per directory and we see later how to merge ?

kbond:
> prefix/key: no "directory" allowed (no ":", no "/")

I think we need to allow a directory even when a prefix is used. For instance, `assets/icons/heroicons: hero` could have `hero:24:outline:user`/`hero:24:solid:user`

smnandre:
As discussed with @kbond (some _weeks_ ago 😊 ) , i'm closing in favor of #2156

PR #1597, opened 2024-03-09
[Icons] Add InMemory registry to allow easier tests

smnandre:
Will come handy to test the IconRenderer

### DIFF
--- a/src/Icons/tests/Unit/Registry/InMemoryIconRegistryTest.php
+++ b/src/Icons/tests/Unit/Registry/InMemoryIconRegistryTest.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Icons\Tests\Unit\Registry;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\UX\Icons\Exception\IconNotFoundException;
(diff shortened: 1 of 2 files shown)


kbond:
As discussed in Slack, do we want to make this a configuration option?

Allowing this to be used as the only registry in your test environment could save a ton of http requests for apps that extensively use the _on demand_ system.

smnandre:
Yep that was on my mind..  But i rather go step by step there, as i'm sure i need the InMemoryRegistry to add feature + test the Renderer :)

You ok if we add the config/test thing in another PR ?

src/Icons/tests/Unit/Registry/InMemoryIconRegistryTest.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Symfony\UX\Icons\Tests\Unit\Registry;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\UX\Icons\Registry\InMemoryIconRegistry;
+use Symfony\UX\Icons\Svg\Icon;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ */
+final class InMemoryIconRegistryTest extends TestCase
```
norkunas, line 22: `throw new IconNotFoundException(sprintf('Icon "%s" not found.', $name));`
is not covered :)

kbond:
Could you show an example of how you intend this to be used by end users?

kbond:
Thanks Simon.

smnandre:
> Could you show an example of how you intend this to be used by end users?

First example there (internal so that's not the answer you expected ... but it's a start)

https://github.com/symfony/ux/pull/1606

PR #1598, opened 2024-03-09
[Icons] Add Iconify tests

smnandre:
+ remove Icon implementation


src/Icons/src/Iconify.php
```diff
@@ -25,10 +25,12 @@
  */
 final class Iconify
 {
+    private const API_BASE_URI = 'https://api.iconify.design';
```
kbond, line 28: Let's use this constant in the bundle config.

kbond (changes requested):
Looks good, just want to note that the iconify API doesn't return 404 in some cases. Just want to ensure this is handled.

src/Icons/src/Iconify.php
```diff
@@ -66,15 +74,15 @@ public function fetchIcon(string $prefix, string $name): Icon
 
     public function fetchSvg(string $prefix, string $name): string
     {
-        $content = $this->http
-            ->request('GET', sprintf('/%s/%s.svg', $prefix, $name))
-            ->getContent()
-        ;
+        $response = $this->http->request('GET', sprintf('/%s/%s.svg', $prefix, $name));
```
kbond, line 77: I'm not sure a non 200 is ever returned. Is this covered by a test? (Same with the other spots we check for a 200)

smnandre: Yes if the set does not exists / is not available. That's why i added the tests :) 

* https://iconify.design/docs/api/collections.html#error-response
* https://iconify.design/docs/api/collection.html#error-response
* https://iconify.design/docs/api/icon-data.html#error-response

kbond: Okay, I couldn't remember the exact scenarios were a 404 wasn't returned when I thought it should be.

smnandre: When the prefix exists but not the icon i think (within the collection endpoint)

smnandre:
Closed and partially merged into #1617

PR #1601, opened 2024-03-09
[Icons] Cache icon set list

smnandre:
We should keep the list of iconset locally .. it will help a lot to limit the number of http request / checks / etc

(pr used as example, this would need a bit more test ^^)

### DIFF
--- a/src/Icons/src/Iconify.php
+++ b/src/Icons/src/Iconify.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\Icons;
 use Symfony\Component\HttpClient\Exception\JsonException;
 use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Component\HttpClient\ScopingHttpClient;
+use Symfony\Contracts\Cache\CacheInterface;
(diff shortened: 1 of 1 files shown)


src/Icons/src/Iconify.php
```diff
@@ -68,13 +81,25 @@ public function fetchSvg(string $prefix, string $name): string
...
         return $content;
     }
+
+    private function fetchCollections(): array
+    {
+        $response = $this->http->request('GET', '/collections');
+        if (200 !== $response->getStatusCode()) {
+            throw new \RuntimeException('Could not fetch icon collections from iconify.design.');
+        }
+        if (null === $collections = $response->toArray()) {
```
norkunas, line 99: `toArray` never returns null

smnandre: Omg what did wrote ?

(the idea behind this PR was to prevent some unhandled 404 (when the icon set does not exists).. but with your refacto this is probably no more usefull)

smnandre: We can close it now i think ?

kbond:
Closing in favor of #1594

PR #1606, opened 2024-03-10
[Icons] Test IconRenderer 

smnandre:
Create tests for the IconRenderer

Add internal docblock to clarify the attributes precedence

(+ one CS fix passing by)

-- 

+ moved the InMemoryRegistrey in the `tests` directory

As @kbond  says, it won't be used by final users so it's better there

### DIFF
--- a/src/Icons/tests/Unit/IconRendererTest.php
+++ b/src/Icons/tests/Unit/IconRendererTest.php
@@ -1,0 +1,175 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Icons\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\UX\Icons\Exception\IconNotFoundException;
+use Symfony\UX\Icons\IconRegistryInterface;
+use Symfony\UX\Icons\IconRenderer;
+use Symfony\UX\Icons\Svg\Icon;
+use Symfony\UX\Icons\Tests\Util\InMemoryIconRegistry;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ */
+class IconRendererTest extends TestCase
+{
+    public function testRenderIcon(): void
+    {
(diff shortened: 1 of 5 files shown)


src/Icons/tests/Unit/IconRendererTest.php
```diff
@@ -0,0 +1,170 @@
...
+            'foo' => ['', $iconAttrs],
+        ]);
+        $iconRenderer = new IconRenderer($registry, $defaultAttrs);
+
+        $svg = $iconRenderer->renderIcon('foo', $renderAttr);
+        $this->assertStringStartsWith($expectedTag, $svg);
+    }
+
+    /**
+     * @return array [iconAttributes, defaultAttributes, renderAttributes, expectedTag]
```
norkunas: Is this valid phpdoc syntax?
Also you return a iterable/generator, not an array 🙂

src/Icons/tests/Unit/IconRendererTest.php
```diff
@@ -0,0 +1,170 @@
...
+
+        $iconRenderer->renderIcon('foo');
+    }
+
+    public function testRenderIconThrowsExceptionWhenAttributesAreInvalid(): void
+    {
+        $registry = $this->createRegistry(['foo' => '<path d="M0 0L12 12"/>']);
+        $iconRenderer = new IconRenderer($registry);
+
+        $this->expectException(InvalidArgumentException::class);
```
norkunas: Exception message should also be tested, no?

src/Icons/tests/Unit/IconRendererTest.php
smnandre: A iterable/generator of array but yes. I'll delete all as no one will add tests

src/Icons/tests/Unit/IconRendererTest.php
smnandre: No i just want to test exceptions are thrown from Icon. 

The message is not defined there so won't test it.

kbond (approved):

kbond:
can you fix the cs

smnandre:
Good now

kbond:
needs the license on `IconRendererTest`

smnandre:
Arf sorry

kbond:
Thanks Simon.

PR #1607, opened 2024-03-10
[Icons] Compute viewBox with iconset data (+tests)

smnandre:
There were a bug for icons with no individual "height" attribute.

Now it fallbacks on IconSet metada.

And i started a test class for Iconify


kbond (approved):

kbond:
Thanks Simon.

PR #1608, opened 2024-03-11
[Icons] Rework cache warm

smnandre:
The warm cache took a lot of time because

- the regexp was very open (due to the non-prefix icon names)
- twig templates were listed multiple times (due to the namespace system)

And a looot of false positives caused lot of cache locks for nothing.

On ux.symfony.com project with this component, the PR make the icons detected from 1000+ to 1 :)
And the command is now way quicker to run.

The prefix-less icons are warmed by simply listing the SVG at the root of the asset/icons directory

### DIFF
--- a/src/Icons/config/services.php
+++ b/src/Icons/config/services.php
@@ -50,15 +50,16 @@ return static function (ContainerConfigurator $container): void {
             ])
             ->tag('twig.runtime')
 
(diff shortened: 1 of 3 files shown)


kbond (approved):

kbond:
Thanks Simon.

PR #1609, opened 2024-03-11
[Twig] allow stringable objects as attribute values

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | n/a
| License       | MIT

Allows passing a stringable objects as attribute values.

```twig
{# currently #}
<twig:Form :action="stringableUriObject"> {# error because stringableUriObject isn't scalar #}

{# with this PR #}
<twig:Form :action="stringableUriObject"> {# auto converts stringableUriObject to string #}
```

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -41,6 +41,10 @@ final class ComponentAttributes implements \IteratorAggregate, \Countable
             function (string $carry, string $key) {
                 $value = $this->attributes[$key];
 
+                if ($value instanceof \Stringable) {
+                    $value = (string) $value;
+                }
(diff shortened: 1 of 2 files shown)


seb-jean:
What will this change? :)

kbond:
> What will this change? :)

I've updated the PR description to show what will change.

smnandre:
Great ! Thanks you :) 

Minor **if you have time**:
The description _may_ be _a bit_ missleading, as your PR has no impact on how you pass arguments to components, but how attributes are rendered.
Maybe illustrate with something in the "component view" instead of the "calling template" ?

smnandre (approved):

PR #1617, opened 2024-03-12
[Icons] Add Icon + Iconify tests

smnandre:
Add missing tests on Icon attributes (html + with/out)

(before making changes to the Renderer)


norkunas:
How it's possible to use `attributes.without` when using `<twig:UX:Icon />` ?
I mean - I have css class that sets for all icons width/height, so there is no point to render icons with these attributes, how I could remove them from the output?

smnandre:
This is planned, and i understand very well your need, as i have the same :)

smnandre:
FYI this PR is one of the steps required ;)

src/Icons/src/Iconify.php
```diff
@@ -26,12 +26,14 @@
  */
 final class Iconify
 {
+    private const API_BASE_URI = 'https://api.iconify.design';
```
kbond: Can you use this in the configuration?

kbond (changes requested):

src/Icons/src/Iconify.php
smnandre: Do we really need to make this configurable ? :| 

If "yes", i'd then let a null default value.

But using a const value in configuration ? That seems a bit weird. Not that i don't want, this just surprises me :)

smnandre: It's done

kbond:
Thank you Simon.

PR #1631, opened 2024-03-19
[LiveComponent] Fix idiomorph dependency (fix #1567)

### LINKED ISSUE #1567: [LiveComponent] Running `npm install --force` fails with "This is related to npm not being able to find a file."
When trying to install the latest version of UX Live Component I can successfully install the package with composer.
When I run `npm install --force` however the command fails with an error message that suggests it can't find a file (see image).

npm_output

My `package.json` points at a file in my vendor folder:
```json
"@symfony/ux-live-component": "file:vendor/symfony/ux-live-component/assets",
```
...and upon inspection the directory in question it appears to look fine, containing it's own `package.json` alongside a `dist` and `styles` directory.

I have attatched the npm log file in case it's useful at all.

[2024-03-01T13_00_54_787Z-debug-0.log](https://github.com/symfony/ux/files/14461498/2024-03-01T13_00_54_787Z-debug-0.log)

smnandre:
Current idiomorph dependency does not work for some users with npm

Now it's available as a ES module it should fix the problem

Thanks @pcmmf for the hand

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -588,7 +588,7 @@ var Idiomorph = (function () {
          * @returns {boolean}
          */
         function ignoreValueOfActiveElement(possibleActiveElement, ctx) {
-            return ctx.ignoreActiveValue && possibleActiveElement === document.activeElement && possibleActiveElement !== document.body;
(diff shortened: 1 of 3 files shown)


WebMamba (approved):
Thanks @smnandre! Looks good to me!

kbond:
Thanks Simon.

PR #1633, opened 2024-03-20
Fix typo on only and without methods documentation

JeanDavidDaviet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | 
| License       | MIT

There are two typos in the `only` and `without` examples.
The intermediate representation still have the style and the classes attributes respectively.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -1025,13 +1025,13 @@ Extract specific attributes and discard the rest:
 
 .. code-block:: html+twig
 
+    {# render component #}
+    {{ component('MyComponent', { class: 'foo', style: 'color:red' }) }}
+
     {# templates/components/MyComponent.html.twig #}
     <div{{ attributes.only('class') }}>
       My Component!
     </div>
-
-    {# render component #}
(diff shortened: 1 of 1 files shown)


kbond:
The reason these extra attributes were included was to explicitly show that they are excluded when rendering the component. Am I missing something - does it not work as expected?

WebMamba:
Agree with @kbond here, I think the documentation show the right things here

smnandre:
We may want to reverse the order to make it easier

1/ Calling template

2/ Component template

3/ HTML result

kbond:
TBH, I don't have a good use-case for the only/without methods. If someone has a good use-case, let's update the docs to show.

JeanDavidDaviet:
thx @kbond for your input I understand more the reason.
I think @smnandre's idea is better as you can better see the initial values and compare it with the final result.
I've updated the PR. 
I personnaly think it makes more sense this way, what do you think ?

smnandre (approved):

kbond:
Thank you Jean-David.

PR #1602, opened 2024-03-09
[Icons] Some doc updates

smnandre:
Step by step...

PR #1641, opened 2024-03-21
[Doc] Tweaks and fixes for UX Icons docs

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

I think we should mention some of the popular icon sets so people googling for them and Symfony can find this page. Also, let's show the config file path in the config example. I'm not sure if we want that file path to be `config/packages/ux_icons.yaml`

In #1602 we're making more changes to docs, so, if this PR is merged, I think it should be merged before #1602. Thanks!

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -1,8 +1,10 @@
 Symfony UX Icons
 ================
 
-Renders local and remote SVG icons in your Twig templates. It also gives direct
-access to over 200,000 open source vector icons from tens of icon sets.
+This package provides utilities to work with SVG icons. It renders both local
+and remote icons in your Twig templates. It provides direct access to over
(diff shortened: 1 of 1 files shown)


smnandre:
I had the intention to open another doc PR next week, with a list of 10/12 big icon sets with their prefix and an example

(something i started to work on as command) 

Capture d’écran 2024-03-10 à 23 59 50

Capture d’écran 2024-03-11 à 00 27 58

Do you want to make a first pass ? Wait for me ?

javiereguiluz:
There's no hurry ... so we could consider this PR "as is" ... and when your command is ready, etc. we can update the docs to add that table with the Top 10 icon sets and their prefixes. Thanks

kbond (approved):

kbond:
Thank you Javier.

PR #1647, opened 2024-03-22
[Twig] boolean (true) CVA variants

### LINKED ISSUE #1614: [TwigComponent] CVA - boolean variants not working
Sadly, the boolean support for variants doesn't work. It's [supposed two work](https://www.youtube.com/watch?v=T-Zv73yZ_QI&t=475s). 

The problem is `true` passed as string to the Twig. Is there a way to solve this issue

```twig
{% props theme = "primary", size = null, text = null, disabled = false %}

<button
    class="{{ cva({
        base: "btn",
        variants: {
            theme: {
                primary: "btn-primary",
                secondary: "btn-secondary",
                ghost: "btn-ghost",
                link: "btn-link",
            },
            size: {
                xs: "btn-xs",
                sm: "btn-sm",
                md: "btn-md",
                lg: "btn-lg",
            },
            disabled: {
                true: "btn-disabled",
            },
        },
        compoundVariants: [],
    }).apply({ theme, size }, attributes.render("class")|default("")) }}"
    {{ attributes.without("class") }}
>
    {% if block("content") is defined %}
        {{ block("content") }}
    {% elseif text|length %}
        {{ text }}
    {% endif %}
</button>
```

image

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1614
| License       | MIT

This is an attempt to solve #1614. As discussed in that issue, it seems like only a `true` should be valid. Because of this, I opted for the variant value(s) to be a `string` or `list<string>` - this is used when the recipe value is `true`. Examples:

```twig
{% props size = 'md', disabled = false %}

{% set button = cva({
    variants: {
        size: {
            sm: 'alert-sm',
            md: 'alert-md',
            lg: 'alert-lg',
        },
        disabled: 'opacity-50 cursor-not-allowed',
        disabled: ['opacity-50', 'cursor-not-allowed'], {# also valid #}
    },
}) %}

<button class="{{ button.apply({size, disabled}, attributes.render('class'), 'flex p-4') }}">
    ...
</button>
```

**TODO**
- [ ] docs
- [ ] changelog

### DIFF
--- a/src/TwigComponent/src/CVA.php
+++ b/src/TwigComponent/src/CVA.php
@@ -26,8 +26,8 @@ final class CVA
 {
     /**
      * @var string|list<string|null>|null
-     * @var array<string, array<string, string|list<string|null>>|null the array should have the following format [variantCategory => [variantName => classes]]
-     *                                                ex: ['colors' => ['primary' => 'bleu-8000', 'danger' => 'red-800 text-bold'], 'size' => [...]]
+     * @var array<string, string|list<string>|array<string, string|list<string|null>>|null the array should have the following format [variantCategory => [variantName => classes]]
+     *                                                                                     ex: ['colors' => ['primary' => 'bleu-8000', 'danger' => 'red-800 text-bold'], 'size' => [...]]
      * @var array<array<string, string|array<string>>> the array should have the following format ['variantsCategory' => ['variantName', 'variantName'], 'class' => 'text-red-500']
      * @var array<string, string>|null
      */
@@ -53,6 +53,14 @@ final class CVA
         }
 
         foreach ($recipes as $recipeName => $recipeValue) {
+            if (isset($this->variants[$recipeName]) && (\is_string($this->variants[$recipeName]) || (\is_array($this->variants[$recipeName]) && array_is_list($this->variants[$recipeName])))) {
+                if (true === $recipeValue) {
+                    $classes .= ' '.implode(' ', (array) $this->variants[$recipeName]);
+                }
+
+                continue;
+            }
+
             if (!isset($this->variants[$recipeName][$recipeValue])) {
                 continue;
             }
--- a/src/TwigComponent/tests/Fixtures/templates/class_merge.html.twig
+++ b/src/TwigComponent/tests/Fixtures/templates/class_merge.html.twig
@@ -1,1 +1,2 @@
 <twig:Alert color='red' size='lg' class='dark:bg-gray-600'/>
+<twig:Alert color='red' size='lg' class='dark:bg-gray-600' disabled />
--- a/src/TwigComponent/tests/Fixtures/templates/components/Alert.html.twig
+++ b/src/TwigComponent/tests/Fixtures/templates/components/Alert.html.twig
@@ -1,4 +1,4 @@
-{% props color = 'blue', size = 'md' %}
+{% props color = 'blue', size = 'md', disabled = false %}
 
 {% set alert = cva({
         base: ['alert'],
@@ -18,7 +18,8 @@
                 sm: 'rounded-sm',
                 md: 'rounded-md',
                 lg: 'rounded-lg',
-            }
+            },
+            disabled: 'disable',
         },
         compoundVariants: [{
             color: ['red'],
@@ -30,6 +31,6 @@
         }
 }) %}
 
-<div class="{{ alert.apply({color, size}, attributes.render('class'), 'flex p-4') }}">
+<div class="{{ alert.apply({color, size, disabled}, attributes.render('class'), 'flex p-4') }}">
     ...
 </div>
--- a/src/TwigComponent/tests/Integration/ComponentExtensionTest.php
+++ b/src/TwigComponent/tests/Integration/ComponentExtensionTest.php
@@ -261,6 +261,7 @@ final class ComponentExtensionTest extends KernelTestCase
         $output = self::getContainer()->get(Environment::class)->render('class_merge.html.twig');
 
         $this->assertStringContainsString('class="alert alert-red alert-lg font-semibold rounded-md dark:bg-gray-600 flex p-4"', $output);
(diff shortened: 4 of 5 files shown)


seb-jean:
When `disabled` is false, how could we associate CSS classes with it ?

Kocal:
To me it should only accept an array with `true` and `false` as keys

kbond:
> When disabled is false, how could we associate CSS classes with it ?

Yeah, you can't.

I sort of followed [this](https://stitches.dev/docs/variants#boolean-variants) (I don't know for sure that `false` is not supported) but simplified the implementation.

> To me it should only accept an array with true and false as keys

I went down this road initially but, since a _real_ `bool` cannot be an array key, I'd have to fudge `0|'false'` and `1|'true'` to represent `false` and `true` respectively. This felt a bit brittle to me.

src/TwigComponent/src/CVA.php
```diff
@@ -53,6 +53,14 @@ public function resolve(array $recipes): string
...
         foreach ($recipes as $recipeName => $recipeValue) {
+            if (isset($this->variants[$recipeName]) && (\is_string($this->variants[$recipeName]) || (\is_array($this->variants[$recipeName]) && array_is_list($this->variants[$recipeName])))) {
+                if (true === $recipeValue) {
+                    $classes .= ' '.implode(' ', (array) $this->variants[$recipeName]);
+                }
+
+                continue;
+            }
+
             if (!isset($this->variants[$recipeName][$recipeValue])) {
```
smnandre, line 64: Hmm.  I fear this become hard to understand for the next ones :) 

I'll try another way, you'll tell me if it's worth the change 🤷

seb-jean:
So, I think it would be interesting to have something like this (to make it clear 😄):

```twig
{% props size = 'md', disabled = false %}

{% set button = cva({
    variants: {
        size: {
            sm: 'alert-sm',
            md: 'alert-md',
            lg: 'alert-lg',
        },
         disabled: {
             true: 'opacity-50 cursor-not-allowed',
             false: 'bg-blue-600 text-orange-400',
         }
    },
}) %}

<button class="{{ button.apply({size, disabled}, attributes.render('class'), 'flex p-4') }}">
    ...
</button>
```

smnandre:
So we're back to the suggestion:  either all your "keys" are boolean, either all are strings ?

src/TwigComponent/src/CVA.php
smnandre: There is something i don't understand in the algorithm will ask you questions about this, but don"t wait for me in this PR, that's 100% implementation and can be handled way after :)

kbond:
> So we're back to the suggestion: either all your "keys" are boolean, either all are strings ?

But keys can't be real booleans so we have to say `"true" === true`, `"false" === false` and/or `1 === true`, `0 === false`. Something like `array<string,string|string[]>|array<"true"|"false",string|string[]>`

smnandre:
Do as you think it's best, and you'll have my full support :))

cavasinf:
https://cva.style/docs/examples/other-use-cases

> cva is really just a fancy way of managing a string…

It would/should treat boolean as a string, even `false`.

------

I think it will be great if these configs work:

```twig
<twig:Alert /> {# Default value #}
```

```twig
<twig:Alert disabled /> {# true #}
```

```twig
<twig:Alert disabled="" /> {# true #}
```

```twig
<twig:Alert disabled="disabled" /> {# true #}
```

```twig
<twig:Alert disabled="anyStringValue" /> {# true #}
```

```twig
<twig:Alert disabled="true" /> {# true #}
```

```twig
<twig:Alert disabled="false" /> {# false #}
```

```twig
<twig:Alert disabled="{{ true }}" /> {# true #}
```

```twig
<twig:Alert disabled="{{ false }}" /> {# false #}
```

```twig
{% set trueBoolean = true %}
<twig:Alert disabled="{{ trueBoolean }}" /> {# true #}
```

```twig
{% set trueBoolean = true %}
<twig:Alert :disabled="trueBoolean" /> {# true #}
```

----------

We achieve that by normalizing the option to a string, and anything but `false` or `0` is `true`.

```php
#[AsTwigComponent]
final class Alert
{
    public string $disabled = 'false';

    #[PreMount]
    {

        $resolver->setDefaults(
            [
                'disabled' => 'false',
            ]
        );

        $resolver->setAllowedTypes('disabled', ['bool', 'string']);
        $resolver->setNormalizer('disabled', function(Options $options, string|bool $value) {
            if (is_bool($value)) {
                $strBool = $value ? 'true' : 'false';
            } else {
                $strBool = $value;
            }

            return preg_match('/^(false|0)$/i', strtolower($strBool)) === 1 ? 'false' : 'true';
        });

    }
}
```

smnandre:
How is this related to CVA ?

cavasinf:
Because variants cannot be `true` only, but `'true'` AND `'false'`.


```twig
cva({
    variants: {
        disabled: 'opacity-50 cursor-not-allowed',
    },
})
```

VS

```twig
cva({
    variants: {
         disabled: {
             true: 'opacity-50 cursor-not-allowed',
             false: 'bg-blue-600 text-orange-400',
         }
    },
})
```

As the doc mentioned above: https://cva.style/docs/examples/other-use-cases

So today, we have to cast boolean value/key to full letter "string boolean".

smnandre:
The moment #1653 is merged, i'll rework this code from @kbond  (yeah i changed to original code ... so i'll assume this and won't ask him to adapt this PR haha) and we'll ship an easier DX, as multiple people ask it :)

kbond:
I agree with @cavasinf in that all keys should be strings/"string booleans". I'll rework this now that #1653 is merged and we can do another round of discussion.

kbond:
I'm in favor of the alternate implementation in #1710. WDYT @seb-jean, @Kocal, @cavasinf?

Kocal:
I'm also in favor of #1710, which keep the usage and maintenance very simple! :)

cavasinf:
I agree with #1710 as well. 👍

PR #1650, opened 2024-03-24
fix: Changed getParent to BaseEntityAutocompleteType instead of depre…

alessandro-podo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | 
| License       | MIT

if you create a new class with the make:autocomplete command, it still use the deprecated class "ParentEntityAutocompleteType" for getParent() return. 
It was deprecated in 93355877d69f95b36479956f430f44b57ae8faf1 

This pull request changes it to the new class "BaseEntityAutocompleteType" in the template

### DIFF
--- a/src/Autocomplete/src/Maker/MakeAutocompleteField.php
+++ b/src/Autocomplete/src/Maker/MakeAutocompleteField.php
@@ -28,7 +28,7 @@ use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\UX\Autocomplete\Form\AsEntityAutocompleteField;
-use Symfony\UX\Autocomplete\Form\ParentEntityAutocompleteType;
+use Symfony\UX\Autocomplete\Form\BaseEntityAutocompleteType;
 
 /**
  * @author Ryan Weaver <ryan@symfonycasts.com>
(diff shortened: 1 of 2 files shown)


smnandre:
Could you add a minimal test ? Or is it too complex ?


(i remember i started this exact PR and wonder if i did stop for this exact reason 😶‍🌫️ )

alessandro-podo:
I'm not sure what exactly you want a test for? @smnandre 
So far I can't find a test for the Commando. 

I can't find any tests for the make commandos in the whole repo. 
If the whole commando is to be tested, wouldn't a new pull request be better?

smnandre:
It's not because something was not tested that it can never be, fortunately!

But i can understand how it feels when you see no other test .. :)

smnandre:
You need to change the content of the template, as the options exposed are not the same in the new class.

alessandro-podo:
That's true. I try to create a test, but i have a problem the with the interactive Question. 
I tried to create the test like [here described.](https://symfony.com/doc/current/components/console/helpers/questionhelper.html#testing-a-command-that-expects-input)

But i get the error `Symfony\Component\Console\Exception\MissingInputException: Aborted.`

```
        self::bootKernel();
        $application = new Application(self::$kernel);
        $command = $application->find('make:autocomplete-field');

        self::assertSame('Generates an Ajax-autocomplete form field class for symfony/ux-autocomplete.', $command->getDescription());

        $commandTester = new CommandTester($command);
        $commandTester->setInputs(['Category', 'CategoryAutocompleteField']);

        $commandTester->execute([],['interactive'=>true]);
``` 

Sorry, but I am not sure how to solve this. :(

smnandre:
Nevermind the test, but if you could update the generated code to follow the new usage of the class (you may look at the doc for reference, or even the class source code)

smnandre:
Done in #1699

PR #1653, opened 2024-03-25
[Twig] CVA - Rework implementation + documentation

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | ..
| License       | MIT

Try another implementation + add some tests.

Updated/added a lot of documentation in the CVA class.

All tests are green but i'm not 100% sure this made no change at all :| 

@WebMamba  I let you decide what you want to do with it (i needed a class to test some things :) ) .. really open to any choice :) 

@kbond that does not include your modifications i think :|

### DIFF
--- a/src/TwigComponent/src/CVA.php
+++ b/src/TwigComponent/src/CVA.php
@@ -12,107 +12,124 @@
 namespace Symfony\UX\TwigComponent;
 
 /**
- * @author Mathéo Daninos <matheo.daninos@gmail.com>
+ * Class Variant Authority (CVA) resolver.
  *
- * CVA (class variant authority), is a concept from the js world.
- * https://cva.style/docs
- * The UI library shadcn is build on top of this principle
- * https://ui.shadcn.com
- * The concept behind CVA is to let you build component with a lot of different variations called recipes.
+ * The CVA concept is used to render multiple variations of components, applying
+ * a set of conditions and recipes to dynamically compose CSS class strings.
+ *
+ * @see https://cva.style/docs
+ *
+ * @doc https://symfony.com/bundles/ux-twig-component/current/index.html
+ *
+ * @author Mathéo Daninos <matheo.daninos@gmail.com>
  *
  * @experimental
  */
 final class CVA
 {
     /**
-     * @var string|list<string|null>|null
-     * @var array<string, array<string, string|list<string|null>>|null the array should have the following format [variantCategory => [variantName => classes]]
-     *                                                ex: ['colors' => ['primary' => 'bleu-8000', 'danger' => 'red-800 text-bold'], 'size' => [...]]
-     * @var array<array<string, string|array<string>>> the array should have the following format ['variantsCategory' => ['variantName', 'variantName'], 'class' => 'text-red-500']
-     * @var array<string, string>|null
+     * @var list<string|null>
+     */
+    private readonly array $base;
+
+    /**
+     * @param string|list<string|null> $base The base classes to apply to the component
      */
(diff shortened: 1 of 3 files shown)


src/TwigComponent/src/CVA.php
```diff
@@ -12,107 +12,134 @@
...
     }
 
-    public function apply(array $recipes, ?string ...$classes): string
+    /**
+     * @deprecated since Symfony 2.17. Use {@see apply()} instead.
+     */
+    public function resolve(array $recipes): string
     {
-        return trim($this->resolve($recipes).' '.implode(' ', array_filter($classes)));
+        trigger_deprecation('symfony/ux-twig-component', '2.17', 'The method "%s()" is deprecated and will be remove in 3.0. Use "apply()" instead.', __METHOD__);
```
kbond: I think we should just kill this method (and not worry about deprecating). I doubt anyone is using it.

smnandre: Good & thanks!

src/TwigComponent/src/Twig/ComponentExtension.php
```diff
@@ -87,25 +87,29 @@ public function finishEmbeddedComponentRender(): void
...
+     * } $cva
      */
     public function cva(array $cva): CVA
     {
         return new CVA(
-            $cva['base'] ?? null,
-            $cva['variants'] ?? null,
-            $cva['compoundVariants'] ?? null,
-            $cva['defaultVariants'] ?? null,
+            $cva['base'] ?? [],
```
nicolas-grekas: `''`?

smnandre: Yep thanks!

src/TwigComponent/src/CVA.php
```diff
@@ -12,107 +12,124 @@
...
  */
 final class CVA
 {
     /**
-     * @var string|list<string|null>|null
-     * @var array<string, array<string, string|list<string|null>>|null the array should have the following format [variantCategory => [variantName => classes]]
-     *                                                ex: ['colors' => ['primary' => 'bleu-8000', 'danger' => 'red-800 text-bold'], 'size' => [...]]
-     * @var array<array<string, string|array<string>>> the array should have the following format ['variantsCategory' => ['variantName', 'variantName'], 'class' => 'text-red-500']
-     * @var array<string, string>|null
+     * @var list<string>
```
WebMamba: Are you sure list<string> include null values? Maybe list<string|null>

WebMamba (approved):
Thanks @smnandre Looks go to me! Thanks for the rework!

src/TwigComponent/src/CVA.php
smnandre: What is the use case to pass null values ?

WebMamba: https://github.com/symfony/ux/issues/1570

smnandre: Could we filter those values in the Twig function and/or the constructor ?

WebMamba: Actually we test for null so I guess your code already works with nullable 😉

smnandre: Ok so just update the docblock, right ?

smnandre: Updated!

kbond (approved):

kbond:
Thanks Simon.

PR #1660, opened 2024-03-25
feat(icons): add `ux:icons:lock` command to "mass import" used icons

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | n/a
| License       | MIT

There has been valid concern over using the on-demand system in production. Icons could change/be removed or the iconify service could be down. This new command searches your twig files and imports valid iconify.design icons in your twig file and mass imports them locally. By default, they aren't overwritten but using the `--force` option enables this.

I think, the official best practice should be:
- on-demand enabled in development only
- part of your workflow includes running `ux:icons:lock` (could be automated)
- (future) `ux:icons:lint` to be used in your CI to ensure all icons are available locally

This could/should be reflected in the official recipe (TODO).

### DIFF
--- a/src/Icons/config/iconify.php
+++ b/src/Icons/config/iconify.php
@@ -12,6 +12,7 @@
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Symfony\UX\Icons\Command\ImportIconCommand;
+use Symfony\UX\Icons\Command\LockIconsCommand;
 use Symfony\UX\Icons\Iconify;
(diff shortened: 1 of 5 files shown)


smnandre:
So if i understand:
* during dev, you use ux_icon(foo), foo is magically downloaded from iconify and stored in cache
* if you call the lock method, it's saved in assets/icons (and can be committed)
* if you call another time lock with --force you can override the local one

This seem really sensible!

(+1 for the lint command idea)

PR #1661, opened 2024-03-25
[CI] Freeze Foundry 1.37

smnandre:
To keep the tests green on the CI, I suggest we stay on Foundry 1.37

Since 1.38, deprecations are triggered to prepare the 2.0, but they make the CI fail on Autocomplete and LiveComponent

We'll release the constraint when we can handle the migration :)

WDYT @kbond ?


kbond (approved):

kbond:
Excellent idea, thanks Simon!

PR #1666, opened 2024-03-29
Create ISSUE_TEMPLATE / config

smnandre:
Update: i now suggest we start with a issue type selector panel, to move on step by step and keep easy-to-fill bug reports / suggestions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
@@ -34,7 +34,7 @@ body:
   - type: textarea
     id: logs
     attributes:
-      label: Relevant log output
+      label: Relevant log/stacktrace output
       description: Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
       render: shell
 
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
@@ -1,0 +1,40 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug", "triage"]
+projects: ["symfony/ux"]
+body:
+  - type: markdown
+    attributes:
+    value: |
+      Thanks for taking the time to fill out this bug report!
+    - type: dropdown
+    id: package
+    attributes:
+      label: Package
+      description: What package is concerned by this $$
+      options:
+        - Autocomplete
+        - Live Component
+        - Icons
+        - Translator
+        - Twig Component
+        - Autre
+      default: 0
+    validations:
+      required: true
+    - type: checkboxes
+    id: terms
+    attributes:
+      label: Code of Conduct
+      description: By submitting this issue, you agree to follow our [Code of Conduct](https://example.com)
+      options:
+        - label: I agree to follow this project's Code of Conduct
+          required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
```diff
@@ -0,0 +1,40 @@
...
+    attributes:
+      label: Package
+      description: What package is concerned by this $$
+      options:
+        - Autocomplete
+        - Live Component
+        - Icons
+        - Translator
+        - Twig Component
+        - Autre
```
evertharmeling, line 22: `Autre` -> `Other` :) 

Should we add all components?

```
        - Autocomplete
        - Chartjs
        - Cropperjs
        - Dropzone
        - Icons
        - LazyImage
        - LiveComponent
        - Notify
        - React
        - StimulusBundle
        - Svelte
        - Swup
        - TogglePassword
        - Translator
        - Turbo
        - TwigComponent
        - Typed
        - Vue
```

smnandre: I had the idea to select only the most frequent in support... as a big list is always something a bit frightening :|

evertharmeling: haha yeah I get it! Although the list would probably not change that often and it could be a bit more directive / clear for the bug reporter (and thus bugfixer). At least we should change the `Autre` to `Other` or are you trying to sneak in little bits of french 😅

smnandre: Hahah yes of course baguette... 

I have really no strong opinion, not even sure it's better than "free textarea'... i think we should try something and iterate (and i'm not against starting with all of them :) )

i'm waiting the "staff" to give their opinion (when they have time, this has a -5000 prioriy like no others before 😅 )

.github/ISSUE_TEMPLATE/BUG_REPORT.yaml
```diff
@@ -0,0 +1,40 @@
...
+    attributes:
+      label: Code of Conduct
+      description: By submitting this issue, you agree to follow our [Code of Conduct](https://example.com)
+      options:
+        - label: I agree to follow this project's Code of Conduct
+          required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant log output
```
94noni:
```suggestion
      label: Relevant log/stacktrace output
```

wdyt? a log can bo top level, a stack trace can help dig faster :)

javiereguiluz:
This looks like something useful to have. Is there anything else left to do before merging? Thanks!

smnandre:
I think we need to discuss what information we want to get from this form ? Any feedback / advice ?

smnandre:
## I changed my mind

Well we have too many packages i'm not sure this is still a good idea (and carsonbot helps us to sort things)

packages

(dummy data but this gives a good idea of what i would look like)

## Suggestion

💡 I think the best thing today would be a "issue selector panel" (and keep very free / simple forms for now for features and bugs)

Almost exactly the same as php-cs-fixer: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/new/choose

phpcs

So If you're ok to test 'll open a PR monday with the items i had in mind :) 

cc @WebMamba @kbond @Kocal

Kocal:
That's would be great yet!
Do you plan to keep the `select` for the packages? It liked it a lot

smnandre:
Closing this one: replaced by #2084 for the first part / then we will deal form by form :)

PR #1672, opened 2024-04-02
[Website] Fix some links in the UX Icons page

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

@tareqas found and reported these errors to me ... so let's fix them!

PR #1673, opened 2024-04-02
[Site] Add "Icons" menu item

smnandre:
* Add Icon menu item in nav
* Remove dead code
* Add test
* Fix footer links

### DIFF
--- a/ux.symfony.com/templates/_header.html.twig
+++ b/ux.symfony.com/templates/_header.html.twig
@@ -29,6 +29,10 @@
             <div class="AppNav_menu">
                 <a href="{{ path('app_turbo') }}" class="AppNav_item">Turbo</a>
                 <a href="{{ path('app_live_component') }}" class="AppNav_item">Live <span>Components</span></a>
+                <a href="{{ path('app_icons') }}" class="AppNav_item">
+                    <span class="AppNav_badge">New</span>
+                    Icons
+                </a>
                 <a href="{{ path('app_packages') }}" class="AppNav_item">Packages</a>
                 <a href="{{ path('app_demos') }}" class="AppNav_item">Demos</a>
             </div>
--- a/ux.symfony.com/assets/styles/sections/_nav.scss
+++ b/ux.symfony.com/assets/styles/sections/_nav.scss
@@ -107,19 +107,28 @@
   }
 }
 
-
 .AppNav_badge {
   position: absolute;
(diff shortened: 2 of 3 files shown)


smnandre:
Oops.. some override https://github.com/symfony/ux/pull/1672 sorry :)

smnandre:
I inserted changes from #1672

javiereguiluz (approved):

ux.symfony.com/templates/_header.html.twig
```diff
@@ -29,6 +29,10 @@
             <div class="AppNav_menu">
                 <a href="{{ path('app_turbo') }}" class="AppNav_item">Turbo</a>
                 <a href="{{ path('app_live_component') }}" class="AppNav_item">Live <span>Components</span></a>
+                <a href="{{ path('app_icons') }}" class="AppNav_item">
```
WebMamba, line 32: Why add this on the navbar? I thing it's better to add this on the packages list? WDYT?

smnandre: @WebMamba It's what was plannned with Ryan and Kevin.. we wanted to showcase it as people come here not only to see the package doc/demo (as for the other packages), but to access the Icons search !

WebMamba: Ha yes ok good idea!

WebMamba:
Hey @smnandre this is detail, but I think the new at the top of the icon nav item should be more visible. Maybe we can make it bigger? Or add a background around? In mobile, the new is far from the world Icon so we don't get that the new is for icon. See some screenshots here:
Capture d’écran 2024-04-05 à 23 40 39
Capture d’écran 2024-04-05 à 23 40 13

smnandre:
> Hey @smnandre this is detail, but I think the new at the top of the icon na [...]

Fixed by ... removing it as it has not the same goal on desktop where we want from the homepage to show it.. In the menu were someone already look at this different pages it has less use i think.

WebMamba (approved):
Thanks @smnandre it looks good! 

There a few bugs in mobile that we should fix in other PRs. I gonna do an issue for it 😁

javiereguiluz:
Let's merge this to give Icons more exposure ... and we can tweak it later if we don't like anything. Thanks Simon! and thanks Matheo for the review.

PR #1679, opened 2024-04-04
[Demo] Infinite scroll 1/2

smnandre:
New demo in two parts: part 1 here

(second part will showcase the automatic load via intersection observer)

Capture d’écran 2024-04-04 à 02 03 50
Capture d’écran 2024-04-04 à 02 03 57


seb-jean:
Nice ❤️

ux.symfony.com/src/Twig/ProductGrid.php
```diff
@@ -0,0 +1,74 @@
...
+
+    #[LiveProp]
+    public int $page = 1;
+
+    public function __construct(private readonly EmojiCollection $emojis)
+    {
+    }
+
+    #[LiveAction]
+    public function more(): void
```
94noni, line 37:
```suggestion
    public function next(): void
```

?

ux.symfony.com/src/Service/EmojiCollection.php
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @internal
+ */
+final class EmojiCollection implements IteratorAggregate, \Countable
+{
+    private array $emojis;
+
+    public function __construct(array $emojis = [])
+    {
+        $this->emojis = $emojis ?: $this->loadEmojis();
```
94noni, line 21: is it "ok" to do it in constructor ?
I know its memory array, but to have better code, perhaps compute this outside of constructor ?

ux.symfony.com/templates/components/ProductGrid.html.twig
```diff
@@ -0,0 +1,43 @@
+<div class="ProductGrid" {{ attributes.defaults(stimulus_controller('scroll')) }}>
+
+    <div class="p-4">
+        <div id="results" style="display: flex; gap: 1rem; flex-direction: column;">
+
+            {% if this.page > 1 %}
+                {#  🦊 #}
+                {# Adding a fake "previous page" div is enough to trick the system #}
+                {# It must have the same ID than the original page #}
+                <div class="ProductGrid_page" id="page--{{ this.page - 1 }}" data-live-ignore="true"></div>
```
94noni, line 10:
```suggestion
                <div class="ProductGrid_page" id="page--{{ this.currentPage }}" data-live-ignore="true"></div>
```

and add inside the component ?

ux.symfony.com/templates/components/ProductGrid.html.twig
```diff
@@ -0,0 +1,43 @@
...
+                {% for item in this.items %}
+                    <article class="ProductGrid_item" data-num="{{ item.id }}"
+                             style="--color: {{ item.color }};">
+                        <div class="ProductGrid_media">
+                            <svg>
+                                <use href="#svg-tshirt"/>
+                            </svg>
+                            <span>{{ item.emoji }}</span>
+                        </div>
+                        <data value="{{ item.id }}">$ {{ item.id + 1 }}<small>.99</small></data>
```
94noni, line 24:
```suggestion
                        <data value="{{ item.id }}">$ {{ item.id + 1 }}<small>.{{ random(0, 99) }}</small></data>
```

^^

ux.symfony.com/templates/components/ProductGrid.html.twig
```diff
@@ -0,0 +1,43 @@
...
+                {% endfor %}
+            </div>
+
+            {% if this.hasMore %}
+                <div style="display: grid; place-content: center;padding: 4rem;">
+                    <button
+                        data-action="live#action"
+                        data-live-action-param="more"
+                        data-scroll-target="loader"
+                        class="btn btn-primary">Load More (page {{ this.page + 1 }})
```
94noni, line 35:
```suggestion
                        class="btn btn-primary">Load More (page {{ this.nextPage }})
```

ux.symfony.com/src/Twig/ProductGrid.php
smnandre: Not sure, as the idea is there to push the "load more" idea, instead of a "next page" ... maybe loadMore() ?

ux.symfony.com/templates/components/ProductGrid.html.twig
smnandre: I don't get it ? The currentPage is the one with results afterwards, there the idea is to mimick the existence of the results from the previous page, to "trick" the morphing into believing it has to append instead of replacing

ux.symfony.com/src/Service/EmojiCollection.php
smnandre: In real life, with a service, you would probably do it in a dedicated runtime getter, indeed. But in order to do short demo and showcase the code in the page, we often have to trade some "best practices" for better readability. 

As shown in the page, the goal is really  here to insist on the LiveProp / LiveAction so i'd like to keep the rest of the code minimal.

WDYT ?

ux.symfony.com/templates/components/ProductGrid.html.twig
smnandre: Same as before, i do agree in an ideal world, but that would add getters in the component class, and i'd like it to be shortest possible to focus on the demo behaviour.

(Your poiints are valid, and thank for posting them ... i'm sorry if this may feel a bit like "ok but no", i hope i made my point clear enough concerning the reasons of those "missing" things)

94noni: Yes no worry i was just commenting while reviewing this PR, like speaking outloud dont feel sorry ^^
I Took some inspiration on it passing by :)

smnandre: Maybe we should find a way to showcase a "limited" version of the code in the demo pages, while still having best-practices code in the source for everyone wanting to take inspiration... 

That's a interesting thing to note in todo-list!

ux.symfony.com/assets/controllers/browser-controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+
+    connect() {
+
+    }
+
+    disconnect() {
+
+    }
+
```
kbond:
```suggestion
```

ux.symfony.com/assets/controllers/browser-controller.js
```diff
@@ -0,0 +1,20 @@
...
+    reduce() {
+        this.element.classList.toggle('Browser--reduced');
+    }
+
```
kbond:
```suggestion
```

kbond (commented):
Just some small things.

kbond:
Thank you Simon.

PR #1681, opened 2024-04-04
Update RecipeFileTree.php

94noni:
such texts are display on hover on homepage
I think it is relevant to at least document that such files are optional sometime aka with assetmapper

### DIFF
--- a/ux.symfony.com/src/Model/RecipeFileTree.php
+++ b/ux.symfony.com/src/Model/RecipeFileTree.php
@@ -26,8 +26,8 @@ class RecipeFileTree
             ->addFile('assets/controllers/hello_controller.js', 'An example controller. Add it to any element with <code class="text-nowrap">data-controller="hello"</code>')
             ->addDirectory('assets/styles')
             ->addFile('assets/styles/app.css', 'Your main CSS file')
-            ->addFile('package.json', 'Holds your node dependencies, most importantly Stimulus & Webpack Encore.')
-            ->addFile('webpack.config.js', 'Configuration file for Webpack Encore: the tool that processes and combines all of your CSS and JS files.')
+            ->addFile('package.json', 'Holds your node dependencies, most importantly Stimulus (when using Webpack Encore).')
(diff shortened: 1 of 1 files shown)


ux.symfony.com/src/Model/RecipeFileTree.php
```diff
@@ -26,8 +26,8 @@ public function __construct()
             ->addFile('assets/controllers/hello_controller.js', 'An example controller. Add it to any element with <code class="text-nowrap">data-controller="hello"</code>')
             ->addDirectory('assets/styles')
             ->addFile('assets/styles/app.css', 'Your main CSS file')
-            ->addFile('package.json', 'Holds your node dependencies, most importantly Stimulus & Webpack Encore.')
-            ->addFile('webpack.config.js', 'Configuration file for Webpack Encore: the tool that processes and combines all of your CSS and JS files.')
+            ->addFile('package.json', 'Holds your node dependencies, most importantly Stimulus (when using Webpack Encore).')
+            ->addFile('webpack.config.js', 'Configuration file for Webpack Encore: the tool that processes and combines all of your CSS and JS files (not needed with AssetMapper).')
```
smnandre, line 30: We will support webpack encore, but i don't think we want to showcase it in the "default case", as it won't be there for people installing a new Symfony webapp. 

So those two files could probably be explained on the documentation, but right there it's kinda the opposite of what we're trying to promote for easy plug-n-play DX.

WDYT ?

94noni: you mean those can be removed completely ?
Why not, and somewhere else document that such file may exists if webpack encore is still use in the app

smnandre: I mean.. it's kinda odd to brand "assetmapper" as "no-npm-tools" and showcase them here, right ? So maybe we can do the jumb.

Another idea would be to have some tabbed content here with "assetmapper / webpack" but i feel on this place, maybe we should be the most direct/simple we can.

smnandre:
I will soon open a PR to remove this block entirely (our visitors do not look for the same things as when this website was first launched) ... agree to close this one ?

94noni:
Yes make sens, thx for the ping

PR #1682, opened 2024-04-04
Update homepage.html.twig

94noni:
following previous PR, with the bloc showcasing hello_controller.js, direct after seeing markdown-controller.js seems not logic (even if stimulus under the hood is ok for both identifiers)

Capture d’écran 2024-04-04 à 22 27 06


ux.symfony.com/templates/main/homepage.html.twig
```diff
@@ -71,7 +71,7 @@
         </div>
         <div class="col-12 col-md-6  mt-3 mt-md-0">
             <twig:CodeBlock
-                filename="assets/controllers/markdown-controller.js"
+                filename="assets/controllers/markdown_controller.js"
```
smnandre, line 74: I think today we used "-controller" as naming convention..

Not sure we should change for only one :|

94noni: What about renaming then the hello-controller?

smnandre: If you think that'll improve the coherence yes for sure!

smnandre:
(same thing here, ok for you @94noni ?)

Issue #1676, opened 2024-04-03
[LiveComponent] JavaScript hooks are not fired for custom controller on mount


YummYume:
Hello ! I believe this issue is the same as #1038, which was never truly resolved. When a component mounts, and you listen to pretty much any event like `connect` or `render:finished`, the event is never fired. However, it **is** registered, as events that are triggered later on are indeed fired correctly. The `hooks` property of the component also contains the callbacks I gave it.

This is a dead-simple reproduction of my case (I can't share the real code), using the example from the [documentation](https://symfony.com/bundles/ux-live-component/current/index.html#javascript-component-hooks) :

```ts
import { Controller } from '@hotwired/stimulus';
import { getComponent, type Component } from '@symfony/ux-live-component';

export default class MyController extends Controller<HTMLElement> {
  component: Component | null = null;

  async initialize() {
    this.component = await getComponent(this.element);

    this.component.on('render:finished', (component: Component) => {
      console.log('render:finished', component);
    });
  }
}
```

```twig
<ul
    class="{{ ('flex ' ~ attributes.render('class')|default)|tailwind_merge }}"
    style="gap: {{ gap }}px; {{ attributes.render('style')|default }}"
    {{ attributes.defaults({
        'data-controller': 'my-controller',
    }) }}
>
    {# ... #}
</ul>
```

Nothing gets logged on the first initial render, but updating the model later on (with a simple button for example) works just fine. What I'm suspecting is that there's a race condition happening, where my controller loads after the events are dispatched. I tried removing the lazy load on the controller, or even add the `defer` attribute on the Live Component, but no success. I'd really like to see this work and possibly control the first render myself (I need to get the screen width of the user and update the component accordingly, ideally showing a loader in the meantime).

Side notes :

- Would it be possible to improve the TypeScript definitions for `on` so that it would automatically type the function arguments based on the event ?
- The TypeScript definitions for `component.set` and `component.render` always return a `Promise<default>`, however in the [documentation](https://symfony.com/bundles/ux-live-component/current/index.html#working-with-the-component-in-javascript), no `await` or `.then` is present between the calls, is this intended ?

smnandre:
Could you show what the generated HTML looks like for your component root tag (before any script execution) ? 

```
<ul
    class="{{ ('flex ' ~ attributes.render('class')|default)|tailwind_merge }}"
    style="gap: {{ gap }}px; {{ attributes.render('style')|default }}"
    {{ attributes.defaults({
[...]

PR #1685, opened 2024-04-05
[Live] Add signature overload for on and off methods of component

YummYume:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no
| Issues        |
| License       | MIT

Following some feedback in #1676. This PR only affects TypeScript typing.

Provides automatic typing for the `on` and `off` events of a component, depending on the event's name. Also changes the return type allowed from an event listener to allow passing an asynchronous function. This is a pretty common thing to do (the need to use `await`), but TypeScript/linters will often yell at you for doing so if the event listener does not expect a `Promise` to be returned.

(also fixed a typing issue with `NodeJS.Timer` being used instead of `number`)

### DIFF
--- a/src/LiveComponent/assets/dist/Component/index.d.ts
+++ b/src/LiveComponent/assets/dist/Component/index.d.ts
@@ -1,8 +1,24 @@
 import { BackendInterface } from '../Backend/Backend';
 import ValueStore from './ValueStore';
+import BackendRequest from '../Backend/BackendRequest';
 import { ElementDriver } from './ElementDriver';
 import { PluginInterface } from './plugins/PluginInterface';
 import BackendResponse from '../Backend/BackendResponse';
+type MaybePromise<T = void> = T | Promise<T>;
+export type ComponentHooks = {
+    'connect': (component: Component) => MaybePromise;
+    'disconnect': (component: Component) => MaybePromise;
+    'request:started': (requestConfig: any) => MaybePromise;
+    'render:finished': (component: Component) => MaybePromise;
+    'response:error': (backendResponse: BackendResponse, controls: {
+        displayError: boolean;
+    }) => MaybePromise;
+    'loading.state.started': (element: HTMLElement, request: BackendRequest) => MaybePromise;
+    'loading.state.finished': (element: HTMLElement) => MaybePromise;
+    'model:set': (model: string, value: any, component: Component) => MaybePromise;
+};
+export type ComponentHookName = keyof ComponentHooks;
+export type ComponentHookCallback<T extends string = ComponentHookName> = T extends ComponentHookName ? ComponentHooks[T] : (...args: any[]) => MaybePromise;
 export default class Component {
     readonly element: HTMLElement;
     readonly name: string;
@@ -30,8 +46,8 @@ export default class Component {
     addPlugin(plugin: PluginInterface): void;
     connect(): void;
     disconnect(): void;
-    on(hookName: string, callback: (...args: any[]) => void): void;
-    off(hookName: string, callback: (...args: any[]) => void): void;
+    on<T extends string | ComponentHookName = ComponentHookName>(hookName: T, callback: ComponentHookCallback<T>): void;
+    off<T extends string | ComponentHookName = ComponentHookName>(hookName: T, callback: ComponentHookCallback<T>): void;
     set(model: string, value: any, reRender?: boolean, debounce?: number | boolean): Promise<BackendResponse>;
     getData(model: string): any;
     action(name: string, args?: any, debounce?: number | boolean): Promise<BackendResponse>;
@@ -55,3 +71,4 @@ export default class Component {
     _updateFromParentProps(props: any): void;
 }
 export declare function proxifyComponent(component: Component): Component;
+export {};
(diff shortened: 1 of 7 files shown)


smnandre:
Thank you for that @YummYume !

Could you rebuild the dist file ? It feels you updated those files "manually" (i may be wrong on this, TS is not my strongest force 😅 )

From the root of the monorepo, this should be enough i think

npm run build 

(This is why the CI is red `Symfony UX / Check for UnBuilt JS Dist Files (pull_request) Failing after 1m`)

YummYume:
This is weird :thinking: I didn't touch the `dist` files I simply ran the `yarn build` command. And somehow the Autocomplete controller needed to be rebuilt. Now it's fine.

No idea why the LiveComponent tests aren't passing, though, they're all green.

smnandre:
Related to dependencies / Symfony 7.1.. and don't bother with Autocomplete if you're not working on it ... or you won't ever end this PR haha

src/LiveComponent/assets/src/PollingDirector.ts
```diff
@@ -55,7 +55,7 @@ export default class {
             };
         }
 
-        const timer = setInterval(() => {
+        const timer = window.setInterval(() => {
```
WebMamba, line 58: I am not sure about this fix. setInterval is native function well supported why use the one from window?

src/LiveComponent/assets/dist/Component/index.d.ts
```diff
@@ -30,8 +33,34 @@ export default class Component {
     addPlugin(plugin: PluginInterface): void;
     connect(): void;
     disconnect(): void;
-    on(hookName: string, callback: (...args: any[]) => void): void;
-    off(hookName: string, callback: (...args: any[]) => void): void;
+    on(hookName: 'connect', callback: (component: Component) => MaybePromise): void;
```
WebMamba: Instead of having all of those definitions why not just use type ComponentHookName?
on(hookName: ComponentHookName, callback: (component: Component) => MaybePromise): void;

src/LiveComponent/assets/src/PollingDirector.ts
YummYume: Using `setInterval` or `window.setInterval` results in the same thing in a browser context. However, when using `setInterval`, TypeScript was mistakenly thinking that this was the Nodejs `setInterval` function, which doesn't return a number like in a browser, which was causing a typing error warning during builds.

src/LiveComponent/assets/dist/Component/index.d.ts
YummYume: Because the callbacks receive different arguments depending on the event. By explicitly overloading this method, TypeScript (and your IDE) now knows which events are available, and the arguments that they each receive. It's a much nicer developer experience, and it should work for any IDE even when not using TypeScript.

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -122,11 +126,31 @@ export default class Component {
      *     * loading.state:finished (element: HTMLElement) => {}
      *     * model:set (model: string, value: any, component: Component) => {}
      */
-    on(hookName: string, callback: (...args: any[]) => void): void {
+    on(hookName: 'connect', callback: (component: Component) => MaybePromise): void;
```
squrious: What do you think about adding callback types too? Like `ConnectHook` or `ConnectCallback`

squrious: Just a suggestion, I'm not so comfortable with TypeScript, but what would you think about something like that?

```ts
type Hooks = {
    'connect': (component: Component) => MaybePromise,
    'disconnect': (component: Component) => MaybePromise,
    'request:started': (requestConfig: any) => MaybePromise
    'render:finished': (component: Component) => MaybePromise
    'response:error': (backendResponse: BackendResponse, controls: {displayError: boolean}) => MaybePromise
    'loading.state.started': (element: HTMLElement, request: BackendRequest) => MaybePromise
    'loading.state.finished': (element: HTMLElement) => MaybePromise
    'model:set': (model: string, value: any, component: Component) => MaybePromise,
}

type ComponentHookName = keyof Hooks;

type HookCallback<T extends string> = T extends ComponentHookName
    ? Hooks[T]
    : (...args: any[]) => MaybePromise;

class Component {
    on<T extends ComponentHookName|string>(name: T, callback: HookCallback<T>): void {
        this.hooks.register(hookName, callback);
    }
    off<T extends ComponentHookName|string>(name: T, callback: HookCallback<T>): void {
        this.hooks.unregister(hookName, callback);
    }
}


```

YummYume: Hi @squrious ! I tested this approach, and after a few tweaks, it gives the same result as the current implementation, and is a bit easier to use for maintaining many events, rather than overloading the methods manually. I personally wouldn't be opposed to it. :smile:

squrious: Nice :smile:

YummYume: So I went ahead and implemented your idea with the conditional, it's way more maintainable and achieves the same thing. :+1:

smnandre (approved):

src/LiveComponent/assets/src/PollingDirector.ts
WebMamba: Ok thanks for your answer!

WebMamba (approved):
Sounds good to me! Thanks a lot @YummYume!

kbond:
Awesome first contribution @YummYume, thank you!

PR #1688, opened 2024-04-05
[Icons] Fix Icon missing width/height

### LINKED ISSUE #1687: [Icons] Width / Height error when using an icon
Hi!
I've just installed the bundle and tried it out.
So far the 2 icons below are working fine.

```HTML
<twig:UX:Icon name="flowbite:user-solid" class="w-6 h-6"/>
<twig:UX:Icon name="tabler:sun" class="w-6 h-6"/>
```
If I try to use a heroicon the following error appears:
```HTML
<twig:UX:Icon name="heroicons:sun" class="w-6 h-6"/>
```
`Error rendering "UX:Icon" component: The icon "heroicons:sun" does not have a width or height.`

I've tried it with another icon (heroicons:sun-solid) and the same error appears.
Seems to me as there is something wrong with the heroicons?

Best regards
Alex

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1687
| License       | MIT

Extract width height from global response attributes.

### DIFF
--- a/src/Icons/src/Iconify.php
+++ b/src/Icons/src/Iconify.php
@@ -65,8 +65,8 @@ final class Iconify
             throw new IconNotFoundException(sprintf('The icon "%s:%s" does not exist on iconify.design.', $prefix, $name));
         }
 
-        $height = $data['icons'][$name]['height'] ?? $this->sets()[$prefix]['height'] ?? null;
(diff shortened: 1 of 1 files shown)


smnandre:
(CI fails unrelated -- LiveComponent vs Symfony 7.1 TypeInfo)

javiereguiluz:
Thanks for fixing this bug Simon.

PR #1689, opened 2024-04-06
[Doc] Add resolve condition names to UX Svelte documentation

kasperh90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

This PR will enhance the UX Svelte documentation. As of yet, some internal functionality will not work in Svelte unless you add ``browser`` and ``svelte`` to the ``conditionNames``. I think it is valuable to mention this in the documentaiton.

### DIFF
--- a/src/Svelte/doc/index.rst
+++ b/src/Svelte/doc/index.rst
@@ -51,6 +51,22 @@ Next, install a package to help Svelte:
 That's it! Any files inside ``assets/svelte/controllers/`` can now be rendered as
 Svelte components.
 
+If you are using Svelte 4, you will have to add ``browser``, ``import`` and ``svelte``
+to the ``conditionNames`` array. This is necessary as per `the Svelte 4 migration guide`_
+for bundlers such as webpack, to ensure that lifecycle callbacks are internally invoked.
+
+To modify the ``conditionNames`` array, append the following changes to the bottom
+of your ``webpack.config.js`` file:
+
+.. code-block:: javascript
+
+    // webpack.config.js
+
+    // module.exports = Encore.getWebpackConfig();
+    const config = Encore.getWebpackConfig();
+    config.resolve.conditionNames = ['browser', 'import', 'svelte'];
+    module.exports = config;
+
 Usage
 -----
 
@@ -146,3 +162,4 @@ https://symfony.com/doc/current/contributing/code/bc.html
 .. _`Svelte`: https://svelte.dev/
 .. _`the Symfony UX initiative`: https://symfony.com/ux
 .. _StimulusBundle configured in your app: https://symfony.com/bundles/StimulusBundle/current/index.html
+.. _the Svelte 4 migration guide: https://svelte.dev/docs/v4-migration-guide#browser-conditions-for-bundlers


src/Svelte/doc/index.rst
```diff
@@ -51,6 +51,19 @@ Next, install a package to help Svelte:
 That's it! Any files inside ``assets/svelte/controllers/`` can now be rendered as
 Svelte components.
 
+To make internal functions work in Svelte, such as ``onMount``, ``onDestroy``, etc,
```
WebMamba: Could you add a link to the documentation you mention?

kasperh90: Yes, sure! I got this link: 
https://svelte.dev/docs/v4-migration-guide#browser-conditions-for-bundlers

Should I add it to the documentation?

WebMamba: Maybe we should mention those steps are needed when you want to use svelte 4? You don't have this kind of issue in version 3?

src/Svelte/doc/index.rst
```diff
@@ -51,6 +51,19 @@ Next, install a package to help Svelte:
...
+you should add ``browser`` and ``svelte`` to the ``conditionNames`` array. To do so, append the following changes
+to the bottom of your ``webpack.config.js`` file:
+
+.. code-block:: javascript
+
+    // webpack.config.js
+
+    // module.exports = Encore.getWebpackConfig();
+    const config = Encore.getWebpackConfig();
+    config.resolve.conditionNames = ['browser', 'svelte'];
```
WebMamba: In the documentation you send, they motion also an import item. Do you think we should added to?

> conditionNames: ['svelte', 'browser', 'import']

src/Svelte/doc/index.rst
WebMamba: Yes perfect!

kasperh90: Great. I will do it tomorrow. 👍

src/Svelte/doc/index.rst
kasperh90: Agree. I will add it too.

src/Svelte/doc/index.rst
kasperh90: Yes, you are right. I will mention this.

kasperh90:
@WebMamba I think we need to consider one more thing:

When I create a new Symfony app, install the UX Svelte bundle, install all the necessary NPM dependencies and type ``npm run watch`` in the terminal, I get the following warning:

image

That's how it came to my attentien that when you use Svelte 4, you will have to do a bit more configuration where you will have to add ``svelte`` and ``browser`` to ``conditionNames`` array. The warning in the terminal provides the following link: https://github.com/sveltejs/svelte-loader#resolveconditionnames

Should I add this too?

WebMamba:
Hummm I think it's fine, the error message is really good. Just specify in the doc that this config is needed for svelte 4

javiereguiluz:
Kasper, thanks for this nice contribution and for your patience during the review process (thanks also to Matheo for the nice review).

Your contribution was perfectly correct, but while merging I changed the code example to show a `diff` block (see https://github.com/symfony/ux/commit/22a47f41bcba89ac9b6aa29159928ca2b7e4c491). I think that way it's easier to know what to add/remove in the config file. Thanks!

kasperh90:
@javiereguiluz you are welcome! And using the ``diff`` block is brilliant. 👍

PR #1690, opened 2024-04-07
[Icons] Set aria-hidden="true" when title/label not set

### LINKED ISSUE #1629: [Icons][DX] Accessibility helpers
Based on the proposed documentation update [here](https://github.com/symfony/ux/pull/1602/files#diff-fe1ff4f12fbf2e62fcb5d1b3516d7c5acf18ac4229ebd74534a1b47fb469715cR214) from @smnandre, what do you think about the following helpers:

- `ux_icon_accessible(icon, label, attributes)` for informative and functional icons, will enforce the label parameter and always add the `aria-label` attribute
- `ux_icon_decorative(icon, attributes)` for obviously decorative icons, will always add the `aria-hidden` attribute set to `true`

I can try to come with a PR if the idea get some traction, but I'm fine if someone else takes it :)

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #1629
| License       | MIT

Add "aria-hidden=true" when the icon has no title, aria-label or aria-labelledby attribute 

(option enabled per default)


src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -43,6 +43,15 @@ public function getConfigTreeBuilder(): TreeBuilder
...
                 ->end()
+                ->arrayNode('renderer')
+                    ->info('Options for the icon renderer.')
+                    ->children()
+                        ->booleanNode('auto_aria_hidden')
+                            ->info('Automatically set aria-hidden="true" when no label or title is provided.')
+                            ->defaultTrue()
+                        ->end()
+                    ->end()
+                ->end()
```
WebMamba: I don't think we need to add a configuration here. I think document the behaviour should be enough

WebMamba (commented):
Why instead of an aria-hidden attribute not fill the aria-label with the name from the metadata ?

smnandre:
> Why instead of an aria-hidden attribute not fill the aria-label with the name from the metadata ?

I don't remember precisely where, but i posted some links about it in the original issue discussion... 

aria-hidden is really the best thing to do when an icon is used with a label (like the green "open" badge on top of this screen)

But your solution could absolutely be something to consider 

(that's why i'm inclined to keep the configuration, there are so many use cases i'd rather no "force" things for users that don't want some attributes.... or have very good reasons to not want them)

src/Icons/tests/Unit/IconRendererTest.php
```diff
@@ -159,6 +159,45 @@ public static function provideAttributesWithDefaultAttributesCases()
...
+        $iconRenderer = new IconRenderer($registry, [], ['auto_aria_hidden' => true]);
+
+        $svg = $iconRenderer->renderIcon('foo', $attributes);
+        $this->assertSame($expectedSvg, $svg);
+    }
+
+    /**
+     * @return iterable<array{array<string, string>, string}>
+     */
+    public static function provideAriaHiddenCases(): iterable
```
yguedidi, line 182: missing the test for `aria-label`

src/Icons/tests/Unit/IconRendererTest.php
```diff
@@ -159,6 +159,45 @@ public static function provideAttributesWithDefaultAttributesCases()
...
+    {
+        yield 'no attributes' => [
+            [],
+            '<svg aria-hidden="true"><path d="M0 0L12 12"/></svg>',
+        ];
+        yield 'aria-hidden attribute' => [
+            ['aria-hidden' => 'true'],
```
yguedidi, line 189: I'd add a case where it's forced to false and another attribute is passed to, e.g. `aria-label`

yguedidi (changes requested):
Thank you @smnandre for taking this!
so in the end no special `label` attribute accepted? I'm fine with the `title` one only, just curious

yguedidi (approved):

smnandre:
> > Why instead of an aria-hidden attribute not fill the aria-label with the name from the metadata ?
> 
> I don't remember precisely where, but i posted some links about it in the original issue discussion...
> 
> aria-hidden is really the best thing to do when an icon is used with a label (like the green "open" badge on top of this screen)
> 
> But your solution could absolutely be something to consider
> 
> (that's why i'm inclined to keep the configuration, there are so many use cases i'd rather no "force" things for users that don't want some attributes.... or have very good reasons to not want them)

Ok for you @WebMamba  ?

WebMamba:
Hum no 😁 I don't think it should be a configuration, it should be a behavior that can be overwrite directly when you use the component. And we should document it

smnandre:
> Hum no 😁 I don't think it should be a configuration, it should be a behavior that can be overwrite directly when you use the component. And we should document it

So people would need to add "aria-hidden: false" for all their icons ? 

That's a niche use case, so why not yes

kbond:
> so in the end no special label attribute accepted?

Is `label` a valid html attribute? If not, I'd prefer keeping as is to keep the complexity to a minimum.

> I don't think it should be a configuration, it should be a behavior that can be overwrite directly when you use the component.

I think this should be the default behaviour for the reasons discussed in #1629 but I want to hear your hesitation @WebMamba. Is there an alternate way we can achieve? My thought for this was adding to the `default_icon_attributes` configuration option but because some logic is required, this seems better.

smnandre:
> Is `label` a valid html attribute? If not, I'd prefer keeping as is to keep the complexity to a minimum.

Yep that was my reasoning to not implement this in the end. I don't want to confuse users with non-existing svg attributes.

src/Icons/src/IconRenderer.php
```diff
@@ -54,10 +48,8 @@ private function getIconAttributes(string $name, array $attributes): array
         $iconAttributes = $this->defaultIconAttributes;
 
         // Add aria-hidden attribute
-        if ($this->rendererConfiguration['auto_aria_hidden'] ?? false) {
-            if ([] === array_intersect(['aria-hidden',  'aria-label', 'aria-labelledby', 'title'], array_keys($attributes))) {
-                $iconAttributes['aria-hidden'] = 'true';
-            }
+        if ([] === array_intersect(['aria-hidden',  'aria-label', 'aria-labelledby', 'title'], array_keys($attributes))) {
```
WebMamba, line 51: simple curiosity where did you get this list?

WebMamba (approved):
Thanks @smnandre! This DX looks good! Sorry to ask you to rewrite all the tests 😁

src/Icons/src/IconRenderer.php
smnandre: > An image’s alt attribute.
    A SVG’s child [title element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title).
    Using aria-label or aria-labelledby on an image or SVG.
    Using [a title attribute](https://w3c.github.io/html-aam/#img-element) on an image or SVG.
    
https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html

src/Icons/tests/Integration/RenderIconsInTwigTest.php
```diff
@@ -26,13 +26,13 @@ public function testRenderIcons(): void
...
+                <li id="second"><svg viewBox="0 0 24 24" fill="currentColor" class="w-6 h-6" aria-hidden="true"><path fill-rule="evenodd" d="M7.5 6a4.5 4.5 0 1 1 9 0 4.5 4.5 0 0 1-9 0ZM3.751 20.105a8.25 8.25 0 0 1 16.498 0 .75.75 0 0 1-.437.695A18.683 18.683 0 0 1 12 22.5c-2.786 0-5.433-.608-7.812-1.7a.75.75 0 0 1-.437-.695Z" clip-rule="evenodd"></path></svg></li>
+                <li id="third"><svg viewBox="0 0 24 24" fill="currentColor" class="w-6 h-6" aria-hidden="true" data-action="string &quot;with&quot; quotes"><path fill-rule="evenodd" d="M19.916 4.626a.75.75 0 0 1 .208 1.04l-9 13.5a.75.75 0 0 1-1.154.114l-6-6a.75.75 0 0 1 1.06-1.06l5.353 5.353 8.493-12.74a.75.75 0 0 1 1.04-.207Z" clip-rule="evenodd"></path></svg></li>
+                <li id="fifth"><svg viewBox="0 0 24 24" fill="currentColor" class="h-8 w-8" aria-hidden="true"><path fill-rule="evenodd" d="M7.5 6a4.5 4.5 0 1 1 9 0 4.5 4.5 0 0 1-9 0ZM3.751 20.105a8.25 8.25 0 0 1 16.498 0 .75.75 0 0 1-.437.695A18.683 18.683 0 0 1 12 22.5c-2.786 0-5.433-.608-7.812-1.7a.75.75 0 0 1-.437-.695Z" clip-rule="evenodd"></path></svg></li>
```
kbond, line 32: Could we change at least one of these to (additionally) prove that the aria behaviour is working?

kbond (changes requested):

src/Icons/tests/Integration/RenderIconsInTwigTest.php
smnandre: Good idea, just added two cases in here!

kbond (approved):

kbond:
Thanks Simon.

PR #1692, opened 2024-04-07
[Doc] Fix Icons doc

JorickPepin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | 
| License       | MIT

This sentence appears twice

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -167,9 +167,6 @@ a subdirectory, the *name* will be ``subdirectory:icon_name``.
 Icons On-Demand
 ~~~~~~~~~~~~~~~
 
-`ux.symfony.com/icons`_ has a huge searchable repository of icons
-from many different sets. This package provides a way to include any icon found on this site _on-demand_.
-
 `ux.symfony.com/icons`_ has a huge searchable repository of icons from many
 different sets. This package provides a way to include any icon found on this
 site *on-demand*:


src/Icons/doc/index.rst
```diff
@@ -167,9 +167,6 @@ a subdirectory, the *name* will be ``subdirectory:icon_name``.
 Icons On-Demand
 ~~~~~~~~~~~~~~~
 
-`ux.symfony.com/icons`_ has a huge searchable repository of icons
```
WebMamba, line 170: @smnandre is it the first one or the second one we should keep? the on-demande is different

smnandre: I think in RST it's  `_`  to emphasis

JorickPepin: the first does not render correctly

https://symfony.com/bundles/ux-icons/current/index.html#icons-on-demand

image

smnandre: Well thne it's `**like that**` 😅

javiereguiluz: Yes, the `_ ... _` is italics in Markdown. In RST, italics is `* ... *` and bold is `** ... **`

But here the problem is that this phrase is wrongly repeated, so let's remove it entirely.

javiereguiluz:
Thank you Jorick.

PR #1709, opened 2024-04-09
[Twig] Handle aria-* attribute boolean values

### LINKED ISSUE #1701: [TwigComponent][DX] Better support for boolean ARIA attributes (i.e. aria-hidden)
In short, a boolean attribute in ARIA is an [enumerated attribute](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/ARIA) in the form of strings "true" and "false". This is different from HTML boolean attributes like `checked`, `open` or `disabled`.

In [`ComponentAttributes`](https://github.com/symfony/ux-twig-component/blob/2.x/src/ComponentAttributes.php#L64-L65) we correctly handle HTML boolean attributes but there is not special "treatment" for ARIA enumerated attributes.

Many developers, especially those just starting out or inexperienced with ARIA (me including), could write the following:

```twig
{# templates/components/Icon.html.twig #}
{% props icon %}

<i
    {{ attributes.defaults({
        class: html_classes("fa-solid", icon),
        "aria-hidden": true, {# Wrong! It should be a string #}
    }) }}
></i>
```

Which would produce and invalid HTML (in the sense of the ARIA enumerated attributes explained above):

```html
<i class="fa-solid fa-user" aria-hidden></i> <!-- Wrong! It should be a string -->
```

To make things easier and more DX, I would suggest to automatically convert true and false to "true" and "false", if the attribute is an ARIA one. Could be easy as:

```php
// Transform ARIA enumerated attribute in case of boolean value
if (str_starts_with($key, 'aria-')) {
  $value = match($value) {
      true => "true",
      false => "false",
      default => $value,
  };
}
```

Or more concise:

```php
if (is_bool($value) && str_starts_with($key, 'aria-')) {
  $value = $value ? "true" : "false";
}
```

... just before returing. WDYT?

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1701 
| License       | MIT

Convert aria-* boolean attributes to their expected string values.

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -51,6 +51,10 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
                     $value = (string) $value;
                 }
 
+                if (\is_bool($value) && str_starts_with($key, 'aria-')) {
+                    $value = $value ? 'true' : 'false';
+                }
+
                 if (!\is_scalar($value) && null !== $value) {
                     throw new \LogicException(sprintf('A "%s" prop was passed when creating the component. No matching "%s" property or mount() argument was found, so we attempted to use this as an HTML attribute. But, the value is not a scalar (it\'s a %s). Did you mean to pass this to your component or is there a typo on its name?', $key, $key, get_debug_type($value)));
                 }
@@ -83,6 +87,10 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
 
         if ($value instanceof \Stringable) {
             $value = (string) $value;
+        }
+
+        if (\is_bool($value) && str_starts_with($attribute, 'aria-')) {
+            $value = $value ? 'true' : 'false';
         }
 
         if (!\is_string($value)) {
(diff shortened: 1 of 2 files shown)


kbond (approved):
@WebMamba, any thoughts/objections?

WebMamba (approved):
Perfect! Sound good to me! Thanks @smnandre

javiereguiluz:
Nice bug fix! Thanks Simon.

norkunas:
This is a BC break for us :D `aria-valuemin`, `aria-valuemax` and `aria-valuenow` are number attributes for a progress component.

Example from component:
`attributes.defaults({ 'aria-valuemin': value is not null ? 0 : false, 'aria-valuemax': value is not null ? 100 : false, 'aria-valuenow': value is not null ? value : false })`

Previously these attributes were hidden and now they just rendered as `aria-valuemin="false"`

smnandre:
@norkunas ~~could you open a new issue ? (i cannot "move" your comment to a new one)~~

I'd like to find a coherent solution for both Icon and TwigComponent, fixing the BC (sorry for that) 


(updated)

smnandre:
Sorry did not see the other comment ;)

PR #1717, opened 2024-04-10
[Icons] aria-hidden="true" doc

smnandre:
Add documentation to explain why & how this works.

(moved the first paragraph down to end with the feature / code examples)


kbond (approved):

kbond:
Thanks Simon.

PR #1719, opened 2024-04-11
 [LiveComponent] PreMount hook extra data loss

### LINKED ISSUE #1704: [TwigComponent] Missleading doc between `Passing & Rendering Attributes` and `PreMount validation`
Rendering ["extra" Attributes](https://github.com/symfony/ux/blob/0f8b952f08862caf7ed76b2b5b93b6b5bfe6227f/src/TwigComponent/doc/index.rst?plain=1#L223-L224) with a [`PreMount` hook + `OptionsResolver`](https://github.com/symfony/ux/blob/0f8b952f08862caf7ed76b2b5b93b6b5bfe6227f/src/TwigComponent/doc/index.rst?plain=1#L453-L454) will not work if followed as the doc explains it.

As from the example, if the Alert component looks like this:
```php
#[AsTwigComponent]
final class Alert
{
    public string $message;

    #[PreMount]
    {
        // validate data
        $resolver->setDefaults(['type' => 'success']);
        $resolver->setAllowedValues('type', ['success', 'danger']);
        $resolver->setRequired('message');
        $resolver->setAllowedTypes('message', 'string');

    }
}
```

Calling it like this:
```twig
<twig:Alert id="alert_id" message="My message"/>
```
 Will throw an `UndefinedOptionsException` error:
> [!CAUTION]
> Error rendering "Alert" component: The option "id" does not exist. Defined options are: "message", "type".

We may need to resolve this incoherence by:
1. Add an extra paragraph in the `PreMount` hook part, to explain how to do it with extra attributes.
2. Re-write the "example" preMount function doc like this:

```diff
#[PreMount]
public function preMount(array $data): array
{
    // validate data
    $resolver = new OptionsResolver();
+   $resolver->setIgnoreUndefined();
+
    $resolver->setDefaults(['type' => 'success']);
    $resolver->setAllowedValues('type', ['success', 'danger']);
    $resolver->setRequired('message');
    $resolver->setAllowedTypes('message', 'string');

-   return $resolver->resolve($data);
+   return array_merge($data, $resolver->resolve($data));
}
```
-------

I could have done a PR directly, but I want to talk about it first and discuss if it is necessary or not.

cavasinf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1704 
| License       | MIT

### DIFF
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -184,7 +184,8 @@ final class ComponentFactory
             $newData = $component->{$method->name}($data);
 
             if (null !== $newData) {
-                $data = $newData;
+                // merge data/options so that extra data is not lost between hooks
+                $data = $newData + $data;
             }
         }
 
--- a/src/TwigComponent/composer.json
+++ b/src/TwigComponent/composer.json
@@ -38,6 +38,7 @@
         "symfony/css-selector": "^5.4|^6.0|^7.0",
         "symfony/dom-crawler": "^5.4|^6.0|^7.0",
         "symfony/framework-bundle": "^5.4|^6.0|^7.0",
+        "symfony/options-resolver": "^6.0|^7.0",
         "symfony/phpunit-bridge": "^6.0|^7.0",
         "symfony/stimulus-bundle": "^2.9.1",
         "symfony/stopwatch": "^5.4|^6.0|^7.0",
--- a/src/TwigComponent/composer.json
+++ b/src/TwigComponent/composer.json
@@ -38,7 +38,7 @@
         "symfony/css-selector": "^5.4|^6.0|^7.0",
         "symfony/dom-crawler": "^5.4|^6.0|^7.0",
         "symfony/framework-bundle": "^5.4|^6.0|^7.0",
-        "symfony/options-resolver": "^6.0|^7.0",
+        "symfony/options-resolver": "^6.4|^7.0",
         "symfony/phpunit-bridge": "^6.0|^7.0",
         "symfony/stimulus-bundle": "^2.9.1",
         "symfony/stopwatch": "^5.4|^6.0|^7.0",
--- a/src/TwigComponent/tests/Fixtures/Component/WithOptionsResolver.php
+++ b/src/TwigComponent/tests/Fixtures/Component/WithOptionsResolver.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Symfony\UX\TwigComponent\Tests\Fixtures\Component;
+
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+use Symfony\UX\TwigComponent\Attribute\PreMount;
+
+/**
+ * @author Florian Cavasin <cavasinf.info@gmail.com>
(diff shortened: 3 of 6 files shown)


src/TwigComponent/src/ComponentFactory.php
```diff
@@ -184,7 +184,8 @@ private function preMount(object $component, array $data, ComponentMetadata $com
             $newData = $component->{$method->name}($data);
 
             if (null !== $newData) {
-                $data = $newData;
+                // merge data/options so that extra data is not lost between hooks
+                $data = $newData + $data;
```
cavasinf, line 188: How do you guys emphasize/warn not to write something for future devs?

We should never simplify `$data = $newData + $data` to `$data += $newData` as it will produce unexpected behavior

kbond: I wonder if we should use `$data = array_merge($data, $newData)`. Maybe it's just me, but this is more explicit than using the `+` operator (as this operator always trips me up).

cavasinf: @Kocal pointed it out for _using less opcodes_
https://github.com/symfony/ux/issues/1704#issuecomment-2046929715

kbond: Heh, I did read that too. Let's keep using + then. Will the test fail if the sides are switched? That's the best way to ensure this doesn't get changed in the future.

cavasinf: FYI:
- `$newData + $data` = `array_merge($data, $newData)`
- `$data += $newData` = `array_merge($newData, $data)`

-------

> Will the test fail if the sides are switched?

Yes, that's why I've added the `PreMount` inside the test. 
And another one: `Symfony\UX\TwigComponent\Tests\Integration\ComponentExtensionTest::testCanRenderTheSameComponentMultipleTimes`

> That's the best way to ensure this doesn't get changed in the future.

Some will think they need to change the tests, no?

kbond: > Yes, that's why I've added the PreMount inside the test.

Ok, perfect!

> Some will think they need to change the tests, no?

I'm not concerned. The note and tests are all we can do.

smnandre, line 188: Hmm i'm not sure here .... as you cannot "transform" data with preMount event there. 

That would be a big BC break :/ 

I would really prefer we only update documentation for those who want to uses partial validation with OptionResolver if we can.

cavasinf: > as you cannot "transform" data with preMount event there

What do you mean by that? Example?

smnandre: Let's say you call a component like that

<twig:Alert level=warning />

In your preMount method, you transform "level" into "class", and so you remove level

If you merge newData and data, will "level" be dropped.. or kept as an attribute ?

cavasinf: Yeah you right, removed data will be back due to the base array concat..
If it is a feature, we maybe need a test to keep that.

cavasinf:
Closing as it breaks the "PreMount remove props" feature.
Only doc should be updated

PR #1720, opened 2024-04-11
[LiveComponent] Add Events listening infos in Profiler and DebugCommand

mbuliard:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | No issue
| License       | MIT

Hello,

First of all, thank you for your work and dedication.

Playing with LiveComponents on my spare time, I have a SPA using several events to communicate between components.
In several occasion, it took me a little time to find out bugs in my application concerning these events.

That's why I propose these small enhancements for debug concerning events in LiveComponent :
1) In TwigComponentDebugCommand, allow to filter components by the events they are listening to :
```bash
bin/console debug:twig-component --listening buyEvent
```

2) Still in TwigComponentDebugCommand, when displaying a unique Component, add a line to display the events it listens to, if any :
Capture d’écran du 2024-04-11 16-22-41

3) In Profiler, add a line to display the events it listens to, if any :
Capture d’écran du 2024-04-11 16-28-02

What do you think of it ? I will add test in a second time.

### DIFF
--- a/src/TwigComponent/src/Command/TwigComponentDebugCommand.php
+++ b/src/TwigComponent/src/Command/TwigComponentDebugCommand.php
@@ -18,9 +18,11 @@ use Symfony\Component\Console\Completion\CompletionSuggestions;
 use Symfony\Component\Console\Helper\TableSeparator;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Finder\Finder;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
 use Symfony\UX\TwigComponent\ComponentFactory;
@@ -49,6 +51,11 @@ class TwigComponentDebugCommand extends Command
         $this
             ->setDefinition([
                 new InputArgument('name', InputArgument::OPTIONAL, 'A component name or part of the component name'),
+                new InputOption(
+                    name: 'listening',
+                    mode: InputOption::VALUE_REQUIRED,
+                    description: 'Filter components list to display only those listening to the given action'
+                ),
             ])
             ->setHelp(
                 <<<'EOF'
@@ -83,7 +90,7 @@ EOF
             return Command::SUCCESS;
         }
 
-        $components = $this->findComponents();
+        $components = $this->findComponents($input->getOption('listening'));
         $this->displayComponentsTable($io, $components);
 
         return Command::SUCCESS;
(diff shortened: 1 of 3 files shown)


src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -314,4 +334,16 @@ private function getAnonymousComponentProperties(ComponentMetadata $metadata): a
 
         return $properties;
     }
+
+    private function resolveEventsListening(string $class): array
```
kbond, line 338: The duplication of this method is unfortunate. Can we use/adjust [this method](https://github.com/symfony/ux/blob/a13900f005c88c6548ebf9daa2ebb79da8340614/src/LiveComponent/src/Attribute/AsLiveComponent.php#L131)?

kbond (changes requested):
This looks useful @mbuliard! Just a suggestion on the implementation and yeah, could you add a test?

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
mbuliard: Thanks for the suggestion ! I was not aware of this method. I removed the @internal to use it, but don't you think it could be a separate service ?

kbond: > I removed the https://github.com/internal to use it

Can you keep `@internal` on the method. We don't want this to be used by end-users.

> but don't you think it could be a separate service ?

Can you explain this a bit more?

mbuliard: > Can you explain this a bit more?

Since the method is use elsewhere and its logic is not inherently linked to the attribute, I would think of creating a separate service let's say `LiveListenersResolver` and call it in the attribute and in my modifications.

WebMamba (commented):
Just tried the PR locally and it's work great! I like the DX! Thanks @mbuliard

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
kbond: Fair point, keeping it internal enables us to make this kind of refactor in the future. For now, let's keep as is.

kbond (commented):
Just realized this breaks the data collectors and debug command if using twig component w/o live component.

Instead of adding a bunch of checks in twig component (we'd like to keep twig component from having knowledge of live component as much as possible), what about making the command and data collector extendable. We'd have extended versions in live component (and swap their services).

_This would be outside the scope of this PR._

@WebMamba, @smnandre, wdyt?

smnandre:
I'm really thinking more and more about "splitting" Twig & Live component to be honest...

So this "extensible system" you suggest @kbond is for me a step in the "good" direction whatever we do next 👍

kbond:
Hey @mbuliard, #1722 should be merged soon. Basically, what's going to be needed here is:
1. Add the new CLI option to `LiveComponentDebugCommand` (this will require some extension points to `TwigComponentDebugCommand`)
2. Add the new profiler logic to `LiveComponentDataCollector` (this will require some extension points to `TwigComponentDataCollector`)
3. Add new profiler info to `live_component.html.twig`

Making changes to Twig Components that are required by Live Components in the same PR is a bit tricky. You need to temporarily add the following to live component's `composer.json` while you're working on it:

```json
"repositories": [
    {
        "type": "path",
        "url": "../TwigComponent"
    }
],
```

Then run composer update for live components.

I apologize in advance for the added complexity, I can definitely jump in and help if you run into problems.

mbuliard:
Hi @kbond , thanks for the update.

I will rework  my PR in the end of the week if yours is merged, otherwise it may be next week.

I hope to sort it out by myself, but will call for help if needed, thanks !

kbond:
> I hope to sort it out by myself, but will call for help if needed, thanks !

I'm hoping to resolve https://github.com/symfony/ux/pull/1722#issuecomment-2069518347 soon in a way that will make this PR easier.

Kocal:

PR #1722, opened 2024-04-11
[Live] override data collector & debug command

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

When live component is installed, swap the debug command and data collector to a live component version.

This refactor enables us to add additional live-component specific details to both the debug command and the profiler.

Helps with #1720.

### DIFF
--- a/src/LiveComponent/templates/Collector/live_component.html.twig
+++ b/src/LiveComponent/templates/Collector/live_component.html.twig
@@ -1,0 +1,1 @@
+{% extends '@TwigComponent/Collector/twig_component.html.twig' %}
--- a/src/TwigComponent/src/Command/TwigComponentDebugCommand.php
+++ b/src/TwigComponent/src/Command/TwigComponentDebugCommand.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\UX\TwigComponent\Command;
 
-use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Completion\CompletionInput;
 use Symfony\Component\Console\Completion\CompletionSuggestions;
@@ -28,7 +27,6 @@ use Symfony\UX\TwigComponent\ComponentMetadata;
 use Symfony\UX\TwigComponent\Twig\PropsNode;
 use Twig\Environment;
 
-#[AsCommand(name: 'debug:twig-component', description: 'Display components and them usages for an application')]
 class TwigComponentDebugCommand extends Command
 {
     private readonly string $anonymousDirectory;
@@ -42,6 +40,16 @@ class TwigComponentDebugCommand extends Command
     ) {
         parent::__construct();
         $this->anonymousDirectory = $anonymousDirectory ?? 'components';
+    }
+
+    public static function getDefaultName(): ?string
+    {
+        return 'debug:twig-component';
+    }
+
+    public static function getDefaultDescription(): ?string
+    {
+        return 'Display components and their usages for an application';
     }
 
     protected function configure(): void
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": ">=8.1",
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
-        "symfony/ux-twig-component": "^2.8",
+        "symfony/ux-twig-component": "^2.17",
         "twig/twig": "^3.8.0"
     },
     "require-dev": {
@@ -38,6 +38,7 @@
         "doctrine/orm": "^2.9.4",
         "doctrine/persistence": "^2.5.2|^3.0",
         "phpdocumentor/reflection-docblock": "5.x-dev",
+        "symfony/console": "^5.4|^6.0|^7.0",
         "symfony/dependency-injection": "^5.4|^6.0|^7.0",
         "symfony/expression-language": "^5.4|^6.0|^7.0",
         "symfony/form": "^5.4|^6.0|^7.0",
--- a/src/LiveComponent/src/Command/LiveComponentDebugCommand.php
+++ b/src/LiveComponent/src/Command/LiveComponentDebugCommand.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Command;
+
+use Symfony\UX\TwigComponent\Command\TwigComponentDebugCommand;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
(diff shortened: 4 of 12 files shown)


src/LiveComponent/templates/Collector/live_component.html.twig
```diff
@@ -0,0 +1,8 @@
+{% extends '@TwigComponent/Collector/twig_component.html.twig' %}
+
+{% block menu %}
+    <span class="label{{ collector.components is empty ? ' disabled' }}">
+        <span class="icon">{{ source('@TwigComponent/Collector/icon.svg') }}</span>
+        <strong>Twigsss Components</strong>
```
WebMamba: typo here I think

WebMamba (commented):
Hey @kbond I really like the idea!

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -44,6 +42,16 @@ public function __construct(
         $this->anonymousDirectory = $anonymousDirectory ?? 'components';
     }
 
+    public static function getDefaultName(): ?string
```
WebMamba, line 45: What do you think about adding an alias debug:live-component so people don't get confuse by typing debug:twig-component when they want to debug a live component ?

src/LiveComponent/templates/Collector/live_component.html.twig
kbond: Haha yes, I was ensuring the proper template was being rendered.

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
kbond: Great idea!

smnandre:
I’ll look at this sunday, but dont wait for me if you need before :)

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -28,7 +27,6 @@
 use Symfony\UX\TwigComponent\Twig\PropsNode;
 use Twig\Environment;
 
-#[AsCommand(name: 'debug:twig-component', description: 'Display components and them usages for an application')]
```
smnandre, line 31: Are most of the commands in the symfony repository using this attribute ?

src/LiveComponent/templates/Collector/live_component.html.twig
smnandre: So the file just need the {% extends %} tag right ? 

Or do we want two distinct menu entries ? (in which case i'll do another icon if you want .. loved working on the twig one ^^)

kbond: > So the file just need the {% extends %} tag right ?

For now, yes - but I'm sort of rethinking this PR (more info below).

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
kbond: They are, yes. I opted to remove to reduce duplication.

kbond:
Upon reflection, I'm thinking this PR isn't the correct direction. When adding more _live component information_ to the debug command and profiler panel, refactoring the parent twig component classes will almost always be required. This sort of defeats the purpose of this change...

What about:
1. Completely separate `debug:live-component` command
    - `debug:twig-component` would still show live components but we could add a "for more information on your live components, run debug:live-component..." message
3. Completely separate live component profiler panel
    - Just like above, the twig component panel would show live components but direct users to see more detail in the live component panel

There will likely be some duplication but perhaps we can mitigate with helper classes or, in the case of the console, maybe the descriptor system (I've never used this so not sure this is viable).

I'm curious for thoughts from others.

smnandre:
In general, concerning Twig & Live Component, i'm 100% for duplication over this "extend" thing we have today. Let met give my reasons before you bite me :) 

* for me TwigComponent should be moved into the TwigBundle or TwigBridge (one day, i'm not saying there is any kind of urgency there)
* for performance reasons, i'd like the event dispatchers to be more separated
* the Attribute extending another make us really dependent, and that -for the moment- forbid us to have dedicated features for TwigComponent (not ported to LiveComponent i mean)... but some of them could make sense.
* the "import from bundle" rules clearly have different implication, as TwigComponent "do nothing" until called, wereas LiveComponent expose code to the external world. 

So i'm all for your points. We just need to think about the "how" :)

kbond:
Your points all make sense to me and solidify my idea to split the commands/panels. @WebMamba, wdyt?

WebMamba:
> When adding more live component information to the debug command and profiler panel, refactoring the parent twig component classes will almost always be required.

Do you have examples in mind?

I am not sure about that since LiveComponent is TwigComponent, I don't see why we should expose different information. More information but not different.
But tell me if you have concrete examples.

kbond:
https://github.com/symfony/ux/pull/1720 is the example that prompted this PR. I can imagine having details on actions and query parameters in the future.

Admittedly, I'm on the fence on what to do here. Today, I feel more inclined to proceed with this PR as is 😆

smnandre:
There is nothing preventing us to change later i guess, so i'll support any choice you guys make :)

WebMamba:
As long as we don't have a good reason, let's leave it as it is 😁

kbond:
Ok, I'll finalize and merge this PR this week so we can unblock #1720.

kbond:
I think this is ready.

WebMamba (approved):
Currently if you run bin/console debug:live-component you also got simple TwigComponent and anonymous component. Should we filter the list to only display Live component ? If we do that should add filters to the debug:twig-component command (bin/console debug:twig-componet --anonymous, bin/console debug:twig-componet --twig-component, bin/console debug:twig-componet --live)?

kbond:
> Currently if you run bin/console debug:live-component you also got simple TwigComponent and anonymous component. Should we filter the list to only display Live component ? If we do that should add filters to the debug:twig-component command (bin/console debug:twig-componet --anonymous, bin/console debug:twig-componet --twig-component, bin/console debug:twig-componet --live)?

Hah, now I'm back on the fence on what to do here. As I mentioned in #1720, it's going to be pretty cumbersome to add live-specific features to the command/profiler panel. I'm back wondering if there should be two completely separate commands and profiler panels...

smnandre:
>  I'm back wondering if there should be two completely separate commands and profiler panels...

Still my opinion there

smnandre:
I'm now persuaded having distinct commands is the better solution.

LiveComponent does not deal with anonymous templates, TwigComponent do not deal with queryurl, events, etc...

How can i help you on this ?

WebMamba (approved):

WebMamba (changes requested):
Wait few things before merging 

- [ ] the debug:live-component does not appear on the command list when you do a bin/console since debug:live-component is just an alias
- [ ] the debug:live-component should filter and should show only live component
- [ ] the panel live component is name twig component and have the same icon
- [ ] the panel live component should only display live component 

WDYT?

kbond:
Closing in favour of #1812

PR #1723, opened 2024-04-12
[Autocomplete] Use `choice_value` in the `WrappedEntityTypeAutocompleter` in `EntityAutocompleteField`

jakubtobiasz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | 
| License       | MIT

Currently despite setting the `choice_value` the `\Symfony\UX\Autocomplete\Form\WrappedEntityTypeAutocompleter` uses the id value of the entity. This PR allows to override the autocomplete result value in the entity autocomplete fields.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.17.0
+
+-   Allow `choice_value` option in entity autocomplete fields #1723
+
 ## 2.16.0
 
 -   Missing translations added for many languages #1527 #1528 #1535
--- a/src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
+++ b/src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
@@ -100,6 +100,16 @@ final class WrappedEntityTypeAutocompleter implements OptionsAwareEntityAutocomp
 
     public function getValue(object $entity): string
     {
+        $choiceValue = $this->getFormOption('choice_value');
+
+        if (\is_string($choiceValue) || $choiceValue instanceof PropertyPathInterface) {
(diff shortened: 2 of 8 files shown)


kbond (commented):
@jakubtobiasz, thanks for this PR! Do you think this could be considered a bug fix?

jakubtobiasz:
Hi @kbond,
considering the definition from the Wikipedia

> A software bug is an error, flaw or [fault](https://en.wikipedia.org/wiki/Fault_(technology)) in the design, development, or operation of [computer software](https://en.wikipedia.org/wiki/Computer_software) that causes it to produce an incorrect or unexpected result, or to behave in unintended ways.

The 
> that causes it to produce an (...) unexpected result

part convinces me that might be categorized as a bugfix. We were pretty surprised it doesn't work that way :D.

---

I've updated the description, the question is whether there should be a `2.16.1` version or we should expect `2.17` as a next version.

kbond:
> We were pretty surprised it doesn't work that way :D.

Me too!

Anyway, it's just semantics and I really just wanted to clarify that this "felt like the intended use" which you have done.

The next version will be 2.17.0 so no worries about the changelog.

src/Autocomplete/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.17.0
+
+-   Allow defining custom values for the autocomplete results with using `choice_value` option in entity autocomplete fields
```
smnandre:
```suggestion
-   Allow `choice_value` option in entity autocomplete fields #1723
```

(just a suggestion)

smnandre (approved):
Thank you @jakubtobiasz !

kbond:
Thank you Jacob.

norkunas:
this was a duplicate of #1328

PR #1724, opened 2024-04-12
[Icons] Fix Icon aliases not found

### LINKED ISSUE #1721: [Icons] Icons with 2 (alias) names are not found
I have searched via https://ux.symfony.com/icons?query=x&set=flowbite and found the "X" Icon.

According to the site it should be named:  `flowbite:x-outline`

`<twig:ux:icon name="flowbite:x-outline" />`

If it is used in twig it will throw an error that the icon is not found.

The reason behind it is that it expects the Iconify response to include that name in the array:

Iconify.php L64
```PHP
if (!isset($data['icons'][$name]['body']))
```

In my case it finds the right icon but the name included in the array is `close-outline` and not `x-outline`.

So it complains: `Error rendering "UX:Icon" component: Icon "flowbite:x-outline" not found.`

﻿﻿﻿regards Alex

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1721
| License       | MIT

I'm adding a dedicated functional test because aliases will require specific attention/code regarding automatic class generation.

Update: other CS fails fixed in https://github.com/symfony/ux/pull/1725

### DIFF
--- a/src/Icons/src/Iconify.php
+++ b/src/Icons/src/Iconify.php
@@ -61,14 +61,19 @@ final class Iconify
             throw new IconNotFoundException(sprintf('The icon "%s:%s" does not exist on iconify.design.', $prefix, $name));
         }
 
+        $nameArg = $name;
+        if (isset($data['aliases'][$name])) {
+            $name = $data['aliases'][$name]['parent'];
+        }
+
         if (!isset($data['icons'][$name]['body'])) {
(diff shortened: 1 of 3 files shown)


kbond (approved):

kbond:
Thanks for the fix Simon. And thanks for reporting @alexz707!

PR #1727, opened 2024-04-13
[Site] Fix LiveMemory components

smnandre:
Fix banner on livememory page

### DIFF
--- a/ux.symfony.com/assets/styles/demos/live-memory.min.css
+++ b/ux.symfony.com/assets/styles/demos/live-memory.min.css
@@ -1,10 +1,15 @@
 /** -----------------------------------------------------------------
  * LiveMemory demo
+ *
+ * This file is suffixed with `.min.css` to prevent it from being
+ * processed by the CDN compression system, which would break the
+ * most recent CSS syntax (as layers) used in this file.
  * --------------------------------------------------------------- */
 
 @import "live-memory/app/reset";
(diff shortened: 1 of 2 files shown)


ux.symfony.com/assets/styles/demos/live-memory.min.css
```diff
@@ -5,6 +5,7 @@
 @import "live-memory/app/reset";
 @import "live-memory/app/root";
 
+@import "../components/Banner";
```
kbond, line 12: Not exactly in the scope of this PR, but why is this file suffixed with `.min.css`? My first thought was this is a built asset that shouldn't be commited.

smnandre: It's due to the CDN that mess with CSS files (using a compression that does not understand some new CSS syntaxes).. the only way we found to avoid that was to add the ".min." part in the file name...

kbond: Would you mind throwing a comment at the top of the file explaining this. Future me will thank you!

smnandre: Absolutely, should have done it before ! 

Done

kbond (approved):

kbond:
Thanks Simon.

PR #1728, opened 2024-04-13
[Site] Lighten bootstrap

smnandre:
Remove unused components

To be honest, i have something a bit more "radical" in mind ... but let's go there slowly :)


ux.symfony.com/assets/styles/app.scss
```diff
@@ -26,16 +26,16 @@
 @import "../../vendor/twbs/bootstrap/scss/tables";
 @import "../../vendor/twbs/bootstrap/scss/forms";
 @import "../../vendor/twbs/bootstrap/scss/buttons";
-@import "../../vendor/twbs/bootstrap/scss/transitions";
-@import "../../vendor/twbs/bootstrap/scss/dropdown";
- @import "../../vendor/twbs/bootstrap/scss/button-group";
+// @import "../../vendor/twbs/bootstrap/scss/transitions";
```
kbond, line 29: Why commented instead of removed?

smnandre: To keep track of what's in the standard distribution VS what we "removed" 

It's scss so it won't appear in compiled code, but i can clean this if you want, i know what the next step is ... and it won't require this file   👼

smnandre: (but don't know _when_)

kbond: Oh I see, understood!

kbond:
Thanks Simon.

PR #1740, opened 2024-04-16
[Site] upgrade to PHP 8.3

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

Upgrades the site to use PHP 8.3

PR #1742, opened 2024-04-16
[RFC][Site] replace `highlight.js` with `tempest/highlight`

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

_(depends on #1740)_

I couldn't remove `highlight.js` entirely as `translation-demo-controller.js` still requires it and I don't think we can easily refactor it out. I don't really want to maintain two syntax highlighters (+styling) so I'm not sure this PR is worth it...


ux.symfony.com/templates/components/CodeLine.html.twig
```diff
@@ -13,7 +13,7 @@
     <pre
         {% if height %} style="--height: {{ height }}" {% endif %}
         data-clipboarder-target="source"
-    ><code {% if highlight %} data-code-highlighter-target="codeBlock" {% endif %}>{{ __code ? __code|trim|raw : code }}</code></pre>
+    ><code>{{ __code ? __code|trim|raw : code }}</code></pre>
```
kbond, line 16: I broke the highlighting here so this needs to be fixed. Does highlight.js detect the language?

.github/workflows/ux.symfony.com.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
 
             - uses: shivammathur/setup-php@v2
               with:
-                php-version: 8.1
+                php-version: 8.3
```
smnandre, line 24: Hahaha ... hidden in a PR :))

smnandre:
> I don't think we can easily refactor it out

It's not that hard i think. Let's add this to my ever-growing-todo-list 😅

Kocal:
That's a nice improvement, especially for [web perfs](https://twitter.com/HugoAlliaume/status/1779192127169171754) :eyes:

For `translator-demo-controller.js`, maybe for a next iteration, we can try to highlight code blocks with `tempest/highlight` and update them in front-side. Or maybe use LiveComponents, but it can be overkill

WebMamba (commented):
Humm sorry but I think if do this move we should completly remove highligh.js even in translator-demo-controller 😅

kbond:
> Humm sorry but I think if do this move we should completly remove highligh.js even in translator-demo-controller

I agree, closing.

ux.symfony.com/src/Service/CommonMark/TempestHighlightExtension.php
```diff
@@ -0,0 +1,31 @@
...
+use League\CommonMark\Environment\EnvironmentBuilderInterface;
+use League\CommonMark\Extension\CommonMark\Node\Block\FencedCode;
+use League\CommonMark\Extension\CommonMark\Node\Inline\Code;
+use League\CommonMark\Extension\ExtensionInterface;
+use Tempest\Highlight\CommonMark\CodeBlockRenderer;
+use Tempest\Highlight\CommonMark\InlineCodeBlockRenderer;
+use Tempest\Highlight\Highlighter;
+
+/**
+ * TODO: remove this class when/if https://github.com/tempestphp/highlight/pull/102 is merged.
```
smnandre, line 14: Good: https://github.com/tempestphp/highlight/pull/102

PR #1753, opened 2024-04-18
[Stimulus] Allow Twig 3.9

smnandre:
As long Twig or Live component restrict Twig (until we implement yield ready) no need to restrict it here...

... and that would ease IRL tests for Twig&Live :)


kbond:
Thanks Simon.

PR #1755, opened 2024-04-18
[LazyImage] Cache BlurHash, close #2

### LINKED ISSUE #2: LazyImage: cache blurhash
The docs say:

> You should try to generate small BlurHash images as generating the image can be CPU-intensive.

So it would be nice if the hashes could be cached somehow :)

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #2 
| License       | MIT

Hi everyone, this PR is a proposal for #2, the second oldest issue on this repository 🤯 

I've also though about decorates the `BlurHashInterface` if option `cache` is set, this way we don't modify the current `BlurHash`. Let me know what is preferred :)

### DIFF
--- a/src/LazyImage/tests/BlurHash/BlurHashTest.php
+++ b/src/LazyImage/tests/BlurHash/BlurHashTest.php
@@ -12,7 +12,11 @@
 namespace Symfony\UX\LazyImage\Tests;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
+use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\UX\LazyImage\BlurHash\BlurHashInterface;
+use Symfony\UX\LazyImage\BlurHash\CachedBlurHash;
 use Symfony\UX\LazyImage\Tests\Kernel\TwigAppKernel;
 
 /**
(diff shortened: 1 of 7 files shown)


src/LazyImage/tests/BlurHash/BlurHashTest.php
```diff
@@ -37,6 +40,54 @@ public function testEncode()
...
+        static::assertEmpty($cache->getValues());
+
+        $this->assertSame(
+            'L54ec*~q_3?bofoffQWB9F9FD%IU',
+            $blurHash->encode(__DIR__.'/../Fixtures/logo.png')
+        );
+
+        static::assertSame(
+            'L54ec*~q_3?bofoffQWB9F9FD%IU',
+            $cache->getItem(array_key_first($cache->getValues()))->get()
```
Kocal: The generated cache key was different between my machine and the CI, so I'm using `array_key_first($cache->getValues())` instead 🤷🏻

kbond (commented):
This is nice! I think I'd prefer as a decorator.

src/LazyImage/tests/BlurHash/BlurHashTest.php
norkunas: Maybe would be enough to set `framework.cache.prefix_seed` to get stable namespaces/keys :)

Kocal: To be more clear:
- on my machine the cache key is `blurhash.70d3aaba5c301af7`
- but on the CI it is `blurhash.788394be043a8824` (see https://github.com/symfony/ux/actions/runs/8745742961/job/24001335802?pr=1755#step:7:108 & https://github.com/symfony/ux/actions/runs/8745742961/job/24001338516?pr=1755#step:7:73)

The different part is the result of `hash('xxh3', '...')` and is not related to the cache prefix seed 😕  

And today we've learned, the namespace (generated through `framework.cache.prefix`) is not even used when adapter is an `ArrayAdapter` (https://github.com/symfony/cache/blob/7.0/DependencyInjection/CachePoolPass.php#L118-L119) :D

Kocal:
Rewritten as a decorator, tests have been adapted too

src/LazyImage/tests/BlurHash/BlurHashTest.php
norkunas: Ah 😲

Kocal: Well, in fact it's super simple, since we compute the cache key with the filename, here `__DIR__.'/../Fixtures/logo.png'`, paths are different between my machine and CI :) 

But I we should keep this behaviour as is, I don't think we want to manipulate the filename (from public directory or not, from a URL or not, ...).

kbond (approved):

kbond:
Thanks Hugo, only 270 issues to go!

Kocal:
I just realized that the decorator `CachedBlurHash` is useless when you call `data_uri_thumbnail` (which seems to be the primary usage of this package):
- `CachedBlurHash#createDataUriThumbnail` calls `BlurHash#createDataUriThumbnail`
- `BlurHash#createDataUriThumbnail` calls `BlurHash#encode`, not `CachedBlurHash#encode`
- The cache is only used when directly calling `CachedBlurHash#encode` 

I'm working on it

PR #1759, opened 2024-04-20
[LiveComponent] hide property-info deprecations in Symfony 7.1

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

This fixes the LiveComponent CI on 7.1.

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -43,7 +43,7 @@
         "symfony/form": "^5.4|^6.0|^7.0",
         "symfony/framework-bundle": "^5.4|^6.0|^7.0",
         "symfony/options-resolver": "^5.4|^6.0|^7.0",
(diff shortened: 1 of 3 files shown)


kbond:
Admittedly, this was mostly fixed via trial and error as I'm not too familiar with the new component. The tests pass!

WebMamba (commented):
@kbond this PR make me think that we should maybe review the SymfonyUX release process. I think to fully use this PropertyInfo component we need to do more refactoring (mostly in the LiveComponentHydrator I think). And this is adding a lot if statement to use one component or an other depending of the user Symfony version, something I feel a bit wobbly... What do you think about releasing one major version each time Symfony release one new major version. So at each major version we can fully use the last Symfony major?

kbond:
I've opted to ignore these deprecations until a time I better understand the `symfony/type-info` component.

PR #1760, opened 2024-04-20
[Turbo] fix sca

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

Not sure how to fix the failing test.

### DIFF
--- a/src/Turbo/phpstan.neon.dist
+++ b/src/Turbo/phpstan.neon.dist
@@ -35,3 +35,13 @@ parameters:
 			message: "#^Parameter \\$stimulus of method Symfony\\\\UX\\\\Turbo\\\\Bridge\\\\Mercure\\\\TurboStreamListenRenderer\\:\\:__construct\\(\\) has invalid type Symfony\\\\WebpackEncoreBundle\\\\Twig\\\\StimulusTwigExtension\\.$#"
 			count: 1
 			path: src/Bridge/Mercure/TurboStreamListenRenderer.php
+
+		-
+			message: "#^Call to an undefined method Doctrine\\\\ORM\\\\Event\\\\OnFlushEventArgs\\:\\:getEntityManager\\(\\)\\.$#"
+			count: 1
+			path: src/Doctrine/BroadcastListener.php
+
+		-
+			message: "#^Call to an undefined method Doctrine\\\\ORM\\\\Event\\\\PostFlushEventArgs\\:\\:getEntityManager\\(\\)\\.$#"
+			count: 1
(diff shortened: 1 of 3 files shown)


src/Turbo/phpstan.neon.dist
```diff
@@ -35,3 +35,13 @@ parameters:
 			message: "#^Parameter \\$stimulus of method Symfony\\\\UX\\\\Turbo\\\\Bridge\\\\Mercure\\\\TurboStreamListenRenderer\\:\\:__construct\\(\\) has invalid type Symfony\\\\WebpackEncoreBundle\\\\Twig\\\\StimulusTwigExtension\\.$#"
 			count: 1
 			path: src/Bridge/Mercure/TurboStreamListenRenderer.php
+
+		-
+			message: "#^Call to an undefined method Doctrine\\\\ORM\\\\Event\\\\OnFlushEventArgs\\:\\:getEntityManager\\(\\)\\.$#"
```
kbond, line 40: I thought phpstan understood `method_exists()` 🤷🏻

src/Turbo/src/Doctrine/BroadcastListener.php
```diff
@@ -136,7 +136,7 @@ private function storeEntitiesToPublish(EntityManagerInterface $em, object $enti
                 foreach ($options as $option) {
                     $this->broadcastedClasses[$class][] = $option->newInstance()->options;
                 }
-            } elseif ($this->annotationReader && $options = $this->annotationReader->getClassAnnotations($r ?? new \ReflectionClass($class))) {
+            } elseif ($this->annotationReader && $options = $this->annotationReader->getClassAnnotations(new \ReflectionClass($class))) {
```
dunglas: Why this change?

kbond: `$r` is never null here, it's always set in the above if statement.

dunglas: So we should only keep `$r`, isn't it?

kbond: Heh, yes, you're right

smnandre:
> Not sure how to fix the failing test.

Related to the deprecation / session, no ?

kbond:
> Related to the deprecation / session, no ?

I don't think so, it's a timeout error in `Symfony\UX\Turbo\Tests\BroadcastTest::testBroadcastWithProxy`

PR #1761, opened 2024-04-20
[LazyImage] Fix cache implementation

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Following https://github.com/symfony/ux/pull/1755#issuecomment-2067586119, `BlurHash#createDataUriThumbnail` could not use the cache because it call `$this->encode()`, which refers to `BlurHash#encode` and not `CachedBlurHash#encode`.

So I went back with my previous implementation by injecting (or not) a `CacheAdapter` implementation to `BlurHash`, and use it if necessary. 

I've cheated a bit to adapt test, because I really want to ensure `$cache` is used or not.


kbond (approved):

kbond:
Good catch, thanks Hugo.

PR #1766, opened 2024-04-22
[LazyImage] refactor to `twig.runtime` & support `intervention/image` 3

### LINKED ISSUE #1465: [Lazy-Image] Support for intervention/image v3
Hi there ! 

While inspecting my dependencies, I just noticed that `intervention/image` had a new v3 version. 

My project use `ux-lazy-image`. I was able to upgrade intervention/image to the last version, and few problems appears. 

On LazyImageExtension, the instance of ImageManager need a driver now (Driver class or an instance of driver). Could be interesting to check if we have Imagick or GD available at this moment to pass with correct driver. However, if we need to support both v2 and v3, i think we should add more adjustment: 

```php

class LazyLoadExtension extends Extension implements PrependExtensionInterface
{
    public function load(array $configs, ContainerBuilder $container)
    {
        if (class_exists(ImageManager::class)) {
            $definition = new Definition(ImageManager::class);
            if (class_exists(\Imagick::class)) {
                $definition->addArgument(\Intervention\Image\Drivers\Imagick\Driver::class);
            } else if (class_exists(\GdImage::class)) {
                $definition->addArgument(\Intervention\Image\Drivers\Gd\Driver::class);
            }
            $container
                ->setDefinition('lazy_image.image_manager', $definition)
                ->setPublic(false)
            ;
        }
        // ...

```

Then, in the BlurHash service, there is few changes in the API provided by the imageManager. 

I listed following : 
* make() -> read()
* pixel() -> drawPixel() (and x,y in first args)
* canvas() -> create()
* pickColor return now a ColorInterface object, that can be converted in array with $color->toArray()
* getWidth() and getHeight() -> width() and height() on Image object

After made all this changes, I ended up with a "Unable to decode input" exception :see_no_evil: 

## Possible solutions

As a quickfix and to avoid the issue for new user that directly install the latest version of intervention/image, we can add a conflic to that v3.

If we don't want to support both version, we can work on the BlurHash service to migrate with the new API (https://image.intervention.io/v3/introduction/upgrade)

If we wan't keep support of that version, and to avoid lot of "method_exist" in the BlurHash service, we maybe should add a InterventionV3BlurHash service and inject it in BlurHash interface when we detect such a version (don't know if it's possible to guess version in the bundle extension by the way). 

WDYT ? 

I can handle modifications after your feedback.

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1465
| License       | MIT

- Refactored the code to use `twig.runtime`
- Support `intervention/image` 3


src/LazyImage/src/Twig/BlurHashExtension.php
```diff
@@ -22,28 +22,11 @@
...
-    public function __construct(BlurHashInterface $blurHash)
-    {
-        $this->blurHash = $blurHash;
-    }
-
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('data_uri_thumbnail', [$this, 'createDataUriThumbnail']),
-            new TwigFunction('blur_hash', [$this, 'blurHash']),
```
Kocal: Why those changes?

kbond: Using a twig runtime increase performance slightly (doesn't need to load the BlurHash service on every request).

Kocal: Oh, I didn't see you add tag `twig.runtime` to `lazy_image.blur_hash`, so yeah that's logical. 

I'm usually familiar with a dedicated `<TwigExtension>Runtime`, implementing `Twig\Extension\RuntimeExtensionInterface`.

Doing things like you did does not break decoration from userland? or any custom implementation?

kbond: > I'm usually familiar with a dedicated <TwigExtension>Runtime, implementing Twig\Extension\RuntimeExtensionInterface.

Yeah, that's just a marker interface for autoconfiguration now.

> Doing things like you did does not break decoration from userland? or any custom implementation?

Hmm, yes, I think this will be an issue if a user decides to decorate `BlurHash` as I don't believe tags are carried over to the decorated service... My gut says this isn't a realistic scenario but what do you think?

Kocal: Who knows... Before #1755 I had planned to write my own implementation for caching :D 

Well, to me that would be nice to have the possibility to decorate the default `BlurHashInterface`, for example add logging, maybe use a fast implementation in tests (to speed up them), ...

kbond: Okay fair enough, I'll revert this change.

Kocal: Thanks :)

Kocal: Thanks :)

PR #1771, opened 2024-04-23
[LazyImage] Re-add forgotten twig.runtime

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Hi!

It looks like the Twig Runtime has been removed in #1766 while processing my review, so let's add it back :)

### DIFF
--- a/src/LazyImage/src/DependencyInjection/LazyImageExtension.php
+++ b/src/LazyImage/src/DependencyInjection/LazyImageExtension.php
@@ -23,6 +23,7 @@ use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\UX\LazyImage\BlurHash\BlurHash;
 use Symfony\UX\LazyImage\BlurHash\BlurHashInterface;
 use Symfony\UX\LazyImage\Twig\BlurHashExtension;
+use Symfony\UX\LazyImage\Twig\BlurHashRuntime;
 
 /**
  * @author Titouan Galopin <galopintitouan@gmail.com>
@@ -40,7 +41,6 @@ class LazyImageExtension extends Extension implements PrependExtensionInterface
             $container
                 ->setDefinition('lazy_image.image_manager', new Definition(ImageManager::class))
                 ->addArgument(BlurHash::intervention3() ? Driver::class : [])
-                ->setPublic(false)
             ;
         }
 
@@ -60,9 +60,13 @@ class LazyImageExtension extends Extension implements PrependExtensionInterface
 
         $container
             ->setDefinition('twig.extension.blur_hash', new Definition(BlurHashExtension::class))
-            ->addArgument(new Reference('lazy_image.blur_hash'))
             ->addTag('twig.extension')
-            ->setPublic(false)
+        ;
+
+        $container
+            ->setDefinition('twig.runtime.blur_hash', new Definition(BlurHashRuntime::class))
+            ->addArgument(new Reference('lazy_image.blur_hash'))
+            ->addTag('twig.runtime')
         ;
     }
 
--- a/src/LazyImage/src/Twig/BlurHashExtension.php
+++ b/src/LazyImage/src/Twig/BlurHashExtension.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\UX\LazyImage\Twig;
 
-use Symfony\UX\LazyImage\BlurHash\BlurHashInterface;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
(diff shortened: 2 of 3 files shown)


src/LazyImage/src/Twig/BlurHashExtension.php
```diff
@@ -22,28 +21,9 @@
...
-        ];
-    }
-
-    public function createDataUriThumbnail(string $filename, int $width, int $height, int $encodingWidth = 75, int $encodingHeight = 75): string
-    {
-        return $this->blurHash->createDataUriThumbnail($filename, $width, $height, $encodingWidth, $encodingHeight);
-    }
-
-    public function blurHash(string $filename, int $encodingWidth = 75, int $encodingHeight = 75): string
+    public function getFunctions(): iterable
```
smnandre, line 24:
```suggestion
    /**
    * @return TwigFunction[]
    */ 
    public function getFunctions(): array
```

src/LazyImage/src/Twig/BlurHashExtension.php
```diff
@@ -22,28 +21,9 @@
...
     {
-        return $this->blurHash->encode($filename, $encodingWidth, $encodingHeight);
+        yield new TwigFunction('data_uri_thumbnail', [BlurHashRuntime::class, 'createDataUriThumbnail']);
+        yield new TwigFunction('blur_hash', [BlurHashRuntime::class, 'blurHash']);
```
smnandre, line 27:
```suggestion
       return [
           new TwigFunction('data_uri_thumbnail', [BlurHashRuntime::class, 'createDataUriThumbnail']),
            new TwigFunction('blur_hash', [BlurHashRuntime::class, 'blurHash']),
        ];
```

src/LazyImage/src/DependencyInjection/LazyImageExtension.php
```diff
@@ -60,10 +61,16 @@ public function load(array $configs, ContainerBuilder $container)
...
-            ->addArgument(new Reference('lazy_image.blur_hash'))
             ->addTag('twig.extension')
             ->setPublic(false)
         ;
+
+        $container
+            ->setDefinition('twig.runtime.blur_hash', new Definition(BlurHashRuntime::class))
+            ->addArgument(new Reference('lazy_image.blur_hash'))
+            ->addTag('twig.runtime')
+            ->setPublic(false)
```
smnandre:
```suggestion
```

Services are private per default

src/LazyImage/src/Twig/BlurHashExtension.php
Kocal: `iterable` is a valid return type here, by doing this way we don't create an useless array in memory, and Twig [only iterates on `$extension->getFunctions()`](https://github.com/twigphp/Twig/blob/4a7de4ab5603ad8f199222cae3951fc271a8c45d/src/ExtensionSet.php#L443-L445), so that's perfectly fine.

We even use `iterable` + `yield` in the repo:
- https://github.com/symfony/ux/blob/80b36bcbeb2902a55495db2afd544f738c73148f/src/Notify/src/Twig/NotifyExtension.php#L25-L28
- https://github.com/symfony/ux/blob/80b36bcbeb2902a55495db2afd544f738c73148f/src/Turbo/src/Twig/TwigExtension.php#L33-L36

Also, the PHPDoc is useless since `@return TwigFunction[]` is already defined here https://github.com/twigphp/Twig/blob/4a7de4ab5603ad8f199222cae3951fc271a8c45d/src/Extension/ExtensionInterface.php#L58-L63

src/LazyImage/src/DependencyInjection/LazyImageExtension.php
Kocal: Will remove it on the 2 other services, thanks

src/LazyImage/src/Twig/BlurHashExtension.php
smnandre: I'm pretty sure the performance impact is null :) 

But okay to let iterable :) 

Now, just to give my reason to suggest this, i do think array would be more accurate, as the doc of the interface states 

> Returns a list of functions to add to the existing list.

;)

src/LazyImage/src/DependencyInjection/LazyImageExtension.php
smnandre: Thanks!

src/LazyImage/src/Twig/BlurHashExtension.php
smnandre: (you can resolve this one too ;) )

kbond (approved):

kbond:
Good catch, thanks Hugo.

PR #1774, opened 2024-04-24
[Turbo] add options to EventSource Mercure

### LINKED ISSUE #291: [UX Turbo Mercure] EventSource is missing withCredentials
Hi, isn't `{withCredentials: true}` missing in https://github.com/symfony/ux/blob/09fca7dab0cd0e3a4db4ae4884b977f8d14bea00/src/Turbo/Bridge/Mercure/Resources/assets/src/turbo_stream_controller.ts#L36 ?
WIthout it the mercureAuthorization cookie will not be sent to mercure hub and it returns HTTP 401 (unless configured anonymous).

ytilotti:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | n/a
| Issues        | Fix #291 
| License       | MIT

To prevent 401 with Mercure & Turbo stream on different domain, add an options for EventSource is required.
Exemple of code:
```twig
<div id="messages" {{ turbo_stream_listen('https://example.com/books/1', eventSourceOptions = {withCredentials: true}) }}</div>
```

### DIFF
--- a/src/Turbo/assets/src/turbo_stream_controller.ts
+++ b/src/Turbo/assets/src/turbo_stream_controller.ts
@@ -17,12 +17,14 @@ export default class extends Controller {
     static values = {
         topic: String,
         hub: String,
+        eventSourceOptions: Object,
     };
     es: EventSource | undefined;
     url: string | undefined;
 
     declare readonly topicValue: string;
     declare readonly hubValue: string;
+    declare readonly eventSourceOptionsValue: object;
     declare readonly hasHubValue: boolean;
     declare readonly hasTopicValue: boolean;
 
@@ -40,7 +42,7 @@ export default class extends Controller {
 
     connect() {
         if (this.url) {
-            this.es = new EventSource(this.url);
+            this.es = new EventSource(this.url, this.eventSourceOptionsValue);
             connectStreamSource(this.es);
         }
     }
--- a/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
+++ b/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
@@ -21,7 +21,7 @@ use Twig\Environment;
 interface TurboStreamListenRendererInterface
 {
     /**
-     * @param string|object $topic
+     * @param string|object       $topic
      * @param array<string,mixed> $eventSourceOptions
      */
     public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions = []): string;
--- a/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
+++ b/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
@@ -23,5 +23,5 @@ interface TurboStreamListenRendererInterface
     /**
      * @param string|object $topic
      */
-    public function renderTurboStreamListen(Environment $env, $topic): string;
+    public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions): string;
 }
--- a/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
+++ b/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
@@ -22,6 +22,7 @@ interface TurboStreamListenRendererInterface
 {
     /**
      * @param string|object $topic
+     * @param array<string,mixed> $eventSourceOptions
      */
-    public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions): string;
+    public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions = []): string;
 }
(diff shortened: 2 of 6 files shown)


src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
```diff
@@ -23,5 +23,5 @@ interface TurboStreamListenRendererInterface
     /**
      * @param string|object $topic
      */
-    public function renderTurboStreamListen(Environment $env, $topic): string;
+    public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions): string;
```
smnandre: This would be a BC break.. :/

src/Turbo/assets/src/turbo_stream_controller.ts
```diff
@@ -40,7 +42,7 @@ export default class extends Controller {
 
     connect() {
         if (this.url) {
-            this.es = new EventSource(this.url);
+            this.es = new EventSource(this.url, this.eventSourceOptionsValue);
```
smnandre, line 45: What would happen if evenSourceOptionsValue is not set ?

ytilotti: Object empty by default via stimulus.

src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
ytilotti: Right, need to add a default value. Thanks.

```php
public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions = []): string;
```

tibobaldwin:
But I think the `mercureAuthorization` cookie is not sent. Is it normal ?
It's the Twig Mercure extension that creates the cookie : https://github.com/symfony/mercure/blob/main/src/Twig/MercureExtension.php#L42.

ytilotti:
You have right @tibobaldwin for the first call, but after the Turbo takes over.
if you are using private updates, Turbo needs to add withCredentials.

tibobaldwin:
But then, how do I add the `mercure()` Twig function?

seb-jean:
Great, this PR is very interesting :).
But in fact, how to add this Twig helper which was added in https://github.com/symfony/mercure/pull/62?

ytilotti:
Sorry @tibobaldwin & @seb-jean, but I think it's not the place for these questions. Check the issue or the [documentation](https://symfony.com/doc/current/mercure.html#subscribing)?

But an example:
```twig
{% block javascripts %}
    <script>
        const eventSource = new EventSource("{{ mercure('https://example.com/books/1', { subscribe: ['https://example.com/books/1'] })|escape('js') }}", {
            withCredentials: true
        });
    </script>
{% endblock %}

{% block content %}
<div id="messages" {{ turbo_stream_listen('https://example.com/books/1', eventSourceOptions = {withCredentials: true}) }}
    Content
</div>
{% endblock %}
```

Without this PR, you need to create your own turbo stream controller and call like this:
```twig
...

{% block content %}
<div id="messages" data-controller="turbo-stream" data-turbo-stream-topic-value="https://example.com/books/1" data-turbo-stream-hub-value="{{ mercure() }}"
    Content
</div>
{% endblock %}
```

seb-jean:
Thanks. This is mainly to set the cookie in fact.

gremo:
Why this isn't merged yet? Seems totally fine to me, and very helpful.

src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
```diff
@@ -21,7 +21,8 @@
 interface TurboStreamListenRendererInterface
 {
     /**
-     * @param string|object $topic
+     * @param string|object       $topic
+     * @param array<string,mixed> $eventSourceOptions
      */
-    public function renderTurboStreamListen(Environment $env, $topic): string;
+    public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions = []): string;
```
smnandre, line 27: You cannot add parameter on an interface, even with default values, without breaking existing implementation.


```php
interface FooInterface
{
    public function bar(string $a, array $eventSourceOptions = []): string;
}

class AppFoo implements FooInterface
{
    // existing method matching previous interface
    public function bar(string $a): string
    {
        return $a;
    }
}
```

This throws a PHP Fatal error as a the concrete implementation is not compatible anymore ([see here](https://3v4l.org/0elpc)).

```
Fatal error: Declaration of AppFoo::bar(string $a): string must be compatible with FooInterface::bar(string $a, array $eventSourceOptions = []): string in php-wasm run script on line 11
```

So to make this change in a BC way, you need:
* to comment the new argument in the interface
* to use func_get_args to fetch the value in the concrete class

You can see here a recent example of Symfony PR with similar BC layer: https://github.com/symfony/symfony/pull/54531

Poke me if you need a hand here!

smnandre (changes requested):
I'm very sorry @ytilotti i must have missed the re-review request.

I just made a comment about the BC break and how to solve it.

Also, could you add a test here ?

I think we will be good for the next release :)

seb-jean:
@ytilotti I think https://github.com/symfony/ux/pull/2447 answers your issue :).

smnandre:
Indeed closing here as the feature is now implemented

PR #1775, opened 2024-04-24
#1770 [Autocomplete] Ignore the ExtraLazyChoiceLoader if the parent $loader is null

### LINKED ISSUE #1770: [Autocomplete] AJAX Autocomplete not working after upgrade to v.2.17 - ExtraLazyChoiceLoader problem
I have my own custom autocompleteField class which worked properly till now. After upgrading the ux-autocomplete to v.2.17 and replacing the parent class `ParentEntityAutocompleteType::class` with the new one `BaseEntityAutocompleteType::class` I receive the following error: 
`Symfony\UX\Autocomplete\Form\ChoiceList\Loader\ExtraLazyChoiceLoader::__construct(): Argument #1 ($decorated) must be of type Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface, null given, called in D:\laragon\www\youcook.online\vendor\symfony\ux-autocomplete\src\Form\BaseEntityAutocompleteType.php on line 41
`

Here is the code of my class:

```
public function configureOptions(OptionsResolver $resolver)
    {
        $resolver->setDefaults([
            'class' => RecipeCategory::class,
            'placeholder' => 'placeholder.recipe_category',
            'choices' =>  $this->treeBuilder->buildTree(),
            'choice_label' => function ($category) {
                return sprintf('%s %s', str_repeat('-', $category->level), $category);
            },
        ]);
    }

    {
    }
```

Perhaps the problem has something to do with the choices option. I pass the `$this->treeBuilder->buildTree()` which returns an array. This seemed to be okay in the past, but somehow is problematic now.

vesselind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| Issues        | Fix #1770 
| License       | MIT

When a custom autoload field class is created and `choices` option is passed, an error occurs: `Symfony\UX\Autocomplete\Form\ChoiceList\Loader\ExtraLazyChoiceLoader::__construct(): Argument #1 ($decorated) must be of type Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface, null given`.

This should fix the problem, described in #1770.

### DIFF
--- a/src/Autocomplete/src/Form/BaseEntityAutocompleteType.php
+++ b/src/Autocomplete/src/Form/BaseEntityAutocompleteType.php
@@ -38,6 +38,10 @@ final class BaseEntityAutocompleteType extends AbstractType
     public function configureOptions(OptionsResolver $resolver): void
     {
         $choiceLoader = static function (Options $options, $loader) {
+            if (null === $loader) {
+                return null;
+            }
+            
             return new ExtraLazyChoiceLoader($loader);
         };
 


kbond:
Thanks for the bug report and fix @vesselind!

PR #1779, opened 2024-04-24
[LiveComponent] Add a note about render:finished event

smnandre:
The `render:finished` event is not dispatched when the component is initially rendered in Twig

As a couple of users have missed this information while reading the documentation, I suggest a little admonition to emphasis it :)


kbond:
Thanks Simon.

PR #1780, opened 2024-04-24
[TwigComponent] Throws if exposed public prop is uninitialized

### LINKED ISSUE #1686: [TwigComponent] PreMount VS required props
First time using TwigComponent, but works like a charm good job 👍 
I've stumbled upon a case that I can't accomplish without feeling like "it's too much code".

Simple case:
- Required props `title`.
- Pass **Entity** to the props which is `Stringable`.
- Component used like this:
   ```twig
   <twig:DropdownHeaderItem title="{{ app.user }}"/>
   ```

Here's the Component:
```php
// src/Twig/Components/DropdownHeaderItem.php

#[AsTwigComponent]
final class DropdownHeaderItem extends DropdownItem
{
    public string $title;
}
```

And his twig:
```twig
{# templates/components/DropdownHeaderItem.html.twig #}

<h6
        {{ attributes }}
        class="dropdown-header {{ this.extraClass }}" {{ macros.attr_to_html(this.attr) }}
>
    {{ this.title }}
</h6>
```

Sadly, this is not enough to make the `title` option required AND with an explicit error message.
If you call the component with no parameters:
```twig
<twig:DropdownHeaderItem/>
```

This is the error that is displayed:
image

Two problems:
1. It show the error INSIDE his component template and not where it is used.
2. The stack trace (as usual with twig) doesn't show where the error occured in twig (only controller with `->render()...`)

Looking [at the doc](https://symfony.com/bundles/ux-twig-component/current/index.html#premount-hook), we should use PreMount hook with the OptionsResolver like this:
```php
#[AsTwigComponent]
final class DropdownHeaderItem extends DropdownItem
{
    public string $title;

    #[PreMount]
    {
        $resolver->setRequired('title');
        $resolver->setAllowedTypes('title', ['string', \Stringable::class]);

    }
}
```

I'm not a fan of that because:
1. Nowadays we can do a lot with only PHP8 (declarations/types/attributes) for simple cases.
    PHP auto `__toString()` call if property typed to `string` when class implements the method.
3. The string reference `'title'` to the property `$title` (maintainability).
4. You need to declare ALL other properties, otherwise you'll get `The option "xxxx" does not exist`
5. All the default value from the PHP declaration, need to be redefined IN the OptionsResolver.
6. +9 lines of code just for a required.

Is there any limitation, to not be able to print required error without the OptionResolver?

smnandre:
Fix #1686, fix ~~#1704~~

### DIFF
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -141,7 +141,11 @@ final class ComponentRenderer implements ComponentRendererInterface
     private function exposedVariables(object $component, bool $exposePublicProps): \Iterator
     {
         if ($exposePublicProps) {
-            yield from get_object_vars($component);
+            $publicProps = get_object_vars($component);
+            if ($uninitializedProps = array_diff_key(get_class_vars($component::class), $publicProps)) {
+                throw new \LogicException(sprintf('Cannot expose uninitialized property "$%s" from "%s".', array_keys($uninitializedProps)[0], $component::class));
+            }
+            yield from $publicProps;
         }
 
         $class = new \ReflectionClass($component);


kbond:
@cavasinf, does this sufficiently fix the two above issues for you?

cavasinf:
Doesn't resolve #1704 because it's more "doc" related.
But certainly #1686 👍

smnandre:
Ok to merge this PR while continuing the discussion about the remaining problems @cavasinf  ?

cavasinf:
Yes 👍 
Remove the #1704 dependency as it will be closed with this PR 😉

smnandre:
(fabbot resolved in another PR)

kbond (approved):

kbond:
Thanks Simon.

norkunas:
This was a BC break..
I have components that have some optional props, and not needed to be initialized if not used

PR #1781, opened 2024-04-25
[LazyImage] Abstract image content fetching

### LINKED ISSUE #1773: [LazyImage] Usage of `file_get_contents` for external images
Hi everyone,

This is more of an RFC/question regarding LazyImage and file_get_contents.

I haven't yet attempted to use `data_uri_thumbnail('https://...')` with an external image (let's say https://symfony.com/doc/6.2fr//the-fast-track/_images/infrastructure.png), but I assume it should work, leveraging [file_get_contents](https://github.com/symfony/ux/blob/83f3e2f/src/LazyImage/src/BlurHash/BlurHash.php#L96).

However, here lies the issue: I'm not particularly fond of using`file_get_contents` for downloading resources over HTTP. Even if we can use it, we're limited in configuring any options (e.g., proxy, authentication, timeout...).

I have two suggestions at the moment:

1. Add support for PSR-7 and PSR-18. Inject a `Psr\Http\Client\ClientInterface` instance, and if we detect that $filename starts with https?://, we utilize the client to download the file.
2. Alternatively, create an abstraction over `file_get_contents`, with an interface and a default implementation using`file_get_contents`. End-users can then use their own implementation, making it easier for us to implement and maintain on our side.

WDYT? Thanks!

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #1773 
| License       | MIT

Hi,

This PR is a proposal for #1773, logic to fetch content image is now customizable thanks to closure, meaning that an end-user can use whatever solution he wants to fetch the image content (HttpClient, Flysystem, ...).

I'm not 100% about the `Closure::fromCallable(new Reference($config['fetch_image_content']))` tho

### DIFF
--- a/src/LazyImage/doc/index.rst
+++ b/src/LazyImage/doc/index.rst
@@ -99,9 +99,37 @@ blurred, data-uri thumbnail of the image:
 
 The ``data_uri_thumbnail`` function receives 3 arguments:
 
--  the server path to the image to generate the data-uri thumbnail for ;
+-  the path to the image to generate the data-uri thumbnail for ;
 -  the width of the BlurHash to generate
 -  the height of the BlurHash to generate
+
+Customizing images fetching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, ``data_uri_thumbnail`` fetches images using the `file_get_contents`_ function.
+It works well for local files, but you may want to customize it to fetch images from a remote server, `Flysystem`_, etc.
+
+To do so you can create a invokable class, the first argument is the filename to fetch:
+
+.. ::
+
+    namespace App\BlurHash;
+
+    class FetchImageContent
+    {
+        public function __invoke(string $filename): string
+        {
+            // Your custom implementation here to fetch the image content
+        }
+    }
+
+Then you must configure the service in your Symfony configuration:
+
+.. code-block:: yaml
+
+    # config/packages/lazy_image.yaml
+    lazy_image:
+        fetch_image_content: 'App\BlurHash\FetchImageContent'
 
 Performance considerations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -190,3 +218,5 @@ https://symfony.com/doc/current/contributing/code/bc.html
 .. _`BlurHash implementation`: https://blurha.sh
 .. _`StimulusBundle`: https://symfony.com/bundles/StimulusBundle/current/index.html
 .. _StimulusBundle configured in your app: https://symfony.com/bundles/StimulusBundle/current/index.html
+.. _`file_get_contents`: https://www.php.net/manual/en/function.file-get-contents.php
+.. _`Flysystem`: https://flysystem.thephpleague.com
--- a/src/LazyImage/src/BlurHash/BlurHash.php
+++ b/src/LazyImage/src/BlurHash/BlurHash.php
@@ -25,10 +25,17 @@ use Symfony\Contracts\Cache\CacheInterface;
  */
 class BlurHash implements BlurHashInterface
 {
+    private \Closure $fetchImageContent;
+
+    /**
+     * @param (callable(string): string)|null $fetchImageContent
+     */
     public function __construct(
         private ?ImageManager $imageManager = null,
         private ?CacheInterface $cache = null,
+        ?callable $fetchImageContent = null,
     ) {
+        $this->fetchImageContent = $fetchImageContent ? $fetchImageContent(...) : file_get_contents(...);
     }
 
     public static function intervention3(): bool
@@ -74,8 +81,10 @@ class BlurHash implements BlurHashInterface
 
     private function generatePixels(string $filename, int $encodingWidth, int $encodingHeight): array
     {
+        $imageContent = ($this->fetchImageContent)($filename);
+
         if (self::intervention3()) {
-            $image = $this->imageManager->read($filename)->scale($encodingWidth, $encodingHeight);
+            $image = $this->imageManager->read($imageContent)->scale($encodingWidth, $encodingHeight);
             $width = $image->width();
             $height = $image->height();
             $pixels = [];
@@ -93,7 +102,7 @@ class BlurHash implements BlurHashInterface
         }
 
         // Resize image to increase encoding performance
-        $image = $this->imageManager->make(file_get_contents($filename));
+        $image = $this->imageManager->make($imageContent);
         $image->resize($encodingWidth, $encodingHeight, static function ($constraint) {
             $constraint->aspectRatio();
             $constraint->upsize();
--- a/src/LazyImage/src/DependencyInjection/LazyImageExtension.php
+++ b/src/LazyImage/src/DependencyInjection/LazyImageExtension.php
@@ -46,13 +46,24 @@ class LazyImageExtension extends Extension implements PrependExtensionInterface
 
         $container
             ->setDefinition('lazy_image.blur_hash', new Definition(BlurHash::class))
-            ->setArgument(0, new Reference('lazy_image.image_manager', ContainerInterface::NULL_ON_INVALID_REFERENCE))
+            ->setArguments([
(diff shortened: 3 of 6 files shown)


src/LazyImage/src/BlurHash/BlurHash.php
```diff
@@ -74,8 +78,10 @@ private function doEncode(string $filename, int $encodingWidth = 75, int $encodi
 
     private function generatePixels(string $filename, int $encodingWidth, int $encodingHeight): array
     {
+        $imageContent = ($this->fetchImageContent)($filename);
```
smnandre, line 84: Maybe just a quick check on the result ?

src/LazyImage/src/BlurHash/FetchImageContent.php
```diff
@@ -0,0 +1,23 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LazyImage\BlurHash;
+
+/**
+ * @author Hugo Alliaume <hugo@alliau.me>
+ */
+final class FetchImageContent
```
smnandre: Is this necessary (genuine question) ? Is there a way to define the closure "file_get_content(...)" in the container ?

src/LazyImage/src/BlurHash/FetchImageContent.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    public function __invoke(string $filename): string
+    {
+        return file_get_contents($filename);
```
smnandre: You risk the TypeError here, if file_get_contents fails it will return "false", not following the "string" return type.

I'd add a if there and throw an exception if the image is not "gettable"

src/LazyImage/doc/index.rst
```diff
@@ -99,10 +99,39 @@ blurred, data-uri thumbnail of the image:
...
+-  the path to the image to generate the data-uri thumbnail for ;
 -  the width of the BlurHash to generate
 -  the height of the BlurHash to generate
 
+Customizing images fetching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, images fetched in ``data_uri_thumbnail`` are fetched using the function
+`file_get_contents`_. It works well for local files, but you may want to customize it
+to fetch images from a remote server, from `Flysystem`_, etc...
```
smnandre:
```suggestion
to fetch images from a remote server, `Flysystem`_, etc.
```

src/LazyImage/src/BlurHash/FetchImageContent.php
Kocal: At first I had this kind of approach in `BlurHash` constructor, but I felt like I had to create an invokable class instead:
```php
$this->fetchImageContent ??= fn (string $filename) => file_get_contents($filename);
```

> Is there a way to define the closure "file_get_content(...)" in the container?

I really don't know :/

src/LazyImage/src/BlurHash/FetchImageContent.php
Kocal: Yup you're right!

What about adding `mixed` and let Intervention does the verification?

src/LazyImage/doc/index.rst
```diff
@@ -99,10 +99,39 @@ blurred, data-uri thumbnail of the image:
...
 
--  the server path to the image to generate the data-uri thumbnail for ;
+-  the path to the image to generate the data-uri thumbnail for ;
 -  the width of the BlurHash to generate
 -  the height of the BlurHash to generate
 
+Customizing images fetching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, images fetched in ``data_uri_thumbnail`` are fetched using the function
```
smnandre: > By default, ``data_uri_thumbnail`` fetch images using the 'file_get_contents' function.

?

src/LazyImage/src/BlurHash/FetchImageContent.php
smnandre: It's more the job of the "fetch_image_content" i think, no ?

src/LazyImage/src/BlurHash/BlurHash.php
```diff
@@ -25,7 +25,11 @@
  */
 class BlurHash implements BlurHashInterface
 {
+    /**
+     * @param \Closure(string): string $fetchImageContent
+     */
     public function __construct(
+        private \Closure $fetchImageContent,
```
kbond:
```suggestion
     * @param callable(string): string $fetchImageContent
     */
    public function __construct(
        private callable $fetchImageContent,
```

src/LazyImage/src/BlurHash/FetchImageContent.php
kbond: > `$this->fetchImageContent ??= fn (string $filename) => file_get_contents($filename);`

Let's go with this to keep this PR simpler.

src/LazyImage/src/BlurHash/BlurHash.php
kbond: This avoids the need the `\Closure::fromCallable()` stuff in the extension.

kbond (changes requested):

src/LazyImage/src/BlurHash/BlurHash.php
Kocal: AFAIK `callable` can not be used as a property type, `Closure` is suggested instead, see https://wiki.php.net/rfc/typed_properties_v2#callable_type

smnandre:
```diff
- $this->fetchImageContent ??= fn (string $filename) => file_get_contents($filename);
+ $this->fetchImageContent ??= file_get_contents(...);
```

src/LazyImage/src/BlurHash/BlurHash.php
kbond: Sorry, I should have been more clear, this is what I meant:

```php
    private \Closure $fetchImageContent;

    /**
     * @param callable(string): string $fetchImageContent
     */
    public function __construct(
        ?callable $fetchImageContent = null,
        private ?ImageManager $imageManager = null,
        private ?CacheInterface $cache = null,
    ) {
        if (!$fetchImageContent) {
            $fetchImageContent = fn(string $v) => file_get_contents($v);
        }

        $this->fetchImageContent = $fetchImageContent(...);
    }
```

Kocal: Let's go for it, thanks :)

src/LazyImage/src/BlurHash/FetchImageContent.php
Kocal: In fact that's not really doable, Intervention type with `mixed` and let the correct handler handle the value (SplFileInfo, Base64, Color, ...):  
image

IMO we should leverage the validation to Intervention

src/LazyImage/src/DependencyInjection/LazyImageExtension.php
```diff
@@ -46,7 +46,11 @@ public function load(array $configs, ContainerBuilder $container)
 
         $container
             ->setDefinition('lazy_image.blur_hash', new Definition(BlurHash::class))
-            ->setArgument(0, new Reference('lazy_image.image_manager', ContainerInterface::NULL_ON_INVALID_REFERENCE))
+            ->setArguments([
+                new Reference('lazy_image.image_manager', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+                null, // $cache
+                null, // $fetchImageContent
+            ])
```
Kocal, line 53: Looks like I can't set the 3rd argument if the 2nd argument has not been set, which can be the case if we have no cache configured, that's why I've changed to `->setArguments`

Kocal:
Reviews have been applied, thanks :)

smnandre (approved):

kbond (approved):

kbond:
Thank you Hugo.

PR #1782, opened 2024-04-25
[TwigComponent] ingore content when printed as string 

### LINKED ISSUE #1021: [TwigComponent] The Component HTML syntax  is not preserved when printed as (raw) string
The verbatim tag marks sections as being raw text that should not be parsed, but this is not the case when using [the html syntax for components](https://symfony.com/bundles/ux-twig-component/current/index.html#component-html-syntax).  

In my use case, I've created a bundle that generates dashboard code using twig components that I've defined in my bundle, 

I want the generated code to have html syntax, but it doesn't, for exemple : 

```twig
{% verbatim %}
<twig:DashliteResponsiveTable :data="data">
  <twig:block name="header">
      <twig:DashliteResponsiveTableColumn value="Nom" col />
      <twig:DashliteResponsiveTableColumn tools />
  </twig:block>

    <twig:DashliteResponsiveTableColumn :value="item.name" />
    <twig:DashliteResponsiveTableColumn tools>
        <twig:DashliteResponsiveTableAction path="" icon="edit" />
        <twig:DashliteDeleteButton path="" :item="item" />
    </twig:DashliteResponsiveTableColumn>
</twig:DashliteResponsiveTable>
{% endverbatim %}
```

renders 

```
{% component 'DashliteResponsiveTable' with { data: data } %}
    {% block header %}
        {{ component('DashliteResponsiveTableColumn', { value: 'Nom', col: true }) }}
        {{ component('DashliteResponsiveTableColumn', { tools: true }) }}
    {% endblock %}

    {% block content %}
        {{ component('DashliteResponsiveTableColumn', { value: item.name }) }}
        {% component 'DashliteResponsiveTableColumn' with { tools: true } %}
            {% block content %}
                {{ component('DashliteResponsiveTableAction', { path: , icon: 'edit' }) }}
                {{ component('DashliteDeleteButton', { path: , item: item }) }}
            {% endblock %}
        {% endcomponent %}
    {% endblock %}
{% endcomponent %}
```

excepted result : 

```
<twig:DashliteResponsiveTable :data="data">
  <twig:block name="header">
      <twig:DashliteResponsiveTableColumn value="Nom" col />
      <twig:DashliteResponsiveTableColumn tools />
  </twig:block>

    <twig:DashliteResponsiveTableColumn :value="item.name" />
    <twig:DashliteResponsiveTableColumn tools>
        <twig:DashliteResponsiveTableAction path="" icon="edit" />
        <twig:DashliteDeleteButton path="" :item="item" />
    </twig:DashliteResponsiveTableColumn>
</twig:DashliteResponsiveTable>
```

how can I keep this syntax when generating code using twig itself? 

if this requires a PR I'd be happy to contribute just give me instructions and what I can do

Thanks.

bernard-ng:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1021 
| License       | MIT

When a component is displayed as a character string, it should not be transformed. This is particularly useful for code generation, as described in issue 

this solution needs to be improved to handle various complex cases, for example, the possibility of taking into account filters such as “raw” etc... and that of not taking into account spaces between “{{” and the character string, for example.

I'd be delighted to have some guidance in this direction, but for now here's a quick fix

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -70,6 +70,16 @@ class TwigPreLexer
                 }
             }
 
+            // ignore content when printed as string, see #1021
+            if ($this->consume('{{ "')) {
+                $output .= $this->consumeUntil('" }}');
+                $this->consume('" }}');
+
+                if ($this->position === $this->length) {
+                    break;
+                }
+            }
+
             if ($this->consume('{% embed')) {
                 $inTwigEmbed = true;
                 $output .= '{% embed';
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -366,5 +366,14 @@ final class TwigPreLexerTest extends TestCase
             '<twig:foobar bar="baz" {{ ...attr }}>content</twig:foobar>',
             '{% component \'foobar\' with { bar: \'baz\', ...attr } %}{% block content %}content{% endblock %}{% endcomponent %}',
         ];
+        yield 'ignore_printed_component' => [
+            '{{ "<twig:Alert/>" }} <twig:Alert/>',
+            '<twig:Alert/> {{ component(\'Alert\') }}',
+        ];
+
+        yield 'file_ended_with_printed_component' => [
+            '{{ "<twig:Alert/>" }}',
+            '<twig:Alert/>',
+        ];
     }
 }


smnandre:
I'm not entirely sure there. This seems to me very similar to the escape requirement to display the `{{` delimiter.

https://twig.symfony.com/doc/3.x/templates.html#escaping

bernard-ng:
It's prevent component transformation but it's optimal I need guidance please, how can we handle "raw" filter and printing more effectively

PR #1784, opened 2024-04-26
[Site] Grouped fixes

smnandre:
See commit names :)


ux.symfony.com/importmap.php
```diff
@@ -84,25 +82,25 @@
         'package_specifier' => 'vue/dist/vue.esm-bundler.js',
     ],
     'swup' => [
-        'version' => '4.6.0',
+        'version' => '3.1.1',
```
Kocal, line 85: Why downgrading Swup that much?

smnandre: I did not touch the import map... weird :|

Kocal: Mmmh yup, maybe it was out of sync? 
Or maybe it's due to `importmap:install` in composer' auto-script? Since you touched dependencies... Otherwise I don't really have explanation

smnandre: Yep, seems logical/expected: https://github.com/symfony/ux-swup/blob/7a93c159b26f02f1ff0d9d64465f7fe6dfc7b78c/assets/package.json#L22

kbond: I've noticed this lately too: when doing recipe updates - ~I think it might be a flex bug~ oh, didn't notice the package.json in swup.

smnandre: No it's the opposite i think it's update when we build the packages :|

kbond: Ok, yeah I did notice it when doing recipe updates on ux.symfony.com. So this change is legit?

smnandre: Yes i think

kbond:
Thanks Simon.

PR #1790, opened 2024-04-26
[Doc] Remove the --dir option in debug:twig-component command

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

I tried adding the `--dir` option and the command showed an error: `The "--dir" option does not exist.`

Looking at the source code, it looks like this option doesn't exist:

https://github.com/symfony/ux/blob/4b3fcaea426d7493e7151c97ba63697be0cc5061/src/TwigComponent/src/Command/TwigComponentDebugCommand.php#L47-L66

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -118,7 +118,7 @@ and any other components by running:
 
 .. code-block:: terminal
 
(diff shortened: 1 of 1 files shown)


kbond:
Thanks Javier.

PR #1793, opened 2024-04-26
[Site] Configure Twig-CS-Fixer

### LINKED ISSUE #1166: [CS] Use Twig CS Fixer
With all the Twig / HTML we may write in a near future...  would that be a good idea to use TwigCSFixer ?

https://github.com/VincentLanglet/Twig-CS-Fixer

smnandre:
(and apply rules)

Fully isolate website GA workflows from the components

Fix #1166


kbond (approved):

kbond:
Thanks Simon.

PR #1795, opened 2024-04-27
Adjust GHA to account for fabbot

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | n/a
| License       | MIT

fabbot now handles cs

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,17 +11,6 @@ on:
             - 'ux.symfony.com/**'
 
 jobs:
-    coding-style-php:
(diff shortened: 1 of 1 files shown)


smnandre:
Does he uses our phpcsfixer config ?

smnandre (approved):

kbond:
> Does he uses our phpcsfixer config ?

This is my understanding, yes.

---

This PR has been stuck on pending and I see a few others like this. Some seem to pass ok though...

PR #1798, opened 2024-04-28
[Icons] Use lowercase <twig:ux:icon /> in documentation

smnandre:
As previously discussed with @kbond, we should use only one format in documentation / website / etc.

(both will work, it really is "just" a communication thing)

This PR update the documentation and the README (website is already OK)


kbond (approved):

kbond:
Thanks Simon.

PR #1805, opened 2024-04-30
[TwigComponent] Fix aria attribute cannot be removed

smnandre:
Handle only "true" (the original need) and restore using false to remove an attribute

cf #1709 and #1797


kbond (approved):

kbond:
Thanks Simon.

PR #1809, opened 2024-04-30
[Translator] Create valid constant names for keys starting with numeric chars

wolfgangweintritt:
If the key of a translation started with a numeric character, the constant's name would also start with a numeric character, thus resulting in a JS syntax error. Prefixing the constant's name with an underscore makes it valid.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

### DIFF
--- a/src/Translator/src/TranslationsDumper.php
+++ b/src/Translator/src/TranslationsDumper.php
@@ -166,7 +166,7 @@ TS
 
         $prefix = 0;
         do {
-            $constantName = s($translationId)->ascii()->snake()->upper()->toString().($prefix > 0 ? '_'.$prefix : '');
+            $constantName = s($translationId)->ascii()->snake()->upper()->replaceMatches('/^(\d)/', '_$1')->toString().($prefix > 0 ? '_'.$prefix : '');
             ++$prefix;
         } while (\in_array($constantName, $alreadyGenerated, true));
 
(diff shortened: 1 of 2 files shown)


kbond:
Thanks for this fix @wolfgangweintritt!

@Kocal, mind taking a look at this?

src/Translator/src/TranslationsDumper.php
```diff
@@ -166,7 +166,7 @@ private function generateConstantName(string $translationId): string
 
         $prefix = 0;
         do {
-            $constantName = s($translationId)->ascii()->snake()->upper()->toString().($prefix > 0 ? '_'.$prefix : '');
+            $constantName = preg_replace('/^(\d)/', '_$1', s($translationId)->ascii()->snake()->upper()->toString().($prefix > 0 ? '_'.$prefix : ''));
```
Kocal: LGTM, but I would be in favor to use `->replaceMatches` from `s()` instead of calling `preg_replace` directly

Kocal (approved):

src/Translator/src/TranslationsDumper.php
wolfgangweintritt: Thanks for the review! Yeah I agree - I've changed it.

kbond:
Thank you Wolfgang, great first PR!

PR #1810, opened 2024-04-30
[TwigComponent]: Fix silently failing tests

adrienlucas:
| Q             | A
| ------------- | ---
| Bug fix?      no
| New feature?  no 
| Issues        | Fix silently failing tests
| License       | MIT

This tries to fix two little mistakes  : 
 - the `expectException` should never be used twice in a test (because it does not actually try/catch, the first exception will stop the execution of the test and will be catched outside).
 - the piece of code that is expected to throw an exception should always come last in the test (because again, no try/catch, so any line after the throwing line will not be executed).


smnandre:
Thank you @adrienlucas !

smnandre (approved):

kbond (approved):

kbond:
Excellent catch Adrien, thank you!

PR #1814, opened 2024-05-01
[Site] Add robots.txt & sitemap.xml

smnandre:


ux.symfony.com/src/Controller/MainController.php
```diff
@@ -29,4 +30,13 @@ public function homepage(UxPackageRepository $packageRepository): Response
...
     }
+
+    #[Route(path: '/robots.txt', name: 'app_robots')]
+    public function __invoke(Request $request): Response
+    {
+        $response = $this->render('robots.txt.twig');
+        $response->headers->set('Content-Type', 'text/plain');
+
+        return $response;
+    }
```
Kocal, line 41: What about creating the file in `public/`? 
So the file can be delivered by the webserver instead of the Symfony app

Kocal: Ah, because you use `{{ url('app_sitemap') }}`

smnandre: To be fair, you're right, and i think we will soon render it once after the deployment ... but i need to check first how it goes :)

kbond (approved):

kbond:
Thanks Simon.

PR #1820, opened 2024-05-02
[TwigComponent] Allow input props to have the same name as context variables

squrious:
| Q            | A   |
|--------------|-----|
| Bug fix?     | yes |
| New feature? | no  |
| Issues       | -   |
| License      | MIT |

It looks like #1652 introduced an issue with input props and parent context.

If:
- A component prop has the same name as a variable in the parent context
- The context variable is used to set the prop value as an input prop
- The component is rendered with the embed strategy

Then, the default value of the prop is used instead of the one passed to the component.

Example: 

```twig
{# templates/components/Hello.html.twig #}

{% props name = 'John' %}
<div {{ attributes }}>
    Hello {{ name }}!
</div>
```

### Render embedded

```twig
{% set name = 'Bryan' %}
<twig:Hello :name="name"></twig:Hello>
```

Wrong output: 
```html
<div>
    Hello John!
</div>
```

### Render with function

```twig
{% set name = 'Bryan' %}
<twig:Hello :name="name" />
```

Correct output: 
```html
<div>
    Hello Bryan!
</div>
```

### DIFF
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -117,8 +117,9 @@ final class ComponentRenderer implements ComponentRendererInterface
             // first so values can be overridden
             $context,
             // add the context in a separate variable to keep track
-            // of what is coming from outside the component
-            ['__context' => $context],
+            // of what is coming from outside the component, excluding props
+            // as they override initial context values
+            ['__context' => array_diff_key($context, $props)],
             // keep reference to old context
             ['outerScope' => $context],
             // add the component as "this"
@@ -180,7 +181,7 @@ final class ComponentRenderer implements ComponentRendererInterface
             $name = $attribute->name ?? (str_starts_with($method->name, 'get') ? lcfirst(substr($method->name, 3)) : $method->name);
 
             if ($method->getNumberOfRequiredParameters()) {
-                throw new \LogicException(sprintf('Cannot use %s on methods with required parameters (%s::%s).', ExposeInTemplate::class, $component::class, $method->name));
+                throw new \LogicException(sprintf('Cannot use "%s" on methods with required parameters (%s::%s).', ExposeInTemplate::class, $component::class, $method->name));
             }
 
             if ($attribute->destruct) {
--- a/src/TwigComponent/src/Twig/PropsNode.php
+++ b/src/TwigComponent/src/Twig/PropsNode.php
@@ -57,7 +57,7 @@ class PropsNode extends Node
             if (!$this->hasNode($name)) {
                 $compiler
                     ->indent()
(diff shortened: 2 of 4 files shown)


src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -180,7 +182,7 @@ private function exposedVariables(object $component, bool $exposePublicProps): \
             $name = $attribute->name ?? (str_starts_with($method->name, 'get') ? lcfirst(substr($method->name, 3)) : $method->name);
 
             if ($method->getNumberOfRequiredParameters()) {
-                throw new \LogicException(sprintf('Cannot use %s on methods with required parameters (%s::%s).', ExposeInTemplate::class, $component::class, $method->name));
+                throw new \LogicException(sprintf('Cannot use "%s" on methods with required parameters (%s::%s).', ExposeInTemplate::class, $component::class, $method->name));
```
smnandre, line 184: ❤️

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -113,12 +113,14 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
...
+        $propsKeys = array_keys($props);
         $variables = array_merge(
             // first so values can be overridden
             $context,
             // add the context in a separate variable to keep track
-            // of what is coming from outside the component
-            ['__context' => $context],
+            // of what is coming from outside the component, excluding props
+            // as they override initial context values
+            ['__context' => array_filter($context, static fn ($key) => !\in_array($key, $propsKeys), \ARRAY_FILTER_USE_KEY)],
```
smnandre:
```suggestion
            ['__context' => array_diff_key($context, $props)],
```

(I have a doubt)

smnandre:
Maybe add a test case when name is passed with "null" ?

squrious:
Thanks for the review @smnandre!

> Maybe add a test case when name is passed with "null" ?

I wasn't sure about that. If you pass null, the prop is considered unset, so yes we can't pass a null value explicitly. 

But there is a systematic use of `isset` in the PropsNode class, so I thought it was intended for some reason. Would you prefer to replace them with  `array_key_exists`?

src/TwigComponent/src/ComponentRenderer.php
squrious: Good fabbot

src/TwigComponent/src/ComponentRenderer.php
squrious: Thanks for the tip! Fixed :)

smnandre:
Oh we lost some hair with @WebMamba last time, i will trust you on this ^^

WebMamba (approved):
Thanks a lot @squrious, I over thinking this part, it's much simpler and less buggy! Great job!

kbond:
This is great, thanks Nicolas!

PR #1821, opened 2024-05-02
[Site] Fix input range color

smnandre:
Previous value made it impossible to use without a background (ex: charts JS demo)


WebMamba (approved):

kbond:
Thanks Simon.

PR #1822, opened 2024-05-03
[Site] Remove light mode & clean templates

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #.
| License       | MIT

* Dark mode for everyone
* Restructure template organisation
* Rework banner
* A lot of tiny things


javiereguiluz:
I'd like to ask you to reconsider the removal of light mode. For folks living in countries with lots of Sun and for folks working in well-lit workplaces, light mode is a must 🙏 

But, if maintaining an additional theme takes a lot of effort, then go ahead and remove it (better spend that effort on other more useful things). Thanks!

smnandre:
Thank you for your comment @javiereguiluz 

As you mentioned, this is primarily a matter of resources.

Currently, I have three demos on my task list, approximately fifty new assets, and four substantial components. Over the next few weeks and months, we plan to create _**many**_ new pages and components, accompanied by numerous design tasks.

For a limited time, I'd like to focus solely on one theme to gain the agility I currently lack.

My original "plan":

**This week:**
* Remove the light mode (this PR)
* Start removing unused CSS

**In the following weeks:**
* Significantly reduce our use of Bootstrap
* Strengthen our HTML/CSS codebase
* Implement all the new features we have planned

**Subsequently:**
* Reintroduce the light theme.
* Add options for dimmed, contrast settings, etc.

I realize this approach may not be ideal for everyone, and I apologize in advance if we decide to proceed.

I'm not making any promises yet, but I will take some time to consider alternative plans.

Let’s pause for 24 hours before making any decisions.

javiereguiluz:
Simon, it sounds like a good plan because removing the light mode would be temporary and it will be hopefully readded in the future 👍 

Just another comment: if you want to go faster and be more productive, I think you should use Bootstrap components more instead of less. Modern Bootstrap allows to customize almost everything using CSS properties, so you can use Bootstrap without looking like default Bootstrap.

Kocal:
Hey, I was like 😱 when reading the PR's title, but I'm more reassured after reading the whole PR. Yeah, please keep the light theme 🥲 🙏🏻 

People are too "hyped" by the dark theme, but the truth is that white text on dark background is harder to read than black text on light background.

More generally, you may want to use a light theme during the day, and a dark theme during the night, this way you don't have to increase a lot your monitor's brightness (which damages your eyes).

Thanks 🙏🏻

WebMamba (commented):
Hey I was reviewing this PR, and I end up with a question: The problem we have here is it really because it's hard to maintain two theme or because the theming system is not implemented well? If we struggle to have two theme because the code is not good enough we should refactor this code! In front development theming is a huge topics, and since SymfonyUX is made also for front end development we should show the exemple and demonstrate that is possible to have a correct design system with SymfonyUX.

I think we should take a little step back here, look over and find the reason that made this theming task over complicated. Make a list, and try to figure out if some issue can't be solve.
For example when digging into the scss file I see that a lot of colors and value are hard coded. Maybe we can start by that?

WDYT ?

smnandre:
> it's hard to maintain two theme 

Never said that... i said it's hard "right now", so we agree on this. It should be easy and i'm confident it will be once we're done here.

> For example when digging into the scss file I see that a lot of colors and value are hard coded.

I know, and even if i added some lately, i can garantee you there are less than when i started working on the website 6-8 monthes ago. I'm sure you're not pointing at anynone saying that, but we need to have some pragmatic approach here. 

The website is open-source, but it's not a project like the "Symfony Demo", and its PHP code does not follow all best practices neither, because we focused on delivering content to showcase the SymfonyUX packages, and i think on this point there have been massive improvments in the past monthes. 

> we should refactor this code

Did you read my comments or am i that bad expressing myself (please don't answer haha 😆 )

That's entirely the concept of this PR, which will lead to other ones as I said. 

We need to move quickyl, as we have many parrallel projects / content incoming.

I could have said: let's have a long and "V-shape" project, and when we're ready we will switch all the front at once, but i'm convinced it's not the good idea here... 

So the idea here is to lighten temporarely the front-end tools we use, as Bootstrap has a very specific way of doing things, to **_rebuild_** on saner grounds.

--- 

I bounce on what @javiereguiluz said about resources, i'm personnaly 100% convinced our energy should be spent on 3 things 
* features: creating more component and/or adding new cool stuff on the existing ones, and fixing all the little bugs/unexpected/magical behaviours
* DX: easing the usage of Symfony UX for the everyday life of developers. Meaning better errors handling, debug tools, makers, documentation (which tool / method for which need, etc..)
* spreading the word: as we discussed many time together, and as we saw at Symfony Live Paris, a lot of people do not know what is SymfonyUX, what the big components allow, and we need to spread the word. For this the cookbook is a brilliant idea / thing, i repeat once more a blog / newsfeed is required too, as is rebranding the UX communication materials ("baseline" / "punchlines" / even github tags...) to make thinks a lot more in harmony with the current state of the project

So, the website should be a _tool_ dedicated to those goals, and that should make us win time in fullfilling them, not somethings that make us doubt or think twice.

Today, it is not that tool, and to became this tool, we need major changes in the way it's coded (in HTML/Twig mainly, CSS beeing a consequence of it).

So what i'm pushing here is to do those changes "live", with the minor drawback to temporarely disable features (knowing that most of them did not exist only a couple monthes before)

The final goal is the same as it always was: a website to promote and showcase "Symfony UX" components in the best way we can.

smnandre:
PR replaced by #1827 - so i'm closing this one.

I'll _try_ to refacto the front + implement all the new features while keeping the light mode. 

But i'm not making _any_ promise here.

PR #1825, opened 2024-05-03
[Autocomplete] Fix grouped options order

### LINKED ISSUE #1616: Display Problem UX Autocomplete / TomSelect
I have the problem with the newest version, that on grouped Autocomplete/TomSelect, the selected item gets duplicated to the end of the list, or sometimes the grouping completely disappears.
Like in this example:
Before selecting a different item:
image
After selecting the other item:
image

With other UX Autocompletes the items are additionally added to the bottom of the list as new items. 

Tested on safari and firefox. 
symfony/ux-autocomplete	2.16.0
tom-select	2.3.1

Does anyone have these effects too?

zavarock:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1616 
| License       | MIT

The bug is being caused by TomSelect not preserving the order of the option elements in the select as we select the options in the dropdown. In this case, the MutationObserver callback uses the optgroup element as a parameter to "store" the group (if any) to which the option belongs. However, once the option is selected, it no longer has an optgroup as its parentElement (a problem caused by the aforementioned bug).

As I see it, there is no need to "store" the option's group since, in any case, all <option> elements usually have a unique [value], and even if not, it will still work as expected.

A callback was added for options added through [addOption](https://github.com/orchidjs/tom-select/blob/69180fa9e79060060f1824fbde85537579d0005d/src/tom-select.ts#L1636), but the caveat is that it needs the parameter user_created=true to trigger the 'option_add' event.

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -195,11 +195,9 @@ class default_1 extends Controller {
     }
     createOptionsDataStructure(selectElement) {
         return Array.from(selectElement.options).map((option) => {
-            const optgroup = option.closest('optgroup');
             return {
                 value: option.value,
                 text: option.text,
-                group: optgroup ? optgroup.label : null,
             };
         });
     }
@@ -216,7 +214,7 @@ class default_1 extends Controller {
         if (filteredOriginalOptions.length !== filteredNewOptions.length) {
             return false;
         }
-        const normalizeOption = (option) => `${option.value}-${option.text}-${option.group}`;
+        const normalizeOption = (option) => `${option.value}-${option.text}`;
         const originalOptionsSet = new Set(filteredOriginalOptions.map(normalizeOption));
         const newOptionsSet = new Set(filteredNewOptions.map(normalizeOption));
         return (originalOptionsSet.size === newOptionsSet.size &&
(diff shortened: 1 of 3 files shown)


simonsolutions:
Is the PR planned to be released in the near future?

IndraGunawan:
@smnandre @kbond any chance to review this?

smnandre:
Oh i'm so sorry @IndraGunawan .. i remember seeing this PR.... and i forgot to come back on it ! 

For Autocomplete it seems a very smart & efficient fix: thank you _**very much**_ for this! 

Just, before we merge i'd like to be sure this does not add problems/side effets when used with LiveComponent
(to be transparent: i'm 90% certain this will in fact solve/improve things..... but i'd like to be 100%)

So... did you try within a LiveComponent ? 
I'm not talking about coding a test or adding anything more in this PR..  But at least one of us should test manually to check this new behaviour has no impact.

I can help if you want / don't have time, feel free to ask!

A good example would be to add groups to the dependant form fields demo of the ux.website (present in the same repository) and see if everything is still working.

smnandre:
⚠️ @kbond 

I marked this as reviewed/accepted because the PR in itself is ok !

Let's just wait we check everything is ok with LiveComponent before merging.

zavarock:
@smnandre Really appreciate your feedback because I myself didn't know if it was a good solution. I was waiting for feedback from you.
Regarding LiveComponent, to be honest, I haven't tested with it yet, but I will later and let you know.

smnandre:
If you don't find time, poke me later in the week i'll make a test or two

zavarock:
@smnandre Sorry for the delay.
As you suggested, I grouped the options from the dependent form on the UX website and everything seems fine.

ezgif-6-bc9fee1be0

Would you like me to update this PR with the changes to the dependent form?

smnandre:
@zavarock thank you for taking some time to check this! 

No need to put it in the PR, and i'd rather keep this particular demo the simplest possible (as it's one of the most seen...)


But after this PR, it could be a good idea to add a couple of demo / code samples for Autcomplete (only if you want and have time of course)

zavarock:
@smnandre Sure, no problems!
I just have two questions if you don't mind: 
1) You're talking about adding a new demo [on this page](https://ux.symfony.com/autocomplete), right?
2) Is it better to create a new PR to avoid mixing things? Or can I just update this one?

smnandre:
1. More probably a dedicated page. I will post a page during the week to explain how to do this properly (short answer, same thing as the LiveComponent demos... but i'll try to isolate more the demos)

2. I personnaly find better to separate code PRs and website ones.. but that's 100% open to debate :)

simonsolutions:
How about the status to integrate that issue in the next release, as it is important for the application to have a proper grouped dropdown. As i read it the issue is fixed and working, the demo could be a separate thing to update.

simonsolutions:
When can we expect the merge into the next release?

IndraGunawan:
@smnandre i'm so sorry for following this up over and over again but can we wrap up this PR in the near future?

smnandre (approved):

Kocal (approved):

smnandre:
Thank you @zavarock.

smnandre:
@IndraGunawan : it's merged 😄

IndraGunawan:
yay, thank you @smnandre

thank you @zavarock for your great work 😀

PR #1828, opened 2024-05-06
[LiveComponent] Tokenize classes on all allowed whitespaces

### LINKED ISSUE #1819: [LiveComponent] Morph error trying to apply invalid classes
When live components try to apply changes to elements for some reason the different classes are picked up with whitespaces or even empty strings. Browser console then throws `Uncaught (in promise) DOMException: DOMTokenList.remove: The empty string is not a valid token.`.

This happened here when an element was morphed that probably shouldn't have been in the first place (and `data-skip-morph` doesn't solve the problem).

---

For example when an inline menu (with ID and `data-skip-morph`) is displayed after clicking an element only the class "hidden" changes. Then after updating the live component the element and its menu are still there (but the menu should be invisible, as backend renders context menus hidden of course):

The changes that are picked up:
- `addedClasses: Set [ "w-full" ]`.
- `removedClasses: Set(3) [ "", "hidden\n", "w-full\n" ]`

---

In my case, simply filtering class changes at least prevents the error and didn't have any side effects here:

```JS
className = className.trim();

if (className === '') {
    return;
}
```

https://github.com/symfony/ux/blob/80b36bcbeb2902a55495db2afd544f738c73148f/src/LiveComponent/assets/src/Rendering/ElementChanges.ts#L13-L23

Am I doing something wrong?

aleho:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1819 
| License       | MIT

The class attribute can be tokenized by any white space as defined by the HTML spec.

When a class attribute contains whitespaces other than a space change tracking fails and invalid classes are applied resulting in `Uncaught (in promise) DOMException: DOMTokenList.remove: The empty string is not a valid token.`.

### DIFF
--- a/src/LiveComponent/assets/src/Rendering/ExternalMutationTracker.ts
+++ b/src/LiveComponent/assets/src/Rendering/ExternalMutationTracker.ts
@@ -189,16 +189,8 @@ export default class {
     private handleClassAttributeMutation(mutation: MutationRecord, elementChanges: ElementChanges) {
         const element = mutation.target as Element;
 
-        const previousValue = mutation.oldValue;
-        const previousValues = previousValue ? previousValue.split(' ') : [];
-        previousValues.forEach((value, index) => {
-            const trimmedValue = value.trim();
-            if (trimmedValue !== '') {
-                previousValues[index] = trimmedValue;
-            } else {
-                previousValues.splice(index, 1);
-            }
-        });
+        const previousValue = mutation.oldValue || '';
+        const previousValues: string[] = previousValue.match(/(\S+)/gu) || [];
 
         const newValues: string[] = [].slice.call(element.classList);
         const addedValues = newValues.filter((value) => !previousValues.includes(value));
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1822,17 +1822,8 @@ class ExternalMutationTracker {
     }
     handleClassAttributeMutation(mutation, elementChanges) {
         const element = mutation.target;
-        const previousValue = mutation.oldValue;
-        const previousValues = previousValue ? previousValue.split(' ') : [];
-        previousValues.forEach((value, index) => {
-            const trimmedValue = value.trim();
-            if (trimmedValue !== '') {
-                previousValues[index] = trimmedValue;
-            }
-            else {
-                previousValues.splice(index, 1);
-            }
-        });
+        const previousValue = mutation.oldValue || '';
+        const previousValues = previousValue.match(/(\S+)/gu) || [];
         const newValues = [].slice.call(element.classList);
         const addedValues = newValues.filter((value) => !previousValues.includes(value));
         const removedValues = previousValues.filter((value) => !newValues.includes(value));
--- a/src/LiveComponent/assets/test/Rendering/ExternalMutationTracker.test.ts
+++ b/src/LiveComponent/assets/test/Rendering/ExternalMutationTracker.test.ts
@@ -114,7 +114,7 @@ describe('ExternalMutationTracker', () => {
         element.classList.remove('second-class');
         element.classList.add('second-class');
         // add new (with some whitespace to be sneaky)
-        element.setAttribute('class', ` ${element.getAttribute('class')} new-class `)
+        element.setAttribute('class', ` ${element.getAttribute('class')} \n    new-class `)
         // remove
         element.classList.remove('first-class');
         // add then remove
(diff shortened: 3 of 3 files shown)


aleho:
I don't understand why `.splice()` was used instead of a new array. Can't be for performance reasons, or can it?

aleho:
Also, forgot to mention: I'd consider a regex to tokenize the string a more elegant solution.

Any objections against that?

smnandre:
Could you rebuild the packages ? (that's the origin of the CI failures)

(yarn run build from the repository root i think)

aleho:
Of course! Didn't find anything in the docs and wasn't sure whether that's something that actually has to be done in the PR or will be done on release.

smnandre:
Yep this would need some CONTRIB file.. i'll open an issue :)

smnandre:
https://github.com/symfony/ux/issues/1836 :)

smnandre:
Do you think you could add a quick test, to prevent any future regression ? 😊

aleho:
I would like to but I didn't see any browser tests in LiveComponents.

Could you point me in the right direction where actual browser integration is tested for `live_controller.js`?

aleho:
Never mind. I figured LiveComponent doesn't have any browser tests yet and tried to add them.

This PR got bigger than expected, please let me know if anything's missing.

smnandre:
wow indeed :) I'll look in the week as soon as i can ! Thank you for the efforts :)

aleho:
@smnandre Any news? Did you have time to have a quick look at it?

smnandre:
Hello @aleho ! Sorry i've been crazy busy those last weeks... 

Thank you **very much** for this!

I'm not sure we need right now a full panther stack with panther :| (really not sure there, but could this test be done like the others ?)

I'll let @kbond (that is his domain much more than mine to be honest) and @WebMamba check and give their opinion :)) 

My _personal_ impresion: we should maybe try to make the suggested changes in a smaller PR, and address the "panther tests" in a distinct one ? 🤷

aleho:
> Hello @aleho ! Sorry i've been crazy busy those last weeks...

No worries, we all know how it is.

> I'm not sure we need right now a full panther stack with panther :| (really not sure there, but could this test be done like the others ?)

There are already Panther tests for Turbo, which this PR is somewhat based on. The reason I think there's no other way to "reliably" test this is that it has to be run in a JS engine in a browser to break. Being the author of this PR I might be biased though. ;)


> My _personal_ impresion: we should maybe try to make the suggested changes in a smaller PR, and address the "panther tests" in a distinct one ? 🤷

I can do that. On the other hand, the bigger part of this PR will only affect the tests / GitHub actions, so it's not going to have a lot of impact even if it breaks (and, as I said, Panther is already used anyway). Moreover this bug fix would not have a valid test case if it is split until the other part is merged as well.

aleho:
This could have been a nice fix for 2.18, just saying 😉.

I'm still fine with either solution! Please just tell me how you'd like to move forward, but, please, don't let this sit for another month and then another for a release. 😭

smnandre:
We've all been crazy busy around here these last weeks... it's an open source project you know, and we have limited time after work, family and stuff.. 

And believe me when i say we would all love to have more time 😄 

I just poked other regulars to come review this / give you feedback ... so that will be soon i think :)

src/LiveComponent/assets/src/Rendering/ExternalMutationTracker.ts
```diff
@@ -189,16 +189,8 @@ export default class {
...
-        previousValues.forEach((value, index) => {
-            const trimmedValue = value.trim();
-            if (trimmedValue !== '') {
-                previousValues[index] = trimmedValue;
-            } else {
-                previousValues.splice(index, 1);
-            }
-        });
+        const previousValue = mutation.oldValue || '';
+        const previousValues = previousValue.match(/(\S+)/gu) || [];
```
smnandre: Could you keep juste this file 
(and the generated files in dists), 

and add just a test or two in `assets/test/Rendering/ExternalMutationTracker.test.js` ? 


I'm not sure we should create and mainain an entire new "env" ... and i'm sure i don't want to need webpack locally :)) 

(i'm trying to slim the PR down to merge it rapidely :) )

aleho: I added tests and made sure they break before the fixes are applied.

WebMamba (approved):
Thanks for your work! I just tested it locally and it work great!

smnandre (approved):
Thank you very much for this fix @aleho !

(... and for your patience ;) )

kbond:
Thanks a lot Alexander!

PR #1829, opened 2024-05-06
Fixed event data passing example

manuelkiessling:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| License       | MIT

Looks like some obsolete data access approach via a $data array was still in the docs.

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2728,7 +2728,7 @@ You can also pass extra (scalar) data to the listeners::
         // ...
 
         $this->emit('productAdded', [
-            'product' => $product->getId(),
+            'productId' => $product->getId(),
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -2728,18 +2728,18 @@ You can also pass extra (scalar) data to the listeners::
         // ...
 
         $this->emit('productAdded', [
-            'product' => $product->getId(),
+            'productId' => $product->getId(),
```
smnandre, line 2731: That will make the Product entity example (L2745-2752) not working :|

smnandre: Could you keep the "product" var name and fix the $data[...] you spotted ? :)

smnandre:
Done in #2402

PR #1827, opened 2024-05-06
[Site] Clean templates

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #
| License       | MIT

Replaces #1822 

All the clean / fixes / CS   (without removing the light mode for now)

PR #1831, opened 2024-05-06
[Site] Icons colors

smnandre:
* don't invert colored icons
* add 3 icon set cards
* re-add SVG code highlighting

(to merge after #1827)


kbond:
Thanks Simon.

PR #1841, opened 2024-05-12
[Live] Remove obsolete internal

smnandre:
The `component_url` Twig function is not used (for a lot of time), is internal and.... unused.

So this PR removes it.

(i'm starting a little inventory of the code to see what is possible in term of separation between the components)


kbond:
Hmm, not sure about this. While the code to provide the twig function is internal, the twig function itself isn't internal and might be used.

If I recall correctly, this function was added as a way to lazily load twig components with turbo. It might be used in the wild. I wonder though, now that we have true lazy components, is it still needed? Maybe we should deprecate and see?

smnandre:
Oh i did not know that, i imagined it was some internal thing from the start. Let's close it then :)

PR #1845, opened 2024-05-13
[TwigComponent] Update DOC preMount validation data code

### LINKED ISSUE #1704: [TwigComponent] Missleading doc between `Passing & Rendering Attributes` and `PreMount validation`

cavasinf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #1704 
| License       | MIT

Fixes the missing doc part as discussed in #1704

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -468,16 +468,32 @@ component use a ``PreMount`` hook::
         {
             // validate data
             $resolver = new OptionsResolver();
+            $resolver->setIgnoreUndefined(true);
+
             $resolver->setDefaults(['type' => 'success']);
             $resolver->setAllowedValues('type', ['success', 'danger']);
             $resolver->setRequired('message');
             $resolver->setAllowedTypes('message', 'string');
 
-            return $resolver->resolve($data);
+            return $resolver->resolve($data) + $data;
         }
 
         // ...
     }
+
+.. note::
+
+   In its default configuration, the OptionsResolver treats all props.
+   However, if more props are passed than the options defined in the OptionsResolver, an error will be prompted, indicating that one or more options do not exist.
+   To avoid this, use the `ignoreUndefined()` method with `true`. See `ignore not defined options`_ for more info.
+   
+      $resolver->setIgnoreUndefined(true);
+   
+   The major drawback of this configuration is that the OptionsResolver will remove every non-defined option when resolving data. 
+   To maintain props that have not been defined within the OptionsResolver, combine the data from the hook with the resolved data.
+   
+      return $resolver->resolve($data) + $data;
+
 
 The data returned from ``preMount()`` will be used as the props for mounting.
 
@@ -1654,3 +1670,4 @@ https://symfony.com/doc/current/contributing/code/bc.html
 .. _`CVA (Class Variant Authority)`: https://cva.style/docs/getting-started/variants
 .. _`shadcn/ui`: https://ui.shadcn.com
 .. _`tales-from-a-dev/twig-tailwind-extra`: https://github.com/tales-from-a-dev/twig-tailwind-extra
+.. _`ignore not defined options`: https://symfony.com/doc/current/components/options_resolver.html#ignore-not-defined-options


src/TwigComponent/doc/index.rst
```diff
@@ -468,12 +468,14 @@ component use a ``PreMount`` hook::
...
             $resolver = new OptionsResolver();
+            $resolver->setIgnoreUndefined(true);
+
             $resolver->setDefaults(['type' => 'success']);
             $resolver->setAllowedValues('type', ['success', 'danger']);
             $resolver->setRequired('message');
             $resolver->setAllowedTypes('message', 'string');
 
-            return $resolver->resolve($data);
+            return $resolver->resolve($data) + $data;
```
cavasinf, line 478: Should we:
1. add code comment about why we do that?
2. add DOC paragraph about why we do that?
3. stay as is.

smnandre: I think i'd add a entire block of code then, and let this one as it is to not confuse anyone.

Something like 

- this block 
- bloc "info" or "tip" in rst with a short paragraph and just 3 lignes of code 

```php
            $resolver->setIgnoreUndefined(true);
            
            // ... 
            
            return $resolver->resolve($data) + $data;
```

What do you think ?

cavasinf: I've tried to be "short", trust me 😭

smnandre: No i mean it's perfect, but i think having a dedicated block would be more limpid!

src/TwigComponent/doc/index.rst
```diff
@@ -468,17 +468,33 @@ component use a ``PreMount`` hook::
...
         }
 
         // ...
     }
 
+.. note::
+
+   In its default configuration, the OptionsResolver treats all props. 
+   However, if more props are passed than the options defined in the OptionsResolver, an error will be prompted, indicating that one or more options are not defined. 
+   To avoid this, it is necessary to use the `ignoreUndefined()` method with `true`. See `ignore not defined options`_ for more info.
```
WebMamba:
```suggestion
   To avoid this, use the `ignoreUndefined()` method with `true`. See `ignore not defined options`_ for more info.
```

src/TwigComponent/doc/index.rst
```diff
@@ -468,17 +468,33 @@ component use a ``PreMount`` hook::
...
+   
+      $resolver->setIgnoreUndefined(true);
+   
+   The major drawback of this configuration is that the OptionsResolver will remove every non-defined option when resolving data.
+   In order to maintain the props that have not been defined within the option resolver, it is necessary to combine the data from the hook with the result of the resolved data.
```
WebMamba:
```suggestion
The major drawback of this configuration is that the OptionsResolver will remove every non-defined option when resolving data. To maintain props that have not been defined within the OptionsResolver, combine the data from the hook with the resolved data.
```

WebMamba (commented):
Minor: some sugestion to make syntax a bit shorter to improver readability

src/TwigComponent/doc/index.rst
cavasinf: Shorter indeed

src/TwigComponent/doc/index.rst
```diff
@@ -468,17 +468,32 @@ component use a ``PreMount`` hook::
...
+            return $resolver->resolve($data) + $data;
         }
 
         // ...
     }
 
+.. note::
+
+   In its default configuration, the OptionsResolver treats all props. 
+   However, if more props are passed than the options defined in the OptionsResolver, an error will be prompted, indicating that one or more options are not defined.
```
cavasinf: @WebMamba `are not defined` is ambiguous.. (not passed or not exist?)

I tend to do:
- `are undefined`
- `do not exist`

```suggestion
   However, if more props are passed than the options defined in the OptionsResolver, an error will be prompted, indicating that one or more options do not exist. 
```

WebMamba: I tend more for do not exist

WebMamba (approved):
Thanks @cavasinf! Documenting is key, and this PR make think much more clear!

kbond:
Excellent, thank you Florian!

PR #1874, opened 2024-05-29
Fix the example with Expression Language

seb-jean:
@stof said in https://github.com/symfony/symfony-docs/pull/17905: 
> When evaluating the expression, the ExpressionLanguage does not have access to private properties of the object, as it runs from its outside. And contrary to Twig, ExpressionLanguage does not have the magic . operator that tries to find a getter when it cannot use the property.

I then improved the documentation :).


kbond:
Thank you @seb-jean.

PR #1876, opened 2024-05-30
[LazyImage] Fix code block rendering in docs

mttsch:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/LazyImage/doc/index.rst
+++ b/src/LazyImage/doc/index.rst
@@ -109,9 +109,7 @@ Customizing images fetching
 By default, ``data_uri_thumbnail`` fetches images using the `file_get_contents`_ function.
 It works well for local files, but you may want to customize it to fetch images from a remote server, `Flysystem`_, etc.
 
-To do so you can create a invokable class, the first argument is the filename to fetch:
-
-.. ::
+To do so you can create a invokable class, the first argument is the filename to fetch::
 
     namespace App\BlurHash;
 


mttsch:
Currently renders incorrectly on website (https://symfony.com/bundles/ux-lazy-image/current/index.html#customizing-images-fetching):

Screenshot 2024-05-30 at 09 21 13

I did not check how it will render after this change!

Kocal:
Hi, I've used `::` because the linter was not happy about it:
> Please do not use ".. code-block:: php", use "::" instead.

And so you have the same issue https://github.com/symfony/ux/actions/runs/9298650239/job/25590944633?pr=1876 

IDK how to fix it properly tho

Kocal:
Hum, maybe removing the newline between `.. ::` and `namespace App\BlurHash;`?

mttsch:
When checking for existing examples, I mistakenly used examples from the vendor folder. Now I searched for existing examples that are not in vendor, though I still have not checked the built result.

kbond:
Thank you Matthias.

Kocal:
> When checking for existing examples, I mistakenly used examples from the vendor folder. Now I searched for existing examples that are not in vendor, though I still have not checked the built result.

Looks like you were right, the code block is now properly displayed, thanks :)
image

PR #1877, opened 2024-05-31
[TwigComponent] Disable profiler integration via configuration

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #... 
| License       | MIT

Add a new configuration key allowing to disable profiler integration (enabled per default in debug mode)

(following several performance issues caused by the current DataCollector implementation)


kbond:
Thanks Simon.

PR #1887, opened 2024-06-04
[Demo] Infinite Scroll 2/2 

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix 
| License       | MIT

This took some time but....  what a beautiful collection of TShirts 😸 

Capture d’écran 2024-06-04 à 20 40 14


seb-jean:
I just tested it, and it works perfectly! :clap:

ux.symfony.com/templates/demos/live_component/infinite_scroll_2.html.twig
```diff
@@ -0,0 +1,153 @@
...
+    </div>
+
+    <section class="py-5">
+        <div class="container container-xxl">
+            <div class="code-description text-center">
+                <h2>So... how does it work?</h2>
+                <p>
+                    Let's focus on two aspects. But first, <code>did you scroll</code> all the way to the end?
+                    <br>
+                    You might have missed some important details about performance... 🍿
```
WebMamba, line 40: Maybe we could explain the performance details you mention, I don't get what you are thinking about 🙃

ux.symfony.com/templates/demos/live_component/infinite_scroll_2.html.twig
```diff
@@ -0,0 +1,153 @@
...
+
+                This trick is very similar to the one used in the
+                [`Part 1/2`]({{ url('app_demo_live_component_infinite_scroll') }})
+                of this "Infinite Scroll" demo with LiveComponent.
+
+                ### Previous Results
+
+                🐯 For the previous page results, we add one fake item
+                corresponding to the last item of the previous page,
+                with the same `id`. This allows the **addition of new elements** after the existing ones.
```
WebMamba, line 66: Maybe we could go a little deeper here and explain what morphism is and what morphism does with this same id?
And I think this part should be after the observer part?

WebMamba (commented):
Thanks @smnandre, just tested locally it's working great!

ux.symfony.com/templates/demos/live_component/infinite_scroll_2.html.twig
smnandre: The "750ms delay" ... there is a pragraph at the end of the grid

ux.symfony.com/templates/demos/live_component/infinite_scroll_2.html.twig
smnandre: > Maybe we could go a little deeper here and explain what morphism is and what morphism does with this same id?

I think this would require a dedicated page (and i have the perfect slide for this remember ? 😆 )

smnandre: > And I think this part should be after the observer part?

Oh i remember now... we need to define why the div 'previous' / 'current' / 'next' at first, because elsewhen this is impossible to understand ... :/ 


To be honest i'm not entirely ok with the final result, and i really want to create something more interactive / steps by steps highlither, etc.. 

I'd say "this is ok" for this one, but i'm really trying a lot of things these days to "more"

ux.symfony.com/templates/components/ProductGrid2.html.twig
```diff
@@ -0,0 +1,97 @@
...
+                <div style="max-width: 36rem;">
+                    <h4>Did you know?</h4>
+                    <p>
+                        This demo adds a fixed delay of <code>750ms</code> before each page load to simulate slow
+                        network conditions. Or maybe just to let you enjoy the sweet animations... 🍿
+                    </p>
+                    <p>Wonder how it would feel without? Why not get the code and try it yourself?</p>
+                </div>
+            </div>
+        {% endif %}
```
smnandre, line 72: This is the performance thing @WebMamba  :)

kbond:
Looks great, thanks Simon!

Issue #1800, opened 2024-04-29
[UX-Icons] Allow Reuse


Jibbarth:
When an SVG has been added once in a page, it can be reused with the `<use xlink:href=#id>` later in the same page to avoid duplication of svg. 

It could be great if ux_icon would have a `allowReuse` parameters to transform icon to use a previously added. 

For example : 

```twig
    {{ ux_icon('fa6-brands:symfony', {}, true)}}
    {{ ux_icon('fa6-brands:symfony', {'style':'color:red'}, true)}}
```

Would produce following html 

```html
<div class="example-wrapper">
[...]

Issue #1767, opened 2024-04-22
[Icons] Alias for icons


Xbirdfr:
Hi !

I love the icon initiative, a very great ressource !

i'm comming with an idea, a thing that actually I use a lot in my custom projects: 

an alias system for icons
Ex: 

```
# config/packages/ux_icons.yaml
ux_icons:
    alias:
      - "home": "material-symbols:home"
```

```
# templates/base.html.twig
...
<a href="#">{{ ux_icon('home') }} Homepage </a>
...
```
this will ensure consistency of the icon throughout the project, and to be able to easily change / update it in one go

Or maybe a way to create a custom twig extension extending ux_icons already exist ?

kbond:
Hey @Xbirdfr, thanks for the idea! This is a bit similar to https://github.com/symfony/ux/pull/1596 where we've been discussing multiple icon registries and aliasing. That issue is more about including multiple directories of icons.

What you're thinking is using `ux_icon('home')` would lazy load (on-demand) `material-symbols:home`?

Xbirdfr:
Hi @kbond ,

not necessarily this particular icon, but rather an alias:icon format

like today I want my alias "home" to refer to "material-symbols:home",
but tomorrow, I think the "solar:home-outline" will better fit my app, so I juste change the alias from :
```
# config/packages/ux_icons.yaml
ux_icons:
    alias:
      - "home": "material-symbols:home"
```

to

```
# config/packages/ux_icons.yaml
ux_icons:
    alias:
      - "home": "solar:home-outlinee"
```

refresh the cache and voila

Kocal:
What about using symbolic links instead? `assets/icons/home.svg` can refers to `assets/icons/solar/home-outline.svg`

kbond:
> What about using symbolinks instead?

That could work today for sure.

Xbirdfr:
> What about using symbolinks instead? `assets/icons/home.svg` can refers to `assets/icons/solar/home-outline.svg`

Yes, nice solution 👌

But I think a file based / config based way will ne more stable

smnandre:
> But I think a file based / config based way will ne more stable

We will add several configuration parameters and "namespace aliases" are one of my top priorities, as all icon sets are not equal currently on a DX perspective (typing `bi:heart` VS `material-icon-foo-bar:heart` is not _fair_)

We've been very busy with support / minor bugs / documentation recently but we're going to work on Icon very very soon :)

kbond:
[...]

PR #1889, opened 2024-06-05
[Icons] Expose IconRendererInterface

smnandre:
As the renderer won't change without BC (at least via the twig function/component).. i think it's ok to expose an interface here, allowing people to use it / decorate / etc..

Only thing i think we don't want (for now): add other public methods (like exposing the Icon object)

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #... 
| License       | MIT

### DIFF
--- a/src/Icons/src/IconRenderer.php
+++ b/src/Icons/src/IconRenderer.php
@@ -16,7 +16,7 @@ namespace Symfony\UX\Icons;
  *
  * @internal
  */
-final class IconRenderer
+final class IconRenderer implements IconRendererInterface
 {
     public function __construct(
         private readonly IconRegistryInterface $registry,
@@ -32,8 +32,6 @@ final class IconRenderer
      *
      * Precedence order:
      *   Icon file < Renderer configuration < Renderer invocation
-     *
-     * @param array<string,string|bool> $attributes
      */
     public function renderIcon(string $name, array $attributes = []): string
     {
--- a/src/Icons/config/services.php
+++ b/src/Icons/config/services.php
@@ -50,6 +50,8 @@ return static function (ContainerConfigurator $container): void {
             ])
             ->tag('twig.runtime')
 
+        ->alias('Symfony\UX\Icons\IconRendererInterface', '.ux_icons.icon_renderer')
+
         ->set('.ux_icons.icon_finder', IconFinder::class)
             ->args([
                 service('twig'),
--- a/src/Icons/src/IconRendererInterface.php
+++ b/src/Icons/src/IconRendererInterface.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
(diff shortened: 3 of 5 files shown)


WebMamba (commented):
I am not sure about this? Why do you think we need an interface? We don't have any plant for other IconRenderer? And why user would need to create their own Renderer? I think this a bit to early for this abstraction

smnandre:
> Why do you think we need an interface? 

I prefer exposing an interface than a Twig Extension + TwigComponent to be fair ... it's more future-proof. And the contract is already ensured so that would not change a thing i guess.

> We don't have any plant for other IconRenderer? 

No, same for the Router, Serializer, FormBuilder, etc.. it makes testing and autowiring easier for users :) 

> And why user would need to create their own Renderer? 

Not create their own... but decorate :) I personnaly would like to do it easily (as i prefer to wrap my icons in span for instance)

> I think this a bit to early for this abstraction

Again, as we offer a form of contract with the TwigExtension, i think we should make this official (i mean, only one method is not much ^^)

kbond:
In my packages, I'd make `IconRenderer` autowireable but in Symfony core, I believe interfaces are preferred, even if there's only 1 implementation. It's easier for someone to decorate.

@smnandre, is an implementation that handles aliases in config (like requested in #1767) what we'd add? I'd like to implement #1800 at some point, does this interface prevent this?

smnandre:
Nope it does not !

Alias/config handling is on the way (started on one of my branches) and the interface is perfect to abstract this implementation i guess 

Regardind SVG symbol / sprites / reuses / etc... It can be either coded via this interface or a new explicit method call but nothing blocking here

kbond (approved):
If we already have a need for an alternate implementation/decorator, I think the interface makes sense.

Making this autowireable allows rendering icons outside of twig - like provide the icon string in a json endpoint.

> And why user would need to create their own Renderer?

I guess if the user wanted to take over the entire system this would be the place.

src/Icons/src/IconRenderer.php
```diff
@@ -16,7 +16,7 @@
  *
  * @internal
  */
-final class IconRenderer
+final class IconRenderer implements IconRendererInterface
```
kbond, line 19:
```suggestion
final class RegistryIconRenderer implements IconRendererInterface
```

This could be done in the follow-up PR for #1767.

smnandre: So i revert the change, or i commit your suggestion ?

kbond: We can leave as is

kbond:
@WebMamba, have we adequately addressed your concerns?

smnandre:
@WebMamba  ?

javiereguiluz:
Let's merge this one because the given arguments were compelling.

Thanks @smnandre!

PR #1899, opened 2024-06-08
[TwigComponent] Add note in doc about composition that component live in their own context

### LINKED ISSUE #1875: [TwigComponent] Parent function does not work in the content of a twig component
I tried the following stupid code:

```twig
{% block choice_row %}
    <twig:Select label="{{ form.vars.label|default(name|humanize)|trans }}">
        {{ parent() }}
    </twig:Select>
{% endblock %}
```

But the `parent()` function does not work inside a twig component. 😅 It was a bit disturbing. Not sure it's a bug or a feature request to be honest.

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #1875
| License       | MIT

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -337,6 +337,19 @@ You can even give the block default content. See
 :ref:`Passing HTML to Components via Block <embedded-components>`
 for more info.
 
+Composition
+~~~~~~~~~~~
+
+One of the most interesting things to do when working with components is composition.
+
+.. code-block:: html+twig
+
+    <twig:Alert>
+        <twig:Icon name="success"/> Nice work!
+    </twig:Alert>
+
+You can have components inside your components! This is great! However, be aware that every component lives in its own context. This means that your child component does not know that it is inside a parent component; you cannot use any method like parent() inside the child component to access the parent. You should keep your components independent of the context. This rule is really important to respect, as it will guarantee that your components will always work the same way regardless of where you use them. If you need to access some parent data, you should pass properties down to your child or use events.
+
 Fetching Services
 -----------------
 
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -337,18 +337,32 @@ You can even give the block default content. See
 :ref:`Passing HTML to Components via Block <embedded-components>`
 for more info.
 
-Composition
-~~~~~~~~~~~
+Composition vs Inheritance
+~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-One of the most interesting things to do when working with components is composition.
+Symfony tend to follow the best pratice of the component architecture, so it's highly recommended to use Composition instead of Inheritance to reuse code betwen component (see: https://legacy.reactjs.org/docs/composition-vs-inheritance.html).
+
+Let's take an exemple. Some component don't know their children ahead of time. This is common for components like Alert or Dialog that represent generic “boxes”. We recommend you
+to use the block content to pass children directly into their output :
+
+.. code-block:: html+twig
+
+    <div class="alert">
+       {% block content %}
+       {% endblock %}
+    </div>
+
+This lets other components pass arbitrary children to them by nesting the elements :
 
 .. code-block:: html+twig
 
     <twig:Alert>
-        <twig:Icon name="success"/> Nice work!
+        <twig:Icon name="success"/> 
+        <p>Nice work!</p>
     </twig:Alert>
 
-You can have components inside your components! This is great! However, be aware that every component lives in its own context. This means that your child component does not know that it is inside a parent component; you cannot use any method like parent() inside the child component to access the parent. You should keep your components independent of the context. This rule is really important to respect, as it will guarantee that your components will always work the same way regardless of where you use them. If you need to access some parent data, you should pass properties down to your child or use events.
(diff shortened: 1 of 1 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -337,6 +337,19 @@ You can even give the block default content. See
...
+
+One of the most interesting things to do when working with components is composition.
+
+.. code-block:: html+twig
+
+    <twig:Alert>
+        <twig:Icon name="success"/> Nice work!
+    </twig:Alert>
+
+You can have components inside your components! This is great! However, be aware that every component lives in its own context. This means that your child component does not know that it is inside a parent component; you cannot use any method like parent() inside the child component to access the parent. You should keep your components independent of the context. This rule is really important to respect, as it will guarantee that your components will always work the same way regardless of where you use them. If you need to access some parent data, you should pass properties down to your child or use events.
```
smnandre: Nice ! 

Two small comments:
- maybe split in two lines, to emphasis some words, and improve even more readability ? 
- i think we want to add a bit of nuance here or you may frighten people 😆 

Instead of 
> You should keep your components independent of the context 

Maybe 

> You should try to keep (...)

or  

> It is generally recommanded to ...

Same for the last sentence ?

Nek-: It's more like `you must`. Actually using the context is just unsupported. And I slightly disagree with the tone of this piece of documentation that suggest composition to explain inheritance does not work.

IMO it should be named something like `Composition enforced` if you want to get people straight to your point 😉 .

smnandre: In fact it's a bit more complicated than that.

There is a difference between _anonymous_ and _class-based_ components, there is a difference betwen "`include`" and "`embed`" components, and i don't even start with `live` ones haha

But this works perfectly

```twig

    {% set aa = 'AA' %}

    <twig:Browser>

        {{ dump(aa) }}

    </twig:Browser>

```

Capture d’écran 2024-06-09 à 14 18 56


A component should not know an outside one existence.

There it's something else :)

Nek-: Ok, it just feels weird to me to have an example unrelated to the comment following then.

smnandre: My point was only that I prefer to avoid binary formulations, as things in life are rarely that simple, and few things in those components are that clear-cut. :)

WebMamba:
Hey @smnandre and @Nek- I completely changed the strategy of this PR, and the ton! Tell me if you think this is better now! 😁

src/TwigComponent/doc/index.rst
```diff
@@ -670,6 +697,10 @@ You can also add more, named blocks:
         </twig:block>
     </twig:Alert>
 
+.. note::
+
+    Because your TwigComponent live inside his own specific context, you can't use the parent function inside the block of your component template.
+
```
smnandre, line 703: This is not true :) 

https://symfony.com/bundles/ux-twig-component/current/index.html#passing-html-to-components-via-blocks

src/TwigComponent/doc/index.rst
```diff
@@ -337,6 +337,33 @@ You can even give the block default content. See
...
+.. code-block:: html+twig
+
+    <twig:Alert>
+        <twig:Icon name="success"/> 
+        <p>Nice work!</p>
+    </twig:Alert>
+
+Anything inside the <twig:Alert>  tag gets passed into the Alert component as a content block. Since Alert renders {% block content %} inside a <div>, the passed elements appear in the final output.
+You can go deeper on the usage of block, on this part of the :ref:`documentation <embedded-components>`
+
```
smnandre, line 366: Does this require a new section ? I feel this is more or less the same as what's already explained in the [Passing HTML to Components Via Blocks](https://symfony.com/bundles/ux-twig-component/current/index.html#passing-html-to-components-via-blocks) section, or did i missread something ?

src/TwigComponent/doc/index.rst
```diff
@@ -337,6 +337,33 @@ You can even give the block default content. See
 :ref:`Passing HTML to Components via Block <embedded-components>`
 for more info.
 
+Composition vs Inheritance
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Symfony tend to follow the best pratice of the component architecture, so it's highly recommended to use Composition instead of Inheritance to reuse code betwen component (see: https://legacy.reactjs.org/docs/composition-vs-inheritance.html).
```
smnandre, line 343: This link should use the RST syntax or the doctor ill not be happy :)

src/TwigComponent/doc/index.rst
WebMamba: Maybe my formulation is bad but I was trying to explain the issue @Nek- had. How to you explain that in a few lines?

src/TwigComponent/doc/index.rst
WebMamba: I think so. I feel like this is one of the biggest traps people are falling into. This part is more about what to choose, and the block part is more about how to do it. 
I tried to add more guidelines to the doc

src/TwigComponent/doc/index.rst
Nek-: You should maybe say that parent refers to its own context instead. Adding that it acts like embed element may also be a nice addition.

Kocal:

Thanks!

Status: Needs Work

PR #1893, opened 2024-06-06
Revert #1694 "[LiveComponent] set LiveArg value to null if empty string"

smnandre:
This reverts commit 538cc61f9fa5858b423ec7e341d909c718d126de. temporarely (we need to release a version)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1694
| License       | MIT

Let's see...

PR #1901, opened 2024-06-09
[Live] Fix nullable DateTime hydration with empty value

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

Currently when we send an empty string as value, the `new DateTime('')` call
creates an object with the current time, which I think is not correct and the
object should be set to null instead.

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -531,6 +531,10 @@ final class LiveComponentHydrator
                 throw new BadRequestHttpException(sprintf('The model path "%s" was sent an invalid data type "%s" for a date.', $propertyPathForError, get_debug_type($value)));
             }
 
+            if ('' === $value) {
+                return $allowsNull ? null : throw new BadRequestHttpException(sprintf('The model path "%s" was sent invalid date data "%s". Setting empty date value is not allowed, send a non-empty value or make the property nullable.', $propertyPathForError, $value, $dateFormat)) ;
+            }
+
             if (null !== $dateFormat) {
                 return $className::createFromFormat($dateFormat, $value) ?: throw new BadRequestHttpException(sprintf('The model path "%s" was sent invalid date data "%s" or in an invalid format. Make sure it\'s a valid date and it matches the expected format "%s".', $propertyPathForError, $value, $dateFormat));
             }
--- a/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
+++ b/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
@@ -307,6 +307,21 @@ final class LiveComponentHydratorTest extends KernelTestCase
             ;
         }];
 
+        yield 'empty DateTime hydrates correctly' => [function () {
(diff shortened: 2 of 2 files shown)


src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -531,6 +531,10 @@ private function hydrateObjectValue(mixed $value, string $className, bool $allow
                 throw new BadRequestHttpException(sprintf('The model path "%s" was sent an invalid data type "%s" for a date.', $propertyPathForError, get_debug_type($value)));
             }
 
+            if ('' === $value) {
+                return $allowsNull ? null : throw new BadRequestHttpException(sprintf('The model path "%s" was sent invalid date data "%s". Setting empty date value is not allowed, send a non-empty value or make the property nullable.', $propertyPathForError, $value, $dateFormat)) ;
+            }
+
```
WebMamba, line 537: We can't do that this is a BC break. If we want to implement this fix we have to add an option to activate or not your fix and to have it not activate by default.

WebMamba (commented):
Yes it sound better for me to have it as null for empty string. Let's wait for what the others thing about it. Thanks for your PR @1ed 😁

smnandre:
I'm pretty sure this would have been fixed with the [LiveArg implementation](https://github.com/symfony/ux/pull/1694) we almost sent in 2.18. 
At least, it was the same root cause: empty string data handle.

But as we realized it has a minor bug, we had to [revert](https://github.com/symfony/ux/pull/1893) it.

So I'd prefer that we find a more generic way to fix these cases...  or i fear every specific type issue will need specific patches, with less coherence and a lot of similar code in various places.

Would you like to have a look at it ?

smnandre:
@1ed did you have time to read the other PR i'd be curious to have your opinion... 

They may not but they are really due to the same thing 
(empty data-attribute or input in HTML sending "" and so interpreted as empty strings in PHP)

Maybe you would have a fresh look / opinion on that matter ?

jannes-io:
@smnandre instead of reverting #1694 , we could look at implementing this:

> Perhaps in the supports method of LiveArgValueResolverTrait we could, after confirming that it's a LiveArg, move the value from _component_action_args to just the attributes but not return any values if the type is expected to be an object. Then the doctrine MapEntity will run, and the value will be where MapEntity expects it to be, and resolves the argument.

smnandre:
We had no choice but to revert it, sadly, as it created bugs in rare occasion.

i’m 100% open to any solution :)

jannes-io:
> We had no choice but to revert it, sadly, as it created bugs in rare occasion.

are the bugs documented somewhere or could we add tests that cover current behaviour (other than the ones we already had)? I could take out the bugs but I’d have to be aware of the current situation 😅

smnandre:
@jannes-io No, still [this problem](https://github.com/symfony/ux/pull/1694#issuecomment-2124931327) with Doctrine / other valueresolvers, but this was enough to _not_ release it like that  😅

Kocal:

Thanks!

Status: Needs Work

PR #1904, opened 2024-06-10
[Translator] Fix changing dump directory using AssetMapper

### LINKED ISSUE #1642: [Translator] Move the `var/translations` folder in the `build_dir`
Currently, UX Translator bundle uses the `var/translations` directory 

> When warming up the Symfony cache, all of your translations will be dumped as JavaScript into the var/translations/ directory. For a better developer experience, TypeScript types definitions are also generated aside those JavaScript files.

https://symfony.com/bundles/ux-translator/current/index.html#usage

# Directory not found / not writable

But this directory is no standardized, and is not "buildable" (as the ... build directory).

And... this directory is **not** created by the the [Kernel](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Component/HttpKernel/Kernel.php) (it does it for `cache_dir` and `build_dir`)

It may be then created during cache:warming but for most of the standard deployment tools, any folder not explicitely marked as writable is not..

And so this happens:  (folder not writable)

```console
W: + mkdir var/translations
W: mkdir: cannot create directory ‘var/translations’: File exists
```

And this happens (asset not found)

```console
  W: 11:36:57 WARNING   [asset_mapper] Unable to find asset "../translator" imported from "/app/assets/controllers/translator-demo-controller.js". Try adding ".js" to the end of the import - i.e. "../translator.js".
  ```
 
  
# Use the build dir  ?
  
For me the best choice would be to use the `build_dir`.

There are other individual solutions, but the build dir is made for this and would remove any future problem too. 
  
So my question is...  ..... can we do it or is this a BC ?   And..... how ? (as it's currently in the importmap of people)

### PR DESCRIPTION
**Pull request by maelanleborgne**
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1642
| License       | MIT

The configuration node `ux_translator.dump_directory` wasn't correctly taken into account when using UxTranslator with AssetMapper.  
This PR fixes the configuration prepend of AssetMapper to use the custom path + adds a "caution" section in the docs to remind developers to also update their `importmap.php`.

### DIFF
--- a/src/Translator/doc/index.rst
+++ b/src/Translator/doc/index.rst
@@ -138,6 +138,12 @@ file. 2 of the new items are::
         'path' => 'var/translations/configuration.js',
     ],
 
+.. caution::
+
+    If you change the ``dump_directory`` in your configuration file, you will need to
+    replace the default ``var/translations/***`` with your new path in the
+    ``importmap.php`` file.
+
 These are then imported in your ``assets/translator.js`` file. This setup is
 very similar to working with WebpackEncore. However, the ``var/translations/index.js``
 file contains *every* translation in your app, which is not ideal for production
--- a/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
+++ b/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
@@ -43,12 +43,12 @@ class UxTranslatorExtension extends Extension implements PrependExtensionInterfa
         if (!$this->isAssetMapperAvailable($container)) {
             return;
         }
-
+        $config = $container->getExtensionConfig('ux_translator')[0];
         $container->prependExtensionConfig('framework', [
             'asset_mapper' => [
                 'paths' => [
                     __DIR__.'/../../assets/dist' => '@symfony/ux-translator',
-                    '%kernel.project_dir%/var/translations' => 'var/translations',
+                    $config['dump_directory'] => '@app/translations',
                 ],
             ],
         ]);

### CODE REVIEW

**Review by Kocal — APPROVED** [review]
Good catch!

File: src/Translator/doc/index.rst [hunk 1]
```diff
@@ -138,6 +138,12 @@ file. 2 of the new items are::
         'path' => 'var/translations/configuration.js',
     ],
 
+.. caution::
+
+    If you change the ``dump_directory`` in your configuration file, you will need to
```
WebMamba, line 143: And add an: if you use AssetMapper.

src/Translator/src/DependencyInjection/UxTranslatorExtension.php
```diff
@@ -43,12 +43,12 @@ public function prepend(ContainerBuilder $container)
...
             return;
         }
-
+        $config = $container->getExtensionConfig('ux_translator')[0];
         $container->prependExtensionConfig('framework', [
             'asset_mapper' => [
                 'paths' => [
                     __DIR__.'/../../assets/dist' => '@symfony/ux-translator',
-                    '%kernel.project_dir%/var/translations' => 'var/translations',
+                    $config['dump_directory'] => '@app/translations',
```
WebMamba, line 51: Maybe we should add a check that the dum_directory config exists? Or are we 100% sure that this conf exist

WebMamba (commented):
Nice thanks @maelanleborgne 😁

src/Translator/src/DependencyInjection/UxTranslatorExtension.php
Kocal: The directory is created during the cache warmup

src/Translator/doc/index.rst
smnandre: It is already in the "Using with AssetMapper" so i'd say we don't need here

src/Translator/src/DependencyInjection/UxTranslatorExtension.php
smnandre, line 51: Could you add a functionnal test for this ? That will prevent any future regression

maelanleborgne: Not sure how to test for this. I don't think I can access the config, and I can't functionally test this without requiring asset mapper in require-dev, which is not wanted. If you have any input on how I can retrieve the config from the test scope I'd be glad to hear about it :)

maelanleborgne: And the configuration is defined with a default value so it should always be set

kbond:
Thank you @maelanleborgne.

smnandre:
Well....  bad week for UX   🐞 🐞 🐞  

Here is my importmap

Capture d’écran 2024-06-12 à 06 01 13

I did a composer update and...

image


To fix i just needed to replace my "var/" by  "@app" links ... (and then it seems cool thanks @maelanleborgne)
Capture d’écran 2024-06-12 à 06 03 43

But we cannot do this like that//// or this will crash for everyone (with no clue/help how to fix)

So can we find something today.. or should we revert before the complaints start ? 😅

maelanleborgne:
Seems like this is caused by a small misleading info from the doc on what lines to add in `importmap.php` :
```php
'@app/translations' => [
    'path' => 'var/translations/index.js',
],
'@app/translations/configuration' => [
    'path' => 'var/translations/configuration.js',
],
```
but importmap usually uses relative paths
```php
'@app/translations' => [
    'path' => './var/translations/index.js',
],
'@app/translations/configuration' => [
    'path' => './var/translations/configuration.js',
],
```

This worked because the associated 'namespace' in `asset_mapper.paths` was a hardcoded `var/translations`.  
Now with this PR changing the namespace to a more generic one, `var/translations` no longer resolves to `%kernel.project_dir%/var/translations`.
Preventing a BC here is quite tough, I see only two solutions but both are bad : updating `importmap.php` on the fly, or keeping a reference to the dump_dir with 'var/translations` as a namespace
```php
$container->prependExtensionConfig('framework', [
    'asset_mapper' => [
        'paths' => [
            __DIR__.'/../../assets/dist' => '@symfony/ux-translator',
            $config['dump_directory'] => '@app/translations',
            $config['dump_directory'] => 'var/translations', // or '%kernel.project_dir%/var/translations' => 'var/translations'
        ],
    ],
```

Anyway, we can revert for now, I'll go fix the doc in the meantime.

src/Translator/src/DependencyInjection/UxTranslatorExtension.php
kbond: I shouldn't have ignored this conversation 😭

PR #1906, opened 2024-06-10
[LiveComponent] improve errors when LiveArg is missing for live action and live listener

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix 
| License       | MIT

This PR improves the error message when a LIveArg is missing on live action or live event. Here is the new error and the previous one.

Capture d’écran 2024-06-10 à 16 43 38
Capture d’écran 2024-06-10 à 16 44 19

### DIFF
--- a/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
@@ -28,6 +28,7 @@ use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
 use Symfony\Component\Security\Csrf\CsrfToken;
 use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 use Symfony\Contracts\Service\ServiceSubscriberInterface;
+use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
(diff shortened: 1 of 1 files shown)


src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -274,6 +275,14 @@ public function onKernelException(ExceptionEvent $event): void
             return;
         }
 
+        if ($event->getThrowable() instanceof RuntimeException) {
+            if (str_starts_with($event->getThrowable()->getMessage(), 'Could not resolve argument $key')) {
```
smnandre, line 279: Can we use a class/interface here?

smnandre:
What will be displayed if the LiveComponent is not tagged with controller_arguments (or does not extends AbstractController) ?

smnandre:
Hmm ok ... 

It's because of the NotTaggedControllerValueResolver registered only in debug.

We should see this message instead, already better.
Capture d’écran 2024-06-14 à 22 45 01

And, guess what..... all this is related to our LiveArg implementation 🤖 

So i suggest we take some time (when we can) to make a good, stable & durable fix for LiveArg 

.. and we won't have this problem anymore :à

Kocal (changes requested):
Checking on the exception message is a bit fragile, can't we check on the exception class name or where the extension has been thrown?

Also, tests are missing.

Kocal:

PR #1907, opened 2024-06-10
[Site] Social metadata

smnandre:
Add  FB/Twitter metadata, link the twitter acount and update demo images.

Social images for ackages will be done later

--

TODO After
- [x] bigger social demo images (1200x675) -> will oversize them via Pixelmator
- [ ] one social image per package (on it)


kbond:
Thanks Simon.

PR #1910, opened 2024-06-11
[Docs][Stimulus] use camelCase for target naming

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Hello! 

I had some trouble following the doc for use `stimulus_target` and how to use them inside my stimulus controller. 

As we can see [on the stimulus target doc](https://stimulus.hotwired.dev/reference/targets#naming-conventions), the value of target should be in camelCase as it's mapped in property in the controller. 

IE 
```
<div data-my-controller-target="myElement">
```

```
import { Controller } from "@hotwired/stimulus"

  static targets = [ "myElement"]
  // …
  foo() {
	this.myElementTarget.innerHTML = "...";
  }
}
```

So in this PR, I changed examples for `stimulus_target` usage.

### DIFF
--- a/src/StimulusBundle/doc/index.rst
+++ b/src/StimulusBundle/doc/index.rst
@@ -337,24 +337,24 @@ For example:
 
 .. code-block:: html+twig
 
-    <div {{ stimulus_target('controller', 'a-target') }}>Hello</div>
-    <div {{ stimulus_target('controller', 'a-target second-target') }}>Hello</div>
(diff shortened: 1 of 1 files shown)


smnandre (approved):

kbond:
Thank you Jibé.

PR #1911, opened 2024-06-11
[TwigComponent][Fix] Revert throwing exception when public property is unitialized

### LINKED ISSUE #1909: [Twig Component] BC break in latest release (2.18)?
Hey,

Lately symfony ux 2.18 was released and it appears to include a breaking change in it:

> Throws if exposed public prop is uninitialized by @smnandre in https://github.com/symfony/ux/pull/1780

`Error rendering "FooComponent" component: Cannot expose uninitialized property "$prop" from "App\Twig\Components\FooComponent".`

Luckily we only had 2 places that had issue by doing the update, but it was working fine before. It was possible to check for `is defined` or use the `??` operator to compensate for a non-initialized value. Even tho I see why this change makes sense, but having to deal with it in a minor version update feels not fully right, so I wanted to open this discussion if that is something that should be addressed or not :) 

No matter what, thank you for all your efforts on moving forward with Symfony UX, thanks for your effort and passion <3

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #1909
| License       | MIT

Revert #1780

If someone want to give another go at the original problem i'd appreciate.

### DIFF
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -142,11 +142,7 @@ final class ComponentRenderer implements ComponentRendererInterface
     private function exposedVariables(object $component, bool $exposePublicProps): \Iterator
     {
         if ($exposePublicProps) {
-            $publicProps = get_object_vars($component);
-            if ($uninitializedProps = array_diff_key(get_class_vars($component::class), $publicProps)) {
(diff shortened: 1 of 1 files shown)


kbond:
Thanks Simon.

PR #1913, opened 2024-06-12
[Translator] Revert Fix changing dump directory using AssetMapper

maelanleborgne:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Issues        | https://github.com/symfony/ux/pull/1904#issuecomment-2162066941
| License       | MIT

[This PR ](https://github.com/symfony/ux/pull/1904) introduced a bug : when using 'var/translations' in the `importmap.php` file instead of a relative path, it would break the twig `importmap()` function.

~~This PR aims to rectify this regression by keeping a 'var/translations' namespace pointing to the dump_directory. If this is not accepted, the previous PR should be reverted to prevent users using 'var/translations' in there `importmap.php` to face this issue, but the documentation should still be updated to display an example using the relative path instead of a misleading namespace.~~

This PR reverts the change. A word in the doc to say that the `dump_directory` conf is not compatible with asset_mapper would be nice.

### DIFF
--- a/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
+++ b/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
@@ -43,12 +43,12 @@ class UxTranslatorExtension extends Extension implements PrependExtensionInterfa
         if (!$this->isAssetMapperAvailable($container)) {
             return;
         }
-        $config = $container->getExtensionConfig('ux_translator')[0];
+
         $container->prependExtensionConfig('framework', [
             'asset_mapper' => [
                 'paths' => [
                     __DIR__.'/../../assets/dist' => '@symfony/ux-translator',
-                    $config['dump_directory'] => '@app/translations',
+                    '%kernel.project_dir%/var/translations' => 'var/translations',
                 ],
             ],
         ]);
--- a/src/Translator/doc/index.rst
+++ b/src/Translator/doc/index.rst
@@ -138,12 +138,6 @@ file. 2 of the new items are::
         'path' => 'var/translations/configuration.js',
     ],
 
-.. caution::
-
-    If you change the ``dump_directory`` in your configuration file, you will need to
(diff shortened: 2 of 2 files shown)


src/Translator/src/DependencyInjection/UxTranslatorExtension.php
```diff
@@ -43,11 +43,12 @@ public function prepend(ContainerBuilder $container)
...
             return;
         }
-        $config = $container->getExtensionConfig('ux_translator')[0];
+        $config = $this->processConfiguration(new Configuration(), $container->getExtensionConfig('ux_translator'));
         $container->prependExtensionConfig('framework', [
             'asset_mapper' => [
                 'paths' => [
                     __DIR__.'/../../assets/dist' => '@symfony/ux-translator',
+                    $config['dump_directory'] => 'var/translations', // Keeping for BC, see https://github.com/symfony/ux/pull/1904#issuecomment-2162066941
                     $config['dump_directory'] => '@app/translations',
```
Kocal: You  will only have `'var/translations'` or `'@app/translations'` in the array, since the key is the same for both values

maelanleborgne: You're right, when we tested it worked because we had it the other way around.  
There's no way to fix this because even with 
```php
asset_mapper' => [
                'paths' => [
                    __DIR__.'/../../assets/dist' => '@symfony/ux-translator',
                    $config['dump_directory'] => '@app/translations',
                    '%kernel.project_dir%/var/translations' => 'var/translations',
                ],
            ],
```
the keys will be the same when running the default conf.

smnandre (approved):
Yep and it happens don't worry... 

(some even say only the best of the best do this 😇  😆 )

... and we should have been more attentive.

Thank you :)

kbond:
Thank you @maelanleborgne.

Issue #1891, opened 2024-06-06
[LiveComponent] Allow disabling the error modal in production


YummYume:
Hello :wave:

Live components have a [great error modal](https://github.com/symfony/ux/blob/2c360c6bc95a035b46f99f2038f844311571a1c6/src/LiveComponent/assets/src/Component/index.ts#L454) when something goes wrong on the server, however, it currently displays even in production, and I believe it's often not a wanted behaviour to display an error. I'm aware that it's possible to use the `response:error` event to stop the modal from rendering, but it would be great if we could do that globally (for all live components at once). Should it be a good idea to dispatch the error event on the window as well for that purpose ? It could then be easy to dispatch a toast each time an error occurs.

Another issue with the current handling of errors is that loading states are not removed, which means that, for example, a disabled submit button will stay disabled. This is explained in #1463. I believe this can be easily fixed by triggering the `loading.state:finished` event after [handling the error](https://github.com/symfony/ux/blob/2c360c6bc95a035b46f99f2038f844311571a1c6/src/LiveComponent/assets/src/Component/index.ts#L310). What do you all think ?

PS : The modal could use some accessibility improvements, I'm up for a PR

smnandre:
Hey @YummYume ! 

What a nice, complete bug report :) 

I agree with your vision and i'd go even further: i believe we should not only "allow disabling"  but "force disable" when `kernel.debug !==true`


I'd probably split your ideas in 3 distinct PR
* one handling the visibility/usage of the error modal depending on a configuration param ( = kernel.debug per default)
* one related to event (but i do think this is already possible... but maybe not 😅 )
* one related to fixing the loading states and stuff

This last one could even include some debug/log capacities depending on env too ? 

So if you want to start one, two or the three, any help would be very welcomed and appreciated :)

And don't hesitate to start with minimal one / poc _if you want_ and we can iterate from this !

YummYume:
Hi !

I agree that it should definitely not be present in production at all, and people should be "forced" to add their own error handling. At the very least, the modal shouldn't be present in production (and probably not tests either ?), and the component should still work, by simply removing the loaders and keeping the same state as before. It would maybe even be a good option to add a `data-error`, to show something only when an error occurs ? It could be an interesting idea.

I'll gladly start working on the third point you mentioned, which is to remove loading states (which is crucial in my opinion). By debug/log, you mean in the browser's console ? I'll also check to see if it's possible to catch the event globally, but I'm not sure since I think the events are not native JS events.

[...]

PR #1916, opened 2024-06-12
[LiveComponent][Live] Add `data-error` attribute + fix loading behaviour on error

YummYume:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        | #1891
| License       | MIT

This PR adds a new `LoadingPlugin` to live components. It allows the use of a `data-error` that works almost like the `data-loading` attribute, but for errors. It also fixes the loading state not clearing when an error occurs, and a typo in the hook names in the types.

Little demo of the before/after of this PR :

Before :

https://github.com/symfony/ux/assets/61044908/64d4f70c-773c-4279-88ee-a39662ecd97c

After :

https://github.com/symfony/ux/assets/61044908/f729ae13-79f2-40e4-93c2-d40361c9e652


smnandre:
Thank you very much @YummYume ! I'll review in depth during week-end

A first comment though: i'm not sure we should expose the same API / possibilities, as such errors "should not" happen and we don't want to maintain/document too many things if they are not used.

So maybe could we select only one or two methods ?  What do you think ?

Regarding the attribute, `data-error` maybe is a bit too common, especially with the added CSS.
Maybe `data-live-error` ? 

(The documentation seems great, i think i'd inverse the 2 sections and start with the "what" before the "how")

src/LiveComponent/assets/dist/Component/index.d.ts
```diff
@@ -13,8 +13,8 @@ export type ComponentHooks = {
     'response:error': (backendResponse: BackendResponse, controls: {
         displayError: boolean;
     }) => MaybePromise;
-    'loading.state.started': (element: HTMLElement, request: BackendRequest) => MaybePromise;
-    'loading.state.finished': (element: HTMLElement) => MaybePromise;
+    'loading.state:started': (element: HTMLElement, request: BackendRequest) => MaybePromise;
+    'loading.state:finished': (element: HTMLElement) => MaybePromise;
```
smnandre: Took me some time to understand this ... as you have some CS changes mixed in here :) 

Could you -if you have time- open a PR just for this "bug fix" ? 

I know it's only TS and not _really_ a bug, but people will probably see these values in their IDE... and i'd really like them to see the good ones instead 😅  

And also to ensure we merge the fix asap .... because this one will depend on the path we choose to take :)

YummYume:
@smnandre Thanks for the feedback! :smile:  By that, you mean only allowing the `show|hide` directives ? I currently implemented the same directives as the `data-loading` attribute because they were quite easy to add, but didn't add support for any modifier (like delay, models, and such...) as it wouldn't make much sense for an error state.

And I thought the same about the `data-error` attribute, I just wasn't sure if I should try to align its name with the `data-loading` one (since they are very similar). I will change it. Good point for the documentation as well !

src/LiveComponent/assets/dist/Component/index.d.ts
YummYume: Will do! I just added the typing to the hooks to not have to remember the event names and having a better DX, but I was wondering why my event wasn't firing... Only to find out there was a typo in the typing :sweat_smile:

WebMamba (commented):
Wohaaaa! Thanks @YummYume for your PR! It's so cool! I love the path taking here!

I noticed that the data-live-error="addClass(text-red-500)" does not work. I didn't dig to see why but here is my template : 
```twig
<div {{ attributes }}>
    <p data-live-error="addClass(text-red-500)">Hello</p>
    <button
            data-action="live#action"
            data-live-action-param="error"
    >
        Throw
    </button>
    <span data-live-error="show">An error occurred ! Please try again.</span>
</div>
```

the class text-red-500 is not added but the span is well added. Maybe you could have a clue.

I think we should disable the pop up we the error by default in production, and actually even in dev ? Do you think this is still needed since we have the request in the profiler ?

YummYume:
Thanks @WebMamba !

I just tried it, and it does work for me, the class is added on the element with the attribute.

image

Is it maybe just that the class does nothing because you're not using Tailwind ?

And about the error modal, I'm not sure, I think it's still a great and more direct debugging tool than the profiler. But for more in depth debugging, it's of course not ideal.

smnandre:
Wherer are we on this one ? are you blocked by something ? Can i help you in any way @YummYume  ?

YummYume:
Hey @smnandre ! I'll work on resolving the conflicts ASAP but other than that, no, I think it was all good !

If there are any suggestions or changes wanted to the current behavior, I'm all ears.

smnandre:
A couple of questions / comments, but to me we're almost there 😄 

I'm still a bit doubtfull we need all the modifier and/or should maybe refactor the code avoid doubling the code with the existing ones, but this (refactor the existing) could be done in a later step / PR.

Do we really need the showError method for instance ? I mean do you think all the JS code can be simplified, more "direct" ?  (genuine question)

Also, i'm not a big fan of the added CSS and would love for use to remove it entirely in future versions, so is there any way to handle this feature without adding CSS ?

YummYume:
Rebase is done.

Some of the code could perhaps be unified, but there was already a big part done (all the parsing is already shared between the loading and error plugins). I also made sure not to include too complex behaviors, such as targeting a specific action (which makes no sense here) or using modifiers with the attributes.

So with all that, I don't see a reason why **not** include them ? At least it stays consistent with the loading plugin. I did my best to make the plugin readable and easy to understand for any further modification, especially since this PR is technically only the first part of our long discussion on how to *properly* handle errors.

I agree with you for the CSS. On top of that, it doesn't **always** work. There can still a short window where you can actually see the element, until this style is loaded. Which is why (and I added this part in the doc as well), I always add my own `display: none;` class on my `data-loading` or `data-live-error` elements. I think the best solution would simply to remove the `live.css` stylesheet and warn the user about the need to add their own hidden class to their elements.

src/Dropzone/assets/src/style.css
```diff
@@ -60,7 +60,7 @@
 }
 
 .dropzone-preview-button::before {
-    content: '×';
+    content: "×";
```
smnandre, line 63: Let's remove this :) 

We'll make the change in another small PR

src/Dropzone/assets/dist/style.min.css
```diff
@@ -1 +1 @@
-.dropzone-container{position:relative;display:flex;min-height:100px;border:2px dashed #bbb;align-items:center;padding:20px 10px}.dropzone-input{position:absolute;display:block;top:0;left:0;width:100%;height:100%;opacity:0;cursor:pointer;z-index:1}.dropzone-preview{display:flex;align-items:center;max-width:100%}.dropzone-preview-image{flex-basis:0;min-width:50px;max-width:50px;height:50px;margin-right:10px;background-size:contain;background-position:50% 50%;background-repeat:no-repeat}.dropzone-preview-filename{word-wrap:anywhere}.dropzone-preview-button{position:absolute;top:0;right:0;z-index:1;border:none;margin:0;padding:0;width:auto;overflow:visible;background:0 0;color:inherit;font:inherit;line-height:normal;-webkit-font-smoothing:inherit;-moz-osx-font-smoothing:inherit;-webkit-appearance:none}.dropzone-preview-button::before{content:'×';padding:3px 7px;cursor:pointer}.dropzone-placeholder{flex-grow:1;text-align:center;color:#999}
```
smnandre, line 1: (same)

smnandre:
Thank you for the quick answers, make sense! 

Let's update the src/LiveComponent CHANGELOG (it's gonna be 2.20) 

@kbond @WebMamba @Kocal if one of you can give a second check ?

src/Dropzone/assets/src/style.css
YummYume: I never made that change, I think that's just the linter ?

src/LiveComponent/assets/src/Component/plugins/ErrorPlugin.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    // Utils
+    private showElement(element: HTMLElement | SVGElement) {
+        element.style.display = 'revert';
+    }
+
+    private hideElement(element: HTMLElement | SVGElement) {
+        element.style.display = 'none';
+    }
```
Kocal, line 199: If the dev used `style="display: flex"` on element, what will happens when calling `showElement`? 

Can we add/remove a custom CSS class instead, like `data-live-hidden`? It would be safer than manipulating `element.style.display` IMO.

src/LiveComponent/assets/src/Component/plugins/ErrorPlugin.ts
```diff
@@ -0,0 +1,224 @@
...
+    private addClass(element: HTMLElement | SVGElement, classes: string[]) {
+        element.classList.add(...combineSpacedArray(classes));
+    }
+
+    private removeClass(element: HTMLElement | SVGElement, classes: string[]) {
+        element.classList.remove(...combineSpacedArray(classes));
+
+        if (element.classList.length === 0) {
+            element.removeAttribute('class');
+        }
```
Kocal, line 210: I think this is not necessary

src/LiveComponent/doc/index.rst
```diff
@@ -928,6 +928,8 @@ The following hooks are available (along with the arguments that are passed):
 * ``loading.state:finished`` args ``(element: HTMLElement)``
 * ``model:set`` args ``(model: string, value: any, component: Component)``
 
+.. _loading:
```
Kocal, line 931: Not a RST syntax expert here, but it's not missing something?

src/LiveComponent/doc/index.rst
```diff
@@ -1039,6 +1041,74 @@ was just changed using the ``model()`` modifier:
...
+In addition to using the ``response:error`` hook, you can also use the special ``data-live-error``
+attribute to show (or hide) an element when the component is in an error state:
+
+.. code-block:: html+twig
+
+    <!-- show only when the component is in an error state -->
+    <span data-live-error>An error occurred ! Please try again.</span>
+
+    <!-- equivalent, longer syntax -->
+    <span data-live-error="show">An error occurred ! Please try again.</span>
```
```suggestion
    <span data-live-error>An error occurred! Please try again.</span>

    <!-- equivalent, longer syntax -->
    <span data-live-error="show">An error occurred! Please try again.</span>
```

src/LiveComponent/assets/src/Component/plugins/ErrorPlugin.ts
smnandre: In fact i believe it is, or the next morph will consider this a change .. or let's say i _think_ i remember this

src/LiveComponent/assets/src/Component/plugins/ErrorPlugin.ts
YummYume: Well, that sadly doesn't work, but this is a current issue with the `data-loading` attribute as well (see https://github.com/symfony/ux/issues/102).
A simple workaround is to do something like `class="hidden" data-live-error="addClass(flex) removeClass(hidden)"` instead. That will work.

I also don't think adding more custom CSS would be a good idea, as discussed above, and we would probably have to use `!important` any way to avoid the style not taking effect.

src/LiveComponent/assets/src/Component/plugins/ErrorPlugin.ts
YummYume: I'm not sure if this behavior changed with the switch to Idiomorph though

src/LiveComponent/doc/index.rst
YummYume: The syntax is a bit weird, but this is basically used for referencing anchors

Kocal:

PR #1923, opened 2024-06-16
[TwigComponent] Add documentation for MakerBundle installation

### LINKED ISSUE #480: [Live][Twig] document `make:twig-component`
`symfony/maker-bundle` [1.46.0](https://github.com/symfony/maker-bundle/releases/tag/v1.46.0) provides a [`make:twig-component`](https://github.com/symfony/maker-bundle/pull/1019) maker that can make both twig and live components.

techSolstice:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #480 
| License       | MIT

docs: Add documentation for a quick way to get the component boilerplate in.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -61,6 +61,7 @@ Let's get this thing installed! Run:
 You can now use the `MakerBundle`_ to get the skeleton of the Twig component made:
 
 .. code-block:: terminal
+
     $ php bin/console make:twig-component
      The name of your twig component (ie Notification):
      > Notification
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -58,7 +58,28 @@ Let's get this thing installed! Run:
 
     $ composer require symfony/ux-twig-component
 
-That's it! We're ready to go! If you're not using Symfony Flex, add a config
+You can now use the `MakerBundle`_ to get the skeleton of the Twig component made:
+
+.. code-block:: terminal
+    $ php bin/console make:twig-component
+     The name of your twig component (ie Notification):
+     > Notification
+
+     Make this a live component? (yes/no) [no]:
+     > no
+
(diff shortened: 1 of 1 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -72,6 +94,8 @@ file to control the template directory for your components:
             # Namespace & directory for components
             App\Twig\Components\: 'components/'
 
+That's it! We're ready to go!
```
smnandre, line 97: I prefer the way it was before. For most of new users, composer require will be enough.

src/TwigComponent/doc/index.rst
```diff
@@ -58,7 +58,29 @@ Let's get this thing installed! Run:
...
+     Make this a live component? (yes/no) [no]:
+     > no
+
+     created: src/Twig/Components/Notification.php
+     created: templates/components/Notification.html.twig
+
+
+      Success!
+
+This will also setup the config.
```
smnandre, line 78: What do you mean here ? How does the maker bundle set the local config ?

src/TwigComponent/doc/index.rst
```diff
@@ -58,7 +58,29 @@ Let's get this thing installed! Run:
...
+
+Similarly, to make a `live component`_, enter ``yes`` when prompted whether
+to make this a live component.
```
smnandre, line 81: I don't think we should talk about LiveComponent that early in the page.

smnandre:
Thank you for the suggestion, @techSolstice.

My primary concern with the current implementation is that using the maker is not strictly related to “installation.”

If we want to document it, I suggest we create a dedicated section, similar to how it’s done for controllers: [Symfony Documentation - Generating Controllers](https://symfony.com/doc/current/controller.html#generating-controllers).

Would you be able to adapt this PR in that direction?

smnandre:
Added in #2384

PR #1927, opened 2024-06-19
[Translator] Add "by-domain" translation dumper

### LINKED ISSUE #1225: [Translator] Filter dumped domains
In my project where I'm using https://github.com/willdurand/BazingaJsTranslationBundle currently, I'm using the `js` domain for translation keys used in JS code instead of keeping them in the `messages` domain. This allows me to reduce the size of the dumped translation a lot by including only this domain in the dumped JS code (I have less than 100 keys in the JS file, while I have more than 2000 keys in the `messages` domain).
It would be great if the bundle could have a configuration setting to define the list of domains to dump for the JS code instead of dumping them all (the default could still be to dump them all).

maelanleborgne:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1225 
| License       | MIT

Will follow #1930 

- Adds a new translation dumper to dump translations by domains in separated files. Theses files can be imported and registered with the new `registerDomain(domain)` js function, and accessed using the existing `trans` function by passing the translation key as a first argument.

TODO:
- [x] Add documentation
- [ ] Add entry to changelog
- [ ] Add better TS typing to the new functions/object
- [ ] Add a TS declaration file for the new domains files ?

### DIFF
--- a/src/Translator/assets/dist/translator_controller.js
+++ b/src/Translator/assets/dist/translator_controller.js
@@ -190,6 +190,7 @@ function getPluralizationRule(number, locale) {
 
 let _locale = null;
 let _localeFallbacks = {};
+let _registeredTranslations = {};
 function setLocale(locale) {
     _locale = locale;
 }
@@ -211,6 +212,9 @@ function trans(message, parameters = {}, domain = 'messages', locale = null) {
     }
     if (typeof locale === 'undefined' || null === locale) {
         locale = getLocale();
+    }
+    if (typeof message === 'string') {
+        message = getRegisteredMessage(message, domain);
     }
     if (typeof message.translations === 'undefined') {
         return message.id;
@@ -242,5 +246,27 @@ function trans(message, parameters = {}, domain = 'messages', locale = null) {
     }
     return message.id;
 }
+function registerDomain(domainTranslations) {
+    for (const [domainName, translationsByLocale] of Object.entries(domainTranslations)) {
+        _registeredTranslations[domainName] = translationsByLocale;
+    }
+}
+function getRegisteredMessage(key, domain) {
+    var _a;
+    var _b;
+    let message = { id: key, translations: {} };
+    for (const domainName of [domain, domain + '+intl-icu']) {
+        if (typeof _registeredTranslations[domainName] === 'undefined') {
+            continue;
+        }
+        for (const [locale, translations] of Object.entries(_registeredTranslations[domainName])) {
+            if (typeof translations[key] !== 'undefined') {
+                (_a = (_b = message.translations)[domainName]) !== null && _a !== void 0 ? _a : (_b[domainName] = {});
+                message.translations[domainName][locale] = translations[key];
+            }
+        }
+    }
+    return message;
+}
 
-export { getLocale, getLocaleFallbacks, setLocale, setLocaleFallbacks, trans };
+export { getLocale, getLocaleFallbacks, registerDomain, setLocale, setLocaleFallbacks, trans };
--- a/src/Translator/assets/dist/translator_controller.js
+++ b/src/Translator/assets/dist/translator_controller.js
@@ -190,7 +190,7 @@ function getPluralizationRule(number, locale) {
 
 let _locale = null;
 let _localeFallbacks = {};
-let _registeredTranslations = {};
+const _registeredTranslations = {};
 function setLocale(locale) {
     _locale = locale;
 }
@@ -254,7 +254,7 @@ function registerDomain(domainTranslations) {
 function getRegisteredMessage(key, domain) {
     var _a;
     var _b;
-    let message = { id: key, translations: {} };
+    const message = { id: key, translations: {} };
     for (const domainName of [domain, domain + '+intl-icu']) {
         if (typeof _registeredTranslations[domainName] === 'undefined') {
             continue;
--- a/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
+++ b/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
@@ -17,6 +17,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\UX\Translator\Dumper\Front\FrontFileDumperInterface;
 
 /**
  * @author Hugo Alliaume <hugo@alliau.me>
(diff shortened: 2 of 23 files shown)


src/Translator/src/DependencyInjection/UxTranslatorExtension.php
```diff
@@ -35,7 +35,24 @@ public function load(array $configs, ContainerBuilder $container)
...
-        $container->getDefinition('ux.translator.translations_dumper')->setArgument(0, $config['dump_directory']);
+        $dumperConfigDefinition = $container->getDefinition('ux.translator.translations_dumper.configuration');
+        $dumperConfigDefinition->setArgument(0, $config['dump_directory']);
+
+        $webpackDumperDefinition = $container->getDefinition('ux.translator.translations_dumper.webpack');
+        $webpackDumperDefinition->setArgument(0, $config['dump_directory']);
+
+        $assetMapperDumperDefinition = $container->getDefinition('ux.translator.translations_dumper.asset_mapper');
+        // Setting to the default value because the folder isn't configurable with asset mapper
+        $assetMapperDumperDefinition->setArgument(0, $this->processConfiguration($configuration, [])['dump_directory']);
```
smnandre: I think we should use some configuration for the user to chose which service he wants to use.. 
... and alias it as "TranslationDumper".

src/Translator/src/FileDumper/AssetMapperTranslationsFileDumper.php
```diff
@@ -0,0 +1,100 @@
...
+            $domain = $translationDetails['domain'];
+            $locale = $translationDetails['locale'];
+            $domainFileName = str_ends_with($domain, MessageCatalogueInterface::INTL_DOMAIN_SUFFIX)
+                ? substr($domain, 0, -\strlen(MessageCatalogueInterface::INTL_DOMAIN_SUFFIX))
+                : $domain;
+
+            $domainSplit[$domainFileName] ??= [];
+            $domainSplit[$domainFileName][$translationId] ??= ['id' => $translationId, 'translations' => []];
+            $domainSplit[$domainFileName][$translationId]['translations'][$domain] ??= [];
+            $domainSplit[$domainFileName][$translationId]['translations'][$domain][$locale] = $translationDetails['message'];
```
smnandre: So you'd keep all the locale in one file ?

src/Translator/src/FileDumper/AssetMapperTranslationsFileDumper.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        foreach ($catalogues as $catalogue) {
+            $locale = $catalogue->getLocale();
+            foreach ($catalogue->getDomains() as $domain) {
+                if (\in_array($domain, $this->excludedDomains, true)) {
+                    continue;
+                }
+                if (0 !== \count($this->domains) && !\in_array($domain, $this->domains, true)) {
+                    continue;
+                }
```
smnandre: I think we maybe should use "includedDomain" / "excludedDomain" or something similar, maybe just me but the difference in naming makes it less clear

src/Translator/src/FileDumper/AssetMapperTranslationsFileDumper.php
maelanleborgne: Yes that's the plan right now, otherwise we'd have to import too many modules (1 per combination of domain/locale). Though a filter on the dumped locales could help remove the unnecessary ones. Unless you have something else in mind ?

src/Translator/src/FileDumper/TranslationConfigDumper.php
```diff
@@ -0,0 +1,64 @@
...
+    public function dump(MessageCatalogueInterface ...$catalogues): void
+    {
+        $this->filesystem->mkdir($this->dumpDir);
+        $this->filesystem->remove($this->dumpDir.'/configuration.js');
+        $this->filesystem->remove($this->dumpDir.'/configuration.d.ts');
+
+        $this->filesystem->dumpFile($this->dumpDir.'/configuration.js', sprintf(
+            "export const localeFallbacks = %s;\n",
+            json_encode($this->getLocaleFallbacks(...$catalogues), \JSON_THROW_ON_ERROR)
+        ));
```
smnandre: Maybe a simple concatenation here ?

src/Translator/src/FileDumper/TranslationConfigDumper.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->filesystem->dumpFile($this->dumpDir.'/configuration.d.ts', <<<'TS'
+import { LocaleType } from '@symfony/ux-translator';
+
+export declare const localeFallbacks: Record<LocaleType, LocaleType>;
+TS
+        );
```
smnandre: Is this file static ?

src/Translator/src/FileDumper/WebpackTranslationsDumper.php
```diff
@@ -72,16 +73,16 @@ public function dump(MessageCatalogueInterface ...$catalogues): void
...
+
+    public function setExcludedDomains(array $domains = []): void
+    {
+        $this->excludedDomains = $domains;
+    }
+
+    public function setDomains(array $domains): void
+    {
+        $this->domains = $domains;
     }
```
smnandre: Wondering if this should either
* be in an abstractDumper
* be in a new TranslationDumpProvider (find a better name^^)

Because there is a certain amount of duplicated code there

src/Translator/tests/FileDumper/AssetMapperTranslationsDumperTest.php
```diff
@@ -0,0 +1,285 @@
...
+    }
+};
+JS, file_get_contents(self::$translationsDumpDir.'/domains/messages.js'));
+
+        $this->assertStringContainsString(<<<'JS'
+export default {
+    "post.num_comments": {
+        "id": "post.num_comments",
+        "translations": {
+            "foobar": {
```
smnandre: is this "domain as key" necessary ?

src/Translator/assets/dist/translator_controller.js
```diff
@@ -242,5 +249,34 @@ function trans(message, parameters = {}, domain = 'messages', locale = null) {
...
+        if (typeof _registeredTranslations[id].translations[domainName] === 'undefined') {
+            _registeredTranslations[id].translations[domainName] = message.translations[domainName];
+            continue;
+        }
+        throw new Error(`The domain "${domainName}" has already been registered for the message "${id}".`);
+    }
+}
+function resetRegisteredDomains() {
+    _registeredTranslations = {};
+}
```
smnandre: What is the use case ?

src/Translator/src/FileDumper/WebpackTranslationsDumper.php
```diff
@@ -30,8 +30,11 @@
  * @phpstan-type Locale string
  * @phpstan-type MessageId string
  */
-class TranslationsDumper
```
smnandre: This would be a massive BC break... not sure the "experimental" would be enough :)

Maybe should we keep it as alias / deprecated ? 

@kbond ?

maelanleborgne: What about keeping the TranslationDumper as a wrapper that would call either or both the Webpack and AssetMapper dumper + the config dumper ?

smnandre: Why not ! That could reduce some code duplication too!

Kocal (commented):
Hi, and thanks for your contribution! :)

I didn't look for everything, but isn't there too much things for a single PR? 
I see dump per domains, global domains filters, but also a new dumper for AssetMapper... is it too much?

maelanleborgne:
> Hi, and thanks for your contribution! :)
> 
> I didn't look for everything, but isn't there too much things for a single PR? I see dump per domains, global domains filters, but also a new dumper for AssetMapper... is it too much?

Hi, I thought indeed about making two PRs : one for the filtering and another one for the 'per-domain' dump (which is what the AssetMapperDumper does, there won't be any change on the actual webpack-oriented behavior). I made it in one PR because it all came from the same issue, but I can still split the PR if need be.

src/Translator/assets/dist/translator_controller.js
maelanleborgne: It was mainly to reset the translations between tests, but I changed the behavior to allow overriding already registered translations, so I got rid of this method.

smnandre:
Hi @maelanleborgne : do you need anything on this one ?

Kocal:

PR #1930, opened 2024-06-20
[Translator] Add configuration to filter dumped translations by domain

### LINKED ISSUE #1225: [Translator] Filter dumped domains

maelanleborgne:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1225 
| License       | MIT

Add two config entries `included_domains` and `excluded_domains` to filter which translation domains should be dumped.
If `included_domains` is empty, it is not taken into account.

TODO:
- [ ] Add entry to changelog

### DIFF
--- a/src/Translator/doc/index.rst
+++ b/src/Translator/doc/index.rst
@@ -65,7 +65,7 @@ After installing the bundle, the following file should be created, thanks to the
 Usage
 -----
 
-When warming up the Symfony cache, all of your translations will be dumped as JavaScript into the ``var/translations/`` directory.
+When warming up the Symfony cache, your translations will be dumped as JavaScript into the ``var/translations/`` directory.
 For a better developer experience, TypeScript types definitions are also generated aside those JavaScript files.
 
 Then, you will be able to import those JavaScript translations in your assets.
@@ -74,6 +74,24 @@ Don't worry about your final bundle size, only the translations you use will be 
 .. note::
 
     This package requires the `translator` to be enabled in your Symfony application. If you don't use the `translator` service, the warmup command will not generate any translations.
+
+Configuring the dumped translations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, all your translations will be exported. You can restrict the dumped messages by either
+including or excluding translation domains in your ``config/packages/ux_translator.yaml`` file:
+
+.. code-block:: yaml
+
+    ux_translator:
+            domains: ~    # Include all the domains
+
+            domains: foo  # Include only domain 'foo'
+            domains: '!foo' # Include all domains, except 'foo'
+
+            domains: [foo, bar]   # Include only domains 'foo' and 'bar'
+            domains: ['!foo', '!bar'] # Include all domains, except 'foo' and 'bar'
+
 
 Configuring the default locale
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/src/Translator/src/DependencyInjection/Configuration.php
+++ b/src/Translator/src/DependencyInjection/Configuration.php
@@ -13,6 +13,7 @@ namespace Symfony\UX\Translator\DependencyInjection;
 
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 
 /**
  * @author Hugo Alliaume <hugo@alliau.me>
@@ -28,6 +29,59 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->scalarNode('dump_directory')->defaultValue('%kernel.project_dir%/var/translations')->end()
+                ->arrayNode('domains')
+                    ->info('List of domains to include/exclude from the generated translations. Prefix with a `!` to exclude a domain.')
+                    ->children()
+                        ->scalarNode('type')
+                            ->validate()
+                                ->ifNotInArray(['inclusive', 'exclusive'])
+                                ->thenInvalid('The type of domains has to be inclusive or exclusive')
+                            ->end()
+                        ->end()
+                        ->arrayNode('elements')
+                            ->scalarPrototype()->end()
+                        ->end()
+                    ->end()
+                    ->canBeUnset()
+                    ->beforeNormalization()
+                        ->ifString()
+                        ->then(fn ($v) => ['elements' => [$v]])
+                    ->end()
+                    ->beforeNormalization()
+                        ->ifTrue(function ($v) { return \is_array($v) && is_numeric(key($v)); })
+                        ->then(function ($v) { return ['elements' => $v]; })
+                    ->end()
+                    ->validate()
+                        ->always(function ($v) {
+                            $isExclusive = null;
+                            $elements = [];
+                            if (isset($v['type'])) {
+                                $isExclusive = 'exclusive' === $v['type'];
+                            }
+                            foreach ($v['elements'] as $domain) {
+                                if (str_starts_with($domain, '!')) {
+                                    if (false === $isExclusive) {
+                                        throw new InvalidConfigurationException('You cannot mix included and excluded domains.');
+                                    }
+                                    $isExclusive = true;
+                                    $elements[] = substr($domain, 1);
+                                } else {
+                                    if (true === $isExclusive) {
+                                        throw new InvalidConfigurationException('You cannot mix included and excluded domains.');
+                                    }
+                                    $isExclusive = false;
+                                    $elements[] = $domain;
+                                }
+                            }
+
+                            if (!\count($elements)) {
+                                return null;
+                            }
+
+                            return ['type' => $isExclusive ? 'exclusive' : 'inclusive', 'elements' => array_unique($elements)];
+                        })
+                    ->end()
+                ->end()
             ->end()
         ;
 
--- a/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
+++ b/src/Translator/src/DependencyInjection/UxTranslatorExtension.php
@@ -35,7 +35,15 @@ class UxTranslatorExtension extends Extension implements PrependExtensionInterfa
         $loader = (new PhpFileLoader($container, new FileLocator(\dirname(__DIR__).'/../config')));
         $loader->load('services.php');
 
-        $container->getDefinition('ux.translator.translations_dumper')->setArgument(0, $config['dump_directory']);
+        $dumperDefinition = $container->getDefinition('ux.translator.translations_dumper');
+        $dumperDefinition->setArgument(0, $config['dump_directory']);
+
+        if (isset($config['domains'])) {
+            $method = 'inclusive' === $config['domains']['type'] ? 'addIncludedDomain' : 'addExcludedDomain';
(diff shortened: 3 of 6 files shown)


src/Translator/doc/index.rst
```diff
@@ -65,12 +65,28 @@ After installing the bundle, the following file should be created, thanks to the
...
+excluding translation domains in your ``config/packages/ux_translator.yaml`` file:
+
+.. code-block:: yaml
+
+    ux_translator:
+        # By listing the domains you want to include :
+        included_domains: ['messages', 'admin']
+
+        # Or by excluding the domains you don't want :
+        excluded_domains: ['validators', 'security']
```
Kocal:
```suggestion
        # By listing the domains you want to include:
        included_domains: ['messages', 'admin']

        # Or by excluding the domains you don't want:
        excluded_domains: ['validators', 'security']
```

No space before `:` in english

src/Translator/src/DependencyInjection/Configuration.php
```diff
@@ -28,6 +28,14 @@ public function getConfigTreeBuilder(): TreeBuilder
...
             ->children()
                 ->scalarNode('dump_directory')->defaultValue('%kernel.project_dir%/var/translations')->end()
+                ->arrayNode('excluded_domains')
+                    ->scalarPrototype()->end()
+                    ->defaultValue([])
+                ->end()
+                ->arrayNode('included_domains')
+                    ->scalarPrototype()->end()
+                    ->defaultValue([])
+                ->end()
```
Kocal: IMO if we configure the domains we want to exclude, we don't need to configure the domains we want to include, and vice-versa.

What about adding a validation telling the user that `excluded_domains` and `included_domains` can not be filled at the same time? :)

src/Translator/src/TranslationsDumper.php
```diff
@@ -94,6 +107,12 @@ private function getTranslations(MessageCatalogueInterface ...$catalogues): arra
         foreach ($catalogues as $catalogue) {
             $locale = $catalogue->getLocale();
             foreach ($catalogue->getDomains() as $domain) {
+                if (\in_array($domain, $this->excludedDomains, true)) {
+                    continue;
+                }
+                if (0 !== \count($this->includedDomains) && !\in_array($domain, $this->includedDomains, true)) {
```
Kocal: Can be simplified to:
```suggestion
                if ($this->includedDomains && !\in_array($domain, $this->includedDomains, true)) {
```

Kocal (commented):
Thanks for your contribution and splitting #1927 into two PR, that was a lot easier to review! :)

src/Translator/doc/index.rst
```diff
@@ -65,12 +65,28 @@ After installing the bundle, the following file should be created, thanks to the
...
+By default, all your translations will be dumped as JavaScript. You can restrict the dumped translations by including or
+excluding translation domains in your ``config/packages/ux_translator.yaml`` file:
+
+.. code-block:: yaml
+
+    ux_translator:
+        # By listing the domains you want to include :
+        included_domains: ['messages', 'admin']
+
+        # Or by excluding the domains you don't want :
```
smnandre: If this is really a "OR" we should say it explicitely (something like "you could either .. or ")

src/Translator/src/DependencyInjection/Configuration.php
```diff
@@ -28,6 +28,14 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->children()
                 ->scalarNode('dump_directory')->defaultValue('%kernel.project_dir%/var/translations')->end()
+                ->arrayNode('excluded_domains')
+                    ->scalarPrototype()->end()
```
smnandre: Could you add a tiny comment/doc there (improve DX when debug config) ?

(for both)

src/Translator/src/TranslationsDumper.php
```diff
@@ -84,6 +87,16 @@ public function dump(MessageCatalogueInterface ...$catalogues): void
         );
     }
 
+    public function setExcludedDomains(array $domains): void
```
smnandre:
```suggestion
    /**
    * @param list<string> $domains
    */
    public function setExcludedDomains(array $domains): void
```

?

src/Translator/src/TranslationsDumper.php
```diff
@@ -84,6 +87,16 @@ public function dump(MessageCatalogueInterface ...$catalogues): void
...
+    {
+        $this->excludedDomains = $domains;
+    }
+
+    public function setIncludedDomains(array $domains): void
+    {
+        $this->includedDomains = $domains;
```
smnandre: Maybe check here if there is excludedDomains ?

src/Translator/src/TranslationsDumper.php
```diff
@@ -94,6 +107,12 @@ private function getTranslations(MessageCatalogueInterface ...$catalogues): arra
...
             $locale = $catalogue->getLocale();
             foreach ($catalogue->getDomains() as $domain) {
+                if (\in_array($domain, $this->excludedDomains, true)) {
+                    continue;
+                }
+                if (0 !== \count($this->includedDomains) && !\in_array($domain, $this->includedDomains, true)) {
+                    continue;
+                }
                 foreach ($catalogue->all($domain) as $id => $message) {
                     $realDomain = $catalogue->has($id, $domain.MessageCatalogueInterface::INTL_DOMAIN_SUFFIX)
```
smnandre, line 123: Maybe leverage CatalogueMetadataAwareInterface here ?

src/Translator/src/DependencyInjection/UxTranslatorExtension.php
```diff
@@ -35,7 +35,15 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = (new PhpFileLoader($container, new FileLocator(\dirname(__DIR__).'/../config')));
         $loader->load('services.php');
 
-        $container->getDefinition('ux.translator.translations_dumper')->setArgument(0, $config['dump_directory']);
+        $dumperDefinition = $container->getDefinition('ux.translator.translations_dumper');
+        $dumperDefinition->setArgument(0, $config['dump_directory']);
+
+        if (0 !== \count($config['excluded_domains'])) {
```
smnandre:
```suggestion
        if ($config['excluded_domains']) {
```

src/Translator/src/DependencyInjection/UxTranslatorExtension.php
```diff
@@ -35,7 +35,15 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $dumperDefinition->addMethodCall('setExcludedDomains', [$config['excluded_domains']]);
+        }
+        if (0 !== \count($config['included_domains'])) {
```
smnandre:
```suggestion
        if ($config['included_domains']) {
```

src/Translator/src/DependencyInjection/Configuration.php
```diff
@@ -28,6 +28,20 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->children()
                 ->scalarNode('dump_directory')->defaultValue('%kernel.project_dir%/var/translations')->end()
+                ->arrayNode('excluded_domains')
+                    ->info('List of domains to exclude from the generated translations')
+                    ->scalarPrototype()->end()
+                    ->defaultValue([])
+                ->end()
+                ->arrayNode('included_domains')
+                    ->info('List of domains to include in the generated translations')
```
smnandre: Add info that all domains will be included by default ?

src/Translator/doc/index.rst
```diff
@@ -65,12 +65,28 @@ After installing the bundle, the following file should be created, thanks to the
...
 -----
 
-When warming up the Symfony cache, all of your translations will be dumped as JavaScript into the ``var/translations/`` directory.
+When warming up the Symfony cache, your translations will be dumped as JavaScript into the ``var/translations/`` directory.
 For a better developer experience, TypeScript types definitions are also generated aside those JavaScript files.
 
 Then, you will be able to import those JavaScript translations in your assets.
 Don't worry about your final bundle size, only the translations you use will be included in your final bundle, thanks to the `tree shaking <https://webpack.js.org/guides/tree-shaking/>`_.
 
+Configuring the dumped translations
```
smnandre, line 78:
```suggestion
Restrict the translation domains
```

?

src/Translator/doc/index.rst
```diff
@@ -65,12 +65,28 @@ After installing the bundle, the following file should be created, thanks to the
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, all your translations will be dumped as JavaScript. You can restrict the dumped translations by either
```
```suggestion
By default, all your translations will be exported. You can restrict the dumped messages by either
```

smnandre:
So were are we there ? :) Do you need help for something ? Is the doc ready ?

maelanleborgne:
> So were are we there ? :) Do you need help for something ? Is the doc ready ?

@smnandre Yes that should be it, the doc as well. I'm just not sure what version I should use in the changelog entry.

smnandre:
2.19 no ?

smnandre:
The failing test surprises me a bit ... @WebMamba any idea ? I feel we may have exposed more than we wanted

WebMamba:
Looks like this is related to the PR: https://github.com/symfony/ux/pull/1940.

Hummm the test looks flaky. I don't think this is related to this PR though

kbond (commented):
Looks good, just questioning the config choice

src/Translator/doc/index.rst
```diff
@@ -65,12 +65,28 @@ After installing the bundle, the following file should be created, thanks to the
...
+including or excluding translation domains in your ``config/packages/ux_translator.yaml`` file:
+
+.. code-block:: yaml
+
+    ux_translator:
+        # By listing the domains you want to include:
+        included_domains: ['messages', 'admin']
+
+        # Or by excluding the domains you don't want:
+        excluded_domains: ['validators', 'security']
```
kbond: What about using the same style monolog uses for channel config:

```suggestion
    ux_translator:
        domains: ~    # Include all the domains

        domains: foo  # Include only domain 'foo'
        domains: '!foo' # Include all domains, except 'foo'

        domains: [foo, bar]   # Include only domains 'foo' and 'bar'
        domains: ['!foo', '!bar'] # Include all domains, except 'foo' and 'bar'
```

smnandre: So this would throw a invalid config exception ? 

```yaml
        domains: ['!foo', 'bar'] 
```

kbond: I guess just mimic what monolog does in this case.

maelanleborgne: Thanks for the suggestion, I've pushed the changes.

kbond (approved):

kbond:
This is great @maelanleborgne, thank you!

PR #1931, opened 2024-06-21
[Turbo] Modernize code (ppp, types..)

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| Issues        | Fix #...
| License       | MIT

### DIFF
--- a/src/Turbo/src/Bridge/Mercure/Broadcaster.php
+++ b/src/Turbo/src/Bridge/Mercure/Broadcaster.php
@@ -40,20 +40,13 @@ final class Broadcaster implements BroadcasterInterface
      */
     public const TOPIC_PATTERN = 'https://symfony.com/ux-turbo/%s/%s';
 
-    private $name;
-    private $hub;
+    private ?ExpressionLanguage $expressionLanguage;
 
-    /** @var ExpressionLanguage|null */
-    private $expressionLanguage;
-
-    public function __construct(string $name, HubInterface $hub)
-    {
-        $this->name = $name;
-        $this->hub = $hub;
-
-        if (class_exists(ExpressionLanguage::class)) {
-            $this->expressionLanguage = new ExpressionLanguage();
-        }
+    public function __construct(
+        private string $name,
+        private HubInterface $hub,
+    ) {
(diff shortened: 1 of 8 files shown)


src/Turbo/src/Bridge/Mercure/Broadcaster.php
```diff
@@ -40,17 +40,15 @@ final class Broadcaster implements BroadcasterInterface
...
     public const TOPIC_PATTERN = 'https://symfony.com/ux-turbo/%s/%s';
 
-    private $name;
-    private $hub;
-
-    /** @var ExpressionLanguage|null */
+    /**
+     * @var ExpressionLanguage|null
+     */
     private $expressionLanguage;
```
Kocal: Can't we type it without PHPDoc? (note the assignment)
```suggestion
    private ExpressionLanguage|null $expressionLanguage = null;
```

src/Turbo/src/Broadcaster/TwigBroadcaster.php
```diff
@@ -21,20 +21,15 @@
...
-    {
-        $this->broadcaster = $broadcaster;
-        $this->twig = $twig;
-        $this->templatePrefixes = $templatePrefixes;
-        $this->idAccessor = $idAccessor ?? new IdAccessor();
+    public function __construct(
+        private BroadcasterInterface $broadcaster,
+        private Environment $twig,
+        private array $templatePrefixes = [],
+        private IdAccessor $idAccessor = new IdAccessor(),
```
wouterj: This is a BC break, the argument is no longer nullable.

src/Turbo/src/Bridge/Mercure/Broadcaster.php
smnandre: Well spoted. I set the value in the constructor.

src/Turbo/src/Broadcaster/TwigBroadcaster.php
smnandre: Absolutely, thank you. fixed

kbond:
Thank you Simon.

PR #1936, opened 2024-06-22
[Site] Use ux-icons (1/2)

smnandre:
First pass to switch from the old custom Icon component to ux-icons

Next step (once feature available on ux-icons)
* configure class & data-attributes auto-attributes
* use <twig:ux:icon instead of <twig:Icon


smnandre:
Let's ignore the cs-twig for now... that will require some work to fix the Twig CS Fixer "UX parser" i suppose 

(poke @VincentLanglet if you have an idea to where i should start)

VincentLanglet:
> Let's ignore the cs-twig for now... that will require some work to fix the Twig CS Fixer "UX parser" i suppose
> 
> (poke @VincentLanglet if you have an idea to where i should start)

This is supposed to be fixed in 2.10.0 version https://github.com/VincentLanglet/Twig-CS-Fixer/releases/tag/2.10.0
I assume you need to bump the version 
https://github.com/symfony/ux/blob/2.x/ux.symfony.com/composer.lock#L11840-L11841

smnandre:
Oh that's a great news :) Thank you !

smnandre:
Remaining cs-twig failures are fixed in https://github.com/symfony/ux/pull/1935

kbond (approved):

kbond:
Thanks Simon.

PR #1940, opened 2024-06-25
[TwigComponent] Ignore array|sequence in exception message

smnandre:
Twig switched from using "array" to "sequence" in 3.11 (#4106)

So let's go simple and test start/end of the exception message there


WebMamba (approved):

kbond:
Thanks Simon.

PR #1941, opened 2024-06-26
Update index.rst Usage React with AssetMapper

xDeSwa:
Completing the documentation of usage with UX React with Asset Mapper

### DIFF
--- a/src/React/doc/index.rst
+++ b/src/React/doc/index.rst
@@ -98,16 +98,36 @@ Using with AssetMapper
 Because the JSX format isn't pure JavaScript, using this library with AssetMapper
 requires some extra steps.
 
-#. Compile your ``.jsx`` files to pure JavaScript files. This can be done by
+.. note:: Compile your ``.jsx`` files to pure JavaScript files. This can be done by
    installing Babel and the ``@babel/preset-react`` preset. Example:
    https://github.com/symfony/ux/blob/2.x/ux.symfony.com/package.json
 
-#. Point this library at the "built" controllers directory that contains the final
+1- Create package.json in your project. This file is not included in the **Standard package**. (code below)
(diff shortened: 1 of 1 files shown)


src/React/doc/index.rst
```diff
@@ -98,16 +98,38 @@ Using with AssetMapper
...
+  },
+  "license": "MIT",
+  "private": true,
+  "scripts": {
+    "build": "yarn build-react && yarn build-svelte",
+    "build-react": "babel assets/react --out-dir assets/build/react --presets=@babel/preset-react",
+    "build-svelte": "node bin/compile_svelte.js"
+  }
+ }
+
```
WebMamba, line 126: I think we don't need to add the content of the package.json. Let's stay simple and say that we need a package.json and to install few dependencies.

src/React/doc/index.rst
```diff
@@ -98,16 +98,38 @@ Using with AssetMapper
...
+2- Point this library at the "built" controllers directory that contains the final
    JavaScript files:
 
 .. code-block:: yaml
 
-    # config/packages/react.yaml
+    # config/packages/asset_mapper.yaml
```
WebMamba, line 132: Why change the file name ?

src/React/doc/index.rst
```diff
@@ -119,6 +141,33 @@ Also, inside of your ``.jsx`` files, when importing another component, use the
     // use PackageList.js even though the file is named PackageList.jsx
     import PackageList from '../components/PackageList.js';
 
+3- Install the required Node.js packages
+
+.. code-block:: terminal
+
+    $ npm install
```
WebMamba, line 148: You use npm and below yarn. Let's stay on yarn

PR #1946, opened 2024-06-29
[LiveComponent] Add 'live_action' twig function

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | N/A
| License       | MIT

Adds a new `live_action` twig function.

This helps with creating the proper attributes to call a live action. I can never remember the syntax and need to look up the docs every time to add a live action. This will make the process much easier.

Example Usage:
```diff
- <button data-action="live#action" data-live-action-param="save">Save</button>
+ <button {{ live_action('save') }}>Save</button>
```

Additional Parameters:
```diff
<button
-  data-action="live#action"
-  data-live-action-param="addItem"
-  data-live-id-param="{{ item.id }}"
-  data-live-item-name-param="CustomItem"
+  {{ live_action('addItem', {'id': item.id, 'itemName': 'CustomItem'}) }}
>Add Item</button>
```

Adding Modifiers:
```diff
<button
-  data-action="live#action:prevent"
-  data-live-action-param="debounce(300)|save"
+  {{ live_action('save', {}, {'debounce': 300, 'prevent': true}) }}
 >Save</button>
```

#### TODO:

- [x] Add CHANGELOG entry
- [x] Update docs

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 -   Add `submitForm()` to `TestLiveComponent`.
+-   Add `live_action` Twig function
 
 ## 2.18.0
 
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -28,6 +28,7 @@
     "require": {
         "php": ">=8.1",
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
+        "symfony/stimulus-bundle": "^2.9",
         "symfony/ux-twig-component": "^2.8",
         "twig/twig": "^3.8.0"
     },
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -1106,6 +1106,17 @@ You can also add several "modifiers" to the action:
 The ``debounce(300)`` adds 300ms of "debouncing" before the action is executed.
 In other words, if you click really fast 5 times, only one Ajax request will be made!
 
+You can also use the ``live_action`` twig helper function to render the attributes:
+
+.. code-block:: html+twig
+
+    <button {{ live_action('resetMax') }}>Reset Min/Max</button>
+
+    {# with modifiers #}
+
+    <button {{ live_action('save', {}, {'debounce': 300}) }}>Save</button>
+
+
 Actions & Services
 ~~~~~~~~~~~~~~~~~~
 
@@ -1157,6 +1168,12 @@ You can also pass arguments to your action by adding each as a
             data-live-id-param="{{ item.id }}"
             data-live-item-name-param="CustomItem"
         >Add Item</button>
+    </form>
+
+    {# or #}
+
+    <form>
+        <button {{ live_action('addItem', {'id': item.id, 'itemName': 'CustomItem' })>Add Item</button>
     </form>
 
 In your component, to allow each argument to be passed, add
--- a/src/LiveComponent/src/Twig/LiveComponentExtension.php
+++ b/src/LiveComponent/src/Twig/LiveComponentExtension.php
@@ -25,6 +25,7 @@ final class LiveComponentExtension extends AbstractExtension
     {
         return [
             new TwigFunction('component_url', [LiveComponentRuntime::class, 'getComponentUrl']),
+            new TwigFunction('live_action', [LiveComponentRuntime::class, 'liveAction'], ['is_safe' => ['html_attr']]),
         ];
     }
 }
--- a/src/LiveComponent/src/Twig/LiveComponentRuntime.php
+++ b/src/LiveComponent/src/Twig/LiveComponentRuntime.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\LiveComponent\Twig;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\StimulusBundle\Helper\StimulusHelper;
 use Symfony\UX\TwigComponent\ComponentFactory;
 
 /**
@@ -28,6 +29,7 @@ final class LiveComponentRuntime
         private ComponentFactory $factory,
         private UrlGeneratorInterface $urlGenerator,
         private LiveComponentMetadataFactory $metadataFactory,
+        private StimulusHelper $stimulusHelper,
     ) {
     }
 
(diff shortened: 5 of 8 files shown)


smnandre:
Hi @pierredup ! 

I like the idea very much (in term of DX), but internally i think we should leverage the StimulusAttribute already present in the Stimulus-Bundle  ...  https://symfony.com/bundles/StimulusBundle/current/index.html#stimulus-action

... what do you thinkl ?

smnandre:
Depending on your answer on the previous question, this comment could be useless ;) 

We'd need to specify the trigger, as even if an action is often trigger by the "default event" on the DOM element it's registerered on, but this can be more specific / diverse

Examples here: https://stimulus.hotwired.dev/reference/actions#keyboardevent-filter

pierredup:
> internally i think we should leverage the StimulusAttribute already present in the Stimulus-Bundle

I did initially start to go down this route, but then noticed the StimulusBundle is not a direct dependency of the LiveComponent, so didn't want to use classes that's not guaranteed to be available. Will then just add the StimulusBundle as a dependency to LiveComponent

> We'd need to specify the trigger, as even if an action is often trigger by the "default event" on the DOM element it's registerered on, but this can be more specific / diverse

Cool, yeah that makes sense, will add it to the function

kbond (approved):
I love this - you're totally right, the syntax is hard to remember!

src/LiveComponent/doc/index.rst
```diff
@@ -1106,6 +1106,21 @@ You can also add several "modifiers" to the action:
 The ``debounce(300)`` adds 300ms of "debouncing" before the action is executed.
 In other words, if you click really fast 5 times, only one Ajax request will be made!
 
+You can also use the ``live_action`` twig helper function to render the attributes:
+
+.. code-block:: html+twig
+
+    <button
+        {{ live_action('resetMax') }}
+    >Reset Min/Max</button>
```
smnandre:
```suggestion
    <button {{ live_action('resetMax') }}>Reset Min/Max</button>
```

I'm hesitating ... wdyt ?

src/LiveComponent/src/Twig/LiveComponentRuntime.php
```diff
@@ -45,4 +47,25 @@ public function getComponentUrl(string $name, array $props = []): string
...
+
+    public function getLiveAction(string $actionName, array $parameters = [], array $modifiers = [], ?string $event = null): string
+    {
+        $attributes = $this->stimulusHelper->createStimulusAttributes();
+
+        $prevent = false;
+        if (\array_key_exists('prevent', $modifiers) && $modifiers['prevent']) {
+            $prevent = true;
+            unset($modifiers['prevent']);
+        }
```
smnandre: I'm thinkink we should maybe not deal with this and let the user type 'validate:prevent' for instance... 

because these are not modifiers for use like 'addClass' or 'delay' but Stimulus ones.. 

```twig
<button {{ live_action('save:prevent') }}>
```

wdyt ?

src/LiveComponent/composer.json
```diff
@@ -28,6 +28,7 @@
     "require": {
         "php": ">=8.1",
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
+        "symfony/stimulus-bundle": "^2.9",
```
smnandre, line 31: I know this is something that was not done before, but don't know/remember the implications...

@kbond any opinion there ?

smnandre: Maybe can we _not_ require it there, and throw an exeption on call if the component is not installed ?

src/LiveComponent/src/Twig/LiveComponentExtension.php
```diff
@@ -25,6 +25,7 @@ public function getFunctions(): array
     {
         return [
             new TwigFunction('component_url', [LiveComponentRuntime::class, 'getComponentUrl']),
+            new TwigFunction('live_action', [LiveComponentRuntime::class, 'getLiveAction'], ['is_safe' => ['html_attr']]),
```
smnandre:
```suggestion
            new TwigFunction('live_action', [LiveComponentRuntime::class, 'liveAction'], ['is_safe' => ['html_attr']]),
```

Feels a bit weird to use '_get_' here, as if we returned the Attribute instance... maybe `liveAction` or `renderLiveAction` to be precise ?

src/LiveComponent/tests/Unit/Twig/LiveComponentRuntimeTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+        $props = $runtime->getLiveAction('action-name', ['pro1' => 'val1', 'prop2' => 'val2'], ['debounce' => 300, 'prevent' => true]);
+        $this->assertSame('data-action="live#action:prevent" data-live-pro1-param="val1" data-live-prop2-param="val2" data-live-action-param="debounce(300)|action-name"', \html_entity_decode($props));
+
+        $props = $runtime->getLiveAction('action-name', [], ['debounce' => 300, 'prevent' => true]);
+        $this->assertSame('data-action="live#action:prevent" data-live-action-param="debounce(300)|action-name"', \html_entity_decode($props));
+
+        $props = $runtime->getLiveAction('action-name', [], [], 'keydown.esc');
+        $this->assertSame('data-action="keydown.esc->live#action" data-live-action-param="action-name"', \html_entity_decode($props));
+    }
```
smnandre, line 43: Could you just add a test case or two to check the escaping ? (for instance with quotes in param values) 

...and maybe a check that the param value is stringable ?

src/LiveComponent/src/Twig/LiveComponentRuntime.php
smnandre: If we keep this implementation, you should unset the modifier even if it is false or null or this will generate an modifier the JS live controller won't be able to parse anyway

src/LiveComponent/composer.json
kbond: Yeah, I wondered about this also. As this is an optional feature, maybe a hard requirement is overkill... but.. I'm thinking to leave as is and remove if there is a problem.

src/LiveComponent/src/Twig/LiveComponentRuntime.php
```diff
@@ -14,7 +14,12 @@
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\StimulusBundle\Helper\StimulusHelper;
 use Symfony\UX\TwigComponent\ComponentFactory;
+use function array_pop;
+use function array_unshift;
+use function explode;
+use function implode;
```
smnandre: This is not in the code style of symfony, afaik ... could you remove them ? :)

src/LiveComponent/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 -   Add `submitForm()` to `TestLiveComponent`.
+-   Add `live_action` twig function
```
smnandre:
```suggestion
-   Add `live_action` Twig function
```

smnandre (commented):
Thank you @pierredup ! 

Just two last tiny details and good!

src/LiveComponent/src/Twig/LiveComponentRuntime.php
pierredup: Ah, default settings in my IDE that does this :)

kbond (approved):

kbond:

PR #1947, opened 2024-06-29
[Site] Update vendors

smnandre:
Update vendors
Update Symfony 7.1
Update import map

Fix minor bug:
* remove "icon.php" file
* remove global bind autowiring


kbond:
Thanks Simon.

PR #1949, opened 2024-06-30
[Site] Add Badge & Cluster

smnandre:
badges

### DIFF
--- a/ux.symfony.com/templates/cookbook/show.html.twig
+++ b/ux.symfony.com/templates/cookbook/show.html.twig
@@ -56,9 +56,37 @@
             </div>
         </div>
 
+        <div class="container container-xl x-4 pt-4 px-md-5 pt-md-5">
+            <div class="cluster">
+                <twig:Badge:Author author="{{ cookbook.author }}" />
+                <twig:Badge:Date label="Published" date="{{ cookbook.publishedAt }}" />
+            </div>
+        </div>
+
     </article>
 {% endblock %}
 
 {% block aside %}
-    {{ include('_aside.html.twig') }}
+    {% embed '_aside.html.twig' %}
+        {% block links %}
+             <twig:DocsLink
+                    title="PHP Packages"
+                    text="Symfony UX bundles & components."
+                    url="{{ url('app_packages') }}"
+                    icon="lucide:package"
+                />
+                <twig:DocsLink
+                    title="Interactive Demos"
+                    text="With commented source code."
+                    url="{{ url('app_demo_live_component') }}"
+                    icon="lucide:monitor-check"
+                />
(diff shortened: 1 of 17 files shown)


javiereguiluz (commented):
In my opinion, `cluster` is an unconventional name, so it could be confusing.

I'd prefer to use a more common name like `stack`. E.g. https://primer.style/components/stack

If this component is only to display a group of badges, we could call it `BadgeGroup` (this is common in other component libraries; e.g. https://primer.style/components/label and https://primer.style/components/label-group)

smnandre:
Cluster is common in design system and modern CSS layout compositions too ;) 
* https://design-system.w3.org/layouts/cluster.html
* https://every-layout.dev/layouts/cluster/
* https://chrisburnell.com/article/cluster-composition/

A cluster is a bit more than an horizontal stack (but it would work here so i agree to use stack if you guys prefer)

The idea is to start using some layout / composition utilities on one hand (dealing with spacing and alignments)... and components (dealing with their inside) 

Especially if we soon start using container queries..

So for me a BadgeGroup would not really bring something to the table.

(open to debate ofc)

javiereguiluz:
@smnandre thanks for sharing these resources. I was **very wrong** and indeed `cluster` is a common word related to CSS layouts. Cheers!

smnandre:
I was surprised and -as always- probably did express myself badly ..

My intention was never to say you were _wrong_.. just wanted to bring some sources to illustrate. 

Really sorry if i have made you feel any other way.

ux.symfony.com/templates/cookbook/show.html.twig
```diff
@@ -56,9 +56,37 @@
             </div>
         </div>
 
+        <div class="container container-xl x-4 pt-4 px-md-5 pt-md-5">
+            <div class="cluster">
+                <twig:Badge:Author author="{{ cookbook.author }}" />
+                <twig:Badge:Date label="Published" date="{{ cookbook.publishedAt }}" />
+            </div>
+        </div>
```
WebMamba, line 64: What do you think about putting the badges at the top of the page between the title and the image ?

WebMamba (commented):
What do you think about adding the author also on the page metadata ?

ux.symfony.com/templates/cookbook/show.html.twig
smnandre: No opinion

smnandre:
> What do you think about adding the author also on the page metadata ?

We can discuss it in another issue ? Or ill remove the badges from the cookbook for now. 

This PR was just about merging badge (and was originally intended for the démos).

kbond:
Thanks Simon.

Issue #1950, opened 2024-07-01
[Translator] TranslationsDumper builds invalid constant names


aleho:
If the translation sources contain strings instead of identifiers the dumper builds invalid constant names.

```yaml
"The passwords don't match.": "Die Passwörter stimmen nicht überein."
"Your password should be at least {{ limit }} characters.": "Ihr Passwort sollte mindestens {limit} Zeichen lang sein."
```

results in

```js
export const THE_PASSWORDS_DON'T_MATCH. = {"id":"The passwords don't match.","translations":{"validators+intl-icu":{"en":"The passwords don't match.","de":"Die Passw\u00f6rter stimmen nicht \u00fcberein."}}};
export const YOUR_PASSWORD_SHOULD_BE_AT_LEAST_{{_LIMIT_}}_CHARACTERS. = {"id":"Your password should be at least {{ limit }} characters.","translations":{"validators+intl-icu":{"en":"Your password should be at least {limit} characters.","de":"Ihr Passwort sollte mindestens {limit} Zeichen lang sein."}}};
```

Are messages as IDs not supported or is this a bug?
If bug, should the IDs be run through e.g. `AsciiSlugger`?

jmsche:
See https://github.com/symfony/symfony/issues/57612

aleho:
@jmsche Ah! Didn't consider there could have been a breaking change in the String component.

Kocal:
Note that it also broke the Translator demo on the Symfony UX website: 
<img width="1543" alt="image" src="https://github.com/symfony/ux/assets/2103975/b6bd42aa-e7f3-45c4-b74d-c53616fe6976">
<img width="504" alt="image" src="https://github.com/symfony/ux/assets/2103975/9122f087-fd86-46e9-a8ac-37c248a8f5fb">

smnandre:
Do we need to fix something or is this going to be fixed in next SF release ?

Kocal:
This has been reverted, let's wait for a new Symfony release

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

aleho:
Yes, this is fixed.

smnandre:
Thank you for this update!

PR #1951, opened 2024-07-01
[Dropzone] Remove event listeners on disconnect

aleho:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| License       | MIT

Events added in controller connect were not removed on disconnect. This would lead to events being triggered twice if a dropzone input is, e.g., moved in the DOM.

### DIFF
--- a/src/Dropzone/assets/dist/controller.js
+++ b/src/Dropzone/assets/dist/controller.js
@@ -1,11 +1,19 @@
 import { Controller } from '@hotwired/stimulus';
 
 class default_1 extends Controller {
+    initialize() {
+        this.clear = this.clear.bind(this);
+        this.onInputChange = this.onInputChange.bind(this);
+    }
     connect() {
         this.clear();
-        this.previewClearButtonTarget.addEventListener('click', () => this.clear());
-        this.inputTarget.addEventListener('change', (event) => this.onInputChange(event));
+        this.previewClearButtonTarget.addEventListener('click', this.clear);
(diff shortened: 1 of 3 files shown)


src/Dropzone/assets/dist/controller.js
```diff
@@ -3,10 +3,16 @@ import { Controller } from '@hotwired/stimulus';
 class default_1 extends Controller {
     connect() {
         this.clear();
-        this.previewClearButtonTarget.addEventListener('click', () => this.clear());
-        this.inputTarget.addEventListener('change', (event) => this.onInputChange(event));
+        this.clear = this.clear.bind(this);
```
norkunas: Isn't it better to bind in `initialize` stage?

aleho: Ah, yes, didn't think about that!

I'm never binding in connect and always use properties.

```js
class A {
    connect() {
        someElement.addEventListener('click', this._click);
    }

    _click = (event) => {
        …
    };
```

smnandre (approved):

kbond:
@aleho, is the failing translator test related to this change?

aleho:
@kbond Afaik that's related to the bug introduced by a change to the string component (breaking various parts of Symfony and the UX demos too).

(symfony/ux#1950 and symfony/symfony#57612)

kbond:
Good catch, thanks Alexander.

aleho:
Thank you @kbond!

PR #1957, opened 2024-07-04
[LiveComponent] Fix (de)hydration of composite and/or foreign ID entities

MatTheCat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | N/A
| License       | MIT

This PR fixes two issues (tests included) with the `DoctrineEntityHydrationExtension`:

- Entities with composite IDs couldn’t be hydrated because in this case the object manager needs the properties as keys. Throwing if the `0` key does not exist prevents this to work, so this PR removes this check.
- Dehydrated entities whose ID is a foreign key would bear the corresponding entity (or proxy), preventing their future correct hydration. This PR dehydrates ID values as well if needed.


kbond (approved):

kbond:
This is great, thanks Mathieu!

Issue #1839, opened 2024-05-10
AlpineJS specific attributes are not rendered (`@click` and `x-on:click`)


loevgaard:
I love the components, so thank you <3

I have a button component and I would like to be able to do:
```twig
<twig:Button value="Toggle" x-on:click="open = !open"/>
```

but the `x-on:click` attribute is not rendered with this Twig:

```twig
<button{{ attributes }}>{{ value }}</button>
```

instead it just strips the attribute and outputs:

```twig
<button>Toggle</button>
```

If I try to use the `@click` syntax, Twig throws an exception.

norkunas:
I think this collides with newly implemented [nested attributes](https://symfony.com/bundles/ux-twig-component/current/index.html#nested-attributes).
In your case it would be accessible with `attributes.nested('x-on')` which in your case is unexpected :)

kbond:
Ah yes, I think @norkunas is correct. Anyone have any thoughts on how to solve?

> If I try to use the @click syntax, Twig throws an exception.

I think this could be a html-twig component parsing issue.

smnandre:
What does the expection say ? In the end i'm not surprised as it's not valid HTML syntax. But if we can improve this let's do it :)

loevgaard:
> I think this collides with newly implemented [nested attributes](https://symfony.com/bundles/ux-twig-component/current/index.html#nested-attributes).
> In your case it would be accessible with attributes.nested('x-on') which in your case is unexpected :)

Now that you pointed me in the right direction, it's totally fine for me :)

I will let you guys close the issue if you believe this is not something you want to 'fix'. For me, when I know what I know now, it's perfect :)

Once again, thanks for a great component!

WebMamba:
I think this is legit issue we should keep it open. I don't really know how to solve it... Should we detect alpine attribute? Or should we add a prefix so that the attributes are ignore by  the nested? Should we add a prefix for every nested attribute like <twig:Form twig:row:label='password' />?

loevgaard:
I don't know if it's possible, but there's also the option to use another separator for nested attributes, e.g. dot:

```twig
<twig:Button value="Toggle" nested.attribute="Value"/>
```

But then again, that might conflict with something else...

WebMamba:
> But then again, that might conflict with something else...

Yes this is my worries!

kbond:
Yeah, no matter what syntax we use, the possibility for conflicts exists. Some quick ideas:

1. Make the nested separator configurable
[...]

PR #1959, opened 2024-07-04
[TwigComponent] escape nesting separator with `::`

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix (partial) #1839
| License       | MIT

If you require a Twig component attribute that includes a `:` (like `x-on:click`) you'll trigger the [nested attribute system](https://symfony.com/bundles/ux-twig-component/current/index.html#nested-attributes). This PR allows escaping `:` with `::`. So to render `x-on:click="!open"`, you can to write `x-on::click="!open"`.

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -21,7 +21,7 @@ use Symfony\WebpackEncoreBundle\Dto\AbstractStimulusDto;
  */
 final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Countable
 {
-    private const NESTED_REGEX = '#^([\w-]+):(.+)$#';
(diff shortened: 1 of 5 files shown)


src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -64,6 +64,8 @@ function (string $carry, string $key) {
                     $value = 'true';
                 }
 
+                $key = str_replace('::', ':', $key);
```
smnandre, line 67: In a regex to enforce it's only for nested attributes ? (so after a word / before another one) ?

smnandre:
So we close this one or do we want to have both ?

kbond:
I'm going to close this for now.

PR #1960, opened 2024-07-04
[TwigComponent] allow attributes to be prefixed with `@`

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix (partial) #1839
| License       | MIT

Allow attributes to be prefixed with `@` (like `@click="!open"`).


kbond:
Currently, this PR only adds a failing test demonstrating the problem reported in #1839. I could use some help to fix!

smnandre:
In [TwigPreLexer](https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/Twig/TwigPreLexer.php) , i guess `consumeAttributeName` should use another regexp than the one in `consumeComponentName`

But we need a lot of tests for this, as they cannot be class props !

PR #1966, opened 2024-07-06
[TwigComponent] Allow @ prefix in nested attributes

smnandre:
This was not compiled

```twig
<twig foo:@bar="baz" />
```

Should fix #1960


kbond:
Thanks Simon.

PR #1937, opened 2024-06-24
[Map] Create Map component

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #38 
| License       | MIT

Hi :)

This PR is a proposal for #38 a replacement for #39.

Symfony UX Map is a new Symfony UX component that makes it easy to create, customize and use interactive JavaScript maps.

The package ships with:
- a simple and generic way to create a map, with markers and info-windows
- two Bridge, [Leaflet](https://leafletjs.com/index.html) and [Google Maps](https://developers.google.com/maps/documentation/javascript)
- a way to define provider-specific options
- a nice and fluent API
- Flex recipes are also ready: https://github.com/symfony/recipes/pull/1329

# Example

## Bridge configuration

```env
# .env
UX_MAP_DSN=leaflet://default
```

```yaml
# config/packages/ux_map.yaml
ux_map:
    renderer: '%env(UX_MAP_DSN)%'
```

## Map creation

An example to render a map, custom center and zoom, some markers and info windows:
```php
namespace App\Controller;

use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Attribute\Route;
use Symfony\UX\Map\InfoWindow;
use Symfony\UX\Map\MapFactoryInterface;
use Symfony\UX\Map\Marker;
use Symfony\UX\Map\Point;

final class ContactController extends AbstractController
{
    #[Route('/contact')]
    public function __invoke(): Response
    {
            // 1. Create a new map instance
            $myMap = (new Map());
                ->center(new Point(46.903354, 1.888334))
                ->zoom(6)
            ;
    
            // 2. You can add markers, with an optional info window
            $myMap
                ->addMarker(new Marker(
                    position: new Point(48.8566, 2.3522), 
                    title: 'Paris'
                ))
                ->addMarker(new Marker(
                    position: new Point(45.7640, 4.8357), 
                    title: 'Lyon',
                    // With an info window
                    infoWindow: new InfoWindow(
                        headerContent: '<b>Lyon</b>',
                        content: 'The French town in the historic Rhône-Alpes region, located at the junction of the Rhône and Saône rivers.'
                    )
                ));
    
            // 3. And inject the map in your template to render it
            return $this->render('contact/index.html.twig', [
                'my_map' => $myMap,
            ]);
    }
}
```

## Map rendering 

[...]

PR #1969, opened 2024-07-07
Clean and homogenize `src/*/.gitignore`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Following @smnandre 's comment (https://github.com/symfony/ux/pull/1937#discussion_r1667323358), it looks like `src/*/.gitignore` files reference files that does not exist in the context of a Package. 
For example, ignoring `.php_cs.cache` from a package does not make sense, since this file is created by PHP-CS-Fixer at the **repository root**.

Now, all the `src/*/.gitignore` contain at least this list: 
```gitignore
/assets/node_modules/
/vendor/
/composer.lock
/phpunit.xml
/.phpunit.result.cache
```

### DIFF
--- a/src/Cropperjs/.gitignore
+++ b/src/Cropperjs/.gitignore
@@ -1,4 +1,5 @@
-vendor
-composer.lock
-.php_cs.cache
-.phpunit.result.cache
+/assets/node_modules/
+/vendor/
+/composer.lock
+/phpunit.xml
+/.phpunit.result.cache
--- a/src/Icons/.gitignore
+++ b/src/Icons/.gitignore
@@ -1,4 +1,7 @@
-/vendor
+/assets/node_modules/
+/vendor/
 /composer.lock
-/var
+/phpunit.xml
 /.phpunit.result.cache
+
+/var
--- a/src/Turbo/tests/app/.gitignore
+++ b/src/Turbo/tests/app/.gitignore
@@ -1,0 +1,2 @@
+/var/
+/public/build/
--- a/src/Autocomplete/.gitignore
+++ b/src/Autocomplete/.gitignore
@@ -1,5 +1,7 @@
+/assets/node_modules/
+/vendor/
 /composer.lock
 /phpunit.xml
-/vendor/
-/var/
 /.phpunit.result.cache
+
+/var
--- a/src/Chartjs/.gitignore
+++ b/src/Chartjs/.gitignore
@@ -1,4 +1,5 @@
+/assets/node_modules/
 /vendor/
-.phpunit.result.cache
-.php_cs.cache
-composer.lock
+/composer.lock
+/phpunit.xml
+/.phpunit.result.cache
--- a/src/Dropzone/.gitignore
+++ b/src/Dropzone/.gitignore
@@ -1,3 +1,5 @@
+/assets/node_modules/
 /vendor/
-.phpunit.result.cache
-composer.lock
+/composer.lock
+/phpunit.xml
+/.phpunit.result.cache
--- a/src/LazyImage/.gitignore
+++ b/src/LazyImage/.gitignore
@@ -1,4 +1,5 @@
-vendor
-composer.lock
-.php_cs.cache
-.phpunit.result.cache
+/assets/node_modules/
+/vendor/
+/composer.lock
+/phpunit.xml
+/.phpunit.result.cache
--- a/src/LiveComponent/.gitignore
+++ b/src/LiveComponent/.gitignore
@@ -1,5 +1,7 @@
+/assets/node_modules/
+/vendor/
 /composer.lock
 /phpunit.xml
-/vendor/
-/var/
 /.phpunit.result.cache
+
+/var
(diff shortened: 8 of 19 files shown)


src/Icons/.gitignore
```diff
@@ -1,4 +1,7 @@
-/vendor
+/assets/node_modules/
```
smnandre, line 1:
```suggestion
```

(same everywhere not needed :) )

src/Icons/.gitignore
```diff
@@ -1,4 +1,7 @@
...
+/vendor/
 /composer.lock
-/var
+/phpunit.xml
```
smnandre, line 4:
```suggestion
```

Not needed (if "you" add things locally, you can use local ignore)

src/Icons/.gitignore
```diff
@@ -1,4 +1,7 @@
...
 /.phpunit.result.cache
+
```
smnandre, line 6:
```suggestion
```

src/Cropperjs/.gitignore
```diff
@@ -1,4 +1,5 @@
-vendor
-composer.lock
-.php_cs.cache
-.phpunit.result.cache
+/assets/node_modules/
```
smnandre, line 1: Should be in your local GIT config
```suggestion
```

smnandre (commented):
Can we double-check the .gitattributes in the same pass ?

smnandre:
To me, i think this should be something like 


.gitignore
```
/vendor
/composer.lock
/.phpunit.result.cache
```



.gitattributes
```
/.git* export-ignore
/.symfony.bundle.yaml export-ignore
/assets/src export-ignore
/assets/test export-ignore
/phpunit.xml.dist export-ignore
/tests export-ignore
```

(without the assets line when/if not needed)


Probably also add when necessary
* vitesst.config (ex: turbo)
* docker-compose (ex: turbo)
* phpstan (ex: turbo)


Be careful Turbo has some custom CI ;)

src/Cropperjs/.gitignore
Kocal: Mmmh not really, some projects commit their `node_modules` folder,  plus I don't really think there is a lot of people that globally git ignore this folder

src/Icons/.gitignore
Kocal: It was on purpose, common files across all components, and some specific files

smnandre: Could be done without empty space ;)

src/Cropperjs/.gitignore
smnandre: Well let's add it only when necessary then 😅

smnandre:
So is there still something to agree on here? Can i help you on something @Kocal ?

Kocal:
I didn't take a look since, maybe tonight!

smnandre:
I don't like adding generated files (phpunit/phpcs/.. cache) ... **but**  if we use CI actions that commit (auto-CS ?) this makes sense.

But i also realised all of those are already ignored in the root .gitignore 

```.gitignore
.doctor-rst.cache
.php-cs-fixer.cache
node_modules
yarn-error.log
/composer.lock
/vendor
```

(we may add phpunit.result.cache here)

--- 


So, after your explanations, the package folders could contain:

### .gitignore

```.gitignore
# not sure if here .. maybe in the global gitignore
/.php-cs-fixer.cache 
/.phpunit.result.cache 

# is the root one not enough ?
/assets/node_modules/

/vendor/
/composer.lock
/phpunit.xml
```

(with aditional files depending on the package content)

### .gitattributes

```.gitignore
/.gitattributes export-ignore
/.gitignore export-ignore
/.symfony.bundle.yaml export-ignore
/assets/src export-ignore
/assets/test export-ignore
/phpunit.xml.dist export-ignore
/tests export-ignore
```

(with aditional files depending on the package content)


```.gitattributes
# And soon
/doc/ export-ignore # as no one want to read raw RST file in their vendor folder
/phpstan.dist.neon export-ignore # if not in the root repo
```

--- 

Side question @Kocal :
Would it be something complicated / with consequences to put the node_module in the root directory of each package ? Or maybe "not advised" ? 


---


So we _may_ only need to update the root .gitignore one for now :)

Kocal:
> Would it be something complicated / with consequences to put the node_module in the root directory of each package ? Or maybe "not advised" ?

Not complicated, but really not wanted. With this kind of project structure you want to use Yarn workspaces (added in #9 :D)

src/Turbo/tests/app/.gitignore
```diff
@@ -0,0 +1,2 @@
+/var/
+/public/build/
```
smnandre, line 2:
```suggestion
/public/build/
/node_modules/
/yarn.lock
```

Those were ignored and no more with your changes.

But i like the idea of a specific file

smnandre: Just double check Turbo and let's merge this one, i'll prepare a dedicated one for .gitattributes to not you to feel harassed :)

Thanks a lot for taking the time

smnandre:
Rebased / clean in #https://github.com/symfony/ux/pull/2243 .. LGTM !

PR #1976, opened 2024-07-11
[Fix] Handle `loading="lazy"` for LiveComponent only

### LINKED ISSUE #1975: `loading` attribute without Live Component
Hi,

I would like to use the `loading` attribute on my Twig Component but it adds `data-action="live:appear->live#$render"` which I do not want to have. How to avoid this? 

Thanks.

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1975
| License       | MIT

A TwigComponent with loading=lazy was interpreted as a live and its attributes were changed.


WebMamba (approved):

kbond:
Thanks Simon.

PR #1983, opened 2024-07-13
[Site] Leverage PHP-CS-Fixer parralel runner

smnandre:
(From symfony/symfony  #57687 😅 )


javiereguiluz:
Nice one! Thanks Simon.

PR #1984, opened 2024-07-13
[Icons] Add PHPStan

smnandre:
Add phpstan for Icon package

I fixed the easiest, but remains 7 errors

### DIFF
--- a/src/Icons/src/Command/ImportIconCommand.php
+++ b/src/Icons/src/Command/ImportIconCommand.php
@@ -52,7 +52,8 @@ final class ImportIconCommand extends Command
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
-        $names = $input->getArgument('names');
+        /** @var list<string> $names */
+        $names = (array) $input->getArgument('names');
         $result = Command::SUCCESS;
 
         foreach ($names as $name) {
@@ -81,6 +82,7 @@ final class ImportIconCommand extends Command
 
             $this->registry->add(\sprintf('%s/%s', $prefix, $name), $svg);
 
+            /** @var array<string, string> $license */
             $license = $this->iconify->metadataFor($prefix)['license'];
 
             $io->text(\sprintf(
--- a/src/Icons/src/Command/LockIconsCommand.php
+++ b/src/Icons/src/Command/LockIconsCommand.php
@@ -81,6 +81,7 @@ final class LockIconsCommand extends Command
 
             $this->registry->add(\sprintf('%s/%s', $prefix, $name), $svg);
 
+            /** @var array<string, string> $license */
             $license = $this->iconify->metadataFor($prefix)['license'];
             ++$count;
 
--- a/src/Icons/src/Command/SearchIconCommand.php
+++ b/src/Icons/src/Command/SearchIconCommand.php
@@ -79,7 +79,9 @@ EOF
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
+        /** @var string $prefix */
         $prefix = $input->getArgument('prefix');
+        /** @var string|null $name */
         $name = $input->getArgument('name');
 
         $iconSets = $this->findIconSets($prefix);
@@ -107,12 +109,18 @@ EOF
         if (1 < \count($iconSets) && $prefix !== $iconSet['prefix']) {
             $choices = array_combine(array_keys($iconSets), array_column($iconSets, 'name'));
             $choice = $io->choice('Select an icon set', array_values($choices));
-            if (!$choice || false === $prefix = array_search($choice, $choices, true)) {
+            if (!$choice || false === $prefix = array_search($choice, $choices, true) || !is_string($prefix)) {
                 $io->error('No icon set selected.');
 
                 return Command::INVALID;
             }
             $iconSet = $iconSets[$prefix];
+        }
+
+        if (!is_string($prefix) || !is_string($name)) {
+            $io->error('Invalid icon set or icon name.');
+
+            return Command::INVALID;
         }
 
         $io->write(\sprintf('Searching <comment>%s</comment> icons "<comment>%s</comment>"...', $iconSet['name'], $name));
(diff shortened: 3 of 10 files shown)


src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -64,17 +66,23 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+    /**
+     * @param array<string, mixed> $mergedConfig
+     */
+    protected function loadInternal(array $mergedConfig, ContainerBuilder $container): void
     {
         $loader = new PhpFileLoader($container, new FileLocator(__DIR__.'/../../config'));
         $loader->load('services.php');
 
-        if (isset($container->getParameter('kernel.bundles')['TwigComponentBundle'])) {
+        if (is_array($kernelBundles = $container->getParameter('kernel.bundles')) && isset($kernelBundles['TwigComponentBundle'])) {
```
Kocal, line 85: Can't we use ?? instead? This will saves us 2 function calls:
```suggestion
        if ($container->getParameter('kernel.bundles')['TwigComponentBundle'] ?? null) {
```

src/Icons/src/Command/ImportIconCommand.php
```diff
@@ -52,7 +52,8 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
-        $names = $input->getArgument('names');
+        /** @var list<string> $names */
+        $names = (array) $input->getArgument('names');
```
Kocal, line 56: I believe we can't use https://github.com/phpstan/phpstan-symfony ?

src/Icons/src/Command/ImportIconCommand.php
```diff
@@ -81,6 +82,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
             $this->registry->add(\sprintf('%s/%s', $prefix, $name), $svg);
 
+            /** @var array<string, string> $license */
             $license = $this->iconify->metadataFor($prefix)['license'];
```
Kocal, line 86: What about typing `Iconify::metadataFor()` return type instead?

src/Icons/src/Command/LockIconsCommand.php
```diff
@@ -81,6 +81,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
             $this->registry->add(\sprintf('%s/%s', $prefix, $name), $svg);
 
+            /** @var array<string, string> $license */
```
Kocal, line 84: Same thing here

src/Icons/src/Command/SearchIconCommand.php
```diff
@@ -107,14 +109,20 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         if (1 < \count($iconSets) && $prefix !== $iconSet['prefix']) {
             $choices = array_combine(array_keys($iconSets), array_column($iconSets, 'name'));
             $choice = $io->choice('Select an icon set', array_values($choices));
-            if (!$choice || false === $prefix = array_search($choice, $choices, true)) {
+            if (!$choice || false === $prefix = array_search($choice, $choices, true) || !is_string($prefix)) {
```
Kocal: Can be simplified to that, right?
```suggestion
            if (!$choice || !is_string($prefix = array_search($choice, $choices, true))) {
```

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -64,17 +66,23 @@ public function getConfigTreeBuilder(): TreeBuilder
         return $builder;
     }
 
+    /**
+     * @param array<string, mixed> $config
+     */
```
Kocal, line 71: Can't this be inferred from `Extension` ?

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: 85     Cannot access offset 'TwigComponentBundle' on array|bool|float|int|string|UnitEnum.

src/Icons/src/Command/ImportIconCommand.php
smnandre: I was thinking it was more for symfony "apps" than bundle, but maybe ?

src/Icons/src/Command/ImportIconCommand.php
smnandre: string, mixed :/

src/Icons/src/Command/LockIconsCommand.php
smnandre: Not sure what you mean here. Could you give me a hint ?

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: 69     Method Symfony\UX\Icons\DependencyInjection\UXIconsExtension::getConfiguration() has parameter $config with no value type specified in iterable type array.  
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type

src/Icons/src/Command/ImportIconCommand.php
Kocal: Nah that's more for Symfony app, with a Kernel/Container instance

src/Icons/src/DependencyInjection/UXIconsExtension.php
Kocal: I mean, typing it in https://github.com/symfony/symfony/blob/1a16ebc32598faada074e0af12a6a698d2964a5e/src/Symfony/Component/DependencyInjection/Extension/ConfigurationExtensionInterface.php#L29

smnandre: I'm all for it, but that's way outside what we can do here ^^

src/Icons/src/Command/ImportIconCommand.php
smnandre: Recent Symfony versions have some type-specific getter i think there. We just need to wait a bit :)

smnandre:
Should we already merge those fixes ?

Kocal:
@smnandre do you have a idea in mind, or PHPStan support is no longer relevant?

smnandre:
I'm not sure for now, to be honest... I might have misunderstood, but it seemed like there wasn't much enthusiasm for this

PR #1989, opened 2024-07-14
fix null props not overwrite parent context

### LINKED ISSUE #1970: [Twig Components] - scoped props ?
(i'm using only anonymous components)

Hey, the current behavior is that inner components can acces their parents props. This is useful but this creates some edge cases that can be annoying. For exemple, i'm creating a sort of UI library and some nested components can share attributes with same name like in my case : **icon** 
Here my badge can have an icon sometime but i dont want one in this case, so i have to set it to null
```twig
//--------- in both my components

{% props icon = null, ..... %}

{# ... some code  ... #}




//------- in my template

<twig:Nav:Link icon="folder-plus">

    {# ... what i'd like to do  ... #}
    <twig:Badge>.......</twig:Badge>

    {# ... what I have to do ( i guess ? )  ... #}
    <twig:Badge :icon="null">.......</twig:Badge>

</twig:Nav:Link>
```
I would like to avoid giving different name to same prop on each of my components ( badgeIcon, buttonIcon, ..... ) or that weird null thing 
Is there a way to make some specific props "private" or "scoped", as they would be defined in the components, and only accessible in this component ant not it's children.
In my ideal world, i think it would be great to have something like : 
```twig
{# could be a function or a fiter #}
{% props scoped(icon = null), ..... %}

{# or a second twig tag specific to this behavior #}  
{% scopedProps icon = null, ... %}
```

Does it make sense ?

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1970 
| License       | MIT

So weird! The code was there but at the wrong place!

### DIFF
--- a/src/TwigComponent/src/Twig/PropsNode.php
+++ b/src/TwigComponent/src/Twig/PropsNode.php
@@ -73,7 +73,23 @@ class PropsNode extends Node
                 ->raw(";\n")
                 ->outdent()
                 ->write('}')
-                ->write("\n");
+                ->write("\n")
+            ;
+
+            // overwrite the context value if a props with a similar name and a default value exist
+            if ($this->hasNode($name)) {
+                $compiler
+                    ->write('if (isset($context[\'__context\'][\''.$name.'\'])) {')
+                    ->raw("\n")
+                    ->indent()
+                    ->write('$context[\''.$name.'\'] = ')
+                    ->subcompile($this->getNode($name))
+                    ->raw(";\n")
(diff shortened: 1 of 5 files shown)


smnandre:
Could we add a test or two to prevent regression in the future ?

smnandre (approved):
Thanks!!

javiereguiluz:
Thanks for fixing this bug Matheo.

PR #1990, opened 2024-07-15
[LiveComponent] Allow updates in arrays of DTOs

### LINKED ISSUE #1953: [LiveComponent] Can not update parent component with array of DTOs
I started working with Live Components and noticed the following behaviour:

Imagine having a `ParentComponent` and a `ChildComponent`. `ParentComponent` holds an array of DTOs like that:

```php
#[AsLiveComponent]
final class ParentComponent
{
    use DefaultActionTrait;

    /**
     * @var IntWrapper[]
     */
    #[LiveProp(writable: true)]
    public array $values = [];

    public function mount(): void
    {
        $one = new IntWrapper();
        $one->value = 1;

        $two = new IntWrapper();
        $two->value = 2;

        $three = new IntWrapper();
        $three->value = 3;

        $this->values = [$one, $two, $three];
    }
}
```

The `ChildComponent` holds a single instance of the DTO:

```php
#[AsLiveComponent]
final class ChildComponent
{
    use DefaultActionTrait;

    #[LiveProp(writable: ["value"])]
    public IntWrapper $value;

    public function mount(): void
    {
        $this->value = new IntWrapper();
        $this->value->value = 10;
    }
}
```

I then render the `ParentComponent` with the `ChildComponent` inside:

`ParentComponent.html.twig`
```twig
<div {{ attributes }}>
    {% for key, val in values %}
        <span>{{ val.value }}</span>
        <twig:ChildComponent data-model="values[{{ key }}].value:value.value"/>
    {% endfor %}
</div>
```

`ChildComponent.html.twig`
```twig
<div {{ attributes }}>
    <span>nested:</span> <input type="number" data-model="value.value">
</div>
```

By binding `values[{{ key }}].value` to `value.value` I expected the `ParentComponent` to update, when I change the value in the input field of the `ChildComponent`, as described [in the docs](https://symfony.com/bundles/ux-live-component/current/index.html#updating-a-parent-model-from-a-child).

However the value in the `ParentComponent` is not updated. The problem is in the https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/LiveComponentHydrator.php#L365 where the property path is adjusted before setting the updated value.

In the presented example, the expected output of the `adjustPropertyPathForData` would be `"[0].value"` since we are changing a property of an object inside an array. However the actual output is `"[0][value]"`. The `$currentValue` inside the method is initially an array, but the value is not being updated as the `foreach`-loop processes the parts of the property path. 

I'm not sure if this is a bug, or if our use case is just not (yet) possible with the current state of Live Components. Possible workarounds are either working with arrays only, without using DTOs, or implementing `\ArrayAccess` on our DTOs. 

I hope I could make my issue clear, please let me know what you think about it.

cgrabenstein:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no 
| Issues        | Fix #1953  
| License       | MIT

This PRs allows changes in arrays of DTOs, see the linked issue for more background information.

There is one limitation though: For now it is not possible to generally set nested properties to be writable (i.e. something like `writable: 'items.*.prop'`, or `items[].prop`). You either have to set everything to writable (`writable: true`), or a very specific element (`writable: items.1.prop`).

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -370,6 +370,7 @@ final class LiveComponentHydrator
         foreach ($parts as $part) {
             if (\is_array($currentValue)) {
                 $finalPropertyPath .= \sprintf('[%s]', $part);
+                $currentValue = $this->propertyAccessor->getValue($rawPropertyValue, $finalPropertyPath);
 
                 continue;
             }
@@ -379,6 +380,10 @@ final class LiveComponentHydrator
             }
 
             $finalPropertyPath .= $part;
+
+            if (null !== $currentValue) {
+                $currentValue = $this->propertyAccessor->getValue($rawPropertyValue, $finalPropertyPath);
+            }
         }
 
         return $finalPropertyPath;
--- a/src/LiveComponent/tests/Fixtures/Dto/HoldsArrayOfDtos.php
+++ b/src/LiveComponent/tests/Fixtures/Dto/HoldsArrayOfDtos.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Dto;
+
+final class HoldsArrayOfDtos
+{
+    /**
+     * @var Address[] $addresses
+     */
+    public array $addresses;
+}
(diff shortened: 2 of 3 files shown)


WebMamba:
Thanks a lot for this PR @cgrabenstein! I was pretty sure most of the code was already one! Nice work! 🧡

WebMamba (approved):
I just tested, it work well! Thanks a lot @cgrabenstein to took the time to look at it! And to write good tests! See you soon! 🧡

cgrabenstein:
Thank you for checking it. What about the fabbot.io errors? I did not change the exception messages, so I'm not sure if I should apply the suggested patch.

kbond:
Great first contribution Christoph, thank you!

PR #1991, opened 2024-07-15
[Doc] Mention that boolean properties require the dynamic syntax

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

This confused me. On Symfony Slack folks mentioned that changing this behavior would be bad for DX (e.g. we can do it in components backend by a PHP class with typed properties but we cannot for anonymous components). So, I think it might be enough to mention this in the docs. Thanks.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -295,6 +295,18 @@ prefix the attribute with ``:`` or use the normal ``{{ }}`` syntax:
     // pass object, array, or anything you imagine
     <twig:Alert :foo="['col' => ['foo', 'oof']]" />
 
+Boolean props require using the dynamic syntax:
+
+.. code-block:: html+twig
+
+    {# in this example, the 'false' value is passed as a string
+       (so it's converted automatically to the true boolean value) #}
+    <twig:Alert message="..." withCloseButton="false" />
+
+    {# in the following examples, the 'false' value is passed as a boolean property #}
+    <twig:Alert message="..." :withCloseButton="false" />
+    <twig:Alert message="..." withCloseButton="{{ false }}" />
+
 Don't forget that you can mix and match props with attributes that you
 want to render on the root element:
 


Kocal:
That's remind this ESLint rule for Vue.js: https://eslint.vuejs.org/rules/no-boolean-default

Ideally when you conceive your component, you don't want to use boolean props with a default value to `true`,  to enforce HTML standard. 

Depending on whether or not you want your close button to be displayed by default, you will need to have a property `withCloseButton` or `withoutCloseButton` default to `false`, meaning that you can simply use a prop with the static syntax like this:
```twig
<twig:Alert message="..." withCloseButton/>

<twig:Alert message="..." withoutCloseButton />
```

src/TwigComponent/doc/index.rst
```diff
@@ -295,6 +295,18 @@ prefix the attribute with ``:`` or use the normal ``{{ }}`` syntax:
     // pass object, array, or anything you imagine
     <twig:Alert :foo="['col' => ['foo', 'oof']]" />
 
+Boolean props require using the dynamic syntax:
```
WebMamba, line 298: Instead of focusing just on bool, maybe we can do a full example of the dynamic syntax, with scalar, bool, array, and objects, ...

javiereguiluz: I think that the dynamic syntax is already explained for objects, arrays, expressions, etc.

The problem is that when using scalar values, only booleans are problematic: `someProp="foo"` (`foo` is a string in PHP), `someProp="5"` (will be casted to `(int) 5` in PHP) but `someProp="false"` will be casted to `true` in PHP because it's not parsed as a proper boolean value.

That was my reasoning for mentioning only this. Thanks.

WebMamba: Ha yes ok thanks @javiereguiluz for the explanation!

WebMamba (approved):

kbond:
Thanks Javier.

PR #1992, opened 2024-07-17
[LiveComponent] Add a new helper to interact with forms in functional tests

yoye:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| License       | MIT

Add a new method in `TestLiveComponent` to interact with live components that contains forms. Submitted data must be passed as an array like you'll do with `KernelBrowser`.

```php
$component = $this->createLiveComponent(name: Component::class);
$component->submitForm(['foo' => ['bar' => 'baz']]); // The method will change this values in ['foo.bar' => 'baz'] as expected by the component.
```

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3659,7 +3659,7 @@ uses Symfony's test client to render and make requests to your components::
 
             // call live action with file uploads
             $testComponent
-                ->save('processUpload', files: ['file' => new UploadedFile(...)]);
+                ->call('processUpload', files: ['file' => new UploadedFile(...)]);
 
             // emit live events
             $testComponent
@@ -3671,6 +3671,10 @@ uses Symfony's test client to render and make requests to your components::
             $testComponent
                 ->set('count', 99)
             ;
+
+            // Submit form data
+            $testComponent
+                ->submitForm(['form' => ['input' => 'value']], 'save');
 
             $this->assertStringContainsString('Count: 99', $testComponent->render());
 
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -125,6 +125,13 @@ final class TestLiveComponent
         return $this->client()->getResponse();
     }
 
+    public function submitForm(array $formValues, ?string $action = null): self
+    {
+        $flattenValues = $this->flattenFormValues($formValues);
+
+        return $this->request(['updated' => $flattenValues, 'validatedFields' => array_keys($flattenValues)], $action);
+    }
+
     private function request(array $content = [], ?string $action = null, array $files = []): self
     {
         $csrfToken = $this->csrfToken();
@@ -204,5 +211,20 @@ final class TestLiveComponent
         $this->performedInitialRequest = true;
 
         return $this->client;
+    }
+
+    private function flattenFormValues(array $values, string $prefix = ''): array
+    {
+        $result = [];
+
+        foreach ($values as $key => $value) {
+            if (\is_array($value)) {
+                $result += $this->flattenFormValues($value, $prefix.$key.'.');
+            } else {
+                $result[$prefix.$key] = $value;
+            }
+        }
+
+        return $result;
     }
 }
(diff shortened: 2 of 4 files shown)


kbond (approved):
This is beautiful! I thought adding this would be much more complex!

src/LiveComponent/doc/index.rst
```diff
@@ -3659,7 +3659,7 @@ uses Symfony's test client to render and make requests to your components::
 
             // call live action with file uploads
             $testComponent
-                ->save('processUpload', files: ['file' => new UploadedFile(...)]);
+                ->call('processUpload', files: ['file' => new UploadedFile(...)]);
```
smnandre, line 3662: Well spotted!

src/LiveComponent/doc/index.rst
```diff
@@ -3672,6 +3672,10 @@ uses Symfony's test client to render and make requests to your components::
                 ->set('count', 99)
             ;
 
+            // Submit form data
```
smnandre, line 3675: Maybe we should create an entire block "Test LiveComponent with Forms" (or something like that), wdyt ? 

If "no", i think some comment to precise what is "form" and "save" here could avoid some missunderstandings.

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -205,4 +212,19 @@ private function client(): KernelBrowser
...
         return $this->client;
     }
+
+    private function flattenFormValues(array $values, string $prefix = ''): array
+    {
+        $result = [];
+
+        foreach ($values as $key => $value) {
+            if (\is_array($value)) {
+                $result += $this->flattenFormValues($value, $prefix.$key.'.');
```
smnandre, line 222: No difference array_is_list / map ? (geniune question)

Does this work with fields accepting multiple values ?

smnandre:
I let some questions about details, but overall very happy to see this PR: thank you very much @yoye ! 

And i second @kbond, I thought it would be much more complex so .. 👏

src/LiveComponent/doc/index.rst
yoye: As Test helper is only one block, I didn't want to change that. But I understand that my array is a bit confusing.

src/LiveComponent/src/Test/TestLiveComponent.php
yoye: Don't know how to use `array_is_list` here. With `array_map` I can't handle keys.

Yes, I was a little undecided to not add a test for that method. At first I've thinked to add a public static method in `Symfony\UX\LiveComponent\Util\LiveFormUtility` which could be easily tested. But I don't know if it make sence as it will be created only for test purpose.

But this will work. 
```php
    public function testFlatten(): void
    {
        $data = [
            'foo' => ['bar' => 'baz'],
            'collection' => ['one', 'two', 'three'],
            'collection_with_sub' => [
                'sub_one' => ['child_one', 'child_two'],
                'sub_two' => ['child_three', 'key' => 'child_four'],
            ],
        ];

        $this->assertSame([
            'foo.bar' => 'baz',
            'collection.0' => 'one',
            'collection.1' => 'two',
            'collection.2' => 'three',
            'collection_with_sub.sub_one.0' => 'child_one',
            'collection_with_sub.sub_one.1' => 'child_two',
            'collection_with_sub.sub_two.0' => 'child_three',
            'collection_with_sub.sub_two.key' => 'child_four',
        ], TestLiveComponent::flattenFormValues($data));
    }
```

src/LiveComponent/doc/index.rst
smnandre: Let's keep as it is.. 

Who knows, mabe if you add other nice helpers...  we will be forced to add a dedicated block ? :))

src/LiveComponent/src/Test/TestLiveComponent.php
smnandre: Good for me then, thank you for taking time to answer :)

smnandre (approved):
Thanks!

kbond:
So great, thanks @yoye!

PR #1994, opened 2024-07-17
[Icon] Emphasis HTML syntax requires TwigComponent

### LINKED ISSUE #1993: [Icons][Twig-Component] Update documentation for installation ux icons
Hi everyone,

I just initialized a new symfony project, and I wanted to first install the ux icons library but it doesn't work until twig components is installed.

You should add at the beginning of the ux icons documentation that you must first install ux twig component then ux icons

smnandre:
First time HTML syntax is shown, use "tip" (green borders) instead of "note" 

Fix: #1993

| Before | After |
| - | - |
| Capture d’écran 2024-07-17 à 21 17 54 - Grande | Capture d’écran 2024-07-17 à 21 29 14 |

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -180,7 +180,7 @@ tag:
     <!-- you can also add any HTML attributes -->
     <twig:ux:icon name="user-profile" height="16" width="16" aria-hidden="true" />
 
-.. note::
+.. tip::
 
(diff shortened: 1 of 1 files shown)


94noni (approved):
Makes sens indeed

javiereguiluz (approved):
Yes, this is better. Thanks!

PR #1999, opened 2024-07-19
[TwigComponent] Add components for Symfony forms

### LINKED ISSUE #1010: [TwigComponent] Provide form components for UX
Symfony form with twig provides many `form_*` functions, but that's ugly when combining with new html-like component syntax.
Maybe TwigComponent could provide generic components for these functions?
So for example we could write:
```twig
<twig:Form :form="form">
  <twig:FormErrors :form="form" />
  <twig:FormRow :form="form.fullName" />
  <div>
    <twig:FormLabel :form="form.fullName" />
    <twig:FormWidget :form="form.fullName" />
    <twig:FormErrors :form="form.fullName" />
  </div>
</twig:Form>
```
etc.

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1010
| License       | MIT

```twig
<twig:Form :form="form" />
```

```twig
<twig:Form :form="form">
  <twig:Form:Errors :field="form" />
  <twig:Form:Row :field="form.name" />
</twig:Form>
```

```twig
<twig:Form :form="form">
  <twig:Form:Errors :field="form" />
  <div>
    <twig:Form:Label :field="form.name" label="Your name" />
    <twig:Form:Errors :field="form.name" />
    <twig:Form:Widget :field="form.name" />
    <twig:Form:Help :field="form.name" />
  </div>
</twig:Form>
```

### DIFF
--- a/src/TwigComponent/composer.json
+++ b/src/TwigComponent/composer.json
@@ -38,6 +38,7 @@
         "symfony/css-selector": "^5.4|^6.0|^7.0",
         "symfony/dom-crawler": "^5.4|^6.0|^7.0",
         "symfony/framework-bundle": "^5.4|^6.0|^7.0",
+        "symfony/form": "^5.4|^6.0|^7.0",
         "symfony/phpunit-bridge": "^6.0|^7.0",
         "symfony/stimulus-bundle": "^2.9.1",
         "symfony/stopwatch": "^5.4|^6.0|^7.0",
--- a/src/TwigComponent/config/form_components.php
+++ b/src/TwigComponent/config/form_components.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\DependencyInjection\Loader\Configurator;
+
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symfony\UX\TwigComponent\Twig\Form\FormComponent;
+use Symfony\UX\TwigComponent\Twig\Form\FormErrorsComponent;
+use Symfony\UX\TwigComponent\Twig\Form\FormHelpComponent;
+use Symfony\UX\TwigComponent\Twig\Form\FormLabelComponent;
+use Symfony\UX\TwigComponent\Twig\Form\FormRowComponent;
+use Symfony\UX\TwigComponent\Twig\Form\FormWidgetComponent;
+
+return static function (ContainerConfigurator $container) {
+    $container->services()
+        ->set('ux.twig_component.form_component', FormComponent::class)
+            ->tag('twig.component', [
+                'key' => 'Form',
+                'template' => '@TwigComponent/Component/Form.html.twig',
+                'expose_public_props' => true,
+                'attributes_var' => 'attributes',
+            ])
+
+        ->set('ux.twig_component.form_row_component', FormRowComponent::class)
+            ->tag('twig.component', [
+                'key' => 'Form:Row',
+                'template' => '@TwigComponent/Component/FormRow.html.twig',
+                'expose_public_props' => true,
+                'attributes_var' => 'attributes',
+            ])
+
+        ->set('ux.twig_component.form_label_component', FormLabelComponent::class)
+            ->tag('twig.component', [
+                'key' => 'Form:Label',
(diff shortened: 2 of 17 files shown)


smnandre:
Hey @norkunas ! 

**TL;DR;** i like this, but think it's not something we should have in the TwigComponent but more in the incoming "recipe" system. 

---

This features clearly has a lot of supporters, and i can see the _sexyness_ in its syntax.

My reaction across different "scopes" :) 

## Component Implementation

In term of code, not sure if you really need classes there, as they do nothing more than an anonymous component.
I suppose it's because we still are discussing anonymous components namespacing/importing ? 

Also, i would use "form" or "form_view" instead of "form" to be coherent with Symfony form functions / theming.


```
{{ form_label(field, label, { label_attr: attributes.all() }) }}
```

The problem here for me is that this offer a lot less possibilities than their Twig functions / blocks equivalents.

Lastly, in term of performance, as the TwigComponent work today, this could have a negative impact on performances.. if this is "just" a syntaxic sugar, i wonder if we should not compile them directly their Twig function / blocks ? 


## UX Bundle

We cannot "reserve" the <twig:Form prefix like this, as many people already use it in their apps.

I also believe (not tested to be honest) it would work weirdly with the form theming / overriding in Symfony

But above all, i really think we should orientate (re-orientate?) this bundle as a fondation for other ones, not as concrete solution/bundles.

One day (difficult _right now_ for multiple reasons, but i'm 100% convinced this is the good direction) i'd like to move everything linked with Forms in one single UXForm component.

## Recipe ?

The recipe system would allow us to make available components, code to install in userland. And a `<twig:Form>` syntax was one of the first @kbond and @weaverryan discussed back then :) 

@kbond @WebMamba are _very_ motivated to work on this system after holiday, and i'm also eager to find good DX for people to use it.

--- 

So my _personal_ opinion on this PR: i don't think we should merge it in TwigComponent right now, but i'd love we have something very similar as one of our first recipes.

In the meantime, we also could show in documentation and/or the website how one user can do this if they want to ?

norkunas:
> In term of code, not sure if you really need classes there, as they do nothing more than an anonymous component.

Anonymous components, does not add strict typehints, so I chose class based component to have proper typehints for props.

> The problem here for me is that this offer a lot less possibilities than their Twig functions / blocks equivalents.

Can you elaborate what are less possibilities with components?
There is always some restrictions if you use something higher level, nobody forces to it, you can always fallback to old way.

> Lastly, in term of performance, as the TwigComponent work today, this could have a negative impact on performances.. if this is "just" a syntaxic sugar, i wonder if we should not compile them directly their Twig function / blocks ?

Probably I agree :thinking: but forms are not the components that are repeatedly 100 times per page..

> We cannot "reserve" the <twig:Form prefix like this, as many people already use it in their apps.

That's why I think we should not have a global registry and allow component imports #801.

My aim was to trigger a discussion about this, so thank you

smnandre:
> Anonymous components, does not add strict typehints, so I chose class based component to have proper typehints for props.

That's a valid point, even if one could add the check is done right after in the method. But not the same DX i agree.

> That's why I think we should not have a global registry and allow component imports #801.

Did not saw #801 ... but i can tell you, with HTML syntax with no prefix and dynamic import names it's almost impossible, as it would be a hell to parse.. 

.. but on a similar PR i was saying this could be done, but only with NodeVisitors (like trans_default_domain or form_theme for instance) .... and only per template (needing a lot of those tags when people do embeds (same pb with trans_default_domain or form_theme ;) )

> Can you elaborate what are less possibilities with components?

You prevent any form theme to be used i guess.. but also you allow only 'attr' to be passed, but form_row should accept label argument too for instance (same for all methods)

> My aim was to trigger a discussion about this, so thank you

Very happy to have this discussion too! I love when "end users" explain their need and bring differents solutions to the table :)

norkunas:
> You prevent any form theme to be used i guess..

That's not true :) you can apply theme on the whole form or on different fields in the same way as before, I even tested on a real project :)

> but also you allow only 'attr' to be passed, but form_row should accept label argument too for instance (same for all methods)

I missed it, but I think that's doable

smnandre:
> you can apply theme on the whole form or on different fields in the same way as before, I even tested on a real project :)

Well that's a good news! 

Just to check some "not-frequent-but" use cases 

```
{# templates/contact.html.twig #}
 
{% form_theme form 'my_theme.html.twig' %}

<twig:Form :form="form">
  <twig:Form:Errors :field="form" />
   {# ... #}
  
  <twig:Form:Row :field="form.name" />

</twig:Form>

```

Form errors would use `my_theme` here ?  Will "form.name" too ? 

If you define a block in templates/contact and use form_theme with self, same thing ?

PR #2001, opened 2024-07-20
Fix DOMException error when trying to remove non-existent classes

alexandrunastase:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | - 
| License       | MIT

### Description
In some cases the ElementChanges.applyElement tries to remove classes, but gives none as argument. This causes a DOMException which breaks the rest of the flow. 

image

### Fix:
I added a check to ensure the remove method only gets called when there's something to be removed.

### Versions
ux-live-component version: 2.18.1
ux-turbo version: 2.18.0
@hotwired/turbo version: 8.0.5

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1670,7 +1670,9 @@ class ElementChanges {
     }
     applyToElement(element) {
         element.classList.add(...this.addedClasses);
-        element.classList.remove(...this.removedClasses);
+        if (this.removedClasses.size > 0) {
+            element.classList.remove(...this.removedClasses);
+        }
         this.styleChanges.getChangedItems().forEach((change) => {
             element.style.setProperty(change.name, change.value);
             return;
--- a/src/LiveComponent/assets/src/Rendering/ElementChanges.ts
+++ b/src/LiveComponent/assets/src/Rendering/ElementChanges.ts
@@ -64,7 +64,9 @@ export default class ElementChanges {
 
     applyToElement(element: HTMLElement): void {
         element.classList.add(...this.addedClasses);
-        element.classList.remove(...this.removedClasses);
+        if (this.removedClasses.size > 0) {
+            element.classList.remove(...this.removedClasses);
+        }
 
         this.styleChanges.getChangedItems().forEach((change) => {
             element.style.setProperty(change.name, change.value);


WebMamba (commented):
Hey @alexandrunastase thanks a lot for taking the time to work on this PR! Do you have any reproducers for this weird behavior?  We should add a test also

smnandre:
I'd be curious to see the use case too ! 

classList.remove should accept an iterator (not an empty string but that should not be the case there...)

https://dom.spec.whatwg.org/#ref-for-dom-domtokenlist-remove

smnandre:
I just checked and ... this is entirely valid


```js
const foo = document.body;

console.log(foo.classList.toString())

foo.classList.remove(...[]);
foo.classList.add(...[]);

foo.classList.add('bar');
console.log(foo.classList.toString())

foo.classList.remove(...[]);
console.log(foo.classList.toString())
```


So there may be some "space pollution" here... if you could share a bit more code  that'd be great ... or ideally a small reproducer as @WebMamba   suggested. Thank you!

alexandrunastase:
Thanks for the feedback. It's a slightly more complex setup, with quite a few dependencies, but I'll try to re-create it.

alexandrunastase:
@smnandre  @WebMamba  You can find a reproducer project here: https://github.com/alexandrunastase/reproduce-dom-exception . Steps to reproduce are in the README.md file. I'll look into the tests in the coming days.

From what I noticed the issue is partially caused also by the fact that I broke up the classes on two lines for the date. In the reproducer it looks silly, but in the real project there were a lot of tailwind classes that made sense to be broken up. Some for example were related to the focus/peer behavior.

alexandrunastase:
So I looked a bit more into the issue and it seems there were two problems. One was passing empty string to the set, causing the error I posed in the initial message, and there were another issue that was appearing after fixing that one caused by classes with a "\n" ending. 

Both of this issues are fixed in https://github.com/symfony/ux-live-component/commit/a1a26c249c4b71e4425fffd88278f6286971944a , which is already merged to 2.x branch. 

I'm not familiar with how the release process for the ux-live-component works, but would nice if there would be a new tag that would contain these changes also. 

In retrospect, I'm not sure how initially the DOMException was fixed by the change made in the PR, but if I see that one again, I'll create another issue for it. Meanwhile I'll close the PR and wait for the new tag to be created. Thanks for your support and for looking into this!

WebMamba:
Thanks a lot for your time! Even if this PR has been close, I really appreciate your work! See you around!

PR #2006, opened 2024-07-24
[React] Improve error handling in `resolveReactComponent`

teklakct:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | 
| License       | MIT

## WHAT

Improve error handling and readability in `window.resolveReactComponent` function

- Enhance error messages for clarity:
  - Indicate when a module exists but lacks a default export.
  - (still) List available controllers when a specified controller does not exist.

With those changes, when resolving a module with no default export, an error message with a hint will be shown: 

```
React controller "${name}" could not be resolved. Ensure the module exports the controller as default.
```

## WHY

I found that when someone creates a module without a default export then the error message is quite confusing. 
For example:
```js
// NoDefaultExportComponent.jsx
export const SomeRandomName = () => {
    return <div>Hello</div>;
}
```
Whenever there will be an element like this 

```html
<div data-controller="NoDefaultExportComponent">
   ...
</div>
```
An error like below in console appear:

> Error connecting controller
>
> React controller "**NoDefaultExportComponent**" does not exist. Possible values: SomeComponent, AnotherComponent, **NoDefaultExportComponent**

Which is not exactly true. The module `NoDefaultExportComponent` was found but there is no required default export.

### DIFF
--- a/src/React/assets/src/register_controller.ts
+++ b/src/React/assets/src/register_controller.ts
@@ -37,6 +37,12 @@ export function registerReactControllerComponents(context: __WebpackModuleApi.Re
             const possibleValues = Object.keys(reactControllers).map((key) =>
                 key.replace('./', '').replace('.jsx', '').replace('.tsx', '')
             );
+
+            if (possibleValues.includes(name)) {
+                throw new Error(`
+                    React controller "${name}" could not be resolved. Ensure the module exports the controller as a default export.`);
+            }
+
             throw new Error(`React controller "${name}" does not exist. Possible values: ${possibleValues.join(', ')}`);
         }
 
--- a/src/React/CHANGELOG.md
+++ b/src/React/CHANGELOG.md
@@ -1,9 +1,13 @@
 # CHANGELOG
 
+## 2.26.0
+
+-   Improve error handling when resolving a React component
+
 ## 2.21.0
 
--   Add `permanent` option to the `react_component` Twig function, to prevent the 
-    _unmounting_ when the component is deconnected and immediately re-connected. 
+-   Add `permanent` option to the `react_component` Twig function, to prevent the
+    _unmounting_ when the component is deconnected and immediately re-connected
 
 ## 2.13.2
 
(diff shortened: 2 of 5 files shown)


teklakct:
Alternatively function `importAllReactComponents` may omit every module without default export. Something like this

```diff
    const importAllReactComponents = (r: __WebpackModuleApi.RequireContext) => {
-        r.keys().forEach((key) => (reactControllers[key] = r(key).default));
+        r.keys().forEach((key) => {
+            if (r(key).default !== undefined) {
+                reactControllers[key] = r(key).default;
+            }
+        });
    };
```

However, with these changes, it will be more difficult to determine if a module or default export is missing.

src/React/assets/src/register_controller.ts
```diff
@@ -37,6 +37,13 @@ export function registerReactControllerComponents(context: __WebpackModuleApi.Re
             const possibleValues = Object.keys(reactControllers).map((key) =>
                 key.replace('./', '').replace('.jsx', '').replace('.tsx', '')
             );
+
+            if (possibleValues.includes(name)) {
+                throw new Error(
+                    `React controller "${name}" could not be resolved. Ensure the module exports the controller as default.`
+                );
```
smnandre:
```suggestion
                throw new Error(`React controller "${name}" could not be resolved. Ensure the module exports the controller as default.`);
```

smnandre:
Hi @teklakct ! Thanks for this 

Could you rebuild the JS dist files ? (that's the reason of the CI failure)

From the root of the monorepo: `yarn run build`

(you may need to rebase before)

src/React/assets/src/register_controller.ts
teklakct: I ran `biome check` and applied the suggested formatter changes

teklakct:
@smnandre sorry for the delay. I must have missed the notification earlier.

smnandre:
@teklakct don't be :) I also lost tracking of this PR to be honest! 

Seems good to me !

smnandre (approved):

Kocal (approved):
Thanks for working on this, and sorry for the delay.

Do you think we could do the same for UX Vue and UX Svelte?

Kocal:
Thank you @teklakct.

PR #2013, opened 2024-07-30
[LiveComponent] Fix issue with onUpdated not triggered on nested array proprety

TheDutchScorpion:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | 
| License       | MIT

When modifying a array property the `onUpdated` hook is only triggered when the exact key is specified. 
```
#[LiveProp(writable: true, onUpdated: [‘firstname’ => ‘onUpdated’, ‘lastname’ => ‘onUpdated’]]
```

This PR will add the ability to use wildcards within the key. 
```
#[LiveProp(writable: true, onUpdated: [‘*’ => ‘onUpdated’]]
```
Or
```
#[LiveProp(writable: true, onUpdated: [‘values.*’ => ‘onUpdated’]]
```


smnandre:
Hi @TheDutchScorpion:  thanks for this PR.

Could you add some test here?

(especially one with something that did not work before and now would work?)

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -663,7 +663,8 @@ private function processOnUpdatedHook(object $component, string $frontendName, L
 
         foreach ($onUpdated as $propName => $funcName) {
             if (LiveProp::IDENTITY === $propName) {
-                if (!$dehydratedUpdatedProps->hasPropValue($frontendName)) {
+                if (!$dehydratedUpdatedProps->hasPropValue($frontendName)
+                    && !$dehydratedUpdatedProps->hasNestedPathsForProperty($frontendName)) {
                     continue;
```
smnandre: I'm not sure this is the right place... for a LiveProp IDENTITY the only thing that can change is... the prop itself no ? Is this not something we should do line 680+ ?

TheDutchScorpion: I don't think thats the best place either to check if there are changes to a array property. My intention was to tigger the onUpdated whenever there is a change within the array without providing a specific nested path.

TheDutchScorpion: Maybe I'm wrong, but it's also strange that if the `$propMetadata->onUpdated()` is a string, the update hook only triggers if the property has been completely changed to another property and not if part of it has been modified.

smnandre: I get that, but i fear doing so here will have side effects .. wdyt ?

smnandre:
I just saw you changed a bit your PR :) 

Could you update your first message with a basic case scenario, describing what you wanted todo, what did not work and what will after your PR ? 

It's easier for everyone to review and understand the PR :)

kevinmade:
Could this branch be reviewed and merged?

kevinmade:
@smnandre

smnandre:
I'd like to get the real use case behind this, as i don't see right now how a map can be mapped to a form (without it being a collection)

Then i'd like to see is before/after for 
* basic list of scalars (ints, strings)
* nested arrays
* free map (like this one) 
* collection of objects

I think we need to restrict this syntax to "*" => "method", or that does not make a lot of sense to me.

At core, this is a callback when a props is updated, triggered because the value will change. That does not feels logical to call multiple times the same function.

Other comments regarding code :)

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -686,13 +686,13 @@ private function processOnUpdatedHook(object $component, string $frontendName, L
...
             $key = \sprintf('%s.%s', $frontendName, $propName);
-            if (!$dehydratedUpdatedProps->hasPropValue($key)) {
-                continue;
-            }
+            $fullPaths = $dehydratedUpdatedProps->searchFullPathsForProperty($key);
 
-            $this->ensureOnUpdatedMethodExists($component, $funcName);
-            $propertyOldValue = $dehydratedOriginalProps->getPropValue($key);
-            $component->{$funcName}($propertyOldValue);
+            foreach ($fullPaths as $fullPath) {
```
smnandre, line 691: This loop concern one callback for a prop. It should not be called multiple times. I this you could "just" replace the prevous "hasPropValue" to some "matchPropValue" or something like this and let the rest flow.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -686,13 +686,13 @@ private function processOnUpdatedHook(object $component, string $frontendName, L
...
+                $this->ensureOnUpdatedMethodExists($component, $funcName);
+                $propertyOldValue = $dehydratedOriginalProps->getPropValue($fullPath);
```
smnandre, line 693: Let's not compute this every loop

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -686,13 +686,13 @@ private function processOnUpdatedHook(object $component, string $frontendName, L
...
+                $component->{$funcName}($propertyOldValue);
```
smnandre, line 694: Here the receiving callable as no way to get what is "old value". If a "*" is used, we should pass the full prop

src/LiveComponent/src/Util/DehydratedProps.php
```diff
@@ -113,6 +113,13 @@ public function getNestedPathsForProperty(string $prop): array
         return $nestedPaths;
     }
 
+    public function searchFullPathsForProperty(string $prop): array
```
smnandre, line 116: Let's add some tests if possible (with failing / problematic cases, not just happy path)

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -686,13 +686,13 @@ private function processOnUpdatedHook(object $component, string $frontendName, L
             }
 
             $key = \sprintf('%s.%s', $frontendName, $propName);
-            if (!$dehydratedUpdatedProps->hasPropValue($key)) {
-                continue;
-            }
+            $fullPaths = $dehydratedUpdatedProps->searchFullPathsForProperty($key);
```
smnandre, line 689: Let's do this only under conditions, and keep previous code if no "*"

smnandre:
Upon rereading my message, I realize it came across more strongly than I intended—apologies for that. Sorry as well for the delayed response, and thank you for your contribution @TheDutchScorpion and for the poke @kevinmade

Kocal:

PR #1996, opened 2024-07-19
[Translator] Fix CS on TranslatorCompilerPass

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Issues        | Fix #... 
| License       | MIT

PR #2014, opened 2024-07-30
chore: check if fabbot is correctly working

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Issues        | Fix #... 
| License       | MIT

PR opened for Fabien.

 Following #1996, we need to understand if something is wrong with fabbot or not.
⚠️ **This PR should not be merged!**

### DIFF
--- a/src/Icons/src/Command/LockIconsCommand.php
+++ b/src/Icons/src/Command/LockIconsCommand.php
@@ -1,14 +1,5 @@
 <?php
 
-/*
- * This file is part of the Symfony package.
- *
- * (c) Fabien Potencier <fabien@symfony.com>
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 namespace Symfony\UX\Icons\Command;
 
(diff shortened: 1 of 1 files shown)


javiereguiluz:
`fabbot` seems to be working ... so maybe it was a temporary error?

Kocal:
fabbot is green when it's supposed to be red, the code here is not valid (run PHP-CS-Fixer locally and see the diff). 

But I'm already seeing this problem with Fabien!
This afternoon he fixed the issue `Unable to parse the PHP CS Fixer configuration: %sSandboxed code attempted to call non-whitelisted class: PhpCsFixer\\Runner\\Parallel\\ParallelConfigFactory`, but there is still more one issue.

smnandre:
(side topic: if we also could make a global pass with fabbot, to fix all the exceptions format at once, that'd be great)

Kocal:
Okay, here fabbot is working as expected: 
image


Closing for now, let's observe other PR :eyes:

PR #2021, opened 2024-08-01
[Doc] Minor syntax fix in Twig Components docs

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

The code blocks were unnecessarily indented and the rendered output wasn't correct.

The usage of `#.` for numbered lists is not mandatory, but it's generally recommended because that way the list items are autonumbered and you can add or remove items later without taking care of renumbering the other items.


smnandre (approved):
Thanks you @javiereguiluz 😄

WebMamba (approved):

Issue #1150, opened 2023-09-28
[LiveComponent] Invalid CSRF token exceptions are not correctly managed


simondaigre:
**Scenario** 
- Enable CSRF in Symfony
- Without being logged in, browse a page with a Symfony form
- Fill the form then submit it
- Go to /login, login normally then go back to the form with your browser and re-submit it

**Actual behaviour (without Live component)** 
- Form error is correctly displayed with HTTP 422 and "Invalid CSRF token" message.

**Actual behaviour (with form in a Live component)**
- Error 500 is thrown and a popup appears with the exception

Error is thrown here : https://github.com/symfony/ux/blob/d993b1283e32aa190553261e9fda0a3d4e72e0f1/src/LiveComponent/src/EventListener/LiveComponentSubscriber.php#L112-L117

This issue occurs because CSRF are reset on each login/logout (which is normal behaviour)
I think we need to find a better way to handle this, since in prod environnements this leads to a 500 error to end users.

kbond:
Ok, so it's not the form's csrf that's causing the problem but the live component's csrf...

I agree we should provide a better method to handle this but not quite sure how yet. I think, currently, you can handle with a [component js hook](https://symfony.com/bundles/ux-live-component/current/index.html#javascript-component-hooks) but it would be complex. Feels like we could use something similar to loading states but for errors.

simondaigre:
@kbond thanks for the idea. Possible workaround here : https://github.com/symfony/symfony/pull/51724

smnandre:
Not sure it would be a solution viable for the entire "symfony stack", but in UX packages, i think a lazy load of the CSRF would not be too hard ... and (yeah i'm a broken record) this would help performance / sobriety. (the CSRF token on public page restrict the caching strategies)

gremo:
I have the same problem, getting random invalid csfr token on components with data-pool enabled. Does any solution exist?

smnandre:
@gremo thank you for reporting this! 

Could you open a new issue please with ideally a small reproducer ?

gremo:
@smnandre I currently receiving those errors in my inbox from an app that is already published. I deleted those emails a few moments ago 😄 but I not sure that the strack trace alone would be helpuful. New errors expected soon btw.

Right now I can't replicate that error while developing 😢

smnandre:
Ok i'm closing this issue for now, and you create a new one when you have those emails again, ok for you ? 
(it's easier for us to maintain / track)

[...]

PR #2022, opened 2024-08-01
[LiveComponent] Update CSRF token after component request

tijnema:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Issues        |
| License       | MIT

When CSRF is enabled on live components, each request will received a new CSRF token in the response. Currently all subsequent requests made by live components will use the initial CSRF token received. This PR solves this, by updating the CSRF token with the one received in the last response.

### DIFF
--- a/src/LiveComponent/assets/dist/Backend/Backend.d.ts
+++ b/src/LiveComponent/assets/dist/Backend/Backend.d.ts
@@ -13,6 +13,7 @@ export interface BackendInterface {
     }, files: {
         [key: string]: FileList;
     }): BackendRequest;
+    updateCsrfToken(csrfToken: string): void;
 }
 export interface BackendAction {
     name: string;
@@ -28,4 +29,5 @@ export default class implements BackendInterface {
     }, files: {
         [key: string]: FileList;
     }): BackendRequest;
+    updateCsrfToken(csrfToken: string): void;
 }
--- a/src/LiveComponent/assets/dist/Backend/RequestBuilder.d.ts
+++ b/src/LiveComponent/assets/dist/Backend/RequestBuilder.d.ts
@@ -2,7 +2,7 @@ import type { BackendAction, ChildrenFingerprints } from './Backend';
 export default class {
     private url;
     private method;
-    private readonly csrfToken;
+    private csrfToken;
     constructor(url: string, method?: 'get' | 'post', csrfToken?: string | null);
     buildRequest(props: any, actions: BackendAction[], updated: {
         [key: string]: any;
@@ -15,4 +15,5 @@ export default class {
         fetchOptions: RequestInit;
     };
     private willDataFitInUrl;
+    updateCsrfToken(csrfToken: string): void;
 }
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -2052,6 +2052,9 @@ class Component {
                 return response;
             }
             this.processRerender(html, backendResponse);
+            if (this.element.dataset.liveCsrfValue) {
+                this.backend.updateCsrfToken(this.element.dataset.liveCsrfValue);
+            }
             this.backendRequest = null;
             thisPromiseResolve(backendResponse);
             if (this.isRequestPending) {
@@ -2325,6 +2328,9 @@ class RequestBuilder {
         const urlEncodedJsonData = new URLSearchParams(propsJson + updatedJson + childrenJson + propsFromParentJson).toString();
         return (urlEncodedJsonData + params.toString()).length < 1500;
     }
+    updateCsrfToken(csrfToken) {
+        this.csrfToken = csrfToken;
+    }
 }
 
 class Backend {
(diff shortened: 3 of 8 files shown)


smnandre:
Hi @tijnema ! Do you thing you could show me a live example of the problem, or a small reproducer? 

I've never had problems with csrf tokens but i often use components without so 😅  ..

tijnema:
We see this in various ways, most common however is that CSRF tokens will reset on login (like issue #1150). After login we rerender live components to refresh the tokens. With this PR applied, the actions work normally. 

To reproduce, just use a simple live component with action, set session life time short, wait for session to expire and press action. It will fail with "CSRF token invalid". You could use the following stimulus controller to reload all components every 3 seconds: 

```lang=js
import {Controller} from '@hotwired/stimulus';
import {getComponent} from '@symfony/ux-live-component';

    connect() {
        setInterval(this.reloadLiveComponents, 3000);
    }

    async reloadLiveComponents() {
        const componentElements = document.querySelectorAll('[data-controller~="live"]');
        for (const componentElement of componentElements) {
            await getComponent(componentElement).then(component => {
                component.render();
            });
        }
    }
}
```

This makes sure the component gets new CSRF token. 
With this PR, action still works after session expired

smnandre:
> We see this in various ways, most common however is that CSRF tokens will reset on login (like issue https://github.com/symfony/ux/issues/1150). After login we rerender live components to refresh the tokens. With this PR applied, the actions work normally.

If you have two form on a page "Foo" currently, with csrf tokens. You open another page and log in. You come back on the page "Foo" and submit one of the form. What happens ?

>  just use a simple live component with action, set session life time short, wait for session to expire and press action

Is this not the same for a classic Form... or am i missing something here ?

>  With this PR, action still works after session expired

Do we want this ?

matijn-made:
> > We see this in various ways, most common however is that CSRF tokens will reset on login (like issue #1150). After login we rerender live components to refresh the tokens. With this PR applied, the actions work normally.
> 
> If you have two form on a page "Foo" currently, with csrf tokens. You open another page and log in. You come back on the page "Foo" and submit one of the form. What happens ?

It depends. If there's a mechanism which reloads the live component (Like the stimulus controller I posted above), it works. Otherwise it will fail with an invalid CSRF. 

> 
> > just use a simple live component with action, set session life time short, wait for session to expire and press action
> 
> Is this not the same for a classic Form... or am i missing something here ?

Yes and no. With a classic form, CSRF is usually stored in a hidden input `_csrf_token`. Updating CSRF after login with JS is as simple as updating this input field, or rerender the whole form server side and it should work. 

> 
> > With this PR, action still works after session expired
> 
> Do we want this ?

Yes. It's still safe because the CSRF will be bound to the users' session. It will be the new session which has been started, instead of it's previous expired one.

tijnema:
> > We see this in various ways, most common however is that CSRF tokens will reset on login (like issue #1150). After login we rerender live components to refresh the tokens. With this PR applied, the actions work normally.
> 
> If you have two form on a page "Foo" currently, with csrf tokens. You open another page and log in. You come back on the page "Foo" and submit one of the form. What happens ?

It depends. If there's a mechanism which reloads the live component (Like the stimulus controller I posted above), it works. Otherwise it will fail with an invalid CSRF. 

> 
> > just use a simple live component with action, set session life time short, wait for session to expire and press action
> 
> Is this not the same for a classic Form... or am i missing something here ?

Yes and no. With a classic form, CSRF is usually stored in a hidden input `_csrf_token`. Updating CSRF after login with JS is as simple as updating this input field, or rerender the whole form server side and it should work. 

> 
> > With this PR, action still works after session expired
> 
> Do we want this ?

Yes. It's still safe because the CSRF will be bound to the users' session. It will be the new session which has been started, instead of it's previous expired one.

smnandre:
> It depends. If there's a mechanism which reloads [..]. Otherwise it will fail with an invalid CSRF.

So currently there is no difference with the behaviour of standard Form, right ?

> Updating CSRF after login with JS is as simple as updating this input field, or rerender the whole form server side and it should work.

Is there something here you cannot do with LiveComponent ? You can re-render the whole component too right ? 

> Yes. It's still safe because the CSRF will be bound to the users' session. It will be the new session which has been started, instead of it's previous expired one.

But the form contains checksum and data from another user.


I'm not convinced for now, so let's wait other opinions on this.

smnandre:
Hello @tijnema! 

So it seems the team agrees with you on the concept "If a component send another CSRF token, it should be updated". So waiting was not a bad idea :) 

As you see it, is your PR ready or does it need some changes/updates ?

smnandre:
Friendly @tijnema ? Is this PR ready to merge for you ?  :)

tijnema:
Sorry, I was enjoying holiday. Good to hear the team agrees with me :). As far as I'm concerned, this PR is ready to merge.

smnandre:
I'm just waiting for someone else to check/validate and let's merge it ! 

 .... because we may need soon to adapt things here 😅  cc PR 58095  on symfony/symfony

Thank you @tijnema !

smnandre (approved):
(Good for me, i'd like someone else to check)

devath0 (approved):
Works to fix the bug in active components!

Kocal (approved):

smnandre:
It took time, but here we go, this is in now. Thank you very much @tijnema.

PR #2023, opened 2024-08-01
[Icons] Add `ignore_not_found` config option

### LINKED ISSUE #2008: ux_icon triggers an exception when icon doesn't exists
### Description

The twig `ux_icon` function always triggers an exception when the icon name doesn't exists.

In my opinion, since icons are usually not a critical feature (most of the time only for design purpose), this should trigger an exception only in `dev` mode by default and the exception should be catched with an error message in logs in `prod` mode.

Maybe by adding an option to the UX Icons component to configure this?

### Example

I have a feature in my app where the admin can add icons to display in the app.
One icon has been added with a typo and this put few pages of our site down.

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #2008
| License       | MIT

Allow to silence error during rendering when an icon is not found (then a warning is log)

### DIFF
--- a/src/Icons/src/DependencyInjection/UXIconsExtension.php
+++ b/src/Icons/src/DependencyInjection/UXIconsExtension.php
@@ -58,6 +58,10 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
                         ->end()
                     ->end()
                 ->end()
+                ->booleanNode('ignore_not_found')
+                    ->info('Ignore error when an icon is not found.')
+                    ->defaultFalse()
+                ->end()
             ->end()
         ;
 
@@ -69,7 +73,7 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
         return $this;
     }
 
-    protected function loadInternal(array $mergedConfig, ContainerBuilder $container): void // @phpstan-ignore-line
+    protected function loadInternal(array $mergedConfig, ContainerBuilder $container): void
     {
         $loader = new PhpFileLoader($container, new FileLocator(__DIR__.'/../../config'));
         $loader->load('services.php');
@@ -94,6 +98,10 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
 
         $container->getDefinition('.ux_icons.icon_renderer')
             ->setArgument(1, $mergedConfig['default_icon_attributes'])
+        ;
+
+        $container->getDefinition('.ux_icons.twig_icon_runtime')
+            ->setArgument(1, $mergedConfig['ignore_not_found'])
         ;
 
         if ($mergedConfig['iconify']['enabled']) {
--- a/src/Icons/src/Twig/UXIconRuntime.php
+++ b/src/Icons/src/Twig/UXIconRuntime.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Icons\Twig;
+
+use Psr\Log\LoggerInterface;
+use Symfony\UX\Icons\Exception\IconNotFoundException;
+use Symfony\UX\Icons\IconRendererInterface;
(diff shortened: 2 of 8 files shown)


src/Icons/src/Twig/UXIconRuntime.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @param array<string, bool|string> $attributes
+     */
+    public function renderIcon(string $name, array $attributes = []): string
+    {
+        try {
+            return $this->iconRenderer->renderIcon($name, $attributes);
+        } catch (IconNotFoundException $e) {
+            if ($this->ignoreNotFound) {
+                return '';
```
Seb33300, line 43: Can you log the catched error?

Seb33300 (changes requested):

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -58,6 +58,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->end()
                     ->end()
                 ->end()
+                ->booleanNode('ignore_not_found')
+                    ->info('Ignore error when an icon is not found.')
+                    ->defaultFalse()
```
yceruto, line 63: It could default to `%kernel.debug%` here or in the recipe

src/Icons/src/Twig/UXIconRuntime.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @param array<string, bool|string> $attributes
+     */
+    public function renderIcon(string $name, array $attributes = []): string
+    {
+        try {
+            return $this->iconRenderer->renderIcon($name, $attributes);
+        } catch (IconNotFoundException $e) {
+            if ($this->ignoreNotFound) {
```
yceruto, line 41: Because it's a Twig rendering context, I'm wondering if we should use `Environment::isStrictVariables()` instead. I can't picture a case where the `twig.strict_variables` config is different from `ux_icons.ignoreNotFound`. They serve the same purpose for me.

kbond: I'm thinking this also, at least until we're asked for a separate config option.

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: I think the idea in #2008 was to default to "not kernel.debug". 

I this change anything in prod, in would be a BC break, so maybe the recipe is a good solution, indeed

src/Icons/src/Twig/UXIconRuntime.php
smnandre: If by "using Environment::isStrictVariables()" you mean "do _not_ throw if strictVariable  = false", this would be a BC break, no ?

yceruto: Not throwing the exception in the prod env doesn't seem like a critical behavior change to me. In fact, that's exactly what the reporter is requesting (aligned with how Twig behaves). As suggested https://github.com/symfony/ux/pull/2023#discussion_r1701592117 we could log the error instead.

The configs (`strict_variables` and `ignore_not_found`) appear to be opposite due to their naming, but they actually represent the same concept/purpose.

smnandre:
Update: add a logger / log a warning when icon is not found

smnandre:
@kbond @yceruto 

--

I've been -more than once- told we cannot do this or that for way, way smaller BC breaks. 

But if you guys say it's ok, that's ok for me. 

--

`ignore_missing` exist in Twig (include, source) too. It was my inspiration here for the name ... and felt very similar 

Every other filter / function i know of throw in Twig when an error occurs

The only silent thing is for getAttribute.

To me an icon (which should be in cache) is nothing like an attribute on an (existing) object.

The asset function throw when an asset does not exist (example: a svg icon)

--

If we use strict_variables here, I _cannot_ have exception in prod for missing icons anymore. 

(i would need to set strict_variables to true -- something i don't desire for obvious reasons)

--

Not in the mood for the subtext of this discussion.. so just tell me your decision and i'll update the PR.

kbond:
This looks fine to me as is!

PR #804, opened 2023-04-18
[Site] More Functional Code, 2x Complex Live Components Examples, few fixes

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (with respect to Twig errors)
| New feature?  | no
| Tickets       | None
| License       | MIT

Hi!

Prepping ux.symfony.com for 2.8 - very excited about this! Join me on a tour:

## 1) Invoice Creator Live Component Demo

https://user-images.githubusercontent.com/121003/232886502-7f8da8f5-fe8a-4251-8e92-c2b56895b1ee.mp4

Parent component + collection of child components communicating by emitting events!

## 2) Product form + modal to create a Category

https://user-images.githubusercontent.com/121003/232888282-ab12c4a1-8153-4fe5-babf-cbf4a313c47b.mp4

Opens a Bootstrap modal with a child component inside - and closes that modal from a `LiveAction`.

### 3) Much nicer code blocks

Before:

Screenshot 2023-04-18 at 3 45 34 PM

After:

Screenshot 2023-04-18 at 3 45 41 PM

* "Copy" button
* Link to view the file on GitHub
* use statements are collapsed (click to show them)
* "Expand code" to show longer code blocks

### 4) Explanations on longer examples

Some examples deserve pointing out a few of the more complex parts. Some of the live component demos now do this:

Screenshot 2023-04-18 at 3 48 42 PM

### 5) Using a LOT of new features

[Smart rendering](https://symfony.com/bundles/ux-live-component/current/index.html#the-smart-re-render-algorithm), [emitting events](https://symfony.com/bundles/ux-live-component/current/index.html#communication-between-components-emitting-events), [browser events](https://symfony.com/bundles/ux-live-component/current/index.html#dispatching-browser-javascript-events), etc, etc - MANY new features are being used on the site now. 

We're also using (in som
e places) the new `<twig:Component>` syntax from #771. Even though editor support (e.g. for auto-completion) doesn't exist yet, I LOVE this. For example, to render a code block:

```html
<twig:CodeBlock filename="src/Twig/SearchPackages.php" />

{# or more complex - logic for extracting" a specific Twig block lives in the CodeBlock.php component #}
<twig:CodeBlock
    filename="templates/main/homepage.html.twig"
    targetTwigBlock="markdown_example"
    :showTwigExtends="false"
    :stripExcessHtml="true"
/>
```

Thanks everyone for the work, review & testing on a lot of this new stuff. Using the new features on the site has been BLAST.

PR #2030, opened 2024-08-07
Fix: Original Throwable get lost during component render

cuchac:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | 
| License       | MIT

When non-Exception Throwable is thrown during component rendering, original Throwable is lost together with it's stacktrace. Without stacktrace it is hard to debug the problem. This simple change fixes the problem in the same way as in https://github.com/cuchac/ux/blob/f69b9656f318464253590a6575da65f34262d931/src/TwigComponent/src/Twig/ComponentExtension.php#L127  where `previous: $e` is used instead of  `previous: $e->getPrevious()`

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentExtension.php
+++ b/src/TwigComponent/src/Twig/ComponentExtension.php
@@ -120,10 +120,6 @@ final class ComponentExtension extends AbstractExtension implements ServiceSubsc
             throw $e;
         }
 
-        if (!($e instanceof \Exception)) {
-            $e = new \Exception($e->getMessage(), $e->getCode(), $e->getPrevious());
-        }
-
-        throw new RuntimeError(\sprintf('Error rendering "%s" component: %s', $name, $e->getMessage()), previous: $e);
+        throw new RuntimeError(\sprintf('Error rendering "%s" component: "%s"', $name, $e->getMessage()), previous: $e);
     }
 }


smnandre:
Hi @cuchac 

I remember i almost made the same change .... and reverted because other scenarios made it not better... but don't remember precisely why/when.

So... let's solve this :) 

Do you have a simple scenario / use case that we can reproduce to play around ?
I'll check on my side to find why i went back :)

cuchac:
Hello @smnandre 

when I for example access undefined variable inside component PHP file, I will get following error page:
image

There are two stack traces but the original stack trace that can be used to solve the issue is lost. Without stack trace I can only guess where the problem inside the component is.

smnandre:
Just wondering, what if you remove entirely this block (i'm not remembering why it's there is the first place)

```
if (!($e instanceof \Exception)) {
    // ... 
}
```

smnandre:
Found the context / reasons here: 

* https://github.com/symfony/ux/pull/1132  
* https://github.com/symfony/ux/pull/804#pullrequestreview-1390839702

src/TwigComponent/src/Twig/ComponentExtension.php
```diff
@@ -121,9 +121,9 @@ private function throwRuntimeError(string $name, \Throwable $e): void
         }
 
         if (!($e instanceof \Exception)) {
-            $e = new \Exception($e->getMessage(), $e->getCode(), $e->getPrevious());
+            $e = new \Exception($e->getMessage(), $e->getCode(), $e);
         }
```
cuchac: I believe this whole part is useless. Why converting `Throwable` to `Exception`? RuntimeError can have `Throwable` as previous parameter. Wrapping Throwables in Twig RuntimeError is correct in my opinion, but wrapping Throwables in another Exception is useless, it just adds another stacktrace and also causes the problem with lost original stack trace because of `$e->getPrevious()` error.

cuchac:
> Just wondering, what if you remove entirely this block (i'm not remembering why it's there is the first place)
> 
> ```
> if (!($e instanceof \Exception)) {
>     // ... 
> }
> ```

If I delete that `if`, it will fix my issue. I believe it can be deleted.

cuchac:
Hello @smnandre, I've changed the PR to remove the problematic `if`.

src/TwigComponent/src/Twig/ComponentExtension.php
smnandre: I'm feeling the same, with no absolute certitude of the scenarios that probably justified this in the first place.

smnandre:
@kbond if you have time to check, i'm not sure if the initial reasons still apply.. but it feels like a good idea :)

cuchac:
@smnandre I've looked at the original commit and I don't see any reason for that `if` statement. I understand why Twig RuntimeException should be thrown, but I don't understand why "repack" non-exceptions. I believe it was mistake from the beginning.

kbond:
Thanks for this Joe. I'm not sure why this was done previously but this is the right move.

PR #2032, opened 2024-08-08
[Map] Fix custom attribute "data-controller" usage

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

I'm trying the new UX Map component on my website, to migrate from my homemade solution, and it looks like I forgot to **fully** implement the `data-controller` special behavior.

Calling `render_map(map, { 'data-controller': 'my-map' })` will now correctly be rendered.


javiereguiluz:
Thanks for fixing this bug Hugo.

PR #2035, opened 2024-08-08
[Map][Leaflet] Fix Popup's automatic-content

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Leaflet's Popup does not have `headerContent`, but GoogleMaps does. 

To prevent too many issues when developpers change for another Map Bridge, with Simon we took the decision to render `headerContent` and `content` data in Leaflet Popup's content.

But it looks like I've did things wrong, if I don't specify `headerContent`, then a `null` text is displayed in the popup:
Capture d’écran 2024-08-08 à 14 01 11

Because in JavaScript, `null + 'foo'` gives `nullfoo` :(
Capture d’écran 2024-08-08 à 14 04 29

With this PR, the popup is now nicely displayed:
Capture d’écran 2024-08-08 à 14 08 14

### DIFF
--- a/src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
+++ b/src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
@@ -35,7 +35,7 @@ class map_controller extends AbstractMapController {
     }
     doCreateInfoWindow({ definition, marker, }) {
         const { headerContent, content, rawOptions = {}, ...otherOptions } = definition;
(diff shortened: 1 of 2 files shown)


javiereguiluz:
Nice fix! I like the _(concise and modern)_ JS code that you used here. Thanks Hugo.

Kocal:
Thanks for quick merging Javier :)

PR #2036, opened 2024-08-08
[Map][Leaflet] Fix Marker inside-circle

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

With Leaflet, the original marker's icon (that you can find at https://www.jsdelivr.com/package/npm/leaflet?tab=files&path=dist%2Fimages), has a white circle in it, [see example](https://leafletjs.com/index.html):
Capture d’écran 2024-08-08 à 14 44 03

In #1937, in my first iteration I've added Leaflet icons to Symfony AssetMapper/ImportMap but this was toooo much work.
So, with Simon, we have agreed to use an inlined SVG instead, I've used the one from https://www.jsdelivr.com/package/npm/leaflet?tab=files&path=src%2Fimages, but it looks like the circle is transparent instead of white:
Capture d’écran 2024-08-08 à 14 39 05

With this PR:
Capture d’écran 2024-08-08 à 14 44 37


javiereguiluz:
Nice. Thanks Hugo.

If we receive feedback about this white circle from folks using maps in "dark mode", we can iterate this and think about possible solutions. But, I'd say that right now this is not a problem, so let's merge this. Thanks.

PR #2039, opened 2024-08-09
[Map] Change "BeforeConnect" for "BeforeCreate" in documentation

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Technically the code is not wrong, but when listening for events `ux:map:marker:before-create` or `ux:map:info-window:before-create`, you can expect the method to be named `*BeforeCreate` and not `*BeforeConnect`.


javiereguiluz:
Yes, it looks more consistent now. Thanks Hugo.

PR #2040, opened 2024-08-09
[Map] Add "extra" data for markers and infowindows

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

Actually when creating Marker and InfoWindow with Symfony UX Map, you can only pass:
- Marker: `position`, `title`, and `infoWindow`
- InfoWindow: `headerContent`, `content`, `position`, `opened` and `autoClose`

This is great, but too strict. 
Adding a new property `extra`, defined by the developper and **only used by the developper** (in specific Stimulus events), is IMHO a good way to open for more extensibility.

Note that the new example in the documentation is not complete yet, a 2nd PR will appears soon to customize the Marker Icon :)


javiereguiluz:
Thanks to your description I understand the need for this. Also, I like the `extra` option name because it's short and easy to understand. Thanks Hugo!

PR #2045, opened 2024-08-11
[Map] Add the possibility to not configure map zoom/center if fit bounds to markers

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

Following a review from @smnandre in #1937 or in our Slack DMs.

Forcing the developper to explicitly configure the center/zoom of the map does not make sense if `fitBoundsToMarker` feature is enabled.

Also, I prefer throw exceptions instead of automatically enable `fitBoundsToMarker` if there are markers and center/zoom are not configured, it's more explicit like that and it mimic the Leaflet/GoogleMaps behaviors (the map can't be rendered if no center/zoom, and `fitBoundsToMarker` is a UX Map's feature).

### DIFF
--- a/src/Map/tests/MapTest.php
+++ b/src/Map/tests/MapTest.php
@@ -24,7 +24,7 @@ class MapTest extends TestCase
     public function testCenterValidation(): void
     {
         self::expectException(InvalidArgumentException::class);
-        self::expectExceptionMessage('The center of the map must be set.');
+        self::expectExceptionMessage('The map "center" must be explicitly set when not enabling "fitBoundsToMarkers" feature.');
 
         $map = new Map();
         $map->toArray();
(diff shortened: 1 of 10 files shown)


src/Map/tests/MapTest.php
```diff
@@ -24,7 +24,7 @@ class MapTest extends TestCase
     public function testCenterValidation(): void
     {
         self::expectException(InvalidArgumentException::class);
-        self::expectExceptionMessage('The center of the map must be set.');
+        self::expectExceptionMessage('The map "center" must be explicitly set when not enabling "fitBoundsToMarkers" feature.');
```
smnandre, line 27: I still think we should make it true per default 🙊

Kocal: 😠

Kocal: More seriously, fit bounds to marker is not a feature from Google Maps or Leaflet, it's a feature from UX Map, and enable it automatically depending of certains conditions looks wrong to me, because it behaves differently than GMaps/Leaflet in this situation.

If we make it `true` by default, we **should** disable it (if not explicitly enabled) if the dev provide a center / zoom configuration.

But, that's true it can offers a better DX, because actually if you create a map (without configuring the zoom / center, fit bounds to marker), nothing is displayed (either grey map, or errors in console).
If we enable it by default (and disable it if necessary), the user will see its map without any necessary configuration

kbond:
Thanks Hugo.

PR #2058, opened 2024-08-13
[Map] Fix doc link

smnandre:
As `https://symfony.com/ux` now redirects to the "FrontEnd page" of Symfony docs --that _barely_ mention Symfony UX-- i think directing users to the website could be less missleading.


kbond:
Thanks Simon.

PR #2063, opened 2024-08-14
[CI] Test ux.symfony.com with local UX packages

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Following #1965 and #2060, I was a bit surprised to see ux.symfony.com tests were executed with UX packages coming from `2.x` branch, instead of UX packages from the same PR.

I believe doing this could help us for the future and prevent some regressions.

I've updated the script `.github/build-packages.php` to automatically replaces **all UX packages requires(-dev)** from all significant `composer.json` files from the repo:
image

### DIFF
--- a/.github/build-packages.php
+++ b/.github/build-packages.php
@@ -9,11 +9,13 @@ require __DIR__.'/../vendor/autoload.php';
 use Symfony\Component\Finder\Finder;
 
 $finder = (new Finder())
-    ->in([__DIR__.'/../src/*/', __DIR__.'/../src/*/src/Bridge/*/'])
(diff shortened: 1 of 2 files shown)


smnandre:
I think we need both (one to check all components with their "current dev version"... but also one like before, to simulate if aa user checks @dev with stable components), no ? 

(not for the ux website probably, but the build package is used for all tests right ? )

Kocal:
> I think we need both (one to check all components with their "current dev version"... but also one like before, to simulate if aa user checks @dev with stable components), no ? 

Nothing changed here, packages tests are still running like before (by using the "local" packages).

This same behaviour has been applied to ux.symfony.com project to take advantage of installing (and testing a bit) our UX packages in a real application.

kbond:
Thanks Hugo.

PR #2053, opened 2024-08-13
[Site] UX Map page

smnandre:
(very simple page for now, but i'd like to deploy _a_ page asap... and then we'll have more time for demos, docs...)

ux-map-screen

⚠️  Will need rebase / composer after 2.19.2 (and some check on the controller.json)

Closes #2042

PR #2069, opened 2024-08-15
[ux.symfony.com] Fix rendering/color issues on Map page

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

It looks like we have some visual issues in light mode on the Map page, we can't see the menu:
Capture d’écran 2024-08-15 à 07 42 21

And we miss the `composer require ...` thing.

I've Replaced the usage of the new component `Hero` (introduced in #2053) by the good old `PackageHeader` component.

This is the quickest solution right now to correctly render UX Map page, and it gives us some times to improve this `Hero` component and re-use it in all UX package pages.

**EDIT:** I've seen with @smnandre for some minor adjustements:
image
image


smnandre (approved):

ux.symfony.com/templates/ux_packages/map.html.twig
```diff
@@ -3,15 +3,18 @@
...
+        eyebrowText: 'Seamless Maps Integration',
+    } %}
+        {% block title_header %}
+            Embed <em>interactive maps</em> in a breeze!
+        {% endblock %}
+
+        {% block sub_content %}
+            Decouple your code from your map provider: <em>Google Maps</em>, <em>LeaftLet</em>.
+        {% endblock %}
+    {% endcomponent %}
```
WebMamba, line 17: Why not use the HTML syntax?

ux.symfony.com/templates/ux_packages/map.html.twig
```diff
@@ -3,15 +3,18 @@
...
+    {% component PackageHeader with {
+        package: 'map',
+        eyebrowText: 'Seamless Maps Integration',
+    } %}
+        {% block title_header %}
+            Embed <em>interactive maps</em> in a breeze!
+        {% endblock %}
+
+        {% block sub_content %}
+            Decouple your code from your map provider: <em>Google Maps</em>, <em>LeaftLet</em>.
```
WebMamba, line 15: I don't really understand this syntax what do you think about something like: Seamlessly connect with any map provider, like Google Maps or Leaflet.

WebMamba (commented):
Tested it locally and it work great 👍

ux.symfony.com/templates/ux_packages/map.html.twig
smnandre: It's not used elsewhere for now. I used it in the first version of this page with the new component but here it feels more coherent.

But all this will really soon change so i don't mind :)

ux.symfony.com/templates/ux_packages/map.html.twig
smnandre: _Seamless_ is used in the eyebrow, and _decouple_ felt a good thing to expose as it is an important promise of the component, but i agree this sentence could be more catchy / efficient

ux.symfony.com/templates/ux_packages/map.html.twig
Kocal: I've just go for the fastest (all other packages use this syntax), hoping this PR could quickly be merged and deployed :)

kbond:
Thanks Hugo.

PR #2072, opened 2024-08-15
[Cookbook] English / doc / code corrections

smnandre:
One commit per "category"

@WebMamba I don't get the Stimulus button example... how would it work ?

### DIFF
--- a/ux.symfony.com/cookbook/component_architecture.md
+++ b/ux.symfony.com/cookbook/component_architecture.md
@@ -1,6 +1,6 @@
 ---
-title: Component architecture
-description: Rules and pattern to work with components
+title: Component Architecture
+description: Rules and patterns for working with components
 image: images/cookbook/component-architecture.png
 tags: 
     - JavaScript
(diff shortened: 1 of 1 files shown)


ux.symfony.com/cookbook/component_architecture.md
```diff
@@ -11,25 +11,25 @@ published_at: '2024-08-02'
 
 ## Introduction
 
-In Symfony UX exist two packages: [TwigComponents](https://symfony.com/bundles/ux-twig-component/current/index.html) and [LiveComponent](https://symfony.com/bundles/ux-live-component/current/index.html).
-Those two packages allow you to create reusable components in your Symfony application.
-But the component architecture is not exclusive to Symfony, it's a design pattern that can be applied to any programming language or framework.
-And the Javascript world already implements this architecture for long time, on many different frameworks like React, Vue, or Svelte.
-So, a set of rules and patterns has already be defined to work with components. This is why Symfony UX tries to be as close as possible to those rules.
-So, let's see what those rules are!
+In Symfony UX, there are two packages: [TwigComponents](https://symfony.com/bundles/ux-twig-component/current/index.html) and [LiveComponent](https://symfony.com/bundles/ux-live-component/current/index.html).
```
Kocal, line 14: General question, should we backtick TwigComponents and LiveComponent?

smnandre: Also should use plural on both or none.

WebMamba (approved):

WebMamba:
> @WebMamba I don't get the Stimulus button example... how would it work 

Sorry, I don't get what you mean. Do you think the example is not relevant?

smnandre:
> > @WebMamba I don't get the Stimulus button example... how would it work
> 
> Sorry, I don't get what you mean. Do you think the example is not relevant?

No sorry, i have (as too often) miss-expressed myself :) To be working, it would need some more code.. i know the goal is to be synthetic and didactic, but i feel it a bit missleading. We speak about state transition when clicked, but we don't show the code to do this. 

What do you recon here ? (and i'll follow your vision ;)) : 
* simplified code to give an idea of the component behaviour (but is not _copy-pastable_ as it)
* exact code (that can be a bit too much to illustrate the article)

WebMamba:
I think adding a state is not much if you thing that it could be relevant. So lets add this state

smnandre:
Let's merge as it is if you're okay, and we'll see the example if someone ask questions :))

kbond (approved):

Kocal (approved):

PR #2073, opened 2024-08-16
[Map] Rename `render_map` Twig function `ux_map`

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #... 
| License       | MIT

### Context

As we recently saw with the `cva` function and forgot to do here: we should prefix every new component / package / Twig function or filter with `ux` to avoid conflict with existing codebases... and with future symfony or twig new methods.

### This PR

This PR renames `render_map` (inspired by charts hjs "render_chart") into `ux_map` (in line with what we did for UX Icons)

```diff
<div>
-    {{ render_map(map) }}
+    {{ ux_map(map) }}
</div>
```

### Next

We can keep the render_map for _some_ time, but it's marked as deprecated right now, and i think this is the good moment to update documentation, to limit future BC break / changes for users (even if we are in experimental)

### Website

Once this once is merged, the website page will need to be updated

### DIFF
--- a/src/Map/CHANGELOG.md
+++ b/src/Map/CHANGELOG.md
@@ -2,4 +2,9 @@
 
 ## Unreleased
 
+-   Rename `render_map` Twig function `ux_map`
+-   Deprecate `render_map` Twig function
+
+## 2.19
+
 -   Component added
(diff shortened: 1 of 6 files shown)


Kocal (approved):
Thanks!


TIL about `deprecated`/`deprecating_package` :)

kbond (approved):

kbond:
Thanks Simon.

PR #2075, opened 2024-08-17
[root] Upgrade JS build dependencies

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Some maintenance on our root JS dependencies, upgrading Rollup, TypeScript, Babel, ...

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -94,9 +94,16 @@ module.exports = {
         resolve(),
         typescript({
             filterRoot: packageRoot,
-            include: ['src/**/*.ts'],
+            include: [
+                'src/**/*.ts', 
+                // TODO: Remove for the next major release
+                // "@rollup/plugin-typescript" v11.0.0 fixed an issue (https://github.com/rollup/plugins/pull/1310) that 
+                // cause a breaking change for UX React users, the dist file requires "react-dom/client" instead of "react-dom"
+                // and it will break for users using the Symfony AssetMapper without Symfony Flex (for automatic "importmap.php" upgrade).
+                '**/node_modules/react-dom/client.js'
+            ],
             compilerOptions: {
-                outDir: 'dist',
+                outDir: '.',
                 declaration: true,
                 emitDeclarationOnly: true,
             }
--- a/package.json
+++ b/package.json
@@ -10,19 +10,19 @@
         "check-format": "biome format"
     },
     "devDependencies": {
-        "@babel/core": "^7.15.8",
-        "@babel/preset-env": "^7.15.8",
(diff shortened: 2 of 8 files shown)


rollup.config.js
```diff
@@ -96,7 +96,7 @@ module.exports = {
             filterRoot: packageRoot,
             include: ['src/**/*.ts'],
             compilerOptions: {
-                outDir: 'dist',
+                outDir: '.',
```
Kocal, line 106: Because of https://github.com/rollup/plugins/pull/1378

src/React/assets/dist/render_controller.js
```diff
@@ -1,25 +1,7 @@
 import React from 'react';
-import require$$0 from 'react-dom';
+import { createRoot } from 'react-dom/client';
```
Kocal: I updated `ux-react`'s importmap to require `react-dom/client` instead.

smnandre: Better add an entry in the React CHANGELOG then, to be safe ?

smnandre (commented):
I trust you on this topic, you're way more qualified than me here 😅

smnandre (approved):

src/React/assets/dist/render_controller.js
Kocal: You're right!

Kocal: Hum, in fact, thinking twice, this can be considered as a breaking change for people not using Symfony Flex (because the importmap won't be automatically updated).

I really prefer:
```js
import { createRoot } from 'react-dom/client';
```
over:
```js
import require$$0 from 'react-dom';

var createRoot;

var m = require$$0;
if (process.env.NODE_ENV === 'production') {
  createRoot = m.createRoot;
  m.hydrateRoot;
} else {
  var i = m.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
  createRoot = function(c, o) {
    i.usingClientEntryPoint = true;
    try {
      return m.createRoot(c, o);
    } finally {
      i.usingClientEntryPoint = false;
    }
  };
}
```

But the React package is not experimental, and I don't think we plan to release a v3.0 soon.

I will try to find a way to reuse the previous code

Kocal: Fixed by adding `'**/node_modules/react-dom/client.js'` to Rollup's plugin for TypeScript

WebMamba (approved):

kbond:
Thanks Hugo.

PR #2083, opened 2024-08-18
[Map] Improve web performances by not calling "createInfoWindow" if unnecessary

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

This afternoon we spoke a bit with @simondaigre, and he told me about InfoWindow performance issues when you have too many of them (and depending your device).

If we have 1.000 InfoWindow, hidden by default and only visible after clicking on a marker, why should be **only** create the InfoWindow at that time?

This PR introduce lazy-creation of InfoWindow, it should improve web-performances on the user's browsers.

I've used the following PHP code:
```php

       $map = (new Map())
            ->center(new Point(0, 0))
            ->zoom(2)
            ->fitBoundsToMarkers()
            ->options(new GoogleOptions(
                mapId: 'XXX',
            ))
        ;

        for ($i = 0; $i < 1_000; $i++) {
            $map->addMarker(
                new Marker(
                    position: new Point(random_int(-90, 90), random_int(-180, 180)),
                    title: 'Marker ' . $i,
                    infoWindow: new InfoWindow(
                        content: 'Info window ' . $i,
                        opened: $i % 499 === 0
                    )
                )
            );
        }
```

## Benchmarks 

I've did some benchmarks for both Google and Leaflet:
- with 1.000 markers, 
- on a Macbook M1 Pro, CPU Throttle x6

### Google

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>Capture d’écran 2024-08-18 à 18 28 54
 <td>Capture d’écran 2024-08-18 à 18 30 29
</table> 

`connect()` takes 936ms instead of ~1,80s, we won ~900ms.

### Leaflet

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>Capture d’écran 2024-08-18 à 18 45 10
 <td>Capture d’écran 2024-08-18 à 18 45 28
</table> 

`connect()` takes ~1s instead of ~1,20s, we won ~200ms

---

The next steps for performance optimizations are:
- Lazyloading markers and infoWindows, in order to keep a good TTFB 
- Clusters!

### DIFF
--- a/src/Map/assets/src/abstract_map_controller.ts
+++ b/src/Map/assets/src/abstract_map_controller.ts
@@ -111,12 +111,14 @@ export default abstract class<
     protected createInfoWindow({
         definition,
         marker,
+        onMarkerClick = false,
     }: {
         definition: MarkerDefinition<MarkerOptions, InfoWindowOptions>['infoWindow'];
         marker: Marker;
+        onMarkerClick?: boolean;
     }): InfoWindow {
         this.dispatchEvent('info-window:before-create', { definition, marker });
-        const infoWindow = this.doCreateInfoWindow({ definition, marker });
+        const infoWindow = this.doCreateInfoWindow({ definition, marker, onMarkerClick });
         this.dispatchEvent('info-window:after-create', { infoWindow, marker });
 
         this.infoWindows.push(infoWindow);
@@ -127,9 +129,11 @@ export default abstract class<
     protected abstract doCreateInfoWindow({
         definition,
         marker,
+        onMarkerClick,
     }: {
         definition: MarkerDefinition<MarkerOptions, InfoWindowOptions>['infoWindow'];
         marker: Marker;
+        onMarkerClick: boolean;
     }): InfoWindow;
 
     protected abstract doFitBoundsToMarkers(): void;
--- a/src/Map/src/Bridge/Google/assets/src/map_controller.ts
+++ b/src/Map/src/Bridge/Google/assets/src/map_controller.ts
@@ -121,7 +121,13 @@ export default class extends AbstractMapController<
         });
 
         if (infoWindow) {
-            this.createInfoWindow({ definition: infoWindow, marker });
+            if (infoWindow.opened) {
+                this.createInfoWindow({ definition: infoWindow, marker });
+            } else {
+                marker.addListener('click', () => {
+                    this.createInfoWindow({ definition: infoWindow, marker, onMarkerClick: true });
+                });
+            }
         }
 
         return marker;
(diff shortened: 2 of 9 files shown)


src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
```diff
@@ -38,14 +38,21 @@ class map_controller extends AbstractMapController {
...
         const marker = L.marker(position, { title, ...otherOptions, ...rawOptions }).addTo(this.map);
         if (infoWindow) {
-            this.createInfoWindow({ definition: infoWindow, marker });
+            if (infoWindow.opened) {
+                this.createInfoWindow({ definition: infoWindow, marker });
+            }
+            else {
+                marker.on('click', () => {
+                    this.createInfoWindow({ definition: infoWindow, marker, onMarkerClick: true });
+                });
```
smnandre, line 47: Does not seem a good idea to run every time.

src/Map/assets/src/abstract_map_controller.ts
```diff
@@ -111,12 +111,14 @@ export default abstract class<
     protected createInfoWindow({
         definition,
         marker,
+        onMarkerClick = false,
```
smnandre, line 114: Not sure this should be exposed like this, seems to complexify the contract, while not beeing this "explanatory"

src/Map/src/Bridge/Google/assets/src/map_controller.ts
```diff
@@ -146,24 +154,19 @@ export default class extends AbstractMapController<
...
         });
 
-        if (definition.opened) {
+        if (definition.opened || onMarkerClick) {
             infoWindow.open({
                 map: this.map,
                 shouldFocus: false,
                 anchor: marker,
             });
         }
```
smnandre, line 163: Confusing to me... this mixes runtime values and definition values

smnandre:
I feel like the current implementation could be improved, but i don't think we could/should avoid exposing this internal code.. and the loop seems a bit odd.

Why not having separate create / open methods ?

src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
Kocal: What do you mean?

smnandre: "createInfoWindow" will be called on every "click" ?

Kocal: Ah, yes you're right.

Hum, I may have better ideas to rewrite the PR, but I don't have time right now, let's put it in draft.

Kocal:
Let's re-open it later

PR #2085, opened 2024-08-18
[Typed] Update index.rst

tilimac:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| License       | MIT

I tried to use the onStop event without success but according to the documentation, you have to use onComplete to mark the end of typing


smnandre:
According to the documentation, these two methods exist (and are different)

https://mattboldt.github.io/typed.js/docs/

PR #2091, opened 2024-08-21
[Site] Support page

smnandre:
Minimal / simple page to list support options

Capture d’écran 2024-08-22 à 06 23 30

You had @WebMamba something like this in mind ?

### DIFF
--- a/ux.symfony.com/assets/styles/app.scss
+++ b/ux.symfony.com/assets/styles/app.scss
@@ -85,6 +85,7 @@
 @import "components/PackageHeader";
 @import "components/PackageBox";
 @import "components/Cookbook";
+@import "components/SupportBox";
 @import "components/Tabs";
 @import "components/Tag";
 @import "components/Terminal";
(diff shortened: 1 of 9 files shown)


kbond:
I like this!

Kocal:
What about the logo to the left of the title?

smnandre:
> What about the logo to the left of the title?

Two reasons i kept it top of the title
* i wanted to change a bit from the "PackageBox", have some variety in card layouts
* i have something in mind for the top right space but i'd like to see if the page is used, first :) 

But no problem if you prefer, i don't really mind here :))

kbond:
Make sense. Let's roll with this!

Kocal:
> > What about the logo to the left of the title?
> 
> Two reasons i kept it top of the title
> 
> * i wanted to change a bit from the "PackageBox", have some variety in card layouts
> * i have something in mind for the top right space but i'd like to see if the page is used, first :)
> 
> But no problem if you prefer, i don't really mind here :))

Okay, let's keep it that way then :)

Kocal (approved):

smnandre:
Capture d’écran 2024-08-22 à 23 17 03

Kocal (approved):
Works fine in light/dark modes, and in mobile too:
Capture d’écran 2024-08-23 à 00 34 12
image
image

Kocal:
Thank you @smnandre.

Issue #1649, opened 2024-03-23
[Autocomplete] incompatible with FrankenPHP's worker mode


pyatnitsev:
![image](https://github.com/symfony/ux/assets/4361764/b5299614-91fd-439f-850b-2ad93808c852)

I use https://ux.symfony.com/autocomplete with standard doctrine.
first request when cursor pointed to field was ok, but no one more. All next requests ends with 500 error with log, that I have attached above.

This issue related to worker mode. I disable it, now autocomplete works file.

So, I can create a minimal app, that reproduce issue.

related issue for FrankenPHP: https://github.com/dunglas/frankenphp/issues/683

smnandre:
Is frankenPhp in worker mode fully supported by Symfony/Doctrine/Twig currently ? 
(**genuine question**)

pyatnitsev:
Yes it is. 
I run it in production. Worker mode support symfony and this components (doctrine and twig), but I have issue with symfony UX's autocomplete component.

smnandre:
> So, I can create a minimal app, that reproduce issue.

Yep, that would make it easier to understand what's happening there, thank you. But as you can see, my knowledge of worker modes is quite... limited 😅

bpacholek:
Hey @pyatnitsev not sure if you still experience the issue or you have managed to overcome it, but at least until symfony/ux is officially supported with worker modes (of any runner, not only frankenphp) you can use a decorator:

1. Create a file in your project. For instance: `src/Form/Autocomplete/WorkerAwareWrappedEntityTypeAutocompleter.php`.
2. Use this code:

```php
<?php

namespace App\Form\Autocompleter;

use Doctrine\ORM\EntityRepository;
use Doctrine\ORM\QueryBuilder;
use Symfony\Bundle\SecurityBundle\Security;
use Symfony\Component\Form\FormFactoryInterface;
use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
use Symfony\UX\Autocomplete\Doctrine\EntityMetadataFactory;
use Symfony\UX\Autocomplete\Doctrine\EntitySearchUtil;
use Symfony\UX\Autocomplete\Form\WrappedEntityTypeAutocompleter;
use Symfony\UX\Autocomplete\OptionsAwareEntityAutocompleterInterface;

final class WorkerAwareWrappedEntityTypeAutocompleter implements OptionsAwareEntityAutocompleterInterface
{
    private WrappedEntityTypeAutocompleter $innerAutocompleter;

[...]

PR #2094, opened 2024-08-23
[Autocomplete] Reset form state on each request for applications that reuse the Symfony application between requests

ashceo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | 
| License       | MIT

Hi there, in ux-autocomplete there is an WrappedEntityTypeAutocompleter::getForm which on first call saves form instance in WrappedEntityTypeAutocompleter $form property, it has method setOptions which is called on each request in EntityAutocompleteController, in PHP-FPM everything will work as expected where container is created on each request, but with Swoole or similar where container is not created on each request rather only first request on first EntityAutocompleteController request everything is also fine, but on second request WrappedEntityTypeAutocompleter  will have an instance of form from previous request which is fine but the call of setOptions in controller will throw exception.

So in Swoole on second and next requests $this->form already contains an instance of form from previous request because DI Container is not created on each request as in PHP-FPM (Container is created only once) and exception is thrown. So, Symfony have an kernel.reset / kernel.terminate event which probably can be used here to reset $this->form in WrappedEntityTypeAutocompleter because they are fired after request is completed.

### DIFF
--- a/src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
+++ b/src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
@@ -19,6 +19,7 @@ use Symfony\Component\Form\FormFactoryInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\PropertyAccess\PropertyPathInterface;
+use Symfony\Contracts\Service\ResetInterface;
 use Symfony\UX\Autocomplete\Doctrine\EntityMetadata;
 use Symfony\UX\Autocomplete\Doctrine\EntityMetadataFactory;
 use Symfony\UX\Autocomplete\Doctrine\EntitySearchUtil;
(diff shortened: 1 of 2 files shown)


src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
```diff
@@ -188,4 +189,10 @@ public function setOptions(array $options): void
 
         $this->options = $options;
     }
+
+    public function reset(): void
```
smnandre, line 193:
```suggestion
    /**
     * @internal
     */
    public function reset(): void
```

smnandre: Well nevermind, the entire class is internal 😶

smnandre:
Thank you @dotdevio for this PR! 

Could you maybe add a test? 

Also.. I'd see this more as a feature than a bug, as swoole (or similar runners) compatibility was not something supported before..  what do you think ?

ashceo:
> Thank you @dotdevio for this PR!
> 
> Could you maybe add a test?
> 
> Also.. I'd see this more as a feature than a bug, as swoole (or similar runners) compatibility was not something supported before.. what do you think ?

This is not a feature because Symfony supports multiple runtimes and this bug will be not only in Swoole but any other runtime like FrankenPHP, RoadRunner etc. Service state shoud reset in this case for sure, no BC breaks on PHP-FPM. I would like add a test but i dont see how i can test here and i dont found any tests for WrappedEntityTypeAutocompleter. Actually there is noothing to break the app, just unsetting variable on kernel.reset event.

Kocal (approved):

smnandre:
Ok! Thank you for the fix @dotdevio :)

smnandre (approved):

1ed:
This fixes https://github.com/symfony/ux/issues/1649

kbond:
Perfect, thanks @dotdevio!

PR #2107, opened 2024-08-29
[Doc] Fix a Twig function name in UX Map docs

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

I've started using the UX Map component in a Symfony app. So far it's great, thanks. I found a minor error in the docs about the name of the function to use.

### DIFF
--- a/src/Map/doc/index.rst
+++ b/src/Map/doc/index.rst
@@ -124,10 +124,11 @@ To render a map in your Twig template, use the ``ux_map`` Twig function, e.g.:
 
 .. code-block:: twig
 
-    {{ ux_map(my_map) }}
-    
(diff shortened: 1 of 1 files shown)


Kocal:
Hi @javiereguiluz, the function has been renamed to `ux_map` in #2073, but it looks like it has not be released yet. 
That's why you are seeing `ux_map` in the doc, but `render_map` in the code 😕

javiereguiluz:
Ah ok! Thanks for checking this. I've reverted the function name change and kept the other doc changes. Thanks.

Kocal (approved):

PR #2113, opened 2024-08-31
[Map][Leaflet] Rename `leaflet` to `L` in event's payloads

### LINKED ISSUE #2111: [UX Map] Rename a variable
When using UX Map and Leaflet bridge to customize a marker, you have to do this:

```js
const { definition, leaflet } = event.detail;
// ...
```

As suggested by @Kocal, the `leaflet` variable should probably be named `L` because Leaflet uses `L` in all its docs. Now that the bridge is so new (and experimental), we could even rename the variable wihout considering BC, etc.

Thanks

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2111 
| License       | MIT

### DIFF
--- a/src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
+++ b/src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
@@ -18,7 +18,7 @@ class map_controller extends AbstractMapController {
             prefix: 'ux:map',
             detail: {
                 ...payload,
-                leaflet: L,
+                L,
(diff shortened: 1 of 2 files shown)


javiereguiluz (approved):
Thanks!

Kocal:
Failing checks will be fixed by #2114.

kbond:
Thanks Hugo.

PR #2117, opened 2024-09-01
[Map] Render map from Twig with `ux_map()` and  `<twig:ux:map />` 

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #...
| License       | MIT

* ~~add an MapFactory (internal)~~
* allow `ux_map` Twig function to render map from named arguments
* add basic TwigComponent `<twig:ux:map ... />`

### Create Map from Twig

```twig

{{ ux_map(
    center: [51.5074, 0.1278],
    zoom: 3,
    markers:  [
        { position: [51.5074, 0.1278], title: 'London' },
        { position: [48.8566, 2.3522], title: 'Paris' },
        {
            position: [40.7128, -74.0060],
            title: 'New York',
            infoWindow: { content: 'Welcom to <b>New York</b>' }
        },
    ],
    attributes: {
        class: 'foo',
        style: 'height: 800px; width: 100%; border: 4px solid red; margin-block: 10vh;',
    }
) }}

```

ux-map

### DIFF
--- a/src/Map/CHANGELOG.md
+++ b/src/Map/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 ## 2.20
 
--   Rename `render_map` Twig function `ux_map`
--   Deprecate `render_map` Twig function
+-   Deprecate `render_map` Twig function (will be removed in 2.21). Use 
+    `ux_map` or the `<twig:ux:map />` Twig component instead.
+-   Add `ux_map` Twig function (replaces `render_map` with a more flexible 
+    interface)
+-   Add `<twig:ux:map />` Twig component
 
 ## 2.19
 
--- a/src/Map/config/twig_component.php
+++ b/src/Map/config/twig_component.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Symfony\UX\Map\Twig\UXMapComponent;
+use Symfony\UX\Map\Twig\UXMapComponentListener;
+use Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent;
+
+return static function (ContainerConfigurator $container): void {
+    $container->services()
+        ->set('.ux_map.twig_component_listener', UXMapComponentListener::class)
+            ->args([
+                service('ux_map.twig_runtime'),
+            ])
+            ->tag('kernel.event_listener', [
+                'event' => PreCreateForRenderEvent::class,
+                'method' => 'onPreCreateForRender',
+            ])
+
+        ->set('.ux_map.twig_component.map', UXMapComponent::class)
+            ->tag('twig.component', ['key' => 'UX:Map'])
+    ;
+};
--- a/src/Map/src/Marker.php
+++ b/src/Map/src/Marker.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\UX\Map;
 
+use Symfony\UX\Map\Exception\InvalidArgumentException;
+
 /**
  * Represents a marker on a map.
  *
@@ -19,7 +21,8 @@ namespace Symfony\UX\Map;
 final readonly class Marker
 {
     /**
-     * @param array<string, mixed> $extra Extra data, can be used by the developer to store additional information and use them later JavaScript side
+     * @param array<string, mixed> $extra Extra data, can be used by the developer to store additional information and
+     *                                    use them later JavaScript side
      */
     public function __construct(
         private Point $position,
@@ -29,6 +32,14 @@ final readonly class Marker
     ) {
     }
 
+    /**
+     * @return array{
+     *     position: array{lat: float, lng: float},
+     *     title: string|null,
+     *     infoWindow: array<string, mixed>|null,
+     *     extra: object,
+     * }
+     */
     public function toArray(): array
     {
         return [
@@ -37,5 +48,29 @@ final readonly class Marker
             'infoWindow' => $this->infoWindow?->toArray(),
             'extra' => (object) $this->extra,
         ];
+    }
+
+    /**
+     * @param array{
+     *     position: array{lat: float, lng: float},
+     *     title: string|null,
+     *     infoWindow: array<string, mixed>|null,
+     *     extra: object,
+     * } $marker
+     *
+     * @internal
+     */
+    public static function fromArray(array $marker): self
+    {
+        if (!isset($marker['position'])) {
+            throw new InvalidArgumentException('The "position" parameter is required.');
+        }
+        $marker['position'] = Point::fromArray($marker['position']);
+
+        if (isset($marker['infoWindow'])) {
+            $marker['infoWindow'] = InfoWindow::fromArray($marker['infoWindow']);
+        }
+
+        return new self(...$marker);
     }
 }
--- a/src/Map/src/Point.php
+++ b/src/Map/src/Point.php
@@ -43,4 +43,16 @@ final readonly class Point
             'lng' => $this->longitude,
         ];
     }
+
+    /**
+     * @param array{lat: float, lng: float}|array{0: float, 1: float} $point
+     */
+    public static function fromArray(array $point): self
+    {
+        if (isset($point['lat'], $point['lng'])) {
+            return new self($point['lat'], $point['lng']);
+        }
+
+        return new self(...$point);
+    }
 }
--- a/src/Map/src/Twig/MapRuntime.php
+++ b/src/Map/src/Twig/MapRuntime.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Map\Twig;
+
+use Symfony\UX\Map\Map;
+use Symfony\UX\Map\Marker;
+use Symfony\UX\Map\Point;
+use Symfony\UX\Map\Renderer\RendererInterface;
+use Twig\Extension\RuntimeExtensionInterface;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
+ */
+final class MapRuntime implements RuntimeExtensionInterface
+{
+    public function __construct(
+        private readonly RendererInterface $renderer,
+    ) {
+    }
+
+    /**
+     * @param array<string, mixed> $attributes
+     * @param array<string, mixed> $markers
+     */
+    public function renderMap(
+        ?Map $map = null,
+        array $attributes = [],
+        ?array $markers = null,
+        ?array $center = null,
+        ?float $zoom = null,
+    ): string {
+        if ($map instanceof Map) {
+            if (null !== $center || null !== $zoom || $markers) {
+                throw new \InvalidArgumentException('You cannot set "center", "markers" or "zoom" on an existing Map.');
+            }
+
+            return $this->renderer->renderMap($map, $attributes);
+        }
+
+        $map = new Map();
+        foreach ($markers ?? [] as $marker) {
+            $map->addMarker(Marker::fromArray($marker));
+        }
+        if (null !== $center) {
+            $map->center(Point::fromArray($center));
+        }
+        if (null !== $zoom) {
+            $map->zoom($zoom);
+        }
+
+        return $this->renderer->renderMap($map, $attributes);
+    }
+}
--- a/src/Map/src/Twig/UXMapComponent.php
+++ b/src/Map/src/Twig/UXMapComponent.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Map\Twig;
+
+use Symfony\UX\Map\Marker;
+use Symfony\UX\Map\Point;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
+ */
+final class UXMapComponent
+{
+    public ?float $zoom;
+
+    public ?Point $center;
+
+    /**
+     * @var Marker[]
+     */
+    public array $markers;
+}
--- a/src/Map/src/Twig/UXMapComponentListener.php
+++ b/src/Map/src/Twig/UXMapComponentListener.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Map\Twig;
+
+use Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
+ */
+final class UXMapComponentListener
+{
+    public function __construct(
+        private MapRuntime $mapRuntime,
+    ) {
+    }
+
+    public function onPreCreateForRender(PreCreateForRenderEvent $event): void
+    {
+        if ('ux:map' !== strtolower($event->getName())) {
+            return;
+        }
+
+        $attributes = $event->getInputProps();
+        $map = array_intersect_key($attributes, ['markers' => 0, 'center' => 1, 'zoom' => 2]);
+        $attributes = array_diff_key($attributes, $map);
+
+        $html = $this->mapRuntime->renderMap(...$map, attributes: $attributes);
+        $event->setRenderedString($html);
+        $event->stopPropagation();
+    }
+}
--- a/src/Map/tests/Kernel/TwigComponentKernel.php
+++ b/src/Map/tests/Kernel/TwigComponentKernel.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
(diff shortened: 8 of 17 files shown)


src/Map/CHANGELOG.md
```diff
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
--   Rename `render_map` Twig function `ux_map`
+-   Add `<twig:ux:map />` Twig component
+-   Update `ux_map` Twig function to accept an array 
 -   Deprecate `render_map` Twig function
+-   Rename `render_map` Twig function `ux_map`
```
Kocal: Do you think can we put:
```
-   Deprecate `render_map` Twig function
-   Rename `render_map` Twig function `ux_map`
```
in the first lines instead? 

It will be easier to understand the changelog when reading from top to bottom, so we can have something like:
```suggestion
-   Deprecate `render_map` Twig function
-   Rename `render_map` Twig function `ux_map
-   Add `<twig:ux:map />` Twig component
-   Update `ux_map` Twig function to accept an array 
```

src/Map/config/twig_component.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\UX\Map\Twig\UXMapComponentListener;
+
+return static function (ContainerConfigurator $container): void {
+    $container->services()
+        ->set('.ux_map.twig_component_listener', UXMapComponentListener::class)
+            ->args([
+                service('ux_map.renderers'),
+            ])
+            ->tag('kernel.event_listener', [
+                'event' => 'Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent',
```
Kocal: Can we use an import instead?

```suggestion
                'event' => PreCreateForRenderEvent::class,
```

src/Map/CHANGELOG.md
Kocal: Also, it is `<twig:ux:map />` or `<twig:UX:Map />` (I see you used `UX:Map` in `twig.component` tag)? Or is it can insensitive?

src/Map/src/MapFactory.php
```diff
@@ -0,0 +1,186 @@
...
+     *    position: array{lat: float, lng: float},
+     *    title?: string,
+     *    infoWindow?: array{
+     *      content?: string,
+     *      headerContent?: string,
+     *    },
+     *  }>,
+     * } $array The array representation of the map
+     */
+    public static function fromArray(array $array): Map
```
Kocal: Do you think it can makes sense to use the Symfony OptionsResolver here, to replace the multiple checks on `$array` properties?

src/Map/src/MapFactory.php
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private static function createMarker(array $marker): Marker
+    {
+        if (!\array_key_exists('position', $marker)) {
+            throw new \InvalidArgumentException('The "position" parameter is required.');
+        }
+        if (!\is_array($marker['position'])) {
+            throw new \InvalidArgumentException(\sprintf('The "position" parameter must be an array, "%s" given.', get_debug_type($marker['position'])));
+        }
```
Kocal: `createMarker` can be a hot point depending of markers count, do we really want to do all those checks? 
Couldn't we assume the user will use correct markers definition, and if it's not the case, we leverage all of this to PHP types instead?

Same comments for `infoWindow`, `title`, `extra`, etc... IMO you are doing too much things here, can't we simplify this code a lot?

src/Map/src/MapFactory.php
```diff
@@ -0,0 +1,186 @@
...
+
+        $lat = (float) $point['lat'];
+        $lng = (float) $point['lng'];
+        unset($point['lat'], $point['lng']);
+
+        if (\count($point) > 2) {
+            throw new \InvalidArgumentException(\sprintf('Unknown Point parameters: "%s".', implode('", "', array_keys($point))));
+        }
+
+        return new Point($lat, $lng);
```
Kocal: Given my previous comments, I imagine something like this (exceptions message can be rewritten): 
```suggestion
        return new Point(
            (float) $point['lat'] ?? throw new \InvalidArgumentException('Parameter "lat" is missing in "Point" definition.'), 
            (float) $point['lng'] ?? throw new \InvalidArgumentException('Parameter "lng" is missing in "Point" definition.')
        );
```

src/Map/src/Twig/MapRuntime.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * @param Map|array<string, mixed> $map
+     * @param array<string, mixed>     $attributes
+     */
+    public function renderMap(Map|array $map, array $attributes = []): string
+    {
+        if (\is_array($map)) {
+            if (isset($map['attr'])) {
```
Kocal: I don't see where `attr` is defined, am I missing something? Why do we need this key if we already have `$attributes`?

src/Map/src/Twig/UXMapComponent.php
```diff
@@ -0,0 +1,26 @@
...
+ * @internal
+ */
+final class UXMapComponent
+{
+    public int $zoom;
+
+    public array $center;
+
+    public array $markers;
+}
```
Kocal: Can be typed with PHPDoc

Kocal (commented):
Thanks for working on this!
It looks nice to me, but I have many comments on the validation part where I'm not 100% sure of its legitimacy (we use PHP types, and I'm afraid all those code/function calls can downgrade performances)

src/Map/src/MapFactory.php
smnandre: At first, I did not want to bloat in dependencies, but this is the exact role of ConfigResolver so ... if you too think we should leverage it, i'm all in :)

src/Map/src/MapFactory.php
smnandre: The main idea here is to provide a way to use "non-typed php objects" directly from Twig (hard-coded twig maps of data, json-like objects, etc).. so the performance trade-off is irrelevent imho in _this_ situation.

I wanted to simplify the data structure but decided to map the PHP classes for now, but something like this would be even better to me (and would require less code/checks/etc)

```twig
{{ ux_map({
    center: [4.51, 16.22],
    markers: [
       {lat: 45, lng: 22, content: 'sdsdf', id: 'store-22'}
   ]
)} }}
```

Also we 100% could accept "4.51 16.22" strings as point descriptor.

This is why a Factory made sense to me, for now and for the future. So now the (very valid) questions you raises here will help us decide what we need "now" and how we want to implement this!

---

In a near future i believe we will be able to offer dedicated tag/sub-component to declare markers, that will solve a lot of things DX-wise (and performance wise, as we then will be able to instanciate on-demand the infowindows)

src/Map/src/MapFactory.php
smnandre: Alternatively we could add static functions to the verious DTO classes (Point, Marker, etc) ... which seems the best place to implement this..

wdyt ?

smnandre: As I said, fromString() could also be interesting for some of them... 

and this could help later implement specific bridge options in the same manner.

src/Map/config/twig_component.php
smnandre: Yep thx

src/Map/CHANGELOG.md
smnandre: I think we use it some times "recent -> old" and some time in the other way so i really spent time to figure something.... and i hoped for some comment like this suggesting a direction haha :)

smnandre: I'v updated the CHANGELOG, tell me what you think, i tried to be the most DX-friendly possible

smnandre: > Also, it is <twig:ux:map /> or <twig:UX:Map /> (I see you used UX:Map in twig.component tag)? Or is it can insensitive?

Both `<twig:ux:icon` and now `<twig:ux:map `are indeed case-insensitive. This is something we need to sort now, i'm opening an issue to discuss it :)

smnandre: https://github.com/symfony/ux/issues/2118

src/Map/src/Twig/MapRuntime.php
smnandre: I'd like to offer a very simple function, like form ones ... as chaining to object in twig function calls is very midly readable..

So instead of 

```twig
{{  ux_map({ 
    center: ...
    zoom: ...
    markers: {
       ...
    }
}, {
   class: 'foo',
}) }} 
```

This allows

```twig
{{  ux_map({ 
    center: ...
    zoom: ...
    markers: {
       ...
    },
    attr: {
       class: 'foo',    
    }
}) }} 
```

The little things :)

src/Map/src/Twig/UXMapComponent.php
smnandre: Add types (this is purely informational for now, as the component is bypassed at runtime by the listener)

src/Map/src/MapFactory.php
Kocal: > At first, I did not want to bloat in dependencies, but this is the exact role of OptionsResolver so ... if you too think we should leverage it, i'm all in :)

We can try with the OptionsResolver yes, but we need to be careful about performances I think.

> Alternatively we could add static functions to the verious DTO classes (Point, Marker, etc) ... which seems the best place to implement this..
> 
> wdyt ? 

I think this alternative is more interesting, this way we can rely on PHP native typing, with a nice DX, and stays dependency-free.

src/Map/src/Twig/MapRuntime.php
Kocal: Thanks for the explanation, I can see a bit more clearly now.

But I'm not 100% fan, what about an alternative where we can abuse Twig named parameter? :
```twig
{{ ux_map(
    center={lat: 51.5074, lng: 0.1278},
    zoom=10,
    markers= [
        {lat: 51.5074, lng: 0.1278, title: 'London'},
        {lat: 48.8566, lng: 2.3522, title: 'Paris'},
        {lat: 40.7128, lng: -74.0060, title: 'New York'},
    ],
    attributes={
        class: 'foo',
        style: 'height: 400px; width: 100%;',
    }
) }}

```

Kocal:
See https://github.com/symfony/ux/pull/2152 as a lighter (but not complete yet) alternative

src/Map/src/Twig/MapRuntime.php
smnandre: If this is the only thing, no problem, just remove the first `{ `and last `}`

```diff
- {{ ux_map({
+ {{ ux_map(
    center: ... 
    zoom: ... 
   markers: { ... }, 
    attr: { class: 'foo', } 
- }) }}
+ ) }}
```


(named parameters now uses "`:`" and this will be the recommanded code style)

smnandre:
Yeah i'll use the fromArray static from your PR and we'll see... but i hope this will simplify a lot this one. 

Thanks !

src/Map/src/Twig/MapRuntime.php
smnandre: Well... not that simple... we need to keep "argumentes" as second arg of the function.

smnandre:
Integrated the ideas fom #2152

src/Map/src/Twig/MapRuntime.php
```diff
@@ -0,0 +1,59 @@
...
+            $map->center(Point::fromArray($center));
+        }
+        if (null !== $zoom) {
+            $map->zoom($zoom);
+        }
+        if (null !== $markers) {
+            foreach ($markers as $marker) {
+                $map->addMarker(Marker::fromArray($marker));
+            }
+        }
```
Kocal: We can make `$markers` not nullable, and simplify the code to:
```suggestion
        foreach ($markers as $marker) {
            $map->addMarker(Marker::fromArray($marker));
        }
```

Kocal: If the user pass a `Map` and `center` option, it will actually mutate the `Map` instance, are we sure about that? 

Shouldn't we call `->center`, `->zoom`, `->addMarker` only when the user didn't pass a `Map`?

src/Map/src/Twig/UXMapComponentListener.php
Kocal, line 1: I'm not sure to understand what this file does. Does it help with `ux_map()` or `<twig:ux:map>`?

Kocal (commented):
That looks much better to me, thanks :)

Still minor comments tho

src/Map/src/Twig/MapRuntime.php
smnandre: I was wondering too. Maybe we should do some cloning here ?

src/Map/src/Twig/UXMapComponentListener.php
Kocal: (yes it does :p, for other people: it's a faster alternative to _real_ Twig components)

src/Map/src/Twig/MapRuntime.php
Kocal: We can, but I don't know how much it can impact performances. 

I think we can throw an error if users try to re-configure center/zoom/markers on-the-fly, on an already existing Map, telling them to update the Map itself directly. 

WDYT?

smnandre: Better now ?

src/Map/src/Twig/MapRuntime.php
```diff
@@ -0,0 +1,63 @@
...
+    ): string {
+        if ($map instanceof Map) {
+            if (null !== $center || null !== $zoom || $markers) {
+                throw new \InvalidArgumentException('You cannot set "center", "markers" or "zoom" on an existing Map.');
+            }
+            return $this->renderer->renderMap($map, $attributes);
+        }
+
+        $map = new Map();
+        foreach ($markers as $marker) {
```
Kocal:
```suggestion
        foreach ($markers ?? [] as $marker) {
```
if you keep `null` as default value

src/Map/src/Marker.php
```diff
@@ -38,4 +49,28 @@ public function toArray(): array
...
+     * } $marker
+     *
+     * @internal
+     */
+    public static function fromArray(array $marker): self
+    {
+        if (!isset($marker['position'])) {
+            throw new InvalidArgumentException('The "position" parameter is required.');
+        }
+        $marker['position'] = Point::fromArray($marker['position']);
```
Kocal, line 68:
```suggestion
        $marker['position'] = Point::fromArray($marker['position'] ?? throw new InvalidArgumentException('The "position" parameter is required.'));
```

src/Map/src/Point.php
```diff
@@ -43,4 +43,16 @@ public function toArray(): array
             'lng' => $this->longitude,
         ];
     }
+
+    /**
+     * @param array{lat: float, lng: float}|array{0: float, 1: float} $point
+     */
+    public static function fromArray(array $point): self
+    {
+        if (isset($point['lat'], $point['lng'])) {
```
Kocal, line 52:
```suggestion
        if ($point['lat'] ?? null && $point['lng'] ?? null) {
```

so we can allow users to pass `0` as lat/lng

src/Map/src/Twig/MapRuntime.php
Kocal: Yes, thanks!

src/Map/src/Twig/UXMapComponentListener.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function onPreCreateForRender(PreCreateForRenderEvent $event): void
+    {
+        if ('ux:map' !== strtolower($event->getName())) {
+            return;
+        }
+
+        $attributes = $event->getInputProps();
+        $map = array_intersect_key($attributes, array_flip(['zoom', 'center', 'markers']));
```
Kocal: Can't we directly replace `array_flip(['zoom', 'center', 'markers'])` by `['zoom' => 0, 'center' => 1, 'markers' => 2])`? Since the output will always be identical

Kocal (approved):

Kocal (commented):
(sorry I approved by mistake)

src/Map/src/Point.php
smnandre:
```php
$pos = ['lat' => 0, 'lng' => 0];

echo isset($pos['lat'], $pos['lng']);

# 1
```

The only case would be `$pos = [0, 0]` and i don't mind that much 😅

src/Map/src/Twig/UXMapComponentListener.php
smnandre: I'm curious, why so ?

src/Map/src/Marker.php
smnandre: I'm not fan of exceptions inside a parenthesis to be honest..

src/Map/src/Twig/MapRuntime.php
smnandre: Yep

src/Map/src/Twig/UXMapComponentListener.php
Kocal: We can remove an useless function call, but I don't know if PHP/OPCache already optimize this call since we pass a static array.

smnandre: Ok i'll update it for the 3ops win 

(even if i'm really not a big fan if nano-optimizations like this :) )

src/Map/src/Marker.php
smnandre: I prefer not to change this one for now, but thanks for all the small optimizations suggested.. in the end i know this adds up 😃

src/Map/tests/Kernel/TwigComponentKernel.php
```diff
@@ -0,0 +1,63 @@
...
+use Symfony\Bundle\TwigBundle\TwigBundle;
+use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\UX\Map\UXMapBundle;
+use Symfony\UX\StimulusBundle\StimulusBundle;
+use Symfony\UX\TwigComponent\TwigComponentBundle;
+
+/**
+ * @internal
```
Kocal, line 24: Do we need to mark test files internal?

Kocal (approved):

Kocal:
Thanks @smnandre for working on this feature, this is much appreciated.

smnandre:
Thanks @Kocal for the help / suggestions / feedback ❤️ 

I'll add doc asap, but I now can work on demos with twig content, and i think it's gonna be more effective to showcase the component!

PR #2122, opened 2024-09-02
[Turbo] Render TurboStream (wip)

### LINKED ISSUE #2068: [Feature Idea] [Turbo Streams] [ux-turbo] Render view as Turbo Stream
# Preface
Before I dive in, to give some context I am dealing with the issue described below, and have implemented a solution, but I am not sure if this is the best approach, and I'm not even sure that I'm not misusing something in a way that creates the issue.
I am working on a project that is transitioning to using Symfony UX. We transitioned all the JS to Stimulus, and are making the application progressively more SPA-like with Turbo. A big part of that is using streams.

# The problem

## 1) Explosion of .turbo.stream.html.twig files
Often times I need to update a small piece of the page and find myself taking that piece of twig, splitting it as a partial and then including that in the original twig, and in a new twig with .turbo.stream in the name that looks like this:
```twig
<turbo-stream action="replace" target="smallPiece">
    <template>
        {{ include('_smallPiece.html.twig', parameters) }}
    </template>
</turbo-stream>
```
Soon the codebase became littered with those tiny turbo-stream wrapper files.
However there are also a few instances where I have more than one of these streams in the same twig.

## 2) Setting the request format when responding with a Turbo Stream
This is a minor complaint, but ties in with the proposal very well. My gripe here is that whenever I am converting a response to a turbo stream I need to remember to DI the Request and do `$request->setRequestFormat(TurboBundle::STREAM_FORMAT);`
This feels like a chore and adds visual clutter. I would much prefer to have a separate rendering method that would take care of it.

```PHP
public function renderSmallPiece(Request $request): Response
{
    $request->setRequestFormat(TurboBundle::STREAM_FORMAT);
    
    return $this->render('_smallPiece.turbo.stream.html.twig', ['param' => 'value']);
}
```

# The solution

I have a working solution, that may soon become the standard in my organisation, but I figured it's a good idea to pitch it here and see if there is a better approach, or get confirmation that I'm on the right track. If there is interest I will polish it up and submit a PR soon.

The proposed solution is a renderAsStream method as described above that is similar to the render method but takes two extra parameters - action and target. It wraps the view with the turbo-stream and template tags, and then sets the response's 'Content-Type' to TurboBundle::STREAM_MEDIA_TYPE, also eliminating problem 2).

Considering the case for multiple streams in the same response I also created a renderTurboStreams method that takes a collection of streams (defined by target, action, view, parameters) and renders them all the same way as above, but in one response.

The example above would become:
```PHP
public function renderSmallPiece(): Response
{
    return $this->renderAsStream('smallPiece', TurboAction::REPLACE, '_smallPiece.html.twig', ['param' => 'value']);
}
```

My current implementation depends on the abstractController from the framework bundle. It looks something like this:

```PHP
protected function renderTurboStreams(TurboStreamCollection $streams): Response
{
    $response = $this->render('TurboResponse/turboStreamsView.html.twig', ['streams' => $streams->all(),]);
    $response->headers->set('Content-Type', TurboBundle::STREAM_MEDIA_TYPE);
    
    return $response;
}
```

The $streams parameter is an array of TurboStream objects - readonly with target, action, view, parameters
The turboStreamsView iterates over them and wraps each one with the turbo-stream and template tags with the corresponding action and target.

## **NB**
**This was done to prevent code duplication from the render method. However it means the new functionality must either be added to the AbstractController, or as my current solution it lives in TurboAbstactController which extends the frameworkBundle's AbstractController. Since the frameworkBundle is decoupled from the UX stuff I guess this is more plausible.  
In order to use this one would only need to change their controllers to extend the new TurboAbstractController**
---

The renderAsStream method is just a shorthand for when you only need one stream. It takes (target, action, view, parameters), creates a TurboStream object from the parameters, adds it to a collection and invokes the renderTurboStreamsMethod:

```PHP
protected function renderAsStream(string $target, TurboAction $action, $view, array $parameters = []): Response
{
    $streams = (new TurboStreamCollection())->add($target, $action, $view, $parameters);
    
    return $this->renderTurboStreams($streams);
}
```

It only serves the purpose of reducing boilerplate

The example from above:
```PHP
public function renderSmallPiece(): Response
{
    return $this->renderAsStream('smallPiece', TurboAction::REPLACE, '_smallPiece.html.twig', ['param' => 'value']);
}
```
can use the other method instead:
```PHP
public function renderSmallPiece(): Response
{
    return $this->renderTurboStreams(
        (new TurboStreamCollection())
            ->add('smallPiece', TurboAction::REPLACE, '_smallPiece.html.twig', ['param' => 'value'])
    );
}
```

and in the case of needing more than one stream:
```PHP
public function renderSmallPiece(): Response
{
    return $this->renderTurboStreams(
        (new TurboStreamCollection())
            ->add('smallPiece', TurboAction::REPLACE, '_smallPiece.html.twig', ['param' => 'value'])
            ->add('otherPiece', TurboAction::REPLACE, '_otherPiece.html.twig', ['param' => 'value'])
    );
}
```

I am curious if others have the same issue and if so do you find this useful.
If there is a fundamental flaw in my approach that I've overlooked please let me know.
If I am overthinking this and missing an easy solution please point me in the right direction.
If this is the right direction, but there is a better way I or some tweaks are needed I would welcome community feedback.

If there is interest in this I can release it as a bundle soon, and in the ideal scenario it will become an official part of the Symfony UX initiative

DRaichev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #2068 
| License       | MIT

This is a WIP to aid in further discussions of the RFC implementation

@smnandre As discussed in #2068 I have opened a PR with some rough outline of the implementation

It is far from ready, but I think it is a good start. I can continue to work on this, but I will need some direction.
I understand now what you meant earlier - the bundle is an independent component not tied to the symfony framework's AbstractController, nor HttpFoundation nor even Twig

So I think my initial concept should be split into 2 parts - one is the abstract implementation of rendering streams, or rather wrapping a view in a turbo stream, and the other part should be specific to the full-fat symfony framework and it will  build on top of the first to provide an easy way to go from a bunch of turbo streams to a HttpFoundation\Response with the correct content type, etc.

### DIFF
--- a/src/TurboStreamRender/src/Core/TurboStream.php
+++ b/src/TurboStreamRender/src/Core/TurboStream.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+
+namespace src\Core;
+
+use src\Enum\TurboAction;
+
+readonly class TurboStream
+{
+    public function __construct(
+        public string $target,
+        public TurboAction $action,
+        public string $view,
+        public array $parameters,
+    ) {
+    }
+}
--- a/src/TurboStreamRender/src/Core/TurboStreamCollection.php
+++ b/src/TurboStreamRender/src/Core/TurboStreamCollection.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+
+namespace src\Core;
+
+use src\Enum\TurboAction;
+
+class TurboStreamCollection
+{
+
+    /**
+     * @param TurboStream[] $streams
+     */
+    public function __construct(private array $streams = [])
+    {
+    }
+
+    public function add(string $target, TurboAction $action, $view, array $parameters = [])
+    {
+        return $this->addStream(new TurboStream($target, $action, $view, $parameters));
+    }
+
+    public function addStream(TurboStream $stream): self
+    {
+        $this->streams[] = $stream;
+
+        return $this;
+    }
+
+    /**
+     * @return TurboStream[]
+     */
+    public function all(): array
+    {
+        return $this->streams;
+    }
+}
--- a/src/TurboStreamRender/src/Core/TurboStreamRenderer.php
+++ b/src/TurboStreamRender/src/Core/TurboStreamRenderer.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace src\Core;
+
+use src\Enum\TurboAction;
+use TurboStreamRendererInterface;
+
+class TurboStreamRenderer implements TurboStreamRendererInterface
+{
+    /**
+     * This is a hacky way to reuse the code of renderView form AbstractController - just pass the method as a callable to the renderer constructor.
+     * @var callable $renderer
+     */
+    protected $renderer = null;
+
+    public function __construct(callable $renderer)
+    {
+        $this->renderer = $renderer;
+    }
+
+    /**
+     * @throws \InvalidRendererException
+     */
+    public function renderView(string $view, array $parameters = []): string
+    {
+        try {
+            $renderer = $this->renderer;
+            return $renderer($view, $parameters);
+        }catch (\Throwable $e) {
+            throw new \InvalidRendererException($e->getMessage(), $e->getCode(), $e);
+        }
+    }
+
+    public function renderAsStreamView(string $target, TurboAction $action, string $view, array $parameters = []): string
+    {
+        $streams = (new TurboStreamCollection())->add($target, $action, $view, $parameters);
+
+        return $this->renderTurboStreamsView($streams->all());
+    }
+
+    /**
+     * @throws \InvalidRendererException
+     */
+    public function renderTurboStreamsView(array $streams): string
+    {
+        $response = "";
+        foreach ($streams as $stream) {
+            $content = $this->renderView($stream['view'], $stream['parameters']);
+            $response .= $this->wrapWithTurboStream($stream['target'], $stream['action'], $content);
+            $response .= "\n";
+        }
+
+        return $response;
+    }
+
+    private function wrapWithTurboStream(string $target, TurboAction $action, string $content): string
+    {
+        return <<<HTML
+        <turbo-stream action="$action->value" target="$target">
+            <template>
+                $content
+            </template>
+        </turbo-stream>
+        HTML;
+    }
+}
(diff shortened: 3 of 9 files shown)


smnandre:
Nice, thank you!

Could you change the PR status to DRAF, that way everyone knows it's a work-in-progress ? 

I think the best now would be some examples of what it would allow (like concrete scenarios / usages, showing the final goal and where the DX / entrypoints would be used/called) 

I'll have a better look later in the week :)

javiereguiluz:
Question: is this idea something we still want to do ... or is it enough with the changes that we merged in #2227 and #2196?

DRaichev:
> Question: is this idea something we still want to do ... or is it enough with the changes that we merged in #2227 and #2196?

I think the referenced changes cover 90% of what I wanted to achieve.
In fact it is very similar to what I had as a WIP implementation at my company.

In terms of UX what's missing/different is the ability to render a twig file.
With those new changes you can use the TurboResponse and pass it streams, but they can only get rendered html.

What I'd like to do is pass a .twig file along with the required params, and the turbo action and it would render the twig to html, wrap that in a turbo stream and then wrap the stream(s) in the TurboResponse.

What I did in my internal WIP is make a render function that uses the regular render and takes the content of that and adds it. I can do the same now, but it is kind of ugly. 
I am using the full symfony framework and I would like to use the abstract controller's methods, but they are private and on top of that some of the logic is split in a way that makes it inconvenient to reuse.

As an intermediate step I will use the improvements added in #2196 and #2227.
In order to achieve my vision some code would need to change in the framework-bundle AbstractController which I think is outside of the scope/control of this community, but if you have any advice, or think it is possible I'd appreciate it. If not I think this issue & PR can be closed.

P.S.

One more thing - the current implementation also has no generic action (only the predefined turbo actions) and what's more the class is marked as final, which does not allow me to add support for more actions. I think the wrap function should be public or have a public function "custom" that exposes the functionality. 

I really like this library: https://github.com/marcoroth/turbo_power
It adds additional simple actions for example I use set/delete attributes to control enable/disable UI elements

Maybe this should be added, while my other idea can be ditched as it is too dependent on the framework bundle / specific renderer implementation

DRaichev:
Since this overlaps some functionality implemented in #2196, the difference is moved to another PR #2298

PR #2127, opened 2024-09-03
[Icons] Icon aliases

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #
| License       | MIT

This PR introduce a first implementation of this feature, asked in #1767 

Icon **aliases** can now be defined in the bundle configuration, to improve the DX for most used icons in a project (logo, socials, ...)

```yaml
ux_icons:
    aliases:
        dots: 'clarity:ellipsis-horizontal-line'
```

Now

```twig
    {{ ux_icon('dots') }}

    {# is the same as #}

    {{ ux_icon('clarity:ellipsis-horizontal-line') }}
```

Or... with the HTML syntax

```twig
    <twig:ux:icon name="dots" />

    {# is the same as #}

    <twig:ux:icon name="clarity:ellipsis-horizontal-line" />
```

### DIFF
--- a/src/Icons/CHANGELOG.md
+++ b/src/Icons/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.20.0
+
+-   Add `aliases` configuration option to define icon alternative names.
+
 ## 2.19.0
 
 -   Add `ignore_not_found` option to silence error during rendering if the 
--- a/src/Icons/config/services.php
+++ b/src/Icons/config/services.php
@@ -56,6 +56,8 @@ return static function (ContainerConfigurator $container): void {
         ->set('.ux_icons.icon_renderer', IconRenderer::class)
             ->args([
                 service('.ux_icons.icon_registry'),
+                abstract_arg('default_icon_attributes'),
+                abstract_arg('icon_aliases'),
             ])
 
         ->alias('Symfony\UX\Icons\IconRendererInterface', '.ux_icons.icon_renderer')
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -336,6 +336,30 @@ Now, all icons will have the ``fill`` attribute set to ``currentColor`` by defau
     # renders "user-profile.svg" with fill="red"
     {{ ux_icon('user-profile', {fill: 'red'}) }}
 
+Icon Aliases
+~~~~~~~~~~~~
+
+You can define aliases for icons in your configuration, which is helpful if 
+you want to use a different or shorter name for an icon in your templates:
+
+.. code-block:: yaml
+
+    # config/packages/ux_icons.yaml
+    ux_icons:
+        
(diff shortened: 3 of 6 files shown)


javiereguiluz:
Just asking ... can you use `:` in the alias name? E.g. if someone wants to group all aliases under a common prefix (e.g. `site:...` or `acme:...`)

smnandre:
That's for sure the next step!

Xbirdfr:
Hello, thank you for this implementation! 

To have thought a little about this functionality, I propose the following modification:

Put an @ before the name of the alias

 This allows you to add an AliasNotFound error which would be easier to debug, and to only check the configuration file when calling an alias. 
Also it increases readability: @dots directly gives us the information that the icon is an alias, and it also allows us to do things like 
@user:login @user:logout @admin:login

javiereguiluz:
@Xbirdfr the problem with that is that it introduces a new convention that all users of UX Icon must learn, which causes friction.

In any case, is `@` forbidden in icon aliases? If it's not, what you are asking it's already possible to do. If it's forbidden, let's double check if we should really forbid it.

Xbirdfr:
@javiereguiluz I understand that, but as Icones Alias will be a new feature, I will cause less friction 

Actually, if `@dots` is not referenced in the config, an IconNotFound error will be thrown, and this is not the good error message 

With an @ we clearly know if we want an alias or not while using ux icon, and allow a better readability and debugging. 

But it's maybe too much and I understand this, but I really think this will be a good way to handle alias

smnandre:
> oblem with that is that it introduces a new convention that all users of UX Icon must learn, which causes friction.

@javiereguiluz yep absolutely! 

And we decided to avoid it in the TwigComponent shared by bundles, so it would be a bit strange to introduce it here.

smnandre:
>  @user:login @user:logout @admin:login

@Xbirdfr  This is precisely what i did not want to handle _for now_.  We have the ` : ` to identify a namespace/ icon set. 

Full disclosure: i will open a PR after this one to implement the long-waiting "configuration per icon sets/directorys" ... so let's talk about namespaces aliases afterwards 😅 

I promised the simple aliases long time ago, and we have been poked again recently ... so this is more the context of this PR :) 

Personally, I will use the aliases in several projects, to have simple "one-word-domain-icon-names"  for the 10/12 most used icons (which come from different sets -- excluding here any partial mapping). Something like this:

```yaml
ux_icons:
    aliases:
    
        # Logos
        github: 'simple-icons:github' 
        twitter: ' pajamas:x'
        php: 'bxl:php'

       # Actions
       delete: 'tdesign:delete'
       add: 'material-symbols:add'
       
       # ...
```

Kocal (approved):

WebMamba (approved):

javiereguiluz:
Simon, thanks for yet another nice feature contribution!

Issue #2070, opened 2024-08-15
[TwigComponent] Cva in html_extra


smnandre:
It seems the cva function will be merged into twig/html-extra (see https://github.com/twigphp/Twig/pull/4006) 

How do we handle the transition ? 

* Add a conflict with the future version of html-extra here seems not a good idea, as this would be blocking for future updates
* Add a deprecation layer here ?
* Ask people to require twig/html-extra ? 
* What do we do about the documentation ?

Kocal:
We should be careful here since the `cva()` definition changes from here and from https://github.com/twigphp/Twig/pull/4006:
1. actual: `public function cva(array $cva): CVA`
2. https://github.com/twigphp/Twig/pull/4006: `public static function cva(array|string|null $base = null, ?array $variants = null, ?array $compoundVariants = null, ?array $defaultVariant = null): Cva`

This first one takes a big array, where the second one takes advantage of Twig named arguments (and that's super nice! I prefer this syntax too), but, we should find a way to correctly handle the migration from UX to twig/html-extra for `cva`. 

If https://github.com/twigphp/Twig/pull/4006 is merged, it has no interest to keep CVA's code in UX and we will need to remove it, in a **major** version.

If release a new major version is not wanted - which I agree because it will affect all other UX packages -, I think we can do it smoothly by:
1. makes `symfony/ux-twig-component` requires `twig/extra-html`
2. implement a "bridge" in `symfony/ux-twig-component` to nicely migrate and warn the user about passing an array to `cva`
3. trigger a deprecation, telling the user to manually requires `twig/extra-html` and pass named arguments
4. tag a new minor version

Then, for a new major version, remove this deprecation layer and `twig/extra-html` dependency.

For the documentation, I believe we can do something like:
> Since 2.2x, `cva` function has moved to `twig/extra-html`. 
> We recommend you to upgrade your `cva` usages to (...),  since `cva` will be removed from UX Twig Components (...).

WDYT?

smnandre:
> 1. makes symfony/ux-twig-component requires twig/extra-html

I'm not in favor of requiring an external package (this would need twig-extra too) ... to deprecate one

And that does not change the fact someone updating twig/extra-html will break its app, no ? Two functions cannot be registered with the same name i think.

smnandre:
@WebMamba renamed the `twig/html-extra` function  `html_cva` to ease this transition / diminish any conflict risks (thank you for this)

So indeed, next steps here:
[...]

PR #2138, opened 2024-09-09
[Map] Adjust changelogs and fix `render_map` deprecated version

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

While working on #2070, I noticed Map's changelogs were still containing `Unreleased` instead of the version number.
I've also changed the version where `render_map` twig function have been deprecated, to match the reality (the 2.20 will be the next release!).

### DIFF
--- a/src/Map/CHANGELOG.md
+++ b/src/Map/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-## Unreleased
+## 2.20
 
 -   Rename `render_map` Twig function `ux_map`
 -   Deprecate `render_map` Twig function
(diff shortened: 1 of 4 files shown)


smnandre (approved):
I think "Unreleased" was just a convention (because some component have no changes for multiple versions) so this was to avoid writing wrong "previous + 1" versions in the changelog.

Now i'm 100% ok with both methods, and we'll have a better look for next release :)

smnandre:
I'm more concerned by the failing js tests :|

Kocal:
> I'm more concerned by the failing js tests :|

I gonna look asap

PR #2144, opened 2024-09-10
[TwigComponent] Deprecate `cva` twig function

### LINKED ISSUE #2070: [TwigComponent] Cva in html_extra

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #2070 
| License       | MIT

Use the new `Twig\DeprecatedCallableInfo` added in https://github.com/twigphp/Twig/pull/4291, it help us to specify the package where the alternative can be found, which was not the case before.


src/TwigComponent/doc/index.rst
```diff
@@ -1158,6 +1158,11 @@ Component with Complex Variants (CVA)
 
     The ``cva`` function was added in TwigComponents 2.16.
 
+.. deprecated:: 2.20
+
+    The ``cva`` function was deprecated in TwigComponents 2.20, and will be removed in 3.0.
+    The function is now called `html_cva`_ and is now part of ``twig/html-extra:^3.12`` package.
```
smnandre:
```suggestion
    The function is now provided by the ``twig/html-extra:^3.12`` package under the name `html_cva`_.
```

?

src/TwigComponent/src/Twig/ComponentExtension.php
```diff
@@ -42,7 +43,15 @@ public function getFunctions(): array
     {
         return [
             new TwigFunction('component', [$this, 'render'], ['is_safe' => ['all']]),
-            new TwigFunction('cva', [$this, 'cva']),
+            new TwigFunction('cva', [$this, 'cva'], [
+                'deprecation_info' => new DeprecatedCallableInfo(
```
smnandre: This won't be usable by anyone not perfectly up to date.. and i don't want to force people upgrade for a deprecation :/ 

Can't we use the other syntax ? Or just trigger in the method ?

Kocal: That was my feeling too, but this new deprecation feature was legit developed for us :p 

But you're right, I'm not a fan of forcing a Twig update to display a deprecation message more appropriately. Let's do it simple

Kocal: Well, I think we don't really have the choice in fact, since options `deprecated`, `alternative` are deprecated in favor of `deprecation_info`, see checks for TwigComponent, LiveComponent, Map, etc...

Those deprecations will appear to the end user... right? 
I think we will have to wait for a new Twig release and use `deprecation_info` everywhere needed.

smnandre: No no no :) 

You cannot assume users will have the new Twig version installed. 

One way could be to check the existence of the new DeprecatedCallableInfo class and act accordingly .. wdyt ?

smnandre (approved):
(Let's update the src/TwigComponent/CHANGELOG too if you have time)

PR #2146, opened 2024-09-10
Upgrade Vitest, use a real browser for Map Bridges tests

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Follow https://github.com/symfony/ux/pull/2138#issuecomment-2338794108.

It looks like Google Maps changed a bit their implementation and now manipulates [`SVGImageElement`](developer.mozilla.org/en-us/docs/web/api/svgimageelement) and [`SVGAnimatedString`](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString): 
image

But, this API is not supported by https://github.com/capricorn86/happy-dom. 

Instead of reimplementing things ourselves, I think it's more clever to use a **real browser** (here Chromium) through [Playwright](playwright.dev/), when using [Vitest's Browser mode](https://vitest.dev/guide/browser/).

This PR also upgrade Vitest from ^0.34 to ^2.0, and Vite from 4.5.0 to 5.4.3, which comes with https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated (that I fixed in a second commit)


Kocal:
For failing tests in `tests-php`, see https://github.com/symfony/ux/pull/2144#discussion_r1751613174

smnandre:
> Symfony UX / tests-js (pull_request) Successful in 2m 

Could you isolate them ? I don't think we need to run all tests in Browser every time we update a README file (i may exagerate a little bit here 👼 )

Kocal:
> > Symfony UX / tests-js (pull_request) Successful in 2m
> 
> Could you isolate them ? I don't think we need to run all tests in Browser every time we update a README file (i may exagerate a little bit here 👼 )

I don't get what you mean, Playwright/browser tests are ran only for Map's bridges when their **code** change.
Vitest's Browser mode is explicitly enabled for Map's bridges, and the CI run `tests-js` when code (outside documentation) has been modified. 🤔 

Note that, in previous `tests-js` CI jobs, the duration is a bit higher than before but that's fine IMO, see https://github.com/symfony/ux/actions/runs/10644425330/job/29509168600?pr=2113

smnandre:
Well i did not know... i'm sorry for asking, and... thank you very very much for all this :) 💐

smnandre (approved):

Kocal:
c520963a50d5847453f3395aaa0d4ffb

PR #2149, opened 2024-09-11
[Icons] Add support for int/float attribute to `ux_icon` function

### LINKED ISSUE #2148: [TwigComponent][Icons] Inconsistency between allowed attributes
Twig components allows to provide an integer as attibute values and casts it, while `UX:Icon` doesn't allow it.

Valid example:
```twig
<twig:Test a="1" b="{{ 2 }}">
  test
</twig:Test>
```
Currently invalid:
```twig
<twig:UX:Icon a="1" b="{{ 2 }}" />
```

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2148 
| License       | MIT

I'm not really sure why do we have checks on attributes values, do you remember why @kbond? 🙏🏻


src/Icons/tests/Fixtures/templates/template1.html.twig
```diff
@@ -2,8 +2,10 @@
     <li id="first">{{ ux_icon('user', {class: 'h-8 w-8'}) }}</li>
     <li id="second">{{ ux_icon('user', {"aria-label": "AriaLabel"}) }}</li>
     <li id="third">{{ ux_icon('sub:check', {'data-action': 'string "with" quotes'}) }}</li>
+    <li id="fourth">{{ ux_icon('sub:check', { height: 24, width: "24" }) }}</li>
```
smnandre, line 5: Is this valid SVG ? oO

src/Icons/src/Icon.php
```diff
@@ -189,8 +189,8 @@ public function withAttributes(array $attributes): self
                 throw new \InvalidArgumentException(\sprintf('Invalid attribute name "%s".', $name));
             }
 
-            if (!\is_string($value) && !\is_bool($value)) {
-                throw new \InvalidArgumentException(\sprintf('Invalid value type for attribute "%s". Boolean or string allowed, "%s" provided. ', $name, get_debug_type($value)));
+            if (!\is_string($value) && !\is_bool($value) && !\is_int($value) && !\is_float($value)) {
```
smnandre, line 192: Do you have any real-life usage for a float ? 
```suggestion
            if (!\is_string($value) && !\is_bool($value) && !\is_int($value)) {
```

smnandre: You need to adjust the docblock here too. And probably on the renderer. But i'd like to have more explanation than "fix" :))

smnandre (changes requested):
It is not a bug fix..

But i'm fully open to discuss the feature _if_ there is a real usage behind :)

smnandre:
This works already

```twig
<twig:ux:icon height="48" width="64" />
```

So the problem is when you want to pass some dynamic value (and i really would love to have an example) 

```twig
<twig:ux:icon height="48" width="{{ 64 / 2 }}" />
```


But i'm not ok to change it in Icon ... as a SVG tag does not have "float" or "int" values, but only strings.

So maybe we can (string) values in the Renderer :|

src/Icons/tests/Fixtures/templates/template1.html.twig
smnandre: Ok so @norkunas gave me an example.... `data-id="22"` is indeed perfectly valid (and legit)

Kocal:
> This works already
> 
> ```
> <twig:ux:icon height="48" width="64" />
> ```
> 
> So the problem is when you want to pass some dynamic value (and i really would love to have an example)
> 
> ```
> <twig:ux:icon height="48" width="{{ 64 / 2 }}" />
> ```

Yes, that's what is described in issue #2148.

> But i'm not ok to change it in Icon ... as a SVG tag does not have "float" or "int" values, but only strings.

In absolute terms, all the values you pass to HTML attributes are strings, but are then interpreted specially.
For example, `<input type="range">` which accepts two attributes [`min`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#min) and [`max`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#max), you pass strings but they need to be a valid (_stringified_) number.

But here, we do not speak about the SVG tag but the `UX:Icon` component instead. Which IMO should behave like other `TwigComponent` and correctly auto-cast int/string values, ex:
```php
#[AsTwigComponent]
final class AutocastInt
{
    public string $min;
    public int $max;
}
```
```twig
<twig:AutocastInt min="1" max="10" />
<twig:AutocastInt :min="1" :max="10" />
<twig:AutocastInt min="{{ 1 }}" max="{{ 10 }}" />
```
which renders as expected:
image


Like in Vue or React/JSX, the following codes are correctly interpreted: 
```vue
<input type="range" :min="1" max="120" />
```
```jsx
<input type="range" min="1" max={10} />
```
---

Anyway, I don't really understand why do we need such this validation, can't we rely on `ComponentAttributes` and let it deal with attributes rendering?

smnandre:
> But here, we do not speak about the SVG tag but the UX:Icon component instead. Which IMO should behave like other TwigComponent and correctly auto-cast int/string values, ex:

This is not directly related to the UX:Icon `component` (that does not really exist, in fact 😓 ) ..  so we can either force the string in the Listener, in the Renderer, or ... in the Icon as you did.

But i don't want to store int as "int", but as strings in Icon attributes property.. so what do you prefer ?

smnandre:
As i said, I agree to cast it, just not internally :)

src/Icons/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 2.20.0
 
 -   Add `aliases` configuration option to define icon alternative names.
+-   Fix `Icon` component to accept `integer` or `float` attributes.
```
smnandre:
```suggestion
-   Add support for `int` and `float` attribute values in `<twig:ux:icon />`.
```

smnandre (approved):

smnandre:
Thanks for fixing this bug Hugo.

PR #2154, opened 2024-09-11
[Dropzone] Enable file replacement via "drag-and-drop"

### LINKED ISSUE #1480: [UX Dropzone] drag'n'drop replacement fail
Hello,

With symfony ux dropzone, you can drag and drop files to add them to the dropzone.
Once a file is in the zone (file A), if you want to replace it with another file (file B), drag'n'drop doesn't work: file A isn't replaced in the dropzone by file B.
And file B opens in a new browser tab.

Is this a known problem? How can I solve it, please?

Thanks

uxdropzone

rrr63:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1480
| License       | MIT

Adding `dragenter` event to allow file replacement

### DIFF
--- a/src/Dropzone/CHANGELOG.md
+++ b/src/Dropzone/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.20
+
+-   Enable file replacement via "drag-and-drop"
+
 ## 2.13.2
 
 -   Revert "Change JavaScript package to `type: module`"
--- a/src/Dropzone/assets/dist/controller.d.ts
+++ b/src/Dropzone/assets/dist/controller.d.ts
@@ -13,5 +13,7 @@ export default class extends Controller {
     clear(): void;
     onInputChange(event: any): void;
     _populateImagePreview(file: Blob): void;
+    onDragEnter(): void;
+    onDragLeave(event: any): void;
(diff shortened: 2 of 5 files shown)


smnandre:
Hi @rrr63 ! Thank you for this PR

Could you add just a screenshot or two, for other to understand what you're adding ?

I'm wondering, you add on dragEnter, but you don't revert if the drag is canceled  ? 

Also, it would be great to add a test or two :)

rrr63:
Hi @smnandre ,

You're right, i canceled when the drag leaves, and added a test for both, enter and leave.
Both, enter and leave are in the same test to get the right drag area from enter event.

Demo :
dropzone

smnandre:
Wow, thank you very much! I'll try tonight but this will probably more be this week-end :)

smnandre:
Could you add a entry in the CHANGELOG (inside the Dropzone folder) ? Next version will be 2.20

rrr63:
Sure i can, just copied the title  :)

smnandre (approved):

smnandre:
(did something weird here, but this should be / will be merged ... sorry for my mittens)

smnandre:
Thank you @rrr63.

PR #2155, opened 2024-09-11
[DataTables] Add DataTables UX integration

pentiminax:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #...
| License       | MIT

I developed an integration of the [DataTables ](https://datatables.net/) library.

I followed the same development pattern as the Chart.js UX integration.

 I haven't found how to allow developer to use his own CSS framework because DataTables uses different npm package for each CSS framework : https://datatables.net/download/npm

I dont know if this package can be relevant for Symfony UX.


src/DataTables/assets/src/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+    declare readonly viewValue: any;
+
+    static values = {
+        view: Object,
+    };
+
+    private table: DataTable | null = null;
+
+    connect() {
+        if (isDataTableInitialized) {
```
pierredup: This will prevent using multiple datatables on a single page. This can be a private prop on the controller

src/DataTables/assets/src/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+            throw new Error('Invalid element');
+        }
+
+        const payload = this.viewValue;
+
+        this.dispatchEvent('pre-connect', {
+            config: payload,
+        });
+
+        new DataTable(this.element as HTMLElement, payload);
```
pierredup: This is a missing an assignment to `this.table`

src/DataTables/doc/index.rst
```diff
@@ -0,0 +1,146 @@
...
+    // ...
+    use Symfony\UX\DataTables\Builder\DataTableBuilderInterface;
+    use Symfony\UX\DataTables\Model\DataTable;
+
+    class HomeController extends AbstractController
+    {
+        #[Route('/', name: 'app_homepage')]
+        public function index(DataTableBuilderInterface $tableBuilder): Response
+        {
+            $table= $tableBuilder->('usersTable');
```
pierredup: This is missing a method call to `createDataTable`

src/DataTables/doc/index.rst
```diff
@@ -0,0 +1,146 @@
...
+                'columns' => ['First name', 'Last name'],
+                'data' => [
+                    ['John', 'Doe'],
+                    ['Jane', 'Smith'],
+                ],
+            ]);
+
+            $table->setOptions([
+                'order' => [
+                    ['idx': 1, 'dir': 'asc']
```
pierredup: This is not a valid array syntax, it should use `=>` instead of `:`

src/DataTables/doc/index.rst
```diff
@@ -0,0 +1,146 @@
...
+        disconnect() {
+            // You should always remove listeners when the controller is disconnected to avoid side effects
+            this.element.removeEventListener('datatables:pre-connect', this._onPreConnect);
+            this.element.removeEventListener('datatables:connect', this._onConnect);
+        }
+
+        _onPreConnect(event) {
+            // The table is not yet created
+            // You can access the config that will be passed to "new DataTable()"
+            console.log(event.detail.config);
```
pierredup, line 109: Looking at the Controller, it looks like this should be `event.detail.table` instead

pierredup (commented):
Just a very quick initial review

src/DataTables/doc/index.rst
pentiminax: No, for the pre-connect event the config is being dispatched : 

```js
this.dispatchEvent('pre-connect', {
    config: payload,
});
```

But for the connect event, it's the table :

```js
this.dispatchEvent('connect', { table: this.table });
```

smnandre:
Hello @pentiminax ! Thank you very much for this PR.

We need to discuss it and all the team is not available at the moment, but i'll come back to you as soon as i can

Kocal:

Thanks!

Status: Needs Work

pentiminax:
Hi @Kocal 

I’ve done a lot of work on the DataTables bundle since opening this PR: https://github.com/pentiminax/ux-datatables

Would you like me to backport all the changes to this one?

Kocal:
@pentiminax I'm sorry about the silence here. You asked a direct question back in April about retargeting and backporting your work, and nobody answered it for five months, and before that the discussion @smnandre mentioned in 2024 never actually happened. That's on us, not you.

In the meantime your standalone bundle, `pentiminax/ux-datatables`, has clearly become the real project: 139 releases, v0.85.0 shipped on September 2, a commit as recently as yesterday, and around 11,500 downloads total (roughly 1,390 a month). This PR is a two-year-old snapshot next to that.

Given that, the right call is to close this rather than fork a living project into a frozen copy here that would lag behind what you're already shipping. I'm closing it, but the door isn't shut for good: if you ever want to talk about bringing it into symfony/ux properly, let's have that conversation for real instead of through a stale PR. Thanks for the work and for your patience with us.

pentiminax:
@Kocal Thanks a lot for the message! No worries at all, I completely understand.

I’m happy to keep working on the standalone bundle for now, and I’d definitely be open to discussing a proper integration into Symfony UX in the future.

Thanks again for taking the time to reply and for the kind words!

PR #2156, opened 2024-09-12
[Icons] Configure icon sets: path, alias & icon attributes

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #
| License       | MIT

I'm opening now if you guys have feedback, and i will add more tests the next couple days

> [!NOTE]  
> This PR brings 3 new features: iconset aliases, custom path and attributes.  So let's focus on that 😄 

```yaml
ux_icons:
    icon_dir: '%kernel.project_dir%/assets/icons'

    # Default icon set
    default_icon_attributes:
        class: 'icon'
        fill: 'currentColor'
        width: '24'
        height: '24'

    icon_sets:
        # Local icon set    (prefix mapped to a local directory)
        flags:
            path: '%kernel.project_dir%/assets/images/flags'

        # Remote icon set   (prefix mapped to an icon set identifier)
        lu:
            alias: 'lucide'

        # Icon set configuration (work for local/remote)
        oc:
            icon_attributes:
                class: 'oc-icon'       # Replace the default class
                stroke: 'none'      # Add a new attribute
                fill: false         # Use "false" to remove a default attribute
```

### Default attributes

```twig

<twig:ux:icon name="anything" />

{# Renders #}
<svg class="icon" fill="currentColor" width="24" height="24"> ... </svg>
```

### Local icons: path

```twig

<twig:ux:icon name="flags:fr" />

{# Renders:  #}
<svg class="icon" fill="currentColor" width="24" height="24">
    <!-- file assets/images/flags/fr.svg -->
</svg>
```

### Remote icons: alias 

```twig

<twig:ux:icon name="lu:circle-check" />

{# Renders:  #}
<svg class="icon" fill="currentColor" width="24" height="24">
    <!-- file lucide:circle-check.svg from iconify -->
</svg>
```

### Icon set attributes

```twig

<twig:ux:icon name="oc:check" />

{# Renders:  #}
{# Attributes are merged into the default attributes #}
<svg class="oc-icon" stroke="none" width="24" height="24">
    <!-- file oc:check.svg from local or remote -->
</svg>
```

### DIFF
--- a/src/Icons/src/DependencyInjection/UXIconsExtension.php
+++ b/src/Icons/src/DependencyInjection/UXIconsExtension.php
@@ -42,32 +42,66 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
                 ->variableNode('default_icon_attributes')
                     ->info('Default attributes to add to all icons.')
                     ->defaultValue(['fill' => 'currentColor'])
+                    ->example(['class' => 'icon'])
+                ->end()
+                ->arrayNode('icon_sets')
+                    ->info('Icon sets configuration.')
+                    ->defaultValue([])
+                    ->normalizeKeys(false)
+                    ->useAttributeAsKey('prefix')
+                    ->arrayPrototype()
+                        ->info('the icon set prefix (e.g. "acme")')
+                        ->children()
+                            ->scalarNode('path')
+                                ->info("The local icon set directory path.\n(cannot be used with 'alias')")
+                                ->example('%kernel.project_dir%/assets/svg/acme')
+                            ->end()
+                            ->scalarNode('alias')
+                                ->info("The remote icon set identifier.\n(cannot be used with 'path')")
+                                ->example('simple-icons')
+                            ->end()
+                            ->arrayNode('icon_attributes')
+                                ->info('Override default icon attributes for icons in this set.')
+                                ->example(['class' => 'icon icon-acme', 'fill' => 'none'])
+                                ->normalizeKeys(false)
+                                ->variablePrototype()
+                                ->end()
+                            ->end()
+                        ->end()
+                    ->end()
+                    ->validate()
+                        ->ifTrue(fn (array $v) => isset($v['path']) && isset($v['alias']))
+                        ->thenInvalid('You cannot define both "path" and "alias" for an icon set.')
+                    ->end()
                 ->end()
                 ->arrayNode('aliases')
-                    ->info('Icon aliases (alias => icon name).')
-                    ->example(['dots' => 'clarity:ellipsis-horizontal-line'])
+                    ->info('Icon aliases (map of alias => full name).')
+                    ->example([
+                        'dots' => 'clarity:ellipsis-horizontal-line',
+                        'privacy' => 'bi:cookie',
+                    ])
                     ->normalizeKeys(false)
                     ->scalarPrototype()
                         ->cannotBeEmpty()
                     ->end()
                 ->end()
                 ->arrayNode('iconify')
-                    ->info('Configuration for the "on demand" icons powered by Iconify.design.')
+                    ->info('Configuration for the remote icon service.')
                     ->{interface_exists(HttpClientInterface::class) ? 'canBeDisabled' : 'canBeEnabled'}()
                     ->children()
                         ->booleanNode('on_demand')
-                            ->info('Whether to use the "on demand" icons powered by Iconify.design.')
+                            ->info('Whether to download icons "on demand".')
                             ->defaultTrue()
                         ->end()
                         ->scalarNode('endpoint')
-                            ->info('The endpoint for the Iconify API.')
+                            ->info('The endpoint for the Iconify icons API.')
                             ->defaultValue(Iconify::API_ENDPOINT)
                             ->cannotBeEmpty()
                         ->end()
                     ->end()
                 ->end()
                 ->booleanNode('ignore_not_found')
-                    ->info('Ignore error when an icon is not found.')
+                    ->info("Ignore error when an icon is not found.\nSet to 'true' to fail silently.")
                     ->defaultFalse()
                 ->end()
             ->end()
@@ -94,9 +128,25 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
             $loader->load('asset_mapper.php');
         }
 
+        $iconSetAliases = [];
+        $iconSetAttributes = [];
+        $iconSetPaths = [];
+        foreach ($mergedConfig['icon_sets'] as $prefix => $config) {
+            if (isset($config['icon_attributes'])) {
+                $iconSetAttributes[$prefix] = $config['icon_attributes'];
+            }
+            if (isset($config['alias'])) {
+                $iconSetAliases[$prefix] = $config['alias'];
+            }
+            if (isset($config['path'])) {
+                $iconSetPaths[$prefix] = $config['path'];
+            }
+        }
+
         $container->getDefinition('.ux_icons.local_svg_icon_registry')
             ->setArguments([
                 $mergedConfig['icon_dir'],
+                $iconSetPaths,
             ])
         ;
 
@@ -107,6 +157,7 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
         $container->getDefinition('.ux_icons.icon_renderer')
             ->setArgument(1, $mergedConfig['default_icon_attributes'])
             ->setArgument(2, $mergedConfig['aliases'])
+            ->setArgument(3, $iconSetAttributes)
         ;
 
         $container->getDefinition('.ux_icons.twig_icon_runtime')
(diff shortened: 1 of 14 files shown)


src/Icons/src/IconRenderer.php
```diff
@@ -36,11 +42,19 @@ public function __construct(
...
     public function renderIcon(string $name, array $attributes = []): string
     {
-        $name = $this->iconAliases[$name] ?? $name;
+        $iconName = $this->iconAliases[$name] ?? $name;
+
+        $icon = $this->registry->get($iconName)
+            ->withAttributes($this->defaultIconAttributes);
+
+        if (str_contains($name, ':')) {
+            $prefix = substr($name, 0, strpos($name, ':'));
```
Kocal: We can remove one function call:
```suggestion
        if (false !== $colonPosition = strpos($name, ':')) {
            $prefix = substr($name, 0, $colonPosition);
```

src/Icons/src/IconRenderer.php
```diff
@@ -36,11 +42,19 @@ public function __construct(
...
+            if (isset($this->iconSetsAttributes[$prefix])) {
+                $icon = $icon->withAttributes($this->iconSetsAttributes[$prefix]);
+            }
```
Kocal: I think it can simplified to: 
```suggestion
            $attributes = ($this->iconSetsAttributes[$prefix] ?? []) + $attributes;
```

Kocal (approved):

src/Icons/src/IconRenderer.php
smnandre: I don't know how it's done internally indeed, but the "hot path" is "there is no $prefix key in $iconsSetAttributes[].. does this change something ?

src/Icons/src/IconRenderer.php
smnandre: Thx

src/Icons/src/IconRenderer.php
Kocal: With my suggestion we can get rid of 2 functions calls and 1 access to `$iconSetsAttributes` property, and 1 access to `$this->iconSetsAttributes` array.

If you compare opcode (see https://3v4l.org/TAQmE/vld vs https://3v4l.org/5QdBE/vld), we have 4 less operations.

smnandre: No of course, i was only comparing isset and [] ?? []

I'll update the code (but i think i'll do it on "default + iconSet"  instead of "iconSet + attribute"

smnandre: This should be better now

src/Icons/src/IconRenderer.php
```diff
@@ -36,11 +42,28 @@ public function __construct(
...
+        );
 
-        $icon = $this->registry->get($name)
-            ->withAttributes($this->defaultIconAttributes)
-            ->withAttributes($attributes);
+        // $icon = $icon->withAttributes([
+        //     ...$attributes,
+        //     ...($iconSetAttributes ?? []),
+        //     ...$this->defaultIconAttributes,
+        // ]);
```
Kocal: To remove

src/Icons/src/IconRenderer.php
```diff
@@ -36,11 +42,28 @@ public function __construct(
...
+            $prefix = substr($name, 0, $colonPosition);
+            $iconSetAttributes = $this->iconSetsAttributes[$prefix] ?? [];
+        }
+
+        $icon = $icon->withAttributes(
+            array_replace(
+                $this->defaultIconAttributes,
+                $iconSetAttributes ?? [],
+                $attributes,
+            )
```
Kocal: Not possible to use `+` instead?

smnandre: The order must be kept so i dont think so..  :|

Kocal: Ah? the order matters?

smnandre: Well i just spent an hour, persuaded there were some case it differed, due to the "null"  / "false" .....

..... and no surprise: i was wrong 😮‍💨 

i'll update tomorrow

smnandre: Done, and with this much better tests !  (in the end i chose a ... instead of a +  :)

smnandre: I'll let you check and LGTM :)

Kocal: > Done, and with this much better tests ! (in the end i chose a ... instead of a + :)

Yeah, I find it more readable

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -42,32 +42,66 @@ public function getConfigTreeBuilder(): TreeBuilder
...
                     ->{interface_exists(HttpClientInterface::class) ? 'canBeDisabled' : 'canBeEnabled'}()
                     ->children()
                         ->booleanNode('on_demand')
-                            ->info('Whether to use the "on demand" icons powered by Iconify.design.')
+                            ->info('Whether to download icons "on demand".')
                             ->defaultTrue()
                         ->end()
                         ->scalarNode('endpoint')
-                            ->info('The endpoint for the Iconify API.')
+                            ->info('The endpoint for the iconify icons API.')
```
Kocal:
```suggestion
                            ->info('The endpoint for the Iconify icons API.')
```

src/Icons/src/DependencyInjection/UXIconsExtension.php
```diff
@@ -42,32 +42,66 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                        'privacy' => 'bi:cookie',
+                    ])
                     ->normalizeKeys(false)
                     ->scalarPrototype()
                         ->cannotBeEmpty()
                     ->end()
                 ->end()
                 ->arrayNode('iconify')
-                    ->info('Configuration for the "on demand" icons powered by Iconify.design.')
+                    ->info('Configuration for the remote icon service.')
```
Kocal, line 89: Why not keeping reference to Iconify?

src/Icons/src/IconRenderer.php
```diff
@@ -36,11 +42,16 @@ public function __construct(
...
     {
-        $name = $this->iconAliases[$name] ?? $name;
+        $iconName = $this->iconAliases[$name] ?? $name;
 
-        $icon = $this->registry->get($name)
-            ->withAttributes($this->defaultIconAttributes)
-            ->withAttributes($attributes);
+        $icon = $this->registry->get($iconName);
+
+        if (0 < (int) $pos = strpos($name, ':')) {
```
Kocal, line 49:
```suggestion
        if (false !== $pos = strpos($name, ':')) {
```

src/Icons/src/IconRenderer.php
```diff
@@ -36,11 +42,16 @@ public function __construct(
...
+            $setAttributes = $this->iconSetsAttributes[substr($name, 0, $pos)] ?? [];
+        } elseif ($iconName !== $name && 0 < (int) $pos = strpos($iconName, ':')) {
```
Kocal, line 51:
```suggestion
        } elseif ($iconName !== $name && false !== $pos = strpos($iconName, ':')) {
```

Kocal (approved):

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: Thanks i made changes, reverted, made other ... and in the end lost my way 😅

src/Icons/src/IconRenderer.php
smnandre: No, this avoid me a check if the passed icon name start with a ":" ... and thus ensure the prefix is not empty ;)

src/Icons/src/DependencyInjection/UXIconsExtension.php
smnandre: The key is iconify, and the name and website are given just after ;) 

But as we start to more and more use the term "remote icon" and "local icon" i wanted to add it here

src/Icons/src/IconRenderer.php
smnandre: Same thing :)

javiereguiluz (approved):

javiereguiluz:
Merged! Thanks a lot Simon

seb-jean:
I tried but I can't get the expected result.
How to differentiate the attributes of these 4 subdirectories:

- assets/icons/heroicons/24/outline
- assets/icons/heroicons/24/solid
- assets/icons/heroicons/20/solid
- assets/icons/heroicons/16/solid

I tried the following code:

```yaml
ux_icons:
    icon_sets:
        heroicons:20:solid:
            icon_attributes:
                fill: 'currentColor'
                aria-hidden: true
                data-slot: 'icon'
```

smnandre:
So i'm pretty sure this would not work, but just to signal "heroicons:20:solid" should be enquoted in yaml ;)

smnandre:
Then i'm pretty sure heroicons:20:solid is not a icon set prefix (at least not an remote one), right ?

seb-jean:
Yes, icon set prefix is `heroicons`.

Capture d'écran 2024-10-06 151003

seb-jean:
Could this work with subdirectories then? :)

seb-jean:
I think it's because I put `:` in the name and suddenly it's blocked.

seb-jean:
I did it like this and it works:

```yaml
ux_icons:
    default_icon_attributes: []
    icon_sets:
        heroicons-20-solid:
            path: '%kernel.project_dir%/assets/icons/heroicons/20/solid'
            icon_attributes:
                fill: 'currentColor'
                data-slot: 'icon'
```

```twig
<twig:ux:icon name="heroicons-20-solid:cake" class="h-5 w-5 text-gray-400" />
```

smnandre:
Nice! I was not 100% sure it would :))

PR #2159, opened 2024-09-12
[StimulusBundle] Normalize Stimulus controller name in event name

### LINKED ISSUE #2151: [StimulusBundle] `StimulusAttributes::addAction()` method miss encoding of the controller name in the `$eventName` parameter
Executing of the following code
```php
$attributes->addAction('@foo/controller_1', 'someAction', '@bar/controller_2:someEvent');
```

will produce `data-action` attribute
```html
<div ... data-action="@bar/controller-2:someEvent->foo--controller-1#someAction" ...>
```

The controller name of event is not normalized.

Now, in order for the `data-action` attribute to be generated correctly, need to write the already normalized name of the controller.

```php
$attributes->addAction('@foo/controller_1', 'someAction', 'bar--controller-2:someEvent');
```

This behavior also happens for Twig.

7-zete-7:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2151
| License       | MIT

Adds the ability to use non-normalized Stimulus controller names when describing actions on events from other Stimulus controllers.

### DIFF
--- a/src/StimulusBundle/src/Dto/StimulusAttributes.php
+++ b/src/StimulusBundle/src/Dto/StimulusAttributes.php
@@ -75,7 +75,7 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
         $this->actions[] = [
             'controllerName' => $controllerName,
             'actionName' => $actionName,
-            'eventName' => $eventName,
+            'eventName' => null !== $eventName ? $this->normalizeEventName($eventName) : null,
         ];
 
         foreach ($parameters as $name => $value) {
@@ -216,6 +216,14 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
     private function normalizeControllerName(string $controllerName): string
     {
         return preg_replace('/^@/', '', str_replace('_', '-', str_replace('/', '--', $controllerName)));
+    }
+
+    /**
+     * @see https://stimulus.hotwired.dev/reference/actions
+     */
+    private function normalizeEventName(string $eventName): string
+    {
+        return preg_replace_callback('/^.+(?=:)/', fn (array $matches): string => $this->normalizeControllerName($matches[0]), $eventName);
     }
 
     /**
--- a/src/StimulusBundle/CHANGELOG.md
+++ b/src/StimulusBundle/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.20.1
+
+-   Normalize Stimulus controller name in event name
+
 ## 2.14.2
 
 -   Fix bug with finding UX Packages with non-standard project structure
--- a/src/StimulusBundle/tests/Dto/StimulusAttributesTest.php
+++ b/src/StimulusBundle/tests/Dto/StimulusAttributesTest.php
@@ -148,4 +148,91 @@ final class StimulusAttributesTest extends TestCase
         $this->assertSame('foo="bar baz"', (string) $this->stimulusAttributes);
         $this->assertSame(['foo' => 'bar baz'], $this->stimulusAttributes->toArray());
     }
+
+    /**
+     * @dataProvider provideAddComplexActionData
+     */
+    public function testAddComplexAction(string $controllerName, string $actionName, ?string $eventName, string $expectedAction): void
+    {
+        $this->stimulusAttributes->addAction($controllerName, $actionName, $eventName);
+        $attributesHtml = (string) $this->stimulusAttributes;
+        self::assertSame(\sprintf('data-action="%s"', $expectedAction), $attributesHtml);
+    }
+
+    /**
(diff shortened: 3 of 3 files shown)


src/StimulusBundle/tests/Dto/StimulusActionAttributeTest.php
```diff
@@ -0,0 +1,118 @@
...
+            'expectedAction' => 'foo--controller-1&#x3A;baz->bar--controller-2#log',
+        ];
+        yield 'foo--controller-1:baz@document->bar--controller-2#log:capture' => [
+            'controllerName' => '@bar/controller_2',
+            'actionName' => 'log:capture',
+            'eventName' => '@foo/controller_1:baz@document',
+            'expectedAction' => 'foo--controller-1&#x3A;baz&#x40;document->bar--controller-2#log&#x3A;capture',
+        ];
+
+        // TODO: Add more datasets.
```
7-zete-7: It would be great if someone could suggest datasets that break the current implementation.

src/StimulusBundle/tests/Dto/StimulusActionAttributeTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    protected function setUp(): void
+    {
+        $this->stimulusAttributes = new StimulusAttributes(new Environment(new ArrayLoader()));
+    }
+
+    /**
+     * @dataProvider provideAddActionData
+     */
+    #[DataProvider('provideAddActionData')]
```
smnandre: We did not start this compatibility migration.. i'd rather we do it at once :) 
```suggestion
```

src/StimulusBundle/tests/Dto/StimulusActionAttributeTest.php
smnandre: Do you think there is many of them ? 😅

smnandre (commented):
Thank you @7-zete-7 

Could you update the CHANGELOG ?

src/StimulusBundle/tests/Dto/StimulusActionAttributeTest.php
7-zete-7: Apparently this is enough 😁

7-zete-7:
> Could you update the CHANGELOG ?

Updated

7-zete-7:
No more unnecessary escaping. Great!

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -218,6 +218,20 @@ private function normalizeControllerName(string $controllerName): string
         return preg_replace('/^@/', '', str_replace('_', '-', str_replace('/', '--', $controllerName)));
     }
 
+    /**
+     * Normalize a Stimulus controller's action event name.
+     *
```
Kocal:
```suggestion
```

The method name is enough to explain what it does

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -218,6 +218,20 @@ private function normalizeControllerName(string $controllerName): string
...
+     * @see https://stimulus.hotwired.dev/reference/actions
+     */
+    private function normalizeEventName(?string $eventName): ?string
+    {
+        if (null === $eventName) {
+            return null;
+        }
+
```
Kocal: I suggest to not support nullable string here, but only non-nullable string.
```suggestion
    private function normalizeEventName(string $eventName): string
    {
```

The null check should be done before calling this method.

src/StimulusBundle/tests/Dto/StimulusActionAttributeTest.php
Kocal, line 1: Please move those tests inside `src/StimulusBundle/tests/Dto/StimulusAttributesTest.php` file 🙏🏻

smnandre: Has a change been made since your comment? Cause I dont see the différence now

Kocal: Nope

src/StimulusBundle/src/Dto/StimulusAttributes.php
7-zete-7: Moved `null` value check out of the method.

src/StimulusBundle/tests/Dto/StimulusActionAttributeTest.php
7-zete-7: Moved the test to the common class.

Kocal:
Thank you @7-zete-7.

PR #2160, opened 2024-09-12
[Vue] The idea of ​​making `RenderController` are reactive

7-zete-7:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

The idea is to create a bridge between the reactivity of Stimulus values ​​and the reactivity of Vue.js properties.

This would allow to influence the properties of a Vue.js component from the outside without implementing an additional Stimulus controller for this.

Changes to the properties passed to the Vue.js component inside the component would also reflect their change in the Stimulus values.

I have already implemented something similar in some projects. It would be very convenient if this worked out of the box and was useful to someone else.

### DIFF
--- a/src/Vue/doc/index.rst
+++ b/src/Vue/doc/index.rst
@@ -169,6 +169,70 @@ used for all the Vue routes:
 
 .. _using-with-asset-mapper:
 
+Keep properties are reactive
+----------------------------
+
+All Vue component properties are reactive up to the Stimulus controller `props` value.
+
+Value changes are two-way:
+
+* Any changes of the Stimulus component `props` value will
+    reactively pass new values to the Vue component without re-creating it,
+    as would be the case when passing props between Vue components.
+
+* Any changes to the properties in the Vue component,
+    if those properties are or replicate the behavior of models,
+    will change the Stimulus controller `props` value.
+
+.. code-block:: javascript
+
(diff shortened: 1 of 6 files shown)


smnandre (commented):
I'm not the target user here, so i'll wait opinion of the others :) 

Could you maybe add a couple of tests ... and maybe add this to the documentation with a short description + example ? 

Thank you for sharing and contributing :)

7-zete-7:
Added some tests to make sure it really works.

Fixed the non-working mechanism.

src/Vue/doc/index.rst
```diff
@@ -169,6 +169,70 @@ used for all the Vue routes:
...
+    document.getElementById('likes-component').dataset.vuePropsValue = JSON.stringify({
+        likes: newLikesCount,
+        alreadyLike: isAlreadyLike,
+    });
+
+.. code-block:: javascript
+
+    // get likes component actual props
+    const { likes, alreadyLike } = JSON.parse(document.getElementById('likes-component').dataset.vuePropsValue);
+
```
7-zete-7, line 235: I tried to make examples untied from other components. It is more logical, of course, to use Stimulus controllers to perform actions with parameters.

Kocal:

Thanks!

Status: Needs Work

PR #2161, opened 2024-09-13
 [TwigComponent] Fix wrong int type hint on integer concatination which results in string

DZunke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

I just played around with the Twig Components and stumbled across an exception from the `ComponentTokenParser`. It returns a string within the method `generateEmbeddedTemplateIndex` but the type hint wants to get an integer. I tried to find out  why thid worked for someone but from my PoV there should be a string return type because the concatination of two integer values always leads to a string as far as i know.

So i have here changed the type hint without further explanaition of my use case because it is, for me, more of a general type hint error.

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentTokenParser.php
+++ b/src/TwigComponent/src/Twig/ComponentTokenParser.php
@@ -111,7 +111,7 @@ final class ComponentTokenParser extends AbstractTokenParser
         return [$variables, $only];
     }
 
-    private function generateEmbeddedTemplateIndex(string $file, int $line): int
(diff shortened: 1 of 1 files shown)


smnandre:
This does in fact return an int ... and an int is needed by the ComponentNode later. 

```php
<?php

function foobar(): int
{
    $bar = 0;
    return crc32('aa').++$bar;
}

var_dump(foobar());

# int(1264910951)
```
https://3v4l.org/tLgK5#v8.3.11

PR #2166, opened 2024-09-15
[Site] Clean templates & css

smnandre:
* Clean demo templates
* Remove unused helpers
* Remove unused utilities

### DIFF
--- a/ux.symfony.com/assets/styles/app.scss
+++ b/ux.symfony.com/assets/styles/app.scss
@@ -11,11 +11,53 @@
 @import "../../vendor/twbs/bootstrap/scss/variables";
 @import "../../vendor/twbs/bootstrap/scss/variables-dark";
 
+// Remove Bootstrap variables/overrides
+$theme-colors: map-remove($theme-colors, "info", "light");
+
 // Boostrap remainder of required parts
 @import "../../vendor/twbs/bootstrap/scss/maps";
 @import "../../vendor/twbs/bootstrap/scss/mixins";
 @import "../../vendor/twbs/bootstrap/scss/utilities";
 @import "../../vendor/twbs/bootstrap/scss/root";
+
+// Remove multiple utilities with a comma-separated list
+// Margin
+$utilities: map-remove(
+    $utilities,
+    // Background
+    "bg-opacity",
+    "subtle-background-color",
+
+    // Negative margin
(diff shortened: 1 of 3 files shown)


smnandre:
Is it possible to disable fabbot for the website ? cc @kbond @Kocal

Kocal:
I don't know how to, I believe you should ask to Fabien

ux.symfony.com/assets/styles/app.scss
```diff
@@ -11,12 +11,54 @@
...
+
 // Boostrap remainder of required parts
 @import "../../vendor/twbs/bootstrap/scss/maps";
 @import "../../vendor/twbs/bootstrap/scss/mixins";
 @import "../../vendor/twbs/bootstrap/scss/utilities";
 @import "../../vendor/twbs/bootstrap/scss/root";
 
+// Remove multiple utilities with a comma-separated list
+// Margin
+$utilities: map-remove(
```
javiereguiluz, line 25: I'm not sure about this kind of micro-optimizations. Reasons:

* The full Bootstrap CSS contents take only 30 KBs when compressed with br or zstd, so it's not a problem for most folks
* If we remove common utilities like `border`, they won't work when we try to use them in the future and debugging why can be frustrating.

smnandre: The bootstrap CSS files is currently 90% useless _because_ of all the utilities we do not use/need  ... and the only failure in the PageSpeed results.

pagespeed

As i explained some time ago, i'll rework the entire HTML of the website, the final goal is to be almost "classless", and thus facilitate content creation/aggregation (mainly for demos but also cookbook, articles, docs, etc..).

I put this work in "pause" after the "dark/light theme" debate .. but i need to move on now, and trimming down the dependencies to the only things we really use is a method i'm use to work with. It's then much easier to move/recode things :) 

So i understand your vision and would agree if this rework project was not in progress :)

Kocal (approved):

ux.symfony.com/assets/styles/app.scss
javiereguiluz: OK then. Thanks for the additional insights! And thanks as always for your work here.

javiereguiluz (approved):

PR #2167, opened 2024-09-15
[TwigComponent] Optimize the ComponentLoggerListener

smnandre:
* Clean unused code
* Reduce memory usage
* Optimize class/templates lookup

### DIFF
--- a/src/TwigComponent/src/EventListener/TwigComponentLoggerListener.php
+++ b/src/TwigComponent/src/EventListener/TwigComponentLoggerListener.php
@@ -12,73 +12,36 @@
 namespace Symfony\UX\TwigComponent\EventListener;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\Stopwatch\Stopwatch;
 use Symfony\Contracts\Service\ResetInterface;
-use Symfony\UX\TwigComponent\Event\PostMountEvent;
 use Symfony\UX\TwigComponent\Event\PostRenderEvent;
-use Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent;
-use Symfony\UX\TwigComponent\Event\PreMountEvent;
 use Symfony\UX\TwigComponent\Event\PreRenderEvent;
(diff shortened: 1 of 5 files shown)


src/TwigComponent/src/EventListener/TwigComponentLoggerListener.php
```diff
@@ -12,90 +12,55 @@
...
-use Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent;
-use Symfony\UX\TwigComponent\Event\PreMountEvent;
 use Symfony\UX\TwigComponent\Event\PreRenderEvent;
 
 /**
  * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
  */
 class TwigComponentLoggerListener implements EventSubscriberInterface, ResetInterface
```
kbond:
```suggestion
final class TwigComponentLoggerListener implements EventSubscriberInterface, ResetInterface
```

kbond (approved):
This is great!

javiereguiluz:
Nice! Maybe we can also remove the `symfony/stopwatch` dev dependency here?

https://github.com/symfony/ux/blob/8dd706da2c1ada2174c920c246daa6765b176cef/src/TwigComponent/composer.json#L43

Kocal (approved):

smnandre:
> Nice! Maybe we can also remove the `symfony/stopwatch` dev dependency here?
> 
> https://github.com/symfony/ux/blob/8dd706da2c1ada2174c920c246daa6765b176cef/src/TwigComponent/composer.json#L43

Absolutely, thank you for catching this!

Kocal:
Do... you think we can have some Blackfire profiles to show the performance improvments? 🙏🏻

smnandre:
> Do... you think we can have some Blackfire profiles to show the performance improvments? 🙏🏻

I did not have time yet to configure it 😶‍🌫️  

But i will in the next few days (already started a dedicated repo to measure performance of twig/live components)

PR #2168, opened 2024-09-15
[TwigComponent] Use a RuntimeExtension

smnandre:
* Hide implementation & start decoupling the template from the event system
* Optimize (a bit) the ComponentNode

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentRuntime.php
+++ b/src/TwigComponent/src/Twig/ComponentRuntime.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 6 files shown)


src/TwigComponent/src/Twig/ComponentRuntime.php
```diff
@@ -0,0 +1,60 @@
...
+use Symfony\UX\TwigComponent\Event\PreRenderEvent;
+use Twig\Extension\RuntimeExtensionInterface;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
+ */
+final class ComponentRuntime implements RuntimeExtensionInterface
```
kbond: (This interface isn't required - it's only for autoconfiguration IIRC)

smnandre: Did not know, thanks

kbond (approved):

Kocal (approved):

javiereguiluz (approved):
Nice. Thank you!

PR #2178, opened 2024-09-18
[Map] Improve backend performance when rendering HTML "view" attribute

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

My intuition was right, I knew there was some back optimization to be done on Map :) 

For the context, I've created a Map with 1000 Marker and InfoWindow (it represents the first 1000 cities from France in alphabetical order, but it can also be a store-locator, etc...)
Capture d’écran 2024-09-18 à 12 22 40

The code looks like:
```php
$map = new Map();
$map->center(new Point(48.8566, 2.3522));
$map->zoom(6);

$cities = require_once __DIR__.'/../../config/cities.php';

foreach ($cities as $i => $city) {
    $map->addMarker(
        new Marker(
            position: new Point($city['latitude'], $city['longitude']),
            title: $city['label'],
            infoWindow: new InfoWindow(
                headerContent: '<b>'.$city['label'].'</b>',
                content: 'Welcome in '.$city['label'],
                opened: $i === 300,
            ),
        )
    );
}
```

## Before

Before this PR, [when profiling the page with Blackfire](https://blackfire.io/profiles/a2174954-445c-4007-8364-8277913d27bf/graph), we can see that `StimulusAttributes#toString()` takes ~62% of the response time (~222ms), because of the Twig escaping code (`preg_replace_callback` and `ord()`):  
Capture d’écran 2024-09-18 à 12 26 37

This is long because our `view` attribute can be very big, it contains a lot of data. But, since it's an array with a known structure, etc..., maybe we can... not use Twig escaping strategy?

> **Note**
> I will open an issue on Twig, maybe it already exists, but it may be interesting to see if we can improve things at the root.

**EDIT:** opened https://github.com/twigphp/Twig/issues/4322

## Now

I've removed `view' => $map->toArray()` from the `StimulusAttributes`, and added `'data-symfony--ux-'.$this->getName().'-map--map-view-value="'.htmlentities(json_encode($map->toArray(), flags: JSON_THROW_ON_ERROR)).'"'` at the final rendering place. 

When [profiling the page with Blackfire](https://blackfire.io/profiles/4d5bdd1c-b528-4fae-8341-066a7eb8239b/graph), I can see an [improvment of ~63% on the response time 🚀 ](https://blackfire.io/profiles/compare/efbd14c7-9510-46aa-a661-78104bb29861/graph):
image

I'm not 100% sure, but I believe `htmlentities` and `json_encode` are enough for our usecase? I mean, it still works, and HTML in InfoWindow still works too:
Capture d’écran 2024-09-18 à 12 33 28


Kocal:
Closing in favor of a new PR :)

PR #2180, opened 2024-09-18
[StimulusBundle] Improve `StimulusAttributes` rendering performances by switching to `html` escaping strategy

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

A better alternative to #2178, by changing the escaping strategy for HTML attributes value from `html_attr` to `html`, as indicated by @stof in https://github.com/twigphp/Twig/issues/4322#issuecomment-2358230379.

On my use case (a `Map` with 1000 `Marker` and `InfoWindow`), [I have a performance gain of ~68%](https://blackfire.io/profiles/compare/56f4d7d2-ee56-487f-8a78-420f4037165f/graph):
image

This PR should also improve performances of our packages using the `StimulusAttributes` DTO, like Chart.js, LiveComponent, ...

---

> [!IMPORTANT]
> The initial PR changed a bit, the default rendering strategy does not change, but instead we introduce a new configuration to use the new (and optimized) rendering strategy, see  https://github.com/symfony/ux/pull/2180#issuecomment-2370574032.

### DIFF
--- a/src/StimulusBundle/src/Dto/StimulusAttributes.php
+++ b/src/StimulusBundle/src/Dto/StimulusAttributes.php
@@ -107,54 +107,36 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
 
     public function __toString(): string
     {
-        $controllers = array_map(function (string $controllerName): string {
-            return $this->escapeAsHtmlAttr($controllerName);
-        }, $this->controllers);
-
-        // done separately so we can escape, but avoid escaping ->
-        $actions = array_map(function (array $actionData): string {
-            $controllerName = $this->escapeAsHtmlAttr($actionData['controllerName']);
-            $actionName = $this->escapeAsHtmlAttr($actionData['actionName']);
-            $eventName = $actionData['eventName'];
-
-            $action = $controllerName.'#'.$actionName;
-            if (null !== $eventName) {
-                $action = $this->escapeAsHtmlAttr($eventName).'->'.$action;
-            }
-
-            return $action;
-        }, $this->actions);
-
-        $targets = [];
-        foreach ($this->targets as $key => $targetNamesString) {
-            $targetNames = explode(' ', $targetNamesString);
-            $targets[$key] = implode(' ', array_map(function (string $targetName): string {
-                return $this->escapeAsHtmlAttr($targetName);
-            }, $targetNames));
-        }
-
         $attributes = [];
 
-        if ($controllers) {
-            $attributes[] = \sprintf('data-controller="%s"', implode(' ', $controllers));
+        if ($this->controllers) {
+            $attributes[] = 'data-controller="'.$this->escape(implode(' ', $this->controllers)).'"';
         }
 
-        if ($actions) {
-            $attributes[] = \sprintf('data-action="%s"', implode(' ', $actions));
+        if ($this->actions) {
+            $actions = [];
+            foreach ($this->actions as ['controllerName' => $controllerName, 'actionName' => $actionName, 'eventName' => $eventName]) {
+                $action = $this->escape($controllerName.'#'.$actionName);
+                if (null !== $eventName) {
+                    // done separately so we can escape, but avoid escaping ->
+                    $action = $this->escape($eventName).'->'.$action;
+                }
+
+                $actions[] = $action;
+            }
+
+            $attributes[] = 'data-action="'.implode(' ', $actions).'"';
         }
 
-        if ($targets) {
-            $attributes[] = implode(' ', array_map(function (string $key, string $value): string {
-                return \sprintf('%s="%s"', $key, $value);
-            }, array_keys($targets), $targets));
+        foreach ($this->targets as $k => $v) {
+            $attributes[] = $this->escape($k, 'html_attr').'="'.$this->escape($v).'"';
         }
 
-        return rtrim(implode(' ', [
-            ...$attributes,
-            ...array_map(function (string $attribute, string $value): string {
-                return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
-            }, array_keys($this->attributes), $this->attributes),
-        ]));
+        foreach ($this->attributes as $k => $v) {
+            $attributes[] = $this->escape($k, 'html_attr').'="'.$this->escape($v).'"';
+        }
+
+        return implode(' ', $attributes);
     }
 
     public function toArray(): array
@@ -193,7 +175,7 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
     {
         $escaped = [];
         foreach ($this->toArray() as $key => $value) {
-            $escaped[$key] = $this->escapeAsHtmlAttr($value);
+            $escaped[$key] = $this->escape($value);
         }
 
         return $escaped;
@@ -212,18 +194,18 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
         return (string) $value;
     }
 
-    private function escapeAsHtmlAttr(mixed $value): string
+    private function escape(mixed $value, string $strategy = 'html'): string
     {
         if (class_exists(EscaperRuntime::class)) {
-            return $this->env->getRuntime(EscaperRuntime::class)->escape($value, 'html_attr');
+            return $this->env->getRuntime(EscaperRuntime::class)->escape($value, $strategy);
         }
 
         if (method_exists(EscaperExtension::class, 'escape')) {
-            return EscaperExtension::escape($this->env, $value, 'html_attr');
+            return EscaperExtension::escape($this->env, $value, $strategy);
         }
 
         // since twig/twig 3.9.0: Using the internal "twig_escape_filter" function is deprecated.
-        return (string) twig_escape_filter($this->env, $value, 'html_attr');
+        return (string) twig_escape_filter($this->env, $value, $strategy);
     }
 
     /**
--- a/src/Chartjs/tests/Twig/ChartExtensionTest.php
+++ b/src/Chartjs/tests/Twig/ChartExtensionTest.php
@@ -56,7 +56,7 @@ class ChartExtensionTest extends TestCase
         );
 
         $this->assertSame(
-            '<canvas data-controller="mycontroller symfony--ux-chartjs--chart" data-symfony--ux-chartjs--chart-view-value="&#x7B;&quot;type&quot;&#x3A;&quot;line&quot;,&quot;data&quot;&#x3A;&#x7B;&quot;labels&quot;&#x3A;&#x5B;&quot;January&quot;,&quot;February&quot;,&quot;March&quot;,&quot;April&quot;,&quot;May&quot;,&quot;June&quot;,&quot;July&quot;&#x5D;,&quot;datasets&quot;&#x3A;&#x5B;&#x7B;&quot;label&quot;&#x3A;&quot;My&#x20;First&#x20;dataset&quot;,&quot;backgroundColor&quot;&#x3A;&quot;rgb&#x28;255,&#x20;99,&#x20;132&#x29;&quot;,&quot;borderColor&quot;&#x3A;&quot;rgb&#x28;255,&#x20;99,&#x20;132&#x29;&quot;,&quot;data&quot;&#x3A;&#x5B;0,10,5,2,20,30,45&#x5D;&#x7D;&#x5D;&#x7D;,&quot;options&quot;&#x3A;&#x7B;&quot;showLines&quot;&#x3A;false&#x7D;&#x7D;" class="myclass"></canvas>',
+            '<canvas data-controller="mycontroller symfony--ux-chartjs--chart" data-symfony--ux-chartjs--chart-view-value="{&quot;type&quot;:&quot;line&quot;,&quot;data&quot;:{&quot;labels&quot;:[&quot;January&quot;,&quot;February&quot;,&quot;March&quot;,&quot;April&quot;,&quot;May&quot;,&quot;June&quot;,&quot;July&quot;],&quot;datasets&quot;:[{&quot;label&quot;:&quot;My First dataset&quot;,&quot;backgroundColor&quot;:&quot;rgb(255, 99, 132)&quot;,&quot;borderColor&quot;:&quot;rgb(255, 99, 132)&quot;,&quot;data&quot;:[0,10,5,2,20,30,45]}]},&quot;options&quot;:{&quot;showLines&quot;:false}}" class="myclass"></canvas>',
             $rendered
         );
     }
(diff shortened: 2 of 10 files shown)


src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -108,18 +108,18 @@ public function addAttribute(string $name, string $value): void
...
         // done separately so we can escape, but avoid escaping ->
         $actions = array_map(function (array $actionData): string {
-            $controllerName = $this->escapeAsHtmlAttr($actionData['controllerName']);
-            $actionName = $this->escapeAsHtmlAttr($actionData['actionName']);
+            $controllerName = $actionData['controllerName'];
+            $actionName = $actionData['actionName'];
             $eventName = $actionData['eventName'];
 
-            $action = $controllerName.'#'.$actionName;
+            $action = $this->escape($controllerName.'#'.$actionName, 'html');
```
Kocal: It keeps the same behaviour, but with **one** `$this->escape()` call, `#` is still rendered as `#`.

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -212,18 +212,22 @@ private function getFormattedValue(mixed $value): string
...
     {
-        if (class_exists(EscaperRuntime::class)) {
-            return $this->env->getRuntime(EscaperRuntime::class)->escape($value, 'html_attr');
-        }
-
-        if (method_exists(EscaperExtension::class, 'escape')) {
-            return EscaperExtension::escape($this->env, $value, 'html_attr');
+        static $escaper;
+
+        if (!$escaper) {
```
Kocal: Quick win... I think? I don't wanted to re-execute `N` times the escaping logic, when calling `N` times `$this->escape()`.

But I believe `class_exists` is fast enough to be re-executed multiple times. 
I'm gonna revert it

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -129,7 +129,7 @@ public function __toString(): string
         foreach ($this->targets as $key => $targetNamesString) {
             $targetNames = explode(' ', $targetNamesString);
             $targets[$key] = implode(' ', array_map(function (string $targetName): string {
-                return $this->escapeAsHtmlAttr($targetName);
+                return $this->escape($targetName, 'html');
```
stof: This could even be simplified as you don't need to escape each of the space-separated values separately (and this will allow avoiding looping first on `$this->targets` and then on `$targets` below.

Side note: to be safe, the handling of targets is missing the escaping of the key

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -152,7 +152,7 @@ public function __toString(): string
         return rtrim(implode(' ', [
             ...$attributes,
             ...array_map(function (string $attribute, string $value): string {
-                return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
+                return $attribute.'="'.$this->escape($value, 'html').'"';
```
stof: this should probably escape `$attribute` here (and this one needs to use `html_attr`)

Kocal: I thought about it too, I'm adjusting tests and I will see for more improvments

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -212,18 +212,18 @@ private function getFormattedValue(mixed $value): string
         return (string) $value;
     }
 
-    private function escapeAsHtmlAttr(mixed $value): string
+    private function escape(mixed $value, string $strategy): string
```
smnandre: Is there any place we don't set 'html' ?

smnandre:
Do we still need to call the Runtime then ?  Would this not be sufficient ?

```
 return htmlspecialchars($string, \ENT_QUOTES | \ENT_SUBSTITUTE, $charset);
 ```

File: src/StimulusBundle/src/Dto/StimulusAttributes.php (hunk 5 above)
Comment (stof) [reply src/StimulusBundle/src/Dto/StimulusAttributes.php]: if you fix https://github.com/symfony/ux/pull/2180#discussion_r1765076980, there will be

File: src/StimulusBundle/src/Dto/StimulusAttributes.php (hunk 4 above)
Comment (Kocal) [reply src/StimulusBundle/src/Dto/StimulusAttributes.php]: Done

File: src/StimulusBundle/src/Dto/StimulusAttributes.php (hunk 5 above)
Comment (Kocal) [reply src/StimulusBundle/src/Dto/StimulusAttributes.php]: Yes!

File: src/StimulusBundle/src/Dto/StimulusAttributes.php (hunk 3 above)
Comment (Kocal) [reply src/StimulusBundle/src/Dto/StimulusAttributes.php]: Done

**Comment by Kocal** [comment]
> Do we still need to call the Runtime then ? Would this not be sufficient ?
> 
> ```
>  return htmlspecialchars($string, \ENT_QUOTES | \ENT_SUBSTITUTE, $charset);
> ```

If we assume UTF-8 is used everywhere, yes, but `|escape('html')` does more than that :/

**Comment by smnandre** [comment]
This bundle is the most used / downloaded. So we cannot release a version that will break tests (and maybe code) during a minor...

so we need to find a way out here... what about thinking a new object ? less "stimulus" and more "attributes"  ?

**Review by WebMamba — COMMENTED** [review]
Hey! I am agree with @smnandre here! It's a BC break for me. Maybe we can have a config that allow to switch from html_attr to html ?

**Comment by smnandre** [comment]
Sadly, it's difficult, as this class is directly used by other components / bundle.. and it's neither final or internal :|

**Comment by Kocal** [comment]
It's a BC for tests/assertions yes, but I don't really see how and where it can impact end users.

However, we can't stay ad vitam aeternam with `html_attr` escaping for HTML attribute values, as we all see it is not performant at all. 

Since `StimulusAttributes` is not final nor internal, maybe... we can configure the escaping format through a new method?

**Comment by kbond** [comment]
New bundle config? Deprecate not setting it to the new, desired value?

**Comment by smnandre** [comment]
> New bundle config ? 

We cannot here, this class may be instanciated directly in many projects, as it is not a service.

> Deprecate not setting it to the new, desired value?

This is a good idea. Probably we can add an argument either in the render / toString methods, or maybe more in the constructor..

.. and deprecate no passing it (stating in 3.x it will use the new mode) 

?

**Comment by smnandre** [comment]
Regarding performances, i'm very _very_ happy you found this bottleneck and have a clean method to improve things.

But as always, we cannot "force it" down, and it's been written everywhere Symfony  UX would follow Symfony BC promise.

To be clear, i'm _**not**_ saying we cannot use these optimisations in the ux components
* ux map is still experimental, and this is a very legit case of accepted change 
* we did not made any promise/engagement on the attributes generated by live component, and it could 100% be considered internal .. this also work for chartsjs, etc (but should we release a typed bundle version just for this ? i'm not entirely sure here..)

I'm concerned about two things here:
* the StimulusAttributes class in itself, that can be used elsewhere / tested / etc ... this would be a major BC break, as simple unit test cases would break
* the StimulusBundle twig functions (`stimulus_xxx`), that could be configured by the stimulus bundle but i'm not sure this is worth it

**Comment by kbond** [comment]
> This is a good idea. Probably we can add an argument either in the render / toString methods, or maybe more in the constructor..

> .. and deprecate no passing it (stating in 3.x it will use the new mode)

Constructor then, yes.

**Comment by Kocal** [comment]
I'm reverting to `html_attr` escaping strategy by default, then I will add whatever needed to re-use `html` for escaping html attribute values (without BC breaks)

**Comment by javiereguiluz** [comment]
Folks, I'm a bit lost about the mentioned BC break. Can you please share an example of a BC-break caused by changing this? Thanks!

**Comment by stof** [comment]
there is no BC break for users of the library. But there is a different escaping applied, which can make their *tests* fail if they assert the output of the rendering of `StimulusAttribute`.

Note however that by this argument, any bugfix is also forbidden.

**Comment by Kocal** [comment]
I've updated the PR and introduced a new config `stimulus.use_optimized_attributes_rendering` (default to `null`, which allows me to trigger a deprecation if not explicitly specified):
image
_Note: I'll gladly take any comments for my deprecation messages... 😅_

This configuration is injected to the `StimulusHelper` constructor, then injected to the `StimulusAttributes` constructor.

I've rollbacked tests added in my first commit as it was considered as breaking changes. Instead, I've configured `stimulus.use_optimized_attributes_rendering`/`$useOptimizedRendering` to `false` in order to get rid of deprecation notices (which make tests fails and CI red). I also adapted actual tests to test the _new optimized_ rendering, this way we can clearly see the difference between the actual and new rendering behaviour.

With all those changes, I took a new Blackfire profile on my page with a `Map` and 1000 `Marker`/`InfoWindow`, and did [a new comparaison](https://blackfire.io/profiles/compare/f71cdd8c-c741-47ec-a598-2318a4821b3e/graph):
image

File: src/StimulusBundle/src/Dto/StimulusAttributes.php [hunk 6]
```diff
@@ -22,14 +22,32 @@
...
 
     private array $controllers = [];
     private array $actions = [];
     private array $targets = [];
 
-    public function __construct(private Environment $env)
+    /**
+     * TODO: Remove the second argument and all related code in the next major version.
+     */
+    public function __construct(private Environment $env /* , bool $useOptimizedRendering = false */)
```
stof: no need to do that magic about new argument. PHP does not enforce having a consistent constructor signature in child classes, so adding a new optional constructor argument is not a BC break.

Kocal: Oh, you're right, thanks

src/TwigComponent/tests/Fixtures/Kernel.php
```diff
@@ -54,7 +54,7 @@ protected function configureContainer(ContainerConfigurator $c): void
         $c->extension('twig', [
             'default_path' => '%kernel.project_dir%/tests/Fixtures/templates',
         ]);
-
+
```
stof: this should be reverted

javiereguiluz:
@Kocal impressive performance improvement 🤯 thanks a lot 🙇 

I still think this is not a true BC break and we shouldn't complicate things with the new config option.

nicolas-grekas (commented):
I agree with Javier, no need for the option.
We might reconsider if we get enough reports but without, I wouldn't anticipate this aspect.

src/StimulusBundle/src/DependencyInjection/StimulusExtension.php
```diff
@@ -38,6 +38,12 @@ public function load(array $configs, ContainerBuilder $container): void
             ->replaceArgument(2, $config['controller_paths'])
             ->replaceArgument(3, $config['controllers_json']);
 
+        if (null === $config['use_optimized_attributes_rendering']) {
+            trigger_deprecation('symfony/stimulus-bundle', '2.20', 'The configuration option "use_optimized_attributes_rendering" is deprecated. It is now interpreted as "false", but it will be interpreted as "true" (and removed) in the next major version. Configure "use_optimized_attributes_rendering" to "true" to use the new and optimized rendering system for Stimulus Attributes.', __METHOD__);
```
stof: This deprecation message makes me think I need to *remove* the option to solve it as it says the option is deprecated.
What is deprecated is actually *not configuring it*. See https://github.com/symfony/symfony/blob/eb88eb3eb820208f1e8f9ba89b319811753a4d12/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L88 for the wording of such deprecation in Symfony.

stof (commented):
I'm still not convinced this should be considered a breaking change. The generated HTML string will be different, but this will not change anything for the output of parsing it.

Kocal:
I agree with all of you, I don't consider it a breaking change either. 
The behavior **won't change** in the browser, users won't be (badly) impacted by this modification. 

Let's revert the 2 last commit then, just in case. We will squash the PR at merging.

src/StimulusBundle/src/DependencyInjection/StimulusExtension.php
Kocal: You're right, thanks for the example!

Kocal:
The PR is now back at its original state, there is no `stimulus.use_optimized_attributes_rendering` config anymore.

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -212,18 +201,18 @@ private function getFormattedValue(mixed $value): string
         return (string) $value;
     }
 
-    private function escapeAsHtmlAttr(mixed $value): string
+    private function escape(mixed $value, string $strategy): string
```
nicolas-grekas:
```suggestion
    private function escape(mixed $value, string $strategy = 'html'): string
```

stof (approved):

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -107,54 +107,43 @@ public function addAttribute(string $name, string $value): void
...
-            $attributes[] = \sprintf('data-controller="%s"', implode(' ', $controllers));
+            $attributes[] = 'data-action="'.implode(' ', $actions).'"';
         }
 
-        if ($actions) {
-            $attributes[] = \sprintf('data-action="%s"', implode(' ', $actions));
+        if ($this->targets) {
+            $attributes[] = implode(' ', array_map(function (string $key, string $value): string {
+                return $this->escape($key, 'html_attr').'="'.$this->escape($value, 'html').'"';
+            }, array_keys($this->targets), $this->targets));
```
nicolas-grekas: foreach FTW, seriously :)
```suggestion
            foreach ($this->targets as $k => $v) {
                $attributes[] = $this->escape($k, 'html_attr').'="'.$this->escape($v).'"';
            }
```

src/StimulusBundle/src/Dto/StimulusAttributes.php
```diff
@@ -107,54 +107,43 @@ public function addAttribute(string $name, string $value): void
...
         }
 
-        if ($targets) {
-            $attributes[] = implode(' ', array_map(function (string $key, string $value): string {
-                return \sprintf('%s="%s"', $key, $value);
-            }, array_keys($targets), $targets));
+        if ($this->attributes) {
+            $attributes[] = implode(' ', array_map(function (string $attribute, string $value): string {
```
nicolas-grekas: foreach here also
so many array copies for nothing...

src/StimulusBundle/src/Dto/StimulusAttributes.php
Kocal: Done

src/StimulusBundle/src/Dto/StimulusAttributes.php
Kocal: Indeed, there was still some work to do, thank you :)

nicolas-grekas (approved):
thanks

javiereguiluz:
Thank you Hugo.

PR #2185, opened 2024-09-19
[Translator] Option to throw an exception instead of returning the message as is when the message is not found.

### LINKED ISSUE #2183: [Translator] Add an option to throw an exception instead of returning the message as is when the translation is not found.
For some translations, the `id` (the key in other words) is the same as the translations (e.g. the `Invalid credentials.` in the security domain):

```js
export const INVALID_CREDENTIALS = {"id":"Invalid credentials.","translations":{"security":{"fr":"Identifiants invalides.","en":"Invalid credentials.","de":"Fehlerhafte Zugangsdaten.","it":"Credenziali non valide."}}};
```

Therefore, there is no way in the frontend to distinguish a missing translation from a translation that have the same `id` (key) same as a translation because the only way I found in the frontend to detect if the translation is missing is checking if the result from the `trans` method is the same as the id. For example:

```js
const translationConstant = aServiceThat.findConstantById(messageId);

// I'm voluntarily trying in the default domain (`messages`) even though I know the message isn't there.
let translatedMessage = translator.trans(translator[translationConstant]);

if (translatedMessage === translator[translationConstant].id) {
    // Here how I distinguish a missing translation from a translation that have a same value as his id when no exception is thrown. If an exception was thrown, I could try something else (e.g. search in another domain)
}
```

To solve this problem, I think we could pass an additional to the `trans` method with a default value (`throwWhenNotFound = false`) and catch the exception later.

Crovitche-1623:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #2183
| License       | MIT

### DIFF
--- a/src/Translator/CHANGELOG.md
+++ b/src/Translator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.20.0
+
+-   Add `throwWhenNotFound` function to configure the behavior when a translation is not found.
+
 ## 2.19.0
 
 -   Add configuration to filter dumped translations by domain.
--- a/src/Translator/assets/src/translator.ts
+++ b/src/Translator/assets/src/translator.ts
@@ -38,6 +38,7 @@ import { format } from './formatters/formatter';
 
 let _locale: LocaleType | null = null;
 let _localeFallbacks: Record<LocaleType, LocaleType> = {};
+let _throwWhenNotFound = false;
 
 export function setLocale(locale: LocaleType | null) {
     _locale = locale;
@@ -50,6 +51,10 @@ export function getLocale(): LocaleType {
         document.documentElement.lang || // <html lang="en">
         'en'
     );
+}
+
+export function throwWhenNotFound(enabled: boolean): void {
+    _throwWhenNotFound = enabled;
 }
 
 export function setLocaleFallbacks(localeFallbacks: Record<LocaleType, LocaleType>): void {
@@ -160,6 +165,10 @@ export function trans<
         if (locale) {
             return format(translations[locale], parameters, locale);
         }
+    }
+
+    if (_throwWhenNotFound) {
+        throw new Error(`No translation message found with id "${message.id}".`);
     }
 
     return message.id;
--- a/src/Translator/assets/test/translator.test.ts
+++ b/src/Translator/assets/test/translator.test.ts
@@ -4,6 +4,7 @@ import {
     type NoParametersType,
     setLocale,
     setLocaleFallbacks,
+    throwWhenNotFound,
     trans,
 } from '../src/translator';
 
@@ -11,6 +12,7 @@ describe('Translator', () => {
     beforeEach(() => {
         setLocale(null);
         setLocaleFallbacks({});
+        throwWhenNotFound(false);
         document.documentElement.lang = '';
         document.documentElement.removeAttribute('data-symfony-ux-translator-locale');
     });
@@ -385,6 +387,40 @@ describe('Translator', () => {
             expect(trans(MESSAGE_MULTI_DOMAINS_WITH_PARAMETERS, { '%parameter1%': 'foo' }, 'messages', 'fr')).toEqual(
                 'message.multi_domains.different_parameters'
             );
+        });
+
+        test('missing message should return the message id when `throwWhenNotFound` is false', () => {
+            throwWhenNotFound(false);
+            setLocale('fr');
+
+            const MESSAGE_IN_ANOTHER_DOMAIN: Message<{ security: { parameters: NoParametersType } }, 'en'> = {
+                id: 'Invalid credentials.',
+                translations: {
+                    messages: {
+                        en: 'Invalid credentials.',
+                    },
+                },
+            };
+
+            expect(trans(MESSAGE_IN_ANOTHER_DOMAIN)).toEqual('Invalid credentials.');
+        });
+
+        test('missing message should throw an error if `throwWhenNotFound` is true', () => {
+            throwWhenNotFound(true);
+            setLocale('fr');
+
+            const MESSAGE_IN_ANOTHER_DOMAIN: Message<{ security: { parameters: NoParametersType } }, 'en'> = {
+                id: 'Invalid credentials.',
+                translations: {
+                    messages: {
+                        en: 'Invalid credentials.',
+                    },
+                },
+            };
+
+            expect(() => {
+                trans(MESSAGE_IN_ANOTHER_DOMAIN);
+            }).toThrow(`No translation message found with id "Invalid credentials.".`);
         });
 
         test('message from intl domain should be prioritized over its non-intl equivalent', () => {
--- a/src/Translator/doc/index.rst
+++ b/src/Translator/doc/index.rst
@@ -98,6 +98,27 @@ By default, the default locale is ``en`` (English) that you can configure throug
 #. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute
 #. Or with ``<html lang="your-locale">`` attribute
 
+Detecting missing translations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, the translator will return the translation key if the translation is missing.
+
+You can change this behavior by calling ``throwWhenNotFound(true)``:
+
+.. code-block:: diff
+
+      // assets/translator.js
+
+    - import { trans, getLocale, setLocale, setLocaleFallbacks } from '@symfony/ux-translator';
+    + import { trans, getLocale, setLocale, setLocaleFallbacks, throwWhenNotFound } from '@symfony/ux-translator';
+      import { localeFallbacks } from '../var/translations/configuration';
+
+      setLocaleFallbacks(localeFallbacks);
+    + throwWhenNotFound(true)
+
+      export { trans }
+      export * from '../var/translations';
+
 Importing and using translations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/Translator/assets/dist/translator.d.ts
+++ b/src/Translator/assets/dist/translator.d.ts
@@ -21,6 +21,7 @@ export interface Message<Translations extends TranslationsType, Locale extends L
 }
 export declare function setLocale(locale: LocaleType | null): void;
 export declare function getLocale(): LocaleType;
+export declare function throwWhenNotFound(enabled: boolean): void;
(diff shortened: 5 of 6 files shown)


Kocal (commented):
Please see https://github.com/symfony/ux/issues/2183#issuecomment-2360783282 for discussion / the implementation details

src/Translator/assets/src/translator.ts
```diff
@@ -52,6 +53,18 @@ export function getLocale(): LocaleType {
     );
 }
 
+export function setThrowWhenNotFound(throwWhenNotFound: boolean): void {
+    _throwWhenNotFound = throwWhenNotFound;
+}
```
Kocal:
```suggestion
export function throwWhenNotFound(enabled: boolean): void {
    _throwWhenNotFound = enabled;
}
```

src/Translator/assets/src/translator.ts
```diff
@@ -52,6 +53,18 @@ export function getLocale(): LocaleType {
...
+
+export function getThrowWhenNotFound(): boolean {
+    return (
+        _throwWhenNotFound ||
+        document.documentElement.getAttribute('data-symfony-ux-translator-throw-when-not-found') === 'true' || // <html data-symfony-ux-translator-throw-when-not-found="true">
+        false
+    );
+}
```
Kocal: The usage of `_throwWhenNotFound` will be internal, I don't think we need a function.

src/Translator/assets/src/translator.ts
```diff
@@ -162,5 +176,9 @@ export function trans<
         }
     }
 
+    if (getThrowWhenNotFound()) {
```
Kocal:
```suggestion
    if (_throwWhenNotFound) {
```

src/Translator/assets/src/translator.ts
```diff
@@ -162,5 +176,9 @@ export function trans<
...
+        throw new Error(`No message with this id found: ${message.id}`);
```
Kocal:
```suggestion
        throw new Error(`No translation message with id "${message.id}" is found.`);
```

src/Translator/src/DependencyInjection/Configuration.php
```diff
@@ -29,6 +29,7 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->children()
                 ->scalarNode('dump_directory')->defaultValue('%kernel.project_dir%/var/translations')->end()
+                ->booleanNode('throw_when_not_found')->defaultTrue()->end()
```
Kocal: We don't want to introduce a parameter that can breaks existing apps, in a minor release:
```suggestion
                ->booleanNode('throw_when_not_found')->defaultFalse()->end()
```

Kocal (changes requested):

Kocal:
Hum, my bad, in fact, we totally don't need to update the configuration file at all. 

The function `throwWhenNotFound()` must be explicitly called by developers if they want to, this can be done in the `assets/translator.js` file.

Also, the documentation `src/Translator/doc/index.rst` must be updated too!

Crovitche-1623:
Now I consider my PR ready @Kocal. Let me know what you think.

src/Translator/assets/test/translator.test.ts
```diff
@@ -387,6 +388,27 @@ describe('Translator', () => {
             );
         });
 
+        test('missing message should throw an error if `throwWhenNotFound` is true', () => {
+            throwWhenNotFound(true);
```
Kocal, line 409: You must ensure that the initial state is reset before each tests, `throwWhenNotFound(false)` is missing in the first `beforeEach()` of this file

src/Translator/assets/test/translator.test.ts
```diff
@@ -387,6 +388,27 @@ describe('Translator', () => {
...
+                translations: {
+                    messages: {
+                        en: 'Invalid credentials.',
+                    },
+                },
+            };
+
+            expect(() => {
+                trans(MESSAGE_IN_ANOTHER_DOMAIN);
+            }).toThrow(`No translation message with id "${MESSAGE_IN_ANOTHER_DOMAIN.id}" is found.`);
```
Kocal: Prefer to inline strings in tests, this way we are sure about what they contains:

```suggestion
            }).toThrow(`No translation message with id "Invalid credentials." is found.`);
```

src/Translator/assets/test/translator.test.ts
```diff
@@ -387,6 +388,27 @@ describe('Translator', () => {
...
+
+            // Set the default behavior for the next tests
+            throwWhenNotFound(false);
```
Kocal: To move in `beforeEach`, state resetting should be done **before** a test and not after a previous test, see https://docs.cypress.io/guides/references/best-practices#State-reset-should-go-before-each-test

src/Translator/doc/index.rst
```diff
@@ -98,6 +98,13 @@ By default, the default locale is ``en`` (English) that you can configure throug
 #. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute
 #. Or with ``<html lang="your-locale">`` attribute
 
+Detecting missing translations on the front-end
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, the translator will return the translation key as is when the translation is missing.
+
+You can configure this behavior to throw an error instead by calling ``throwWhenNotFound(true)`` from ``@symfony/ux-translator`` package.
```
Kocal: I think it would be more friendly to display the function call like this:
```suggestion
You can configure this behavior to throw an error instead by calling ``throwWhenNotFound(true)``:

.. code-block:: diff

    // assets/translator.js

    /*
     * This file is part of the Symfony UX Translator package.
     *
     * If folder "../var/translations" does not exist, or some translations are missing,
     * you must warmup your Symfony cache to refresh JavaScript translations.
     *
     * If you use TypeScript, you can rename this file to "translator.ts" to take advantage of types checking.
     */

    -import { trans, getLocale, setLocale, setLocaleFallbacks } from '@symfony/ux-translator';
    +import { trans, getLocale, setLocale, setLocaleFallbacks, throwWhenNotFound } from '@symfony/ux-translator';
    import { localeFallbacks } from '../var/translations/configuration';

    setLocaleFallbacks(localeFallbacks);
    +throwWhenNotFound(true)

    export { trans }
    export * from '../var/translations';
```

@javiereguiluz do you know if we can mix diff + PHP languages highlight? 🙏🏻

src/Translator/assets/test/translator.test.ts
Crovitche-1623: Done.

src/Translator/assets/src/translator.ts
```diff
@@ -162,5 +167,9 @@ export function trans<
         }
     }
 
+    if (_throwWhenNotFound) {
+        throw new Error(`No translation message with id "${message.id}" is found.`);
```
smnandre:
```suggestion
        throw new Error(`No translation message found with id "${message.id}".`);
```

src/Translator/doc/index.rst
```diff
@@ -98,6 +98,36 @@ By default, the default locale is ``en`` (English) that you can configure throug
 #. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute
 #. Or with ``<html lang="your-locale">`` attribute
 
+Detecting missing translations on the front-end
```
```suggestion
Detecting missing translations
```

src/Translator/doc/index.rst
```diff
@@ -98,6 +98,36 @@ By default, the default locale is ``en`` (English) that you can configure throug
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, the translator will return the translation key as is when the translation is missing.
```
```suggestion
By default, the translator will return the translation key if the translation is missing.
```

src/Translator/doc/index.rst
```diff
@@ -98,6 +98,36 @@ By default, the default locale is ``en`` (English) that you can configure throug
...
+
+You can configure this behavior to throw an error instead by calling ``throwWhenNotFound(true)``:
```
```suggestion
You can configure this behavior to throw an error instead by calling ``throwWhenNotFound(true)``:
You can change this behavior by calling ``throwWhenNotFound(true)``:
```

src/Translator/doc/index.rst
```diff
@@ -98,6 +98,36 @@ By default, the default locale is ``en`` (English) that you can configure throug
...
+    // assets/translator.js
+
+    /*
+     * This file is part of the Symfony UX Translator package.
+     *
+     * If folder "../var/translations" does not exist, or some translations are missing,
+     * you must warmup your Symfony cache to refresh JavaScript translations.
+     *
+     * If you use TypeScript, you can rename this file to "translator.ts" to take advantage of types checking.
+     */
```
```suggestion

```

src/Translator/assets/test/translator.test.ts
```diff
@@ -387,6 +389,24 @@ describe('Translator', () => {
             );
         });
 
+        test('missing message should throw an error if `throwWhenNotFound` is true', () => {
```
smnandre, line 408: Could you add a test for `throwWhenNotFound(false)` ?

src/Translator/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.20.0
+
+-   Add option to throw an exception when a translation is missing.
```
smnandre:
```suggestion
-   Add `throwWhenNotFound` function to configure the behavior when a translation is not found.
```

smnandre (approved):

src/Translator/doc/index.rst
smnandre: `.. code-block:: diff`

This is alread used in the symfony docs, but never in the diff+language or language+diff way.

Capture d’écran 2024-09-21 à 21 58 18

The .rst source: 

```rst

Once enabled, the security system redirects unauthenticated visitors to the
``login_path`` when they try to access a secured place (this behavior can
be customized using :ref:`authentication entry points <security-entry-point>`).

Edit the login controller to render the login form:

.. code-block:: diff

      // ...
    + use Symfony\Component\Security\Http\Authentication\AuthenticationUtils;

      class LoginController extends AbstractController
      {
          #[Route('/login', name: 'app_login')]
    -     public function index(): Response
    +     public function index(AuthenticationUtils $authenticationUtils): Response
          {
```

Kocal (approved):

Kocal:

PR #2196, opened 2024-09-23
[Turbo] Add Helper/TurboStream::append() et al. methods

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

For thoughts, this might help, WDYT?

```php
return new TurboResponse(TurboStream::append($target, $html));
```

(Note that I'm also wondering about having a twig component to render streams from the twig side, stay tuned)

### DIFF
--- a/src/Turbo/src/Helper/TurboStream.php
+++ b/src/Turbo/src/Helper/TurboStream.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo\Helper;
+
+/**
+ * @see https://turbo.hotwired.dev/reference/streams
+ */
+final class TurboStream
+{
+    /**
+     * Appends to the element(s) designated by the target CSS selector.
+     */
+    public static function append(string $target, string $html): string
+    {
+        return self::wrap('append', $target, $html);
+    }
+
+    /**
+     * Prepends to the element(s) designated by the target CSS selector.
+     */
+    public static function prepend(string $target, string $html): string
+    {
+        return self::wrap('prepend', $target, $html);
+    }
+
+    /**
+     * Replaces the element(s) designated by the target CSS selector.
+     */
+    public static function replace(string $target, string $html, bool $morph = false): string
+    {
+        return self::wrap('replace', $target, $html, $morph ? ' method="morph"' : '');
+    }
+
+    /**
+     * Updates the content of the element(s) designated by the target CSS selector.
+     */
+    public static function update(string $target, string $html, bool $morph = false): string
+    {
+        return self::wrap('update', $target, $html, $morph ? ' method="morph"' : '');
+    }
+
+    /**
+     * Removes the element(s) designated by the target CSS selector.
+     */
+    public static function remove(string $target): string
+    {
+        return \sprintf('<turbo-stream action="remove" targets="%s"></turbo-stream>', htmlspecialchars($target));
+    }
+
+    /**
+     * Inserts before the element(s) designated by the target CSS selector.
+     */
+    public static function before(string $target, string $html): string
+    {
+        return self::wrap('before', $target, $html);
+    }
+
+    /**
+     * Inserts after the element(s) designated by the target CSS selector.
+     */
+    public static function after(string $target, string $html): string
+    {
+        return self::wrap('after', $target, $html);
+    }
+
+    /**
+     * Initiates a Page Refresh to render new content with morphing.
+     *
+     * @see Initiates a Page Refresh to render new content with morphing.
+     */
+    public static function refresh(?string $requestId = null): string
+    {
+        if (null === $requestId) {
+            return '<turbo-stream action="refresh"></turbo-stream>';
+        }
+
+        return \sprintf('<turbo-stream action="refresh" request-id="%s"></turbo-stream>', htmlspecialchars($requestId));
+    }
+
+    private static function wrap(string $action, string $target, string $html, string $attr = ''): string
+    {
+        return \sprintf(<<<EOHTML
+            <turbo-stream action="%s" targets="%s"%s>
+                <template>%s</template>
+            </turbo-stream>
+            EOHTML, $action, htmlspecialchars($target), $attr, $html);
+    }
+}
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 2.21.0
+
+-   Add `Helper/TurboStream::append()` et al. methods
+-   Add `TurboStreamResponse`
+
 ## 2.19.0
 
 -   Fix Doctrine proxies are not Broadcasted #3139
--- a/src/Turbo/doc/index.rst
+++ b/src/Turbo/doc/index.rst
@@ -374,7 +374,7 @@ Let's discover how to use Turbo Streams to enhance your `Symfony forms`_::
     {% endblock %}
 
 Supported actions are ``append``, ``prepend``, ``replace``, ``update``,
-``remove``, ``before`` and ``after``.
+``remove``, ``before``, ``after`` and ``refresh``.
 `Read the Turbo Streams documentation for more details`_.
 
 Resetting the Form
--- a/src/Turbo/phpstan.neon.dist
+++ b/src/Turbo/phpstan.neon.dist
@@ -45,3 +45,8 @@ parameters:
 			message: "#^Call to an undefined method Doctrine\\\\ORM\\\\Event\\\\PostFlushEventArgs\\:\\:getEntityManager\\(\\)\\.$#"
 			count: 1
 			path: src/Doctrine/BroadcastListener.php
+
+		-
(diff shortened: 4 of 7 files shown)


Kocal (commented):
I'm not a Turbo user, but I think it can be useful yes.

Just, can you add some documentation for this new feature please? And also fix failing checks (IMO we can ignore the error found by PHPStan).

Thanks! :)

nicolas-grekas:
Updated

rvmourik:
Great addition imho, but what about multiple streams in a single response? Or is that the Twig solution you are talking about @nicolas-grekas?

```php
return new TurboResponse(TurboStream::append($target, $html));
```

Regarding DX it is not optimal to have this I guess:

```php
return new TurboResponse(TurboStream::append($target, $html) . TurboStream::replace($target2, $html2));
```

Is this something to consider?

smnandre:
Thanks a lot @nicolas-grekas! More comments tomorrow :)

nicolas-grekas:
@rvmourik concatenation FTW indeed. I don't see a better option. In the end, any alternative would just be a fancy way to do concatenation.

rvmourik:
> @rvmourik concatenation FTW indeed. I don't see a better option. In the end, any alternative would just be a fancy way to do concatenation.

@nicolas-grekas thanks for the quick reply.

I know the TurboResponse is extending the base Response class but if we take a look at the JsonResponse there is also an option to pass an array or a mixed value, when it is mixed in the end is converts it to an ArrayObject. Could the TurboResponse do something similar, always convert the passed stream(s) to be a collection of streams and processing them before passing it to the base Response?

Maybe I am overthinking it? 😇

src/Turbo/src/TurboResponse.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo;
+
+use Symfony\Component\HttpFoundation\Response;
+
+class TurboResponse extends Response
```
smnandre: TurboStreamResponse ?

src/Turbo/src/TurboResponse.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    public function __construct(?string $content = '', int $status = 200, array $headers = [])
+    {
```
smnandre: Both HttpFoundation StreamedResponse and [StreamedJsonResponse](https://github.com/symfony/symfony/blob/f42d8b4e9b3bd4724c24b309dbdbdc11c6e9244d/src/Symfony/Component/HttpFoundation/StreamedJsonResponse.php#L55-L60) have specialized constructors.

What if we made TurboStream a stringable object and have this kind of signature 

```php
/**
 * @param string|Stringable|iterable<string|Stringable> $stream
 */ 
public function __construct(string|iterable $streams = '', int $status = 200, array $headers = [])
{
```

(Genuine suggestion, not sure what would be the best DX here)

src/Turbo/src/TurboResponse.php
```diff
@@ -0,0 +1,26 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo;
+
+use Symfony\Component\HttpFoundation\Response;
+
+class TurboResponse extends Response
+{
+    public function __construct(?string $content = '', int $status = 200, array $headers = [])
```
smnandre:
```suggestion
    /**
     * @param array<string, string|string[]> $headers
     */
    public function __construct(?string $content = '', int $status = 200, array $headers = [])
```

To please the CI :)

javiereguiluz:
If we decide to add some helpers to make using Turbo more user-friendly, what if we all DX and provide a fluent interface to not deal with any internal details?

```php
return TurboResponse()
    ->append($targetId, $html)
    ->prepend($targetId, $html)
    ->replace($targetId, $html, $method = 'morph')
    ->update($targetId, $html, $method = 'morph')
    ->remove($targetId, $html)
    ->before($targetId, $html)
    ->after($targetId, $html)
    // useful if the action (append, prepend, etc.) is a PHP variable:
    ->stream($action, $targetId, $html)
;
```

@rvmourik your example now would look like this:

```php
// before
return new TurboResponse(TurboStream::append($target, $html) . TurboStream::replace($target2, $html2));

// after
return new TurboResponse()
    ->append($target, $html)
    ->append($target2, $html2)
;
```

There's a final consideration, which is not covered either in the current proposal of this PR. In addition to `target` (the targetID), you can use `targets` (a CSS selector) to apply the stream to multiple targets (See https://turbo.hotwired.dev/handbook/streams#actions-with-multiple-targets).

Adding an optional `targets` param would look ugly in my opinion:

```php
return TurboResponse()
    ->append(string|null $targetId, string|null $targets, string $html)
    ->prepend(string|null $targetId, string|null $targets, string $html)
    // ...
```

We could add a `*All()` series of methods:

```php
return TurboResponse()
    ->append(string $targetId, string $html)
    ->appendAll(string $targets, string $html)

    ->prepend(string $targetId, string $html)
    ->prependAll(string $targets, string $html)
    // ...
```

What do you think?

rvmourik:
> If we decide to add some helpers to make using Turbo more user-friendly, what if we all DX and provide a fluent interface to not deal with any internal details?
> 
> ```
> return TurboResponse()
>     ->append($targetId, $html)
>     ->prepend($targetId, $html)
>     ->replace($targetId, $html, $method = 'morph')
>     ->update($targetId, $html, $method = 'morph')
>     ->remove($targetId, $html)
>     ->before($targetId, $html)
>     ->after($targetId, $html)
>     // useful if the action (append, prepend, etc.) is a PHP variable:
>     ->stream($action, $targetId, $html)
> ;
> ```
> 
> @rvmourik your example now would look like this:
> 
> ```
> // before
> return new TurboResponse(TurboStream::append($target, $html) . TurboStream::replace($target2, $html2));
> 
> // after
> return new TurboResponse()
>     ->append($target, $html)
>     ->append($target2, $html2)
> ;
> ```
> 
> There's a final consideration, which is not covered either in the current proposal of this PR. In addition to `target` (the targetID), you can use `targets` (a CSS selector) to apply the stream to multiple targets (See https://turbo.hotwired.dev/handbook/streams#actions-with-multiple-targets).
> 
> Adding an optional `targets` param would look ugly in my opinion:
> 
> ```
> return TurboResponse()
>     ->append(string|null $targetId, string|null $targets, $html)
>     ->prepend(string|null $targetId, string|null $targets, $html)
>     // ...
> ```
> 
> We could add a `*All()` series of methods:
> 
> ```
> return TurboResponse()
>     ->append(string $targetId, $html)
>     ->appendAll(string $targets, $html)
> 
>     ->prepend(string $targetId, string $html)
>     ->prependAll(string $targets, $html)
>     // ...
> ```
> 
> What do you think?

I think this is a good suggestion, maybe in a future PR the same can be done with adding a helper for a regular turbo frame. I find my self often doing one of the following:

a) creating a twig file only with the frame and HTML in it
b) manually building a response like in the helper in this PR en load the twig file between the <turbo-frame> tag in the response.

Thanks all.

src/Turbo/src/TurboResponse.php
nicolas-grekas: Adding this doesn't make sense to me - the parent doesn't declare anything so it doesn't make sense that a child class does it. That'd be random, just to please a tool with no brain.

src/Turbo/src/TurboResponse.php
nicolas-grekas: That's not possible: a response has to extend StreamedResponse if it streams, and not otherwise.
We'd need to have two kind of responses: StreamedTurboStreamResponse and TurboStreamResponse.

BUT, I don't see the need for a StreamedTurboStreamResponse honestly. People should use StreamedResponse with an iterable that uses `TurboStream::*()` methods instead.

nicolas-grekas:
> what if we all DX and provide a fluent interface

Works for me, I added the fluent methods to the TurboStreamResponse class.
Note that I don't know yet if that should be the recommended way to generate turbo stream: the drawback of this approach is that it splits targets from actual HTML, so that it can be easy to forget to update a target after updating a template. Putting the stream tags in the template that they target would improve this by providing locality.

> In addition to target (the targetID), you can use targets (a CSS selector) to apply the stream to multiple targets

You might have missed this in the attached patch: this PR uses only targets, with an "s". That means you have to use `#foo` to target id foo. To me, this is the most empowering and DX-friendly way to express the target. Using target (no "s") provides zero benefits.

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -0,0 +1,97 @@
...
+            return '<turbo-stream action="refresh"></turbo-stream>';
+        }
+
+        return \sprintf('<turbo-stream action="refresh" request-id="%s"></turbo-stream>', htmlspecialchars($requestId));
+    }
+
+    private static function wrap(string $action, string $target, string $html, string $attr = ''): string
+    {
+        return \sprintf(<<<EOHTML
+            <turbo-stream action="%s" targets="%s"%s>
```
nicolas-grekas, line 92: Note the "s" to targets, and the docblock comments: we support only CSS selectors, not pure IDs (CSS selectors are more capable anyway).

nicolas-grekas:
PR ready IMHO (failure unrelated IIUC)

javiereguiluz (approved):

nicolas-grekas:
All green :)

Kocal (approved):

Kocal:
Thank you @nicolas-grekas.

PR #2193, opened 2024-09-23
Fixed typo in links to Symfony UX repo

sblum51:
I fixed typo in the links to Symfony UX repo (before: symfony/symfony-ux).
Note: I wasn't able to deal with the "version" variable.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature? |  no 
| Issues        | 
| License       | MIT

Before :  https://github.com/symfony/symfony-ux/blob/4.4/src/Map/src/Bridge/Leaflet/README.md  (404)
Now : https://github.com/symfony/ux/blob/2.x/src/Map/src/Bridge/Leaflet/README.md

PR #2198, opened 2024-09-23
[Doc] Fix some links in UX Map docs

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

Related to #2193. The `{version}` placeholder only works for Symfony Docs and I'm not sure it can be extended to also work on UX. So, let's just hardcode the `2.x` version number because this is fine.


Kocal:
Thank you @javiereguiluz.

PR #2200, opened 2024-09-23
[LiveComponent] Fix DeferLiveComponentSubscriber service definition

smnandre:
Removes two unused arguments in the service definition

(DeferLiveComponentSubscriber has no constructor)


Kocal:
Thank you @smnandre.

PR #2206, opened 2024-09-25
[Translator] Many performance improvements on the cache warming

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

The Symfony UX Translator cache warmer can be an long task, especially if your app has a lot of translations.

I was able to identify some bottlenecks and fix them, with a dedicated commit for each of them.

My test application have ~25k translation keys (3.119 keys * 8 locales). Of course that's a lot, but I believe some people needs to dump all their translations (e.g. if they build an SPA on Symfony). For "more classic" applications, it is recommended to [use dedicated domains for translations JS-side](https://symfony.com/bundles/ux-translator/current/index.html#configuring-the-dumped-translations).

Anyway, in my application, I was able to reduce the Translator cache warming **from 2m11s** (higher than the reality, because of Blackfire profiling) **to 41s** (higher than the reality, because of Blackfire profiling):
image

The vast majority of improvements come from better use of [the Symfony String component](https://symfony.com/doc/current/string.html).

Blackfire profiles:
1. [Initial](https://app.blackfire.io/profiles/2e264f01-c3a6-46de-92fa-0856b2078010/graph) 
2. [Re-using `s()` instances](https://app.blackfire.io/profiles/b4c5dea7-4309-4cb8-b99f-584ec3a4ddc9/graph)
3. [Removing useless `s()->slice()`](https://app.blackfire.io/profiles/f66b6bdf-39da-4d98-a966-781418b3a6f5/graph)
4. [Caching `$this->message->length()`](https://app.blackfire.io/profiles/1334e3e1-4405-44d1-a327-e446dad7d1d8/graph) as it will always return the same length
5. [Before/After](https://blackfire.io/profiles/compare/8a0df8d0-5d01-47bf-a064-bfed6783c5a3/graph): ~63% 🚀 

Also, since the Translator component is still experimental, I've marked some classes as final and internal, as they should only be used internally. :)

### DIFF
--- a/src/Translator/src/Intl/ErrorKind.php
+++ b/src/Translator/src/Intl/ErrorKind.php
@@ -14,7 +14,7 @@ namespace Symfony\UX\Translator\Intl;
 /**
  * Adapted from https://github.com/formatjs/formatjs/blob/590f1f81b26934c6dc7a55fff938df5436c6f158/packages/icu-messageformat-parser/error.ts#L9-L77.
  *
- * @experimental
+ * @internal
  */
 final class ErrorKind
 {
--- a/src/Translator/src/TranslationsDumper.php
+++ b/src/Translator/src/TranslationsDumper.php
@@ -189,13 +189,14 @@ TS
     {
         static $alreadyGenerated = [];
 
+        $translationId = s($translationId)->ascii()->snake()->upper()->replaceMatches('/^(\d)/', '_$1')->toString();
         $prefix = 0;
         do {
-            $constantName = s($translationId)->ascii()->snake()->upper()->replaceMatches('/^(\d)/', '_$1')->toString().($prefix > 0 ? '_'.$prefix : '');
+            $constantName = $translationId.($prefix > 0 ? '_'.$prefix : '');
             ++$prefix;
-        } while (\in_array($constantName, $alreadyGenerated, true));
+        } while ($alreadyGenerated[$constantName] ?? false);
 
-        $alreadyGenerated[] = $constantName;
+        $alreadyGenerated[$constantName] = true;
 
         return $constantName;
     }
--- a/src/Translator/src/TranslationsDumper.php
+++ b/src/Translator/src/TranslationsDumper.php
@@ -155,21 +155,18 @@ TS
                 }
 
                 $parametersTypes[$domain] = $this->typeScriptMessageParametersPrinter->print($parameters);
-
                 $locales[] = $locale;
             }
         }
 
+        $typeScriptParametersType = [];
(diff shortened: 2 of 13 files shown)


src/Translator/src/Intl/ErrorKind.php
```diff
@@ -15,6 +15,8 @@
  * Adapted from https://github.com/formatjs/formatjs/blob/590f1f81b26934c6dc7a55fff938df5436c6f158/packages/icu-messageformat-parser/error.ts#L9-L77.
  *
  * @experimental
+ *
+ * @internal
```
smnandre:
```suggestion
 * @internal
```

Let's remove `@experimental` everywhere you add "`@internal`" ... as internal classes won't be coverd by BC promise anyway :)

src/Translator/src/TranslationsDumper.php
```diff
@@ -189,13 +186,14 @@ private function generateConstantName(string $translationId): string
     {
         static $alreadyGenerated = [];
 
+        $translationId = s($translationId)->ascii()->snake()->upper()->replaceMatches('/^(\d)/', '_$1')->toString();
```
smnandre, line 189: If you have an ascii string here at one point, you'be probably quicker to compute the upper snake yourself, as the String implementation does some form of camel -> snake 

(cf https://github.com/twigphp/Twig/blob/3.x/src/Util/CallableArgumentsExtractor.php and the whole PR about it)

smnandre (commented):
Trusting you here :)

kbond (approved):

src/Translator/src/TranslationsDumper.php
Kocal: I've just did a quick test with:
```php
$translationId = preg_replace(['/([A-Z]+)([A-Z][a-z])/', '/([a-z0-9])([A-Z])/'], '\1_\2', $translationId);
$translationId = str_replace(['.', '-'], '_', $translationId);
$translationId = strtoupper($translationId);
```

but it does not generate the same translationId than before. Did I miss something?

smnandre: Damn ... :( This string implementation has really too many impacts :( 

Ignore my comment then 😢

smnandre (approved):
🤝

src/Translator/src/TranslationsDumper.php
Kocal: Yeah that's not an easy topic 😅

PR #2210, opened 2024-09-26
[TwigComponent] Fix ux:icon & ux:map renders

smnandre:
Change implementation to avoid many dispatched events for nothing

Add a conflict in symfony/ux-map and symfony/ux-icons  to synchronise the 3 packages to 2.21

Performance gain around 20%


Kocal:
> Performance gain around 20%

Can you link the Blackfire profiles delta plz?

javiereguiluz (approved):

kbond (approved):

javiereguiluz:
Thank you Simon.

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -108,6 +110,10 @@ class_exists(AbstractArgument::class) ? new AbstractArgument(\sprintf('Added in
         $container->register('.ux.twig_component.twig.component_runtime', ComponentRuntime::class)
             ->setArguments([
                 new Reference('ux.twig_component.component_renderer'),
+                service_locator([
+                    'ux:icon' => service('.ux_icons.twig_icon_runtime')->nullOnInvalid(),
+                    'ux:map' => service('ux_map.twig_runtime')->nullOnInvalid(),
```
nicolas-grekas, line 115: chiming here a bit late: this doesn't work
a service locator must point to non-null references
when these are turned into nulls, XmlDumper explodes (and that's a symptom of this code, which is not "legal")

Also, from a design pov, this looks super hardcoded. How can one pass more things to the locator?

This should use a tagged locator instead.

nicolas-grekas: This was spotted by the CI BTW:
https://github.com/symfony/ux/actions/runs/11071829929/job/30764609180

PR #2211, opened 2024-09-26
[TwigComponent] Cache component properties metadata

smnandre:
Currently if we have 50 times the same component in a page, 50 times we use reflection to analyse component class properties and methods.

This PR centralize this task in a dedicated (internal) service and add a cachewarmer to pre-compute metadata during app build.

Significant performance gains here too 
(i won't do charts for every PR but be sure i'm gonna make some before/after once i'm "done" with all this.... in Vienna 👼 )

### DIFF
--- a/src/TwigComponent/src/CacheWarmer/TwigComponentCacheWarmer.php
+++ b/src/TwigComponent/src/CacheWarmer/TwigComponentCacheWarmer.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\CacheWarmer;
+
+use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerInterface;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
+use Symfony\UX\TwigComponent\ComponentProperties;
+
+/**
+ * Warm the TwigComponent metadata caches.
+ *
+ * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
+ */
+final class TwigComponentCacheWarmer implements CacheWarmerInterface, ServiceSubscriberInterface
+{
+    /**
+     * As this cache warmer is optional, dependencies should be lazy-loaded, that's why a container should be injected.
+     */
+    public function __construct(
+        private readonly ContainerInterface $container,
+    ) {
+    }
+
+    public static function getSubscribedServices(): array
+    {
+        return [
+            'ux.twig_component.component_properties' => ComponentProperties::class,
+        ];
+    }
+
+    public function warmUp(string $cacheDir, ?string $buildDir = null): array
+    {
+        $properties = $this->container->get('ux.twig_component.component_properties');
+        $properties->warmup();
+
+        return [];
+    }
+
+    public function isOptional(): bool
+    {
+        return true;
+    }
+}
--- a/src/TwigComponent/src/ComponentProperties.php
+++ b/src/TwigComponent/src/ComponentProperties.php
@@ -1,0 +1,148 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+use Symfony\Component\Cache\Adapter\AdapterInterface;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ *
+ * @internal
+ */
+final class ComponentProperties
+{
+    private const CACHE_KEY = 'ux.twig_component.component_properties';
+
+    /**
+     * @var array<class-string, array{
+     *     properties: array<class-string, array{string, array{string, string, bool}, bool}>,
+     *     methods: array<class-string, array{string, array{string, bool}}>,
+     *  }|null>
+     */
+    private array $classMetadata;
+
+    public function __construct(
+        private readonly PropertyAccessorInterface $propertyAccessor,
+        ?array $classMetadata = [],
+        private readonly ?AdapterInterface $cache = null,
+    ) {
+        $cacheItem = $this->cache?->getItem(self::CACHE_KEY);
+
+        $this->classMetadata = $cacheItem?->isHit() ? [...$cacheItem->get(), ...$classMetadata] : $classMetadata;
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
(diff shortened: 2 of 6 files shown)


src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
```diff
@@ -80,8 +80,13 @@ public function process(ContainerBuilder $container): void
         $factoryDefinition->setArgument(4, $componentConfig);
         $factoryDefinition->setArgument(5, $componentClassMap);
 
+        $componentPropertiesDefinition = $container->findDefinition('ux.twig_component.component_properties');
+        $componentPropertiesDefinition->setArgument(1, array_fill_keys(array_keys($componentClassMap), null));
+
         $debugCommandDefinition = $container->findDefinition('ux.twig_component.command.debug');
         $debugCommandDefinition->setArgument(3, $componentClassMap);
+
+
```
Kocal:
```suggestion
```

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -82,21 +85,29 @@ static function (ChildDefinition $definition, AsTwigComponent $attribute) {
...
+                new AbstractArgument(\sprintf('Added in %s.', TwigComponentPass::class)),
             ])
         ;
 
         $container->register('ux.twig_component.component_stack', ComponentStack::class);
 
+        $container->register('ux.twig_component.component_properties', ComponentProperties::class)
+            ->setArguments([
+                new Reference('property_accessor'),
+                new AbstractArgument('Add in TwigComponentPass'),
```
Kocal:
```suggestion
                new AbstractArgument(\sprintf('Added in %s.', TwigComponentPass::class)),
```

src/TwigComponent/src/ComponentProperties.php
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * @internal
+ */
+final class ComponentProperties
+{
+    private const CACHE_KEY = 'ux.twig_component.component_properties';
+
+    /**
+     * @var array<class-string, array{
+     *    properties: array<class-string, array{string, array{string, string, bool}, bool}>,
```
Kocal:
```suggestion
     *     properties: array<class-string, array{string, array{string, string, bool}, bool}>,
```

src/TwigComponent/src/ComponentProperties.php
```diff
@@ -0,0 +1,148 @@
...
+     *     methods: array<class-string, array{string, array{string, bool}}>,
+     *  }|null>
+     */
+    private array $classMetadata;
+
+    public function __construct(
+        private readonly PropertyAccessorInterface $propertyAccessor,
+        ?array $classMetadata = [],
```
Kocal, line 37: Here we have array of arrays of arrays of arrays and it can be pretty memory-consuming, do you think using DTO intsances here can improve things a bit?

Kocal (approved):

src/TwigComponent/src/CacheWarmer/TwigComponentCacheWarmer.php
```diff
@@ -0,0 +1,55 @@
...
+    public static function getSubscribedServices(): array
+    {
+        return [
+            'ux.twig_component.component_properties' => ComponentProperties::class,
+        ];
+    }
+
+    public function warmUp(string $cacheDir, ?string $buildDir = null): array
+    {
+        $properties = $this->container->get('ux.twig_component.component_properties');
```
WebMamba, line 45: I don't understand why we don't inject the service directly. The #[Lazy] attribute is not enough to make this service lazy?

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -132,63 +132,20 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
             // add attributes
             [$metadata->getAttributesVar() => $mounted->getAttributes()],
         );
-        $event = new PreRenderEvent($mounted, $metadata, $variables);
 
+        // TODO pass only props & vars here
+        // --> no context, outerscope, magic, ..
+
```
WebMamba: Can be remove 😁

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -132,63 +132,20 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
...
+        $event = new PreRenderEvent($mounted, $metadata, $variables);
         $this->dispatcher->dispatch($event);
 
         return $event;
     }
 
-    private function exposedVariables(object $component, bool $exposePublicProps): \Iterator
+    public function warmup(): array
```
WebMamba: I don't think this method is used, right?

WebMamba (commented):
Looks pretty cool! Thanks @smnandre

src/TwigComponent/src/ComponentProperties.php
smnandre: Do you think the difference will be palpable, with something like 20/30ish entries? genuine question

src/TwigComponent/src/CacheWarmer/TwigComponentCacheWarmer.php
smnandre: We cannot use attributes in bundle (no autoconfigure).. and to be honest i took this code from symfony bundles :))

src/TwigComponent/src/ComponentRenderer.php
smnandre: Yep thx!

src/TwigComponent/src/ComponentRenderer.php
smnandre: Oops 😅  removed, thx !

src/TwigComponent/src/ComponentProperties.php
Kocal: I don't think so

kbond (approved):

onEXHovia:
As an alternative, or even as a follow-up, I think it makes sense to replace/improve class `ComponentMetadata`. We could get all possible meta information PostMount/PreMount/ExposeInTemplate hooks etc, not just about properties. 

Structure may look like this:
`ComponentMetadata` - store metadata
`ComponentMetadataFactory` - provider metadata with methods getMetadataFor/hasMetadataFor. Factory methods return `ComponentMetadata` instance.
`ComponentCacheWarmer` - cache warmer call factory `ComponentMetadataFactory`

src/TwigComponent/src/CacheWarmer/TwigComponentCacheWarmer.php
WebMamba: Ok perfect then! I didn't know this syntax

WebMamba (commented):
Are we agree, this is happening only in prod ? I don't see anything prod related

smnandre:
Hey @onEXHovia ! You read my mind 

This is indeed my next move, and i started to extract the props metadata from template too :)

smnandre:
> Are we agree, this is happening only in prod ?

@WebMamba nothing "prod-related" 

For all the environments, during a request, there is no need to compute metadata twice, so the values are "kept".

Also, if you use `cache:warm` all the components metadata are cached in the ssytem cache.

What's maybe missing here is something to invalidate the cache locally if you change a file ? 

Do you have encountered any trouble ?

Kocal:
image

A small rebase before merging? :) 🙏🏻

Kocal (approved):

PR #2220, opened 2024-09-28
[TwigComponent] Fix internal variables are dispatched

smnandre:
The key/values this/computed/outerScope/__props/__context are internal and should not have been exposed in Event

* it does not match the documentation
* it causes non needed memory usage / log
* modifying them would break the entire rendering process

(i'll make another PR soon to simplify & clarify `__props`, `__context`, and `props`)


Kocal (approved):

javiereguiluz:
Thanks Simon.

PR #2226, opened 2024-09-30
[UX Icons] Some updates in the full configuration example docs

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | - below instead -->
| License       | MIT

I propose to also show some options with problematic characters in their names (`:`, `-`) to let readers know how to use them.

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -541,30 +541,30 @@ Full Configuration
 
     # config/packages/ux_icons.yaml
     ux_icons:
-        # The local directory where icons are stored.
+        # The local directory where icons are stored
         icon_dir: '%kernel.project_dir%/assets/icons'
 
-        # Default attributes to add to all icons.
+        # Default attributes to add to all icons
         default_icon_attributes:
-            # Default:
             fill: currentColor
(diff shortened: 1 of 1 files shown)


src/Icons/doc/index.rst
```diff
@@ -541,30 +541,30 @@ Full Configuration
...
     ux_icons:
-        # The local directory where icons are stored.
+        # The local directory where icons are stored
         icon_dir: '%kernel.project_dir%/assets/icons'
 
         # Default attributes to add to all icons.
         default_icon_attributes:
-            # Default:
             fill: currentColor
+            'aria-hidden': true
```
norkunas:
```suggestion
            # 'aria-hidden': true # it is always added automatically unless 'aria-label', 'aria-labelledby' or 'title' is provided
```

Kocal (approved):

src/Icons/doc/index.rst
javiereguiluz: You are right. Using `aria-hidden` can be confusing. I replaced it by `'font-size': '1.25em'`

PR #2227, opened 2024-10-01
Add `<twig:Turbo:Stream:*>` components

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

This PR provides Twig components for turbo-streams:

```html
<twig:Turbo:Stream:Append target="#some-id">
    <div>Append this</div>
</twig:Turbo:Stream:Append>
```

Sibling PR to #2196
See https://github.com/symfony/ux/pull/2196#issuecomment-2386444070 for some background.

### DIFF
--- a/.github/workflows/test-turbo.yml
+++ b/.github/workflows/test-turbo.yml
@@ -8,6 +8,9 @@ on:
         paths:
             - 'src/Turbo/**'
 
+env:
+    SYMFONY_REQUIRE: '>=5.4'
+
 jobs:
     phpstan:
         runs-on: ubuntu-latest
@@ -38,11 +41,11 @@ jobs:
         strategy:
            fail-fast: false
            matrix:
-               php-version: ['8.1', '8.3']
+               php-version: ['8.1', '8.4']
                include:
                   - php-version: '8.1'
                     dependency-version: 'lowest'
-                  - php-version: '8.3'
+                  - php-version: '8.4'
                     dependency-version: 'highest'
 
         services:
@@ -68,6 +71,9 @@ jobs:
               uses: shivammathur/setup-php@v2
               with:
                   php-version: ${{ matrix.php-version }}
+
+            - name: Install symfony/flex
+              run: composer global config allow-plugins.symfony/flex true && composer global require symfony/flex
 
             - name: Install Turbo packages
               uses: ramsey/composer-install@v3
@@ -101,6 +107,8 @@ jobs:
 
             - name: Run tests
               working-directory: src/Turbo
-              run: vendor/bin/simple-phpunit
+              run: |
+                [ 'lowest' = '${{ matrix.dependency-version }}' ] && export SYMFONY_DEPRECATIONS_HELPER=weak
+                vendor/bin/simple-phpunit
               env:
                   SYMFONY_DEPRECATIONS_HELPER: 'max[self]=1'
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 -   Add `Helper/TurboStream::append()` et al. methods
 -   Add `TurboStreamResponse`
+-   Add `<twig:Turbo:Stream:*>` components
 
 ## 2.19.0
 
--- a/src/Turbo/composer.json
+++ b/src/Turbo/composer.json
@@ -52,7 +52,7 @@
         "symfony/property-access": "^5.4|^6.0|^7.0",
         "symfony/security-core": "^5.4|^6.0|^7.0",
         "symfony/stopwatch": "^5.4|^6.0|^7.0",
-        "symfony/ux-twig-component": "^2.0",
+        "symfony/ux-twig-component": "^2.21",
         "symfony/twig-bundle": "^5.4|^6.0|^7.0",
         "symfony/web-profiler-bundle": "^5.4|^6.0|^7.0",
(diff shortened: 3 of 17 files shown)


src/Turbo/src/Twig/Component/TurboStreamComponent.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * @see https://turbo.hotwired.dev/reference/streams
+ */
+class TurboStreamComponent
+{
+    public string $action;
+    public ?string $target;           // A CSS selector - not for "refresh" action
+    public bool $morph = false;       // Only for "replace" and "update" actions
+    public ?string $requestId = null; // Only for "refresh" action
```
Kocal: Won't PHP-CS-Fixer/Fabbot will complain about all spaces before comments?

What about a true phpdoc?

Kocal (approved):
LGTM, can you add some tests to check that components nicely render please? 
Thanks :)

neluba:
I don't know if the symfony phpstorm plugin works with third party components, but it might incorrectly offer autocompletions for all properties on all components, because they're inherited from TurboStreamComponent.

nicolas-grekas:
I figured out I could use anonymous twig components, so here we are. Tags are now like:

```html
<twig:Turbo:Stream:Append target="#some-id">
    <div>Append this</div>
</twig:Turbo:Stream:Append>
```

rvmourik:
> I figured out I could use anonymous twig components, so here we are. Tags are now like:
> 
> ```
> <twig:Turbo:Stream:Append target="#some-id">
>     <div>Append this</div>
> </twig:Turbo:Stream:Append>
> ```

Is there a specific reason why the attribute is called `target` and in the HTML using `targets`?

nicolas-grekas:
> Is there a specific reason why the attribute is called target and in the HTML using targets?

Good question. Turbo accepts both target and targets. The former accepts an HTML id, the latter a CSS selector.
IMHO this is provides bad DX. Using a CSS selector should be the only way: when reading `#id`, it's way clearer that we target an id, and also it's hinting that another CSS selector would work.

I'm for not supporting Turbo's target attribute.
Then, which name should we give to the CSS-selector attribute on our side? `target` looks the most sensible to me.
We could name it `targets`, but the "s" feels unnecessary.

rvmourik:
> > Is there a specific reason why the attribute is called target and in the HTML using targets?
> 
> Good question. Turbo accepts both target and targets. The former accepts an HTML id, the latter a CSS selector. IMHO this is provides bad DX. Using a CSS selector should be the only way: when reading `#id`, it's way clearer that we target an id, and also it's hinting that another CSS selector would work.
> 
> I'm for not supporting Turbo's target attribute. Then, which name should we give to the CSS-selector attribute on our side? `target` looks the most sensible to me. We could name it `targets`, but the "s" feels unnecessary.

I think `target` is the way to go, but maybe elaborate on the target/targets difference in the docs so people don't mistake the target attribute of the component with the HTML target attribute.

Nevertheless, great DX improvements with the two PR's 👍

kbond (approved):
❤️

nicolas-grekas:
#2230 should be fixed for tests to be green.

javiereguiluz (approved):

nicolas-grekas:
All green :)

Kocal (approved):

Kocal:
PR rebased and conflicts fixed, I merge when the CI is green.

Kocal:
Thank you Nicolas! :)

onEXHovia:
Is there any benefit from these components in a real application? This package can also be used without TwigComponent. Maybe we should be closer to the package https://github.com/hotwired/turbo-rails?

nicolas-grekas:
The benefit is structured declaration: you cannot make a typo, the IDE can help you, etc.

nicolas-grekas:
I don't about the turbo-rails package, so feel free to open issues/PRs with the specific things you think would be useful for Symfony apps. I'd be happy to review them.

onEXHovia:
> The benefit is structured declaration: you cannot make a typo, the IDE can help you, etc.

You right, but twig component not part of twig(I hope this will change) how blade+components in laravel. In projects where not used twig components will have to install +1 package, this could be a problem. Instead, we can use only twig functions:

```twig
{% turbo_stream 'apend', dom_id(todo) %}
{% turbo_frame_tag 'target' %}
```

I don't mind twig components with html syntax, but look at PR https://github.com/twigphp/Twig/issues/3867#issuecomment-2388573781 there are some problems with this. My idea is to use standard features twig to cover as many projects as possible.

smnandre:
I love this PR, and this is such a good example of anonymous components exposed by bundles !  (thanks @yceruto again)

(note to myself: I'm thinking we should use this in the documentation to illustrate!)

Thank you @nicolas-grekas ❤️

smnandre:
Also, in my mind there is no "feature" here that would _need_ to be delivered _without_ TwigComponent. Anyone can do something _similar_ with pure Twig if they want (via embeds or macros).. but for those who have the TwigComponent installed, it's a great sugar syntax / DX improvement. 


> I don't mind twig components with html syntax, but look at PR https://github.com/twigphp/Twig/issues/3867#issuecomment-2388573781 there are some problems with this. My idea is to use standard features twig to cover as many projects as possible.



You then can use `{{ component('Turbo:Stream', ... ) }}`  syntax.  


It is 100.00% compatible with Twig standard features and will stay this way.

PR #2231, opened 2024-10-02
[Twig][Icons][Map] fix wiring of `ComponentRuntime`

### LINKED ISSUE #2230: TwigComponentExtension shouldn't hardcode (broken) service locator
Reminder for https://github.com/symfony/ux/pull/2210#discussion_r1784494815

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2230
| License       | MIT

I've opted to not add an interface and keep this extension point internal to other `ux/*` components for now.

### DIFF
--- a/src/Icons/config/services.php
+++ b/src/Icons/config/services.php
@@ -53,6 +53,7 @@ return static function (ContainerConfigurator $container): void {
                 service('logger')->ignoreOnInvalid(),
             ])
             ->tag('twig.runtime')
+            ->tag('ux.twig_component.twig_renderer', ['key' => 'ux:icon'])
 
         ->set('.ux_icons.icon_renderer', IconRenderer::class)
             ->args([
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -17,6 +17,8 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\AbstractArgument;
+use Symfony\Component\DependencyInjection\Argument\ServiceLocatorArgument;
+use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\LogicException;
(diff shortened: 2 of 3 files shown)


src/Icons/config/services.php
```diff
@@ -53,6 +53,7 @@
                 service('logger')->ignoreOnInvalid(),
             ])
             ->tag('twig.runtime')
+            ->tag('.ux.twig_component.twig_renderer', ['key' => 'ux:icon'])
```
nicolas-grekas: can you please double confirm this works with attribute `key`?
should the $needsIndexes argument of TaggedIterator be set to true?
sorry I don't remember, even thought I might have written the logic years ago :)
also, does the tag really need to be "private"? it should be fine to add the the list to me, that's the whole point of the DI container of Symfony: out of the box extensibility

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -110,10 +110,7 @@ class_exists(AbstractArgument::class) ? new AbstractArgument(\sprintf('Added in
         $container->register('.ux.twig_component.twig.component_runtime', ComponentRuntime::class)
             ->setArguments([
                 new Reference('ux.twig_component.component_renderer'),
-                service_locator([
-                    'ux:icon' => service('.ux_icons.twig_icon_runtime')->nullOnInvalid(),
-                    'ux:map' => service('ux_map.twig_runtime')->nullOnInvalid(),
-                ]),
+                new ServiceLocatorArgument(new TaggedIteratorArgument('.ux.twig_component.twig_renderer')),
             ])
             ->addTag('twig.runtime')
```
nicolas-grekas, line 115: BTW, all these service declarations might better live in a services.php config file to me. For another PR :)

src/Icons/config/services.php
kbond: > also, doesn the tag really need to be "private"?

My idea was to keep this functionality internal until a time we add a proper interface/document its usage. It's not exactly specified in Symfony's BC policy but I wanted to ensure no one uses this tag until then for BC purposes.

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
kbond: Hah, agreed! This comes from a time when the component was much simpler and Ryan and I were just hacking on it.

src/Icons/config/services.php
nicolas-grekas: the thing is that changing the name of a tag is hard from a BC PoV, so I'd recommend using a proper name (and no need for any interface nor doc for now at least, ppl are responsible for wiring their things properly)

kbond: Makes sense!

nicolas-grekas (approved):

PR #2234, opened 2024-10-04
[Turbo][Testing] Fix to the right test

GrinWay:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fixing the test
| License       | MIT

In order to test the Turbo loading you don't have to use `assertSelectorTextContains` method, instead of it use `assertSelectorWillContain` cuz it's async javascript


nicolas-grekas (approved):

smnandre:
Well maybe we should change the preceding paragraph then, because this method is not shared with the DomCrawler api :\

kbond:
Good catch, thanks @GrinWay.

PR #2246, opened 2024-10-06
[Site] Skip yarn & npm install during deploy

smnandre:
> To disable the JavaScript dependencies and asset building, set NO_NPM or NO_YARN to 1 depending on your package manager.

cf https://docs.platform.sh/guides/symfony/integration.html#symfony-build

With the correct export syntax :facepalm-sorry:


smnandre:
Update: 20s gain on deployment time

PR #2251, opened 2024-10-09
[LiveComponent] Remove CSRF tokens - rely on same-origin/CORS instead

### LINKED ISSUE #2177: [LiveComponent] Testing forms causes exeption "There is currently no session available."
I'm trying to test my live component which contains a form. The following code raises an exception of type `Symfony\Component\HttpFoundation\Exception\SessionNotFoundException`.

My sample testcode:
```
$formComponent = $this->createLiveComponent(
        name: ProductionForm::class,
)->actingAs($testUser);

$formComponent->set('form_name.title', 'Loreum Ipsum');
```

The user and the field exist. What else is needed to get the test working?

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | could be
| Issues        | Fixes https://github.com/symfony/ux/issues/2177
| License       | MIT

This PR replaces token-based CSRF-protection by same-origin/CORS policies.
It replaces #2250


src/LiveComponent/src/DependencyInjection/Compiler/OptionalDependencyPass.php
```diff
@@ -33,5 +33,10 @@ public function process(ContainerBuilder $container): void
                 ->addTag(LiveComponentBundle::HYDRATION_EXTENSION_TAG)
             ;
         }
+
+        if (!$container->hasDefinition('test.client')) {
+            $container->getDefinition('ux.live_component.event_subscriber')
+                ->setArgument(1, false);
```
nicolas-grekas, line 39: In test mode, we don't require the custom `Accept` content-type to make testing easier (and not break existing tests).

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -344,7 +337,17 @@ private function createResponse(MountedComponent $mounted): Response
...
+            return false;
+        }
+
+        if ($this->testMode) {
+            return true;
+        }
+
+        // Except when testing, require the correct content-type in the Accept header.
+        // This also acts as a CSRF protection since this can only be set in accordance with same-origin/CORS policies.
+        return \in_array(self::HTML_CONTENT_TYPE, $request->getAcceptableContentTypes(), true);
```
nicolas-grekas, line 348: This is the most important part of this PR: we mandate our custom content-type in the `Accept` header.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -344,7 +335,17 @@ private function createResponse(MountedComponent $mounted): Response
...
     private function isLiveComponentRequest(Request $request): bool
     {
-        return $request->attributes->has('_live_component');
+        if (!$request->attributes->has('_live_component')) {
+            return false;
+        }
+
+        if ($this->testMode) {
+            return true;
+        }
```
smnandre, line 344: This feels a bit weird to have this here... don't you think ?

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -24,33 +24,62 @@
...
+
+            switch (\func_num_args()) {
+                case 6: $route = 'ux_live_component';
+                    // no break
+                case 7: $method = 'post';
+                    // no break
+                case 8: $urlReferenceType = UrlGeneratorInterface::ABSOLUTE_PATH;
+                    // no break
+                default:
+            }
```
smnandre, line 71: Nice

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
smnandre: Could we refactor in a way we test this only once per request ?

onEXHovia:
Many developers, for various reasons(anwser stackoverflow), can add `Access-Control-Allow-Origin: * ` header for all requests its loosens Single Origin Policy (SOP) and allows you to execute request using `XMLHttpRequest `  and setting `Accept` header. Classic way with hidden input, seems like more effective or at least an additional layer against CSRF.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
nicolas-grekas: I'm not sure it's worth it: the way to do that would be to use an attribute on the request object, polluting the list of attributes. The check should be fast enough and rare enough.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
nicolas-grekas: I don't :) It's the way to make the new behavior seamless for existing test suites. The alternative (requiring test cases to all add HTTP_ACCEPT to their calls) looks like forcing boilerplate.

nicolas-grekas:
@onEXHovia thanks for the comment but I'm not sure what you're suggesting for this PR?

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
smnandre: In general, not a big fan of having test-or-dev code in the final classes. 

To be honest, I had the _feeling_ this was one of the "untold conventions" in the symfony codebase.. but if you're good with it, i'm good with it :)

smnandre: Suggestion here, what about renaming something like "controlHeader" or even "acceptHeader" instead ? And set it's default value to "true", instead of having a "false" in the constructor ?

src/Autocomplete/composer.json
```diff
@@ -45,7 +45,6 @@
         "symfony/phpunit-bridge": "^6.3|^7.0",
         "symfony/process": "^6.3|^7.0",
         "symfony/security-bundle": "^6.3|^7.0",
-        "symfony/security-csrf": "^6.3|^7.0",
```
smnandre: Good catch! Ok for you if i extract this in another PR ?

smnandre: --> https://github.com/symfony/ux/pull/2264

onEXHovia:
> @onEXHovia thanks for the comment but I'm not sure what you're suggesting for this PR?

If we consider this implementation and current with hidden token, I see two different scenarios. Without documentation developers may not know that live component depends on SOP in js for protections CSRF and added header Access-Control-Allow-Origin could mean potential CSRF.

On the other hand, implementation with token is more explicit. Developer disable it through the component configuration and understands that could mean potential CSRF.

Discussion in #2250 
> Enabling CSRF by default means enabling the session, making requests stateful.

> CSRF protection is not a "feature", nobody is "using it" in the sense that it's a behavior that shouldn't have any visible effect to any normal workflow.

The same can be said about stateless requests, usually no one tries to make all requests stateless. I don't see any reason to remove the token based implementation, as a solution we could add a global configuration that configures the CSRF behavior for all components.

kbond (approved):
This is really nice! Ryan and I always wanted the possibility for stateless live components but csrf was our Achilles heel!

daFish:
This MR helps with #2177 and fixes the issue. My attempt to fix it in #2254 is no longer needed.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
nicolas-grekas: testMode defaults to true... :)

nicolas-grekas: If you mean changing the default behavior, this makes the tests all red - aka it's a BC break.

nicolas-grekas:
I've just added a note about CSRF / CORS in the doc.

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
smnandre: 👍

kbond (approved):

smnandre:
LGTM :) 

@nicolas-grekas can you update the CHANGELOG in the src/LiveComponent repository ?

smnandre (approved):

nicolas-grekas:
CHANGELOG updated

kbond:
Thank you Nicolas.

PR #2254, opened 2024-10-09
[LiveComponent] add empty request to help with live component tests

### LINKED ISSUE #2177: [LiveComponent] Testing forms causes exeption "There is currently no session available."

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2177 
| License       | MIT

This is an attempt to fix #2177 as now there is a session and a request.

### DIFF
--- a/src/LiveComponent/src/Test/InteractsWithLiveComponents.php
+++ b/src/LiveComponent/src/Test/InteractsWithLiveComponents.php
@@ -13,6 +13,10 @@ namespace Symfony\UX\LiveComponent\Test;
 
 use Symfony\Bundle\FrameworkBundle\KernelBrowser;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\Storage\MockFileSessionStorageFactory;
 use Symfony\UX\TwigComponent\ComponentFactory;
 
 /**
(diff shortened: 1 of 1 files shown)


smnandre:
What i am not sure here is that tests are working in a lot of scenarios... i really would like to understand what is specific in your use case :|

daFish:
I'm also still trying to understand the real cause. Looking at the current tests I get the impression, that they might not reflect a real world use case. But I might totally be wrong.

smnandre:
My point is, if this is (as i kinda suppose) related to CSRF and security.. this may be fixed soon 

See: https://github.com/symfony/ux/pull/2251

daFish:
👍🏻 I'm looking forward to #2251 being merged.

mkrauser:
I've explained everything in https://github.com/symfony/ux/issues/2177#issuecomment-2413211416

That PR would help when testing Components with Form using the ComponentWithFormTrait

mkrauser:
I've tested this PR some more. It seems to only fix one issue. When I try to submit the form, the CSRF-Token is invalid. 
It seems, that a new Session is created for the Submit-Request. The one generated in this PR is not used.

daFish:
> My point is, if this is (as i kinda suppose) related to CSRF and security.. this may be fixed soon
> 
> See: #2251

After testing #2251 and disabling CSRF on my form, I can confirm this solves my issue.

PR #1126, opened 2023-09-19
[LiveComponent] Fix morphing issues with an Alpine.js component, close #984

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Tickets       | Fix #984 
| License       | MIT

This PR is a proposal for #984. 
As I encountered the same issue than @richardhj when using Alpine.js with a LiveComponent, and I wanted to fix it ASAP. 

I've added some checks and errors messages in order to help the dev if needed.

Issue #984, opened 2023-07-02
[Live] Access on Morphdom's `onBeforeElUpdated` callback - Alpine.JS compatibility


richardhj:
I use [Alpine.JS](https://alpinejs.dev/) a lot. It is a nice library and complements Live components perfeclty.

Right now I use it for toggling the state Dropdowns.

What happens, when a Live component re-renders, is, that all the states of the components is discarded. All dropdowns of that particular Live component are open.

I learned, that, the state must be copied from the old DOM to the new DOM. [Details.](https://github.com/alpinejs/alpine/issues/826#issuecomment-755299554)

Concretely, what I need to do in my project, is to extend the `onBeforeElUpdated` callback in morphdom as follows:

https://github.com/symfony/ux/blob/f05e5f1460cdb80ea284b0a2e3a3d8d9094261c8/src/LiveComponent/assets/src/morphdom.ts#L38

```diff
onBeforeElUpdated: (fromEl: Element, toEl: Element) => {
            if (fromEl === rootFromElement) {
                return true;
            }

            // skip special checking if this is, for example, an SVG
            if (fromEl instanceof HTMLElement && toEl instanceof HTMLElement) {

+           if (fromEl.__x) {
+                 window.Alpine.clone(fromEl.__x, toEl);
+            }

                if (childComponentMap.has(fromEl)) {
                    const childComponent = childComponentMap.get(fromEl) as Component;

                    childComponent.updateFromNewElementFromParentRender(toEl);

                    return false;
                }

                // if this field's value has been modified since this HTML was
                // requested, set the toEl's value to match the fromEl
                if (modifiedFieldElements.includes(fromEl)) {
                    setValueOnElement(toEl, getElementValue(fromEl));
                }

                // handle any external changes to this element
                const elementChanges = externalMutationTracker.getChangedElement(fromEl);
                if (elementChanges) {
                    // apply the changes to the "to" element so it looks like the
                    // external changes were already part of it
                    elementChanges.applyToElement(toEl);
                }

                // https://github.com/patrick-steele-idem/morphdom#can-i-make-morphdom-blaze-through-the-dom-tree-even-faster-yes
                if (fromEl.isEqualNode(toEl)) {
                    // the nodes are equal, but the "value" on some might differ
                    // lets try to quickly compare a bit more deeply
                    const normalizedFromEl = cloneHTMLElement(fromEl);
                    normalizeAttributesForComparison(normalizedFromEl);

[...]

PR #2259, opened 2024-10-11
[LiveComponent] Fix morphing for Alpine.js v3

Kobo-one:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        |  /
| License       | MIT

This PR fixes live component compatibility for Alpine v3.

Currently, the live component controller already implements the Alpine morph plugin. This works fine for Alpine v2, but since v3, the `__x` property is no longer used by Alpine. Instead, we can check for `_x_dataStack`.

Also, the morph plugin no longer requires the `__x` property; the DOM element should be used instead.

I made sure the changes support both Alpine v2 and v3.

Original implementation and issue:
- https://github.com/symfony/ux/issues/984
- https://github.com/symfony/ux/pull/1126

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1375,14 +1375,14 @@ function executeMorphdom(rootFromElement, rootToElement, modifiedFieldElements, 
                     return false;
                 }
                 if (fromEl instanceof HTMLElement && toEl instanceof HTMLElement) {
-                    if (typeof fromEl.__x !== 'undefined') {
+                    if (typeof fromEl.__x !== 'undefined' || typeof fromEl._x_dataStack !== 'undefined') {
                         if (!window.Alpine) {
                             throw new Error('Unable to access Alpine.js though the global window.Alpine variable. Please make sure Alpine.js is loaded before Symfony UX LiveComponent.');
                         }
                         if (typeof window.Alpine.morph !== 'function') {
                             throw new Error('Unable to access Alpine.js morph function. Please make sure the Alpine.js Morph plugin is installed and loaded, see https://alpinejs.dev/plugins/morph for more information.');
                         }
-                        window.Alpine.morph(fromEl.__x, toEl);
+                        window.Alpine.morph(fromEl.__x || fromEl, toEl);
                     }
                     if (externalMutationTracker.wasElementAdded(fromEl)) {
                         fromEl.insertAdjacentElement('afterend', toEl);
--- a/src/LiveComponent/assets/src/morphdom.ts
+++ b/src/LiveComponent/assets/src/morphdom.ts
@@ -118,11 +118,11 @@ export function executeMorphdom(
 
                 // skip special checking if this is, for example, an SVG
                 if (fromEl instanceof HTMLElement && toEl instanceof HTMLElement) {
-                    // We assume fromEl is an Alpine component if it has `__x` property.
+                    // We assume fromEl is an Alpine component if it has `__x` (Alpine v2) or `_x_dataStack` (Alpine v3) property.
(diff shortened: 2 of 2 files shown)


Kocal (approved):
LGTM

smnandre (approved):

smnandre:
Thank you @Kobo-one

Kocal (changes requested):
Hey, I tried your patch locally on my app (the one I've used when working on https://github.com/symfony/ux/pull/1126), but I'm having an issue.

Below you will find a video of my app and some interaction with a dropdown which contains a LiveComponent inside. 
**I already use Alpine 3** and it works as I expect things to work, but when I test your patch, then the LiveCompoent disappears:

https://github.com/user-attachments/assets/513a0f44-a4cc-4af5-89db-7652283d2a46

Is there a bug somewhere, or is it the good behavior but I did something wrong in my app?

Thanks!

Kobo-one:
Hi, 

My current test setup uses a reversed logic; an alpine element inside of a LiveComponent. The morph is necessary to keep the alpine state when a LiveComponent re-renders with DOM changes. 

But as your video shows there might be an issue with LiveComponents inside an Alpine element. Oddly enough I also tested this and did not see any weird behaviour. Both with an alpine component surrounding a LiveComponent as well as sharing the same DOM element.

Could you share more details about how your setup is structured in Twig?

**That said, this shows the PR has a breaking change and it should not be merged (yet).**

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -1375,14 +1375,14 @@ function executeMorphdom(rootFromElement, rootToElement, modifiedFieldElements,
...
-                    if (typeof fromEl.__x !== 'undefined') {
+                    if (typeof fromEl.__x !== 'undefined' || typeof fromEl._x_dataStack !== 'undefined') {
                         if (!window.Alpine) {
                             throw new Error('Unable to access Alpine.js though the global window.Alpine variable. Please make sure Alpine.js is loaded before Symfony UX LiveComponent.');
                         }
                         if (typeof window.Alpine.morph !== 'function') {
                             throw new Error('Unable to access Alpine.js morph function. Please make sure the Alpine.js Morph plugin is installed and loaded, see https://alpinejs.dev/plugins/morph for more information.');
                         }
-                        window.Alpine.morph(fromEl.__x, toEl);
+                        window.Alpine.morph(fromEl.__x || fromEl, toEl);
```
smnandre, line 1385: Passing fromEl here and not fromEl.__x seems weird to me, as depending on what the window.Alpine.morph will do, the following of the code can be deeply impacted here :| 

cloneHTMLElement(fromEl) for instance would return wrong results if some morphing is done on the element before

Or am I missing something ?

RobinDev:
I get the same bug. I test the patch from @Kobo-one and it's worked for me.

So, if we use Alpine.js in a live component, it could be buggy actually. 

This patch permits to fix that bug creating a bug when using Alpine.js wrapping a live component.

Did you manage it a different way ?

smnandre:
Maybe if you guys could update this PR with a simple test covering the 2 scenarios (alpine wrapping live and vice versa)

(and when i say _test_, a clean page we can run and test in browsers would do the trick i guess)

What do you think ?

Kocal:
> I get the same bug. I test the patch from @Kobo-one and it's worked for me.
> 
> So, if we use Alpine.js in a live component, it could be buggy actually.
> 
> This patch permits to fix that bug creating a bug when using Alpine.js wrapping a live component.
> 
> Did you manage it a different way ?

@RobinDev since you have the same issue, when I wasn't able to reproduce it in https://github.com/symfony/ux/pull/2259#pullrequestreview-2368996095, could you provide use a reproducer please? 🙏🏻

Kocal:

PR #2269, opened 2024-10-13
Add CI workflow to compute diff between files dist files

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

## 🚨 Before merging, we should drop the commit that modify files (in order to generate the diff table)

This PR is purely internal, and aims to display the `assets/dist/*.{js,css}` files diff between `2.x` and a pull-request. Similar to https://github.com/marketplace/actions/pkg-size-action, but fully internal.

I wanted a tool that display dist files size diff for each pull-request, because I was a bit afraid of changes done in https://github.com/symfony/ux/pull/2160.
 
When a PR is opened, it check dist files between the base branch (`2.x`) and the pull request, and it create a GitHub comment. The comment is created by https://github.com/marocchino/sticky-pull-request-comment, and is automatically updated depending of the check state.
If any diff between dist files, then a table is displayed, with a line per file. It shows the  original size and compressed (gzip and brotli) sizes, and also a difference in %. 

## States

_Currently not working on this repository, but you can see them on https://github.com/Kocal/symfony-ux/pull/1_ 

### 
Capture d’écran 2024-10-13 à 11 45 10
When opening a PR

### When an issue happened
Capture d’écran 2024-10-13 à 11 45 19

### When there is no difference between base and PR
Capture d’écran 2024-10-13 à 11 25 31

### When there is difference between base and PR
image

### DIFF
--- a/.github/generate-dist-files-size-diff.mjs
+++ b/.github/generate-dist-files-size-diff.mjs
@@ -1,0 +1,177 @@
+/**
+ * Generate a markdown table with the difference in size of the dist files between the base and the PR.
+ */
+
+/*
+Usage:
+```shell
+BASE_DIST_FILES='{"src/Autocomplete/assets/dist/controller.js":{"size":15382,"size_gz":3716},"src/Chartjs/assets/dist/controller.js":{"size":2281,"size_gz":771},"src/Cropperjs/assets/dist/controller.js":{"size":1044,"size_gz":475}}' \
+PR_DIST_FILES='{"src/Chartjs/assets/dist/controller.js":{"size":1281,"size_gz":171},"src/Cropperjs/assets/dist/controller.js":{"size":1044,"size_gz":475},"src/Cropperjs/assets/dist/style.min.css":{"size":32,"size_gz":66},"src/Dropzone/assets/dist/controller.js":{"size":3199,"size_gz":816},"src/Map/src/Bridge/Google/assets/dist/foo.js":{"size":3199,"size_gz":816}}' \
+GITHUB_REPOSITORY='symfony/ux' \
+GITHUB_HEAD_REF='my-branch-name' \
+  node .github/generate-dist-files-size-diff.mjs
+```
+ */
+
+if (!process.env.BASE_DIST_FILES) {
+    throw new Error('Missing or invalid "BASE_DIST_FILES" env variable.');
+}
+
+if (!process.env.PR_DIST_FILES) {
+    throw new Error('Missing or invalid "PR_DIST_FILES" env variable.');
+}
+
+if (!process.env.GITHUB_REPOSITORY) {
+    throw new Error('Missing or invalid "GITHUB_REPOSITORY" env variable.');
+}
+
+if (!process.env.GITHUB_HEAD_REF) {
+    throw new Error('Missing or invalid "GITHUB_HEAD_REF" env variable.');
+}
+
+/**
+ * Adapted from https://gist.github.com/zentala/1e6f72438796d74531803cc3833c039c?permalink_comment_id=4455218#gistcomment-4455218
+ * @param {number} bytes
+ * @param {number} digits
+ * @returns {string}
+ */
+function formatBytes(bytes, digits = 2) {
+    if (bytes === 0) {
+        return '0 B';
+    }
+    const sizes = [`B`, 'kB', 'MB'];
+    const i = Math.floor(Math.log(bytes) / Math.log(1024));
+
+    return parseFloat((bytes / Math.pow(1024, i)).toFixed(digits)) + ' ' + sizes[i];
+}
+
+/**
+ * @param {number} from
+ * @param {number} to
+ * @returns {number}
+ */
+function computeDiffPercent(from, to) {
+    if (from === to) {
+        return 0;
+    }
+
+    return Math.round((from - to) / from * -100);
+}
+
+/**
+ * @param {number} percent
+ * @returns {string}
+ */
+function formatDiffPercent(percent) {
+    return percent > 0 ? `+${percent}% 📈` : percent < 0 ? `${percent}% 📉` : `${percent}%`;
+}
+
+export function main() {
+    const repoUrl = `https://github.com/${process.env.GITHUB_REPOSITORY}`;
+    /** @type {Record<string, {size: number, size_gz: number}>} */
+    const base = JSON.parse(process.env.BASE_DIST_FILES);
+    /** @type {Record<string, {size: number, size_gz: number}>} */
+    const pr = JSON.parse(process.env.PR_DIST_FILES);
+    let output = '<h1>📊 Dist packagesFiles size difference</h1>\n\n';
+
+    /**
+     * @type {Map<string, {
+     *     meta: {
+     *         packageName: string,
+     *         bridgeName: string,
+     *         url: string,
+     *     },
+     *     files: Set<{
+     *         state: 'added' | 'removed' | 'changed',
+     *         before: {size: number, sizeGz: number},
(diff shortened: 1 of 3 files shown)


Kocal:
It looks like there are permissions issues, even after added `permissions.pull-requests: write` like [mentioned in the documentation](https://github.com/marocchino/sticky-pull-request-comment#basic):
image

Even if I use `permissions: 'write-all'` it does not work either:
image

I suspect a configuration at organization-level...? 

@kbond do you think you can do something about that please? 🙏🏻 

/cc @nicolas-grekas or maybe you Nicolas since you setup the same action on Symfony recipes repositories :) 

Thanks!

.github/generate-dist-files-size-diff.mjs
```diff
@@ -0,0 +1,118 @@
...
+
+    if (files.size === 0) {
+        output += 'ℹ️ No difference in dist files.\n';
+        return output;
+    }
+
+    output += 'Thanks for the PR! Here is the difference in size of the dist files between the base and the PR.\n';
+    output += 'Please review the changes and make sure they are expected.\n\n';
+    output += `<table>
+    <thead><tr><th>File</th><th>Diff (B)</th><th>Diff (%)</th></tr></thead>
```
javiereguiluz: I would simplify the output to display something more similar to pck-size action, which I think it's very readable. Something like this:

```js
    <thead><tr><th>File</th><th>Before (Size / Gzip)</th><th>After (Size / Gzip)</th></tr></thead>
    <tbody>`;
    for (const [file, details] of files.entries()) {
        output += `<tr>
        <td><code>${file}</code> ${details.state === 'added' ? '(new)' : (details.state === 'removed' ? '(deleted)' : '')}</td>
        <td>
            <code>${formatBytes(base[file]?.size || 0)}</code> / <code>${formatBytes(base[file]?.size_gz || 0)}</code>
        </td>
        <td align="right">
            <sup>${formatDiffPercent(details.diff_percent_size)}</sup> (<code>${formatBytes(pr[file]?.size || 0)}</code>)
            / <sup>${formatDiffPercent(details.diff_percent_size_gz)}</sup> <code>${formatBytes(pr[file]?.size_gz || 0)}</code>)
        </td>
    </tr>
```

Kocal: This is what I've did first, but that was ugly because our filenames are very very long (but I think I can do something about it) and it was breaking the sizes into multiple lines.

I will rework that part, thanks

kbond:
Shouldn't this job only run if a dist file has been changed?

Kocal:
> Shouldn't this job only run if a dist file has been changed?

Fixed by using the following config:
```yaml
on:
  pull_request:
    paths:
      - 'src/*/assets/dist/**'
      - 'src/*/src/Bridge/*/assets/dist/**'

```

I will push soon, thanks

.github/generate-dist-files-size-diff.mjs
Kocal: I re-used your code and this is what happens:
image

As you can see, because of line breaks the display is not very nice :/

Kocal: Also, it does not even contain the brotli-compressed size of dist files, which I really want to keep (since it's more efficient than gzip and zstd)

Kocal: I reworked the table rendering to display info in a smarter way:
image


You don't have useless data like `0 B` or `-+100%` displayed anymore, instead it shows you if the file has been added or removed (without displaying size) and the delta (%) only if the file has been changed.
I've regrouped files by packages, and each packages and files has a link.

Kocal:
I've changed the table rendering after @javiereguiluz's comment, it now looks like this:
image

We still need to find out why there is this `Resource not accessible by integration` error, and we are good IMO

.github/generate-dist-files-size-diff.mjs
```diff
@@ -0,0 +1,182 @@
...
+ * @param {number} from
+ * @param {number} to
+ * @returns {number}
+ */
+function computeDiffPercent(from, to) {
+    if (from === to) {
+        return 0;
+    }
+
+    return Number(((from - to) / to * -100).toFixed(1));
```
jmsche: I think you should divide by `from`, not `to`, to compare with the previous file size?

javiereguiluz:
@Kocal I like a lot what you are doing here. Thanks.

Some additional comments ... but I could be wrong, so feel free to ignore them:

* I'd remove Brotli because it's just "the same as Gzip but a bit better", so it doesn't add any relevant information
* I'd round percentages and remove decimals to make them easier to understand: `-4%` is better to me than `-3.9%` and `+74%` is better than `+74.3%`
* I'd put the percentages first, because it's the most relevant info by far. E.g. I don't care that the new file size is `5.56 KB`, I just want to quickly check that it's a `-27%` file size change.

Thanks!

.github/generate-dist-files-size-diff.mjs
Kocal: I don't think so:
- if the file was previously `1000` bytes and is now `1500` bytes, I want to display an increase of `33,33%`
- in the opposite, if the file was previously `1500` bytes and is now `1000` bytes, I want to display a decrease of `50%`
image

Or did I miss something?

jmsche: To me it's the other way:

- previously it was 1000 bytes, now 1500, so it's +50% compared to the previous value
- and if it was 1500 bytes and now 1000, it's -33,33% compared to the previous value

javiereguiluz: @jmsche is right. Going from 1,000 to 1,500 is a +50%, not a +33% _(you are adding 500 to 1,000 ... so you are adding 50% of the original value)_

Kocal:
Thanks Javier, the more we speak about Brotli and the more I think we can remove it (btw [pkg-size-action](https://github.com/marketplace/actions/pkg-size-action) can display Brotli sizes).  Gzip is the most common compression standard for the moment, while Brotli is — I believe —  still a "niche" thing. Brotli users know that their files will be smaller than gzipped files.

I agree for percentage rounding.

When it comes to putting percentages first, I don't really agree with you. Yes, percentage differences are important, but when the browser goes to download a resource, it's the size of the resource that matters, not the percentage difference.
If a file is reduced from 1 MB to 500 KB, we gain 50%, but more importantly we save 500 KB (and that's a really huge gain). If a file is reduced from 10 Kb to 5 Kb, you also gain 50%, but you only save 5KB (which still has an impact, but a much smaller one).
But then, your point of view is valid too and I can understand it, I think here it's two different ways of thinking :)

.github/generate-dist-files-size-diff.mjs
Kocal: ... 🤦🏻 

Thanks guys 😅

Kocal:
Thanks for your reviews, the diff table now looks like this:
image

javiereguiluz:
Looks very nice!! Thanks a lot Hugo.

kbond:
This looks great!

@Kocal:
1. Can you revert the temporary commit?
2. The proper permissions are still todo?

Kocal:
@kbond yes there are still permission issues, and I can revert the temporary commit (when permissions will be fixed, otherwise we won't see the comment)

Kocal:
I've reworked the workflow in two dedicated workflows, hoping for permissions issues to be gone.
We won't have comment before the diff is generated, or if a failure happened.

I reverted the commit that modify dist files for testing purposes. I'm merging ASAP

PR #2270, opened 2024-10-14
Remove stimulus-bundle unnecessary warning on documentation

feymo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | N/A
| License       | MIT

Update LiveComponents documentation by removing the warning requiring Stimulus bundle (as it is now part of the package requirements)

### DIFF
--- a/src/Chartjs/doc/index.rst
+++ b/src/Chartjs/doc/index.rst
@@ -8,10 +8,6 @@ It is part of `the Symfony UX initiative`_.
 Installation
 ------------
 
-.. caution::
-
-    Before you start, make sure you have `StimulusBundle configured in your app`_.
-
 Install the bundle using Composer and Symfony Flex:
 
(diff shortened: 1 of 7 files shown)


Kocal:
Merci Félix :eyes:

I've double-checked our packages and the LiveComponent is not the only one to have a hard requirement on `symfony/stimulus-bundle` and have a document telling people to install/configure the Stimulus Bundle. The other packages are:
- Chart
- Map
- React
- Svelte
- Vue
- Turbo

Can you remove the documentation from those packages as well please? 🙏🏻 

/cc @smnandre we spoke about that some weeks/months ago :)

feymo:
Thank you @Kocal for the double-check 👍🏼 
Packages you've mentioned have also been updated :)

Kocal:
Thanks Félix :)

PR #2277, opened 2024-10-15
Add ability to add inputProps in component inside eventListener

tito10047:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #2276
| License       | MIT

Add ability to modity InputProps inside EventListener

```php
#[AsEventListener(priority: 100)]
class UxIconListener
{
    public function __invoke(PreCreateForRenderEvent $event): void
    {
        if ('ux:icon' !== strtolower($event->getName())) {
            return;
        }
        $attributes = $event->getInputProps();

        if ($attributes['name']!=="my-global-icon"){
            return;
        }
        
        $attributes["class"] = "icon view";
        $attributes["title"]= "this item is public";
        
        $event->setInputProps($attributes);

    }
}
```

### DIFF
--- a/src/TwigComponent/src/Event/PreCreateForRenderEvent.php
+++ b/src/TwigComponent/src/Event/PreCreateForRenderEvent.php
@@ -49,6 +49,12 @@ final class PreCreateForRenderEvent extends Event
         return $this->inputProps;
     }
 
+	public function setInputProps(array $inputProps): self {
+		$this->inputProps = $inputProps;
+
+		return $this;
+	}
+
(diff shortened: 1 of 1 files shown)


smnandre:
This event is not designed for this usage, but to return HTML and bypass the system.

You can achieve similar goal by:
* using [PreRenderEvent](https://symfony.com/bundles/ux-twig-component/current/index.html#prerenderevent)
* decorating the IconRenderer

PR #2282, opened 2024-10-18
[Autocomplete] Only activate clear button plugin when there is a placeholder option

### LINKED ISSUE #402: [Autocomplete] required choice types end up not being required.

Qronicle:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | maybe?
| Issues        | Fix #402 
| License       | MIT

Only add the `clear_button` plugin to TomSelect if the form element has no placeholder option.

Notes:
- I defined 'having a placeholder' as the select having an option with an empty value, as that is what TomSelect also seems to do
- I'm not 100% sure about whether this needs a test or how to even run tests locally

### DIFF
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -227,8 +227,9 @@ _default_1_instances = new WeakSet(), _default_1_getCommonConfig = function _def
     const plugins = {};
     const isMultiple = !this.selectElement || this.selectElement.multiple;
     if (!this.formElement.disabled && !isMultiple) {
-        const firstOption = this.selectElement.options[0];
-        if (firstOption && firstOption.value === '') {
+        const placeholderOptions = Array.from(this.selectElement.options)
+            .filter((option) => option.value === '');
+        if (placeholderOptions.length) {
             plugins.clear_button = { title: '' };
         }
     }
--- a/src/Autocomplete/assets/dist/controller.js
+++ b/src/Autocomplete/assets/dist/controller.js
@@ -15,7 +15,7 @@ LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
 OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
 PERFORMANCE OF THIS SOFTWARE.
 ***************************************************************************** */
-/* global Reflect, Promise, SuppressedError, Symbol */
+/* global Reflect, Promise, SuppressedError, Symbol, Iterator */
 
 
 function __classPrivateFieldGet(receiver, state, kind, f) {
@@ -227,7 +227,10 @@ _default_1_instances = new WeakSet(), _default_1_getCommonConfig = function _def
     const plugins = {};
     const isMultiple = !this.selectElement || this.selectElement.multiple;
     if (!this.formElement.disabled && !isMultiple) {
-        plugins.clear_button = { title: '' };
+        const firstOption = this.selectElement.options[0];
+        if (firstOption && firstOption.value === '') {
+            plugins.clear_button = { title: '' };
+        }
     }
     if (isMultiple) {
         plugins.remove_button = { title: '' };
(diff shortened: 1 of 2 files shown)


src/Autocomplete/assets/dist/controller.js
```diff
@@ -227,7 +227,11 @@ _default_1_instances = new WeakSet(), _default_1_getCommonConfig = function _def
     const plugins = {};
     const isMultiple = !this.selectElement || this.selectElement.multiple;
     if (!this.formElement.disabled && !isMultiple) {
-        plugins.clear_button = { title: '' };
+        const placeholderOptions = Array.from(this.selectElement.options)
+            .filter((option) => option.value === '');
```
smnandre, line 231: Maybe use `Array.from(...).find()` here ?
(especially as the placeholder will be first element in most of the cases)

smnandre: Even `.some()` that will return you the boolean you need L232 ?

smnandre: (yeah.. in the .ts file 😅 )

smnandre:
Form the monorepository root

### Install dependencies
```
yarn install 
```

### Code style
```
yarn run lint
yarn run format
```

### Run Tests
```
yarn run test
```

### Build dist packages 
```
yarn run test
```

(you need to commit both `src/Autocomplete/assets/src/**` and `src/Autocomplete/assets/dist/**` files in this PR)

Please ask if you want/need anymore information :)

Kocal:
Status: Needs work

Qronicle:
Sorry guys, I somehow missed the review notifications. Will try and clear some time to dive back into this.

smnandre:
No worry, we cannot criticize anyone on time things ....  😅 


(well, we cannot criticize anyone on anything just to be clear 😄 )

LucileDT:
@Qronicle do you have a bit of time to fix your PR? :eyes:

Kocal:

PR #2295, opened 2024-10-23
[AutoComplete] Note form requirements

### LINKED ISSUE #2252: [AutoComplete] Weak dependency on symfony/form
I followed these steps to create a bug reproducer:

1. https://symfony.com/doc/5.x/contributing/code/reproducer.html#reproducing-complex-bugs
2. https://symfony.com/bundles/StimulusBundle/current/index.html
3. https://symfony.com/bundles/ux-autocomplete/current/index.html

And I got blocked when adding `class AnyForm extends AbstractType`: without `symfony/form` installed, `AbstractType` is not available.

So I had to install `symfony/form`, and this may not be obvious for newcomers.

Possible solutions:

- add `composer require symfony/form` in the documentation of `ux-autocomplete`
- declare that `symfony/ux-autocomplete` depends on `symfony/form` in https://github.com/symfony/ux/blob/2.x/src/Autocomplete/composer.json`

MrYamous:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #2252 
| License       | MIT

### DIFF
--- a/src/Autocomplete/doc/index.rst
+++ b/src/Autocomplete/doc/index.rst
@@ -33,8 +33,9 @@ needed if you're using AssetMapper):
 Usage in a Form (without Ajax)
 ------------------------------
 
-Any ``ChoiceType`` or ``EntityType`` can be transformed into a
-Tom Select-powered UI control by adding the ``autocomplete`` option:
+If you're using `Symfony Form`_ any ``ChoiceType`` or ``EntityType`` can be
+transformed into a Tom Select-powered UI control by adding the ``autocomplete``
+option:
 
 .. code-block:: diff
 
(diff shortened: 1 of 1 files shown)


smnandre:
Thank you for this PR!

Most users likely already have the Form component installed when they read this documentation, or they’ve just installed Symfony following the installation guide, which includes the component by default. So, I’m not sure it’s necessary to include a "composer require ..." block here.

Instead, we could add a link to the Symfony Form documentation within the text. What do you think?

MrYamous:
> Thank you for this PR!
> 
> Most users likely already have the Form component installed when they read this documentation, or they’ve just installed Symfony following the installation guide, which includes the component by default. So, I’m not sure it’s necessary to include a "composer require ..." block here.
> 
> Instead, we could add a link to the Symfony Form documentation within the text. What do you think?

I fully agree with your point, that's probably overkill and adding a bit of noise to the doc. 
I've updated to remove the `composer require` part and insert Symfony Form link to the original sentence

smnandre (approved):
Perfect, thank you @MrYamous !

javiereguiluz:
Thanks Matthieu!

PR #2298, opened 2024-10-24
[Turbo] Support custom TurboStreamResponse actions

DRaichev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

The current implementation of TurboStream & TurboStreamResponse has no generic action (only the predefined turbo actions) and what's more the class is marked as final, which does not allow adding support for more actions. I think the wrap function should be public or have a public function "custom" that exposes the functionality (I've opted for the latter)

I really like this library: https://github.com/marcoroth/turbo_power
It adds additional simple actions. For example I use set/delete attributes to enable/disable UI elements

This change will allow people to use this or similar libraries, or even build their own custom actions if they want to.

There is no impact to the rest of the functionality or DX

There is a minor change to the wrap function, it now accepts an array of attributes instead of a string. It builds the string from the array of attributes, and adds a leading space. This aims to prevent errors, as the previous implementation needs the $attr argument to start with a blank space otherwise it would produce invalid html

### DIFF
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 -   Add `<twig:Turbo:Stream>` component
 -   Add `<twig:Turbo:Frame>` component
+-   Add support for custom actions in `TurboStream` and `TurboStreamResponse`
 
 ## 2.21.0
 
--- a/src/Turbo/src/Helper/TurboStream.php
+++ b/src/Turbo/src/Helper/TurboStream.php
@@ -86,6 +86,34 @@ final class TurboStream
         return \sprintf('<turbo-stream action="refresh" request-id="%s"></turbo-stream>', htmlspecialchars($requestId));
     }
 
+    /**
+     * Custom action and attributes.
+     *
+     * Set boolean attributes (e.g., `disabled`) by providing the attribute name as key with `null` as value.
+     *
+     * @param array<string, string|int|float|null> $attr
+     */
+    public static function action(string $action, string $target, string $html, array $attr = []): string
+    {
+        if (\array_key_exists('action', $attr) || \array_key_exists('targets', $attr)) {
+            throw new \InvalidArgumentException('The "action" and "targets" attributes are reserved and cannot be used.');
+        }
+
+        $attrString = '';
+        foreach ($attr as $key => $value) {
+            $key = htmlspecialchars($key);
+            if (null === $value) {
+                $attrString .= \sprintf(' %s', $key);
+            } elseif (\is_int($value) || \is_float($value)) {
+                $attrString .= \sprintf(' %s="%s"', $key, $value);
+            } else {
+                $attrString .= \sprintf(' %s="%s"', $key, htmlspecialchars($value));
+            }
+        }
+
+        return self::wrap(htmlspecialchars($action), $target, $html, $attrString);
+    }
+
     private static function wrap(string $action, string $target, string $html, string $attr = ''): string
     {
         return \sprintf(<<<EOHTML
--- a/src/Turbo/src/TurboStreamResponse.php
+++ b/src/Turbo/src/TurboStreamResponse.php
@@ -104,4 +104,20 @@ class TurboStreamResponse extends Response
 
         return $this;
     }
+
+    /**
+     * Custom action and attributes.
+     *
+     * Set boolean attributes (e.g., `disabled`) by providing the attribute name as key with `null` as value.
+     *
+     * @param array<string, string|int|float|null> $attr
+     *
+     * @return $this
+     */
+    public function action(string $action, string $target, string $html, array $attr = []): static
+    {
+        $this->setContent($this->getContent().TurboStream::action($action, $target, $html, $attr));
+
+        return $this;
+    }
 }
--- a/src/Turbo/tests/Helper/TurboStreamTest.php
+++ b/src/Turbo/tests/Helper/TurboStreamTest.php
@@ -76,4 +76,32 @@ class TurboStreamTest extends TestCase
             TurboStream::refresh('a"b')
         );
     }
+
+    public function testCustom(): void
+    {
+        $this->assertSame(<<<EOHTML
+            <turbo-stream action="customAction" targets="some[&quot;selector&quot;]" someAttr="someValue" boolAttr intAttr="0" floatAttr="3.14">
+                <template><div>content</div></template>
+            </turbo-stream>
+            EOHTML,
+            TurboStream::action('customAction', 'some["selector"]', '<div>content</div>', ['someAttr' => 'someValue', 'boolAttr' => null, 'intAttr' => 0, 'floatAttr' => 3.14])
+        );
+    }
+
+    /**
+     * @dataProvider customThrowsExceptionDataProvider
+     *
+     * @param array<string, string|int|float|null> $attr
+     */
+    public function testCustomThrowsException(string $action, string $target, string $html, array $attr): void
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        TurboStream::action($action, $target, $html, $attr);
+    }
+
+    public static function customThrowsExceptionDataProvider(): \Generator
+    {
+        yield ['customAction', 'some["selector"]', '<div>content</div>', ['action' => 'someAction']];
+        yield ['customAction', 'some["selector"]', '<div>content</div>', ['targets' => 'someTargets']];
+    }
 }


setineos:
I think it should also work for `<twig:Turbo:Stream:*>` components.

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,28 @@ public static function refresh(?string $requestId = null): string
...
+     */
+    public static function custom(string $action, string $target, string $html, array $attr = []): string
     {
+        return self::wrap($action, $target, $html, $attr);
+    }
+
+    /**
+     * @param array<string> $attr
+     */
+    private static function wrap(string $action, string $target, string $html, array $attr = []): string
```
nicolas-grekas: we don't need that method anymore, custom is basically a public version of wrap

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,28 @@ public static function refresh(?string $requestId = null): string
...
+    {
+        // Join array elements with a space and prepend a leading space
+        $atrrString = empty($attr) ? '' : ' '.implode(' ', $attr);
```
nicolas-grekas: the API for $attr means people are responsible for encoding their attributes, is that what we want?

nicolas-grekas:
```suggestion
        $atrrString = $attr ? ' '.implode(' ', $attr) : '';
```

nicolas-grekas:
> I think it should also work for <twig:Turbo:Stream:*> components.

This can be done by adding a generic `<twig:Turbo:Stream>` component maybe.

src/Turbo/src/Helper/TurboStream.php
DRaichev: > the API for $attr means people are responsible for encoding their attributes, is that what we want?

Not sure, I don't mind either way, but I think it should be documented if users are responsible for encoding.
What do you think ?

DRaichev: >  $atrrString = $attr ? ' '.implode(' ', $attr) : '';

I'm not sure I like this. I generally don't like implicit checks, although for empty array it makes sense.

nicolas-grekas: I think I'd prefer encoding on our side, and $attr be a name=>value list
about the check, empty() is the worst :)

DRaichev:
@nicolas-grekas  
I just realised if you pass "action" or "targets" into the attributes array those will conflict with the existing attributes.
I think we should throw an exception if those are present

smnandre:
With the incoming [`<twig:Turbo:Frame>`](https://github.com/symfony/ux/pull/2303) and [`<twig:Turbo:Stream>`](https://github.com/symfony/ux/pull/2302), do you think we still need this features ?

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -61,15 +61,15 @@ public static function remove(string $target): string
...
+        return self::custom('before', $target, $html);
     }
 
     /**
      * Inserts after the element(s) designated by the target CSS selector.
      */
     public static function after(string $target, string $html): string
     {
-        return self::wrap('after', $target, $html);
+        return self::custom('after', $target, $html);
```
smnandre: I think we should revert these changes (other ones too in existing methods)

(wrap worked better as method name here)

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,34 @@ public static function refresh(?string $requestId = null): string
         return \sprintf('<turbo-stream action="refresh" request-id="%s"></turbo-stream>', htmlspecialchars($requestId));
     }
 
-    private static function wrap(string $action, string $target, string $html, string $attr = ''): string
+    /**
+     * Custom action and attributes.
+     *
+     * @param array<string, string|int|null> $attr
+     */
+    public static function custom(string $action, string $target, string $html, array $attr = []): string
```
smnandre: I'd rather keep the wrap method, and have this new method use it 

(prepare and escape the $attr string and pass it to self::wrap())

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,34 @@ public static function refresh(?string $requestId = null): string
...
     {
+        if (\array_key_exists('action', $attr) || \array_key_exists('targets', $attr)) {
+            throw new \InvalidArgumentException('The "action" and "targets" attributes are reserved and cannot be used.');
+        }
+
+        $attr['action'] = $action;
+        $attr['targets'] = $target;
+        $attr = array_merge(['action' => $action, 'targets' => $target], $attr);
```
smnandre: Not needed if you call the `wrap` method

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,34 @@ public static function refresh(?string $requestId = null): string
...
+
+        $attrString = '';
+        foreach ($attr as $key => $value) {
+            if (null === $value) {
+                $attrString .= \sprintf(' %s', $key);
+            } else {
+                $attrString .= \sprintf(' %s="%s"', $key, \is_string($value) ? htmlspecialchars($value) : $value);
```
smnandre: If $value is a Stringable object, this can be dangerous.

Special cases for boolean should be added too probably.

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,34 @@ public static function refresh(?string $requestId = null): string
...
+        }
+
+        $attr['action'] = $action;
+        $attr['targets'] = $target;
+        $attr = array_merge(['action' => $action, 'targets' => $target], $attr);
+
+        $attrString = '';
+        foreach ($attr as $key => $value) {
+            if (null === $value) {
+                $attrString .= \sprintf(' %s', $key);
```
smnandre, line 106: Don't we need to encode keys here?

src/Turbo/src/TurboStreamResponse.php
```diff
@@ -104,4 +104,16 @@ public function refresh(?string $requestId = null): static
 
         return $this;
     }
+
+    /**
+     * @param array<string> $attr
+     *
+     * @return $this
+     */
+    public function custom(string $action, string $target, string $html, array $attr = []): static
```
smnandre:
```suggestion
    public function action(string $action, string $target, string $html, array $attr = []): static
```

?

src/Turbo/src/TurboStreamResponse.php
```diff
@@ -104,4 +104,16 @@ public function refresh(?string $requestId = null): static
 
         return $this;
     }
+
+    /**
+     * @param array<string> $attr
```
smnandre:
```suggestion
     * @param array<string, scalar> $attr
```

?

DRaichev:
> With the incoming [`<twig:Turbo:Frame>`](https://github.com/symfony/ux/pull/2303) and [`<twig:Turbo:Stream>`](https://github.com/symfony/ux/pull/2302), do you think we still need this features ?

I think they are complementary. I'd rather use the twig component syntax only in templates and stick to this helper on the php side, plus it's better to have consistent functionality in both, especially for people who dislike the twig component syntax.

src/Turbo/src/TurboStreamResponse.php
DRaichev: Scalar includes bool, and I think passing null as value for boolean attributes is better, as per
https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

So I think `array<string, string|int|float|null>` is the correct type.

src/Turbo/src/Helper/TurboStream.php
DRaichev: I don't really see a case where the key would have special characters, but it doesn't hurt, so better safe than sorry

src/Turbo/src/Helper/TurboStream.php
DRaichev: I've changed it so it covers the types specified in the typehint and would fail if an object is passed

smnandre:
Looks good to me!

Could you check the type of values and throw if not expected (a non-stringable object) ?

DRaichev:
> Looks good to me!
> 
> Could you check the type of values and throw if not expected (a non-stringable object) ?

I believe this is going to throw:
```
foreach ($attr as $key => $value) {
    $key = htmlspecialchars($key);
    if (null === $value) {
        $attrString .= \sprintf(' %s', $key);
    } elseif (\is_int($value) || \is_float($value)) {
        $attrString .= \sprintf(' %s="%s"', $key, $value);
    } else {
        $attrString .= \sprintf(' %s="%s"', $key, htmlspecialchars($value)); // htmlspecialchars expect string so will throw here
    }
}
```

Or do you mean somewhere else ?

DRaichev:
> > Looks good to me!
> > Could you check the type of values and throw if not expected (a non-stringable object) ?
> 
> I believe this is going to throw:
> 
> ```
> foreach ($attr as $key => $value) {
>     $key = htmlspecialchars($key);
>     if (null === $value) {
>         $attrString .= \sprintf(' %s', $key);
>     } elseif (\is_int($value) || \is_float($value)) {
>         $attrString .= \sprintf(' %s="%s"', $key, $value);
>     } else {
>         $attrString .= \sprintf(' %s="%s"', $key, htmlspecialchars($value)); // htmlspecialchars expect string so will throw here
>     }
> }
> ```
> 
> Or do you mean somewhere else ?

So this is ready for merge then, right

smnandre:
After some thinking, why don't we rename the `custom($action)`  method in `action($name, .... ... ... .. )`  ?

DRaichev:
> After some thinking, why don't we rename the `custom($action)` method in `action($name, .... ... ... .. )` ?

Yeah, I was thinking about this when I renamed the one in TurboStreamResponse
It should be either custom in both or action in both.

I don't really have a preference, both work, so I leave the choice to you

smnandre:
Let's use action in both then (sorry for the late comment)

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,45 @@ public static function refresh(?string $requestId = null): string
         return \sprintf('<turbo-stream action="refresh" request-id="%s"></turbo-stream>', htmlspecialchars($requestId));
     }
 
+    /**
+     * Custom action and attributes.
+     * When passing attributes use null value to for boolean attributes (e.g. disabled).
```
smnandre:
```suggestion
     * 
     * Set boolean attributes (e.g., `disabled`) by providing the attribute name as key with `null` as value.
```

src/Turbo/src/Helper/TurboStream.php
```diff
@@ -86,12 +86,45 @@ public static function refresh(?string $requestId = null): string
...
             <turbo-stream action="%s" targets="%s"%s>
                 <template>%s</template>
             </turbo-stream>
-            EOHTML, $action, htmlspecialchars($target), $attr, $html);
+            EOHTML,
+            $action,
+            htmlspecialchars($target),
+            $attr,
+            $html
+        );
```
smnandre: Let's revert this change, as the previous one feels more readable

smnandre (commented):
Last couple of minor suggestions..

Could you update the CHANGELOG in the src/Turbo repository ? 

And let's merge this :)

DRaichev:
Ready to merge

smnandre (approved):
Thank you @DRaichev !

src/Turbo/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # CHANGELOG
+## 2.22.0
+
+-   Add support for custom actions in `TurboStream` and `TurboStreamResponse`
```
smnandre: Arf, saw this just too late... can you move it a couple lines down, there was already a 2.22.0 block 😅

DRaichev: Ah, yes, it changed when synced with the upstream.

DRaichev:
Updated the changelog, now ready to merge

smnandre (approved):
Perfect!

DRaichev:
@smnandre Hey, can we get this merged ?

I am in the process of replacing the usage of the old syntax at work. I've already replaced all basic cases, so 
now I'm waiting for v2.22 to do the `<twig:Turbo:Frame/>` and hopefully we can get this as well so I can do the instances of custom turbo actions.

Kocal:
Hey, I rebasing your PR and will merge it in the next minutes.

Kocal:
Thank you @DRaichev.

smnandre:
@DRaichev we will probably release a 2.22 in the coming weeks, you can use it with the "2.x-dev" in the meanwhile! Thanks for this PR 😄

DRaichev:
@Kocal  @smnandre Thanks for the speedy reaction. And thank you for all the work you're doing on this project. 
I used to hate to work on UX stuff, but thanks to Turbo and Symfony UX I realised you can have great UX without the convoluted mess. Keep up the good work.

smnandre:
Thank you for such a kind message—it really means a lot!

PR #2302, opened 2024-10-25
[Turbo] Add generic `<Turbo:Stream>` component

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | https://github.com/symfony/ux/pull/2298#issuecomment-2434799250
| License       | MIT

I added a generic Turbo Stream component. To use it, you can do this:

```twig
<twig:Turbo:Stream action="turbo_frame_reload" targets="#count-post" />
```

The rendering is:

```twig
<turbo-stream action="turbo_frame_reload" targets="#count-post"></turbo-stream>
```

### DIFF
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.22.0
+
+-   Add `<twig:Turbo:Stream>` component
+
 ## 2.21.0
 
 -   Add `Helper/TurboStream::append()` et al. methods
--- a/src/Turbo/templates/components/Stream.html.twig
+++ b/src/Turbo/templates/components/Stream.html.twig
@@ -1,0 +1,3 @@
+{% props action -%}
+
+<turbo-stream action="{{ action }}" {{- attributes }}></turbo-stream>


Kocal (approved):
Can you also update the CHANGELOG here please? Thanks

smnandre:
Is there any advantage here (apart from the syntax) ?

Because 

```html
<turbo-frame foo bar>foo bar</turbo-frame>
```

and

```twig
<twig:Turbo:Frame foo bar>
foo bar
</twig:Turbo:Frame>
```

have a real difference in term of performance.

But i get the target to cover all bases of turbo with one syntax. 

Thinking outloud here.. not sure what to do.

I wonder if we should not make a search/replace here, instead of calling an embedding component 🤷 
But then, it would be difficult to explain the differences in scope

seb-jean:
> Can you also update the CHANGELOG here please? Thanks

Done

seb-jean:
> Is there any advantage here (apart from the syntax) ?
> 
> Because
> 
> ```
> <turbo-frame foo bar>foo bar</turbo-frame>
> ```
> 
> and
> 
> ```
> <twig:Turbo:Frame foo bar>
> foo bar
> </twig:Turbo:Frame>
> ```
> 
> have a real difference in term of performance.
> 
> But i get the target to cover all bases of turbo with one syntax.
> 
> Thinking outloud here.. not sure what to do.
> 
> I wonder if we should not make a search/replace here, instead of calling an embedding component 🤷 But then, it would be difficult to explain the differences in scope

This comment is rather related to this PR I think no?: https://github.com/symfony/ux/pull/2303

smnandre (approved):

Kocal:
Thank you @seb-jean.

PR #2304, opened 2024-10-26
[StimulusBundle] Check controllers source files for laziness

### LINKED ISSUE #10: Comment "stimulusFetch: 'lazy'" is dropped from Stimulus Controller
Hey :)

Given my Stimulus controller:
```js
import { redo as commandRedo, undo as commandUndo } from "@codemirror/commands";
import { markdown } from "@codemirror/lang-markdown";
import { Compartment, EditorSelection } from "@codemirror/state";
import { githubDark } from "@fsegurai/codemirror-theme-github-dark";
import { githubLight } from "@fsegurai/codemirror-theme-github-light";
import { Controller } from "@hotwired/stimulus";
import { EditorView, basicSetup } from "codemirror";
import { codeLanguages } from "../codemirror/code_languages.js";
import { lightDarkThemeSwitcher } from "../codemirror/light_dark_theme_switcher.js";

/* stimulusFetch: 'lazy' */
    static targets = ["textarea"];
    static classes = ["editor"];

    /* ... */
}
```

After being minified, the comment `/* stimulusFetch: 'lazy' */` disappear from the source code:
image

Which load the controller (even if not needed since there are no `data-controller="markdown-editor"` on the page) and load all the dependencies (a lot): 
image

MatTheCat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix https://github.com/sensiolabs/minify-bundle/issues/10
| License       | MIT

The StimulusBundle allows to mark a controller as lazy using a `/* stimulusFetch: 'lazy' */` comment, but it will be searched in said controller’s *compiled* content. If the compilation removes that comment (it typically happens when using minifiers), then the controller is no longer considered lazy by the `ControllersMapGenerator`.

This PR makes the comment searched in source files, so that its presence doesn’t depend on the compilation’s result.

### DIFF
--- a/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
+++ b/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
@@ -77,7 +77,7 @@ class ControllersMapGenerator
             $name = str_replace(['_', '/', '\\'], ['-', '--', '--'], $name);
 
             $asset = $this->assetMapper->getAssetFromSourcePath($file->getRealPath());
-            $content = $asset->content ?: file_get_contents($asset->sourcePath);
+            $content = file_get_contents($asset->sourcePath);
             $isLazy = preg_match('/\/\*\s*stimulusFetch:\s*\'lazy\'\s*\*\//i', $content);
 
             $controllersMap[$name] = new MappedControllerAsset($asset, $isLazy);
--- a/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php
+++ b/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php
@@ -20,7 +20,7 @@ use Symfony\UX\StimulusBundle\AssetMapper\ControllersMapGenerator;
 use Symfony\UX\StimulusBundle\AssetMapper\MappedControllerAutoImport;
 use Symfony\UX\StimulusBundle\Ux\UxPackageReader;
 
-class ControllerMapGeneratorTest extends TestCase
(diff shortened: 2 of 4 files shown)


Kocal (commented):
Hi @MatTheCat, and thanks for the contribution.

I kind of agree with you, but I don't know how much of a breaking change it can be. :/

MatTheCat:
I guess it would break compilers wanting to interact with the StimulusBundle by adding or removing the `/* stimulusFetch: 'lazy' */` comment in controllers. Wouldn't it be better to decorate or replace the `stimulus.asset_mapper.controllers_map_generator` service in that case?

smnandre:
It feels to me a very internal matter... and a very logical change.

> To make one of your custom controllers lazy, add a special comment on top: [...]

To me it's never an AssetMapper thing, but a source code thing, so.. 

I'd even call this a feature to be honest , as it would remove the following constraint:

> [!NOTE]
>  If you write your controllers using TypeScript, make sure removeComments is **not** set to true in your TypeScript config.

https://symfony.com/bundles/StimulusBundle/current/index.html#lazy-stimulus-controllers

So i'm clearly 👍  on this

@MatTheCat  Do you think you could add a test or two ?

MatTheCat:
Test added!

smnandre (approved):
Thank you @MatTheCat ! Nice fix/feat :)

Kocal (approved):

Kocal:
Thanks!

javiereguiluz:
Thanks for fixing this bug Mathieu.

Issue #2273, opened 2024-10-14
Improve the link to docs on each package


javiereguiluz:
In the current web design, the `Documentation` link is displayed at the bottom of the page. For most folks, this is way below the visible contents when loading the page:

![](https://github.com/user-attachments/assets/3cda6ba9-7272-48b2-baf0-cf0fa563d2c2)

I propose to display a link to the docs right next to the command shown to install the package:

![](https://github.com/user-attachments/assets/9af7e19c-0010-414a-bb91-f70e0d6cf0af)

Don't mind the text and the "design" of my mockup. It's just a quick mockup to give you an idea of the exact location where I'd like this link to be.

Kocal:
That would be a great improvment :)

smnandre:
Sub menus for every package pages is a work in progress... i really hope to release all this before vienna :)

Kocal:
Closing, since #2309 has been merged

PR #2309, opened 2024-10-30
Show a visible link to the docs of each package

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | #2273
| License       | MIT

I know that you mentioned that there's a work-in-progress effort to update some things in the current design (https://github.com/symfony/ux/issues/2273#issuecomment-2417350181) but I think we should fix this issue quickly because it's a problem in many long package pages.

If you don't like the proposed changes here it's fine; let's iterate over them. Tell me what to fix or change and I'll do it. Thanks!

This is how this PR looks at the moment:

#SymfonyHackday :)


smnandre:
I'll open a PR this week-end (ideally sunday afternoon) -- probably loosely based on the stimulus bundle page "menu", with some simplification in markup + CSS, and some things to help us bring the next chapter :) 

But I see no problem to make this change in the meantime!

Do you think we could keep the current (flat) icon, with the color theme of the package ?

smnandre:
@javiereguiluz I'm on it and pretty confident i'll open the PR soon... (@kbond can confirm if needed)

But depending on tonight, i may miss the deadline by a day or two... 
... as i have a clear roadmap on mind .. i needed to make some other changes :) 

Poke me if you want to have an "[avant-première](https://en.wikipedia.org/wiki/Film_screening)" ;)

Kocal:
I'm working on it :eyes:

Kocal:
Since nothing moved since October/November, I still believe this PR is useful until we rewrite packages pages. 

But, I've fixed the icon and its size, to be more consistent with the rest of the application:
image

kbond (approved):

Kocal:
Thank you Javier.

smnandre:
@Kocal  "Since nothing moved since October/November" ... this is not fair 😆  You were in the slack discussions about the redesign, and saw how the scope changed 😅

Kocal:
Of course, but I mean on the live website / pull-requests :)

PR #2316, opened 2024-11-01
[TwigComponent] Optimize TwigPreLexer

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #...
| License       | MIT

Refactor the following methods:
* `check`
* `consume`
* `consumeComponentName`
* `consumeUntil`
* `consumeWhitespace`

Remove the following methods:
* `addDefaultBlock`
* `consumeAttributeName`
* `doesStringEventuallyExist`

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -39,7 +39,7 @@ class TwigPreLexer
             return $input;
         }
 
-        $this->input = $input;
+        $this->input = $input = str_replace(["\r\n", "\r"], "\n", $input);
         $this->length = \strlen($input);
         $output = '';
 
@@ -126,7 +126,8 @@ class TwigPreLexer
                 // open the default block
                 if (!empty($this->currentComponents)
                     && !$this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock']) {
-                    $output .= $this->addDefaultBlock();
+                    $output .= '{% block content %}';
+                    $this->currentComponents[\count($this->currentComponents) - 1]['hasDefaultBlock'] = true;
                 }
 
                 $attributes = $this->consumeAttributes($componentName);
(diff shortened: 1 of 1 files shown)


WebMamba (commented):
Hey Simon! Maybe this is it out of scope of this PR, but what do you think to cache the PreLexer result ? Maybe we can do that in the ComponentLexer

smnandre:
> Hey Simon! Maybe this is it out of scope of this PR, but what do you think to cache the PreLexer result ? Maybe we can do that in the ComponentLexer

I don't think we need to: 
- in prod all templates are cached 
- in dev, the lexer will be called only when the template changed (so we cannot use cache then)

kbond (approved):

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -39,7 +39,7 @@ public function preLexComponents(string $input): string
             return $input;
         }
 
-        $this->input = $input;
```
Kocal, line 42: I am surely wrong but, removing whitespaces won't increase the difficulty for the developper to know where it failed when we indicate the line number in an exception?

Kocal (approved):

src/TwigComponent/src/Twig/TwigPreLexer.php
smnandre: I think i did this for these tests 

```php
$this->line += substr_count($content, "\n");
```

(and Twig Lexer start by doing this too, soo we need to keep the same line numbers)

smnandre: Sorry ii think i missunderstood at first. 

We are not removing any whitespace here (Twig Lexer will do it), we're just normalizing the new lines to simplify things, but there is no change as `\r\n` and `\r` are new line markers as `\n`

PR #2318, opened 2024-11-01
[Turbo] Include minimal layout for Turbo Frames

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| License       | MIT

Since Turbo does not need the content outside of the frame, reducing the amount that is rendered can be a useful optimisation.

This PR is the Symfony version of Turbo Rails: https://github.com/hotwired/turbo-rails/pull/428.

To use it, you must do the following:

```twig
{% extends '@Turbo/frame.html.twig' %}

{% block body %}
    <turbo-frame id="frame_id">
        Content of the Turbo Frame
    </turbo-frame>
{% endblock %}

{# renders as: #}
<!DOCTYPE html>
<html>
    <head>
    </head>
    <body>
        <turbo-frame id="frame_id">
            Content of the Turbo Frame
        </turbo-frame>
    </body>
</html>
```


smnandre:
What would be the use case ?

smnandre:
As i understand the discussions from the issue you linked, the need really is only for people wanting to set specific head metadata (title, canonical, turbo config)... in which case i'd say this template will be really specific per app 

Also, Symfony initially come with a template very similar

```twig
<!DOCTYPE html>
<html>
    <head>
        <meta charset="UTF-8">
        <title>{% block title %}Welcome!{% endblock %}</title>
        <link rel="icon" href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 128 128%22><text y=%221.2em%22 font-size=%2296%22>⚫️</text><text y=%221.3em%22 x=%220.2em%22 font-size=%2276%22 fill=%22%23fff%22>sf</text></svg>">
        {% block stylesheets %}
        {% endblock %}

        {% block javascripts %}
            {% block importmap %}{{ importmap('app') }}{% endblock %}
        {% endblock %}
    </head>
    <body>
        {% block body %}{% endblock %}
    </body>
</html>
```

So i'm not sure we should add another "empty/base" template 🤷  

To only render a frame, you can wrap it in a block and use the `AbstractController::renderBlock`, or (possible) i'm missing something ?

seb-jean:
> As i understand the discussions from the issue you linked, the need really is only for people wanting to set specific head metadata (title, canonical, turbo config)...

Yes, that's exactly it.

smnandre:
But this would be super specific, with probably common code with their codebase, right ? I'm just wondering if this can be really used ? 

Maybe some documentation to explain why an "almost empty" layout can be used seems to me a good idea, no ?

seb-jean:
> But this would be super specific, with probably common code with their codebase, right ? I'm just wondering if this can be really used ?
> 
> Maybe some documentation to explain why an "almost empty" layout can be used seems to me a good idea, no ?

Done.

Kocal (approved):

smnandre:
What i meant was... i'm not sure if a documentation would not be enough?

 The template now makes 6 lines, this could be on documentation. 

Maintaining an almost empty file does not seem a good idea to me. 

Especially as the <html> tag needs a lang attribute, and we should not call app.currentLocale ourselves :| 

Can we just add documentation, remove the added file, and wait to see if this really is a need for users (even then, i believe a file in the bundle recipe would be a better idea)

src/Turbo/doc/index.rst
```diff
@@ -288,6 +288,42 @@ With content:
...
     </turbo-frame>
 
+Minimal layout for Turbo Frames
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.22
+
+    The minimal layout for Turbo Frames was added in Turbo 2.22.
+
+Since Turbo does not need the content outside of the frame, reducing the amount that is rendered can be a useful optimisation. However, this optimisation prevents responses from specifying ``head`` content as well. There are cases where it would be useful for Turbo to have access to items specified in ``head``. To specify the heads, you must then use a minimal layout for frame, rather than no layout. With this, applications can render content into the ``head`` if they want.
```
smnandre: > reducing the amount that is rendered can be a useful optimisation

I'd present first an example of this (with controller code and renderblock for instance)

> However, this optimisation prevents responses from specifying

This part can be removed i think, as the same idea is on the next sentence. And i feel it's a bit fearfull and would not want to alarm users if they dot need any meta tags (which is, as i see it, the most standard case, no ?)

> With this, applications can render content into the ``head`` if they want.

Maybe something like "it allows you to set meta tags while still having a minimal ...." ?

src/Turbo/doc/index.rst
```diff
@@ -288,6 +288,42 @@ With content:
...
+
+.. code-block:: html+twig
+
+    {% extends '@Turbo/frame.html.twig' %}
```
smnandre:
```
  {% extends '@Turbo/frame.html.twig' %}

    {% block head %}
        <meta name="alternative" content="present" />
    {% endblock %}

    {% block body %}
        <turbo-frame id="frame_id">
            Content of the Turbo Frame
        </turbo-frame>
    {% endblock %}
```

```
<!DOCTYPE html>
<html>
  <head>
      <meta name="alternative" content="present" />
  </head>
  <body>
      <turbo-frame id="frame_id">
          Content of the Turbo Frame
      </turbo-frame>
  </body>
</html>
```

Not sure about this gain :)

smnandre:
Really i like the intention @seb-jean  (as always with you ;) )... but i realized even the base.html / app.css / etc are not in the framework but in recipe... and you see we're not  thousand  around here so... ;)

seb-jean:
Does that mean I have to delete the Turbo template?

smnandre:
1/ You don't have to anything
2/ You know better than me what could helper Turbo developpers
3/ See point 1 


;)

Kocal:
Hi, what is the state of this PR?

seb-jean:
Hi Hugo,
It's super specific so I'm going to close this PR.

PR #2325, opened 2024-11-02
Allow to disable nested attributes with !

ottaviano:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | ~
| License       | MIT

With the nested attributes feature, components have become incompatible with the AlpineJS library, which also uses HTML attributes in a similar format, such as x-bind:class="foo" or x-transition:enter="ease-out duration-300".
This PR introduces a way to specify that an attribute containing : and ending with ! should not be treated as a nested attribute.

Propose : 

```twig
<twig:Button x-bind:class!="hey">My button</twig:Button>
```

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.21.0
+
+-  Allow to disable nested attributes with !
+
 ## 2.20.0
 
 -  Add Anonymous Component support for 3rd-party bundles #2019
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -21,7 +21,7 @@ use Symfony\WebpackEncoreBundle\Dto\AbstractStimulusDto;
  */
 final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Countable
 {
-    private const NESTED_REGEX = '#^([\w-]+):(.+)$#';
+    private const NESTED_REGEX = '#^([\w-]+):(.+)(?<!!)$#';
 
     /** @var array<string,true> */
     private array $rendered = [];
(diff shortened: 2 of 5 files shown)


WebMamba (commented):
Could you add some documentation ? What do you think about a section on how to use TwigComponent with Alpine js ?

smnandre:
Hi @ottaviano, thank you for this PR!

We have special handling for aria attributes, classes, stimulus attributes, etc.

I see not problem if we add a special case if a given attribute key:
- start with x- 
- contains a : 

(and I still consider that "x-bind" cannot be a component name)

So i'm still thinking we should have a special treatment here.. because even if the solution you suggest is smart, i really don't like to open the regex like this and allow new caracters here. :| 

What do you think ?

Kocal:
Do we agree this issue would not occurs if the "nested attributes" feature  didn't exist? :/

smnandre:
Once we release and document a feature, we can’t just remove it without consideration—even if I’d love to, especially to resolve some Twig/Live issues! 🙂

ottaviano:
managing each library (like AlpineJs, Vue, etc) independently can add complexity to the code, as we may need to revisit and modify it if a new conflict arises with another library in the future. My approach is to simply enable/disable a nested attribute, ensuring compatibility with any JavaScript library.

Kocal:
Yes, but it add (again) a new syntax for attributes, that should be documented and known by users. 

With #2328, yes it add more work and complexity for us (UX mainteners), but users will be able to copy/paste examples valid Alpine (and others) code without trying to understand why the code does not work when they use the special attributes on a Twig component:
image

ottaviano:
I see your point; I close this PR in favor of https://github.com/symfony/ux/pull/2328
Thank you for your work, can’t wait to test the improvement !

smnandre:
Yep, nothing to do with your PR, i was thinking of DX and incoming support .. i should have been more explicit sorry @ottaviano  :)

PR #2162, opened 2024-09-14
[Map] Adding polygons to google and leaflet + info window

rrr63:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | 
| License       | MIT

Hello, 

In my job i use to work a lot with maps, so i would love to work with symfony ux in next projects.
But i have to work with polygons, so i tried to implement it to the ux-map.

I added polygons on the Map component, and made it work on both bridge : google and leaflet.
I added a window popup on click also, the current version of window popup worked only with markers.
I had to rename `createInfoWindow` to `createInfoWindowMarker`. With that i could add a `createInfoWindowPolygon`.
The only difference with the infoWindow from marker is that his anchor will be the event cursor position, because the area of the polygon could be very large and we want a window on the click position.

You can try by adding a simple polygon on your map :
```php
      ->addPolygon(new Polygon(
                points: [
                    new Point(45.7402741, 3.191185),
                    new Point(45.7314078, 3.1903267),
                    new Point(45.7307488, 3.208952),
                    new Point(45.7402741, 3.2112694)
                ],
                rawOptions:[
                    'fillColor' => 'green',
                    'color' => 'green',
                ],
                infoWindow: new InfoWindow(
                    headerContent: '<b>Polygon</b>',


                    content: 'I am a polygon'
                )
            ))
```
smallgreenrectangle

Here a full example that works with leaflet and google that show some of french regions via an api call

```php
$polygons = json_decode(file_get_contents('api_url'), true);
        $polygons = $polygons['features'];
        
        $googleOptions = (new GoogleOptions())
            ->mapId('mapId')
        ;
    
        $map = (new Map('default', $googleOptions))
            ->center(new Point(45.8079318, 3))
            ->zoom(10)
            ;
    
        foreach ($polygons as $polygon) {
            $points = [];
            $title = $polygon['properties']['nom'];
            foreach ($polygon['geometry']['coordinates'][0] as $point) {
                if (count($point) == 2) {
                    $points[] = new Point($point[1], $point[0]);
                }
            }
    
            $map->addPolygon(new Polygon(
                points: $points,
                rawOptions:[
                    'fillColor' => 'red',
                    'color' => 'red',
                ],
                infoWindow: new InfoWindow(
                    headerContent: '<b>'.$title.'</b>',
                    content: 'The French region of '.$title.'.'
                )
            ));
        }
```
googleregions
leafletregions

I added tests for polygons and polygon info window
[...]

PR #2326, opened 2024-11-02
[DX] Add per-package Yarn scripts (build, watch, test, lint, ...)

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

When working in UX Map, it was very painful to build or watch modifications on `.ts` files, I had investigate how the building process worked, and had to run Rollup like this, which is far from ideal (not documented and not obvious):  
```
./node_modules/.bin/rollup -c --environment INPUT_FILE:src/Map/assets/src/abstract_map_controller.ts -w
./node_modules/.bin/rollup -c --environment INPUT_FILE:src/Map/src/Bridge/Google/assets/src/map_controller.ts -w
./node_modules/.bin/rollup -c --environment INPUT_FILE:src/Map/src/Bridge/Leaflet/assets/src/map_controller.ts -w
```

And I also felt sorry for @rrr63 when he worked on adding Polygons on Map (#2162)...

This PR improves the way developers will work on UX, and makes their lives easier.

**Before**:
- `yarn build` compiled the assets from ALL packages, it was not possible to build packages from only one package (which is useful if you work on a single package)
- no `yarn watch`
- `yarn test` runned tests from ALL packages, like `yarn build`, it was not possible to run tests for only one package

**Now:** 
- at the project root, `build`/`test`/`lint`/`format`/`check-lint`/`check-format` scripts will run on all assets **from all packages**. And it will be faster than before, when processing was sequential, but now it's parallelized.
- at a package root (ex:  `src/Map/assets`), `build`/`test`/`lint`/`format`/`check-lint`/`check-format` scripts will run on all assets **from this package only**
- `build` and `watch` scripts handles both TypeScript and CSS files in a single command

This is a first step to what we spoke about with @smnandre to write a contribution guide.
It is now much more easier and friendlier to tell a developer to run `yarn watch` inside a package root, instead of telling it to run `./node_modules/.bin/rollup -c --environment INPUT_FILE:src/Map/src/Bridge/Google/assets/src/map_controller.ts -w` (and more, if you have multiple files).

image


bin/build_package.js
```diff
@@ -0,0 +1,128 @@
...
+            : undefined;
+        if (!inputStyleFile) {
+            return;
+        }
+
+        console.log('Minifying CSS...');
+        const css = await fs.promises.readFile(inputStyleFile, 'utf-8');
+        const minified = new CleanCSS().minify(css).styles;
+        await fs.promises.writeFile(inputStyleFileDist, minified);
+    };
```
Kocal, line 75: I don't know if `config.css_source` is used somewhere else, I'm not super fan of having this explicit key when you can simply glob `*.css` files.

package.json
```diff
@@ -24,7 +24,7 @@
         "@rollup/plugin-typescript": "^11.1.6",
         "@symfony/stimulus-testing": "^2.0.1",
         "@vitest/browser": "^2.1.1",
-        "clean-css-cli": "^5.6.2",
```
Kocal, line 27: The `cli` version of this package was not necessary, `clean-css` provides a JavaScript API that we can use in our `build_package.js` script.

src/LiveComponent/assets/dist/Component/plugins/QueryStringPluging.d.ts
Kocal, line 1: This file was commited by error, it is `QueryStringPlugin` without the `g` at the end.
Now the building script remove the `dist` directory before building, this kind of issue should not happens anymore.

bin/build_package.js
```diff
@@ -0,0 +1,128 @@
...
+    if (!fs.existsSync(srcDir)) {
+        console.error(`The package directory "${packageRoot}" does not contain a "src" directory.`);
+        process.exit(1);
+    }
+
+    if (fs.existsSync(distDir)) {
+        console.log(`Cleaning up the "${distDir}" directory...`);
+        await fs.promises.rm(distDir, { recursive: true });
+        await fs.promises.mkdir(distDir);
+    }
```
Kocal, line 50: This was not done in the previous script, but this is super useful to ensure we start with a clean `dist` directory.

bin/rollup.js
```diff
@@ -0,0 +1,138 @@
...
+        // React, Vue
+        if (file.includes('assets/src/loader.ts')) {
+            peerDependencies.push('./components.js');
+        }
+    });
+
+    const outDir = path.join(packageRoot, 'dist');
+
+    return {
+        input: inputFiles,
```
Kocal, line 99: We now spawn one `rollup` process per package, instead of one `rollup` process per package per files.

bin/build_package.js
smnandre: Only in "a-package"/assets/styles directories right ? :p

package.json
smnandre: Me looking at this package:

yw8i_b

bin/build_package.js
smnandre: (i'd like to remove the CSS files to be honest)

bin/rollup.js
```diff
@@ -0,0 +1,138 @@
...
+    name: 'move-ts-declarations',
+    writeBundle: async () => {
+        const isBridge = packageRoot.includes('src/Bridge');
+        const globPattern = path.join(packageRoot, 'dist', '**', 'assets', 'src', '**/*.d.ts');
+        const files = glob.sync(globPattern);
+
+        files.forEach((file) => {
+            // a bit odd, but remove first 8 or 11 directories, which will leave
+            // only the relative path to the file
+            const relativePath = file
```
smnandre: 🤔 
Maybe add an example ?

biome.json
```diff
@@ -4,6 +4,8 @@
         "include": [
             "*.json",
             "*.md",
+            "bin/*.js",
+            "test/*.js",
```
smnandre, line 8: Jumping on the occasion, could you please change line 18 ? 👼 

```diff
-            "ignore": ["**/composer.json", "**/vendor", "**/package.json", "**/node_modules"] 
+            "ignore": ["**/composer.json", "**/vendor", "**/package.json", "**/node_modules", "ux.symfony.com/**"]
```

bin/rollup.js
Kocal: Done

biome.json
Kocal: Why?

smnandre: Ho and `**/var/` too :)

Kocal: I don't think this is necessary, since we configured Biome `files.includes` and we don't run on `ux.symfony.com` nor `var` folders

bin/build_package.js
Kocal: For a next iteration, I will try to use [Lightning CSS](https://lightningcss.dev/) instead

biome.json
smnandre: It does when i run check-format from the repo root

image

smnandre:
This is so grea! And it does run very quickly!

Congrats and thank you 🙇

bin/build_package.js
smnandre: For me: we don't need to build CSS

smnandre: I played with Lightning CSS, it's dope.

smnandre (approved):

biome.json
Kocal: Ah, didn't know you had `.json` files under `ux.symfony.com/.....`. Is that a SF 7.2 thing?

Kocal: Yep it is, fixed, thank you

PR #2328, opened 2024-11-03
[TwigComponent] Ignore "nested" for Alpine & Vue attributes

### LINKED ISSUE #1839: AlpineJS specific attributes are not rendered (`@click` and `x-on:click`)

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #1839
| License       | MIT

Have lost time on Twig & the website :sweat:

Alternative implementation of #2325 

Update: Improved `__toString` performance following @Kocal comments

--- 
Now all these attributes are directly rendered 

| Framework       | Prefix | Code Example                                                                                       | Documentation                                                 |
|-----------------|---------|-------------------------------------------------------------------------------------------------------|----------------------------------------------------------------|
| **Alpine.js**   | `x-`    | `<div x-data="{ open: false }" x-show="open"></div>`                                                 | [Documentation Alpine.js](https://alpinejs.dev/)               |
| **Vue.js**      | `v-`    | `<input v-model="message" v-if="show">`                                                              | [Documentation Vue.js](https://vuejs.org/guide/)               |
| **Stencil**     | `@`     | `<my-component @onClick="handleClick"></my-component>`                                               | [Documentation Stencil](https://stenciljs.com/docs/)           |
| **Lit**         | `@`     | `<button @click="${this.handleClick}">Click me</button>`                                             | [Documentation Lit](https://lit.dev/docs/)                    |

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -37,48 +37,43 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
 
     public function __toString(): string
     {
-        return array_reduce(
-            array_filter(
-                array_keys($this->attributes),
-                fn (string $key) => !isset($this->rendered[$key])
-            ),
-            function (string $carry, string $key) {
-                if (
-                    str_contains($key, ':')
-                    && preg_match(self::NESTED_REGEX, $key)
-                    && !preg_match(self::ALPINE_REGEX, $key)
-                    && !preg_match(self::VUE_REGEX, $key)
-                ) {
-                    return $carry;
-                }
+        $attributes = '';
 
-                $value = $this->attributes[$key];
+        foreach ($this->attributes as $key => $value) {
+            if (isset($this->rendered[$key])) {
+                continue;
+            }
 
-                if ($value instanceof \Stringable) {
-                    $value = (string) $value;
-                }
+            if (
+                str_contains($key, ':')
+                && preg_match(self::NESTED_REGEX, $key)
+                && !preg_match(self::ALPINE_REGEX, $key)
+                && !preg_match(self::VUE_REGEX, $key)
+            ) {
+                continue;
+            }
 
-                if (!\is_scalar($value) && null !== $value) {
-                    throw new \LogicException(\sprintf('A "%s" prop was passed when creating the component. No matching "%s" property or mount() argument was found, so we attempted to use this as an HTML attribute. But, the value is not a scalar (it\'s a "%s"). Did you mean to pass this to your component or is there a typo on its name?', $key, $key, get_debug_type($value)));
-                }
+            if (null === $value) {
+                trigger_deprecation('symfony/ux-twig-component', '2.8.0', 'Passing "null" as an attribute value is deprecated and will throw an exception in 3.0.');
+                $value = true;
+            }
 
-                if (null === $value) {
-                    trigger_deprecation('symfony/ux-twig-component', '2.8.0', 'Passing "null" as an attribute value is deprecated and will throw an exception in 3.0.');
-                    $value = true;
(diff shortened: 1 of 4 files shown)


src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -41,7 +43,11 @@ public function __toString(): string
...
             ),
             function (string $carry, string $key) {
-                if (preg_match(self::NESTED_REGEX, $key)) {
+                if (
+                    preg_match(self::NESTED_REGEX, $key)
+                    && !preg_match(self::ALPINE_REGEX, $key)
+                    && !preg_match(self::VUE_REGEX, $key)
+                ) {
                     return $carry;
                 }
```
Kocal: `preg_match` calls can be expensive, and can be avoid by quickly checking if the attribute name contains a `:` (ex: https://github.com/symfony/symfony/pull/28106). If it's not the case we early exit (without calling `preg_match`), otherwise we continue.

```suggestion
                if (
                    str_contains($key, ':')
                    && preg_match(self::NESTED_REGEX, $key)
                    && !preg_match(self::ALPINE_REGEX, $key)
                    && !preg_match(self::VUE_REGEX, $key)
                ) {
                    return $carry;
                }
```

Can you add the same check to `\Symfony\UX\TwigComponent\ComponentAttributes::nested` please? 

Thanks!

PS: Not related to your PR, but I'm not a fan about all the `array_reduce`/`array_filter`/`array_keys` calls where a `foreach` could be enough. We can open a new PR to rewrite and optimize that :)

Kocal (approved):

src/TwigComponent/src/ComponentAttributes.php
smnandre: > Can you add the same check to \Symfony\UX\TwigComponent\ComponentAttributes::nested please?

The str_contains will be enough here, as if someone want to explicitely ask for .nested("x-on") let's not block this.

> I'm not a fan about all the array_reduce/array_filter/array_keys calls where a foreach could be enough. We can open a new PR to rewrite and optimize that :)

Yes, will do leter

smnandre: Done here in the end, i'll let you check

Kocal (approved):

WebMamba:
I am sorry but I don't like the solution taken here. I like way more the initial solution taken here: #2325. I think we should not add code specific to another framework, and I don't think that we should ignore nested attributes without making them implicit for the user. I think we should focus on a solution like the initial solution where the user choose to ignore it implicitly.

Kocal:
What about https://github.com/symfony/ux/pull/2325#issuecomment-2453408427 then?
People will have to adapt their code to make it works with Twig Components?

I don't find that super friendly, it not obvious at all, and can be a source of frustration to the user even if it's documented (and document that you need to explicitly use a new syntax to disable nested attributes to makes your code works, feels like a loss to me)

smnandre:
Are we _really_ considering there is real scenario where someone 
* has a component
* has a child component in it named "x-on" 
* uses nested attributes

?

Regarding specific code, we have some for LiveComponent & Alpine already. On the other hand, opening a new char / syntax in the attributes will be a change with many more problems.

Kocal:
Thank you @smnandre for taking care of this regression.

PR #2332, opened 2024-11-03
Test diff size (4)

Kocal:
I want to see if https://github.com/symfony/ux/pull/2331 works.


Kocal:
Finally 🥹 

But it does not show the `.d.ts` file being removed, gonna investigate. The title is wrong tho

PR #2334, opened 2024-11-03
[LiveComponent] Explicit accessor on live component demo

DocFX:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| License       | MIT

Stumbled upon the documentation, sweated some time on understanding the multiple layers of magic in the demo, thought this would help some people. :)

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -49,6 +49,7 @@ A real-time product search component might look like this::
         >
 
         <ul>
+            {# remember: "`.products`" accessor checks at least if a "`getProducts()`" method exists #}
             {% for product in this.products %}
                 <li>{{ product.name }}</li>
             {% endfor %}


smnandre:
Hi @DocFX and thank you for this PR. 

Would you explain a bit more what you mean by 

> .products accessor checks at least if a getProducts() method exists

Or more specifically what was not clear at first  ? 

(Genuinely trying to understand where we can improve documentation -- more globally i mean)

DocFX:
> Hi @DocFX and thank you for this PR.
> 
> Would you explain a bit more what you mean by
> 
> > .products accessor checks at least if a getProducts() method exists
> 
> Or more specifically what was not clear at first ?
> 
> (Genuinely trying to understand where we can improve documentation -- more globally i mean)

Actually, I feel like this part of documentation uses a little bit too much "magic". 
The example is really precise and does not really separate what is specific to it from what is generic and provided by the bundle.

I really struggled at first to understand what the relation was between all files. The asbtraction layer between the JS and the objects is already a little hard to deal with. It took some time for me to understand that this uses the automatic getter inspired by twig "dot" accessor.

smnandre:
>  It took some time for me to understand that this uses the automatic getter inspired by twig "dot" accessor.

It is not inspired by the Twig "dot" accessor, it is the Twig dot accessor :) 

We have a warning to read the TwigComponent documentation before, but it could be highlighted more i guess.

smnandre:
Thank you for bringing this up and sharing your thoughts. I'm closing this for now as there aren’t any immediate actions we can take on our side. The next version of the website will address some of these points, but at the moment, we’re limited in what we can do. LiveComponents can be a bit challenging to grasp initially, especially without prior knowledge of Twig and TwigComponent.

We appreciate your patience and understanding!

PR #2344, opened 2024-11-06
[TwigComponent] Improve ComponentFactory performances

smnandre:
ComponentFactory Quick Optimization

Minor change with.. mid impact
(on large number of renders)

### DIFF
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\TwigComponent;
 use Psr\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\DependencyInjection\ServiceLocator;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Contracts\Service\ResetInterface;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 use Symfony\UX\TwigComponent\Event\PostMountEvent;
 use Symfony\UX\TwigComponent\Event\PreMountEvent;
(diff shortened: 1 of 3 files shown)


smnandre:
false positive for fabbot

norkunas:
When running let's say frankenphp, this will always stay in memory?

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -141,37 +145,40 @@ public function get(string $name): object
...
-        }
-
         if ($component instanceof AnonymousComponent) {
             $component->mount($data);
 
             return;
         }
 
-        $parameters = [];
+        if (!\array_key_exists($component::class, self::$mountMethods)) {
```
Kocal:
```suggestion
        if (null === (self::$mountMethods[$component::class] ?? null)) {
```

Kocal (approved):

Kocal:
Thanks for this optimization!

However I'm thinking, not now but maybe for a next PR, but do you think things can be more improved if this "find mounts methods" process was done during a cache warmer and persisted in cache? 
I don't speak about storing `ReflectionMethod` in cache (I don't think it is possible), but storing that if `$component::class` **has** `mount` method or not: 
```diff
    private function mount(object $component, array &$data): void
    {
        if ($component instanceof AnonymousComponent) {
            $component->mount($data);

            return;
        }
 
+       if (!self::$hasMountMethod[$component::class]) {
+           return;
+       }

        if (!\array_key_exists($component::class, self::$mountMethods)) {
            try {
                $mountMethod = self::$mountMethods[$component::class] = (new \ReflectionClass($component))->getMethod('mount');
            } catch (\ReflectionException) {
                self::$mountMethods[$component::class] = false;

                return;
            }
        }
```
WDYT?

smnandre:
> However I'm thinking, not now but maybe for a next PR [...]  WDYT?

Absolutely! 

I'm moving step by step, a bit in the blind. The biggest change is yet to come (this week hopefully). This is why i'm only doing work on private/internal methods for now.

Once we have the results i'm targetting, a clean will be very welcome (required?)..  We will then see how to name these classes / methods, and how we can rearrange the whole thing

👍

smnandre:
> When running let's say frankenphp, this will always stay in memory?

Would it be a problem ? Genuine question here :)

norkunas:
Depends on how much memory it will use. ResetInterface could be used to cleanup after request

smnandre:
Updated!

Kocal (approved):

PR #2345, opened 2024-11-06
[Site] Add attribute `xmlns` to icons

seb-jean:
I allow myself to add the `xmlns="http://www.w3.org/2000/svg"` attribute.

What do you think?

### DIFF
--- a/ux.symfony.com/config/packages/ux_icons.yaml
+++ b/ux.symfony.com/config/packages/ux_icons.yaml
@@ -1,7 +1,7 @@
 ux_icons:
     default_icon_attributes:
         class: 'Icon'
-        xmlns: 'http://www.w3.org/2000/svg'
+        xmlns: 'https://www.w3.org/2000/svg'
 
     icon_sets:
 
--- a/ux.symfony.com/config/packages/ux_icons.yaml
+++ b/ux.symfony.com/config/packages/ux_icons.yaml
@@ -1,6 +1,7 @@
 ux_icons:
     default_icon_attributes:
         class: 'Icon'
+        xmlns: 'http://www.w3.org/2000/svg'
 
     icon_sets:
 


smnandre:
~~As this would potentially change how icons are renders in existing apps... what about updating the recipe instead ?~~

Read too quickly sorry!

ux.symfony.com/config/packages/ux_icons.yaml
```diff
@@ -1,6 +1,7 @@
 ux_icons:
     default_icon_attributes:
         class: 'Icon'
+        xmlns: 'http://www.w3.org/2000/svg'
```
smnandre:
```suggestion
        xmlns: 'https://www.w3.org/2000/svg'
```

seb-jean:
I made this addition because the icon has the `xmlns` attribute defined at the start but it is not rendered at the end.

smnandre (approved):

norkunas:
> I made this addition because the icon has the `xmlns` attribute defined at the start but it is not rendered at the end.

But is it needed? :slightly_smiling_face:

seb-jean:
> > I made this addition because the icon has the `xmlns` attribute defined at the start but it is not rendered at the end.
> 
> But is it needed? 🙂

Do you think this is an attribute that is optional?

norkunas:
Yes, browsers renders svg's without it perfectly fine :)

seb-jean:
Well then, I don't know if it's an optional or required attribute 😄.

smnandre:
It is "required" but in reality for this kind usage it makes no difference.

javiereguiluz (approved):

kbond:
I think this clearly states it isn't required for ux-icons:

> Note: The xmlns attribute is only required on the outermost svg element of SVG documents, or inside HTML documents with XML serialization. It is unnecessary for inner svg elements or inside HTML documents with HTML serialization.
> -- https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg

seb-jean:
I am closing this PR for the reasons given by @kbond.

PR #2347, opened 2024-11-06
[StimulusBundle][Doc] Mention `removeComments` will no longer be necessary

MatTheCat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | N/A
| License       | MIT

Once https://github.com/symfony/ux/pull/2304 is released, it will no longer be necessary to set `removeComments` to `false` when using TypeScript, because the `stimulusFetch: 'lazy'` comment will be searched in the source rather than the compiled content.


smnandre (approved):

Kocal:
Thank you @MatTheCat.

PR #2348, opened 2024-11-06
[Autocomplete][Doc] Mention `preload` can be set to `false`

MatTheCat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | N/A
| License       | MIT

Setting `preload` to `false` is useful when you only want to display search results; that way nothing is loaded on focus.


smnandre (approved):

Kocal:
Thank you @MatTheCat.

PR #2351, opened 2024-11-07
[Icons] Prepare lock / warmup optimization

smnandre:
Method needed for both
* lock/warmup optimization
* sprite generation
* search performances

### DIFF
--- a/src/Icons/src/Iconify.php
+++ b/src/Icons/src/Iconify.php
@@ -85,6 +85,54 @@ final class Iconify
         ]);
     }
 
+    public function fetchIcons(string $prefix, array $names): array
+    {
+        if (!isset($this->sets()[$prefix])) {
+            throw new IconNotFoundException(\sprintf('The icon set "%s" does not exist on iconify.design.', $prefix));
+        }
+
+        // Sort to enforce cache hits
+        sort($names);
+        $queryString = implode(',', $names);
+        if (!preg_match('#^[a-z0-9-,]+$#', $queryString)) {
+            throw new \InvalidArgumentException('Invalid icon names.');
+        }
+
+        // URL must be 500 chars max (iconify limit)
+        // -39 chars: https://api.iconify.design/XXX.json?icons=
+        // -safe margin
+        if (450 < \strlen($prefix.$queryString)) {
+            throw new \InvalidArgumentException('The query string is too long.');
+        }
+
+        $response = $this->http->request('GET', \sprintf('/%s.json', $prefix), [
+            'headers' => [
+                'Accept' => 'application/json',
+            ],
+            'query' => [
+                'icons' => strtolower($queryString),
+            ],
+        ]);
+
+        if (200 !== $response->getStatusCode()) {
+            throw new IconNotFoundException(\sprintf('The icon set "%s" does not exist on iconify.design.', $prefix));
+        }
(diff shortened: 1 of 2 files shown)


src/Icons/src/Iconify.php
```diff
@@ -85,6 +85,52 @@ public function fetchIcon(string $prefix, string $name): Icon
...
+            'query' => [
+                'icons' => $queryString,
+            ],
+        ]);
+
+        if (200 !== $response->getStatusCode()) {
+            throw new IconNotFoundException(\sprintf('The icon set "%s" does not exist on iconify.design.', $prefix));
+        }
+
+        $data = $response->toArray();
```
Kocal: Can't we support more than 500 chars by splitting icons and doing multiple requests?

src/Icons/src/Iconify.php
```diff
@@ -85,6 +85,52 @@ public function fetchIcon(string $prefix, string $name): Icon
...
+        // Sort to enforce cache hits
+        $queryString = implode(',', $names);
+        if (!preg_match('#^[\w-]+(,[\w-]+)*$#', $queryString)) {
+            throw new \InvalidArgumentException('Invalid icon names.');
+        }
+
+        // 500 chars max
+        // 42 chars: https://api.iconify.design/mdi.json?icons=
+        if (450 < \strlen($queryString)) {
+            throw new \InvalidArgumentException('The query string is too long.');
```
Kocal: Please use dedicated exception classes

src/Icons/src/Iconify.php
Kocal: Ah, didn't see #2352

Kocal (approved):

src/Icons/src/Iconify.php
kbond: Is the plan to catch this? (I typically only add named exceptions if this is the case - or it's realistic an end-user would want to catch)

src/Icons/src/Iconify.php
```diff
@@ -85,6 +85,52 @@ public function fetchIcon(string $prefix, string $name): Icon
...
     }
 
+    public function fetchIcons(string $prefix, array $names): array
+    {
+        if (!isset($this->sets()[$prefix])) {
+            throw new IconNotFoundException(\sprintf('The icon set "%s" does not exist on iconify.design.', $prefix));
+        }
+
+        // Sort to enforce cache hits
+        $queryString = implode(',', $names);
```
javiereguiluz, line 96: the comment above says "sort" (to get more cache hits) but here you are not sorting the values, just imploding them. The sorting of the `$names` array is missing, right?

src/Icons/src/Iconify.php
Kocal: Nah I don't think so, but I like to have dedicated exceptions and also a base exception class per package (like they do on Symfony components)

smnandre: Not on every component to be far... ;) 

Yet I kinda agree to use custom exception but this should be done in a dedicated PR i think, carefully selecting the one we want to "migrate" and the one we do not. 

To explain myself here, this is internal and "should never" happen. 

But as we all know what "should never" means... i get your point.

kbond (approved):

PR #2352, opened 2024-11-08
[Icons] Fetch icons in batch in Import command

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #... 
| License       | MIT

First usage of #2351 with the import command 

---

**Update:** 

- add icons counts (imported / total)
- handle same-prefix icons per batch
- improve visual feedback

Capture d’écran 2024-11-09 à 19 26 16

---

Next ones:
* lock
* warm-up

### DIFF
--- a/src/Icons/src/Command/ImportIconCommand.php
+++ b/src/Icons/src/Command/ImportIconCommand.php
@@ -18,7 +18,6 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
-use Symfony\UX\Icons\Exception\IconNotFoundException;
 use Symfony\UX\Icons\Iconify;
 use Symfony\UX\Icons\Registry\LocalSvgIconRegistry;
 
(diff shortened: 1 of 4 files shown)


Kocal (approved):

src/Icons/src/Command/ImportIconCommand.php
```diff
@@ -65,33 +63,57 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                "\n",
+            ]);
+
+            foreach (array_chunk($icons, 25, true) as $iconBatch) {
+                foreach ($iconBatch as $name => $fullName) {
+                    $io->writeln(\sprintf('  Importing %s:%s ...', $prefix, $name));
+                }
+
+                try {
+                    $batchResults = $this->iconify->fetchIcons($prefix, $names);
```
javiereguiluz: In this call ... aren't you fetching all icons at once? The `$names` variable contains ALL icon names, right? Shouldn't you create a new variable with up to 25 icons, which is what you want to download thanks to the previous `array_chunk()`?

smnandre: Good catch @javiereguiluz!  

This should indeed be array_keys($iconBatch) here. 

Will fix when the other one is merged (need to rebase anyway)

Thanks!

kbond (approved):

PR #2356, opened 2024-11-09
[Icons] Support aliases in LockIconsCommand

### LINKED ISSUE #2354: [Icons] Add support for icon aliases during icon:lock command
### Context

Currently, the `IconFinder` only scans templates to identify used icons, using the following regex:

```php
$token = '[a-z0-9]+(?:-[a-z0-9]+)*';
$pattern = "#(?:'$token:$token')|(?:\"$token:$token\")#i";
```

This approach means that only icons directly referenced in templates are locked when running the `icon:lock` command.

### Problem

Some icons are defined as aliases in the `ux_icons` configuration file. These aliased icons are not detected by `IconFinder` and therefore are not locked automatically, which can lead to inconsistencies during deployment.

### Request

Would it be possible to add support for scanning icon aliases defined in the `ux_icons` file, in addition to using the regex to search templates? This enhancement would ensure that all icons actually used in the application are properly locked.

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #2354
| License       | MIT

Alias icons are now all locked, and icon set aliases are supported too.

### DIFF
--- a/src/Icons/src/Command/LockIconsCommand.php
+++ b/src/Icons/src/Command/LockIconsCommand.php
@@ -37,6 +37,8 @@ final class LockIconsCommand extends Command
         private Iconify $iconify,
         private LocalSvgIconRegistry $registry,
         private IconFinder $iconFinder,
+        private readonly array $iconAliases = [],
+        private readonly array $iconSetAliases = [],
     ) {
         parent::__construct();
     }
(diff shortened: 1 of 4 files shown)


smnandre:
Test Failures unrelated (JS Map)

Kocal (approved):

kbond (approved):
Nice!

PR #2360, opened 2024-11-10
[Map] Add position and missing closing parenthesis

tacman:
so that it works when blindly cutting and pasting.

### DIFF
--- a/src/Map/doc/index.rst
+++ b/src/Map/doc/index.rst
@@ -116,7 +116,7 @@ You can add markers to a map using the ``addMarker()`` method::
         // You can also pass arbitrary data via the `extra` option in both the marker
         // and the infoWindow; you can later use this data in your custom Stimulus controllers
         ->addMarker(new Marker(
-            // ...
+            position: new Point(45.7740, 4.8351), 
             extra: [
                 'icon_mask_url' => 'https://maps.gstatic.com/mapfiles/place_api/icons/v2/tree_pinlet.svg',
             ],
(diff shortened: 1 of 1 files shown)


src/Map/doc/index.rst
```diff
@@ -116,7 +116,7 @@ You can add markers to a map using the ``addMarker()`` method::
         // You can also pass arbitrary data via the `extra` option in both the marker
         // and the infoWindow; you can later use this data in your custom Stimulus controllers
         ->addMarker(new Marker(
-            // ...
+            position: new Point(45.7740, 4.8351),
```
smnandre, line 119: Could you then remove the one L124 and/or add missing data here.

Or maybe we should use a diff block, with the previous data, wdyt ?

```diff
    ->addMarker(new Marker(
            position: new Point(45.7640, 4.8357), 
            title: 'Lyon',
+            extra: [
+                'icon_mask_url' => 'https://example.com/my-icon.svg',
+           ],
            infoWindow: new InfoWindow(
                headerContent: '<b>Lyon</b>',
+                extra: [
+                    'num_items' => 3,
+                    'includes_link' => true,
                ],
                content: 'The French town in the historic Rhône-Alpes region, located at the junction of the Rhône and Saône rivers.',
            )
))

Kocal (approved):

javiereguiluz:
Good catch, thanks Tac.

src/Map/doc/index.rst
javiereguiluz: @smnandre I prefer to not use `diff` blocks too much because they don't have any syntax highlighting (except the green/red diff blocks) and that's confusing sometimes. Thanks.

PR #2362, opened 2024-11-10
<deleted>

yyaremenko:


yyaremenko:
also, the tests are failing... after docu

nicolas-grekas:
When the numbers come from standards like HTTP, we prefer the actual value.

yyaremenko:
why do you have a whole set of constants for HTTP codes in a Response class then?

yyaremenko:
...and zero response in a week, so classy of symfony devs )) anw.

PR #2364, opened 2024-11-11
[Twig] Cache template class resolution

smnandre:
Cache template class during rendering.

Non negligeable impact on pages with many components.

Capture d’écran 2024-11-11 à 01 59 30

### DIFF
--- a/src/TwigComponent/src/ComponentRenderer.php
+++ b/src/TwigComponent/src/ComponentRenderer.php
@@ -12,6 +12,7 @@
 namespace Symfony\UX\TwigComponent;
 
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\Service\ResetInterface;
 use Symfony\UX\TwigComponent\Event\PostRenderEvent;
 use Symfony\UX\TwigComponent\Event\PreCreateForRenderEvent;
 use Symfony\UX\TwigComponent\Event\PreRenderEvent;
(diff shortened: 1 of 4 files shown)


src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -24,6 +24,8 @@
  */
 final class ComponentRenderer implements ComponentRendererInterface
 {
+    private array $templateClasses = [];
```
kbond, line 28: We should reset this state I think.

smnandre: Done, and added some (not so) minor optim in Factory again :|

smnandre:
Arf.... again LiveComponent hidden contract .. :( 

Will fix later

smnandre:
Cannot reproduce locally :(

smnandre:
Weird. I think is it related to property accessor but it’s weird this différence between local and CI.

Kocal:
Locally, is your `src/LiveComponent/vendor/symfony/ux-twig-component` points to `src/TwigComponent`?

smnandre:
Ok i confirm... the Live test helper does not simulate real conditions. But i think we can deal with it :) 


This was the change responsible for failures 🤷 


```diff
+       if (!$componentMetadata->isAnonymous()) {
            // set data that wasn't set in mount on the component directly
            foreach ($data as $property => $value) {
                if ($this->propertyAccessor->isWritable($component, $property)) {
                    $this->propertyAccessor->setValue($component, $property, $value);
                    unset($data[$property]);
                }
            }
+        }
```

smnandre:
Fabbot wrong, we can move on with this one :)

Kocal (approved):

PR #2376, opened 2024-11-14
[LiveComponent] Fix form elements unsynced on load (need confirm?)

### LINKED ISSUE #1958: [LiveComponent] `ComponentWithFormTrait` out-of-sync with browsers’ changes
While working with [`symfonycasts/dynamic-forms`](https://github.com/SymfonyCasts/dynamic-forms) I observed two issues which seem to stem from the same cause:

- A not-`expanded` `ChoiceType` without placeholder whose value is `null` will appear as a `<select>` with the first `<option>` selected. The state of the component is then out-of-sync.
- When refreshing a page with a form, browsers will populate fields with the value the user previously changed. As this happens without any user interaction, the component won’t be aware of these changes and will keep its default state.

The first issue can be worked around by adding a placeholder, and the second by setting to the fields which must keep their value the `autocomplete` attribute to `off`.

Maybe the component could react to the `load` event to update its state?

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1958
| License       | MIT

Let's **_try_** to fix #1958

Multiple iterations probably needed here

### DIFF
--- a/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts
@@ -30,7 +30,12 @@ export default class implements PluginInterface {
      * the "firstName" model.
      */
     private synchronizeValueOfModelFields(component: Component): void {
-        component.element.querySelectorAll('[data-model]').forEach((element: Element) => {
+        const modelFields = [
+            ...Array.from(component.element.querySelectorAll('[data-model]')),
+            ...Array.from(component.element.querySelectorAll('form,input,select,textarea')),
+        ];
+
+        modelFields.forEach((element: Element) => {
             if (!(element instanceof HTMLElement)) {
                 throw new Error('Invalid element using data-model.');
             }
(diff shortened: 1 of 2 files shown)


src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts
```diff
@@ -30,7 +30,12 @@ export default class implements PluginInterface {
      * the "firstName" model.
      */
     private synchronizeValueOfModelFields(component: Component): void {
-        component.element.querySelectorAll('[data-model]').forEach((element: Element) => {
+        const modelFields = [
+            ...Array.from(component.element.querySelectorAll('[data-model]')),
+            ...Array.from(component.element.querySelectorAll('form,input,select,textarea')),
+        ];
```
Kocal, line 36: No?
```suggestion
        const modelFields = component.element.querySelectorAll('[data-model],form,input,select,textarea')
```

Kocal (approved):

smnandre:
If people want to help, you can check this change on your project and tell here if everything OK

Kocal:

Thanks!

Status: Needs Work

PR #2377, opened 2024-11-14
[Autocomplete] Fix handling of associated properties in DQL joins

### LINKED ISSUE #2370: [Autocomplete] (Searchable Fields): Incorrect search field selection due to similar field names
## Description:
In the MyAutocompleteField implementation, the searchable_fields option includes translations.name and foo.translations.name. However, when both fields share the same name (name), the autocomplete search only uses the first occurrence (translations.name). If the order is reversed, the search instead targets foo.translations.name, which leads to inconsistent search results depending on the field order.

This behavior impacts the usability of the autocomplete functionality, as it prevents the search from working properly across both related entities.

## Steps to reproduce:
    Define searchable_fields as ['translations.name', 'foo.translations.name'].
    Trigger the autocomplete field search.
    Notice that only translations.name is used in the query.
    Reverse the field order to ['foo.translations.name', 'translations.name'].
    Now, only foo.translations.name is considered in the search.

## Expected behavior:

The autocomplete should search across both translations.name and foo.translations.name fields, regardless of their order in searchable_fields.

## Actual behavior:

The search query only considers the first occurrence of name in searchable_fields, ignoring the second field.
Proposed Solution:

Consider updating the query logic to handle multiple fields with the same name explicitly. This could involve aliasing the fields or extending the searchable_fields handling to ensure both translations.name and foo.translations.name are included in the search query.

## Environment:

    Symfony UX Autocomplete
    Symfony Form Component
    PHP 8.3
    
## Code :

```php

namespace App\Form\AutocompleteField\MyAutocomplete;

use Override;
use Symfony\Component\Form\ChoiceList\ChoiceList;
use Symfony\Component\HttpFoundation\RequestStack;
use Symfony\UX\Autocomplete\Form\AsEntityAutocompleteField;
use Symfony\UX\Autocomplete\Form\BaseEntityAutocompleteType;

use function sprintf;

#[AsEntityAutocompleteField(route: 'ux_entity_autocomplete')]
final class MyAutocompleteField extends AbstractType
{
    #[Override]
    public function configureOptions(OptionsResolver $optionsResolver): void
    {
        $optionsResolver
            ->setDefaults([
                'class' => MyEntity::class,
                'placeholder' => 'Choices',
                'searchable_fields' => ['translations.name', 'foo.translations.name'],
                'choice_label' => ChoiceList::label($this, $this->getChoiceLabel(...)),
                'query_builder' => ...,
            ]);
    }

    #[Override]
    {
    }

    public function getChoiceLabel(): string
    {
        return ...;
    }
}
```

HugoSEIGLE:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2370
| License       | MIT

Changes:

- The alias generation logic has been improved to correctly handle nested associations and avoid alias collisions.
- The code was modified to ensure that the parent entity is correctly accounted for when generating aliases, which helps prevent issues in complex joins with multiple associated entities.
- A check was added to verify the existence of a parent property before using it as part of the alias. This prevents logical errors in cases where the parent entity is not directly involved in the join.

### DIFF
--- a/src/Autocomplete/src/Doctrine/EntitySearchUtil.php
+++ b/src/Autocomplete/src/Doctrine/EntitySearchUtil.php
@@ -47,6 +47,7 @@ class EntitySearchUtil
         ];
 
         $entitiesAlreadyJoined = [];
+        $aliasAlreadyUsed = [];
         $searchableProperties = empty($searchableProperties) ? $entityMetadata->getAllPropertyNames() : $searchableProperties;
         $expressions = [];
         foreach ($searchableProperties as $propertyName) {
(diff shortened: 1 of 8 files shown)


smnandre:
This seems very legit: do you think you could add a test or two ?

HugoSEIGLE:
> This seems very legit: do you think you could add a test or two ?

Thank you for the feedback! I’ve already added tests in "Symfony\UX\Autocomplete\Tests\Functional\FieldAutocompleterTest". Let me know if there are fix to add or additional scenarios you'd like me to cover.

smnandre (approved):
Hey @HugoSEIGLE!

Thank you for the added tests! 

To be honest I do feel this entire class deserves a big clean / refacto (code style, performance, and code readability)..

But this is way outside the scope of this PR.. that does fix a bug.. so thank you for this and we'll try to refactor it later.

:)

Kocal (approved):

Kocal:
Thanks Hugo :)

PR #2384, opened 2024-11-16
[TwigComponent] Add tip about `make:twig-component` command

DennisdeBest:
Add information on how to use the maker bundle to craete twig components, especially on how to create them in a subdirectory

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no

#SymfonyHackday :)

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -122,6 +122,26 @@ and any other components by running:
 
 Take a moment to fist pump - then come back!
 
+You can also use the ``make`` commands to generate the component PHP and twig files :
+
+.. code-block:: terminal
+
+    $ php bin/console make:twig-component Alert
+
+If you want your component to reside in a sub directory like ``src/twig/Components/Alert/`` and make a specific component for a Danger alert you can run:
+
+.. code-block:: terminal
+
+    $ php bin/console make:twig-component Alert\\Danger
+
+Or:
+
+.. code-block:: terminal
+
+    $ php bin/console make:twig-component 'Alert\Danger'
+
+This wil render the php file at ``src\twig\Components\Alert\Danger.php`` and the twig file at ``templates\components\Alert\Danger.html.twig``
+
 .. _naming:
 
 Naming Your Component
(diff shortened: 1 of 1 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -122,6 +122,26 @@ and any other components by running:
...
+    $ php bin/console make:twig-component Alert\\Danger
+
+Or:
+
+.. code-block:: terminal
+
+    $ php bin/console make:twig-component 'Alert\Danger'
+
+This wil render the php file at ``src\twig\Components\Alert\Danger.php`` and the twig file at ``templates\components\Alert\Danger.html.twig``
+
```
Kocal:
```suggestion

.. tip

    If you use the Symfony Maker Bundle (...):

    .. code-block:: terminal

        $ php bin/console make:twig-component Alert
```

Is enough, feel free to complete the first sentence

Kocal (changes requested):
Hi @DennisdeBest, while I find this addition very nice, I think it's too much verbose, the Twig Component documentation is already **so long**!

A small paragraph inside a `.. tip` with a simple example and a link to the [MakerBundle](https://symfony.com/bundles/SymfonyMakerBundle/current/index.html) would be enough. 
We do not need to:
- tell about creating a component in a sub-directory, this should be documented in Maker's documentation
- tell what and where files were created, since the command already display this information

Thanks!

src/TwigComponent/doc/index.rst
smnandre: I'd like this block to be later in the doc... maybe with just a link here pointing at this block ? 

I feel it's a bit early to tell people to run "make component" when the full concept, the paths, etc... are not yet explained

Kocal: I'm not sure to understand your point @smnandre, since above this block we already speak about the PHP file and the Twig file: 
image

Kocal (approved):

Kocal:
Thank you @DennisdeBest.

PR #2385, opened 2024-11-17
[Map] Make UX Map compatible with Live Components (and some internal things)

Kocal:
Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

Here’s a corrected and polished version of your text:

---

Hi! 😊

This PR enhances the UX Map to be compatible with Live Components, allowing you to interact with the `Map` directly from your PHP code. You can perform actions and see your map update in real-time from the front-end!

To achieve this, I had to refactor and improve several areas:

1. Due to the hydration and de-hydration processes of Live Components, I ensured that the `toArray` and new `fromArray` methods for `Map`, `Marker`, and all other value objects could rebuild an equivalent object accurately (e.g., `$marker == Marker::fromArray($marker->toArray())`).
2. Since Stimulus monitors value changes, it wasn’t efficient to store everything in a single large `view` object containing `zoom`, `center`, `markers`, etc. These properties were split into individual values to improve performance.
3. Before rendering the map, an `@id` is now automatically computed for each `Marker` and `Polygon` definition. This optimization makes rendering significantly more efficient, avoiding the need to remove and re-render all markers or polygons (and avoiding a visual glitch aswell).

 
https://github.com/user-attachments/assets/c151e64c-7321-46d3-a5c6-cfeaf57beb47

https://github.com/user-attachments/assets/ffb12035-a0c3-48b3-afb7-ddf2c24c318a


Kocal:
Note: Usages and examples can be a bit limited for the moment, but for a next PR I would like to add more methods like `getZoom()` or `removeMarker()` (...)

src/Map/assets/src/abstract_map_controller.ts
```diff
@@ -166,4 +178,50 @@ export default abstract class<
...
+                if (!this.markers.has(marker['@id'])) {
+                    this.createMarker(marker);
+                }
+            });
+
+            if (this.fitBoundsToMarkersValue) {
+                this.doFitBoundsToMarkers();
+            }
+        }
+    }
```
smnandre:
```suggestion
    public markersValueChanged(): void {
        if (!this.map) return;

        const currentKeys = new Set(this.markers.keys());

        this.markersValue.forEach((marker) => {
            if (!this.markers.has(marker['@id'])) {
                this.createMarker(marker);
            }
            currentKeys.delete(marker['@id']);
        });

        currentKeys.forEach((markerId) => {
            this.removeMarker(this.markers.get(markerId));
            this.markers.delete(markerId);
        });

        if (this.fitBoundsToMarkersValue) {
            this.doFitBoundsToMarkers();
        }
    }
```

?

Kocal: Yes, totally, I thought of that too when re-reading the PR :p

src/Map/doc/index.rst
```diff
@@ -313,6 +313,96 @@ Then, you can use this controller in your template:
...
+            return (new Map())
+                ->center(new Point(48.8566, 2.3522))
+                ->zoom(7)
+                ->addMarker(new Marker(position: new Point(48.8566, 2.3522), title: 'Paris', infoWindow: new InfoWindow('Paris')))
+                ->addMarker(new Marker(position: new Point(45.75, 4.85), title: 'Lyon', infoWindow: new InfoWindow('Lyon')))
+            ;
+        }
+    }
+
+Then, you can render the map with ``ux_map()`` in your template:
```
smnandre:
```suggestion
Then, you can render the map with ``ux_map()`` in your component template:
```

src/Map/src/Exception/UnableToDenormalizeOptionsException.php
```diff
@@ -0,0 +1,32 @@
...
+final class UnableToDenormalizeOptionsException extends LogicException
+{
+    public function __construct(string $message)
+    {
+        parent::__construct(\sprintf('Unable to denormalize the map options: %s', $message));
+    }
+
+    public static function missingProviderKey(string $key): self
+    {
+        return new self(\sprintf('the provider key ("%s") is missing in the normalized options.', $key));
```
smnandre:
```suggestion
        return new self(\sprintf('the provider key "%s" is missing in the normalized options.', $key));
```

src/Map/src/Exception/UnableToDenormalizeOptionsException.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public static function unsupportedProvider(string $provider, array $supportedProviders): self
+    {
+        return new self(\sprintf('the provider "%s" is not supported. Supported providers are "%s".', $provider, implode('", "', $supportedProviders)));
```
smnandre, line 30:
```suggestion
        return new self(\sprintf('The provider "%s" is not supported. Supported providers are "%s".', $provider, implode('", "', $supportedProviders)));
```

src/Map/src/Exception/UnableToNormalizeOptionsException.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        parent::__construct(\sprintf('Unable to normalize the map options: %s', $message));
+    }
+
+    /**
+     * @param class-string<MapOptionsInterface> $classOptions
+     */
+    public static function unsupportedProviderClass(string $classOptions): self
+    {
+        return new self(\sprintf('the class "%s" is not supported.', $classOptions));
```
smnandre:
```suggestion
        return new self(\sprintf(The class "%s" is not supported.', $classOptions));
```

src/Map/src/Live/ComponentWithMapTrait.php
```diff
@@ -0,0 +1,74 @@
...
+
+    abstract protected function instantiateMap(): Map;
+
+    public function getMap(): Map
+    {
+        if (null === $this->map) {
+            $this->map = $this->instantiateMap();
+        }
+
+        return $this->map;
```
smnandre:
```suggestion
        return $this->map ??= $this->instantiateMap();
```

Or is it a matter of PHP compat ?

src/Map/src/Live/ComponentWithMapTrait.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * @author Hugo Alliaume <hugo@alliau.me>
+ */
+trait ComponentWithMapTrait
+{
+    /**
+     * @internal
+     */
+    #[LiveProp(hydrateWith: 'hydrateMap', dehydrateWith: 'dehydrateMap')]
```
smnandre, line 31:
```suggestion
    #[LiveProp(hydrateWith: 'hydrateMap', dehydrateWith: 'dehydrateMap')]
```
Should be  private i think, and we maybe should create a dedicated hydration extension here.. wdyt ?

src/Map/src/MapOptionsInterface.php
```diff
@@ -17,6 +17,13 @@
 interface MapOptionsInterface
 {
     /**
+     * @internal
+     */
+    public static function fromArray(array $array): self;
```
smnandre, line 22: We do need it I imagine ? 

Not a big fan of leaking internal methods in interfaces

src/Map/src/MapOptionsNormalizer.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    private const string KEY_PROVIDER = '@provider';
+
+    /**
+     * @var array<string, class-string<MapOptionsInterface>>
+     */
+    public static array $providers = [
+        'google' => MapBridge\Google\GoogleOptions::class,
+        'leaflet' => MapBridge\Leaflet\LeafletOptions::class,
+    ];
```
smnandre, line 37: What about a constant ?

src/Map/src/MapOptionsNormalizer.php
```diff
@@ -0,0 +1,68 @@
...
+
+        if (!isset(self::$providers[$provider])) {
+            throw UnableToDenormalizeOptionsException::unsupportedProvider($provider, array_keys(self::$providers));
+        }
+
+        unset($array[self::KEY_PROVIDER]);
+
+        $class = self::$providers[$provider];
+
+        return $class::fromArray($array);
```
smnandre:
```suggestion
        unset($array[self::KEY_PROVIDER]);

        if (null === $class = self::$providers[$provider] ?? null) {
            throw UnableToDenormalizeOptionsException::unsupportedProvider($provider, array_keys(self::$providers));
        }

        return $class::fromArray($array);
```

src/Map/src/MapOptionsNormalizer.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $provider = array_search($options::class, self::$providers, true);
+        if (!\is_string($provider)) {
+            throw UnableToNormalizeOptionsException::unsupportedProviderClass($options::class);
+        }
+
+        $array = $options->toArray();
+        $array[self::KEY_PROVIDER] = $provider;
+
+        return $array;
```
smnandre, line 64:
```suggestion
        return [
            ...$options->toArray(), 
            self::KEY_PROVIDER => $provider,
        ];
```

src/Map/src/Exception/UnableToDenormalizeOptionsException.php
Kocal: (the "the" is on purpose, see class constructor)

src/Map/src/Renderer/AbstractRenderer.php
```diff
@@ -81,4 +81,21 @@ final public function renderMap(Map $map, array $attributes = []): string
...
+
+        $attrs = $map->toArray();
+
+        foreach ($attrs['markers'] as $key => $marker) {
+            $attrs['markers'][$key]['@id'] = $computeId($marker);
+        }
+
+        foreach ($attrs['polygons'] as $key => $polygon) {
+            $attrs['polygons'][$key]['@id'] = $computeId($polygon);
+        }
```
smnandre, line 97: If the idea is to have reproductible identifiers, i think some more normalization would be required, no ?

src/Map/src/MapOptionsNormalizer.php
Kocal: Won't be useful at all

src/Map/src/Renderer/AbstractRenderer.php
```diff
@@ -81,4 +81,21 @@ final public function renderMap(Map $map, array $attributes = []): string
 
         return \sprintf('<div %s></div>', $stimulusAttributes);
     }
+
+    private function getMapAttributes(Map $map): array
+    {
+        $computeId = fn (array $array) => hash('xxh3', json_encode($array));
```
smnandre: Not sure if a method is required here ?

smnandre:
```suggestion
        $computeId = fn (array $array) => hash('xxh3', json_encode($array, JSON_THROW_ON_ERROR));
```

src/Map/src/Exception/UnableToNormalizeOptionsException.php
```diff
@@ -0,0 +1,32 @@
...
+
+final class UnableToNormalizeOptionsException extends LogicException
+{
+    public function __construct(string $message)
+    {
+        parent::__construct(\sprintf('Unable to normalize the map options: %s', $message));
+    }
+
+    /**
+     * @param class-string<MapOptionsInterface> $classOptions
```
smnandre: I'm having a bit of a freeze every time i read this class 😅 

What do you think we rename it optionsClass everywhere ?

src/Map/doc/index.rst
```diff
@@ -313,6 +313,96 @@ Then, you can use this controller in your template:
...
+
+.. code-block:: html+twig
+
+    <div{{ attributes.defaults() }}>
+        {{ ux_map(map, { style: 'height: 300px' }) }}
+    </div>
```
smnandre:
```suggestion
.. code-block:: html+twig

    <div{{ attributes }}>
        {{ ux_map(map, {style: 'height: 300px'}) }}
    </div>
```

src/Map/src/Exception/UnableToNormalizeOptionsException.php
Kocal: Sure!

src/Map/src/Live/ComponentWithMapTrait.php
Kocal: I put it public otherwise the Live hydration process was not working due to visibility issue. Maybe I can try to add a setter?

src/Map/src/Live/ComponentWithMapTrait.php
Kocal: (copy-pasted from ComponentWithFormTrait)

src/Map/src/MapOptionsNormalizer.php
Kocal: No need to create a new array again IMO

src/Map/src/MapOptionsInterface.php
Kocal: I understand your point, but that's fine if we have `@internal` no?

src/Map/src/Renderer/AbstractRenderer.php
Kocal: As you mentioned in https://github.com/symfony/ux/pull/2385/files#r1851075522, we may want to add more logic to compute the identifier, so I've laid the foundations for the future :)

src/Map/src/Renderer/AbstractRenderer.php
Kocal: That's possible! I think you mean about attributes ordering like it's done in LiveComponent hydration/checksum processes?

Well, since we (de)normalize with fromArray and toArray, the array structure is identical, so I believe it is not necessary...?

src/Map/src/MapOptionsInterface.php
smnandre: I mean, interface and internal are a bit contradictory no ?

src/Map/src/Live/ComponentWithMapTrait.php
smnandre: Let's discuss all this tomorrow :)

src/Map/src/MapOptionsNormalizer.php
smnandre: https://3v4l.org/Sk47Z#v8.3.14

If this array is readonly, let's use a const no ?

src/Map/src/Renderer/AbstractRenderer.php
smnandre: Ok! Just curious about it... and if needed as you said we now how to implement it :)

src/Map/doc/index.rst
```diff
@@ -313,6 +313,96 @@ Then, you can use this controller in your template:
...
+--------------------------
+
+.. versionadded:: 2.22
+
+    The ability to render and interact with a Map inside a Live Component was added in Map 2.22.
+
+To use a Map inside a Live Component, you need to use the ``ComponentWithMapTrait`` trait
+and implement the method ``instantiateMap`` to return a ``Map`` instance.
+
+You can interact with the Map by using `LiveAction`
```
smnandre:
```suggestion
You can interact with the Map by using ``LiveAction``
```

src/Map/src/Exception/UnableToDenormalizeOptionsException.php
smnandre: 😮‍💨  sorry

src/Map/src/Live/ComponentWithMapTrait.php
```diff
@@ -0,0 +1,70 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Map\Live;
+
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\Map\Map;
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+use Symfony\UX\TwigComponent\Attribute\PostMount;
```
smnandre, line 19: Can we do without at first or is it required ? (genuine question)

src/Map/src/MapOptionsNormalizer.php
smnandre: It does not really ... 

I think anywhere you use merge or + or send the array back, spread is quicker. will check again :)

src/Map/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 -   Add method `Symfony\UX\Map\Renderer\AbstractRenderer::tapOptions()`, to allow Renderer to modify options before rendering a Map.
 -   Add `ux_map.google_maps.default_map_id` configuration to set the Google ``Map ID``
+-   Add compatibility with [Live Components](https://symfony.com/bundles/ux-live-component/current/index.html), for the moment only zoom, center, markers and polygons are supported.
```
smnandre:
```suggestion
-   Add `ComponentWithMapTrait` to ease maps integration in  LiveComponent.
```

At least, i'd remove the "for the moment..." as it does not bring real value and give some negatiev tonality to the entry :)

smnandre (approved):
Nice!!

src/Map/src/MapOptionsNormalizer.php
Kocal: Ah! I thought you were suggesting using constants for the keys on the board.

Well, the array is supposed to be readonly yes, **but** it is modified in tests to register the `\Symfony\UX\Map\Tests\DummyOptions` class, and reflection does not help here.

And, thanks for your benchmark! It looks like const is faster than a static property with a lot of iterations, **but** if you reduce the number to 10, 100, or 200, the static property is faster than the const 🤔

src/Map/doc/index.rst
Kocal: RST syntax 😠

src/Map/src/MapOptionsInterface.php
Kocal: I see what you mean, but it seems correct to me, as I want to force developers to implement `fromArray`/`toArray` if someone decide to implement a new bridge.

UX users are not supposed to see `fromArray` and `toArray` methods, they are internal to UX Map, so I believe it's fine?

src/Map/src/MapOptionsNormalizer.php
Kocal: You make me curious:
- if I read this code, we are creating a new array from a spreaded array and a new entry
- if I read my previous code, we re-use the array previously created and add a new entry to it

But maybe I'm wrong and biased due to how this feature works in the JavaScript world (OFC depending of the runtime etc, but I saw many projects improve performances by not using object spreading!).

Do you think an optimization is done by PHP?

src/Map/src/Live/ComponentWithMapTrait.php
Kocal: I think it's nice to keep it

src/Map/src/MapOptionsNormalizer.php
smnandre: Well then... it's a good sign it's way to early for micro optimization :)

src/Map/src/Live/ComponentWithMapTrait.php
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @author Hugo Alliaume <hugo@alliau.me>
+ */
+trait ComponentWithMapTrait
```
smnandre:
```suggestion
/**
 * @author Hugo Alliaume <hugo@alliau.me>
 *
 * @experimental
 */
trait ComponentWithMapTrait
```

smnandre (approved):

PR #2390, opened 2024-11-19
[Translator] Handle W3C locale format on document element

### LINKED ISSUE #2378: [Translator] No translations if detected via HTML lang attribute
Symfony locales are specified with an underscore to split country code from language code, e.g. `de_AT`.

When using UX Translator it correctly detects the current locale if set [via the HTML attribute](https://symfony.com/bundles/ux-translator/current/index.html#configuring-the-default-locale):

```twig
<html lang="{{ app.request.locale }}">
{# <html lang="de_AT"> #}
```

Unfortunately, [according to W3](https://www.w3.org/TR/html401/struct/dirlang.html#h-8.1.1), this format is invalid, as the "subcode" has to be separated by a hyphen.

Using

```twig
<html lang="{{ app.request.locale | replace({'_': '-'}) }}">
{# <html lang="de-AT"> #}
```

doesn't work, because this format is not detected by Translator, resulting in no translations at all (only the string constants are returned).

aleho:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2378
| License       | MIT

According to [W3C](https://www.w3.org/TR/html401/struct/dirlang.html#h-8.1.1
) (and [RFC1766](https://www.ietf.org/rfc/rfc1766.txt)) the valid format for language codes is "primary-code"-"subcode", but Symfony expects underscores as separator, resulting in broken translations.

This changes language codes specified in the correct format for HTML to the Symfony format when reading the "lang" attribute.

### DIFF
--- a/src/Translator/assets/test/translator.test.ts
+++ b/src/Translator/assets/test/translator.test.ts
@@ -35,6 +35,18 @@ describe('Translator', () => {
         });
     });
 
+    describe('getLocale', () => {
+        test('with subcode', () => {
+            // allow format according to W3C
+            document.documentElement.lang = 'de-AT';
+            expect(getLocale()).toEqual('de_AT');
+
+            // or "incorrect" Symfony locale format
+            document.documentElement.lang = 'de_AT';
+            expect(getLocale()).toEqual('de_AT');
+        });
+    });
+
     describe('setLocale', () => {
         test('custom locale', () => {
             setLocale('fr');
--- a/src/Translator/doc/index.rst
+++ b/src/Translator/doc/index.rst
@@ -90,9 +90,9 @@ Configuring the default locale
 
 By default, the default locale is ``en`` (English) that you can configure through many ways (in order of priority):
 
-#. With ``setLocale('your-locale')`` from ``@symfony/ux-translator`` package
-#. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute
-#. Or with ``<html lang="your-locale">`` attribute
+#. With ``setLocale('de')`` or ``setLocale('de_AT')`` from ``@symfony/ux-translator`` package
+#. Or with ``<html data-symfony-ux-translator-locale="{{ app.request.locale }}">`` attribute (e.g., ``de_AT`` or ``de`` using Symfony locale format)
+#. Or with ``<html lang="{{ app.request.locale|replace({ '_': '-' }) }}">`` attribute (e.g., ``de-AT`` or ``de`` following the `W3C specification on language codes`_)
 
 Detecting missing translations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -191,3 +191,4 @@ https://symfony.com/doc/current/contributing/code/bc.html
 .. _`the Symfony UX initiative`: https://ux.symfony.com/
 .. _StimulusBundle configured in your app: https://symfony.com/bundles/StimulusBundle/current/index.html
 .. _`ICU Message Format`: https://symfony.com/doc/current/reference/formats/message_format.html
+.. _`W3C specification on language codes`: https://www.w3.org/TR/html401/struct/dirlang.html#h-8.1.1
--- a/src/Translator/CHANGELOG.md
+++ b/src/Translator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.22.0
+
+-   Support both the Symfony format (`fr_FR`) and W3C specification (`fr-FR`) for locale subcodes.
+
 ## 2.20.0
 
 -   Add `throwWhenNotFound` function to configure the behavior when a translation is not found.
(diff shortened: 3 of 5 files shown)


src/Translator/doc/index.rst
```diff
@@ -91,8 +91,8 @@ Configuring the default locale
 By default, the default locale is ``en`` (English) that you can configure through many ways (in order of priority):
 
 #. With ``setLocale('your-locale')`` from ``@symfony/ux-translator`` package
-#. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute
-#. Or with ``<html lang="your-locale">`` attribute
+#. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute (e.g., `en_US`)
+#. Or with ``<html lang="your-locale">`` attribute (e.g., `en-US`)
```
Kocal: To get closer to reality, I think we can modify the examples to use `app.request.locale`:
```suggestion
#. Or with ``<html data-symfony-ux-translator-locale="{{ app.request.locale }}">`` attribute (e.g., `en_US`)
#. Or with ``<html lang="{{ app.request.locale|replace({ '_': '-' }) }}">`` attribute (e.g., `en-US`)
```

Kocal (changes requested):
Hi @aleho, and thank you for taking the subject!

This is a direct BC break for people using `<html lang="{{ app.request.locale }}">` with a `_` inside the locale, **but** the component **is still experimental**, good news! It means that we can break things :)

However, it miss the CHANGELOG update, and we can improve the documentation a little bit.

Also, can you update the test `getLocale / default locale` to ensure that `expect(getLocale()).toEqual('fr_FR')` when `document.documentElement.lang = 'fr-FR';`?

Thanks!

src/Translator/doc/index.rst
smnandre: Maybe use the occasion to update the previous line too ? 

```
- #. With ``setLocale('your-locale')`` from ``@symfony/ux-translator`` package
+ #. With ``setLocale('your_locale')`` from ``@symfony/ux-translator`` package
```

wdyt ?

Kocal: Let's directly use a real locale, short and long forms

aleho:
> This is a direct BC break for people using `<html lang="{{ app.request.locale }}">` with a `_` inside the locale

Is it? Did I miss something? In my tests it didn't matter whether I used `de`, `de_DE` or `de-DE`, all work with this PR. (See new tests.)

What doesn't work is using `en(-|_)US` if there's no `en_US` configured in Symfony. Neither does `de(-|_)AT` if only `de` and `de_DE` are configured. (But that's unrelated to my PR, it wouldn't fallback to the available primary code before either.)


> However, it miss the CHANGELOG update

If this isn't a BC break, does it still need a changelog entry? Or can this be a bug fix?


Now with tests. Sorry about that, completely forgot to commit them.

src/Translator/doc/index.rst
```diff
@@ -90,9 +90,9 @@ Configuring the default locale
 
 By default, the default locale is ``en`` (English) that you can configure through many ways (in order of priority):
 
-#. With ``setLocale('your-locale')`` from ``@symfony/ux-translator`` package
-#. Or with ``<html data-symfony-ux-translator-locale="your-locale">`` attribute
-#. Or with ``<html lang="your-locale">`` attribute
+#. With ``setLocale('de')`` or ``setLocale('de_AT')`` from ``@symfony/ux-translator`` package
+#. Or with ``<html data-symfony-ux-translator-locale="{{ app.request.locale }}">`` attribute (e.g., `de_AT` or `de` for Symfony locale format)
+#. Or with ``<html lang="{{ app.request.locale|replace({ '_': '-' }) }}">`` attribute (e.g., `de-AT` or `de` for W3C spec compliant format)
```
```suggestion
#. Or with ``<html data-symfony-ux-translator-locale="{{ app.request.locale }}">`` attribute (e.g., ``de_AT`` or ``de`` using Symfony locale format)
#. Or with ``<html lang="{{ app.request.locale|replace({'_': '-'}) }}">`` attribute (e.g., ``de`` or ``de-AT`` or  following HTML specifications)
```

wdyt ?

smnandre:
> If this isn't a BC break, does it still need a changelog entry? Or can this be a bug fix?

Almost a feature here :) 


```diff
# CHANGELOG

+ ## 2.22.0
+ 
+ -  Support both "fr-FR" and "fr_FR" format in html lang attribute

## 2.20.0

```

Something like this maybe ? (wording to illustrate only 😅 )

src/Translator/assets/test/translator.test.ts
```diff
@@ -35,6 +35,18 @@ describe('Translator', () => {
         });
     });
 
+    describe('getLocale', () => {
+        test('with subcode', () => {
+            // allow format according to W3
```
Kocal:
```suggestion
            // allow format according to W3C
```

Kocal (approved):
Good to me after applying @smnandre suggestion

aleho:
@smnandre @Kocal Thanks!

Kocal:
Thank you @aleho.

PR #2397, opened 2024-11-20
[LiveComponent] (Dynamic Form) Do not clear missing values on submit

### LINKED ISSUE #2369: [LiveComponent] (Dynamic Form) Default values disappear for dynamically added fields
I have a LiveComponent for a form. This form has dependent fields which have been implemented with the [dynamic-forms bundle](https://github.com/SymfonyCasts/dynamic-forms). I have a situation where I have select fields that are only added when certain other fields have been filled in. These select fields should have a preselected value. To do this, I set the value in the corresponding fields in the object for the initial form values. The problem is that these values disappear when the fields are added and therefore no value is preselected in the fields. There is already an [issue](https://github.com/SymfonyCasts/dynamic-forms/issues/26) for this in the dynamic-forms bundle. I have now debugged it and found out what the problem is. This is due to the `$clearMissing` parameter of the `Symfony\Component\Form::submit` function. This parameter is `true` by default. The function is called [here](https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/ComponentWithFormTrait.php#L153) without the 2nd parameter. If I set the 2nd parameter to `false`, the problem is solved. I could create a PR for it now, but unfortunately I can't assess whether this would have unwanted side effects. Can anyone say? Is it ok to set the `$clearMissing` parameter to `false` in the `ComponentWithFormTrait` on submit?

momocode-de:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2369  / https://github.com/SymfonyCasts/dynamic-forms/issues/26
| License       | MIT

If you have a LiveComponent for a form with dependent fields which have been implemented with the [dynamic-forms bundle](https://github.com/SymfonyCasts/dynamic-forms), it can happen that the pre-defined value disappears for fields that are displayed later. This is due to the `$clearMissing` parameter of the `Symfony\Component\Form::submit` function. This parameter is `true` by default and therefore I now set it to “false” so that the values are not cleared.

I can't assess whether this would have unwanted side effects. Please check it out and if you think it fits, you can merge it.

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -150,7 +150,7 @@ trait ComponentWithFormTrait
         }
 
         $form = $this->getForm();
-        $form->submit($this->formValues);
+        $form->submit($this->formValues, false);
         $this->shouldAutoSubmitForm = false;
 
         if ($validateAll) {


smnandre:
Thank you for this PR @momocode-de! 

I took time to look at it and one question arises: does this only change solves your DynamicFormBundle problem ? Or is this only a step ? 

Right now i'm a bit hesitant here...  if i(m not reading this badly, it would be a strong behaviour change: 

> When the second parameter $clearMissing is false, like with the "PATCH" method, the validation will only apply to the submitted fields. If you need to validate all the underlying data, add the required fields manually so that they are validated:

> When submitting a form via a "PATCH" request, you may want to update only a few submitted fields. To achieve this, you may pass an optional second boolean argument to submit(). Passing false will remove any missing fields within the form object. Otherwise, the missing fields will be set to null.

https://symfony.com/doc/current/form/direct_submit.html

At least we would need some tests here, and a clear explanation of what would happen in the different scenario with some honest before/after.

Right now (and i'm very open to change my position)... i don't know... maybe you should first try to change this on your side? Because i'm not sure we should take this risk.. What do you think ?

stehled:
Hi @momocode-de, does this really resolve the issue? I've come across this problem myself and tried locally your changes and unfortunately this solve didn't work for me. I understand we both have different environments, but this simple code snippet is still affected. Could you please provide an example of code that was fixed by your change? Maybe there were more steps to resolve the issue which I don't see.

```
$builder
  ->addDependent(
      name: 'value',
      dependencies: ['amount'],
      callback: function (DependentField $field, ?float $amount) {
          $field->add(
              type: NumberType::class,
              options: [
                  'required' => true,
                  'scale' => 2,
                  'disabled' => !$amount,
                  'data' => $amount ?? 0,
              ],
          );
      }
  )
```

cklm:
I just opened https://github.com/SymfonyCasts/dynamic-forms/issues/43, which is related here. I can confirm, that `$form->submit($this->formValues, false);` does solve my problem there.

smnandre:
Well, then .. do you guys think we could add some test here showing maybe the before/after ?

Also @momocode-de if you could rebase ?

After that,LGTM! 

(let's also check/test as much possible in real conditions on the `2.x-dev` branch before the release)

BorbagUskada:
For your information, I have a very similar problem, just that I don't use SymfonyCasts/dynamic-forms, 

My default value for every dynamically-added input is not here. If I save without filling those fields, then edit the form, it is then here.

 Trying to add the `clearMissing: false` does not solve the issue, just like @steheld.

Huluti:
@smnandre just tested and this fix my issue also! (https://github.com/symfony/ux/issues/2369#issuecomment-2540984342)

Kocal:
Tests still need to be added 🙏🏻 

Status: Needs Work

tjveldhuizen:
I struggle with the same issue as @momocode-de, at the moment. In my case the change of the 'clearMissing' parameter does fix this issue. Unfortunately, my test-writing skills are insufficient to create the required tests. I understand the concerns commented by @smnandre. However: would it be a real issue that fields without submitted data are not validated, since later in the submitForm method the validation results for the not submitted fields are removed anyway?

Would it be a way forward to make the parameter configurable (without changing default), so we can at least change it in our project to find out if unexpected behavior occurs?

Jalliuz:
Just setting `clearMissing: false` can have unwanted consequences. The property is passed over to all the sub forms during a submit. Nested CollectionTypes might suddenly behave differently.

When using `clearMissing: true` any collection items not included in $data will be removed. When using `clearMissing: false` they will be preserved.

Jalliuz:
FYI: I have a quick fix for this matter

https://gist.github.com/Jalliuz/dfc7243290d7618ad9af13300c5deef7

Works as a fix, but probably not good enough to make it a pull request

Copy/paste it somewhere and just do `use LiveComponentDynamicFormTrait` in your LiveComponent

smnandre:
No response from the OP here, and the implementation doesn't work for everyone. 

If someone wants to try another approach, feel free to open a PR - we can help, especially if you're new to contributing to Symfony and want to use this as an opportunity.

momocode-de:
Since my PR solution works for me, I haven't tested @Jalliuz 's quick fix yet, but thanks anyway! Of course, it would still be nice if this were officially taken into account in the future, but unfortunately I can't contribute any more to this at the moment.

PR #2399, opened 2024-11-20
[LiveComponent] Refactor elementBelongsToThisComponent

### LINKED ISSUE #2388: [LiveComponent] Race Condition in Live Components - Premature 'render:finished' Event Before Component Registration
There seems to be a bug in `Live Components`. More specifically, in the `live_controller.js` file, there is a logic flaw that breaks the `SetValueOntoModelFieldsPlugin`.

After fetching a request, `executeMorphdom` is performed, and a few steps later, the `render:finished` event is fired. The `SetValueOntoModelFieldsPlugin` subscribes to this event and executes its logic. However, there's an assumption that between `executeMorphdom` and firing the `render:finished` event, `Stimulus` registers all `Live components` and builds the live component map (used in the `elementBelongsToThisComponent` method).

This assumption is not always true. Sometimes `Stimulus` has not finished registering every `Live component`, so the `elementBelongsToThisComponent` method does not work properly, resulting in an exception.

I suggest ensuring that all `Live components` are fully registered after `executeMorphdom` before continuing with the rest of the logic.

```
live_controller.js:397 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'accessRule')
    at parseDeepData (live_controller.js:397:1)
    at getDeepData (live_controller.js:383:42)
    at ValueStore.get (live_controller.js:427:1)
    at ValueStore.has (live_controller.js:430:1)
    at live_controller.js:2708:1
    at NodeList.forEach (<anonymous>)
    at SetValueOntoModelFieldsPlugin.synchronizeValueOfModelFields (live_controller.js:2690:1)
    at live_controller.js:2686:1
    at live_controller.js:1556:1
```

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2388 
| License       | MIT

This method is called from multiple places, sometimes before children component are instanciated and stored in the component map.

So let's back to the quickest selector we have:  the browser DOM selector :)

We'll know tomorrow, but it _could_ solve #2388

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -131,82 +131,6 @@ function getElementAsTagText(element) {
         : element.outerHTML;
 }
 
-let componentMapByElement = new WeakMap();
-let componentMapByComponent = new Map();
-const registerComponent = (component) => {
-    componentMapByElement.set(component.element, component);
-    componentMapByComponent.set(component, component.name);
-};
-const unregisterComponent = (component) => {
-    componentMapByElement.delete(component.element);
-    componentMapByComponent.delete(component);
-};
-const getComponent = (element) => new Promise((resolve, reject) => {
-    let count = 0;
-    const maxCount = 10;
-    const interval = setInterval(() => {
-        const component = componentMapByElement.get(element);
-        if (component) {
-            clearInterval(interval);
-            resolve(component);
-        }
-        count++;
-        if (count > maxCount) {
-            clearInterval(interval);
-            reject(new Error(`Component not found for element ${getElementAsTagText(element)}`));
-        }
-    }, 5);
-});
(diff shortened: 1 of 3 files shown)


Pechynho:
I can confirm that this change fixes the issue I reported in #2388. It might be helpful to add a comment about the potential race condition that could occur, so other developers are aware of it when building new plugins. For now, however, this solution seems sufficient to me.

In the future, it might be necessary to address this issue more specifically and deal with the timing problem, but for now, I would say this solution is acceptable.

Kocal (approved):

smnandre:
Thank you @Pechynho for your time and for showing me where the bug was happening, how it was happening, and for having again time to check if the solution did work.

You cannot imagine how it eases this kind of situation and how it makes a bug almost.. enjoyable :)

I 100% note your comment about the whole "optimistic timing" issue and we will keep it in mind for the incoming changes.

PR #2407, opened 2024-11-26
[Turbo] Add support for providing multiple mercure topics to `turbo_stream_listen`

### LINKED ISSUE #213: [Turbo][Mercure] Allow turbo_stream_listen() to subscribe multiple topics in single connection
Hi! First thanks for turbo and mercure integration, i am playing with it for last several days and it is awesome!

I am struggling with finding a way, how to subscribe to multiple topics using `turbo_stream_listen()` in twig and after some digging around, it seems that it is not currently supported.

I can do it this way, but i would love to avoid writing javascript in twig (as documented in https://symfony.com/doc/current/mercure.html#subscribing):
```twig
<script>
const eventSource = new EventSource("{{ mercure([
    'a',
    'b'
])|escape('js') }}");
</script>
```

But by doing this i am loosing a lot of comfort.

What i tried:
```twig
<div {{ turbo_stream_listen(['a', 'b']) }}></div>  {# Ends with error #}
<div {{ turbo_stream_listen('a,b']) }}></div>  {# Subscribes to single "a,b" topic #}
<div {{ turbo_stream_listen('a&topic=b') }}></div>  {# Subscribes to single "a&topic=b" topic #}
```

That `a&topic=b` variant was my favourite that it would work, but unfortunately it escapes the `&` so final URL is like this, which is why it is treated as single topic:
```
URL: http://localhost:8080/.well-known/mercure?topic=a%26topic%3Db
```

What i am actually trying to achieve is this hub url with single `turbo_stream_listem()`:
```
http://localhost:8080/.well-known/mercure?topic=a&topic=b
```

My current solution is to:
```twig
<div {{ turbo_stream_listen('a') }}></div>
<div {{ turbo_stream_listen('b') }}></div>
```
Which results into two connections to hub and i believe it would be more efficient to have one connection with two subscriptions.

I could probably solve this by writing my own stimulus controller (inspired by turbo_stream_controller.js).

My idea is to update `Symfony\UX\Turbo\Bridge\Mercure\TurboStreamListenRenderer`
```diff
public function renderTurboStreamListen(Environment $env, $topic): string
    {
+        if (\is_array($topic)) {
+            // .. some logic here :-)
+        }

         if (\is_object($topic)) {
```

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #213
| License       | MIT

### DIFF
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -5,6 +5,7 @@
 -   Add `<twig:Turbo:Stream>` component
 -   Add `<twig:Turbo:Frame>` component
 -   Add support for custom actions in `TurboStream` and `TurboStreamResponse`
+-   Add support for providing multiple mercure topics to `turbo_stream_listen`
 
 ## 2.21.0
 
--- a/src/Turbo/assets/dist/turbo_stream_controller.d.ts
+++ b/src/Turbo/assets/dist/turbo_stream_controller.d.ts
@@ -2,14 +2,17 @@ import { Controller } from '@hotwired/stimulus';
 export default class extends Controller {
     static values: {
         topic: StringConstructor;
+        topics: ArrayConstructor;
         hub: StringConstructor;
     };
     es: EventSource | undefined;
     url: string | undefined;
     readonly topicValue: string;
+    readonly topicsValue: string[];
     readonly hubValue: string;
     readonly hasHubValue: boolean;
     readonly hasTopicValue: boolean;
+    readonly hasTopicsValue: boolean;
     initialize(): void;
     connect(): void;
     disconnect(): void;
--- a/src/Turbo/assets/dist/turbo_stream_controller.js
+++ b/src/Turbo/assets/dist/turbo_stream_controller.js
@@ -6,12 +6,19 @@ class default_1 extends Controller {
         const errorMessages = [];
         if (!this.hasHubValue)
             errorMessages.push('A "hub" value pointing to the Mercure hub must be provided.');
-        if (!this.hasTopicValue)
-            errorMessages.push('A "topic" value must be provided.');
+        if (!this.hasTopicValue && !this.hasTopicsValue)
+            errorMessages.push('Either "topic" or "topics" value must be provided.');
         if (errorMessages.length)
             throw new Error(errorMessages.join(' '));
         const u = new URL(this.hubValue);
-        u.searchParams.append('topic', this.topicValue);
+        if (this.hasTopicValue) {
+            u.searchParams.append('topic', this.topicValue);
+        }
+        else {
+            this.topicsValue.forEach((topic) => {
+                u.searchParams.append('topic', topic);
+            });
+        }
         this.url = u.toString();
     }
     connect() {
@@ -29,6 +36,7 @@ class default_1 extends Controller {
 }
 default_1.values = {
     topic: String,
+    topics: Array,
     hub: String,
 };
 
--- a/src/Turbo/assets/src/turbo_stream_controller.ts
+++ b/src/Turbo/assets/src/turbo_stream_controller.ts
@@ -16,24 +16,34 @@ import { connectStreamSource, disconnectStreamSource } from '@hotwired/turbo';
 export default class extends Controller {
     static values = {
         topic: String,
+        topics: Array,
         hub: String,
     };
     es: EventSource | undefined;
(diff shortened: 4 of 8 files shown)


src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
```diff
@@ -21,7 +21,7 @@
 interface TurboStreamListenRendererInterface
 {
     /**
-     * @param string|object $topic
+     * @param string|object|array<string> $topic
```
smnandre: Purely per curiosity, would this not be a BC break? As (purely) hypothetical implementation would need to adapt ? Or do we ignore the docblock here?

smnandre:
Thank you @norkunas 👍 

I guess there were no elegant way to avoid the value/values duo in the controller ?

src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
norkunas: I think widening is not a BC break?

norkunas:
> Thank you @norkunas 👍 
> 
> I guess there were no elegant way to avoid the value/values duo in the controller ? 

Didn't think of a better way without breaking, later the former one could be deprecated

src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
norkunas: An alternative way could be add a new method with `@method` phpdoc, which would accept arrays only

smnandre: Arf it seems to be one https://symfony.com/doc/current/contributing/code/bc.html#changing-interfaces

smnandre: Twig function is expendable (no interface), so we could pass an array of topics here

```php
    /**
     * @param object|string $topic
     */
    public function turboStreamListen(Environment $env, $topic, ?string $transport = null): string
    {
```

smnandre: And -- hacky i admit -- if we have some TopicSet / TopicCollection object we could use it as the interface accept objects

```
    /**
     * @param string|object $topic
     */
    public function renderTurboStreamListen(Environment $env, $topic): string;
```


So we could create this object (or any ArrayObject or similar) in the TwigExtension and pass it to the renderer ?

norkunas: ArrayObject would be too wide I think as it can accept anything, TopicSet/TopicCollection works for me

norkunas: hm, i've basically done with TopicSet and additionaly wrote some tests, but then - how do I create TopicSet in twig? :thinking:

Should twig extension accept an array and convert it to TopicSet?

smnandre: That was my first idea ... but then we could even make TopicSet internal i guess ? What do you think?

If all this feels too much, we can also "go back" on your new interface/method idea, i really don't mind .. and don't want to make you lose time :)

smnandre:
phpstan can be ignored..  are the failue 8.1 related or are they our usual flakky turbo failures ? 😅

src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
norkunas: made it internal :)

norkunas:
> are the failue 8.1 related or are they our usual flakky turbo failures ? 😅

some were related due to changed escape, fixed that with feature detection

smnandre (approved):
Thank you, @norkunas! This PR will definitely make several people happy. 😊

norkunas:
> This PR will definitely make several people happy. 😊

Including myself so I can use it in my project :D

Kocal:
Thanks @norkunas, I've just pushed a minor modification and rebased, I will merge your PR when checks become green.

Kocal:
Thank you @norkunas.

norkunas:
I've hesitated to do that because php 8.1 was declared as min version and phpstorm shown incompatibility

seb-jean:
Thanks @norkunas

PR #2410, opened 2024-11-28
[Icons] use dedicated cache pool 

IndraGunawan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

based on @stof comment on EasyAdminBundle repository https://github.com/EasyCorp/EasyAdminBundle/pull/6507#discussion_r1831159693

> you should probably create a dedicated cache pool inheriting its configuration from cache.system instead of using cache.system directly. this is how core Symfony components work (allowing them to avoid having to care about conflicting keys between the caches of different component, as the key only needs to be unique inside the cache pool)

### DIFF
--- a/src/Icons/config/services.php
+++ b/src/Icons/config/services.php
@@ -24,10 +24,15 @@ use Symfony\UX\Icons\Twig\UXIconRuntime;
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
+        ->set('.ux_icons.cache')
+            ->parent('cache.system')
+            ->private()
+            ->tag('cache.pool')
+
         ->set('.ux_icons.cache_icon_registry', CacheIconRegistry::class)
             ->args([
                 service('.ux_icons.chain_registry'),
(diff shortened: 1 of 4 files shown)


Kocal (approved):

src/Icons/config/services.php
```diff
@@ -24,10 +24,15 @@
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
+        ->set('cache.ux.icons')
```
IndraGunawan: follows the naming format of twig_component cache https://github.com/symfony/ux/blob/2.x/src/TwigComponent/config/cache.php

stof: I would name this `.ux_icons.cache` to respect the best practice of prefixing service ids with the alias of the bundle for namespacing purpose.

Kocal:
Thank you @IndraGunawan.

PR #2423, opened 2024-12-04
Apply natural import order for JS to synchronise the code base with Biome and its configuration

chadyred:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| License       | MIT

Hello,

`Biome` sounds really good :scientist: , it seems to be a really cool tool, guys ! 

It applies a rule for ordering import "naturally" (https://biomejs.dev/analyzer/import-sorting/) so, I apply the command to check and fix

What about apply it "as is" ? Another alternative will be to disabled the rule. What do you think about that detail ?

Maybe, an import needs of a package just above, from another package import, this is what I propose also to disable the rule, WYT ?

Thanks ^^


Kocal (approved):
Seems nice, but by curiosity I don't see any changes about Biome.js changes, did you forgot to commit something? :)

chadyred:
Hey @Kocal,

 I didn't edit the conf (I play with it on my side :p )
 
 I exposed the fact I discovered the tool, I follow the contributing guideline, and I thinked about the rule to sort import has not been discussed / seen / treated before. So I pointed out the fact if we apply to the code base a global check then fix with actual rules of Biome config, we will apply (like in this PR) a lot of changed. If we keep Biome and its conf as is, ok, the PR is good, and the code base is synchronized with Biome conf...that's it!

Kocal:
Yeah configuring how imports are sorted is nice, not doing it is sometimes a waste of time in some projects/teams when reviewing the code, but AFAIK it didn't happen in Symfony UX.

Can you please commit the new Biome.js configuration, so we can merge your PR? :) 


Thanks!

Kocal (changes requested):
Putting a guard, waiting for Biome.js configuration to be updated 🙏🏻

chadyred:
@Kocal I did not change it, it is the default config, I just call`node_modules/.bin/biome check --fix` simply with the existing conf (https://biomejs.dev/analyzer/import-sorting/ : it is enabled by default, it is implicite)

Kocal:
@chadyred Ah!
> This feature is enabled by default but can be opted-out via configuration:

Alright, I thought it was the opposite, and expected to see Biome configuration (or commands) to be updated to enable imports sortings.

However, you used `check` command, and given https://biomejs.dev/reference/cli/ it looks like only `biome check` and `biome ci` commands sort imports 🤔 
It's a bit unfortunate because we use `biome lint` and `biome format` commands, for... historical reason (ESLint and Prettier).

You know, let's iterate, I'm merging this and will open a new PR for `biome check` and `biome ci`. 

Thanks @chadyred :)

Kocal (approved):

Kocal:
Thank you @chadyred.

chadyred:
You are welcome, thanks for your time @Kocal :green_apple:

PR #2426, opened 2024-12-05
[LiveComponent] Handle loose comparison with empty placeholder 

### LINKED ISSUE #2425: [LiveComponents] Bug with required select with empty placeholders
Hello,
    the last version (2.22.0) fixed bug #1332 that selects the first value of select fields when they are required. This works great, even with placeholders, as they are selected if present, which is the intended behaviour. Unfortunately, due to loose comparison, if there is an empty placeholder (that is, a placeholder defined as the empty string), it is detected as a field without placeholder.

The error comes from the line here : https://github.com/symfony/ux/blob/9d1dada215a9809f13c09d90be8048a8f92aa358/src/LiveComponent/src/ComponentWithFormTrait.php#L268 with :
```php
!$child->vars['placeholder']
```

Casting the placeholder to a boolean leads to the empty string being considered as an absence of placeholder, this should be tested this way I think : 
```php
$child['placeholder'] === null
```

To reproduce the bug, try a ChoiceType field that is required, with an empty placeholder (`['placeholder => '']` in the options array of the form), and see that it takes the first value when changing another field of the form.

Not a big deal, but I had to investigate why the behaviour was different in the previous version, so I might not be the only one.

FoxCie:
Fix for a bug that arises when using an empty placeholder in a required select field, that is wrongly treated as an absence of placeholder by LiveComponents.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2425  
| License       | MIT

LiveComponents test placeholders of required selects to detect when to set a default value from the choices (i.e. if the select is required and does not have a placeholder, among other conditions) by just casting the placeholder to a boolean. This leads to empty placeholders (placeholders that are the empty string) to be treated as an absence of placeholders when they should not. This MR fixes this by replacing the condition on placeholders with a more appropriate one.
According to the docs https://symfony.com/doc/current/reference/forms/types/choice.html#placeholder and https://symfony.com/doc/current/reference/forms/types/choice.html#field-variables, the placeholder can be a boolean (`false` indicates that there should not be a placeholder), a string (the text to display for the empty value) or a TranslatableMessage (same as string, but goes through a translator before being displayed). The vars property is either the value of the field, or `null` if there is no placeholder specified. From this, I considered that there is no placeholder either if the placeholder holds `false` or `null`.

I added a simple test that checks the behaviour when there is a required select with an empty placeholder ; we expect the value of the field to be empty in such cases when we do not modify it.

#SymfonyHackday :)

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -265,7 +265,7 @@ trait ComponentWithFormTrait
                 && $child->vars['required']
                 && !$child->vars['disabled']
                 && !$child->vars['value']
-                && !$child->vars['placeholder']
+                && (false === $child->vars['placeholder'] || null === $child->vars['placeholder'])
                 && !$child->vars['multiple']
                 && !$child->vars['expanded']
                 && $child->vars['choices']
--- a/src/LiveComponent/tests/Fixtures/Form/FormWithManyDifferentFieldsType.php
+++ b/src/LiveComponent/tests/Fixtures/Form/FormWithManyDifferentFieldsType.php
@@ -52,6 +52,13 @@ class FormWithManyDifferentFieldsType extends AbstractType
                 ],
                 'placeholder' => 'foo',
             ])
+            ->add('choice_required_with_empty_placeholder', ChoiceType::class, [
+                'choices' => [
+                    'bar' => 2,
+                    'foo' => 1,
+                ],
+                'placeholder' => '',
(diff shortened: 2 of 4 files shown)


Kocal (approved):
Nice catch @FoxCie, and thanks for the PR :)

Kocal:
Thank you @FoxCie.

FoxCie:
Thank you for the quick merge!

PR #2428, opened 2024-12-05
[Meta] Drop format, lint, check-format and check-lint commands, replace them with check and ci commands

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Following https://github.com/symfony/ux/pull/2423#issuecomment-2520352394.

When I've migrated from ESLint and Prettier to Biome.js, I kept commands `yarn lint`, `yarn format` (and `yarn check-`) for historical and simplicity reasons, because it was two separate tools.

_Now_ that we use a single tool, we can use its two dedicated commands:
- [`biome check`](https://biomejs.dev/reference/cli/#biome-check), like `biome format + biome lint + imports sorting`
- [`biome ci`](https://biomejs.dev/reference/cli/#biome-ci), for the CI (it does not write files)

image

cc @chadyred

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,14 +81,14 @@ To help you with assets, you can run the following commands in a specific packag
   - `yarn run build`: build (compile) assets from the package,
   - `yarn run watch`: watch for modifications and rebuild assets from the package,
   - `yarn run test`: run the tests from the package,
-  - `yarn run lint`: lint assets from the package,
-  - `yarn run format`: format assets from the package.
+  - `yarn run check`: run the formatter, linter, and sort imports, and fails if any modifications 
+  - `yarn run check --write`: run the formatter, linter, imports sorting, and write modifications 
 
 Thanks to [Yarn Workspaces](https://yarnpkg.com/features/workspaces), you can also run these commands from the root directory of the project:
   - `yarn run build`: build (compile) assets from **all** packages,
   - `yarn run test`: run the tests from **all** packages,
-  - `yarn run lint`: lint assets from **all** packages,
-  - `yarn run format`: format assets from **all** packages,
+  - `yarn run check`: run the formatter, linter, and sort imports for **all** packages, and fails if any modifications
+  - `yarn run check --write`: run the formatter, linter, imports sorting for **all** packages, and write modifications
 
 ## Useful Git commands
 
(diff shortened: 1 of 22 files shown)


chadyred:
@Kocal Re, you are quicker than me I was doing make a PR, so great to see that :)

A little thing about the "--write", it sounds as a low level of details, I found the "--fix" shorter and it is more humanized...but the "--write" is the first accepted option and "--fix" is an alias...(My :brain: swims too much with cs-fixer, maybe :D)

CONTRIBUTING.md
```diff
@@ -81,14 +81,14 @@ To help you with assets, you can run the following commands in a specific packag
   - `yarn run build`: build (compile) assets from the package,
   - `yarn run watch`: watch for modifications and rebuild assets from the package,
   - `yarn run test`: run the tests from the package,
-  - `yarn run lint`: lint assets from the package,
-  - `yarn run format`: format assets from the package.
+  - `yarn run check`: run the formatter, linter, and sort imports, and fails if any modifications 
+  - `yarn run check --write`: run the formatter, linter, imports sorting, and write modifications
```
chadyred, line 85: I forgot to update the contributing guide :+1:

Kocal: Nah, don't worry that's fine :)

Kocal:
> A little thing about the "--write", it sounds as a low level of details, I found the "--fix" shorter and it is more humanized...but the "--write" is the first accepted option and "--fix" is an alias...(My 🧠 swims too much with cs-fixer, maybe :D)

I understand your point, but I believe it can be the same scenario with `--apply` alias which was... deprecated some versions ago (I was not really happy to open https://github.com/Kocal/BiomeJsBundle/pull/9 because of https://github.com/biomejs/biome/blob/main/CHANGELOG.md#180-2024-06-04, only one month after creating the bundle 😅)

Let's keep `--write`!

Kocal:
> you are quicker than me I was doing make a PR, so great to see that :)

'just put you as co-author ;)

chadyred:
@Kocal so you are "--write" by XP, it is better ^^

smnandre:
Is it possible to have "yarn run check" and "yarn run fix" ? If not i'll do my aliases :)

Kocal:
`yarn run check --write` ?

PR #2433, opened 2024-12-07
[LiveComponent] Adds MultiStep Form Feature

silasjoisten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

This pull request introduces a new feature to Symfony UX LiveComponent that enables developers to easily create multistep forms in their applications. This functionality simplifies the implementation of complex, multi-step workflows while maintaining a clean and structured developer experience.

**Key Features**

1. MultiStep Form Type (MultiStepType): A custom form type that handles step-specific configuration, such as defining the steps and their corresponding fields.
2. ComponentWithMultiStepFormTrait: A reusable trait to manage the form flow, including:
     - Navigation between steps (next, previous, submit).
     - Validation and data persistence for each step.
     - Easy integration with storage mechanisms.
     - Storage Interface (StorageInterface): Provides a standardized way to persist form state across steps, enabling seamless navigation and data retrieval.

**Example Usage**

1. Create the LiveComponent

```php
<?php

declare(strict_types=1);

namespace App\Twig\Component;

use Symfony\UX\LiveComponent\Storage\StorageInterface;
use Symfony\UX\LiveComponent\ComponentWithMultiStepFormTrait;
use App\Form\Test\MyFancyWizardType;
use Symfony\Component\Form\FormFactoryInterface;

#[AsLiveComponent(
    name: 'SuperFancyWizard',
    template: 'components/super_fancy_wizard.html.twig',
)]
final class WizardComponent
{
    use ComponentWithMultiStepFormTrait;

    public function __construct(
        private StorageInterface $storage,
        private FormFactoryInterface $formFactory,
    ) {
    }

    public function onSubmit(): void
    {
        $data = $this->getAllData();

        // Do what ever you want with the data.

        $this->resetForm();
    }

    protected static function formClass(): string
    {
        return MyFancyWizardType::class;
    }

    protected function getFormFactory(): FormFactoryInterface
    {
        return $this->formFactory;
    }

    protected function getStorage(): StorageInterface
    {
        return $this->storage;
    }
}
```

2. Create the FormType using `getParent` to Create a custom MultiStep form type:

```php
<?php

declare(strict_types=1);

namespace App\Form;

use Symfony\UX\LiveComponent\Form\Type\MultiStepType;
use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
use Symfony\Component\Form\Extension\Core\Type\NumberType;
use Symfony\Component\Validator\Constraints\IsTrue;
use Symfony\Component\Validator\Constraints\NotBlank;

final class MyFancyWizardType extends AbstractType
{
    /**
     * @return class-string<AbstractType>
     */
    {
        return MultiStepType::class;
    }

    {
        $resolver->setDefaults([
            'steps' => [
                'general' => static function (FormBuilderInterface $builder): void {
                    $builder
                        ->add('age', NumberType::class, [
                            'label' => 'Age',
                            'constraints' => [
                                new NotBlank(),
                            ],
                        ])
                        ->add('name', TextType::class, [
                            'label' => 'Name',
                            'constraints' => [
                                new NotBlank(),
                            ],
                        ]);
                },
                'contact' => static function (FormBuilderInterface $builder): void {
                    $builder
                        ->add('email', TextType::class, [
                            'label' => 'E-Mail',
                            'constraints' => [
                                new NotBlank(),
                            ],
                        ])
                        ->add('newsletter', CheckboxType::class, [
                            'label' => 'Newsletter',
                            'constraints' => [
                                new IsTrue(),
                            ],
                        ]);
                },
                'bar' => static function (FormBuilderInterface $builder): void {
                    $builder
                        ->add('address', TextType::class, [
                            'label' => 'Address',
                            'constraints' => [
                                new NotBlank(),
                            ],
                        ])
                        ->add('city', TextType::class, [
                            'label' => 'City',
                            'constraints' => [
                                new NotBlank(),
                            ],
                        ]);
                },
            ],
        ]);
    }
}
```

```twig
<div {{ attributes }}>

    <div class="max-w-4xl mx-auto">
        {% for stepName in this.stepNames %}
            <div class="inline-flex space-x-4">
                <span class="{% if stepName == this.currentStepName %}text-red-500{% else %}text-gray-400{% endif %}">{{ stepName }}</span>
                {% if not loop.last %}
                    /
                {% endif %}
            </div>
        {% endfor %}

        {{ form(form) }}
    </div>

    <div class="max-w-4xl mx-auto my-12">
        <button
            {{ live_action('previous') }}
            type="button"
            class="btn-secondary"
            {% if this.isFirst() %}disabled="disabled"{% endif %}
        >
            Previous
        </button>

        {% if not this.isLast() %}
            <button
                {{ live_action('next') }}
                type="button"
                class="btn-primary"
            >
                Next
            </button>
        {% else %}
            <button
                {{ live_action('submit') }}
                type="button"
                class="btn-primary"
            >
                Finish
            </button>
        {% endif %}
    </div>
</div>

```

3. Create a template and benefit from awesome methods to control the MultiStep form

**Acknowledgments**

This feature was collaboratively developed during the #SymfonyHackday following SymfonyCon Vienna 2024. Special thanks to the contributors for their outstanding work:
- @PReimers
- @heahdude
- @Kocal 
- @kbond 
- @WebMamba 

Your efforts and dedication have made this feature possible!

Still on my todo list:

- [ ] Write tests
- [ ] Update docs
- [ ] Update CHANGELOG.md

**Look and feel**

CleanShot 2024-12-07 at 22 28 43

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -51,6 +51,7 @@
         "symfony/serializer": "^5.4|^6.0|^7.0",
         "symfony/twig-bundle": "^5.4|^6.0|^7.0",
         "symfony/validator": "^5.4|^6.0|^7.0",
+        "symfony/string": "^5.4|^6.0|^7.0",
         "zenstruck/browser": "^1.2.0",
         "zenstruck/foundry": "^2.0"
     },
--- a/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
@@ -1,0 +1,282 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent;
+
+use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\UX\LiveComponent\Attribute\LiveAction;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\Storage\StorageInterface;
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+use Symfony\UX\TwigComponent\Attribute\PostMount;
+use function Symfony\Component\String\u;
+
+/**
+ * @author Silas Joisten <silasjoisten@proton.me>
+ * @author Patrick Reimers <preimers@pm.me>
+ * @author Jules Pietri <jules@heahprod.com>
+ */
+trait ComponentWithMultiStepFormTrait
+{
+    use DefaultActionTrait;
+    use ComponentWithFormTrait;
+
+    #[LiveProp]
+    public ?string $currentStepName = null;
+
+    /**
+     * @var string[]
+     */
+    #[LiveProp]
+    public array $stepNames = [];
+
+    public function hasValidationErrors(): bool
+    {
+        return $this->form->isSubmitted() && !$this->form->isValid();
+    }
+
+    /**
+     * @internal
+     *
+     * Must be executed after ComponentWithFormTrait::initializeForm().
+     */
+    #[PostMount(priority: -250)]
+    public function initialize(): void
+    {
+        $this->currentStepName = $this->getStorage()->get(
+            sprintf('%s_current_step_name', self::prefix()),
+            $this->formView->vars['current_step_name'],
+        );
+
+        $this->form = $this->instantiateForm();
+
+        $formData = $this->getStorage()->get(sprintf(
+            '%s_form_values_%s',
+            self::prefix(),
+            $this->currentStepName,
+        ));
+
+        $this->form->setData($formData);
+
+        if ([] === $formData) {
+            $this->formValues = $this->extractFormValues($this->getFormView());
+        } else {
+            $this->formValues = $formData;
+        }
+
+        $this->stepNames = $this->formView->vars['steps_names'];
+
+        // Do not move this. The order is important.
+        $this->formView = null;
+    }
+
+    #[LiveAction]
+    public function next(): void
+    {
+        $this->submitForm();
+
+        if ($this->hasValidationErrors()) {
+            return;
+        }
+
+        $this->getStorage()->persist(
+            sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName),
+            $this->form->getData(),
+        );
+
+        $found = false;
+        $next = null;
+
+        foreach ($this->stepNames as $stepName) {
+            if ($this->currentStepName === $stepName) {
+                $found = true;
+
+                continue;
+            }
+
+            if ($found) {
+                $next = $stepName;
+
+                break;
+            }
+        }
+
+        if (null === $next) {
+            throw new \RuntimeException('No next forms available.');
+        }
+
+        $this->currentStepName = $next;
+        $this->getStorage()->persist(sprintf('%s_current_step_name', self::prefix()), $this->currentStepName);
+
+        // If we have a next step, we need to resinstantiate the form and reset the form view and values.
+        $this->form = $this->instantiateForm();
+        $this->formView = null;
+
+        $formData = $this->getStorage()->get(sprintf(
+            '%s_form_values_%s',
+            self::prefix(),
+            $this->currentStepName,
+        ));
+
+        // I really don't understand why we need to do that. But what I understood is extractFormValues creates
+        // an array of initial values.
+        if ([] === $formData) {
+            $this->formValues = $this->extractFormValues($this->getFormView());
+        } else {
+            $this->formValues = $formData;
+        }
+
+        $this->form->setData($formData);
+    }
+
+    #[LiveAction]
+    public function previous(): void
+    {
+        $found = false;
+        $previous = null;
+
+        foreach (array_reverse($this->stepNames) as $stepName) {
+            if ($this->currentStepName === $stepName) {
+                $found = true;
+
+                continue;
+            }
+
+            if ($found) {
+                $previous = $stepName;
+
+                break;
+            }
+        }
+
+        if (null === $previous) {
+            throw new \RuntimeException('No previous forms available.');
+        }
+
+        $this->currentStepName = $previous;
+        $this->getStorage()->persist(sprintf('%s_current_step_name', self::prefix()), $this->currentStepName);
+
+        $this->form = $this->instantiateForm();
+        $this->formView = null;
+
+        $formData = $this->getStorage()->get(sprintf(
+            '%s_form_values_%s',
+            self::prefix(),
+            $this->currentStepName,
+        ));
+
+        $this->formValues = $formData;
+        $this->form->setData($formData);
+    }
+
+    #[ExposeInTemplate]
+    public function isFirst(): bool
+    {
+        return $this->currentStepName === $this->stepNames[array_key_first($this->stepNames)];
+    }
+
+    #[ExposeInTemplate]
+    public function isLast(): bool
+    {
+        return $this->currentStepName === $this->stepNames[array_key_last($this->stepNames)];
+    }
+
+    #[LiveAction]
+    public function submit(): void
+    {
+        $this->submitForm();
+
+        if ($this->hasValidationErrors()) {
+            return;
+        }
+
+        $this->getStorage()->persist(
+            sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName),
+            $this->form->getData(),
+        );
+
+        $this->onSubmit();
+    }
+
+    abstract public function onSubmit();
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function getAllData(): array
+    {
+        $data = [];
+
+        foreach ($this->stepNames as $stepName) {
+            $data[$stepName] = $this->getStorage()->get(sprintf(
+                '%s_form_values_%s',
+                self::prefix(),
+                $stepName,
+            ));
+        }
+
+        return $data;
+    }
+
+    public function resetForm(): void
+    {
+        foreach ($this->stepNames as $stepName) {
+            $this->getStorage()->remove(sprintf('%s_form_values_%s', self::prefix(), $stepName));
+        }
+
+        $this->getStorage()->remove(sprintf('%s_current_step_name', self::prefix()));
+
+        $this->currentStepName = $this->stepNames[\array_key_first($this->stepNames)];
+        $this->form = $this->instantiateForm();
+        $this->formView = null;
+        $this->formValues = $this->extractFormValues($this->getFormView());
+    }
+
+    abstract protected function getStorage(): StorageInterface;
+
+    /**
+     * @return class-string<FormInterface>
+     */
+    abstract protected static function formClass(): string;
+
+    abstract protected function getFormFactory(): FormFactoryInterface;
+
+    /**
+     * @internal
+     */
+    protected function instantiateForm(): FormInterface
+    {
+        $options = [];
+
+        if (null !== $this->currentStepName) {
+            $options['current_step_name'] = $this->currentStepName;
+        }
+
+        return $this->getFormFactory()->create(
+            type: static::formClass(),
+            options: $options,
+        );
+    }
+
+    /**
+     * @internal
+     */
+    private static function prefix(): string
+    {
+        return u(static::class)
+            ->afterLast('\\')
+            ->snake()
+            ->toString();
+    }
+}
--- a/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
@@ -20,6 +20,7 @@ use Symfony\UX\LiveComponent\Attribute\LiveProp;
 use Symfony\UX\LiveComponent\Storage\StorageInterface;
 use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
 use Symfony\UX\TwigComponent\Attribute\PostMount;
+
 use function Symfony\Component\String\u;
 
 /**
@@ -29,8 +30,8 @@ use function Symfony\Component\String\u;
  */
 trait ComponentWithMultiStepFormTrait
 {
-    use DefaultActionTrait;
     use ComponentWithFormTrait;
+    use DefaultActionTrait;
 
     #[LiveProp]
     public ?string $currentStepName = null;
@@ -49,19 +50,19 @@ trait ComponentWithMultiStepFormTrait
     /**
      * @internal
      *
-     * Must be executed after ComponentWithFormTrait::initializeForm().
+     * Must be executed after ComponentWithFormTrait::initializeForm()
      */
     #[PostMount(priority: -250)]
     public function initialize(): void
     {
         $this->currentStepName = $this->getStorage()->get(
-            sprintf('%s_current_step_name', self::prefix()),
+            \sprintf('%s_current_step_name', self::prefix()),
             $this->formView->vars['current_step_name'],
         );
 
         $this->form = $this->instantiateForm();
 
-        $formData = $this->getStorage()->get(sprintf(
+        $formData = $this->getStorage()->get(\sprintf(
             '%s_form_values_%s',
             self::prefix(),
             $this->currentStepName,
@@ -91,7 +92,7 @@ trait ComponentWithMultiStepFormTrait
         }
 
         $this->getStorage()->persist(
-            sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName),
+            \sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName),
             $this->form->getData(),
         );
 
@@ -117,13 +118,13 @@ trait ComponentWithMultiStepFormTrait
         }
 
         $this->currentStepName = $next;
-        $this->getStorage()->persist(sprintf('%s_current_step_name', self::prefix()), $this->currentStepName);
+        $this->getStorage()->persist(\sprintf('%s_current_step_name', self::prefix()), $this->currentStepName);
 
         // If we have a next step, we need to resinstantiate the form and reset the form view and values.
         $this->form = $this->instantiateForm();
         $this->formView = null;
 
-        $formData = $this->getStorage()->get(sprintf(
+        $formData = $this->getStorage()->get(\sprintf(
             '%s_form_values_%s',
             self::prefix(),
             $this->currentStepName,
@@ -165,12 +166,12 @@ trait ComponentWithMultiStepFormTrait
         }
 
         $this->currentStepName = $previous;
-        $this->getStorage()->persist(sprintf('%s_current_step_name', self::prefix()), $this->currentStepName);
+        $this->getStorage()->persist(\sprintf('%s_current_step_name', self::prefix()), $this->currentStepName);
 
         $this->form = $this->instantiateForm();
         $this->formView = null;
 
-        $formData = $this->getStorage()->get(sprintf(
+        $formData = $this->getStorage()->get(\sprintf(
             '%s_form_values_%s',
             self::prefix(),
             $this->currentStepName,
@@ -202,7 +203,7 @@ trait ComponentWithMultiStepFormTrait
         }
 
         $this->getStorage()->persist(
-            sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName),
+            \sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName),
             $this->form->getData(),
         );
 
@@ -219,7 +220,7 @@ trait ComponentWithMultiStepFormTrait
         $data = [];
 
         foreach ($this->stepNames as $stepName) {
-            $data[$stepName] = $this->getStorage()->get(sprintf(
+            $data[$stepName] = $this->getStorage()->get(\sprintf(
                 '%s_form_values_%s',
                 self::prefix(),
                 $stepName,
@@ -232,12 +233,12 @@ trait ComponentWithMultiStepFormTrait
     public function resetForm(): void
     {
         foreach ($this->stepNames as $stepName) {
-            $this->getStorage()->remove(sprintf('%s_form_values_%s', self::prefix(), $stepName));
+            $this->getStorage()->remove(\sprintf('%s_form_values_%s', self::prefix(), $stepName));
         }
 
-        $this->getStorage()->remove(sprintf('%s_current_step_name', self::prefix()));
+        $this->getStorage()->remove(\sprintf('%s_current_step_name', self::prefix()));
 
-        $this->currentStepName = $this->stepNames[\array_key_first($this->stepNames)];
+        $this->currentStepName = $this->stepNames[array_key_first($this->stepNames)];
         $this->form = $this->instantiateForm();
         $this->formView = null;
         $this->formValues = $this->extractFormValues($this->getFormView());
--- a/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
@@ -188,7 +188,7 @@ trait ComponentWithMultiStepFormTrait
     /**
      * Checks if the current step is the first step.
      *
-     * @return bool True if the current step is the first; false otherwise.
+     * @return bool true if the current step is the first; false otherwise
      */
     #[ExposeInTemplate]
     public function isFirst(): bool
@@ -199,7 +199,7 @@ trait ComponentWithMultiStepFormTrait
     /**
      * Checks if the current step is the last step.
      *
-     * @return bool True if the current step is the last; false otherwise.
+     * @return bool true if the current step is the last; false otherwise
      */
     #[ExposeInTemplate]
     public function isLast(): bool
@@ -232,7 +232,7 @@ trait ComponentWithMultiStepFormTrait
     /**
      * Retrieves all data from all steps.
      *
-     * @return array<string, mixed> An associative array of step names and their data.
+     * @return array<string, mixed> an associative array of step names and their data
      */
     public function getAllData(): array
     {
@@ -265,30 +265,30 @@ trait ComponentWithMultiStepFormTrait
     /**
      * Abstract method to retrieve the storage implementation.
      *
-     * @return StorageInterface The storage instance.
+     * @return StorageInterface the storage instance
      */
     abstract protected function getStorage(): StorageInterface;
 
     /**
      * Abstract method to specify the form class for the component.
      *
-     * @return class-string<FormInterface> The form class name.
+     * @return class-string<FormInterface> the form class name
      */
     abstract protected static function formClass(): string;
 
     /**
      * Abstract method to retrieve the form factory instance.
      *
-     * @return FormFactoryInterface The form factory.
+     * @return FormFactoryInterface the form factory
      */
     abstract protected function getFormFactory(): FormFactoryInterface;
 
     /**
      * @internal
      *
-     * Instantiates the form for the current step.
+     * Instantiates the form for the current step
      *
-     * @return FormInterface The form instance.
+     * @return FormInterface the form instance
      */
     protected function instantiateForm(): FormInterface
     {
@@ -304,9 +304,9 @@ trait ComponentWithMultiStepFormTrait
     /**
      * @internal
      *
-     * Generates a unique prefix based on the component's class name.
+     * Generates a unique prefix based on the component's class name
      *
-     * @return string The generated prefix in snake case.
+     * @return string the generated prefix in snake case
      */
     private static function prefix(): string
     {
--- a/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
@@ -1,6 +1,5 @@
 <?php
 
-declare(strict_types=1);
 
 /*
  * This file is part of the Symfony package.
--- a/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
@@ -25,10 +25,8 @@ use function Symfony\Component\String\u;
 /**
  * Trait for managing multi-step forms in LiveComponent.
  *
- * This trait simplifies the implementation of multistep forms by handling
+ * This trait simplifies the implementation of multi-step forms by handling
  * step transitions, form validation, data persistence, and state management.
- * It provides a structured API for developers to integrate multistep forms
- * into their components with minimal boilerplate.
  *
  * @author Silas Joisten <silasjoisten@proton.me>
  * @author Patrick Reimers <preimers@pm.me>
--- a/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
@@ -23,7 +23,7 @@ use Symfony\UX\TwigComponent\Attribute\PostMount;
 use function Symfony\Component\String\u;
 
 /**
- * Trait for managing multistep forms in Symfony UX LiveComponent.
+ * Trait for managing multi-step forms in LiveComponent.
  *
  * This trait simplifies the implementation of multistep forms by handling
  * step transitions, form validation, data persistence, and state management.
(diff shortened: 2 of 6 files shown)


seb-jean:
Hello, thank you very much for your contribution: the idea is very interesting.
I don't know if it's already done but is there a possibility to stay on the page of the current step when you refresh the page or does it go back to step 1?

silasjoisten:
> Hello, thank you very much for your contribution: the idea is very interesting.
> 
> I don't know if it's already done but is there a possibility to stay on the page of the current step when you refresh the page or does it go back to step 1?

Hello, there is a Storage used which allows you by default to stay on the same step on page reload. So there is the persistence level already included

seb-jean:
Nice!

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,283 @@
...
+            $this->currentStepName,
+        ));
+
+        $this->form->setData($formData);
+
+        if ([] === $formData) {
+            $this->formValues = $this->extractFormValues($this->getFormView());
+        } else {
+            $this->formValues = $formData;
+        }
```
OskarStark:
```suggestion
        $this->formValues = [] === $formData
            ? $this->extractFormValues($this->getFormView())
            : $formData;
```

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,283 @@
...
+            $this->formView->vars['current_step_name'],
+        );
+
+        $this->form = $this->instantiateForm();
+
+        $formData = $this->getStorage()->get(\sprintf(
+            '%s_form_values_%s',
+            self::prefix(),
+            $this->currentStepName,
+        ));
```
OskarStark:
```suggestion
        $formData = $this->getStorage()->get(\sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName));
```

Symfony prefers one liners to avoid merge conflicts

same for the other code parts

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
OskarStark: same change for down below

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,283 @@
...
+        $this->formView = null;
+
+        $formData = $this->getStorage()->get(\sprintf(
+            '%s_form_values_%s',
+            self::prefix(),
+            $this->currentStepName,
+        ));
+
+        // I really don't understand why we need to do that. But what I understood is extractFormValues creates
+        // an array of initial values.
```
OskarStark: Should be kept?

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,283 @@
...
+    {
+        $options = [];
+
+        if (null !== $this->currentStepName) {
+            $options['current_step_name'] = $this->currentStepName;
+        }
+
+        return $this->getFormFactory()->create(
+            type: static::formClass(),
+            options: $options,
```
OskarStark:
```suggestion
            static::formClass(),
            $options,
```

no named arguments, as there is no BC promise for this from Symfony side

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,283 @@
...
+
+    /**
+     * @internal
+     */
+    private static function prefix(): string
+    {
+        return u(static::class)
+            ->afterLast('\\')
+            ->snake()
+            ->toString();
```
OskarStark:
```suggestion
        return u(static::class)->afterLast('\\')->snake()->toString();
```

OskarStark: And is symfony/string still a requirement ?

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: Yes but the second argument is data which will be null

src/LiveComponent/src/Storage/StorageInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace Symfony\UX\LiveComponent\Storage;
+
+/**
+ * @author Silas Joisten <silasjoisten@proton.me>
+ * @author Patrick Reimers <preimers@pm.me>
+ * @author Jules Pietri <jules@heahprod.com>
+ */
+interface StorageInterface
```
OskarStark, line 28: Should have a description

src/LiveComponent/src/Storage/StorageInterface.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public function persist(string $key, mixed $values): void;
+
+    public function remove(string $key): void;
+
+    public function get(string $key, mixed $default = []): mixed;
```
OskarStark: Methods should be described as well

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
...
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Silas Joisten <silasjoisten@proton.me>
+ * @author Patrick Reimers <preimers@pm.me>
+ * @author Jules Pietri <jules@heahprod.com>
+ */
+final class MultiStepType extends AbstractType
```
OskarStark, line 27: Should have a description

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,283 @@
...
+use Symfony\UX\TwigComponent\Attribute\PostMount;
+
+use function Symfony\Component\String\u;
+
+/**
+ * @author Silas Joisten <silasjoisten@proton.me>
+ * @author Patrick Reimers <preimers@pm.me>
+ * @author Jules Pietri <jules@heahprod.com>
+ */
+trait ComponentWithMultiStepFormTrait
```
OskarStark, line 35: should have description

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: yes i required it.

WebMamba:
Hey @silasjoisten ! Thanks for this nice contribution ! You bring something that can be truly useful for a lot of folks! I just don't thinks that a trait in the LiveComponent component is the right place... I think we bring to much that are not LiveComponent related:  StorageInterface, MultiStepType, and so on. I am also scared also that doing this in to a Trait can limit us in to the implementation and on the DX. I think we can go way further on this topics and bring features that can be also front end related (step transition).
So I am not sure where is the right place to put this work, should it be a new component ? A new documentation ? A Cookbook ?
Tell me what do you thinks about that cheers!
Thanks again 😁

silasjoisten:
Hey @WebMamba, 

Thanks for your detailed feedback and for taking the time to review this! 😊 

I firmly believe that this feature fits well within the Symfony UX LiveComponents component. Here’s why:

1. **Alignment with Existing Architecture**:  
   The multistep form concept is built on the same principles as LiveComponents: reactive, state-driven interfaces that rely on server-side logic to manage component behavior. The `StorageInterface` and `MultiStepType` were specifically designed to integrate seamlessly into the LiveComponent ecosystem. Much like `ComponentWithFormTrait`, which manages forms within LiveComponents, this trait extends that functionality to handle multistep forms without introducing unnecessary complexity.

2. **CollectionType as an Example**:  
   Consider how Symfony provides the `CollectionType` as part of its Form component. It’s a focused feature, but it belongs there because it enhances the developer experience for managing collections within forms. Similarly, multistep forms naturally fit within LiveComponents because they extend its reactive capabilities. 

3. **DX and Limitations**:  
   You mentioned that this approach might limit DX or implementation options. Could you elaborate on specific scenarios where you think this might be restrictive? For instance, LiveComponents already allows for tight integration of server-side logic, storage, and frontend rendering. How would moving this feature to another component, cookbook, or package better serve the DX? From my perspective, centralizing this within LiveComponents provides developers with a familiar and consistent experience, reducing the need to integrate external solutions or reinvent the wheel.

4. **Future Enhancements**:  
   While this implementation focuses on server-side management, it lays a strong foundation for frontend-related features like step transitions. Since LiveComponents already facilitates reactive UI updates, extending this feature to include frontend transitions feels like a natural progression within this component.

I’d love to hear your thoughts on these points and where you see potential challenges or conflicts. If you have specific ideas for improving or repositioning the feature, I’m happy to collaborate further. 

Cheers,  
Silas

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
OskarStark: Is this the only place where it is needed? In this case it is preferred to not use a dep but use plain vanilla PHP for it

smnandre:
Hi @silasjoisten! So happy to see you there ❤️ 

First, thank you very very much for this PR! This feature will be warmly welcomed by many users.

Regarding the "where", it is now pretty clear, in my mind, that we will end up with a dedicated "LiveForm" or "UxForm" component/bundle.. probably sooner than later. But right now, we have... LiveComponent :)

Also, I do agree with you regarding the similarity with CollectionType + the DX aspects.

As i see it, we can introduce it in "experimental" anywhere we want, and we'll see until that where it best fits in UX 3.

Next comments in the code :) 

I need to sleep on the "StorageInterface" thing because i'm having mixed feeling about introducing a generic tool like this for now, while we're moving things around here.

src/LiveComponent/composer.json
```diff
@@ -51,6 +51,7 @@
         "symfony/serializer": "^5.4|^6.0|^7.0",
         "symfony/twig-bundle": "^5.4|^6.0|^7.0",
         "symfony/validator": "^5.4|^6.0|^7.0",
+        "symfony/string": "^5.4|^6.0|^7.0",
```
smnandre, line 54: Can we remove it ( if only used for the string/camel stuff ?)

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
+<?php
+
+declare(strict_types=1);
```
smnandre: We don't use strict types in the codebase for now.. 🤷 
```suggestion
```

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+use Symfony\UX\LiveComponent\Attribute\LiveAction;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\Storage\StorageInterface;
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+use Symfony\UX\TwigComponent\Attribute\PostMount;
+
+use function Symfony\Component\String\u;
+
+/**
+ * Trait for managing multistep forms in Symfony UX LiveComponent.
```
smnandre:
```suggestion
 * Trait for managing multi-step forms in LiveComponent.
```

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+ *
+ * This trait simplifies the implementation of multistep forms by handling
+ * step transitions, form validation, data persistence, and state management.
+ * It provides a structured API for developers to integrate multistep forms
+ * into their components with minimal boilerplate.
```
smnandre:
```suggestion
 * This trait simplifies the implementation of multi-step forms by handling
 * step transitions, form validation, data persistence, and state management.
```

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
```
smnandre:
```suggestion
```

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+final class MultiStepType extends AbstractType
+{
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setDefault('current_step_name', static function (Options $options): string {
+                return array_key_first($options['steps']);
+            })
+            ->setRequired('steps');
```
smnandre, line 35: Will require more checks

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
...
+ * @author Silas Joisten <silasjoisten@proton.me>
+ * @author Patrick Reimers <preimers@pm.me>
+ * @author Jules Pietri <jules@heahprod.com>
+ */
+final class MultiStepType extends AbstractType
+{
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setDefault('current_step_name', static function (Options $options): string {
```
smnandre, line 32: Let's discuss this more globally, but i think "current_step" or even "step" could be enough, and improve readability. 

Maybe just me, but i feel the implementation choice (e.g., passing the steps as a map) should not impose constraints on the overall naming convention. 

wdyt ?

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
...
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $options['steps'][$options['current_step_name']]($builder);
+    }
+
+    public function buildView(FormView $view, FormInterface $form, array $options): void
+    {
+        $view->vars['current_step_name'] = $options['current_step_name'];
+        $view->vars['steps_names'] = array_keys($options['steps']);
+    }
```
smnandre, line 47:
```suggestion
    }

    public function getBlockPrefix(): string
    {
        return 'multistep';
    }
    
```

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
...
+        $resolver
+            ->setDefault('current_step_name', static function (Options $options): string {
+                return array_key_first($options['steps']);
+            })
+            ->setRequired('steps');
+    }
+
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $options['steps'][$options['current_step_name']]($builder);
```
smnandre, line 40: This should be checked first i guess ? 😅

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function buildView(FormView $view, FormInterface $form, array $options): void
+    {
+        $view->vars['current_step_name'] = $options['current_step_name'];
+        $view->vars['steps_names'] = array_keys($options['steps']);
```
smnandre, line 46: What about "step" and "steps" (or "current_step" and "steps" ? 

To illustrate my point here: if you look at ChoiceType, it does not use "prefered_choices**_values**" or "prefered_choices**_names**"

src/LiveComponent/src/Form/Type/MultiStepType.php
smnandre, line 27: I think this is where we should describe/document what is a "step".

src/LiveComponent/src/Storage/SessionStorage.php
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Implementation of the StorageInterface using Symfony's session mechanism.
+ *
+ * This class provides a session-based storage solution for managing data
+ * persistence in Symfony UX LiveComponent. It leverages the Symfony
+ * `RequestStack` to access the session and perform operations such as
+ * storing, retrieving, and removing data.
+ *
+ * Common use cases include persisting component state, such as form data
+ * or multistep workflow progress, across user interactions.
```
smnandre, line 26: >  Common use cases include persisting component state, such as form data, or multistep workflow progress, across user interactions.

I'm having doubt if we should push session usage like this, as in the same time we push for stateless components, removed csrf, etc..

And i'm very hesistant to add such feature in another PR :| 

I don't think these two things should be handled as one... 

Will sleep on it... but already can see several alternatives:
- having an internal/experimental storage (just pour the multi-step form)
- using props
- using existing adapters (doctrine, cache)

src/LiveComponent/src/Storage/SessionStorage.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
```
smnandre:
```suggestion
```

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+
+        $this->getStorage()->persist(\sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName), $this->form->getData());
+
+        $this->onSubmit();
+    }
+
+    /**
+     * Abstract method to be implemented by the component for custom submission logic.
+     */
+    abstract public function onSubmit();
```
smnandre, line 227: Not sure what this mean, but onSubmit feels uncommon name.. :/

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+    #[LiveAction]
+    public function submit(): void
+    {
+        $this->submitForm();
+
+        if ($this->hasValidationErrors()) {
+            return;
+        }
+
+        $this->getStorage()->persist(\sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName), $this->form->getData());
```
smnandre, line 219:
```suggestion
        $this->getStorage()->persist(\sprintf('%s_form_values_%s', self::prefix(), $this->currentStepName), $this->form->getData());
```

This feels very implementation specific to me... more on that later :)

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+
+    /**
+     * Checks if the current step is the first step.
+     *
+     * @return bool true if the current step is the first; false otherwise
+     */
+    #[ExposeInTemplate]
+    public function isFirst(): bool
+    {
+        return $this->currentStepName === $this->stepNames[array_key_first($this->stepNames)];
```
smnandre, line 193:
```suggestion
        return $this->currentStep === reset($this->steps);
```

wdyt ?

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+
+    /**
+     * Checks if the current step is the last step.
+     *
+     * @return bool true if the current step is the last; false otherwise
+     */
+    #[ExposeInTemplate]
+    public function isLast(): bool
+    {
+        return $this->currentStepName === $this->stepNames[array_key_last($this->stepNames)];
```
smnandre, line 204:
```suggestion
        return $this->currentStep === end($this->steps);
```

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+            if ($found) {
+                $previous = $stepName;
+
+                break;
+            }
+        }
+
+        if (null === $previous) {
+            throw new \RuntimeException('No previous forms available.');
+        }
```
smnandre, line 167:
```php
$current = array_search($currentStep, $steps, true);

if (!$current) {
    throw new \RuntimeException('No previous steps available.');
}

$previous = $steps[$current - 1];
```

(or similar)

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+        $this->formView = null;
+
+        $formData = $this->getStorage()->get(\sprintf(
+            '%s_form_values_%s',
+            self::prefix(),
+            $this->currentStepName,
+        ));
+
+        $this->formValues = $formData;
+        $this->form->setData($formData);
```
smnandre, line 182: Duplicate with "next" method... should this be the main "getCurrentForm"  or "getStepForm" method or something like that ?

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+     * Throws a RuntimeException if no next step is available.
+     */
+    #[LiveAction]
+    public function next(): void
+    {
+        $this->submitForm();
+
+        if ($this->hasValidationErrors()) {
+            return;
+        }
```
smnandre, line 98: If `!form->isValid`, then a `UnprocessableEntityHttpException` is triggered in the componentwithformtrait, no ? So as i see it, the hasValidationErrors() check is redondant (i may have missed something)

 ```suggestion
        $this->submitForm();
```

File: src/LiveComponent/src/ComponentWithMultiStepFormTrait.php [hunk 29]
```diff
@@ -0,0 +1,315 @@
...
+ * @author Patrick Reimers <preimers@pm.me>
+ * @author Jules Pietri <heahdude@yahoo.fr>
+ */
+trait ComponentWithMultiStepFormTrait
+{
+    use ComponentWithFormTrait;
+    use DefaultActionTrait;
+
+    #[LiveProp]
+    public ?string $currentStepName = null;
```
smnandre, line 41: Should we make this non nullable? I don't see when a multi-step could have "no current step"

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
```diff
@@ -0,0 +1,315 @@
...
+
+        $this->form->setData($formData);
+
+        $this->formValues = [] === $formData
+            ? $this->extractFormValues($this->getFormView())
+            : $formData;
+
+        $this->stepNames = $this->formView->vars['steps_names'];
+
+        // Do not move this. The order is important.
```
smnandre, line 81: Why? 😅

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: Maybe doSubmit ?

src/LiveComponent/src/Form/Type/MultiStepType.php
silasjoisten: What checks you got in mind?

src/LiveComponent/src/Form/Type/MultiStepType.php
silasjoisten: I dont think so, because configure options is already ensuring that. (see the test)

src/LiveComponent/src/Form/Type/MultiStepType.php
silasjoisten: sure we can rename this.

src/LiveComponent/src/Storage/SessionStorage.php
silasjoisten: If it is stateless. okay we could introduce a NullStorage. But in order to be able to reload the page and having a persistent state is quite cool in my opinion.

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: reset can return false which makes it not typesafe enough for me.

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: i would like to use the new php functions array_key_first and array_key_last.

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: with you example the exception is thrown if no current step is available.

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: not sure to be honest. i can test it.

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: We are in a trait and we dont have a constructor or something here. which means it depends on the #[PostMount] hook. if thats not executed it is null by default. i think thats something by design we should not change.

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
silasjoisten: good question :D i cann add a better comment

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
WedgeSama: Why not just `submit` like it is in `FormInterface`?

src/LiveComponent/src/Form/Type/MultiStepType.php
```diff
@@ -0,0 +1,48 @@
...
+        $resolver
+            ->setDefault('current_step_name', static function (Options $options): string {
+                return array_key_first($options['steps']);
+            })
+            ->setRequired('steps');
+    }
+
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $options['steps'][$options['current_step_name']]($builder);
```
WedgeSama, line 40: Allowing to use `FormType` as step can be good too. What do you think?

Maybe with something like that (haven't tested it):
```suggestion
        $step = $options['steps'][$options['current_step_name']];
        if (is_callable()) {
            $step($builder);
        } elseif (is_string($step) && is_a($step, FormTypeInterface::class)) {
            $builder->add('step', $step, [
                'inherit_data' => true,
            ]);
        }
```

WedgeSama, line 40:
```suggestion
        $options['steps'][$options['current_step_name']]($builder, $options);
```

IMO, passing `$options` to keep consistent with `buildForm` method can be good.

WedgeSama (commented):
Raw thought:

IMO, this feature can have its place in Form component, it will be useful for people/project that not using ux yet and will ease transitioning to ux.

And that way `StorageInterface` make sense in the Form component.

src/LiveComponent/src/Form/Type/MultiStepType.php
silasjoisten: sure we can do that but my question is. Why would it make sense to pass the options of the "parent" form StepType to the children? you got an example or use case where it might be useful?

silasjoisten: yes i was thinking about the same. so callback or class string of FormTypeInterface.

WedgeSama: It is more to be consistent with how work the `buildForm` method, where you have access to options.

In your PR, you cannot set your form options yet, so there is no direct example. `instantiateForm` method pass an empty `$options` (except the `current_step_name`) to `FormFactory::create`.

You may want to create step form that is configurable, either from:
- your live component (not yet possible)
- using `FormTypeExtension` from Form component

Is it clearer that way?

src/LiveComponent/src/Form/Type/MultiStepType.php
smnandre: I may miss-read this, but to me at this point there is no certitude that `$options['steps'][$options['current_step_name']]` is a callable  that accept FormBuilderInterface argument ?

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
smnandre: Then currentStep === .. would return false no ?

src/LiveComponent/src/ComponentWithMultiStepFormTrait.php
smnandre: haha the "new" :) 

The "problem" (a detail, to be honest) here is that you first look for a key, then get the value indexed by that key, then compare currentStepName is identical.

But we can update this in time if necessary :)

smnandre:
My overall reaction here would be : i kinda feel 80% of this PR should be in ... the symfony/form component.

Both the "steps" logic and the "storage" logic are not related with Symfony UX and could be used without it.
... and would be very usefull seen the overall demand for it (in live component but in general too)

I'd even go to say LiveComponent could only provide a storage (in the browser DOM, via a LiveProp) for this feature.

chalasr:
I tend to agree that part of the added logic could belong to symfony/form instead. Curious to know @heahdude thoughts on this

Kocal:
Status: Needs work

silasjoisten:
@smnandre @WebMamba @Kocal 

- I created a PR and added the `MultiStepType` to the form component. https://github.com/symfony/symfony/pull/59548 - - I don't know what to do with the `StorageInterface` and the `SessionStorage` i think it does not belong into the form component due to the `Session` dependency.

Regarding the Trait. As far as i can see this is the only thing that belongs to this PR and the symfony UX repository due to its similarities with the other traits.

What else is todo?

smnandre:
Congrat's for the PR (i'll have a look over there too, at least to support it 😻 ) 

I agree with you: a trait / helper could be ideal, in term of both DX and long-term "maintainability". 

Let's see how it goes on the symfony/symfony repository... depending on the result we will better know the ideal move here (what i mean here is: no need to lose time polishing things here as you already did the code, so let's focus on the main PR and if some changes are asked for on symfony you'll have time to adapt here)

Also, to avoid any _quiproquo_ here: we have monthly-_ish_ releases around here, but the 7.3 will not be out until march I guess, so no pressure on this side, as we won't release if the "base" code from symfony/form is not available... right ?

silasjoisten:
Thanks! Yea it depends now on what the guys say regarding the form PR. My only concern is still where to put the Storage interface. I was thinking a lot about it and it does not make sense anywhere. Not framework bundle, not form component (no dependency on session etc.) so the only place left is here where its actual implementation lays. And yes i agree with you. We got time until 7.3 is released before we can continue this.

seb-jean:
I was wondering, will the URL change depending on the steps? Or will the URL stay the same?

PReimers:
@seb-jean The URL will stay the same.

kbond:
https://github.com/symfony/symfony/pull/60212 could help here?

silasjoisten:
Yes it will

RafaelKr:
The official Symfony FormFlow integration is pretty much done and targeted for Symfony 7.4: https://github.com/symfony/symfony/pull/60212#issuecomment-3112620167

Maybe it would make sense to rebuild on this now?

Something I really would appreciate would be the ability to support Dependent Form Fields like in that demo: https://ux.symfony.com/demos/live-component/dependent-form-fields

smnandre:
Yeah, i also would be happy to see how this can be connected/integrated now! :) 

I believe @silasjoisten has a lot on its plate right now... so if i would also love to use/work this feature, there is in fact no real urgency right now (see note below) 😄   

(this message is not directed to your message @RafaelKr don't worry, more a poke at @silasjoisten ;) )

--- 

Side Note 1:

We are not in sync with Symfony releases (for now)... so we cannot merge it before 7.4 is released, in all manner...  or it would be released but not working :| 

---

Side note 2:
 
SymfonyUX  cannot support _officially_ DynamicForms as it's not a Symfony UX package (not a component or bundle we maintain).

That said, we have a very **very** good relationship with SymfonyCasts (where both @weaverryan and @kbond work), so we’ll do our absolute best to keep things in sync. Just we cant make _firm guarantees_ when it comes to external code.

Kocal:
Since Symfony 7.4 released a Multi-Step form feature (https://symfony.com/blog/new-in-symfony-7-4-multi-step-forms), I'm closing this PR.

Thanks for your contribution, that's appreciated!

ludovik-dev:
@Kocal Indeed, Symfony 7.4 has released a Multi-Step form feature, but could this feature be used for integration with LiveComponent?

PR #2439, opened 2024-12-10
[Autocomplete] reorder autocomplete data-controller

IndraGunawan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        |
| License       | MIT

load `symfony--ux-autocomplete--autocomplete` controller first before other controller make other controllers able access the tomselect instance, this will make autocomplete more extensible

having an custom-tomselect controller

```

    connect() {
        console.log(this.element.tomselect)
    }
}
```

and a form field

```
$builder->add('field', ChoiceType::class, [
            'autocomplete' => true,
            'attr' => [
                'data-controller' => 'custom-tomselect',
            ],
        ]);
```

in current behavior it will print out `undefined`, with this change it print tomselect instance

### DIFF
--- a/src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
+++ b/src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
@@ -55,7 +55,7 @@ final class AutocompleteChoiceTypeExtension extends AbstractTypeExtension
         $attr = $view->vars['attr'] ?? [];
 
         $controllerName = 'symfony--ux-autocomplete--autocomplete';
-        $attr['data-controller'] = $controllerName.' '.trim(($attr['data-controller'] ?? ''));
+        $attr['data-controller'] = $controllerName.' '.trim($attr['data-controller'] ?? '');
 
         $values = [];
         if ($options['autocomplete_url']) {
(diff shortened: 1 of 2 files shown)


src/Autocomplete/src/Form/AutocompleteChoiceTypeExtension.php
```diff
@@ -55,7 +55,7 @@ public function finishView(FormView $view, FormInterface $form, array $options):
         $attr = $view->vars['attr'] ?? [];
 
         $controllerName = 'symfony--ux-autocomplete--autocomplete';
-        $attr['data-controller'] = trim(($attr['data-controller'] ?? '').' '.$controllerName);
+        $attr['data-controller'] = $controllerName.' '.trim($attr['data-controller'] ?? '');
```
smnandre, line 58:
```suggestion
        $attr['data-controller'] = trim($controllerName.' '.($attr['data-controller'] ?? ''));
```

smnandre: I'm not sure here.. as i believe we recommand the other way around for now..

https://symfony.com/bundles/ux-autocomplete/current/index.html#extending-tom-select

IndraGunawan: I have used that way to extend tom-select on other places, i completely forgot about that.

thank you @smnandre for responding this PR

closing this PR

PR #2440, opened 2024-12-10
[Autocomplete] Fix Incompatibility filter_query and max_results (doc)

### LINKED ISSUE #769: [Autocomplete] filter_query & max_results conflict
There is a conflict on ux-autocomplete between filter_query and max_results.

With this code, we have 50 results:
```
'max_results' => 50
```

But with this code, we have 10 results:
```
'filter_query' => function(QueryBuilder $qb, string $query, EntityRepository $repository) {
    if (!$query) {
        return;
    }

    $qb->andWhere('entity.name LIKE :filter OR entity.description LIKE :filter')
        ->setParameter('filter', '%'.$query.'%');
},
 'max_results' => 50
 ```

### PR DESCRIPTION
**Pull request by ytilotti**
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix #769 
| License       | MIT

Fix in the documentation the incompatibility between filter_query and max_results.

### DIFF
--- a/src/Autocomplete/doc/index.rst
+++ b/src/Autocomplete/doc/index.rst
@@ -250,7 +250,7 @@ to the options above, you can also pass:
 
 ``filter_query`` (default: ``null``)
     If you want to completely control the query made for the "search results",
-    use this option. This is incompatible with ``searchable_fields``::
+    use this option. This is incompatible with ``searchable_fields`` and ``max_results``::
 
         [
             'filter_query' => function(QueryBuilder $qb, string $query, EntityRepository $repository) {

### CODE REVIEW

File: src/Autocomplete/doc/index.rst [hunk 1]
```diff
@@ -250,7 +250,7 @@ to the options above, you can also pass:
 
 ``filter_query`` (default: ``null``)
     If you want to completely control the query made for the "search results",
-    use this option. This is incompatible with ``searchable_fields``::
+    use this option. This is incompatible with ``searchable_fields`` and ``max_results``::
```
smnandre, line 253: Nice catch here!

Would you also update the BaseAutocompleteType code to make the same check than filter/searcheable_fields ?  https://github.com/symfony/ux/blob/eab7db5c90401d943cdb64394d01ed1579cc9694/src/Autocomplete/src/Form/BaseEntityAutocompleteType.php#L68-L75

ytilotti: Can be a problem because max_results has default to 10.

smnandre (approved):

src/Autocomplete/doc/index.rst
smnandre: Well let's forget then :) Thanks for the fix!

smnandre:
Thank you @ytilotti.

PR #2447, opened 2024-12-12
[Turbo] Pass EventSource options to `turbo_stream_listen`

### LINKED ISSUE #1860: [Turbo] Add `mercure()` to `turbo_stream_listen()`
Hi,

For generate the URL of the Hub and for automatically set the cookie required to use the authorization mechanism, we need Twig helper `mercure()` (PR: https://github.com/symfony/mercure/pull/62)

But how to use it with `turbo_stream_listen()`?

Fan2Shrek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1860 I think
| License       | MIT

This allows to pass options to the EventSource, for example
`{{ turbo_stream_listen('a_topic', 'default', { withCredentials: true }) }}`
Or create specific auth cookie 
`{{ turbo_stream_listen('a_topic', 'default', { subscribe: '*' }) }}`
This functionality is similar to how `mercure()` works

### DIFF
--- a/src/Turbo/CHANGELOG.md
+++ b/src/Turbo/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.24.0
 
 -   Add Twig Extensions for `meta` tags
+-   Add support for authentication to the EventSource via `turbo_stream_listen`
 
 ## 2.22.0
 
--- a/src/Turbo/assets/src/turbo_stream_controller.ts
+++ b/src/Turbo/assets/src/turbo_stream_controller.ts
@@ -18,12 +18,14 @@ export default class extends Controller {
         topic: String,
         topics: Array,
         hub: String,
+        withCredentials: Boolean,
     };
     es: EventSource | undefined;
     url: string | undefined;
 
     declare readonly topicValue: string;
     declare readonly topicsValue: string[];
+    declare readonly withCredentialsValue: boolean;
     declare readonly hubValue: string;
     declare readonly hasHubValue: boolean;
     declare readonly hasTopicValue: boolean;
@@ -50,7 +52,7 @@ export default class extends Controller {
 
     connect() {
         if (this.url) {
-            this.es = new EventSource(this.url);
+            this.es = new EventSource(this.url, { withCredentials: this.withCredentialsValue });
             connectStreamSource(this.es);
         }
     }
--- a/src/Turbo/config/services.php
+++ b/src/Turbo/config/services.php
@@ -18,6 +18,7 @@ use Symfony\UX\Turbo\Broadcaster\ImuxBroadcaster;
 use Symfony\UX\Turbo\Broadcaster\TwigBroadcaster;
 use Symfony\UX\Turbo\Doctrine\BroadcastListener;
 use Symfony\UX\Turbo\Request\RequestListener;
+use Symfony\UX\Turbo\Twig\TurboRuntime;
 use Symfony\UX\Turbo\Twig\TwigExtension;
 
 /*
@@ -47,8 +48,14 @@ return static function (ContainerConfigurator $container): void {
             ->decorate('turbo.broadcaster.imux')
 
         ->set('turbo.twig.extension', TwigExtension::class)
-            ->args([tagged_locator('turbo.renderer.stream_listen', 'transport'), abstract_arg('default')])
             ->tag('twig.extension')
+
+        ->set('turbo.twig.runtime', TurboRuntime::class)
+            ->args([
+                tagged_locator('turbo.renderer.stream_listen', 'transport'),
+                abstract_arg('default_transport'),
+            ])
+            ->tag('twig.runtime')
 
         ->set('turbo.doctrine.event_listener', BroadcastListener::class)
             ->args([
--- a/src/Turbo/doc/index.rst
+++ b/src/Turbo/doc/index.rst
@@ -754,6 +754,9 @@ Let's create our chat::
         </turbo-frame>
     {% endblock %}
 
+If you're using a private hub, you can add ``{ withCredentials: true }``
+as ``turbo_stream_listen()`` third argument to authenticate with the hub
+
 .. code-block:: html+twig
 
     {# chat/message.stream.html.twig #}
--- a/src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
+++ b/src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
@@ -12,18 +12,20 @@
 namespace Symfony\UX\Turbo\Bridge\Mercure;
 
 use Symfony\Component\Mercure\HubInterface;
+use Symfony\Component\Mercure\Twig\MercureExtension;
 use Symfony\UX\StimulusBundle\Helper\StimulusHelper;
 use Symfony\UX\Turbo\Broadcaster\IdAccessor;
-use Symfony\UX\Turbo\Twig\TurboStreamListenRendererInterface;
+use Symfony\UX\Turbo\Twig\TurboStreamListenRendererWithOptionsInterface;
 use Symfony\WebpackEncoreBundle\Twig\StimulusTwigExtension;
 use Twig\Environment;
+use Twig\Error\RuntimeError;
 
 /**
  * Renders the attributes to load the "mercure-turbo-stream" controller.
  *
  * @author Kévin Dunglas <kevin@dunglas.fr>
  */
-final class TurboStreamListenRenderer implements TurboStreamListenRendererInterface
+final class TurboStreamListenRenderer implements TurboStreamListenRendererWithOptionsInterface
 {
     private StimulusHelper $stimulusHelper;
 
@@ -31,6 +33,7 @@ final class TurboStreamListenRenderer implements TurboStreamListenRendererInterf
         private HubInterface $hub,
         StimulusHelper|StimulusTwigExtension $stimulus,
         private IdAccessor $idAccessor,
+        private Environment $twig,
     ) {
         if ($stimulus instanceof StimulusTwigExtension) {
             trigger_deprecation('symfony/ux-turbo', '2.9', 'Passing an instance of "%s" as second argument of "%s" is deprecated, pass an instance of "%s" instead.', StimulusTwigExtension::class, __CLASS__, StimulusHelper::class);
@@ -42,8 +45,12 @@ final class TurboStreamListenRenderer implements TurboStreamListenRendererInterf
         $this->stimulusHelper = $stimulus;
     }
 
-    public function renderTurboStreamListen(Environment $env, $topic): string
+    public function renderTurboStreamListen(Environment $env, $topic /* array $eventSourceOptions = [] */): string
     {
+        if (\func_num_args() > 2) {
+            $eventSourceOptions = func_get_arg(2);
+        }
+
         $topics = $topic instanceof TopicSet
             ? array_map($this->resolveTopic(...), $topic->getTopics())
             : [$this->resolveTopic($topic)];
@@ -53,6 +60,18 @@ final class TurboStreamListenRenderer implements TurboStreamListenRendererInterf
             $controllerAttributes['topics'] = $topics;
         } else {
             $controllerAttributes['topic'] = current($topics);
+        }
+
+        if (isset($eventSourceOptions)) {
+            try {
+                $mercure = $this->twig->getExtension(MercureExtension::class);
+                $mercure->mercure($topics, $eventSourceOptions);
+
+                if (isset($eventSourceOptions['withCredentials'])) {
+                    $controllerAttributes['withCredentials'] = $eventSourceOptions['withCredentials'];
+                }
+            } catch (RuntimeError $e) {
+            }
         }
 
         $stimulusAttributes = $this->stimulusHelper->createStimulusAttributes();
--- a/src/Turbo/src/Twig/TurboRuntime.php
+++ b/src/Turbo/src/Twig/TurboRuntime.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Turbo\Twig;
+
+use Psr\Container\ContainerInterface;
+use Symfony\UX\Turbo\Bridge\Mercure\TopicSet;
+use Twig\Environment;
+use Twig\Extension\RuntimeExtensionInterface;
+
+/**
+ * @author Kévin Dunglas <kevin@dunglas.fr>
+ * @author Pierre Ambroise <pierre27.ambroise@gmail.com>
+ *
+ * @internal
+ */
+final class TurboRuntime implements RuntimeExtensionInterface
+{
+    public function __construct(
+        private ContainerInterface $turboStreamListenRenderers,
+        private readonly string $defaultTransport,
+    ) {
+    }
+
+    /**
+     * @param object|string|array<object|string> $topic
+     * @param array<string, mixed>               $options
+     */
+    public function renderTurboStreamListen(Environment $env, $topic, ?string $transport = null, array $options = []): string
+    {
+        $options['transport'] = $transport ??= $this->defaultTransport;
+
+        if (!$this->turboStreamListenRenderers->has($transport)) {
+            throw new \InvalidArgumentException(\sprintf('The Turbo stream transport "%s" does not exist.', $transport));
+        }
+
+        if (\is_array($topic)) {
+            $topic = new TopicSet($topic);
+        }
+
+        $renderer = $this->turboStreamListenRenderers->get($transport);
+
+        return $renderer instanceof TurboStreamListenRendererWithOptionsInterface
+            ? $renderer->renderTurboStreamListen($env, $topic, $options) // @phpstan-ignore-line
+            : $renderer->renderTurboStreamListen($env, $topic);
+    }
+}
--- a/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
+++ b/src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
@@ -23,5 +23,5 @@ interface TurboStreamListenRendererInterface
     /**
      * @param string|object $topic
      */
-    public function renderTurboStreamListen(Environment $env, $topic): string;
+    public function renderTurboStreamListen(Environment $env, $topic /* , array $eventSourceOptions = [] */): string;
 }
--- a/src/Turbo/src/Twig/TwigExtension.php
+++ b/src/Turbo/src/Twig/TwigExtension.php
@@ -11,9 +11,6 @@
 
 namespace Symfony\UX\Turbo\Twig;
 
-use Psr\Container\ContainerInterface;
-use Symfony\UX\Turbo\Bridge\Mercure\TopicSet;
-use Twig\Environment;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
@@ -28,16 +25,10 @@ final class TwigExtension extends AbstractExtension
     private const REFRESH_SCROLL_RESET = 'reset';
     private const REFRESH_SCROLL_PRESERVE = 'preserve';
 
-    public function __construct(
-        private ContainerInterface $turboStreamListenRenderers,
-        private string $default,
-    ) {
-    }
-
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('turbo_stream_listen', $this->turboStreamListen(...), ['needs_environment' => true, 'is_safe' => ['html']]),
+            new TwigFunction('turbo_stream_listen', [TurboRuntime::class, 'renderTurboStreamListen'], ['needs_environment' => true, 'is_safe' => ['html']]),
             new TwigFunction('turbo_exempts_page_from_cache', $this->turboExemptsPageFromCache(...), ['is_safe' => ['html']]),
             new TwigFunction('turbo_exempts_page_from_preview', $this->turboExemptsPageFromPreview(...), ['is_safe' => ['html']]),
             new TwigFunction('turbo_page_requires_reload', $this->turboPageRequiresReload(...), ['is_safe' => ['html']]),
@@ -45,24 +36,6 @@ final class TwigExtension extends AbstractExtension
             new TwigFunction('turbo_refresh_method', $this->turboRefreshMethod(...), ['is_safe' => ['html']]),
             new TwigFunction('turbo_refresh_scroll', $this->turboRefreshScroll(...), ['is_safe' => ['html']]),
         ];
-    }
-
-    /**
-     * @param object|string|array<object|string> $topic
-     */
-    public function turboStreamListen(Environment $env, $topic, ?string $transport = null): string
-    {
-        $transport ??= $this->default;
-
-        if (!$this->turboStreamListenRenderers->has($transport)) {
-            throw new \InvalidArgumentException(\sprintf('The Turbo stream transport "%s" does not exist.', $transport));
-        }
-
-        if (\is_array($topic)) {
-            $topic = new TopicSet($topic);
-        }
-
-        return $this->turboStreamListenRenderers->get($transport)->renderTurboStreamListen($env, $topic);
     }
 
     /**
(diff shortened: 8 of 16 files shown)


src/Turbo/assets/src/turbo_stream_controller.ts
```diff
@@ -50,7 +52,7 @@ export default class extends Controller {
 
     connect() {
         if (this.url) {
-            this.es = new EventSource(this.url);
+            this.es = new EventSource(this.url, this.eventSourceOptionsValue);
```
smnandre: EventSource does not seems to accept "options" but only `{ withCredentials: bool }` or do i read this wrong ? 

* https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource
* https://html.spec.whatwg.org/multipage/server-sent-events.html#eventsource

src/Turbo/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 -   Add `<twig:Turbo:Frame>` component
 -   Add support for custom actions in `TurboStream` and `TurboStreamResponse`
 -   Add support for providing multiple mercure topics to `turbo_stream_listen`
+-   Add support for providing options to the EventSource via `turbo_stream_listen`
```
smnandre: To move into a new 2.23 section

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -50,6 +56,24 @@ public function turboStreamListen(Environment $env, $topic, ?string $transport =
...
         }
 
-        return $this->turboStreamListenRenderers->get($transport)->renderTurboStreamListen($env, $topic);
+        if (
+            null !== $this->authorization
+            && null !== $this->requestStack
+            && (isset($options['subscribe']) || isset($options['publish']) || isset($options['additionalClaims']))
+            // @phpstan-ignore-next-line
+            && null !== $request = method_exists($this->requestStack, 'getMainRequest') ? $this->requestStack->getMainRequest() : $this->requestStack->getMasterRequest()
+        ) {
```
smnandre: What version does this target ?

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -50,6 +56,24 @@ public function turboStreamListen(Environment $env, $topic, ?string $transport =
...
+                $options['subscribe'] ?? [],
+                $options['publish'] ?? [],
+                $options['additionalClaims'] ?? [],
+                $transport,
+            );
+
+            unset($options['subscribe'], $options['publish'], $options['additionalClaims']);
+        }
+
+        return $this->turboStreamListenRenderers->get($transport)->renderTurboStreamListen($env, $topic, $options);
```
smnandre: You cannot pass options as third parameter, as the interface only declare "options" in docblock.

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -12,19 +12,24 @@
...
  * @author Kévin Dunglas <kevin@dunglas.fr>
+ * @author Pierre Ambroise <pierre27.ambroise@gmail.com>
  */
 final class TwigExtension extends AbstractExtension
 {
     public function __construct(
         private ContainerInterface $turboStreamListenRenderers,
         private string $default,
+        private ?Authorization $authorization = null,
+        private ?RequestStack $requestStack = null,
```
smnandre: We should then declare a Twig Runtime .. as this should probably not be in the hot path.

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -12,19 +12,24 @@
 namespace Symfony\UX\Turbo\Twig;
 
 use Psr\Container\ContainerInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\Mercure\Authorization;
 use Symfony\UX\Turbo\Bridge\Mercure\TopicSet;
```
smnandre: I think you could/should probably use TopicSet here, as it's specific to Mercure and would keep things properly typed / isolated.

Or any strategy that does not impact users using a different solution than Mercure.

You should look at this PR and the discussion, as most challenges are identical

https://github.com/symfony/ux/pull/2407

smnandre (changes requested):

src/Turbo/src/Twig/TwigExtension.php
Fan2Shrek: I'm not really sure how to do that, could you explain it to me ?

src/Turbo/src/Twig/TwigExtension.php
Fan2Shrek: I don't see where I could use this object, could you add more details ?

src/Turbo/src/Twig/TurboStreamListenRendererInterface.php
```diff
@@ -23,5 +23,5 @@ interface TurboStreamListenRendererInterface
     /**
      * @param string|object $topic
      */
-    public function renderTurboStreamListen(Environment $env, $topic): string;
+    public function renderTurboStreamListen(Environment $env, $topic /* , array $eventSourceOptions = [] */): string;
```
smnandre, line 26: Regarding the interface and the BC promise

This is a good way to expand an interface while not making current of previous code break. 

But then it also means some class can implement TurboStreamListenRendererInterface without accepting this new parameter.

If you pass a third arrgument to such class, it will break. 

This is why that kind of changes cannot be done (not my opinion here, it's symfony BC promise: https://symfony.com/doc/current/contributing/code/bc.html#changing-interfaces )

Before calling renderTurboStreamListen() on an object implementing TurboStreamListenRendererInterface,  you  need to check if this object is ready to receive this third parameter. Or the best way to do this is to have a dedicated interface. 

That means we probably need to
* create a new interface
* make the mercure streamlistener implement this interface
* type check this on the twig extension/runtime before passing the argument

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -11,45 +11,19 @@
...
 
-use Psr\Container\ContainerInterface;
-use Symfony\UX\Turbo\Bridge\Mercure\TopicSet;
-use Twig\Environment;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
 /**
  * @author Kévin Dunglas <kevin@dunglas.fr>
+ * @author Pierre Ambroise <pierre27.ambroise@gmail.com>
```
smnandre: Not sure here 😅

src/Turbo/src/Twig/TurboRuntime.php
```diff
@@ -0,0 +1,71 @@
...
+                $options['subscribe'] ?? [],
+                $options['publish'] ?? [],
+                $options['additionalClaims'] ?? [],
+                $transport,
+            );
+
+            unset($options['subscribe'], $options['publish'], $options['additionalClaims']);
+        }
+
+        return $this->turboStreamListenRenderers->get($transport)->renderTurboStreamListen($env, $topic, $options);
```
smnandre: Here you have no way to know if renderTurboStreamListen accept the "options" parameters

src/Turbo/src/Twig/TurboRuntime.php
```diff
@@ -0,0 +1,71 @@
...
+
+        if (!$this->turboStreamListenRenderers->has($transport)) {
+            throw new \InvalidArgumentException(\sprintf('The Turbo stream transport "%s" does not exist.', $transport));
+        }
+
+        if (\is_array($topic)) {
+            $topic = new TopicSet($topic);
+        }
+
+        if (
```
smnandre: This should be implemented in the Mercure streamListenRenderers as Authorization here is only a Mercure one.

Fan2Shrek:
Hey @smnandre,
Thanks for taking the time to explain this to me ! 😊

However, the last commit seems a bit off to me. Could you confirm if that's what you had in mind ?
Also, since the cookie logic was moved to the Mercure TurboStreamListenRenderer, is the Twig runtime still relevant ?

src/Turbo/src/Twig/TwigExtension.php
smnandre: I did not mean "use this one", but if you read the PR linked and the discussion, you will see how it relates to yours.

We decided to introduce a new object to bypass these BC questions.

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
```diff
@@ -55,6 +63,29 @@ public function renderTurboStreamListen(Environment $env, $topic): string
...
+                && (isset($eventSourceOptions['subscribe']) || isset($eventSourceOptions['publish']) || isset($eventSourceOptions['additionalClaims']))
+                && null !== $request = $this->requestStack->getMainRequest()
+            ) {
+                $this->authorization->setCookie(
+                    $request,
+                    $eventSourceOptions['subscribe'] ?? [],
+                    $eventSourceOptions['publish'] ?? [],
+                    $eventSourceOptions['additionalClaims'] ?? [],
+                    $eventSourceOptions['transport'] ?? null,
+                );
```
smnandre: This will throw if the method is called twice during the render of the page.

What are the goal of the various options here ? 

Or, say differently, why not use the mercure extension here ?  (real question i'm not sure to get)

Fan2Shrek: Do you mean the Mercure Twig extension?
If so, it doesn't seem to work 😅. The dependency isn't being resolved, and to be honest, I'm not sure why.

smnandre:
Hi [seb-jean](https://github.com/seb-jean) [Guervyl](https://github.com/Guervyl) [gremo](https://github.com/gremo) [DRaichev](https://github.com/DRaichev)  and anyone who want to :) 


Well.... It’s time to lay all this out clearly! 😊

**Could you share your ideal Developer Experience (DX) suggestions?**
As much in PHP code than in Twig // Give examples of code usage
Once we have a solid understanding of what you’re aiming for, we can figure out the best way to make it happen together.

**Please clarify where and how you’d use this feature:**
For example, in a page template, a component, a full-page view, multiple times, or other contexts. This will help us better target the needs and define responsibilities.

Looking forward to your insights!

smnandre:
@norkunas as i think you are one of the (if not the) person knowing this most, and having experience with this BC problematic... i'd love to have your input on this.

@Fan2Shrek  I've opened a meta-issue to see how we can make changes in a more generic way first, to ease the development of this specific need, but also others after.

--> https://github.com/symfony/ux/issues/2460

seb-jean:
Hi,

This feature would be interesting for me: I would use it with https://github.com/symfony/ux/pull/327.

Let's take the example of GitHub notifications.

I have a `Notification` entity:

```php
// src/Entity/Notification.php
namespace App\Entity;

use Symfony\UX\Turbo\Attribute\Broadcast;

#[Broadcast(topics: ['@="notifications_list_" ~ entity.getUser().getId()'], private: true)]
class Notification
{
    // ...
}
```

With each creation, a notification will be automatically added with the [Broadcast system](https://symfony.com/bundles/ux-turbo/current/index.html#broadcast-options).

A user can then view their own notifications with `{{ turbo_stream_listen('notifications_list_' ~ app.user.id) }}`.

The generated code will be as follows:


```html
<div data-controller="symfony--ux-turbo--mercure-turbo-stream" data-symfony--ux-turbo--mercure-turbo-stream-hub-value="https://symfony.com/.well-known/mercure" data-symfony--ux-turbo--mercure-turbo-stream-topic-value="notifications_list_1"></div>
```

Unfortunately, this is not secure because a malicious person will be able to change the value of the `data-symfony--ux-turbo--mercure-turbo-stream-topic-value` attribute to have notifications from another user.

Here I talked about notifications, but it could also be valid for instant messaging for example.

The idea would then be to secure the `turbo_stream_listen()`.

smnandre:
Let's add this to the conversatioon about stream sources : https://github.com/symfony/ux/issues/2460 

I'd like to keep talking about generic implementation first (but based on _your_ specific need, it's a rich source of feedback / irl usage!)

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
tibobaldwin: I might be getting ahead of myself, but can't we pass something like this to use `mercure` Twig extension?

```php
use Symfony\Component\Mercure\Twig\MercureExtension;

|...]

$mercureExtension = $container->get(MercureExtension::class);
$mercureExtension->mercure($topics, $options);
```

tibobaldwin:
What is blocking this PR? :blush:

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
smnandre: Ithink $twig->getExtension(...) would be enough  but yes 100% 

(and probably check/throw before if it does not exist... while not forcing everyone to have it)

smnandre:
Well, sometimes it feels like I'm talking to the void, but yeah, let's merge this PR! 😅

So, we need to:

    Rebase it
    Add a test or two
    Fix a few minor issues

Then, we just need to prepare a short paragraph of documentation.

And finally, let's hope there will be as many people to help, debug, and test when things change on the Mercure or Turbo side! 👼

Good for you, @Fan2Shrek? If you'd like to help, @tibobaldwin, feel free to jump in! 😊

src/Turbo/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23.0
+
+-   Add support for providing options to the EventSource via `turbo_stream_listen`
```
smnandre: I think we should only use the term "authenticated" or "credential based" event source... because the options is not open to all transports nor all keys

src/Turbo/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23.0
```
smnandre:
```suggestion
## 2.24.0
```

src/Turbo/config/services.php
```diff
@@ -45,9 +46,15 @@
             ->decorate('turbo.broadcaster.imux')
 
         ->set('turbo.twig.extension', TwigExtension::class)
-            ->args([tagged_locator('turbo.renderer.stream_listen', 'transport'), abstract_arg('default')])
             ->tag('twig.extension')
 
+        ->set('turbo.twig.runtime', TurboRuntime::class)
+            ->args([
+                tagged_locator('turbo.renderer.stream_listen', 'transport'),
+                abstract_arg('default'),
```
smnandre:
```suggestion
                abstract_arg('default transport'),
```

src/Turbo/src/Twig/TurboRuntime.php
```diff
@@ -0,0 +1,54 @@
...
+use Symfony\UX\Turbo\Bridge\Mercure\TopicSet;
+use Twig\Environment;
+use Twig\Extension\RuntimeExtensionInterface;
+
+/**
+ * @author Pierre Ambroise <pierre27.ambroise@gmail.com>
+ *
+ * @internal
+ */
+class TurboRuntime implements RuntimeExtensionInterface
```
smnandre:
```suggestion
final class TurboRuntime implements RuntimeExtensionInterface
```

src/Turbo/src/Twig/TurboRuntime.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public function __construct(
+        private ContainerInterface $turboStreamListenRenderers,
+        private string $default,
```
smnandre:
```suggestion
        private readonly string $defaultTransport,
```

? (and update code below)

smnandre (changes requested):
Minor things and let's merge

src/Turbo/src/Twig/TurboRuntime.php
Fan2Shrek: Question for my knowledge: should I remove the `@internal` in this case?

smnandre: Not the same goal i'd say. 

`final`:  you cannot customize this one. Use it, decorate it, or leave it like that.
`@internal `:  we do not promise the methods will stay the same // you are not supposed to call this class directly, if you do it and we change something, you'd be warned 

:)

src/Turbo/doc/index.rst
```diff
@@ -750,6 +750,9 @@ Let's create our chat::
         </turbo-frame>
     {% endblock %}
 
+If you're using a private hub, you can add ``{ withCredentials: true }``
+as the third argument of ``turbo_stream_listen()`` to authenticated with the hub
```
smnandre:
```suggestion
as ``turbo_stream_listen()`` third argument to authenticate with the hub
```

wdyt ?

smnandre:
TwigComponent errors are unrelated

smnandre (approved):
@Fan2Shrek it took too much time, i'm sorry for this. Thank you very much for this work.

@seb-jean & @tibobaldwin  thank you for the recent pokes :)

Fan2Shrek:
Hey @smnandre, thanks for the help! Unfortunately, there's still a problem with the `TurboStreamListenRenderer`. It doesn't create the cookie—I think it's because the class dependencies aren't resolved, but I have no clue why.

Fan2Shrek:
I found that `TurboStreamListenRenderer` is configured by the [Mercure bundle here](https://github.com/symfony/mercure-bundle/blob/77435d740b228e9f5f3f065b6db564f85f2cdb64/src/DependencyInjection/MercureExtension.php#L211C17-L211C19). Should we open a PR there as well?

Using Mercure to create the cookies causes the tests to fail.

smnandre:
I promise you I did not remember this... but now it happens it feels a bit ironic 😆 

Well, let's move on :) 

You can try this: https://github.com/symfony/ux/pull/2447#discussion_r1966598308 

The only think you cannot is "configure" a service from another bundle/bridge/etc.

So if you find a way to inject the service directly it's perfect (just use ignoreOnInvalid to not pollute our code if there were a problem one day)

If you don't, I'll have a look later during the night 🦇   or tomorrow

Fan2Shrek:
I adapted the code to use the Twig Mercure extension.
For the injection part, I'm not good enough to find a proper solution. 😅

smnandre:
## Well well well ... 👼  (1/1)

Capture d’écran 2025-02-23 à 01 21 57

https://github.com/symfony/ux/issues/2460


## Well well well .... 👼  (2/2)

```php
     *
     * @return string The URL of the hub with the appropriate "topic" query parameters (if any)
     */
    public function mercure($topics = null, array $options = []): string
```
[https://github.com/symfony/mercure[•••]src/Twig/MercureExtension.php](https://github.com/symfony/mercure/blob/304cf84609ef645d63adc65fc6250292909a461b/src/Twig/MercureExtension.php#L48-L51)

smnandre:
>  For the injection part, I'm not good enough to find a proper solution. 😅

Does it work ? If not, what is missing here ? How do you test?

Fan2Shrek:
I tested it on a small app with Mercure already set up. The only problem I’m facing is how to inject `TwigEnvironment` to get `MercureExtension`. Since `TurboStreamRenderer` is configured by the `MercureBundle`, it doesn't inject the new dependencies.

I’m not sure what the 'proper' way to solve this problem is.

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
```diff
@@ -23,14 +26,15 @@
...
-final class TurboStreamListenRenderer implements TurboStreamListenRendererInterface
+final class TurboStreamListenRenderer implements TurboStreamListenRendererWithOptionsInterface
 {
     private StimulusHelper $stimulusHelper;
 
     public function __construct(
         private HubInterface $hub,
         StimulusHelper|StimulusTwigExtension $stimulus,
         private IdAccessor $idAccessor,
+        private Environment $twig,
```
smnandre, line 36: did you update the config for this argument ?

Fan2Shrek: Can I? The config is handled by MercureBundle.

smnandre: How so ?

Fan2Shrek: Isn't this the line that handles it?
https://github.com/symfony/mercure-bundle/blob/77435d740b228e9f5f3f065b6db564f85f2cdb64/src/DependencyInjection/MercureExtension.php#L211C17-L211C19

smnandre: Well... back to step1. 

Honestly it makes not much sense to me MercureBundle is the one configuring classes on this bundle :| 

Could you start a conversation on the MercureBundle repository and see what's the best way to handle this ?

Fan2Shrek:
Hey @smnandre, this finally looks good to me :)
Let me know if there are any details to adjust .

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
```diff
@@ -64,12 +63,14 @@ public function renderTurboStreamListen(Environment $env, $topic /* array $event
...
-            if (isset($eventSourceOptions['withCredentials'])) {
-                $controllerAttributes['withCredentials'] = $eventSourceOptions['withCredentials'];
+            try {
+                $mercure = $this->twig->getExtension(MercureExtension::class);
+                $mercure->mercure($topics, $eventSourceOptions);
+
+                if (isset($eventSourceOptions['withCredentials'])) {
+                    $controllerAttributes['withCredentials'] = $eventSourceOptions['withCredentials'];
+                }
+            } catch (RuntimeError $e) {
```
smnandre, line 73: Why this ?

src/Turbo/src/DependencyInjection/Compiler/RegisterMercureHubs.php
```diff
@@ -0,0 +1,51 @@
...
+            if (isset($tag['default']) && $tag['default']) {
+                $container->getDefinition("turbo.mercure.{$name}.renderer")
+                    ->addTag('turbo.renderer.stream_listen', ['transport' => 'default']);
+            }
+
+            $container->register("turbo.mercure.$name.broadcaster", Broadcaster::class)
+                ->addArgument($name)
+                ->addArgument(new Reference($hubId))
+                ->addTag('turbo.broadcaster');
+        }
```
smnandre: Could we have some test here ?

It does not seem to follow the findTaggedServiceIds() doc (it returns a array of tags, not a single one)

https://github.com/symfony/symfony/blob/6477041b05f95511c6c0a11bff062b1e6fd8ece6/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L1321-L1338

src/Turbo/src/DependencyInjection/Compiler/RegisterMercureHubs.php
```diff
@@ -0,0 +1,51 @@
...
+use Symfony\UX\Turbo\Bridge\Mercure\Broadcaster;
+use Symfony\UX\Turbo\Bridge\Mercure\TurboStreamListenRenderer;
+
+/**
+ * This compiler pass ensures that TurboStreamListenRenderer
+ * and Broadcast are registered per Mercure hub.
+ *
+ * @author Pierre Ambroise<pierre27.ambroise@gmail.com>
+ */
+final class RegisterMercureHubs implements CompilerPassInterface
```
smnandre:
```suggestion
final class RegisterMercureHubsPass implements CompilerPassInterface
```

I know we have a couple without in the UX repository right now, but it feels more coherent to follow Symfony codebase conventions

smnandre:
@Fan2Shrek sorry I was off ! Looks good to me, some minor (last) fixes and LGTM :)

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
Fan2Shrek: This can be thrown by `$this->twig->getExtension`, 
Should I remove it anyway?

src/Turbo/src/DependencyInjection/Compiler/RegisterMercureHubs.php
Fan2Shrek: You're right, I've added tests :)

smnandre:
Thank you very much for your work on this @Fan2Shrek !

tibobaldwin:
Can this feature be merged?

Kocal:
There are some conflicts, I'm taking care of them

Kocal (approved):

Kocal:
Thank you @Fan2Shrek.

PR #2450, opened 2024-12-18
[StimulusBundle] Attribute Builder

### LINKED ISSUE #737: PHP functions to add Stimulus JS attributes
It would be nice to have functions or options to add Stimulus attributes on form widgets.

Something like

```php
$builder->add('country', CountryType::class, [
    'attr' = stimulus_target('path/to/controller', 'targetName')
]);
```

or

```php
$builder->add('country', CountryType::class, [
    'stimulus_target' = ['path/to/controller', 'targetName']
]);
```

Same goes for `stimulus_action` and `stimulus_controller`

19Gerhard85:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #737 
| License       | MIT

This PR adds a FromTypeExtension and adds the options `stimulus_controller`, `stimulus_action` and `stimulus_target`.

Supported syntax:

**stimulus_controller**
```php
//Single controller
'stimulus_controller' => 'controllerName',

//Single controller with path
'stimulus_controller' => 'path/controllerName',

//Single controller with path, already in correct nomenclature
'stimulus_controller' => 'path--controllerName',

//Multiple controllers
'stimulus_controller' => ['controllerName1', 'controllerName2'],

//Single controller with values, classes and targets
'stimulus_controller' => [
    'controllerName' => [
        'values' => [
            'key' => 'value'
        ],
        'classes' => [
            'key' => 'value'
        ],
        'targets' => [
            'otherControllerName' => '.targetName'
        ]
    ]
]
```

**stimulus_action**
```php
//Default event, already in correct nomenclature
'stimulus_action' => 'controllerName#actionName'

//Defined event, already in correct nomenclature
'stimulus_action' => 'eventName->controllerName#actionName'

//Default event in array notation
'stimulus_action' => ['controllerName' => 'actionName']

//Defined event in array notation
'stimulus_action' => ['controllerName' => ['eventName' => 'actionName']]

//Defined event in array notation with parameters
'stimulus_action' => ['controllerName' => ['eventName' => ['actionName' => ['param1' => 'value1']]]]
```

**stimulus_target**
```php
//Single target in array notation
'stimulus_target' => ['controllerName' => 'targetName']

//Multiple targets of the same controller, already in correct nomenclature
'stimulus_target' => ['controllerName' => 'targetName1 targetName2']

//Multiple targets of same controller in array notation
'stimulus_target' => ['controllerName' => ['targetName1', 'targetName2']]

//Single/Multiple targets of different controllers
'stimulus_target' => ['controllerName' => ['targetName1', 'targetName2'], 'otherController' => 'otherTargetName']
```

Tests will follow when we finalized the syntax.


src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
```
smnandre:
```suggestion
```

There is no strict_types in the Symfony codebase, afaik :)

smnandre:
Continuing here the discussion from the issue https://github.com/symfony/ux/issues/737#issuecomment-2550799682

> > Do we want only array_like structures ? What about some DTO ?
> 
> Arrays are usually a pain to deal with, and more in this case when you won't be able to use PHPDoc on them. It's like the Twig function `stimulus_*` we introduced a long time ago, because writting plain HTML attributes were painful.
> 
> Using a new kind of `StimulusAttributes` builder, as you suggested @smnandre, is a good idea to me.


The main question is "what is the end goal ?"

This first draft only handle stimulus_*  on the form type, but i'm not sure if this is the expected behaviour. 

How can i set an attribute on the row and another on the widget ? 

Right now user can set attributes (classes but also data-**) on `attr` / `row_attr`  etc..

smnandre:
Maybe simple static method could do most of the needs here? 

```php

Stimulus::controller(....)

Stimulus::target($controller, $target)

```

or builder-like

```php
Stimulus::controller(name)
    ->value(name, value)
    ->class(name, value)
    ->...    
```

19Gerhard85:
> Continuing here the discussion from the issue [#737 (comment)](https://github.com/symfony/ux/issues/737#issuecomment-2550799682)
> 
> > > Do we want only array_like structures ? What about some DTO ?
> > 
> > 
> > Arrays are usually a pain to deal with, and more in this case when you won't be able to use PHPDoc on them. It's like the Twig function `stimulus_*` we introduced a long time ago, because writting plain HTML attributes were painful.
> > Using a new kind of `StimulusAttributes` builder, as you suggested @smnandre, is a good idea to me.
> 
> The main question is "what is the end goal ?"
> 
> This first draft only handle stimulus_* on the form type, but i'm not sure if this is the expected behaviour.
> 
> How can i set an attribute on the row and another on the widget ?
> 
> Right now user can set attributes (classes but also data-**) on `attr` / `row_attr` etc..

The FormTypeExtension works on forms and widgets, as long as the widget extends the AbstractType.

`row_attr` and `choice_attr` implemented now, unfortunately choice_attr isn't working at all, e.g.

```php
        $builder->add('select', ChoiceType::class, [
            'expanded' => true,
            'choices'  => ['foo' => 'foo', 'bar' => 'bar'],
            'row_attr' => [
                'class' => 'rowClass',
                'stimulus_controller' => ['row/Controller']
            ],
            'choice_attr' => [
                'class' => 'choiceClass',
                'stimulus_controller' => ['widget/Controller']
            ],
        ]);
```

`choiceClass` isn't added to the radio button.

```html
<div class="rowClass form-group" data-controller="row--Controller">
    <label class="control-label">Select</label>
    <div id="test_form_select">
        <div class="radio">
            <label for="test_form_select_placeholder" class="">
                <input type="radio" id="test_form_select_placeholder" name="test_form[select]" value="" checked="checked"> None
            </label>
        </div>
        <div class="radio">
            <label for="test_form_select_0" class="">
                <input type="radio" id="test_form_select_0" name="test_form[select]" value="foo"> foo
            </label>
        </div>
        <div class="radio">
            <label for="test_form_select_1" class="">
                <input type="radio" id="test_form_select_1" name="test_form[select]" value="bar"> bar
            </label>
        </div>
    </div>
</div>
```

smnandre:
> row_attr and choice_attr implemented now, unfortunately choice_attr isn't working at all, e.g.

In fact, [choice_attr](https://symfony.com/doc/current/reference/forms/types/choice.html#choice-attr) works "per choice". 

Examples from the doc: 

```php
    'choice_attr' => [
        'Apple' => ['data-color' => 'Red'],
        'Banana' => ['data-color' => 'Yellow'],
        'Durian' => ['data-color' => 'Green'],
    ],
```    

```php
    'choice_attr' => function ($choice, string $key, mixed $value) {
        // adds a class like attending_yes, attending_no, etc
        return ['class' => 'attending_'.strtolower($key)];
    },
```

And there is a static helper to defer the computation

```
    'choice_attr' => ChoiceList::attr($this, function (?Category $category): array {
        return $category ? ['data-uuid' => $category->getUuid()] : [];
    }),
```



> The FormTypeExtension works on forms and widgets, as long as the widget extends the AbstractType.

Yep.. Do we want this is my point :) Or should this be configured / dynamic in any way ?

smnandre:
@19Gerhard85  This extension is a very good idea, thanks for following up and taking charge of the topic! 

And let's keep the motivation high, even if it takes a few iterations! We're aiming for solid choices for the future, so let's stay the course—it’ll be worth it!

## Stimulus logic / object

I don't think we should have the parsing "logic" coded in here..  using a small dedicated object will simplify the code & its maintenance, and allow us to is elsewhere.  This really is the occasion.

## Name

Let's rename the extension StimulusFormExtension, ok for you ?  (or anything with Stimulus in it :) ) 


I let some other comments in the code :)

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\UX\StimulusBundle\Dto\StimulusAttributes;
+use Twig\Environment;
+use Twig\Loader\ArrayLoader;
+
+class FormTypeExtension extends AbstractTypeExtension
+{
+    private StimulusAttributes $stimulusAttributes;
```
smnandre: This is temporary right ?

I think we won't need it and use a new object but if we don't, you can inject the StimulusHelper here

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+
+    public static function getExtendedTypes(): iterable
+    {
+        return [FormType::class];
+    }
+
+    public function buildView(FormView $view, FormInterface $form, array $options): void
+    {
+        if (
```
smnandre: In this method you repeat two times the same code, this can be reduce a lot i think

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+            if (isset($options['stimulus_target'])) {
+                $this->handleTarget($options['stimulus_target']);
+            }
+
+            if (isset($options['stimulus_action'])) {
+                $this->handleAction($options['stimulus_action']);
+            }
+
+            $attributes = array_merge($view->vars['attr'], $this->stimulusAttributes->toArray());
+            $view->vars['attr'] = $attributes;
```
smnandre:
```suggestion
            $view->vars['attr'] = [...$view->vars['attr'], ...$this->stimulusAttributes->toArray()];
```

Avoid unnecessary var

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+        if (
+            isset($options['stimulus_controller'])
+            || !isset($options['stimulus_target'])
+            || !isset($options['stimulus_action'])
+        ) {
+            $this->stimulusAttributes = new StimulusAttributes(new Environment(new ArrayLoader()));
+
+            if (isset($options['stimulus_controller'])) {
+                $this->handleController($options['stimulus_controller']);
+            }
```
smnandre:
```suggestion
            if ($options['stimulus_controller']) {
                $this->handleController($options['stimulus_controller']);
            }
            
            // same for the other ones
```

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+        return [FormType::class];
+    }
+
+    public function buildView(FormView $view, FormInterface $form, array $options): void
+    {
+        if (
+            isset($options['stimulus_controller'])
+            || !isset($options['stimulus_target'])
+            || !isset($options['stimulus_action'])
+        ) {
```
smnandre:
```suggestion
        if ($options['stimulus_controller'] || $options['stimulus_target'] || $options['stimulus_action']) {
```

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+    }
+
+    private function handleTarget(array $targets): void
+    {
+        foreach ($targets as $controllerName => $target) {
+            $this->stimulusAttributes->addTarget($controllerName, \is_array($target) ? implode(' ', $target) : $target);
+        }
+    }
```
smnandre: Can be removed / inline when it is called

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+                $attributes = array_merge($options[$index], $this->stimulusAttributes->toArray());
+                $view->vars[$index] = $attributes;
+            }
+        }
+    }
+
+    private function handleController(string|array $controllers): void
+    {
+        if (\is_string($controllers)) {
+            $controllers = [$controllers];
```
smnandre: Could be handle outside this method i suppose (will call $this->stimulusAttributes->addController(...) every case)

src/StimulusBundle/src/Form/Extension/FormTypeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+        foreach ($targets as $controllerName => $target) {
+            $this->stimulusAttributes->addTarget($controllerName, \is_array($target) ? implode(' ', $target) : $target);
+        }
+    }
+
+    private function handleAction(string|array $actions): void
+    {
+        // 'stimulus_action' => 'controllerName#actionName'
+        // 'stimulus_action' => 'eventName->controllerName#actionName'
+        if (\is_string($actions) && str_contains($actions, '#')) {
```
smnandre: The reason i want another object, we should not have to duplicate this in multiple place. Will open a draft later if i can

smnandre: (i cannot 😅 🕐 )

19Gerhard85:
> > row_attr and choice_attr implemented now, unfortunately choice_attr isn't working at all, e.g.
> 
> In fact, [choice_attr](https://symfony.com/doc/current/reference/forms/types/choice.html#choice-attr) works "per choice".
> 
> Examples from the doc:
> 
> ```
>     'choice_attr' => [
>         'Apple' => ['data-color' => 'Red'],
>         'Banana' => ['data-color' => 'Yellow'],
>         'Durian' => ['data-color' => 'Green'],
>     ],
> ```
> 
> ```
>     'choice_attr' => function ($choice, string $key, mixed $value) {
>         // adds a class like attending_yes, attending_no, etc
>         return ['class' => 'attending_'.strtolower($key)];
>     },
> ```
> 
> And there is a static helper to defer the computation
> 
> ```
>     'choice_attr' => ChoiceList::attr($this, function (?Category $category): array {
>         return $category ? ['data-uuid' => $category->getUuid()] : [];
>     }),
> ```
> 
> > The FormTypeExtension works on forms and widgets, as long as the widget extends the AbstractType.
> 
> Yep.. Do we want this is my point :) Or should this be configured / dynamic in any way ?

:grimacing: Sorry my fault > RTFM :sweat_smile:

19Gerhard85:
> @19Gerhard85 This extension is a very good idea, thanks for following up and taking charge of the topic!
> 
> And let's keep the motivation high, even if it takes a few iterations! We're aiming for solid choices for the future, so let's stay the course—it’ll be worth it!
> ## Stimulus logic / object
> 
> I don't think we should have the parsing "logic" coded in here.. using a small dedicated object will simplify the code & its maintenance, and allow us to is elsewhere. This really is the occasion.
> ## Name
> 
> Let's rename the extension StimulusFormExtension, ok for you ? (or anything with Stimulus in it :) )
> 
> I let some other comments in the code :)

An object is definitely a better choice and easier to maintain. We can also enforce a syntax for the parameters.

I have the next weeks off and will work on a nice object in "Builder style".

```
Stimulus::controller(name)
    ->value(name, value)
    ->class(name, value)
    ->...    
```

19Gerhard85:
@smnandre May you can take a look? I've changed the whole thing :)

smnandre:
I'll look tomorrow 👍

Kocal:

Thanks!

Status: Needs Work

PR #2451, opened 2024-12-18
[LiveComponent] Use container.build_hash instead of kernel.secret in CacheWarmer

### LINKED ISSUE #2419: [LiveComponent] TemplateCacheWarmer needs non available `kernel.secret` to build the template map
In https://github.com/symfony/ux/pull/1497 the `kernel.secret`  was introduced as a way to make the template map reproducible.

This is fine, except the `kernel.secret` value used is a **runtime value** (meant to be injected via `env()` or similar) and not necessarily known at build time. The secret being requested started breaking my build (in Docker) suddenly when upgrading to Symfony 7.2.

Since the usage of the secret is just to provide a stable seed, it doesn't need to use the actual secret, it can use the container checksum, just as similar uses in Symfony proper done recently by @nicolas-grekas, one example is https://github.com/symfony/symfony/pull/56831

It would make sense to swap out the use of the secret (runtime variable) with the container checksum (build time variable).

cc @stof

dkarlovi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2419 
| License       | MIT

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,7 @@
 # CHANGELOG
 
+-   Use container.build_hash instead of kernel.secret in CacheWarmer
+
 ## 2.22.0
 
 -   Remove CSRF tokens - rely on same-origin/CORS instead
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,7 +1,5 @@
 # CHANGELOG
 
--   Use container.build_hash instead of kernel.secret in CacheWarmer
-
 ## 2.22.0
 
 -   Remove CSRF tokens - rely on same-origin/CORS instead
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -18,6 +18,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
+use Symfony\Component\DependencyInjection\Parameter;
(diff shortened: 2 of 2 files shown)


smnandre:
Nice if this solution works for everyone! 

Depending on the feedback people give, we should probably add some tests to prevent problems in the future.

smnandre:
Should we do the same for the FingerprintCalculator / Hydrator ? 

(they should only be called on runtime)

dkarlovi:
AFAIK, they are actually using the secret as intended? The secret is required for cryptography there? I might be mistaken, feel free to correct me.

Edit: correct, I've checked it and the secret usage there is indeed cryptographic and correct. I've actually reverse engineered this specific part of code in a Node app (to enable the Node app to render correct live components), it must use the secret too (the secret is shared with the Node app). It would be cool if you could specific a LiveComponent-specific secret (so, not use `kernel.secret`) to improve the security, but that's not something we should do in this PR. @smnandre

smnandre (approved):
Can you add a line in the changelog ?

dkarlovi:
@smnandre done.

src/LiveComponent/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # CHANGELOG
 
+-   Use container.build_hash instead of kernel.secret in CacheWarmer
```
nicolas-grekas: I don't see this change as requiring a changelog entry, to me it's a "minor" change, not a new "feature", since it doesn't have any user-visible side effect except DX-related things (aka you cannot "build on it", while you can build on a "feature")

But if we want to keep it, I guess this needs a version:
```suggestion
## 2.23.0

-   Use container.build_hash instead of kernel.secret in CacheWarmer
```

dkarlovi: OK, feel free to edit this to your liking or reject the last commit.

smnandre (approved):

smnandre:
Thank you @dkarlovi.

dkarlovi:
@smnandre IMO a bugfix release with this would be in order since it's blocking using Symfony 7.2 in specific cases completely.

PR #2456, opened 2024-12-19
[TwigComponent] Document about unwanted behavior with `ExposeInTemplate` and computed methods

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

As spotted when working on my application, the following code triggers two times the methods `transactions()` because of the presence of `#[ExposeInTemplate]` (because before I used `transactions` instead of `computed.transactions`):

```php
	#[ExposeInTemplate]
    public function transactions(): array
    {
        // SQL query to database
    }
```

```twig
{{ computed.transactions|length }}
```

When I remove `#[ExposeInTemplate]`, then only one SQL query is made.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -382,7 +382,7 @@ use the full path of the template where the macro is defined:
     <twig:Alert>
         {# ❌ this won't work #}
         {% from _self import message_formatter %}
-        
+
         {# ✅ this works as expected #}
         {% from 'path/of/this/template.html.twig' import message_formatter %}
 
@@ -1552,6 +1552,12 @@ are called additional times, the cached value is used.
 
     Computed methods only work for component methods with no required
     arguments.
+
+.. tip::
+
+    Ensure to not use the ``ExposeInTemplate`` attribute on a computed method,
+    otherwise the method will be called twice instead of only once, leading to
+    unnecessary overhead and potential performance issues.
 
 Events
 ------


smnandre:
Nothing illogical to me, but a mention in documentation won't hurt and clarify this case. 👍 

Could you maybe use a  "tip" block  instead of a "warning" one here ?

src/TwigComponent/doc/index.rst
```diff
@@ -1553,6 +1553,12 @@ are called additional times, the cached value is used.
     Computed methods only work for component methods with no required
     arguments.
 
+.. warning::
+
+    Ensure to not use the ``ExposeInTemplate`` attribute on a computed method,
+    otherwise the method will be called twice instead of only once, leading to
```
smnandre, line 1559: It kinda leads to present it as a bug... but it is not illogical.

The calls are not made at the same time so the "twice" is maybe a bit missleading  
(first time to build the context of the templates, before render, and  one by you calling the cache object).

Kocal: Do you think instead, we should fix this behavior? 

I mean, throw an exception if the user tries to call a computed method with `#[ExposeInTemplate]` ?

smnandre: I am not sure we can. BUT we can maybe populate the computed before the render if method as the attribute. There are some limitations but this could work. 

Lets update doc for now and see after if we change and how wdyt ? 

(but id rather not use warning or any tone that can stress or frighten users 😇)

smnandre (approved):
We may want to make all these more clear with schema later.. 

But this precision is usefull right now, thx!

Issue #2455, opened 2024-12-19
[TwigComponent] Unexpected attribute parsing behavior


eokic:
These may be two very different reports with different opinions for each, so please let me know if you'd like me to separate them.

## ~Example 01: Passing attributes as-is~

```twig
<twig:Comp {{ attributes }} /> {# ❌ original question/report #}
<twig:Comp {{ ...attributes }} /> {# ✅ the correct way #}
```

A realistic usecase for this is creating higher-order (wrapper) components.

## Example 02

```twig
<twig:Comp
  {% if condition %}
    disabled
  {% endif %}
/>
```

Most notably, without these conditions, boolean attributes seem impossible (?), but would also just make the syntax more consistent with the way Twig is normally written.

-----

Both result in:

```
Expected attribute name when parsing the "<twig:Comp" syntax.
```

Kocal:
What about `<twig:Comp {{ ...attributes }} />`?

eokic:
> What about `<twig:Comp {{ ...attributes }} />`?

Ah dangit, somehow that documentation bit kept eluding me 🤦🏻. Maybe to mention it again briefly under `#component-attributes`? That's definitely where I've spent most of my time looking for it. Spread seems to work as expected so I'll give it a go, thank you!

Kocal:
> Maybe to mention it again briefly under `#component-attributes`? That's definitely where I've spent most of my time looking for it. 

Mmmh, I don't know, because it is related to the HTML syntax 🤔 

But maybe it can be interesting to also have it in this section, do you feel opening a PR? 🙏🏻

eokic:
> Mmmh, I don't know, because it is related to the HTML syntax 🤔

Completely understandable; I can only speak from my own experience as someone who's been periodically consuming that documentation over the course of a few months and yet still managed to not stumble upon the info in question 😅

> But maybe it can be interesting to also have it in this section, do you feel opening a PR? 🙏🏻

Sure thing, opened as https://github.com/symfony/ux/pull/2457

smnandre:
```
 <twig:Comp
  {% if condition %}
    disabled
  {% endif %}
/>
```

You cannot do this to pass parameters to an include or embed block in Twig. So this is all but a bug :) 

I’m closing here as thanks to you the first item will be better documented.

eokic:
> You cannot do this to pass parameters to an include or embed block in Twig. So this is all but a bug :)

[...]

PR #2457, opened 2024-12-20
[TwigComponent] Documents the spread syntax under the "Component Attribute" section

eokic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | — 
| License       | MIT

Motivated by https://github.com/symfony/ux/issues/2455, adds a syntax reminder in a place where many readers might be looking for it.

Used an inline code sample to keep the note minimal and unobtrusive, but I can update it with a proper code block if needed.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -929,6 +929,10 @@ component's template:
       My Component!
     </div>
 
+.. note::
+
+    Quick reminder that passing an array of attributes to the *Component* requires the spread operator syntax (``<twig:MyComponent {{ ...attributes }} />``). See `Component HTML Syntax`_.
+
 When rendering the component, you can pass an array of html attributes to add:
 
 .. code-block:: html+twig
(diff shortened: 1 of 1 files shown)


smnandre:
If you can link to the html syntax block of the same doc it would be better, we dont want user of TwigComponent only to be confused if they read Live Component doc with our realizing it.

Also you’re right to suggest this and yes I would do a small block with a line or text to explain the need (array or dynamic Attributes) and a little block code showing how to pass and id and a disabled true/false atrribute like that.

Tell me if you need any help!

src/TwigComponent/doc/index.rst
```diff
@@ -929,6 +929,10 @@ component's template:
       My Component!
     </div>
 
+.. note::
+
+    Quick reminder that passing an array of attributes to the *Component* requires the spread operator syntax (``<twig:MyComponent {{ ...attributes }} />``). See `Component HTML Syntax`_.
+
```
smnandre, line 935: Hmm in fact this may be confusing. Using the  `attributes` variable here may be missunderstood as we are talking about the component attributes just after.

What could be done here is a reminder link targetting [Passing & Rendering Attributes](https://symfony.com/bundles/ux-twig-component/current/index.html#passing-rendering-attributes) in the first pargraph of [Component Attributes](https://symfony.com/bundles/ux-twig-component/current/index.html#component-attributes)

smnandre (changes requested):
(cf the comment about component attributes / passing attributes)

Kocal:

PR #2462, opened 2024-12-21
[LiveComponent] Allow configuring secret for fingerprints and checksums

### LINKED ISSUE #2453: [LiveComponent] Allow specifying a live-component specific secret which defaults to kernel.secret

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #2453
| License       | MIT

Allow to configure a dedicated secret (used in FingerprintCalculator and LiveComonentHydrator)

Suggested by @dkarlovi in #2453 
Implementation inspired by [symfony #56840](https://github.com/symfony/symfony/pull/56840)

Should be merged _after_  #2461

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23.0
+
+-   Allow configuring the secret used to compute fingerprints and checksums.
+
 ## 2.22.0
 
 -   Remove CSRF tokens - rely on same-origin/CORS instead
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -12,6 +12,10 @@
 namespace Symfony\UX\LiveComponent\DependencyInjection;
 
 use Symfony\Component\AssetMapper\AssetMapperInterface;
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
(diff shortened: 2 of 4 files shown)


Kocal (approved):
It's an interesting approach, are there any plans to open a PR on symfony/recipes?

smnandre:
Not at all, why?

kbond:
Is there any considerations we need to make regarding: https://symfony.com/blog/new-in-symfony-7-2-optional-secret ?

smnandre:
@kbond  I don't think so, if i understand this correctly

> With these changes, if you enable a feature that requires a secret but have not configured one, Symfony will throw an exception with a clear message explaining how to resolve the issue.

But that's one of the reasons this PR checks for empty secrets: https://github.com/symfony/ux/pull/2461)

Kocal:
> Not at all, why?

For pure user-DX, I believe creating recipes with commented/default options is nice to have

smnandre:
This would be the only options for the live_component package.. so i'm not sure we want to annoy existing apps with a new file for this only..  it's more a power-user to me and can lead to very bad/insecure choices :| 

I guess this is something we will do easier when centralized (at least a bit) in UX Bundle

Kocal:
Makes sense, thanks! :)

kbond (approved):

PR #2464, opened 2024-12-22
[TwigComponent] Minimal support of comment lines

smnandre:
Twig introduced the inline comments in https://github.com/twigphp/Twig/pull/4349

This PR add minimal support for it the PreLexer / HTML syntax

```twig
<twig:Button
    # comment
    bar="bar"
/>
```

I'd like some IRL feedbacks on this one :)

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -233,6 +233,7 @@ class TwigPreLexer
             $isAttributeDynamic = false;
 
             // :someProp="dynamicVar"
+            $this->consumeWhitespace();
             if ($this->check(':')) {
                 $this->consume(':');
                 $isAttributeDynamic = true;
@@ -244,6 +245,7 @@ class TwigPreLexer
 
             // <twig:component someProp> -> someProp: true
             if (!$this->check('=')) {
+                $this->consumeWhitespace();
                 // don't allow "<twig:component :someProp>"
                 if ($isAttributeDynamic) {
                     throw new SyntaxError(\sprintf('Expected "=" after ":%s" when parsing the "<twig:%s" syntax.', $key, $componentName), $this->line);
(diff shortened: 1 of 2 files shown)


Kocal (commented):
Cool! 

I don't have real use-cases from the projets I'm working on, but I believe it will be super-useful (even if I believe some comments can be avoid by using good attribute's name or variable!) 

However, introducing a new syntax means that tools manipulating ASTs (like IDEs, Twig-CS-Fixer, maybe TwigStan, and syntax highlighting) will break, right?

smnandre:
Not at all, they already are compatible. I'm just following Twig syntax here, not adding anything new. But that means adapting the PreLexer we use to convert HTML Syntax into regular Twig code :)

Kocal:
I mean, for tools that does not use our `PreLexer`.

I've tested with Twig-CS-Fixer and no issues so far, but I think it does not support Twig Component at all.
I don't know about TwigStan (which I believe indirectly use our `PreLexer`).

But, for syntax coloration it does not work, and I don't know who/when this new syntax will be added to IDE and other syntax coloration libraries: 
image

smnandre:
> I've tested with Twig-CS-Fixer and no issues so far, but I think it does not support Twig Component at all.

They do, and they integrated inline comments right after Twig release :) 


> But, for syntax coloration it does not work, and I don't know who/when this new syntax will be added

Me neither but... how is this related to this PR ? :) I'm genuinely not sure to get the questions here.

Twig syntax changes are generally pretty quickly integrated into PhpStorm Twig integration (or in the Symfony plugin).

94noni (approved):
Would be very welcome to allow props comment next to/along the usage in the template👍🏻

smnandre:
cc @Kocal @kbond @WebMamba ?

Kocal (approved):

kbond (approved):

Kocal:
Thanks Simon.

PR #2470, opened 2024-12-29
[Autocomplete] Fix invalid createQueryBuilder code example

tacman:
replace with creating it from the entity repository

### DIFF
--- a/src/Autocomplete/doc/index.rst
+++ b/src/Autocomplete/doc/index.rst
@@ -316,7 +316,7 @@ the consequent renders, the default ``query_builder`` will be used::
                 ->add('food', FoodAutocompleteField::class, [
                     'query_builder' => function (EntityRepository $er) {
                         return $er->createQueryBuilder('o')
-                            ->andWhere($qb->expr()->notIn('o.id', [$currentFoodId]));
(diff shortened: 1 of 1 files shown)


src/Autocomplete/doc/index.rst
```diff
@@ -316,7 +316,7 @@ the consequent renders, the default ``query_builder`` will be used::
                 ->add('food', FoodAutocompleteField::class, [
                     'query_builder' => function (EntityRepository $er) {
                         return $er->createQueryBuilder('o')
-                            ->andWhere($qb->expr()->notIn('o.id', [$currentFoodId]));
+                            ->andWhere($er->createQueryBuilder()->expr()->notIn('o.id', [$currentFoodId]));
```
smnandre: Nice catch, thank you!

I think we should use the same style than later in the file (L369..) as i fear calling `createQueryBuilder` twice could be confusing.

```
                       $qb = $er->createQueryBuilder('o');
                       
                       return $qb->andWhere($qb->expr()->notIn('o.id', [$currentFoodId]));
```

Kocal:
Status: Needs Work

src/Autocomplete/doc/index.rst
smnandre: Ended up doing it :)

smnandre:
Thank you @tacman.

PR #2483, opened 2025-01-05
[LiveComponent] Add support for downloading files from LiveActions (Experimental)

### LINKED ISSUE #1516: [LiveComponent] Live Actions cannot handle file downloads
As live actions are treated as regular controller actions, they should also support file responses.

```php

#[LiveAction]
public function submit(): BinaryFileResponse
{
    $this->validate();

    return $this->file($this->exporter->generate($this->items));
}

```

Currently, file downloads are not downloaded but parsed..?

Screenshot 2024-02-18 at 13 28 23

## Edit (current workaround)

```php
public function submit(UriSigner $uriSigner): Response
{
    $this->validate();


    $file = $this->exporter->generate($this->items)

    $url = $this->generateUrl('admin_download', ['file' => base64_encode($file)]);

    return $this->redirect($uriSigner->sign($url));
}
```

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Close #1516 
| License       | MIT

This pull request introduces a new experimental feature for handling file downloads in `LiveActions`.

> [!CAUTION]
> As stated here, this feature -if merged- will keep an experimental status, for everyone to test and see how it goes. We cannot ensure BC promise on features without testing it IRL first.

## TL;DR;

Send file

```php

 #[LiveAction]
public function getDocument(): Response
{
    $file = $this->projectDir . '/docs/legals/terms.pdf');

    return LiveResponse::file($file);
}

```

Stream files

```php

 #[LiveAction]
public function streamThisReport(): StreamedResponse
{
    // Get a stream, or generate
    $resource =  $this->readStream('file/big/report.csv');

    return LiveResponse::streamFile($resource, 'report.csv');
}

```

⚠️  We are working to keep state synchronization and stream working together.

## Also

I started the demo, just need to complete some texts + add a bit of style

## Scope of this

This PR taks a very narrow scope on this thing, i already started working on cool things for upload (as I also needed to for work..)

Let's stay focus on this here (discussions / issues open for any new idea / MR)


src/LiveComponent/assets/src/Backend/BackendResponse.ts
```diff
@@ -13,4 +13,8 @@ export default class {
 
         return this.body;
     }
+
+    async getBlob(): Promise<Blob> {
+        return await this.response.blob();
```
Kocal: Prefer `return` instead of `return await`: 
```suggestion
        return this.response.blob();
```

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -300,15 +300,60 @@ export default class Component {
...
+                    !(
+                        headers.get('Content-Disposition')?.includes('attachment') ||
+                        headers.get('Content-Disposition')?.includes('inline')
+                    ) ||
+                    !headers.get('Content-Disposition')?.includes('filename=')
+                ) {
+                    throw new Error('Invalid LiveDownload response');
+                }
+
+                const fileSize = Number.parseInt(headers.get('Content-Length') || '0');
```
Kocal: Do we really want to assume that if `Content-Length` header is absent, then it is equal to `0`? 
In which case the header can be absent? (I didn't see the rest of the PR yet!)

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -300,15 +300,60 @@ export default class Component {
...
+                if (fileSize > 10000000) {
+                    throw new Error('File is too large to download (10MB limit)');
+                }
```
Kocal: More correct and easier to understand:
```suggestion
                if (fileSize > 10 * 1024 * 1024) {
                    throw new Error('File is too large to download (10MB limit)');
                }
```

Also, why do you want to limit the file size? It can be a nice feature, but shouldn't it be configurable by the developer?

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -300,15 +300,60 @@ export default class Component {
...
+            if (headers.get('X-Live-Download')) {
+                if (
+                    !(
+                        headers.get('Content-Disposition')?.includes('attachment') ||
+                        headers.get('Content-Disposition')?.includes('inline')
+                    ) ||
+                    !headers.get('Content-Disposition')?.includes('filename=')
+                ) {
+                    throw new Error('Invalid LiveDownload response');
+                }
```
Kocal:
```suggestion
                const headerContentDisposition = headers.get('Content-Disposition');
                if (
                    !headerContentDisposition
                    || !(headerContentDisposition?.includes('attachment') || headerContentDisposition?.includes('inline')) 
                    || !headerContentDisposition?.includes('filename=')
                ) {
                    throw new Error('Invalid LiveDownload response');
                }
```

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -300,15 +300,60 @@ export default class Component {
...
+
+                const fileSize = Number.parseInt(headers.get('Content-Length') || '0');
+                if (fileSize > 10000000) {
+                    throw new Error('File is too large to download (10MB limit)');
+                }
+
+                const fileName = headers.get('Content-Disposition')?.split('filename=')[1];
```
Kocal:
```suggestion
                const fileName = headerContentDisposition.split('filename=')[1];
```

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -255,6 +258,13 @@ public function onKernelView(ViewEvent $event): void
             return;
         }
 
+        if ($event->getControllerResult() instanceof BinaryFileResponse) {
+            if (!$event->getControllerResult()->headers->has(self::DOWNLOAD_HEADER)) {
+
+            }
```
Kocal: Dead code

src/LiveComponent/src/LiveDownloadResponse.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function __construct(string|SplFileInfo $file, ?string $filename = null)
+    {
+        if (\is_string($file)) {
+            $file = new SplFileInfo($file);
+        }
+
+        if ((!$file instanceof SplFileInfo)) {
+            throw new \InvalidArgumentException(sprintf('The file "%s" does not exist.', $file));
+        }
```
Kocal: I'm not sure to understand when this code can be executed, `$file` can either be a `string` or `SplFileInfo`, and when it's a `string`, we re-assign `$file` to a `SplFileInfo`.

Kocal: Maybe the condition is wrong?

src/LiveComponent/src/LiveDownloadResponse.php
```diff
@@ -0,0 +1,38 @@
...
+
+        if ($file instanceof SplTempFileObject) {
+            $file->rewind();
+        }
+
+        parent::__construct($file, 200, [
+            self::HEADER_LIVE_DOWNLOAD => 1,
+            'Content-Disposition' => HeaderUtils::makeDisposition(HeaderUtils::DISPOSITION_ATTACHMENT, $filename ?? basename($file)),
+            'Content-Type' => 'application/octet-stream',
```
Kocal: Shouldn't we use the `$file`'s mime type?

src/LiveComponent/src/LiveDownloadResponse.php
```diff
@@ -0,0 +1,38 @@
...
+use Symfony\Component\HttpFoundation\HeaderUtils;
+
+/**
+ * @author Simon André <smn.andre@gmail.com>
+ */
+final class LiveDownloadResponse extends BinaryFileResponse
+{
+    public const HEADER_LIVE_DOWNLOAD = 'X-Live-Download';
+
+    public function __construct(string|SplFileInfo $file, ?string $filename = null)
```
Kocal: What about adding `\Symfony\Component\HttpFoundation\File\File` as type of `$file`? 
Even if `File` already extends `SplFileInfo`, it may be interesting to add it IMO

Kocal: Well, I see that `BinaryFileResponse::__construct()` _only_ accepts `\SplFileInfo|string`, let's keep it as it is.

Kocal (commented):
This solution is so simple 🤩

src/LiveComponent/assets/src/Component/index.ts
smnandre: This is the part we need to discuss. Don't forget the user has NOT accepted this download in any way for now.

Also, you cannot store 1GB in the DOM before saving the file.

We have options after: 
- streaming toward the FileStorage api (where user would then decide where and how to save the file)
- allowing this only on button with attribute "download" (my preference for now)

src/LiveComponent/assets/src/Component/index.ts
smnandre: > In which case the header can be absent?

It should not be. This method is not made to stream movies, but files generated using the liveprops values. 

Elsewhen, developers should send an URL to the JS and then open some stream, or redirect to another page, etc...

> that if Content-Length header is absent, then it is equal to 0

No, absolutely not. But if it is unknown, we need to compute it on the fly then. Not implemented yet, but linked to the other questions below :)

src/LiveComponent/src/LiveDownloadResponse.php
smnandre: I hesitated (keep dep light) and chose not to...

But in fact it's a good idea and Mime is a very small component.

So let's make it a requirement for LiveComponent ? Or only when using downloads / uploads ? 

wdyt ?

src/LiveComponent/src/LiveDownloadResponse.php
smnandre: What would be the goal to add File ?

src/LiveComponent/src/LiveDownloadResponse.php
Kocal: I suggest to put it in `suggest` `composer.json` section (I don't remember what is the rule in Symfony, cc @nicolas-grekas), and then runtime check if `symfony/mime` exists (when using downloads/uploads)

src/LiveComponent/src/LiveDownloadResponse.php
Kocal: Nothing except documentation/DX, when I see `string|`SplFileInfo`, I don't really know I can pass a `Symfony\Component\HttpFoundation\File\File` (except if I know its parent class)

But that's fine, let's keep aligned to `BinaryFileResponse` constructor.

src/LiveComponent/src/LiveDownloadResponse.php
smnandre: Pretty sure there is no suggest in Symfony composer packages..

We have other situations in UX where we do not require a package (i.e. in Autocomplete for Form ..)

My question was more: should we require it for everyone, or keep it as an "runtime" dependency ?

norkunas:
What about streams from flysystem? We'll need to always load files into memory?

src/LiveComponent/src/LiveDownloadResponse.php
Kocal: runtime dependency

smnandre:
> What about streams from flysystem? We'll need to always load files into memory?

With that method, i fear yes. That's why the limit. And browsers can change their policy soon to prevent abuse.

We also can ask for a Filesystem API grant.

Again, will need to test one way or the other, locally i can stream a full 8GB file in one second so ... :|

src/LiveComponent/src/LiveDownloadResponse.php
kbond: Why aren't you deferring to BinaryFileResponse for this logic?

kbond (commented):
This is awesome! I'll let someone else review the frontend code.

For the back end:
1. I wish we weren't restricted to `BinaryFileResponse` - I'd like to be able to use with StreamedResponse and a resource.
2. What are the requirements for a response to trigger as a download on the frontend? Just the special header? Could we add this header in the response event listener (detect if it's a download)?

smnandre:
> Why aren't you deferring to BinaryFileResponse for this logic?

How so ? BinaryFile does not add content disposition different than what you pass here as argument.

smnandre:
> 1. I wish we weren't restricted to BinaryFileResponse - I'd like to be able to use with StreamedResponse and a resource.

I need to continue looking, but the restriction is not here at all, there is just an helper we can provide to offer the best DX possible.

The difficulty is to ensure a ResponseStream will work (without crashing the LiveComponent thread) and this is where problems come from:
- unknown size is a bad signal 
- content type must be set and be legit

And regarding streaming. I made it work perfectly at home. But i have still not see any use case where downloading a 500MB file via a LiveComponent is a good idea 😅 


Anyway we'll need to test IRL a bit to get feedback, issues, etc etc

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -300,15 +300,59 @@ export default class Component {
...
+                if (fileSize > 10000000) {
+                    throw new Error('File is too large to download (10MB limit)');
+                }
+
+                const fileName = headerContentDisposition.split('filename=')[1];
+                if (!fileName) {
+                    throw new Error('No filename found in Content-Disposition header');
+                }
+
+                const blob = await backendResponse.getBlob();
```
kbond, line 330: Does this load the entire file (up to our max) into memory?

smnandre, line 314: It depends... if you trigger the "click" and the file start downloading I think browser play the "pipe" role here.

But during my test I managed to crashed pretty violently Chrome and Safari multiple times.

src/LiveComponent/src/LiveDownloadResponse.php
kbond: What's the reason for the header?

In the frontend, if not a standard/known content type, but has a content disposition (and an other requirements), trigger the download? This way we could say "to enable file downloads, return a response that has a content disposition"

norkunas:
> And regarding streaming. I made it work perfectly at home. But i have still not see any use case where downloading a 500MB file via a LiveComponent is a good idea 😅

But it's a good idea to reduce resource usage in case memory limit is low.

javiereguiluz:
Simon, thanks for this nice new feature!

My only concern is about the name of this class:

```php
return new LiveDownloadResponse($file);
```

`LiveDownloadResponse` is related to files but its name has nothing to do with files. Also, `download` is not strictly correct, because we can download files, embed files in the browser and stream files.

Here are some proposals for your consideration:

```php
return new LiveFileResponse($file);
return new FileLiveResponse($file);
return new FileResponse($file);
return new StreamResponse($file);
return new ResourceResponse($file);
```

smnandre:
We've been working on it with @kbond and decided not to extend Response at all... (users can use whatever Response they want in term of implementation.

...but we would provide a static helper to create the best DX possible.

```php
    // ... 

    return LiveResponse::file($myFile);
}
```

That way we can also offer 


```php
    // ... 

    return LiveResponse::streamFile($myFile);
}
```

And this would be perfect for other ideas we have next (no content, refresh...)


---

We are now trying to see the best way to send files _and_ save component state/update the DOM (not the case in my PR)

javiereguiluz:
Very nice! Thanks for the insights.

src/LiveComponent/assets/src/Component/index.ts
smnandre: No limit

src/LiveComponent/assets/src/Component/index.ts
smnandre: We'll see if this is really required

mahono:
Would it be possible to use a redirect? I would like to do pre-signed urls and let the download be directly handled by an S3 bucket.

kbond:
> Would it be possible to use a redirect? I would like to do pre-signed urls and let the download be directly handled by an S3 bucket.

Yes, that's possible now!

FluffyDiscord:
Can `yield` be used/supported instead of `echo` for streamed response? Asking for a support of running Symfony in worker mode, where echoes are forbidden.

smnandre:
So how do we move on here? @kbond your work on the streams can be added right now ? do we need more tests before opening an alpha/experimental phase ?

ThibaultJlvt:
Hi ! Any news on this PR ?

smnandre:
Open source has taken a back seat lately, but I hope to clean and merge this soon.

If this feature matters to you or your company, feel free to reach out—either by supporting a contributor through GitHub Sponsors or backing Symfony directly via the [official sponsorship page](https://symfony.com/sponsor).

Any support is appreciated and helps keep the community moving :)

Kocal:

Thanks!

Status: Needs Work

PR #2490, opened 2025-01-07
[LiveComponent] Check if accessed vars exist

### LINKED ISSUE #2487: [LiveComponent] Undefined array key "placeholder" after upgrading to 2.22.x
Hello,

recently after upgrading symfony/ux-live-components to 2.22.x I've encountered this exception:

**An exception has been thrown during the rendering of a template ("Warning: Undefined array key "placeholder"").**

Here's the reproducer: https://github.com/maciazek/symfony-ux-live-components-choices-reproducer

I have a custom DatalistType form field, which extends TextType to add HTML datalist into it. Inside DatalistType I'm using 'choices' option to add choices to datalist.
https://github.com/maciazek/symfony-ux-live-components-choices-reproducer/blob/master/src/Form/DatalistType.php#L11-L29
https://github.com/maciazek/symfony-ux-live-components-choices-reproducer/blob/master/templates/form/custom_types.html.twig

I have a form inside live component which is using DatalistType as a child field.
https://github.com/maciazek/symfony-ux-live-components-choices-reproducer/blob/ebe9ff77a7181ed6e0ca4dd5fca7483f38e8ed94/src/Form/ExampleType.php#L9-L30
https://github.com/maciazek/symfony-ux-live-components-choices-reproducer/blob/ebe9ff77a7181ed6e0ca4dd5fca7483f38e8ed94/src/Twig/Components/ExampleForm.php#L13-L23

Everything worked well until updating symfony/ux-live-components to 2.22.x - then I've encountered previously mentioned error.

After some investigation I've found that problem is caused by recent change in ComponentWithFormTrait:
https://github.com/symfony/ux-live-component/blob/5adf780aac0ccbd5cb1dddea683db9aa99b91d4a/src/ComponentWithFormTrait.php#L264-L271

I have 'choices' option in my DatalistType, but it's extending TextType which, as I guess, has no default values for placeholder, multiple and expanded options so they're missing in vars array.

Possible solutions:
1. I can rename option 'choices' to something different, but I still haven't found a better name :)
2. I can set default values for placeholder, multiple and expanded and assign them to vars inside buildView function in my DatalistType - that's my temporary workaround;
3. Maybe these recently added vars tests could be null-coalesced or `array_key_exists`-ed in ComponentWithFormTrait?

Thank you for your great work on the whole UX ecosystem!

maciazek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2487 
| License       | MIT

In my case, checking 'placeholder' for existence is sufficient. But I can imagine some very extreme custom form type which - besides 'choices' - has 'placeholder' option too. In that case, when 'placeholder' is set to false, then 'multiple' option need to be checked for existence, and so on... Considering the most pessimistic scenario, I think at least 'placeholder', 'multiple' and 'expanded' options should be checked.

I'm not sure about the rest - if there could be a situation where 'required', 'disabled' or 'value' option is nonexistent?

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -261,14 +261,13 @@ trait ComponentWithFormTrait
                 continue;
             }
 
-            if (\array_key_exists('choices', $child->vars)
+            if (\array_key_exists('choices', $child->vars) && $child->vars['choices']
                 && $child->vars['required']
                 && !$child->vars['disabled']
                 && !$child->vars['value']
-                && (false === $child->vars['placeholder'] || null === $child->vars['placeholder'])
-                && !$child->vars['multiple']
-                && !$child->vars['expanded']
-                && $child->vars['choices']
+                && \array_key_exists('placeholder', $child->vars) && (false === $child->vars['placeholder'] || null === $child->vars['placeholder'])
+                && \array_key_exists('multiple', $child->vars) && !$child->vars['multiple']
+                && \array_key_exists('expanded', $child->vars) && !$child->vars['expanded']
             ) {
                 if (null !== $firstKey = array_key_first($child->vars['choices'])) {
                     $values[$name] = $child->vars['choices'][$firstKey]->value ?? null;


smnandre:
Hey @maciazek during my reading of all this I found another bug here...

As I wanted to make some changes in your PR... I  ended by opening a new PR for both: https://github.com/symfony/ux/pull/2491 

If you want to have a look at it and confirm it does fix your problem? 

Thank you very much for opening this one!

maciazek:
Of course, I'll check it in my workplace tomorrow as I'm out of office today. Thanks!

smnandre:
I'm closing this one but I'd like to thank you _again_ for stepping up, explaining your case precisely and kindly, then opening a PR to fix the issue. 

🙇‍♂️

PR #2403, opened 2024-11-22
[LiveComponent] fix required select not initialized

dsoriano:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1332 
| License       | MIT

Initialize required select(ChoiceType or EntityType) with the first entry. Not applied to multiple or expanded.

PR #2491, opened 2025-01-08
[LiveComponent] Fix `ComponentWithFormTrait::extractFormValues()` with edge cases

### LINKED ISSUE #2487: [LiveComponent] Undefined array key "placeholder" after upgrading to 2.22.x

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #2487 
| License       | MIT

#2403 fixed an old bug to simulate browser behaviour when a select field has no option selected,  no placeholder, and is required (it then uses the first option)
#2425 and #2426 fixed the way we handled empty strings as placeholders

In #2487 @maciazek explained us how a custom type with a "choices" option became unusable since the last changes.

Also.. while I was reading all this I realized we returned wrong values here when "option groups" were used.. leading to other problems.

I updated the code of `extractFormValues()` method to:
* check if most of the caracteristic keys added in `ChoiceType::buildView` were defined in the `$view->vars`, to ensure we are dealing with a `<select>` field built by a `ChoiceType`
* fetch recursively the value of the first displayed option (even with groups)

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\LiveComponent;
 
+use Symfony\Component\Form\ChoiceList\View\ChoiceGroupView;
 use Symfony\Component\Form\ClearableErrorsInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
@@ -255,25 +256,39 @@ trait ComponentWithFormTrait
                 continue;
             }
 
+            // <input type="checkbox">
             if (\array_key_exists('checked', $child->vars)) {
-                // special handling for check boxes
                 $values[$name] = $child->vars['checked'] ? $child->vars['value'] : null;
                 continue;
             }
 
-            if (\array_key_exists('choices', $child->vars)
-                && $child->vars['required']
-                && !$child->vars['disabled']
-                && !$child->vars['value']
-                && (false === $child->vars['placeholder'] || null === $child->vars['placeholder'])
-                && !$child->vars['multiple']
-                && !$child->vars['expanded']
-                && $child->vars['choices']
+            // <select> - Simulate browser behavior
+            // When no option is selected, browsers send the value of the first
+            // option, when the following conditions are met:
+            if (
+                \array_key_exists('choices', $child->vars)
+                && $child->vars['choices']                  // has defined choices
+                && $child->vars['required']                 // is required
+                && !$child->vars['disabled']                // is not disabled
+                && '' === $child->vars['value']             // has no value set  ("0" can be a value)
+                && !array_diff_key(
+                    /* @see \Symfony\Component\Form\Extension\Core\Type\ChoiceType::buildView() */
+                    array_flip(['choices', 'expanded', 'multiple', 'placeholder', 'placeholder_in_choices', 'preferred_choices']),
+                    $child->vars,
+                )
+                && !$child->vars['expanded']                // is a <select>     (not a radio/checkbox)
+                && !$child->vars['multiple']                // is not multiple
+                && !\is_string($child->vars['placeholder'])  // has no placeholder (empty string is valid)
             ) {
-                if (null !== $firstKey = array_key_first($child->vars['choices'])) {
-                    $values[$name] = $child->vars['choices'][$firstKey]->value ?? null;
-                }
+                $choices = $child->vars['choices'];
+                do {
+                    $choice = $choices[array_key_first($choices)];
+                    if (!$choice instanceof ChoiceGroupView) {
+                        break;
+                    }
+                } while ($choices = $choice->choices);
 
+                $values[$name] = $choice?->value;
                 continue;
             }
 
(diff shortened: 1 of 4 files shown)


maciazek:
I can confirm that this patch resolved my specific case with custom form type and "choices" option. Thank you.

norkunas:
Something is more going on here.
I was getting `An exception has been thrown during the rendering of a template ("Warning: Undefined array key "expanded"").` and so I've tried this patch, then started to get:
`An exception has been thrown during the rendering of a template ("Warning: Undefined array key """).`

While #2490 works for me

maciazek:
> Something is more going on here. I was getting `An exception has been thrown during the rendering of a template ("Warning: Undefined array key "expanded"").` and so I've tried this patch, then started to get: `An exception has been thrown during the rendering of a template ("Warning: Undefined array key """).`
> 
> While #2490 works for me

I think it could happen when form has some ChoiceType field, which has 'choices' set as empty array. In that case
https://github.com/smnandre/ux/blob/cd353fc7a0689b9cee5da252f977acb420176d96/src/LiveComponent/src/ComponentWithFormTrait.php#L269
returns true, and then the error itself comes from:
https://github.com/smnandre/ux/blob/cd353fc7a0689b9cee5da252f977acb420176d96/src/LiveComponent/src/ComponentWithFormTrait.php#L284

Seems that some additional check is needed (if 'choices' is array and if it is not empty?)

smnandre:
Entirely my fault here.

As you can see here i checked for the "choices" existence. 
https://github.com/smnandre/ux/blob/cd353fc7a0689b9cee5da252f977acb420176d96/src/LiveComponent/src/ComponentWithFormTrait.php#L275C33-L275C34

I decided later on to "optimize" and moved the check first, adding an isset in the operation when the key was not defined.... and losing then my original goal: ensure there is one value.

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -255,25 +256,38 @@ private function extractFormValues(FormView $formView): array
...
-                && (false === $child->vars['placeholder'] || null === $child->vars['placeholder'])
-                && !$child->vars['multiple']
-                && !$child->vars['expanded']
-                && $child->vars['choices']
+            // <select> - Simulate browser behavior
+            // When no option is selected, browsers send the value of the first
+            // option, when the following conditions are met:
+            if (
+                isset($child->vars['choices'])              // has defined choices
+                && $child->vars['required']                 // is required
```
smnandre:
```suggestion
                \array_key_exists('choices', $child->vars)
                && $child->vars['choices']                  // has defined choices  
                && $child->vars['required']                 // is required
```

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -255,25 +256,39 @@ private function extractFormValues(FormView $formView): array
...
-                && (false === $child->vars['placeholder'] || null === $child->vars['placeholder'])
-                && !$child->vars['multiple']
-                && !$child->vars['expanded']
-                && $child->vars['choices']
+            // <select> - Simulate browser behavior
+            // When no option is selected, browsers send the value of the first
+            // option, when the following conditions are met:
+            if (
+                \array_key_exists('choices', $child->vars)
+                && $child->vars['choices']                  // has defined choices
```
smnandre:
```suggestion
                && $child->vars['choices']                  // has defined choices
```

smnandre:
This should work better now. Thank you very much to both of you.

Kocal (approved):

PR #2498, opened 2025-01-10
[TwigComponent] Add new ComponentReflection to extract properties from TwigComponent

Halleck45:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | 
| License       | MIT

Hello,

I'm currently trying to build a tool that analyzes `TwigComponent` to document (automatically) the components of a project. A bit like what [StoryBook](https://storybook.js.org/) does, but for Twig.

For this, I need to retrieve the properties of a component. The `TwigComponentDebugCommand` command already does this. This PR extracts the relevant code, and moves it to the new `ComponentPropertiesExtractor` class.

~~A BC is introduced: I need to inject the `ComponentPropertiesExtractor` class into the `TwigComponentDebugCommand`, but the last argument of the constructor is optional. This is normally not a problem: the class is internal~~
~~If necessary, I can keep backward compatibility by building the object directly in the code (rather than injecting it), but in the long run this may introduce testability issues.~~
Edit: fixed thanks to @Kocal suggestion

Thank you for your feedback!

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23.0
+
+- Add `ComponentPropertiesExtractor` to extract component properties from a Twig component
+
 ## 2.20.0
 
 -  Add Anonymous Component support for 3rd-party bundles #2019
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.23.0
 
-- Add `ComponentPropertiesExtractor` to extract component properties from a Twig component
+- Add `ComponentPropertiesExtractor` to extract component properties from a Twig component (experimental)
 
 ## 2.20.0
 
--- a/src/TwigComponent/src/Command/TwigComponentDebugCommand.php
+++ b/src/TwigComponent/src/Command/TwigComponentDebugCommand.php
@@ -21,10 +21,9 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Finder\Finder;
-use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
 use Symfony\UX\TwigComponent\ComponentFactory;
 use Symfony\UX\TwigComponent\ComponentMetadata;
-use Symfony\UX\TwigComponent\Twig\PropsNode;
+use Symfony\UX\TwigComponent\ComponentPropertiesExtractor;
 use Twig\Environment;
 use Twig\Loader\FilesystemLoader;
 
@@ -32,6 +31,7 @@ use Twig\Loader\FilesystemLoader;
 class TwigComponentDebugCommand extends Command
 {
     private readonly string $anonymousDirectory;
+    private readonly ComponentPropertiesExtractor $componentPropertiesExtractor;
 
     public function __construct(
         private string $twigTemplatesPath,
@@ -39,9 +39,11 @@ class TwigComponentDebugCommand extends Command
         private Environment $twig,
         private readonly array $componentClassMap,
         ?string $anonymousDirectory = null,
+        ?ComponentPropertiesExtractor $componentPropertiesExtractor = null,
     ) {
         parent::__construct();
         $this->anonymousDirectory = $anonymousDirectory ?? 'components';
+        $this->componentPropertiesExtractor = $componentPropertiesExtractor ?? new ComponentPropertiesExtractor($this->twig);
     }
 
     protected function configure(): void
@@ -212,12 +214,18 @@ EOF
             ['Template', $metadata->getTemplate()],
         ]);
 
+        $properties = $this->componentPropertiesExtractor->getComponentProperties($metadata);
+        $propertiesAsArrayOfStrings = array_filter(array_map(
+            fn (array $property) => $property['display'],
+            $properties,
+        ));
+
         // Anonymous Component
         if ($metadata->isAnonymous()) {
             $table->addRows([
                 ['Type', '<comment>Anonymous</comment>'],
                 new TableSeparator(),
-                ['Properties', implode("\n", $this->getAnonymousComponentProperties($metadata))],
+                ['Properties', implode("\n", $propertiesAsArrayOfStrings)],
             ]);
             $table->render();
 
@@ -229,7 +237,7 @@ EOF
             new TableSeparator(),
             // ['Attributes Var', $metadata->get('attributes_var')],
             ['Public Props', $metadata->isPublicPropsExposed() ? 'Yes' : 'No'],
-            ['Properties', implode("\n", $this->getComponentProperties($metadata))],
+            ['Properties', implode("\n", $propertiesAsArrayOfStrings)],
         ]);
 
         $logMethod = function (\ReflectionMethod $m) {
@@ -279,81 +287,5 @@ EOF
             ]);
         }
         $table->render();
-    }
-
-    /**
-     * @return array<string, string>
-     */
-    private function getComponentProperties(ComponentMetadata $metadata): array
-    {
-        $properties = [];
-        $reflectionClass = new \ReflectionClass($metadata->getClass());
-        foreach ($reflectionClass->getProperties() as $property) {
-            $propertyName = $property->getName();
-
-            if ($metadata->isPublicPropsExposed() && $property->isPublic()) {
-                $type = $property->getType();
-                if ($type instanceof \ReflectionNamedType) {
-                    $typeName = $type->getName();
-                } else {
-                    $typeName = (string) $type;
-                }
-                $value = $property->getDefaultValue();
-                $propertyDisplay = $typeName.' $'.$propertyName.(null !== $value ? ' = '.json_encode($value) : '');
-                $properties[$property->name] = $propertyDisplay;
-            }
-
-            foreach ($property->getAttributes(ExposeInTemplate::class) as $exposeAttribute) {
-                /** @var ExposeInTemplate $attribute */
-                $attribute = $exposeAttribute->newInstance();
-                $properties[$property->name] = $attribute->name ?? $property->name;
-            }
-        }
-
-        return $properties;
-    }
-
-    /**
-     * Extract properties from {% props %} tag in anonymous template.
-     *
-     * @return array<string, string>
-     */
-    private function getAnonymousComponentProperties(ComponentMetadata $metadata): array
-    {
-        $source = $this->twig->load($metadata->getTemplate())->getSourceContext();
-        $tokenStream = $this->twig->tokenize($source);
-        $moduleNode = $this->twig->parse($tokenStream);
-
-        $propsNode = null;
-        foreach ($moduleNode->getNode('body') as $bodyNode) {
-            foreach ($bodyNode as $node) {
-                if (PropsNode::class === $node::class) {
-                    $propsNode = $node;
-                    break 2;
-                }
-            }
-        }
-        if (!$propsNode instanceof PropsNode) {
-            return [];
-        }
-
-        $propertyNames = $propsNode->getAttribute('names');
-        $properties = array_combine($propertyNames, $propertyNames);
-        foreach ($propertyNames as $propName) {
-            if ($propsNode->hasNode($propName)
-                && ($valueNode = $propsNode->getNode($propName))
-                && $valueNode->hasAttribute('value')
-            ) {
-                $value = $valueNode->getAttribute('value');
-                if (\is_bool($value)) {
-                    $value = $value ? 'true' : 'false';
-                } else {
-                    $value = json_encode($value);
-                }
-                $properties[$propName] = $propName.' = '.$value;
-            }
-        }
-
-        return $properties;
     }
 }
--- a/src/TwigComponent/src/ComponentPropertiesExtractor.php
+++ b/src/TwigComponent/src/ComponentPropertiesExtractor.php
@@ -17,7 +17,7 @@ use Twig\Environment;
 
 /**
  * @author Jean-François Lépine <lepinejeanfrancois@gmail.com>
- * 
+ *
  * @experimental This class is not covered by the BC promise yet
  */
 final class ComponentPropertiesExtractor
@@ -28,17 +28,19 @@ final class ComponentPropertiesExtractor
     }
 
     /**
-     * Returns a list of properties from a Component. 
+     * Returns a list of properties from a Component.
      *
      * Warning: We do not recommend using this method at runtime, as it is rather slow.
+     *
+     * @return array<string, array{display: string, name: string, type: string, default: mixed}>
      */
     public function getComponentProperties(ComponentMetadata $metadata): array
     {
-        if ($medata->isAnonymous()) {
-            return $this->getAnonymousComponentProperties($medata);
+        if ($metadata->isAnonymous()) {
+            return $this->getAnonymousComponentProperties($metadata);
         }
 
-        return $this->getNonAnonymousComponentProperties($medata);
+        return $this->getNonAnonymousComponentProperties($metadata);
     }
 
     /**
--- a/src/TwigComponent/src/ComponentPropertiesExtractor.php
+++ b/src/TwigComponent/src/ComponentPropertiesExtractor.php
@@ -17,6 +17,8 @@ use Twig\Environment;
 
 /**
  * @author Jean-François Lépine <lepinejeanfrancois@gmail.com>
+ * 
+ * @experimental This class is not covered by the BC promise yet
  */
 final class ComponentPropertiesExtractor
 {
--- a/src/TwigComponent/src/ComponentPropertiesExtractor.php
+++ b/src/TwigComponent/src/ComponentPropertiesExtractor.php
@@ -15,6 +15,9 @@ use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
 use Symfony\UX\TwigComponent\Twig\PropsNode;
 use Twig\Environment;
 
+/**
+ * @author Jean-François Lépine <lepinejeanfrancois@gmail.com>
+ */
 final class ComponentPropertiesExtractor
 {
     public function __construct(
--- a/src/TwigComponent/src/ComponentPropertiesExtractor.php
+++ b/src/TwigComponent/src/ComponentPropertiesExtractor.php
@@ -23,7 +23,7 @@ final class ComponentPropertiesExtractor
     }
 
     /**
-     * @return array<string, string>
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
      */
     public function getComponentProperties(ComponentMetadata $medata)
     {
@@ -35,7 +35,7 @@ final class ComponentPropertiesExtractor
     }
 
     /**
-     * @return array<string, string>
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
      */
     private function getNonAnonymousComponentProperties(ComponentMetadata $metadata): array
     {
@@ -81,7 +81,7 @@ final class ComponentPropertiesExtractor
     /**
      * Extract properties from {% props %} tag in anonymous template.
      *
-     * @return array<string, string>
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
      */
     private function getAnonymousComponentProperties(ComponentMetadata $metadata): array
     {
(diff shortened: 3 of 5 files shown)


src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -37,6 +36,7 @@ public function __construct(
         private string $twigTemplatesPath,
         private ComponentFactory $componentFactory,
         private Environment $twig,
+        private readonly ComponentPropertiesExtractor $componentPropertiesExtractor,
```
Kocal: It looks like a BC since the class is neither marked as internal or final. 

If I'm not wrong, an ideal BC layer would looks like this:
```php
private readonly ComponentPropertiesExtractor $componentPropertiesExtractor 

public function __construct(
    private string $twigTemplatesPath,
    private ComponentFactory $componentFactory,
    private Environment $twig,
    private readonly array $componentClassMap,
    ?string $anonymousDirectory = null,
    ?ComponentPropertiesExtractor $componentPropertiesExtractor = null,
) {
    parent::__construct();
    $this->anonymousDirectory = $anonymousDirectory ?? 'components';
    $this->componentPropertiesExtractor = $componentPropertiesExtractor ?? new ComponentPropertiesExtractor($this->twig);
}
```

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -134,11 +135,17 @@ static function (ChildDefinition $definition, AsTwigComponent $attribute) {
...
+            ->setArguments([
+                new Reference('twig'),
+            ]);
+
         $container->register('ux.twig_component.command.debug', TwigComponentDebugCommand::class)
             ->setArguments([
                 new Parameter('twig.default_path'),
                 new Reference('ux.twig_component.component_factory'),
                 new Reference('twig'),
+                new Reference('ux.twig_component.extractor_properties'),
```
Kocal: To move after `$config['anonymous_template_directory']`

Kocal (commented):
Thanks for the feature, that's really interesting and could be really useful for some use cases, and we have new tests! :D  

Just a minor comment about the BC break.

cc @smnandre

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
Halleck45: Good idea, it's done and pushed.

Halleck45:
> Thanks for the feature, that's really interesting and could be really useful for some use cases, and we have new tests! :D
> 
> Just a minor comment about the BC break.

@Kocal I'm glad you like it! 😊 Thanks for the feedback. The BC break is now fixed and pushed!

smnandre:
Hi! I need to digg deeper in the code (later today), but already a big thank you for opening this and contributing! 

First reaction is: some code here is redondant with the ComponentProperties class.

I think we should differenciate here the extraction of props from templates, and props from classes. 

Making ComponentProperties call different sources (maybe more "Loaders" than "extractors" in the end).

--- 

And starting _after that_ building the ComponentMetadataFactory / ies we need.

src/TwigComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23.0
+
+- Add `ComponentPropertiesExtractor` to extract component properties from a Twig component
```
smnandre: I don't think we want to expose this class (with BC promise) for something like this. At least in a first step.

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+final class ComponentPropertiesExtractor
+{
+    public function __construct(
+        private readonly Environment $twig,
+    ) {
+    }
+
+    /**
+     * @return array<string, string>
+     */
```
smnandre: If we create a class for this, we will need to be a little more precise about the types

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,141 @@
...
+        if ($medata->isAnonymous()) {
+            return $this->getAnonymousComponentProperties($medata);
+        }
+
+        return $this->getNonAnonymousComponentProperties($medata);
+    }
+
+    /**
+     * @return array<string, string>
+     */
```
smnandre: Not the good type (you return an array of string => array{name: display: type: ...} for no)

smnandre:
I started this work here : https://github.com/symfony/ux/compare/2.x...smnandre:ux:dev/template-properties

src/TwigComponent/src/ComponentPropertiesExtractor.php
Halleck45: Thanks, that's fixed

src/TwigComponent/CHANGELOG.md
Halleck45: Hi, BC is resolved now. Do you still want me to remove this change?

smnandre: No, my point is i'm not sure we should make this class "public" for now. Because once it's done, we cannot go back. 

And, as i said in the previous message, we need to clarifiy first whose class has which responsibility, the contracts, etc.

WebMamba:
Hey there 👋 Thanks @Halleck45 for your PR! I am interested about your debugging tool! Why not using : https://packagist.org/packages/sensiolabs/storybook-bundle ? What your debugging tool gonna do ? 😁

WebMamba (commented):
I thinks that it's a good idea to move the logic to extract property information into a dedicated service, and we should do the same for the service to find the components. But I thinks all of that should not be separate from the core. I don't have any solution yet 😅

smnandre:
Any thought about my main points @Halleck45 ? (ComponentProperties / ComponentPropertyExtractor  and  the metadata thing)

Kocal (approved):

Halleck45:
> Any thought about my main points @Halleck45 ? (ComponentProperties / ComponentPropertyExtractor and the metadata thing)

I'm not sure, as I'm not very familiar with how `ComponentProperties` is used. I see it's called by the `ComponentRenderer`, and indeed, part of the logic is redundant, but I don't think it's used in the same way.

In the long term, having separate loaders does seem appropriate. However, I think that's a different topic to address later.

That said, we can lay the groundwork. With that in mind, I just committed some changes to introduce stronger typing and more consistent naming. I've introduced a `ComponentReflection` class with the methods `getProperties: ComponentReflectionProperty[]` and `getProperty(): ?ComponentReflectionProperty`, as well as a more structured object (`ComponentReflectionProperty`).

If this works for you, we can move forward with this, or I can revert it and go back to a simpler version.

smnandre:
Hmm that does not feel entirely right to me right now.. :| 

And i understand you need this for something. But we cannot introduce a class that, at least for a part, does what is already done by another, with cache and preloading instead of runtime.

So i'm very sorry for this, but yeah it's probably a safer choice in these conditions to go back at your original proposal. 

The problems of code duplication and maintaining this class afterwards being still here in both case, i'd suggest the following:
* move it into a Debug folder (or something like that in the name, you know, to clearly identify it) .... (and probably not include it when not APP_DEBUG=true, or do i miss something here?)
* but most importantly i'd set this class as experimental, allowing us to revert this change if we roll out a clean Metadata Factory/Loader before the end of the year.

I know this seems not much, but you cannot imagine the time i'm spending on debug/maintenance on tiny things like this, and I really would love to spend more of my time on new things :)

Halleck45:
> Hmm that does not feel entirely right to me right now.. :|
> 

No problem, I just reverted the last commit.

I'll keep [the branch handy anyway](https://github.com/symfony/ux/compare/2.x...Halleck45:ux:properties_extractor_reflection?expand=1), as I think it would be cleaner to have strong typing on what's being extracted.

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,144 @@
...
+{
+    public function __construct(
+        private readonly Environment $twig,
+    ) {
+    }
+
+    /**
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
+     */
+    public function getComponentProperties(ComponentMetadata $medata)
```
Kocal:
```suggestion
    public function getComponentProperties(ComponentMetadata $metadata): array
```

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,144 @@
...
+ */
+final class ComponentPropertiesExtractor
+{
+    public function __construct(
+        private readonly Environment $twig,
+    ) {
+    }
+
+    /**
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
```
Kocal: I recommend you to add a description, and also a warning like it has been done on `\Symfony\Component\Routing\RouterInterface::getRouteCollection()`: 

```suggestion
     * Returns a list of properties from a Component. 
     *
     * Warning: We do not recommend using this method at runtime, as it is rather slow.
```

src/TwigComponent/src/ComponentPropertiesExtractor.php
Kocal: The parameter name must be fixed, and return type added

src/TwigComponent/src/ComponentPropertiesExtractor.php
Kocal: We can be a little bit more precise by setting `string` as the array key:
```suggestion
     * @return array<string, array{display: string, name: string, type: string, default: mixed}>
```

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+        if ($medata->isAnonymous()) {
+            return $this->getAnonymousComponentProperties($medata);
+        }
+
+        return $this->getNonAnonymousComponentProperties($medata);
+    }
+
+    /**
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
```
Kocal:
```suggestion
     * @return array<string, array{display: string, name: string, type: string, default: mixed}>
```

Kocal (commented):
Some minor comments, and I think we will be ready to merge it :)

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,144 @@
...
+namespace Symfony\UX\TwigComponent;
+
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+use Symfony\UX\TwigComponent\Twig\PropsNode;
+use Twig\Environment;
+
+/**
+ * @author Jean-François Lépine <lepinejeanfrancois@gmail.com>
+ */
+final class ComponentPropertiesExtractor
```
smnandre:
```suggestion
/**
 * @author Jean-François Lépine <lepinejeanfrancois@gmail.com>
 * 
 * @experimental This class is not covered by the BC promise yet
 */
final class ComponentPropertiesExtractor
```

src/TwigComponent/src/ComponentPropertiesExtractor.php
smnandre: I don't think this service should be available when not in debug for now.

src/TwigComponent/CHANGELOG.md
```suggestion
- Add `ComponentPropertiesExtractor` to extract component properties from a Twig component (experimental)
```

Kocal (approved):

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+        private readonly Environment $twig,
+    ) {
+    }
+
+    /**
+     * Returns a list of properties from a Component.
+     *
+     * Warning: We do not recommend using this method at runtime, as it is rather slow.
+     *
+     * @return array<string, array{display: string, name: string, type: string, default: mixed}>
```
smnandre, line 35: Sorry i'm just realizing... we only use "display" in the debug command, right?

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+namespace Symfony\UX\TwigComponent;
+
+use Symfony\UX\TwigComponent\Attribute\ExposeInTemplate;
+use Symfony\UX\TwigComponent\Twig\PropsNode;
+use Twig\Environment;
+
+/**
+ * @author Jean-François Lépine <lepinejeanfrancois@gmail.com>
```
smnandre, line 19: As most of this class code comes from the original command code, this feels a bit over over-stating, wdyt ? :/

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+                $type = $property->getType();
+                if ($type instanceof \ReflectionNamedType) {
+                    $typeName = $type->getName();
+                } else {
+                    $typeName = (string) $type;
+                }
+                $value = $property->getDefaultValue();
+                $propertyDisplay = $typeName.' $'.$propertyName.(null !== $value ? ' = '.json_encode(
+                    $value
+                ) : '');
```
smnandre, line 66:
```suggestion
                $propertyDisplay = $typeName.' $'.$propertyName.(null !== $value ? ' = '.json_encode($value) : '');
```

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+
+        return $properties;
+    }
+
+    /**
+     * Extract properties from {% props %} tag in anonymous template.
+     *
+     * @return array<array{display: string, name: string, type: string, default: mixed}>
```
smnandre, line 93:
```suggestion
     * @return array<array{display: string, name: string, type: string, default: ?string}>
```

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+                }
+                $value = $property->getDefaultValue();
+                $propertyDisplay = $typeName.' $'.$propertyName.(null !== $value ? ' = '.json_encode(
+                    $value
+                ) : '');
+                $properties[$property->name] = [
+                    'name' => $propertyName,
+                    'display' => $propertyDisplay,
+                    'type' => $typeName,
+                    'default' => $value,
```
smnandre, line 71: Should be json encoded as the other one

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+    {
+        if ($metadata->isAnonymous()) {
+            return $this->getAnonymousComponentProperties($metadata);
+        }
+
+        return $this->getNonAnonymousComponentProperties($metadata);
+    }
+
+    /**
+     * @return array<string, array{display: string, name: string, type: string, default: mixed}>
```
smnandre, line 47:
```suggestion
     * @return array<string, array{display: string, name: string, type: string, default: ?string}>
```

(once fixed the value json encoding)

src/TwigComponent/src/ComponentPropertiesExtractor.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    private function getNonAnonymousComponentProperties(ComponentMetadata $metadata): array
```
smnandre, line 49: Let's avoid double negation if possible. Anonymous component are a special type of components. 

But "non anonymous components" are just "components".

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -21,27 +21,29 @@
...
         private string $twigTemplatesPath,
         private ComponentFactory $componentFactory,
         private Environment $twig,
         private readonly array $componentClassMap,
         ?string $anonymousDirectory = null,
+        ?ComponentPropertiesExtractor $componentPropertiesExtractor = null,
     ) {
         parent::__construct();
         $this->anonymousDirectory = $anonymousDirectory ?? 'components';
+        $this->componentPropertiesExtractor = $componentPropertiesExtractor ?? new ComponentPropertiesExtractor($this->twig);
```
smnandre, line 46: If we can create it like this, why injecting it ? 

If we need to inject it, then we should trigger a deprecation if not provided here

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -212,12 +214,18 @@ private function displayComponentDetails(SymfonyStyle $io, string $name): void
             ['Template', $metadata->getTemplate()],
         ]);
 
+        $properties = $this->componentPropertiesExtractor->getComponentProperties($metadata);
+        $propertiesAsArrayOfStrings = array_filter(array_map(
+            fn (array $property) => $property['display'],
+            $properties,
+        ));
```
smnandre, line 221:
```suggestion
        $properties = array_column($properties, 'display');
```

1/3

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -212,12 +214,18 @@ private function displayComponentDetails(SymfonyStyle $io, string $name): void
...
+
         // Anonymous Component
         if ($metadata->isAnonymous()) {
             $table->addRows([
                 ['Type', '<comment>Anonymous</comment>'],
                 new TableSeparator(),
-                ['Properties', implode("\n", $this->getAnonymousComponentProperties($metadata))],
+                ['Properties', implode("\n", $propertiesAsArrayOfStrings)],
```
smnandre, line 228:
```suggestion
                ['Properties', implode("\n", $properties)],
```

2/3

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -229,7 +237,7 @@ private function displayComponentDetails(SymfonyStyle $io, string $name): void
             new TableSeparator(),
             // ['Attributes Var', $metadata->get('attributes_var')],
             ['Public Props', $metadata->isPublicPropsExposed() ? 'Yes' : 'No'],
-            ['Properties', implode("\n", $this->getComponentProperties($metadata))],
+            ['Properties', implode("\n", $propertiesAsArrayOfStrings)],
```
smnandre, line 240:
```suggestion
            ['Properties', implode("\n", $properties)],
```

3/3

smnandre:
(last comments and LGTM ;) )

src/TwigComponent/src/ComponentPropertiesExtractor.php
WebMamba: I do not agree here since Jean-Francois is the creator of the file, he should put his name on the file

Kocal: image

smnandre: The code had not been authored here, do you agree ? Or i can create HttpResponse2.php and put my name on it ? Be serious @WebMamba @Kocal  ... 

... did you really took the time to understand what is new / what is not here ?

Halleck45: Hi, 

Sorry! Indeed, I moved some code, and the header should mention both contributors. I didn't think of it. I'll address this along with the other feedback, if that's okay with you.

I can also remove my name; it's not particularly necessary. I'm just trying to follow what I noticed in the surrounding files. Let me know what you prefer, and I'll follow your guidance.

smnandre: Disagreements are rare around here, but always sincere. After discussing with my colleagues, I align with the general opinion. Thank you for the suggestion -not necessary after all- but appreciated!

WebMamba:
Hey @Halleck45! Thanks for your interest, but I am not sure we gonna merge this PR. The reason is that I see the value of exposing such classes to your project, but I think there is less value for the Symfony UX project. We don't have any plan to use this code in another place than in the debug command; this is why I think we don't need to move the code to a dedicated class. Exposing such classes has cost (maintenance, BC, ...), so I don't think it gonna be valuable for us to have this class. Sorry about that, don't take it personally, we really appreciate having you here. @smnandre @Kocal WDYT ?

Kocal:
I think it's fine to make it public but with the `@experimental` annotation, so we are free to not respect the BC promise, but the users must use it carefully.

We can also tag it as internal, with a warning message `This class is not covered by the Symfony backward compatibility promise.` like PHPUnit does.

Halleck45:
No worries about closing this PR. I think I can move forward in another way. Thanks for your feedback anyway!

@WebMamba, I couldn't get storybook-bundle to work, but it makes me want to dig deeper.

See you soon!

smnandre:
(closing there, as discussed, and let's focus on the other one ❤️ )

Issue #2004, opened 2024-07-23
Deprecate the Lazy Image package


javiereguiluz:
As discussed in #1967, the solution proposed by the Lazy Image package (based on the "blurhash" library) was useful in the past. However, nowadays all browsers support many native features to improve the loading of images. See https://web.dev/learn/images/performance-issues

So, maybe we could deprecate this package to reduce the maintenance work? Thanks!

smnandre:
What is the "usual" scenario there?
- deprecate the component (what about the doc ? should we explain other solutions?)
- abandon it ? (after how much time?)

This is very new to me ^^

kbond:
I agree, my plan (whose timeline is unknown at this point), was to create a ["recipe" system](https://github.com/symfony/ux/pull/1347) on ux.symfony.com first, then deprecate the ux components that could easily be replaced by a recipe (once the recipe exists). This allow us to point users to a concrete replacement.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly reminder that this issue exists. If I don't hear anything I'll close this.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

Kocal:
The package has been deprecated since

PR #2501, opened 2025-01-10
[Doc] Add a warning message to not recommend Lazy Image package

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | 
| License       | MIT

Related to #2004. If we don't deprecate the package, at least warn users that it's no longer recommended. Thanks.

### DIFF
--- a/src/LazyImage/doc/index.rst
+++ b/src/LazyImage/doc/index.rst
@@ -1,6 +1,12 @@
 Symfony UX LazyImage
 ====================
 
+.. warning::
+
+    **This package is no longer recommended.** Instead, use the
+    `modern techniques to improve image loading performance`_ natively supported
+    by all major browsers.
+
 Symfony UX LazyImage is a Symfony bundle providing utilities to improve
 image loading performance. It is part of `the Symfony UX initiative`_.
 
@@ -245,6 +251,7 @@ This bundle aims at following the same Backward Compatibility promise as
 the Symfony framework:
 https://symfony.com/doc/current/contributing/code/bc.html
 
+.. _`modern techniques to improve image loading performance`: https://web.dev/learn/images/performance-issues
 .. _`the Symfony UX initiative`: https://ux.symfony.com/
 .. _`BlurHash implementation`: https://blurha.sh
 .. _`StimulusBundle`: https://symfony.com/bundles/StimulusBundle/current/index.html


Kocal (commented):
Thanks Javier for the suggestion.

I'm a bit conflicted, because for me, modern techniques for improving image loading and BlurHash aren't really the same thing:
- On the one hand, we prefer to play with early hints, lazy-loading, loading priorities, in order to **improve performances** (and CrUX)
- but on the other hand, it **is possible** to use BlurHash to **improve the user experience** by displaying a (blurred) image instead of nothing at all.

I already documented how we can use UX LazyImage for the LCP image in #1967 and I still believe it is enough. Maybe we should cover other topics? 

WDYT?

Thanks!

javiereguiluz:
You are right. But at the same time, a few years ago it was common to see these image blurs while they were loading. Today, I can't remember the last time I saw one.

Maybe this is my own bubble so I'm asking: do you know any big and/or popular site that still uses this technique? Thanks.

Kocal:
> Maybe this is my own bubble so I'm asking: do you know any big and/or popular site that still uses this technique? Thanks.

...  not at all! 😅 

Apparently, Mastodon used to use it (https://github.com/woltapp/blurhash#users), but I don't see where. 

Anyway, to mee BlurHash was always more a "smartphone" thing than a "web" thing. 

At least, maybe we can reword your suggestion by telling that using UX LazyImage is not recommended for performances but only for user experience?

smnandre:
I'm for deprecating the package. 

Starting from the documentation here:

> It provides two key features:
>
>    a Stimulus controller to load lazily heavy images, with a placeholder
>   a [BlurHash implementation](https://blurha.sh/) to create data-uri thumbnails for images


## Stimulus controller 

This part is obselete today. 

The source HTML we suggest is the following

```html
<img
        src="[data:image/jpeg;base64,/9j/4AAQSkZJ[...]"
        data-controller="symfony--ux-lazy-image--lazy-image"
        data-symfony--ux-lazy-image--lazy-image-src-value="/assets/images/legos-FTSkITj.jpg"
        alt="A blurhash image that is replaced by the real image on load"
        width="578"
        height="275"
        style="border-radius: 5px;"
    >
```

What was a good idea years back is now...  less than good. On many levels (SEO, accessibility) but it's above all very counter-productive, working against the browser and its asset priority algorithms, forcing it to render one image then a second... without using all the srcset / dimension / formats optimizations. 

## Blurhash

I can understand the visual impact on specific use cases. But this is not about loading then, it's about visual effect, style, mood.  

Technically, a CSS/SVG radial gradient with a blur will give you the same result, without the need for complex image computations, cache, heavy pages. It requires 2 CSS vars per "card" (worst case scenario).

And no activity for a while on the most popular blurhash repository: https://github.com/woltapp/blurhash .. probably a good sign to leave it to the past :) 


Personal conclusion: it's **more than time** to deprecate the package (or at least publish a warning on it a suggested by @javiereguiluz  here 👍 )

--- 

If only we had more hands.... 🙏   There is so much we could do about images, CSS, SVG...  to have a major impact on performances, and UX...

smnandre:
Well, this is right on time : https://developer.mozilla.org/en-US/blog/fix-image-lcp/#resource_load_delay

LGTM, but can we discuss then something a bit more ... radical ?

smnandre (approved):

Kocal:
Thank you @javiereguiluz.

PR #2503, opened 2025-01-13
[CI] Fix and display changed dist files in the CI

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

When working on https://github.com/symfony/stimulus-bridge, I've added a new job CI that shows when dist files are up-to-date or need to be build.

No surprise, I re-used what was on UX, but I found it was not correct: https://github.com/symfony/stimulus-bridge/actions/runs/12743300005/job/35512968204?pr=96
image

The command that checks if files are up-to-date fails, to me it is not working as expected.

Instead, I've simplified things and added a `git diff` to display the changes: 
image

WDYT?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,21 +47,17 @@ jobs:
                   restore-keys: |
                       ${{ runner.os }}-yarn-
             - run: yarn --immutable && yarn build
-            - name: Check if js dist files are current
-              id: changes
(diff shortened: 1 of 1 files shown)


kbond (approved):
Nice!

smnandre (approved):

PR #1772, opened 2024-04-23
[TwigComponent] Fix LiveComponent namespace mapping

smnandre:
I believe this modification fix #1754

**Original problem**

If you render a template @Acme/dashboard.html.twig, embedded Components (live or twig) referenced wrong template (dashboard.html.twig here), making LiveComponent impossible to rerender (this file beeing not referenced in the TemplateMap).

This behaviour was introduced in #1082 (and worked then). But i'm convinced all the work/changes made by Ryan to fix the embed bug (#1247 and related) made this obselete (and in this particular situation even originates a bug)

All tests are green, and i'd like to test it more heavily... but before that : do you think this is a bug fix or is there some kind of BC break here?

PR #2506, opened 2025-01-16
[TwigComponent] Remove obsolete TemplateNameParser

sneakyvv:
It no longer parses (and should not have been stripping namespaces as it did before)

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | 
| License       | MIT

As discussed in https://github.com/symfony/ux/pull/1772#issuecomment-2595130307, the code added in #1070 (& expanded in #1082) was not needed (even incorrect, based upon a decorator TemplateIterator of Shopware).


smnandre (approved):
As discussed on Slack, i'm very happy to not have to digg again in here :) 

And i'm glad we did not do something that created BC.

So everything perfect :)

smnandre:
Thank you @sneakyvv.

PR #2507, opened 2025-01-17
[TwigComponent][Docs] add explicit string cast on string assertions

Valmonzo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        |  
| License       | MIT

Following the examples from the documentation, I realized that if we use the `TestLiveComponent::render` method in `assertStringContainsString` and if we use strict types, then it throw an error because we try to assert a `RenderedComponent` object instead of a string. 

I think that explicitly adding the cast can help.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -1342,7 +1342,7 @@ You can test how your component is mounted and rendered using the
                 data: ['foo' => 'bar'],
             );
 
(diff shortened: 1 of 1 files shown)


Kocal (approved):

Kocal:
Indeed, it's a good documentation improvement that will help with silly mistakes. Thanks @Valmonzo

PR #2510, opened 2025-01-19
[LiveComponent] Update doc on how to use submitForm

### LINKED ISSUE #2509: Testing LiveComponent submitForm doc improvement
hi, 

thanks a lot for the library and especially the TestHelper. However, I found myself in a weird situation when wanting to use the `InteractsWithLiveComponents` trait as mentioned in the [documentation of the LiveComponents](https://symfony.com/bundles/ux-live-component/current/index.html#test-helper) as the following is mentioned in it to help test the submission of the form:

```php
// Submit form data
$testComponent
    ->submitForm(['form' => ['input' => 'value']], 'save');
```

I tried copy-pasting it but I got an error "`Form Validation error` even if I was absolutely certain that my form values were correct. After some time of trying it out, of searching (I ventured around a csrf issue in test which was not the case), the error was that I was not using the correct form name (my form isn't using a `FormType` but an `EventCreationType` and so the code needed is the following:

```php
// Submit form data
$testComponent
    ->submitForm(['event_creation' => [... correct values for my form]], 'save');
```

Instead, the form was submitted with the default formValues of my component which were indeed not correct ;-)

Maybe the documentation could be updated to include in the comment above the code example what 'form' links to? I'll propose these changes in a pull request

yalit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #2509  
| License       | MIT

As described in the issue, I struggled a bit with the usage of the submitForm helper for testing a LiveComponent due to the name of the form used not being correct see #2509.

I propose a small update to the documentation to make it more clear how the formName should be used into the call of submitForm

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3732,9 +3732,9 @@ uses Symfony's test client to render and make requests to your components::
                 ->set('count', 99)
             ;
 
-            // Submit form data
+            // Submit form data ('my_form' for your MyFormType form)
             $testComponent
-                ->submitForm(['form' => ['input' => 'value']], 'save');
+                ->submitForm(['my_form' => ['input' => 'value']], 'save');
 
             $this->assertStringContainsString('Count: 99', $testComponent->render());
 


src/LiveComponent/doc/index.rst
```diff
@@ -3733,6 +3733,8 @@ uses Symfony's test client to render and make requests to your components::
             ;
 
             // Submit form data
+	    // 'form' is the name of the form you're submitting (here FormType). 
+            // e.a : form submitted is EntityCreationType the name of your form would be 'entity_creation' and should replace 'form' in the below submitForm call
```
smnandre: I'd suggest to remove the text and use another word than "form" in the exemple..

Something like "create_comment"  or "my_form" .

wdyt ?

yalit: indeed, I didn't want to update the "core" of the documentation but if it's a possibility, we could indeed do something like:

```php
// Submit form data ('my_form' for your MyFormType form)
$testComponent
    ->submitForm(['my_form' => ['input' => 'value']], 'save');
```

```suggestion
            // 'form' is the name of the form you're submitting (here FormType). 
            // e.a : form submitted is EntityCreationType the name of your form would be 'entity_creation' and should replace 'form' in the below submitForm call
```

yalit: I pushed that modification

yalit: I don't know why i was doing that but when looking at the actual render it was properly aligned... Anyhow, a new commit has been made with a simplification of the comments

Kocal (approved):

smnandre (approved):

Kocal:
Thank you @yalit.

yalit:
Thanks for the approval and merge!

PR #2519, opened 2025-01-22
Update mercure doc configuration

### LINKED ISSUE #2518: Turbo & Mercure documentation does not work
The Turbo / Mercure documentation is inconsistent: 

The following (can be found https://github.com/symfony/ux/blob/2.x/src/Turbo/doc/index.rst?plain=1#L519-L522) does generate an error:
```yaml
# config/packages/turbo.yaml
turbo:
    mercure:
        hubs: [default]
```

> Unrecognized option "mercure" under "turbo". Available options are "broadcast", "default_transport".

I don't get how this can work, as there is no code referring to this option in the configuration: https://github.com/symfony/ux/blob/2.x/src/Turbo/src/DependencyInjection/Configuration.php

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2518 
| License       | MIT

The turbo-mercure-bundle does not exist anymore, so mercure configuration is enough to have mercure and turbo working together

### DIFF
--- a/src/Turbo/doc/index.rst
+++ b/src/Turbo/doc/index.rst
@@ -516,10 +516,14 @@ Otherwise, configure Mercure Hub(s) to use:
 
 .. code-block:: yaml
 
-    # config/packages/turbo.yaml
-    turbo:
-        mercure:
-            hubs: [default]
+    # config/packages/mercure.yaml
+    mercure:
+        default:
+            url: '%env(MERCURE_URL)%'
+            public_url: '%env(MERCURE_PUBLIC_URL)%'
+            jwt:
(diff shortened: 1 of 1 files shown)


Kocal (approved):

smnandre:
There is another occurence of this later ... could you fix it too ? 🤗 

```
turbo:
     mercure:
```

dunglas (approved):

smnandre:
Closing in favor of #2524 that uses real values a user would see, removes the second bloc to remove in the same pages, and fixes minors things passing by?

PR #2523, opened 2025-01-23
[React] Mention Webpack Encoure bundle should be installed first

dukeofgaming:
If not working with AssetMapper.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

This will help newcomers not have to remove the UX React bundle if they wanted it to modify what is created through the Webpack Encore bundle.

### DIFF
--- a/src/React/doc/index.rst
+++ b/src/React/doc/index.rst
@@ -14,6 +14,7 @@ Symfony UX React supports React 18+.
 
 Installation
 ------------
+Make sure you have WebpackEncore if you are not working with AssetMapper. You can install and configure Webpack Encore through the `Symfony Encore bundle`_. This will modify your ``base.html.twig`` template and create ``assets/app.js``, in addition to adding and modifying other configuration, which will in turn help you get started faster with the UX React bundle.
 
 .. note::
 
@@ -156,3 +157,4 @@ https://symfony.com/doc/current/contributing/code/bc.html
 .. _`Symfony UX initiative`: https://ux.symfony.com/
 .. _`Symfony UX React demo`: https://ux.symfony.com/react
 .. _`Turbo`: https://turbo.hotwire.dev/
+.. _`Symfony Encore bundle`: https://symfony.com/doc/current/frontend/encore/installation.html


src/React/doc/index.rst
```diff
@@ -14,6 +14,7 @@ Symfony UX React supports React 18+.
 
 Installation
 ------------
+Make sure you have WebpackEncore if you are not working with AssetMapper. You can install and configure Webpack Encore through the `Symfony Encore bundle`_. This will modify your ``base.html.twig`` template and create ``assets/app.js``, in addition to adding and modifying other configuration, which will in turn help you get started faster with the UX React bundle.
```
smnandre, line 17: There is a note just below, stating "works best with Webpack Encore. If AssetMapper... "

I'd rather keep it on top. I think we should add a simple sentence  here, with a link to either the WebpackEncore documentation, or a section lower in the page. 


Something like

```rst
. note::

    This package works best with WebpackEncore (see  :ref:`Install and configure WebpackEncore`_)
    To use it with AssetMapper, see  :ref:`Using with AssetMapper <using-with-asset-mapper>`. 
```     

That would make this visible, while keeping the top of the page focused on `ux/react`

Kocal: Totally, a simple sentence is enough, no need to duplicate what's written on Webpack Encore setup page.

smnandre: @dukeofgaming still motivated on releasing this doc update?

dukeofgaming: @smnandre will try to get back to this by the end of the week. I recall this causing me some unnecessary pain (had to revert a commit on installing the UX react bundle). just because I had not configured/installed webpack before following the docs on the installation of the UX React bundle

dukeofgaming: @smnandre totally ok with your change :)

Kocal:

Kocal:

Thanks!

Status: Needs Work

PR #2526, opened 2025-01-25
Debugging Components:  Info for PHP

Uplink03:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Issues        | Fix #... 
| License       | MIT

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3715,6 +3715,11 @@ Debugging Components
 Need to list or debug some component issues.
 The `Twig Component debug command`_ can help you.
 
+To debug your PHP code, you can inject a `private LoggerInterface $logger` in your `__construct`.
+Then you can use `$this->logger->debug('Message', ['var' => $var]);` and see the output in `var/log/dev.log`.
+`dump` won't display anything anywhere (not even the debug toolbar).
+`dd` will display in your browser, but will also deadlock your component - you cannot dismiss it and continue working.
+
 Test Helper
 -----------
 


src/LiveComponent/doc/index.rst
```diff
@@ -3715,6 +3715,11 @@ Debugging Components
 Need to list or debug some component issues.
 The `Twig Component debug command`_ can help you.
 
+To debug your PHP code, you can inject a `private LoggerInterface $logger` in your `__construct`.
+Then you can use `$this->logger->debug('Message', ['var' => $var]);` and see the output in `var/log/dev.log`.
+`dump` won't display anything anywhere (not even the debug toolbar).
+`dd` will display in your browser, but will also deadlock your component - you cannot dismiss it and continue working.
```
smnandre, line 3721: This is the same thing in a webpage. `dd` means `dump and die`.

src/LiveComponent/doc/index.rst
```diff
@@ -3715,6 +3715,11 @@ Debugging Components
 Need to list or debug some component issues.
 The `Twig Component debug command`_ can help you.
 
+To debug your PHP code, you can inject a `private LoggerInterface $logger` in your `__construct`.
+Then you can use `$this->logger->debug('Message', ['var' => $var]);` and see the output in `var/log/dev.log`.
+`dump` won't display anything anywhere (not even the debug toolbar).
```
smnandre, line 3720: This is not true. 

During the first render, dumps are shown either in the template or in the debug toolbar.

On following request they are shown on the profiler, like any AJAX request.

src/LiveComponent/doc/index.rst
```diff
@@ -3715,6 +3715,11 @@ Debugging Components
 Need to list or debug some component issues.
 The `Twig Component debug command`_ can help you.
 
+To debug your PHP code, you can inject a `private LoggerInterface $logger` in your `__construct`.
+Then you can use `$this->logger->debug('Message', ['var' => $var]);` and see the output in `var/log/dev.log`.
```
smnandre, line 3719: I'm not sure this is really needed, as it applies to any component / service in Symfony apps (and you can inject a Logger with a lot of other methods,) . 

We could send the user towards the documentation pages on logging, but then it would not be really about _debugging_.

Uplink03:
This was meant as a quick mention of deficiencies when debugging the component in the browser. I didn't put much thought into it, and I'm not really sure how I phrase it better. I'll withdraw the PR and come back if I have a better one.

smnandre:
Thank you for opening the PR and making this effort!  👍

PR #2537, opened 2025-01-30
[LiveComponent] Prevent `__component` property to be serialized when called `JSON.stringify()`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Following https://symfony-devs.slack.com/archives/C01FN4EQNLX/p1738238525455439

With some RUM tools that listen user interactions, some of them can `JSON.stringify()` their payload. But if their payload contains an HTML element with the property `__component` (which represents a `Proxy` to the `LiveComponent` instance), then it tries to be serialized and lead to an error by trying to call an action `toJSON` and trigger an XHR request to `/toJSON`: 
image

Let's change the property assignment with `Object.defineProperty()`, so the property is not enumarable and so not serialized when `JSON.stringify()` is called:
image

cc @jwage

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.23.0
 
 -   Allow configuring the secret used to compute fingerprints and checksums.
+-   Prevent `__component` property to be serialized when called `JSON.stringify()`
 
 ## 2.22.0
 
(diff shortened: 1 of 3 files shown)


jwage (approved):

smnandre:
Thank you @jwage  for the time spent to assist / explain what happened.

@Kocal  🙇  you'be been a legend today

smnandre (approved):

kbond (approved):
Wow, that was a deep dive!

Kocal:
(Added missing changelog)

Issue #2534, opened 2025-01-29
[Autocomplete] How to customize the rendering of items?


daFish:
Hi there,

I want to customize the rendering of items. Currently, I have a custom stimulus controller which listens to the event `autocomplete:pre-connect` to override the option set in https://github.com/symfony/ux/blob/2.x/src/Autocomplete/assets/src/controller.ts#L243C17-L243C73

What other ways could be there?

Kocal:
Hi, looks like listening on `autocomplete:pre-connect` event is the best (and only) option here

daFish:
Thank you, @Kocal. To make my controller more generic I'll work with values here. Let's see how this works out.

PR #2541, opened 2025-01-31
[Autocomplete] Configurable results

J-Ben87:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | 
| License       | MIT

TomSelect accepts by default an array of results, each item containing a "text" and a "value".
Therefore this bundle exposes results formatted like so.

However one may need to expose additional data such as a "disabled" state to customize TomSelect's rendering.

This PR aims to offer this possibility by letting `EntityAutocompleter`s drive the creation of the results array.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 -   Deprecate `ExtraLazyChoiceLoader` in favor of `Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader`
 -   Reset TomSelect when updating url attribute #1505
+-   Add `getAttributes()` method to define additional attributes for autocomplete results #2541
 
 ## 2.22.0
 
--- a/src/Autocomplete/src/AutocompleteResultsExecutor.php
+++ b/src/Autocomplete/src/AutocompleteResultsExecutor.php
@@ -73,10 +73,7 @@ final class AutocompleteResultsExecutor
 
         if (!method_exists($autocompleter, 'getGroupBy') || null === $groupBy = $autocompleter->getGroupBy()) {
             foreach ($paginator as $entity) {
-                $results[] = [
-                    'value' => $autocompleter->getValue($entity),
-                    'text' => $autocompleter->getLabel($entity),
-                ];
+                $results[] = $this->formatResult($autocompleter, $entity);
             }
 
             return new AutocompleteResults($results, $hasNextPage);
@@ -104,10 +101,7 @@ final class AutocompleteResultsExecutor
         $optgroupLabels = [];
 
         foreach ($paginator as $entity) {
-            $result = [
-                'value' => $autocompleter->getValue($entity),
-                'text' => $autocompleter->getLabel($entity),
-            ];
+            $result = $this->formatResult($autocompleter, $entity);
 
             $groupLabels = $groupBy($entity, $result['value'], $result['text']);
 
@@ -123,5 +117,22 @@ final class AutocompleteResultsExecutor
         $optgroups = array_map(fn (string $label) => ['value' => $label, 'label' => $label], array_unique($optgroupLabels));
 
         return new AutocompleteResults($results, $hasNextPage, $optgroups);
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    private function formatResult(EntityAutocompleterInterface $autocompleter, object $entity): array
+    {
+        $attributes = [];
+        if (method_exists($autocompleter, 'getAttributes')) {
+            $attributes = $autocompleter->getAttributes($entity);
+        }
+
+        return [
+            ...$attributes,
+            'value' => $autocompleter->getValue($entity),
+            'text' => $autocompleter->getLabel($entity),
+        ];
     }
 }
--- a/src/Autocomplete/src/EntityAutocompleterInterface.php
+++ b/src/Autocomplete/src/EntityAutocompleterInterface.php
@@ -18,29 +18,49 @@ use Symfony\Bundle\SecurityBundle\Security;
 /**
  * Interface for classes that will have an "autocomplete" endpoint exposed.
  *
- * @method mixed getGroupBy() Return group_by option.
+ * @template T of object
+ *
+ * TODO Remove next lines for Symfony UX 3
+ *
+ * @method array getAttributes(object $entity) Returns extra attributes to add to the autocomplete result.
+ * @method mixed getGroupBy()                  Return group_by option.
  */
 interface EntityAutocompleterInterface
 {
     /**
      * The fully-qualified entity class this will be autocompleting.
+     *
+     * @return class-string<T>
      */
     public function getEntityClass(): string;
 
     /**
      * Create a query builder that filters for the given "query".
+     *
+     * @param EntityRepository<T> $repository
      */
     public function createFilteredQueryBuilder(EntityRepository $repository, string $query): QueryBuilder;
 
     /**
      * Returns the "choice_label" used to display this entity.
+     *
+     * @param T $entity
      */
     public function getLabel(object $entity): string;
 
     /**
      * Returns the "value" attribute for this entity, usually the id.
+     *
+     * @param T $entity
      */
     public function getValue(object $entity): mixed;
+
+    /**
+     * Returns extra attributes to add to the autocomplete result.
+     *
+     * TODO Uncomment for Symfony UX 3
+     */
+    /* public function getAttributes(object $entity): array; */
 
     /**
      * Return true if access should be granted to the autocomplete results for the current user.
@@ -51,6 +71,8 @@ interface EntityAutocompleterInterface
 
     /*
      * Return group_by option.
+     *
+     * TODO Uncomment for Symfony UX 3
      */
     /* public function getGroupBy(): mixed; */
 }
--- a/src/Autocomplete/tests/Fixtures/Autocompleter/CustomAttributesProductAutocompleter.php
+++ b/src/Autocomplete/tests/Fixtures/Autocompleter/CustomAttributesProductAutocompleter.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\UX\Autocomplete\Tests\Fixtures\Autocompleter;
+
+class CustomAttributesProductAutocompleter extends CustomProductAutocompleter
+{
+    public function getAttributes(object $entity): array
+    {
+        return [
+            'disabled' => true,
+            'value' => 'This value should be replaced with the result of getValue()',
+            'text' => 'This value should be replaced with the result of getText()',
+        ];
+    }
+}
--- a/src/Autocomplete/tests/Fixtures/Autocompleter/CustomProductAutocompleter.php
+++ b/src/Autocomplete/tests/Fixtures/Autocompleter/CustomProductAutocompleter.php
@@ -19,6 +19,9 @@ use Symfony\UX\Autocomplete\Doctrine\EntitySearchUtil;
 use Symfony\UX\Autocomplete\EntityAutocompleterInterface;
 use Symfony\UX\Autocomplete\Tests\Fixtures\Entity\Product;
 
+/**
+ * @implements EntityAutocompleterInterface<Product>
+ */
 class CustomProductAutocompleter implements EntityAutocompleterInterface
 {
     public function __construct(
@@ -56,6 +59,11 @@ class CustomProductAutocompleter implements EntityAutocompleterInterface
     public function getValue(object $entity): mixed
     {
         return $entity->getId();
+    }
+
+    public function getAttributes(object $entity): array
+    {
+        return [];
     }
 
     public function isGranted(Security $security): bool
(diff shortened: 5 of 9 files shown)


J-Ben87:
I'm not sure why I am getting failures related to "direct deprecation notices" on [LiveComponent] (see https://github.com/symfony/ux/actions/runs/13069237861/job/36467092283?pr=2541 and https://github.com/symfony/ux/actions/runs/13069237861/job/36467096825?pr=2541) when I didn't update this component nor anything related to `zenstruck/foundry` 🤔

daFish:
Hey @J-Ben87, this is awesome. This will be very useful in the context of #2534.

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -18,30 +18,48 @@
...
      */
     public function getValue(object $entity): mixed;
 
+    /**
+     * Returns the autocomplete result, usually the label as "text" and the value as "value".
+     * May be used to expose extra options such as "disabled" so that tom-select's rendering may be customized.
+     *
+     * @param T $entity
+     */
+    public function getResult(object $entity): array;
```
Kocal: Adding a public method to a non-internal interface is a BC break. 
When upgrading to a minor version, users implementing interface will see their code break because they do not implement this method `getResult()`.

Kocal (changes requested):

src/Autocomplete/src/EntityAutocompleterInterface.php
smnandre: But you can set this method in comment for future version and add it into the concrete class.

But.. i feel this method a bit odd.. the result is the object. So maybe more getAttributes or getItemAttr (something that would be 100% opt-in, like what's done in the form) ?

J-Ben87: Oops, you're 100% right! I'm not used to Open Source development 😅 
I fixed it the way @smnandre suggested (just like the `getGroupBy()` method actually).

J-Ben87: @smnandre Actually the result is not the object: it's an array like:
```
$results[] = [
    'value' => $autocompleter->getValue($entity),
    'text' => $autocompleter->getLabel($entity),
];
```

I named the method after the `$results` variable that was used in the `AutocompleteResultsExecutor`, but if you have a more relevant suggestion, I don't mind changing it 😉

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -73,10 +73,7 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
 
         if (!method_exists($autocompleter, 'getGroupBy') || null === $groupBy = $autocompleter->getGroupBy()) {
             foreach ($paginator as $entity) {
-                $results[] = [
-                    'value' => $autocompleter->getValue($entity),
-                    'text' => $autocompleter->getLabel($entity),
-                ];
+                $results[] = self::getAutocompleterResult($autocompleter, $entity);
```
smnandre:
```suggestion
                $results[] = $this->formatResult($autocompleter, $entity);
```

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -104,10 +101,7 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
         $optgroupLabels = [];
 
         foreach ($paginator as $entity) {
-            $result = [
-                'value' => $autocompleter->getValue($entity),
-                'text' => $autocompleter->getLabel($entity),
-            ];
+            $result = self::getAutocompleterResult($autocompleter, $entity);
```
smnandre:
```suggestion
            $result = $this->formatResult($autocompleter, $entity);
```

src/Autocomplete/src/AutocompleteResultsExecutor.php
```diff
@@ -124,4 +118,19 @@ public function fetchResults(EntityAutocompleterInterface $autocompleter, string
 
         return new AutocompleteResults($results, $hasNextPage, $optgroups);
     }
+
+    /**
+     * @return array<string, mixed>
+     */
+    private static function getAutocompleterResult(EntityAutocompleterInterface $autocompleter, object $entity): array
```
smnandre:
```suggestion
    /**
     * @return array{value: scalar, text: string}
     */
    private function getAutocompleterResult(EntityAutocompleterInterface $autocompleter, object $entity): array
```

I'm having doubt about the whole naming here. but in all cases let's use non-static methods here.

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -18,30 +18,52 @@
 /**
  * Interface for classes that will have an "autocomplete" endpoint exposed.
  *
+ * @template T of object
+ *
+ * @method array getResults(object $entity) Returns the autocomplete result, usually the label as "text" and the value as "value".
```
smnandre: Thinking out loud here, but this is dimishing the whole interface here. 

getLabel and getValue are strongly typed methods, enforcing non-null text, etc. 

Replacing it with a losely typed array does not feel good. 

I'm finishing to review format/code thing and i'll make a suggestion about the more global aspect :)

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -18,30 +18,52 @@
...
+    /**
+     * Returns the autocomplete result, usually the label as "text" and the value as "value".
+     * May be used to expose extra options such as "disabled" so that tom-select's rendering may be customized.
+     *
+     * @param T $entity
+     *
+     * @return array<string, mixed>
+     */
+    /* public function getResult(object $entity): array; */
+
```
smnandre: Well... i'm now pretty sure. 

We have 
* a `getLabel` method (equivalent to [choice_label](https://symfony.com/doc/current/reference/forms/types/choice.html#choice-label) in Form options 
* a `getValue` method (equivalent to [choice_value](https://symfony.com/doc/current/reference/forms/types/choice.html#choice-value) in Form options)

Instead of introducing a new method that englobes/replaces the previously existing one, i'd rather add a new method 

```php
/**
 * @return array<string, string|null>|null
 */
public function getAttributes(object $entity): ?array
```

In your scenario, you could return ['disabled' => true] when you'd want to.

Wdyt ?

src/Autocomplete/src/AutocompleteResultsExecutor.php
smnandre: Well yeah We  should rename this method formatResult() or something like buildResult().

src/Autocomplete/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 -   Deprecate `ExtraLazyChoiceLoader` in favor of `Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader`
 -   Reset TomSelect when updating url attribute #1505
+-   Let the EntityAutocompleter drive results creation, so that extra options may be exposed for tom-select to use them
```
smnandre:
```suggestion
-   Add ``getAttributes()`` method to define...
```

Let's be factual on what it _does_  instead on the why here.

smnandre:
Thank you for the PR, this feature (and the effort to improve things passing by -- docblocks for static analysis)!

Minor changes to match existing DX / wording in symfony/form, and maybe add one or two tests to check everything work if the method is not implemented, or does not override getValue/getText

On an implementation level, you should probably do something like this in the formatResult method

```php

$attributes = [];
if (method_exists(..... 
    ....     
}

return [
   ...$attributes, 
   'value' => $autocompleter->getValue(),
   'text' => $autocompleter->getText(),
];
```

src/Autocomplete/src/EntityAutocompleterInterface.php
smnandre: Understood! See other suggestions below regarding naming but mostly "meaning" of the new method

J-Ben87:
Thanks for your review.

I like your idea of merging extra attributes into an array containing the text and the value.

WDYT of the latest version of my branch?

src/Autocomplete/src/Form/WrappedEntityTypeAutocompleter.php
```diff
@@ -114,6 +114,14 @@ public function getValue(object $entity): string
...
     }
 
+    public function getResult(object $entity): array
+    {
+        return [
+            'value' => $this->getValue($entity),
+            'text' => $this->getLabel($entity),
+        ];
+    }
+
```
smnandre:
```suggestion
```

src/Autocomplete/tests/Fixtures/Autocompleter/CustomAttributesProductAutocompleter.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace Symfony\UX\Autocomplete\Tests\Fixtures\Autocompleter;
+
+class CustomAttributesProductAutocompleter extends CustomProductAutocompleter
+{
+    public function getAttributes(object $entity): array
+    {
+        return ['disabled' => true];
```
smnandre:
```suggestion
        return [
            'disabled' => true,
            'text' => 'This value should be replaced with the result of getText()',
            'value' => 'This value should be replaced with the result of getValue()',            
        ];
```

That way we also prevent any future regression on method precedence

src/Autocomplete/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 -   Deprecate `ExtraLazyChoiceLoader` in favor of `Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader`
 -   Reset TomSelect when updating url attribute #1505
+-   Add a `getAttributes()` method to define extra attributes to add to the autocomplete result.
```
smnandre:
```suggestion
-   Add `getAttributes()` method to define additional attributes for autocomplete results #2541
```

smnandre:
Tiny details, then lets see the CI failures and LGTM :)

Would you be ok (on another PR) to write a paragraph + example about this in the documentation ? I can help if you need / want.

src/Autocomplete/tests/Fixtures/Autocompleter/CustomAttributesProductAutocompleter.php
J-Ben87: Crap, I thought of it last night and wanted to implement it this morning, but you've been too fast ^^

J-Ben87:
Ok everything should be fine now 👍🏻 

Regarding the failing tests however, ~I have no clue why they are failing since they are related to zenstruck/foundry deprecations in another component~...
It says:
```
Since zenstruck/foundry 2.4: In order to use Foundry, you must use the trait "Symfony\Bundle\FrameworkBundle\Test\KernelTestCase" in your "Symfony\UX\LiveComponent\Tests\Unit\Form\ComponentWithFormTest" tests. This will throw an exception in 3.0.
```
However there is no `KernelTestCase` trait available in the project. I assume the error message is inaccurate 🤔 

---

Edit: I found the failing tests and I fixed them 🥳

J-Ben87:
No worries for writing a bit of documentation, could you just point me the place and the kind of doc you expect?

smnandre (approved):
👍  for me! 

Nice PR and thank you @J-Ben87

J-Ben87:
@Kocal Is that all good for you too? Your change request is still pending

src/Autocomplete/src/EntityAutocompleterInterface.php
```diff
@@ -18,30 +18,46 @@
...
      * Returns the "value" attribute for this entity, usually the id.
+     *
+     * @param T $entity
      */
     public function getValue(object $entity): mixed;
 
+    /**
+     * Returns extra attributes to add to the autocomplete result.
+     */
+    /* public function getAttributes(object $entity): array; */
```
Kocal, line 63: Let's add a comment to remember that we should uncomment this line for Symfony UX 3

J-Ben87: Done ✅

Kocal (approved):

J-Ben87:
What is the next step to have this PR merged?
It says that it needs 2 approving reviews by reviewers with write access. Should I ping someone? Or should I just wait for one of them to notice it's ready?

Sorry if it's obvious, but like I said I'm not used to contribute so I'm confused about the next steps 😕

smnandre:
> What is the next step to have this PR merged? It says that it needs 2 approving reviews by reviewers with write access. Should I ping someone? Or should I just wait for one of them to notice it's ready?

@Kocal gave the second approval today, so we will merge it soon :) 

Often, we let some time for all maintainers to see the PR, as we are not all on the same timezone and we all need to find the good balance between day jobs, family life, and open source.

So here, the next steps are: 

**Merged:** 
someone will merge the PR (you will then be able to test it if you use the "2.x-dev" branch)

**Released:** 
Then, on the next UX release, it will be available for everyone to enjoy. We are releasing a new version every-ish month (see [the changelog](https://ux.symfony.com/changelog)). The next one will be the `2.23`, and should be released in the next few days. So you should not have to wait much here.

> Sorry if it's obvious, but like I said I'm not used to contribute so I'm confused about the next steps 😕

It's not obvious :)

And welcome aboard!

smnandre:

J-Ben87:
Thanks for your guidance and your availability 🙏🏻

PR #2550, opened 2025-02-05
[Turbo] Add stream format with request listener

### LINKED ISSUE #2549: [Turbo] Request format is not registered for workers
`BundleInterface::boot()` is only called when the worker is booted.

This leads to the stream format not being available for consecutive requests to the same worker.

I noticed this when trying to find out why a valid Turbo request never returned a stream response.

aleho:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2549
| License       | MIT

BundleInterface::boot() is only called when booting a worker and not on every request. So on FrankenPHP for example Turbo requests would never return a stream when checking with `if ($request->getPreferredFormat() === TurboBundle::STREAM_FORMAT`.

This replaces `TurboBundle::boot()` with a request listener to register the format on every request.

### DIFF
--- a/src/Turbo/config/services.php
+++ b/src/Turbo/config/services.php
@@ -11,11 +11,13 @@
 
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
+use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\UX\Turbo\Broadcaster\BroadcasterInterface;
 use Symfony\UX\Turbo\Broadcaster\IdAccessor;
 use Symfony\UX\Turbo\Broadcaster\ImuxBroadcaster;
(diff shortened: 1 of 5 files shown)


aleho:
The failing tests are in LiveComponents and a timeout in webdriver. I hope they're not related to my changes ;).

smnandre:
Hi @aleho ! Thank you for this PR :) 

Do you think you could add a small test to prevent any future regression?

aleho:
@smnandre I wasn't entirely sure how to test that. Is "simulating" a worker mode by not re-booting the kernel enough?

smnandre:
@aleho i think so!  cc @kbond ?

kbond (approved):
Cool, makes sense!

smnandre:
Thanks @aleho.

PR #2554, opened 2025-02-07
[Map] Add Clustering Algorithms

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Issues        | Fix #... 
| License       | MIT

Provide an interface and the first two PHP implementations of Clustering algorithms
- GridClustering
- Morton

Performances: < 1ms per 1000 Points *

* tested on a non production server (... my iMac 😅 )

**Next Step** will require JS code.

But this is already usefull with PHP classes only, for whomever has large amount of markers

### DIFF
--- a/src/Map/CHANGELOG.md
+++ b/src/Map/CHANGELOG.md
@@ -3,11 +3,11 @@
 ## 2.29.0
 
 -  Add Symfony 8 support
+-  Add `Cluster` class and `ClusteringAlgorithmInterface` with two implementations `GridClusteringAlgorithm` and `MortonClusteringAlgorithm`
 
 ## 2.28
 
 -  Add `minZoom` and `maxZoom` options to `Map` to set the minimum and maximum zoom levels
--  The package is not experimental anymore
 
 ## 2.27
 
--- a/src/Map/src/Cluster/Cluster.php
+++ b/src/Map/src/Cluster/Cluster.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Map\Cluster;
+
+use Symfony\UX\Map\Point;
+
+/**
+ * Cluster representation.
+ *
+ * @implements \IteratorAggregate<int, Point>
+ *
+ * @author Simon André <smn.andre@gmail.com>
+ */
+final class Cluster implements \Countable, \IteratorAggregate
+{
+    /**
+     * @var Point[]
+     */
+    private array $points = [];
+
+    private float $sumLat = 0.0;
+    private float $sumLng = 0.0;
+    private int $count = 0;
+
+    /**
+     * Initializes the cluster with an initial point.
+     */
+    public function __construct(Point $initialPoint)
+    {
+        $this->addPoint($initialPoint);
+    }
+
+    public function addPoint(Point $point): void
+    {
+        $this->points[] = $point;
+        $this->sumLat += $point->getLatitude();
+        $this->sumLng += $point->getLongitude();
+        ++$this->count;
+    }
+
+    /**
+     * Returns the center of the cluster as a Point.
+     */
+    public function getCenter(): Point
+    {
+        return new Point($this->sumLat / $this->count, $this->sumLng / $this->count);
+    }
+
+    /**
+     * @return non-empty-list<Point>
+     */
+    public function getPoints(): array
+    {
+        return $this->points;
+    }
+
+    /**
+     * Returns the number of points in the cluster.
+     */
+    public function count(): int
+    {
+        return $this->count;
+    }
+
+    /**
+     * @return \Traversable<int, Point>
+     */
+    public function getIterator(): \Traversable
+    {
+        return new \ArrayIterator($this->points);
+    }
+}
--- a/src/Map/tests/Cluster/ClusteringPerformanceTest.php
+++ b/src/Map/tests/Cluster/ClusteringPerformanceTest.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Map\Tests\Cluster;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\UX\Map\Cluster\ClusteringAlgorithmInterface;
+use Symfony\UX\Map\Cluster\GridClusteringAlgorithm;
+use Symfony\UX\Map\Cluster\MortonClusteringAlgorithm;
+use Symfony\UX\Map\Point;
+
+class ClusteringPerformanceTest extends TestCase
+{
+    /**
+     * @const array<float>
+     */
+    private const ZOOMS = [
+        2.0,
+        5.0,
+        8.0,
+    ];
+
+    /**
+     * @const array<string>
+     */
+    private const ALGORITHMS = [
+        GridClusteringAlgorithm::class,
+        MortonClusteringAlgorithm::class,
+    ];
+
+    /**
+     * @return iterable<array{0: ClusteringAlgorithmInterface, 1: float}>
+     */
+    public static function algorithmProvider(): iterable
+    {
+        foreach (self::ZOOMS as $zoom) {
+            foreach (self::ALGORITHMS as $algorithm) {
+                yield $algorithm.' '.$zoom => [new $algorithm(), $zoom];
+            }
+        }
+    }
+
+    /**
+     * Scenario 1: Large number of points (50,000), concentrated area (Paris region).
+     *
+     * @dataProvider algorithmProvider
+     */
+    public function testScenarioRegion50000(ClusteringAlgorithmInterface $algorithm, float $zoom): void
+    {
+        $points = $this->generatePoints(50000, 48.8, 49, 2.2, 2.5);
+
+        $this->runPerformanceTest($algorithm, $points, $zoom);
+    }
+
+    /**
+     * Scenario 2: Moderate number of points (5,000), broad area (France and surroundings).
+     *
+     * @dataProvider algorithmProvider
+     */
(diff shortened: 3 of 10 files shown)


smnandre:
(cannot ask for your review @Kocal as you're code owner, i would have otherwise ;) )

Kocal (commented):
Can you provide some examples/use-cases and documentation? I'm not really sure to understand what you are solving right now.

When you say "next step is JS", do you mean https://github.com/Leaflet/Leaflet.markercluster & https://developers.google.com/maps/documentation/javascript/marker-clustering?hl=fr implementation?

smnandre:
You have 30000 markers in database. Sending them all in HTML is non sense. Clustering them in backend is definitively the way to go. That way you only send data you want to display, and fix the performance problem some people had with too many infowindow for instance. 

Moreover it enforces the initial problem this bundle save: abstract the map services and manipulate data / UI from backend. 

People that do want to use custom GoogleMaps SDK Methods wont install this Bundle, so -personal opinion- it’s a mistake to base our choices on what they would do there.

Also this would not coexist with Live trait for now.

So Next step is to continue the path and provide the missing link: making the LiveComponent able to react on map events: zoom or pan, clicks, etc. 

That’s where I am right now, I Hope i’ve clarified my vision and goal for both this feature and more generally this bundle. 

And would love to integrate this Into yours, if you have a différent one :)

Kocal:
Thanks for the explanations! I was asking because I didn't see any modifications on `Map` nor in the documentation, and so I was wondering what you really wanted to do with this PR.

smnandre:
Yeah i'm sorry I miss time and need to push more often, or i end with hundreds of "features ready at 80%" on my computer and never find time to release them.

Here, it's because two unrelated people asked me for the same exact thing last week, I've reopened my "geo" archives and started to prepare the foundations. 

I'd like to end this (documentation included) before 2.24 😅

Kocal (approved):

Kocal:
After a small rebase, that would be perfect 🤓

rvmourik:
Just a question out of curiosity, will this be in 2.24? Thanks in advance

mart-insiders:
Hi @rvmourik @Kocal , 
I was wondering what the current state of this feature is and what is required to finish it. 

At the moment some issues occur when implementing [https://github.com/Leaflet/Leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster) with the current leaflet bridge.
The only possible way to add a leaflet layer with clustered markers at the moment is via javascript. 
This feels like unnecessary iterations over every marker in your stimulus controller.

`import { Controller } from "@hotwired/stimulus";

/* stimulusFetch: 'lazy' */
    connect() {
        this.element.addEventListener('ux:map:connect', this._onConnect);
    }

    disconnect() {
        this.element.removeEventListener('ux:map:connect', this._onConnect);
    }

    async _onConnect(event) {
      await import("leaflet.markercluster/dist/leaflet.markercluster-src.js");

      const { map } = event.detail;

      var clusterMarkerLayer = L.markerClusterGroup({
        spiderfyOnMaxZoom: true,
        showCoverageOnHover: false,
        zoomToBoundsOnClick: true
      });

      map.eachLayer(layer => {
        if (!(layer instanceof L.Marker)) {
            return;
        }

        clusterMarkerLayer.addLayer(layer);
        map.removeLayer(layer);
      })

      map.addLayer(clusterMarkerLayer);
    }
}`

An issue which occurs is Events are not persisted when creating a marker on this different layer.
Would the following changes suffice to resolve this?
- Implementation of a Layers object within the Map object or a reference in the Marker object. 
- Changing the `doCreateMarker` in `map_controller.js` to add a marker to a configured layer or as a layer to the map.

I would like to hear your opinion on this matter.

dannyvw:
@smnandre Do you perhaps have an update for this feature?

smnandre:
100% forgot this one : will rebase/clean tomorrow!

smnandre:
Cleaned a bit: 
- Cluster is now countable + iterable
- remove getCenterLat / getCenterLng with getCenter(): Point
- add PHP 8.1/2 support
- removed readonly classes
- rebased  and changelog updated

Not sure if doc will follow first, or if i first open the PR to integrate this in the components (depending on the time i find)

@Kocal : to me it's ready 

Thx @dannyvw for the poke :)

src/Map/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## 2.28
 
 -  Add `minZoom` and `maxZoom` options to `Map` to set the minimum and maximum zoom levels
+-  Add `Cluster` class, interface for clustering algorithm and two implementations:
+   `GridClusteringAlgorithm` and `MortonClusteringAlgorithm`.
```
Kocal: What about:
```suggestion
-  Add `Cluster` class and `ClusteringAlgorithmInterface` with two implementations `GridClusteringAlgorithm` and `MortonClusteringAlgorithm`.
```

?

src/Map/src/Cluster/Cluster.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Initializes the cluster with an initial point.
+     */
+    public function __construct(Point $initialPoint)
+    {
+        $this->points[] = $initialPoint;
+        $this->sumLat = $initialPoint->getLatitude();
+        $this->sumLng = $initialPoint->getLongitude();
+        $this->count = 1;
```
Kocal: Maybe..? 
```suggestion
        $this->addPoint($initialPoint);
```

src/Map/src/Cluster/Cluster.php
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Returns the center of the cluster as a Point.
+     */
+    public function getCenter(): Point
+    {
+        return new Point($this->sumLat / $this->count, $this->sumLng / $this->count);
+    }
+
+    /**
+     * @return Point[]
```
Kocal: We can be more precise here: 
```suggestion
     * @return non-empty-list<Point>
```

src/Map/tests/Cluster/ClusteringPerformanceTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Scenario 1: Large number of points (50,000), concentrated area (Paris region).
+     *
+     * @dataProvider algorithmProvider
+     */
+    public function testScenarioRegion50000(ClusteringAlgorithmInterface $algorithm, float $zoom): void
+    {
+        $points = $this->generatePoints(50000, 48.8, 49, 2.2, 2.5);
+        $this->runPerformanceTest($algorithm, $points, $zoom);
```
Kocal:
```suggestion
        $points = $this->generatePoints(50000, 48.8, 49, 2.2, 2.5);

        $this->runPerformanceTest($algorithm, $points, $zoom);
```

src/Map/tests/Cluster/ClusteringPerformanceTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Scenario 3: Very large number of points (100,000), global distribution.
+     *
+     * @dataProvider algorithmProvider
+     */
+    public function testScenarioWorld100000(ClusteringAlgorithmInterface $algorithm, float $zoom): void
+    {
+        $points = $this->generatePoints(100000, -90, 90, -180, 180);
+        $this->runPerformanceTest($algorithm, $points, $zoom);
```
Kocal:
```suggestion
        $points = $this->generatePoints(100000, -90, 90, -180, 180);

        $this->runPerformanceTest($algorithm, $points, $zoom);
```

src/Map/tests/Cluster/GridClusteringAlgorithmTest.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $point1 = new Point(10.0, 20.0);
+        $point2 = new Point(-10.0, -20.0); // Far away
+        $algorithm = new GridClusteringAlgorithm();
+
+        // High zoom
+        // Expect two separate clusters
+        $clusters = $algorithm->cluster([$point1, $point2], 5.0);
+
+        $this->assertCount(2, $clusters);
```
Kocal: I think it could have more value (feel free to modify the message if you want)
```suggestion
        $clusters = $algorithm->cluster([$point1, $point2], 5.0);

        $this->assertCount(2, $clusters, 'Two clusters should have created due to the high zoom value.');
```

src/Map/tests/Cluster/GridClusteringAlgorithmTest.php
```diff
@@ -0,0 +1,101 @@
...
+    public function testPointsInSameGridAreClusteredTogether(): void
+    {
+        $point1 = new Point(10.0, 20.0);
+        $point2 = new Point(10.1, 20.1);
+        $algorithm = new GridClusteringAlgorithm();
+
+        // Low zoom
+        $clusters = $algorithm->cluster([$point1, $point2], 1.0);
+
+        $this->assertCount(1, $clusters);
```
Kocal:
```suggestion
        $clusters = $algorithm->cluster([$point1, $point2], 1.0);

        $this->assertCount(1, $clusters, 'One cluster should have been created due to the low zoom value.');
```

src/Map/tests/Cluster/GridClusteringAlgorithmTest.php
Kocal: (same comments for other similar code)

Kocal (approved):

src/Map/src/Cluster/Cluster.php
smnandre: My code initially... but I had alert for division by zero risks in PhpStorm..

But in fact yeah that's better. Just need to add back initial values

src/Map/src/Cluster/Cluster.php
```diff
@@ -0,0 +1,84 @@
...
+final class Cluster implements \Countable, \IteratorAggregate
+{
+    /**
+     * @var Point[]
+     */
+    private array $points = [];
+
+    private float $sumLat;
+    private float $sumLng;
+    private int $count;
```
smnandre:
```suggestion
    private float $sumLat = 0.0;
    private float $sumLng = 0.0;
    private int $count = 0;
```

smnandre:
Thanks for all the feedback / suggestions @Kocal : all commited :)

Kocal (approved):

Kocal:
Rebased, conflicts fixed, and added some documentation.

Will merge when CI becomes green

Kocal:
Thank you @smnandre.

Issue #2118, opened 2024-09-01
[DX][Components] Twig components case sensitivity / HTML syntax


smnandre:
## TL;DR; ⏲️

* Should we adopt a naming convention for component names in UX packages and documentation? If so, which one?
* Should all component names be case-sensitive?

## Context 🔠

While working on UX Map, @Kocal raised a question during [the discussion](https://github.com/symfony/ux/pull/2117#discussion_r1739996205):

> Should it be `<twig:ux:map />` or `<twig:UX:Map />`? (I noticed you used `UX:Map` in the `twig.component` tag). Or is it case-insensitive?

During the development and early stages of UX Icons, @kbond and I decided to switch from `<twig:Ux:Icon />` to `<twig:ux:icon />` both [in the code](https://github.com/symfony/ux/pull/1659) and [in the documentation](https://github.com/symfony/ux/pull/1798). 

This change felt more natural for typing and reading in the documentation. Additionally, since "`<twig`" is lowercase, "`<twig:ux`" seemed more consistent to me. As a result, we made it case-insensitive.

However, this is the only place where it works this way, as user-land component names are still case-sensitive.

## Suggestions 🗳️

I’d like to gather feedback from the team and community on the following points:

### Symfony UX Components
* **A0:** Symfony UX component names (e.g., Map, Icon, and future ones) should be case-insensitive (in HTML syntax).
* **A1:** Symfony UX components should adopt a consistent casing style in documentation, code blocks, and social media.
* **A2:** Symfony UX components should use lowercase casing style in documentation, code blocks, and social media.

### App/User-Land/Bundle Components
* **B0:** Twig/Live Component names should be case-sensitive (as these names may be used in events, logs, etc.).
* **B1:** [NEW] Twig/Live Component names should be unique in a case-insensitive form (preventing `Acme:Foo` and `acMe:foo` from being registered simultaneously).
* **B2:** [NEW] We could (later) suggest the correct name when a user misspells it, or even automatically map it to the correct one.

.

I’m inclined to answer "yes" to all these points.

Let’s discuss them and hear from you if there are any concerns. :)

smnandre:
cc @Kocal @kbond @WebMamba :)

Kocal:
Hey Simon, thanks for opening the discussion! 🙏🏻

## Symfony UX Components 
[...]

PR #2555, opened 2025-02-08
Update documentation and tests to change `twig:ux:component` to `twig:UX:Component`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Applying suggestions for `Symfony UX Components` section in OP's message, see #2118

### DIFF
--- a/src/Icons/CHANGELOG.md
+++ b/src/Icons/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.20.0
 
 -   Add `aliases` configuration option to define icon alternative names.
--   Add support for `int` and `float` attribute values in `<twig:ux:icon />`.
+-   Add support for `int` and `float` attribute values in `<twig:UX:Icon />`.
 
 ## 2.19.0
 
--- a/src/Icons/README.md
+++ b/src/Icons/README.md
@@ -7,7 +7,7 @@ Renders local and remote [SVG icons](https://ux.symfony.com/icons) in your Twig 
 {{ ux_icon('mdi:check', {class: 'w-4 h-4'}) }}
 
 {# .. or Twig Component #}
-<twig:ux:icon name="mdi:check" class="w-4 h-4" />
+<twig:UX:Icon name="mdi:check" class="w-4 h-4" />
 
 {# Render the "check" icon from "mdi" pack with class #}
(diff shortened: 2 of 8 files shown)


smnandre:
Well.... this wasnt my suggestion ^^

And we made this choice because it felt weird to use <twig:UX:Icon  (lower / UPPER / Camel).

Open to change of course, but does not feel the previous topic raised massive interest :) 

> A2: Symfony UX components should use lowercase casing style in documentation, code blocks, and social media.

😅

smnandre:
And this would require a lot of other changes since we acted the ux:icon case (commands, docs, website, ...).

Not strongly opiniated here (really) and open to change, but then i guess we should do it all at once, to not confuse dev/users

src/Icons/tests/Fixtures/templates/template1.html.twig
```diff
@@ -4,7 +4,7 @@
     <li id="third">{{ ux_icon('sub:check', {'data-action': 'string "with" quotes'}) }}</li>
     <li id="fourth">{{ ux_icon('sub:check', { height: 24, width: "24" }) }}</li>
     <li id="fifth"><twig:UX:Icon name="user" class="h-8 w-8" /></li>
-    <li id="sixth"><twig:ux:icon name="sub:check" aria-labelledby="foo" /></li>
+    <li id="sixth"><twig:UX:Icon name="sub:check" aria-labelledby="foo" /></li>
```
smnandre: This one is a test (the various formats) to ensure we can handle them all

Kocal: Oops, good catch

Kocal:
After a little thought, I understand the problem with the following proposed naming `lower:UPPER:PascaleCase` (e.g.: `twig:UX:Map`). I was thinking this way:
- `twig`, because it's the obligatory prefix for filling in a component
- `UX`, because Symfony “UX”, and it can be associated with the PascalCase naming of the component
- `PascalCase`, as documented on https://symfony.com/bundles/ux-twig-component/current/index.html, but also because this convention is the same as with React or Vue (even though Vue also allows kebab-case).

It is true it makes more sense to use `<twig:ux:icon>`.

PR #2572, opened 2025-02-12
[TwigComponent] Fix ComponentTokenParser on 32-bits

### LINKED ISSUE #2568: [TwigComponent] ComponentTokenParser failing to generate template index on 32bit platform
This issue affects **only 32bit** platforms. 

Appending digits to a 32bit integer in [ComponentTokenParser::generateEmbeddedTemplateIndex](https://github.com/symfony/ux/blob/63c88090dcac127e1263bfa5d9e43894ac312a95/src/TwigComponent/src/Twig/ComponentTokenParser.php#L121) is prone to integer overflows on 32bit systems, resulting in errors like:

> Symfony\UX\TwigComponent\Twig\ComponentTokenParser::generateEmbeddedTemplateIndex(): Return value must be of type int, string returned 

I guess, that’s what happened in https://github.com/EasyCorp/EasyAdminBundle/issues/6713

Furthermore, `crc32` itself [can return negative integers](https://www.php.net/manual/en/function.crc32.php#:~:text=many%20crc32%20checksums%20will%20result%20in%20negative%20integers%20on%2032bit%20platforms) on 32bit which can lead to unexpected results (`-` in [compiled class names](https://github.com/twigphp/Twig/blob/f1c9039106ecc5a8182e1536027a6940ac4a8ab8/src/Node/ModuleNode.php#L183)).

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2568
| License       | MIT

- 64-bit users get the original behavior
- 32-bit users get the fix to prevent overflow issues

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentTokenParser.php
+++ b/src/TwigComponent/src/Twig/ComponentTokenParser.php
@@ -118,6 +118,14 @@ final class ComponentTokenParser extends AbstractTokenParser
             $this->lineAndFileCounts[$fileAndLine] = 0;
         }
 
-        return crc32($fileAndLine).++$this->lineAndFileCounts[$fileAndLine];
+        $index = crc32($fileAndLine).++$this->lineAndFileCounts[$fileAndLine];
+
+        if (4 === \PHP_INT_SIZE) {
+            // On 32-bit PHP, the index can be negative or greater than PHP_INT_MAX
+            // we need to convert it to a positive 32-bit integer
+            $index = fmod(abs($index), \PHP_INT_MAX) + 1;
+        }
+
+        return (int) $index;
     }
 }


src/TwigComponent/src/Twig/ComponentTokenParser.php
```diff
@@ -118,6 +118,14 @@ private function generateEmbeddedTemplateIndex(string $file, int $line): int
             $this->lineAndFileCounts[$fileAndLine] = 0;
         }
 
-        return crc32($fileAndLine).++$this->lineAndFileCounts[$fileAndLine];
+        $index = crc32($fileAndLine).++$this->lineAndFileCounts[$fileAndLine];
+
+        if (4 === \PHP_INT_SIZE) {
+            // On 32-bit PHP, the index can be negative
+            // we need to convert it to a positive integer
+            $index = abs($index) % \PHP_INT_MAX + 1;
```
smnscp:
```suggestion
            // On 32-bit PHP, the index can be negative or greater than PHP_INT_MAX
            // we need to convert it to a positive 32-bit integer
            $index = fmod(abs($index), \PHP_INT_MAX) + 1;
```
Using `fmod` here instead of the modulo operator is important, because `abs($index)` might return a float greater than `PHP_INT_MAX`, which `%` would [convert to a negative integer](https://www.php.net/manual/en/language.operators.arithmetic.php#:~:text=Operands%20of%20modulo%20are%20converted%20to%20int), resulting [in a negative](https://www.php.net/manual/en/language.operators.arithmetic.php#:~:text=has%20the%20same%20sign%20as%20the%20dividend) remainder.

smnscp (changes requested):
Hi @smnandre, I tested the patch on a 32-bit system. Unfortunately, `console clear:cache` failed again with:
> syntax error, unexpected token "-", expecting "{"

That’s due to a compiled template component class with a `-` in the class name.
My suggestion solves this problem.

src/TwigComponent/src/Twig/ComponentTokenParser.php
smnscp: FWIW, we could add explicit integer conversion to that line to spare confusion. Up to you.

smnandre:
> Hi @smnandre, I tested the patch on a 32-bit system. Unfortunately, `console clear:cache` failed again with:
> 
> > syntax error, unexpected token "-", expecting "{"
> 
> That’s due to a compiled template component class with a `-` in the class name. My suggestion solves this problem.

Thanks a lot! I've commited your suggested code. Tell me if this is ok for you and LGTM :)

smnscp (approved):
LGTM. Thanks!

smnandre:
Note: test failures are related to deprecations in Twig, that will require a dedicated PR

Kocal (approved):

PR #2587, opened 2025-02-21
[LiveComponent] Fix default select value with preferred choices

1ed:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | -
| License       | MIT

When there are preferred_choices set for a choice type Symfony
renders them before the actual choices so the default value in the
browser will be the first value of the preferred choices.


smnandre:
Thank you @1ed !

--- 

Thinking out loud here... I'm less and less enjoying this whole mecanism. 

I'm starting to wonder if we should not revert it..  instead of trying to decide PHP-side how the browsers behave 🤷‍♂️

src/LiveComponent/src/ComponentWithFormTrait.php
```diff
@@ -288,7 +288,7 @@ private function extractFormValues(FormView $formView): array
                 && !$child->vars['multiple']                // is not multiple
                 && !\is_string($child->vars['placeholder'])  // has no placeholder (empty string is valid)
             ) {
-                $choices = $child->vars['choices'];
+                $choices = $child->vars['preferred_choices'] ?: $child->vars['choices'];
```
smnandre: Could you add some comment here (more or less as the other lines) to keep this always-expanding list ? 

(i'm thinking to the next ones that would need to understand the whole system with no comments 😅 )

src/LiveComponent/tests/Fixtures/Form/FormWithManyDifferentFieldsType.php
```diff
@@ -74,6 +74,28 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     'foo' => 1,
                 ],
             ])
+            ->add('choice_required_with_preferred_choices_array', ChoiceType::class, [
```
smnandre, line 77: Maybe an empty array of preferred choices just to be triple sure ? 

(i know this is the case without setting it -- aka the other tests, but this will prevent any regression if ChoiceType internals / view data change)

smnandre (approved):

1ed:
> Thank you @1ed !
> 
> Thinking out loud here... I'm less and less enjoying this whole mecanism.
> 
> I'm starting to wonder if we should not revert it.. instead of trying to decide PHP-side how the browsers behave 🤷‍♂️

Probably would be better yes, but I don't know the internals, why was this necessary originally, just wanted to fix the bug.

smnandre:
Of course, and thank you again for that ! 

It was really thinking outloud :))

Kocal (approved):

Kocal:
Thank you @1ed.

Issue #2575, opened 2025-02-16
[Stimulus] Create NPM Package for @symfony/stimulus-bundle to Decouple from PHP Components


alinceDev:
### Issue Description:

Currently, certain JavaScript bundles, such as @symfony/stimulus-bundle, are being installed from the vendor package instead of an NPM package. This is evident in the package.json file where the dependency is specified as:

```json
    "@symfony/stimulus-bundle": "file:vendor/symfony/stimulus-bundle/assets",
```

### Problem

This approach creates a tight coupling between the JavaScript and PHP components, making it difficult to manage and install JavaScript dependencies independently. For instance, in a CI environment, this setup requires the installation of PHP components via Composer to install the JavaScript packages, which is not ideal for running tests with Playwright or other JavaScript-focused tasks.

### Request

Could you please create an NPM package for @symfony/stimulus-bundle (and potentially other relevant bundles) to allow for independent installation via NPM? This would streamline the setup process, especially in CI environments, and decouple the JavaScript dependencies from the PHP components.

### Use Case Example

For example, when setting up a CI pipeline to run Playwright tests, it would be beneficial to install the necessary JavaScript packages directly via NPM without needing to install the PHP components through Composer.

If this issue is already documented and I missed the information, I apologize for the oversight. Thank you for considering this enhancement!

smnandre:
Say we do publish on npm the LiveComponent javascript part.

Next release we have a new feature that does work with JS and PHP.

How would you enforce both vendors/symfony/ux-live-component and nodes_modules/vendor/symfony.. are in sync ?

(genuine question)

Kocal:
Maybe we can hook on `postinstall` npm script and read user's `composer.lock` to see if the PHP package has the same version.

Kocal:
Done in #2595, #2612 and #2617

PR #2595, opened 2025-02-23
Normalize package.json files, to prepare publication on NPM

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

This is a first iteration for #2575: rework our `package.json` files in order to make them more "friendly" and work on npm.

Next steps:
1. Add some verifications, to ensure PHP and JS versions of a same package are identical (and so 100% compatible)
2. See how we can deploy JS packages when creating a new UX releases. I'm thinking about a full automated process using the CI with a workflow listening on tag creation, which will be responsible to upgrade the `version` from `package.json` files, commit, push, and publish to npm)
3. Check if we need to manually create the packages on npm before trying to publishing them

### DIFF
--- a/src/Autocomplete/assets/package.json
+++ b/src/Autocomplete/assets/package.json
@@ -7,7 +7,7 @@
         "symfony-ux"
     ],
     "homepage": "https://ux.symfony.com/autocomplete",
-    "repository": "https://github.com/symfony/ux-autocomplete/tree/2.x",
+    "repository": "https://github.com/symfony/ux-autocomplete",
     "type": "module",
     "files": [
         "dist"
--- a/src/Autocomplete/assets/package.json
+++ b/src/Autocomplete/assets/package.json
@@ -1,10 +1,19 @@
 {
     "name": "@symfony/ux-autocomplete",
-    "description": "JavaScript-powered autocompletion functionality for forms.",
+    "description": "JavaScript Autocomplete functionality for Symfony",
+    "license": "MIT",
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux"
+    ],
+    "homepage": "https://ux.symfony.com/autocomplete",
+    "repository": "https://github.com/symfony/ux-autocomplete/tree/2.x",
+    "type": "module",
+    "files": [
+        "dist"
+    ],
     "main": "dist/controller.js",
     "types": "dist/controller.d.ts",
-    "version": "1.0.0",
-    "license": "MIT",
     "scripts": {
         "build": "node ../../../bin/build_package.js .",
         "watch": "node ../../../bin/build_package.js . --watch",
--- a/src/LiveComponent/assets/package.json
+++ b/src/LiveComponent/assets/package.json
@@ -9,7 +9,7 @@
         "components"
     ],
     "homepage": "https://ux.symfony.com/live-component",
-    "repository": "https://github.com/symfony/ux-live-component/tree/2.x",
+    "repository": "https://github.com/symfony/ux-live-component",
     "type": "module",
     "files": [
         "dist"
--- a/src/LiveComponent/assets/package.json
+++ b/src/LiveComponent/assets/package.json
@@ -1,13 +1,24 @@
 {
     "name": "@symfony/ux-live-component",
     "description": "Live Component: bring server-side re-rendering & model binding to any element.",
+    "license": "MIT",
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux",
+        "twig",
+        "components"
+    ],
+    "homepage": "https://ux.symfony.com/live-component",
+    "repository": "https://github.com/symfony/ux-live-component/tree/2.x",
+    "type": "module",
+    "files": [
+        "dist"
+    ],
     "main": "dist/live_controller.js",
     "types": "dist/live_controller.d.ts",
-    "version": "1.0.0",
     "config": {
         "css_source": "styles/live.css"
     },
-    "license": "MIT",
     "scripts": {
         "build": "node ../../../bin/build_package.js .",
         "watch": "node ../../../bin/build_package.js . --watch",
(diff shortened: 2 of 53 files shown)


src/Autocomplete/assets/package.json
```diff
@@ -1,10 +1,19 @@
...
     "name": "@symfony/ux-autocomplete",
-    "description": "JavaScript-powered autocompletion functionality for forms.",
+    "description": "JavaScript Autocomplete functionality for Symfony",
+    "license": "MIT",
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux"
+    ],
+    "homepage": "https://ux.symfony.com/autocomplete",
+    "repository": "https://github.com/symfony/ux/tree/2.x/src/Autocomplete",
```
smnandre: Question: should this be the monorepo or the split one ? 

( https://github.com/symfony/ux-autocomplete [...] ) ?

src/LiveComponent/assets/package.json
```diff
@@ -1,13 +1,24 @@
...
     "description": "Live Component: bring server-side re-rendering & model binding to any element.",
+    "license": "MIT",
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux",
+        "twig",
+        "components"
+    ],
+    "homepage": "https://ux.symfony.com/cropperjs",
+    "repository": "https://github.com/symfony/ux/tree/2.x/src/Cropperjs",
```
smnandre:
```suggestion
    "homepage": "https://ux.symfony.com/live-component",
    "repository": "https://github.com/symfony/ux/tree/2.x/src/LiveComponent",
```

src/Swup/assets/package.json
```diff
@@ -2,7 +2,16 @@
...
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux"
+    ],
+    "homepage": "https://ux.symfony.com/swup",
+    "repository": "https://github.com/symfony/ux/tree/2.x/src/Swup",
+    "type": "module",
+    "files": [
+        "dist"
+    ],
```
smnandre: I think i would not prepare Swup for the future 😶

src/Svelte/assets/package.json
```diff
@@ -1,9 +1,18 @@
...
     "name": "@symfony/ux-svelte",
     "description": "Integration of Svelte in Symfony",
-    "main": "dist/register_controller.js",
-    "version": "1.0.0",
     "license": "MIT",
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux"
+    ],
+    "homepage": "https://ux.symfony.com/Svelte",
```
smnandre:
```suggestion
    "homepage": "https://ux.symfony.com/svelte",
```

src/Typed/assets/package.json
```diff
@@ -2,7 +2,16 @@
...
+    "version": "2.23.0",
+    "keywords": [
+        "symfony-ux"
+    ],
+    "homepage": "https://ux.symfony.com/typed",
+    "repository": "https://github.com/symfony/ux/tree/2.x/src/Typed",
+    "type": "module",
+    "files": [
+        "dist"
+    ],
```
smnandre: Same comment than Swipe .... but i would 100% understand if the simpler for you here is to do them all the same way

src/Autocomplete/assets/package.json
Kocal: Let's use splited repositories

src/LiveComponent/assets/package.json
Kocal: Nice catch 🙈

src/Autocomplete/assets/package.json
skmedix: I wouldn't mind using monorepo here, as it is a go-to place for proposing contributions etc

smnandre: Problem is, JsDeliver, Packagist, etc;... do not work well with monorepo... so it depends on "how" this package.json attribute is used (I have no idea so I will trust you guys on this)

skmedix: Fair enough then 👍

smnandre:
Ho yes one thing. Would the LICENCE be published with these settings? Or do we need to duplicate the repository one in the assets directory ?

Kocal:
We already have [`"license": "MIT"`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#license) defined in `package.json` files, but it's true that's make sense to ship the LICENSE file in javascript packages. 

I'm on it

**EDIT:** it means that we also need a `README.md` file for each packages aswell

Kocal:
Files `README.md` and `LICENSE` added!

Kocal:
`@symfony/ux-map` will need a special treatment, since bridges dist files already include the `dist/abstract_map_controller.js` from `@symfony/ux-map`, it means that there is no need to publish it as a npm package. 

For the same reasons, installing `symfony/ux-map` with composer should not update the `package.json`, which is the case today: 
```json
        "@symfony/ux-leaflet-map": "file:vendor/symfony/ux-leaflet-map/assets",
        "@symfony/ux-map": "file:vendor/symfony/ux-map/assets",
```
**EDIT:** see #2611

Kocal:
Added `"private": true` to `src/Map/assets/package.json`, this way we can easily exclude Map package from `yarn workspaces ...` commands with the help of `--no-private`. 

Next steps:
1. Check PHP / JS packages versions sync 
1. Ensure all packages can be installed correctly from npm (`@symfony/ux-leaflet-map` works fine) manually and automatically with a CI workflow
2. Deploy the packages on npm 
3. Ensure all pacakges can be installed correctly from jsdelivr 
4. Write a CI workflow to automatically bump versions (of all packages) and publish packages (except private ones)

PR #2597, opened 2025-02-24
[Doc][StimulusBundle] Misc doc fixes

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | -
| License       | MIT

This PR is mostly to fix this minor syntax issue in the doc intro:

But, I also propose to remove the list of UX packages because this is very hard to maintain. For example, we're missing a lot of the packages added recently. So, I think it's better to just add a link to the always up-to-date listing on the website.


Kocal (approved):

Kocal:
Thanks Javier!

PR #2602, opened 2025-02-25
[LiveComponent][ComponentWithFormTrait] Force recreation of FormView on PreReRender hook to synchronize FormInstance and FormView.

### LINKED ISSUE #1981: [LiveComponent] How do I manually add a FormError to the Form when my submit-function fails?
Let me preface by saying "yea I know this aint probably the right place to ask", but I am fairly certain this problem is unwanted behaviour...but I'd be glad to be proven wrong.

The LiveComponent would be pretty simple:
```php

#[AsLiveComponent(template: 'some_template.html.twig')]
class SomeLiveComponent extends AbstractController
{
     use DefaultActionTrait, ComponentWithFormTrait;
    
     {
          return $this-createForm(SomeFormType::class);
     }

     #[LiveAction]
     public function save(): ?Response
     {
          try {
               /// ... do some persistence 
          } catch (\Throwable $e) {
               $error = new FormError('An error occurred');
               $this->form->addError($error); // <--- How would I display this error in my template?
          }
     }
```

The template, as simple as can be:
```twig
{# some_template.html.twig #}

{{ form_start(form, 
     {
          'attr': {
               'data-action': 'live#action:prevent',
               'data-live-action-param': 'save'
          }
     })
}}

{{ form_errors(form) }} <!-- The Error should be displayed here -->
{{ form_row(form.someField) }}

<button type="submit" data-loading="action(save)|addAttribute(disabled)">Submit</button>

{{ form_end(form) }}
````

A somewhat ugly workaround would be adding a $formErrors array property to the component and populating it with whatever error-messages occur. You could then simply display all those messages in the template, though due to how detached that is from the normal symfony form workflow I would rather avoid this.

I understand that throwing a UnprocessableEntityHttpException() makes turbo re-render the HTML it receives in the response, so the "save()" function should throw that whenever it fails. But the errors are not added to the response HTML anyway - either because the HTML has already been rendered or the FormErrors are removed manually...though I have no idea

I tried adding a writable LiveProp array property to my component which would be populated with FormErrors throughout the submit process - similar to the hacky $formErrors array property mentioned above. But the "instantiateForm()" function would add these errors manually to the form - but that does not work either.

Any pointers would be helpful.

Mauriceu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1981
| License       | MIT

ComponentWithFormTrait will now recreate the existing FormView in its PreReRender-Hook `submitFormOnRender()` to synchronize FormView and FormInstance to reflect any manual changes to the FormInstance.
This primarily solves the Problem of manually added FormErrors not being rendered.

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -110,12 +110,12 @@ trait ComponentWithFormTrait
     {
         if ($this->shouldAutoSubmitForm) {
             $this->submitForm($this->isValidated);
+        } else {
+            // Recreate the FormView because it has been created by the submitForm() with a FormInterface whose values may
+            // have changed. Basically synchronizes FormView and FormInstance to reflect all manual changes made to the
+            // latter between form submit and the components re-render.
+            $this->getFormView(true);
         }
-
-        // Recreate the FormView because it has been created by the submitForm() with a FormInterface whose values may
-        // have changed. Basically synchronizes FormView and FormInstance to reflect all manual changes made to the
-        // latter between form submit and the components re-render.
-        $this->getFormView(true);
     }
 
     public function getFormView(bool $forceCreate = false): FormView
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -111,11 +111,16 @@ trait ComponentWithFormTrait
         if ($this->shouldAutoSubmitForm) {
             $this->submitForm($this->isValidated);
         }
+
+        // Recreate the FormView because it has been created by the submitForm() with a FormInterface whose values may
+        // have changed. Basically synchronizes FormView and FormInstance to reflect all manual changes made to the
+        // latter between form submit and the components re-render.
+        $this->getFormView(true);
     }
 
-    public function getFormView(): FormView
+    public function getFormView(bool $forceCreate = false): FormView
     {
-        if (null === $this->formView) {
+        if ($forceCreate || null === $this->formView) {
             $this->formView = $this->getForm()->createView();
             $this->useNameAttributesAsModelName();
         }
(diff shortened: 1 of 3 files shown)


Mauriceu:
Recreating the FormView may lead to unwanted behavior regarding the `isValidated` and `validatedFields` properties, as they will be set according to the old FormView. However, an issue may only arise when the underlying Code manually changes the value of a given Form/FormChild, which is discouraged anyway after its been submitted.

Also, for performance reasons I would like to avoid adding a second validation.

Kocal:
It makes sense to me, WDYT @smnandre?

smnandre:
I'm not 100% sure to be honest, looks to me this could have side effects, especially with recent changes. 

I will have a deeper look this week-end.

Also tests here focus only on errors, and not on data. But the added comment states "Recreate the FormView because it has been created by the submitForm() with a FormInterface whose values may have changed".

Mauriceu:
> Also tests here focus only on errors, and not on data.

The data in this case would be the string `manually added form error` existing in the template, as it was added by the LiveAction.
But I'd gladly add more coverage as needed. This is my first official pull request so I am still getting the hang of it.

smnandre:
> This is my first official pull request so I am still getting the hang of it.

Welcome aboard :) 

> The data **in this case** would be the string manually added form error existing in the template

Yes but this change will be executed for every component using this trait, on every re-render.. so there may be other cases :)

Mauriceu:
Thank you! Finally able to pay some of it back \o/ 

The added change will force the recreation of the formView only in cases where the `shouldAutoSubmitForm` property is set to `false`.
As I understood that could only happen in one of two cases:
1. `submitForm()` was called before the PreReRender hook is invoked.
2.  Manually setting the `shouldAutoSubmitForm` property to `false`.

If necessary due to performance implications we could always introduce a function `recreateFormView` to give the User more control.

Another thought: Is it even necessary to create the formView inside the `submitForm()` function? We could extract the formValues directly from the instance, and create the formView last when the PreReRender hook is invoked.

Mauriceu:
Could theoretically potentially maybe also create a wrapper class for the FormInterface tracking all changes. However, I feel like that creates way too much overhead for what this PR is actually trying to solve...manual re-creation of the form is probably the option with the least impact on performance

smnandre:
I'd need another opinion here, i'm very hesitating :)

Kocal:

Thanks!

Status: Needs Work

PR #2607, opened 2025-02-26
[LiveComponent] Fix `PropertyTypeExtractorInterface::getTypes()` deprecation, use TypeInfo ^7.2 `Type`

### LINKED ISSUE #1708: [LiveComponent] Transition to TypeInfo (fix 7.1 deprecations)
LiveComponent CI has failures on PHP 8.3 / high dependency

Symfony 7.1 introduces a new [TypeInfo component](https://github.com/symfony/type-info) that will partially replace PropertyInfo, and triggers a lot of deprecations.

We use this in two places : LivePropMetadata(/Factory) and in the Hydrate/Dehydate methods

Original PR: https://github.com/symfony/symfony/pull/52510
Documentation (wip): https://github.com/symfony/symfony-docs/pull/19554 

That means that we must find a way to handle both for a time, so any good idea very welcome :) 

CI Fail examples here: https://github.com/symfony/ux/actions/runs/8613628361/job/23605334525#step:7:122 

```console
Remaining direct deprecation notices (2793)

  705x: Since symfony/property-info 7.1: The "Symfony\Component\PropertyInfo\PropertyInfoExtractor::getTypes()" method is deprecated, use "Symfony\Component\PropertyInfo\PropertyInfoExtractor::getType()" instead.
    126x in LiveComponentHydratorTest::testCanDehydrateAndHydrateComponentWithTestCases from Symfony\UX\LiveComponent\Tests\Integration
    110x in QueryStringPropsExtractorTest::testExtract from Symfony\UX\LiveComponent\Tests\Functional\Util
    102x in LiveComponentHydratorTest::testCoerceFalseyValuesForScalarTypes from Symfony\UX\LiveComponent\Tests\Integration
    30x in LiveComponentHydratorTest::testCoerceTruthyValuesForScalarTypes from Symfony\UX\LiveComponent\Tests\Integration
    20x in ComponentWithFormTest::testHandleCheckboxChanges from Symfony\UX\LiveComponent\Tests\Functional\Form
    ...

  705x: Since symfony/property-info 7.1: The "Symfony\Component\PropertyInfo\Extractor\PhpStanExtractor::getTypes()" method is deprecated, use "Symfony\Component\PropertyInfo\Extractor\PhpStanExtractor::getType()" instead.
    126x in LiveComponentHydratorTest::testCanDehydrateAndHydrateComponentWithTestCases from Symfony\UX\LiveComponent\Tests\Integration
    110x in QueryStringPropsExtractorTest::testExtract from Symfony\UX\LiveComponent\Tests\Functional\Util
    102x in LiveComponentHydratorTest::testCoerceFalseyValuesForScalarTypes from Symfony\UX\LiveComponent\Tests\Integration
    30x in LiveComponentHydratorTest::testCoerceTruthyValuesForScalarTypes from Symfony\UX\LiveComponent\Tests\Integration
    20x in ComponentWithFormTest::testHandleCheckboxChanges from Symfony\UX\LiveComponent\Tests\Functional\Form
    ...

  691x: Since symfony/property-info 7.1: The "Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor::getTypes()" method is deprecated, use "Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor::getType()" instead.
    112x in LiveComponentHydratorTest::testCanDehydrateAndHydrateComponentWithTestCases from Symfony\UX\LiveComponent\Tests\Integration
    110x in QueryStringPropsExtractorTest::testExtract from Symfony\UX\LiveComponent\Tests\Functional\Util
    102x in LiveComponentHydratorTest::testCoerceFalseyValuesForScalarTypes from Symfony\UX\LiveComponent\Tests\Integration
    30x in LiveComponentHydratorTest::testCoerceTruthyValuesForScalarTypes from Symfony\UX\LiveComponent\Tests\Integration
    20x in ComponentWithFormTest::testHandleCheckboxChanges from Symfony\UX\LiveComponent\Tests\Functional\Form
    ...
```

If anyone has some time / motivation to look at this ❤️  ?

### LINKED ISSUE #2778: [LiveComponent] PropertyInfoExtractor::getTypes() method is deprecated
Triggered by:
https://github.com/symfony/ux/blob/59e840d825d9a095ad74b1e54b51c5eb9267e489/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php#L82

Probably needs a fix like this:
https://github.com/symfony/symfony/blob/5add177fd59074401f7a1d8d65d807779ce84b86/src/Symfony/Component/Validator/Mapping/Loader/PropertyInfoLoader.php#L192-L196

mtarld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Fix #1708, fix #2778
| License       | MIT

Fix the `PropertyTypeExtractorInterface::getTypes()` deprecation and use `PropertyTypeExtractorInterface::getType()` (based on TypeInfo `Type`) instead. 

This PR allows TypeInfo ^7.2 to be used, thanks to the compatibility layers added for `Type::accepts()` and `Type::traverse()` (added in 7.3). In order to reduce frictions when users will upgrade their apps dependencies.

The CI changed a bit too:
- When testing LiveComponent, with PHP 8.2, we install `symfony/property-info:7.1.* symfony/type-info:7.2.*`
- When testing LiveComponent, with PHP 8.3, we install `symfony/property-info:7.2.* symfony/type-info:7.2.*`
- When testing LiveComponent, with PHP 8.4, we install `symfony/property-info:7.3.* symfony/type-info:7.3.*`
- When testing LiveComponent, with PHP 8.4 (and dev deps), we install `symfony/property-info:>=7.3 symfony/type-info:>=7.3`

Allowing us to covers a maximum versions of PropertyInfo and TypeInfo.

### DIFF
--- a/src/LiveComponent/composer.json
+++ b/src/LiveComponent/composer.json
@@ -56,7 +56,9 @@
         "zenstruck/foundry": "^2.0"
     },
     "conflict": {
-        "symfony/config": "<5.4.0"
+        "symfony/config": "<5.4.0",
+        "symfony/type-info": "<7.1",
+        "symfony/property-info": "~7.0.0"
     },
     "config": {
         "sort-packages": true
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -18,19 +18,26 @@ use Symfony\Component\PropertyAccess\Exception\UninitializedPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractor;
-use Symfony\Component\PropertyInfo\Type;
+use Symfony\Component\PropertyInfo\Type as LegacyType;
 use Symfony\Component\Serializer\Exception\ExceptionInterface as SerializerExceptionInterface;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\TypeIdentifier;
+use Symfony\Component\TypeInfo\Type\CollectionType;
+use Symfony\Component\TypeInfo\Type\ObjectType;
+use Symfony\Component\TypeInfo\Type\WrappingTypeInterface;
 use Symfony\Component\Uid\AbstractUid;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
 use Symfony\UX\LiveComponent\Exception\HydrationException;
 use Symfony\UX\LiveComponent\Hydration\HydrationExtensionInterface;
+use Symfony\UX\LiveComponent\Metadata\LegacyLivePropMetadata;
 use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadata;
 use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
 use Symfony\UX\LiveComponent\Metadata\LivePropMetadata;
 use Symfony\UX\LiveComponent\Util\DehydratedProps;
+use Symfony\UX\LiveComponent\Util\TypeHelper;
 use Symfony\UX\TwigComponent\ComponentAttributes;
 use Twig\Environment;
 use Twig\Runtime\EscaperRuntime;
@@ -249,7 +256,7 @@ final class LiveComponentHydrator
      *
      * @throws SerializerExceptionInterface
      */
-    public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, object $parentObject): mixed
+    public function hydrateValue(mixed $value, LivePropMetadata|LegacyLivePropMetadata $propMetadata, object $parentObject): mixed
     {
         if ($propMetadata->hydrateMethod()) {
             if (!method_exists($parentObject, $propMetadata->hydrateMethod())) {
@@ -270,37 +277,65 @@ final class LiveComponentHydrator
                 throw new \LogicException(\sprintf('The LiveProp "%s" on component "%s" has "useSerializerForHydration: true", but the given serializer does not implement DenormalizerInterface.', $propMetadata->getName(), $parentObject::class));
             }
 
-            if ($propMetadata->collectionValueType()) {
-                $builtInType = $propMetadata->collectionValueType()->getBuiltinType();
-                if (Type::BUILTIN_TYPE_OBJECT === $builtInType) {
-                    $type = $propMetadata->collectionValueType()->getClassName().'[]';
+            // BC layer when "symfony/type-info" is not available
+            if ($propMetadata instanceof LegacyLivePropMetadata) {
+                if ($propMetadata->collectionValueType()) {
+                    $builtInType = $propMetadata->collectionValueType()->getBuiltinType();
+                    if (LegacyType::BUILTIN_TYPE_OBJECT === $builtInType) {
+                        $type = $propMetadata->collectionValueType()->getClassName().'[]';
+                    } else {
+                        $type = $builtInType.'[]';
+                    }
                 } else {
-                    $type = $builtInType.'[]';
+                    $type = $propMetadata->getType();
                 }
+
+                if (null === $type) {
+                    throw new \LogicException(\sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
+                }
+
+                if (null === $type) {
+                    throw new \LogicException(\sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
+                }
+
+                if (null === $value && $propMetadata->allowsNull()) {
+                    return null;
+                }
+
+                return $this->serializer->denormalize($value, $type, 'json', $propMetadata->serializationContext());
             } else {
                 $type = $propMetadata->getType();
-            }
 
-            if (null === $type) {
-                throw new \LogicException(\sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
-            }
+                if (null === $type) {
+                    throw new \LogicException(\sprintf('The "%s::%s" object should be hydrated with the Serializer, but no type could be guessed.', $parentObject::class, $propMetadata->getName()));
+                }
 
-            if (null === $value && $propMetadata->allowsNull()) {
-                return null;
-            }
+                if (null === $value && $type->isNullable()) {
+                    return null;
+                }
 
-            return $this->serializer->denormalize($value, $type, 'json', $propMetadata->serializationContext());
-        }
+                $isCollection = false;
+                foreach (TypeHelper::traverse($type) as $t) {
+                    if ($t instanceof CollectionType) {
+                        $isCollection = true;
+                        $type = $t->getCollectionValueType();
 
-        if ($propMetadata->collectionValueType() && Type::BUILTIN_TYPE_OBJECT === $propMetadata->collectionValueType()->getBuiltinType()) {
-            $collectionClass = $propMetadata->collectionValueType()->getClassName();
-            foreach ($value as $key => $objectItem) {
-                $value[$key] = $this->hydrateObjectValue($objectItem, $collectionClass, true, $propMetadata->getFormat(), $parentObject::class, \sprintf('%s.%s', $propMetadata->getName(), $key), $parentObject);
+                        break;
+                    }
+                }
+
+                while ($type instanceof WrappingTypeInterface) {
+                    $type = $type->getWrappedType();
+                }
+
+                $typeString = $type.($isCollection ? '[]' : '');
+
+                return $this->serializer->denormalize($value, $typeString, 'json', $propMetadata->serializationContext());
             }
         }
 
         // no type? no hydration
-        if (!$propMetadata->getType()) {
+        if (!$type = $propMetadata->getType()) {
             return $value;
         }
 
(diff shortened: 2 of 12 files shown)


mtarld:
Actually, didn't notice that PHP 8.1 is still supported by Symfony UX.
Unfortunately, it'll prevent `symfony/type-info` to be installed in that version.

src/LiveComponent/composer.json
```diff
@@ -29,7 +29,9 @@
         "php": ">=8.1",
         "symfony/deprecation-contracts": "^2.5|^3.0",
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
+        "symfony/property-info": "^5.4|^6.0|^7.0",
```
mtarld: This package is a hard dependency, so it has to be moved to the `require` section.

mtarld:
Anyway, I think we must wait for the bump of PHP in `composer.json` (ie: `symfony/ux-live-component:^3.0`).

Kocal:
Added milestone [3.x](https://github.com/symfony/ux/milestone/1)

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -188,7 +188,7 @@ function (ChildDefinition $definition, AsLiveComponent $attribute) {
         $container->register('ux.live_component.metadata_factory', LiveComponentMetadataFactory::class)
             ->setArguments([
                 new Reference('ux.twig_component.component_factory'),
-                new Reference('property_info'),
+                new Reference('type_info.resolver'),
```
stof: this service only exists in symfony/framework-bundle 7.1+ (when symfony/type-info is installed, but this will be the case thanks to your dependency). Things will break when using Symfony 6.4 LTS.

Kocal:
@mtarld Sorry to bring this up again, but are we sure it's not possible at all to ship this PR for 2.x? 
Like, if we remove the `symfony/type-info` from require-dev, and tweak the CI to install it when PHP >=8.2?

mtarld:
The UX LiveComponent is requiring a too low PHP version (lower than TypeInfo), which means that a `composer install` in low deps will fail. Therefore we can't have `symfony/type-info` in the `composer.json` file.

As discussed with @Kocal , the solution for 2.x will be: 
- Telling in the documentation to run `composer require symfony/type-info` in order to kill the deprecation (if they're using a lower version than 7.3, it won't be any deprecation).
- Do the same in the CI. 
- Make sure that the code work for either PropertyInfo and TypeInfo.

Then for 3.x: 
- Remove the BC layer for PropertyInfo's Type
- Bump the minimum PHP version
- Require `symfony/type-info` in `composer.json`

I'll update the PR in that way ASAP :slightly_smiling_face:

smnandre:
> Do the same in the CI.

I'd like to keep the CI testing the actual behaviour users would have with 8.1 🤷 

@mtarld poke me if you want some hand there :)

Kocal:
> I'd like to keep the CI testing the actual behaviour users would have with 8.1 🤷

It won't change for 8.1, since symfony/type-info is not installable, but it will change for higher jobs

mtarld:
Status: Needs Work

src/LiveComponent/composer.json
```diff
@@ -56,7 +56,9 @@
         "zenstruck/foundry": "^2.0"
     },
     "conflict": {
-        "symfony/config": "<5.4.0"
+        "symfony/config": "<5.4.0",
+        "symfony/type-info": "<7.3",
+        "symfony/property-info": "~7.0"
```
Kocal: Otherwise we can't install `symfony/property-info:>=7.1`
```suggestion
        "symfony/property-info": "~7.0.0"
```

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -493,14 +607,26 @@ private function dehydrateValue(mixed $value, LivePropMetadata $propMetadata, ob
...
-        // at this point, we have an object and can assume $propMetadata->getType()
-        // is set correctly (needed for hydration later)
+            return $this->dehydrateObjectValue($value, $propMetadata->getType(), $propMetadata->getFormat(), $parentObject);
+        } else {
+            foreach ($propMetadata->getType()?->traverse() ?? [] as $t) {
+                if ($t instanceof ObjectType) {
+                    return $this->dehydrateObjectValue($value, $t->getClassName(), $propMetadata->getFormat(), $parentObject);
+                }
+            }
+            dd($propMetadata);
```
Kocal: To remove

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -72,43 +75,54 @@ public function createPropMetadatas(\ReflectionClass $class): array
...
 
-    public function createLivePropMetadata(string $className, string $propertyName, \ReflectionProperty $property, LiveProp $liveProp): LivePropMetadata
+    public function createLivePropMetadata(string $className, string $propertyName, \ReflectionProperty $property, LiveProp $liveProp): LivePropMetadata|LegacyLivePropMetadata
     {
-        $type = $property->getType();
-        if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
-            throw new \LogicException(\sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
-        }
+        // BC layer when "symfony/type-info" is not available
+        if (!method_exists($this->propertyTypeExtractor, 'getType')) {
```
Kocal, line 81: Method `$this->propertyTypeExtractor->getTypes()` triggers a deprecation explaining to use `getType()`, which is a bit useless for the end-user since `getTypes()` is internally called by us.

At this point, can we add a deprecation for PHP 8.2+ to tell the user to install `symfony/property-info:^7.1` (which implements `getType()` and ships with `symfony/type-info`)?

src/LiveComponent/src/LiveComponentHydrator.php
Kocal: Added to the mental list: automatically check for undesired functions call (dd, dump, ...)

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
Kocal: Hum, I don't see how we can do it correctly, the deprecation on `getTypes()` has been added in `symfony/property-info:^7.3`. 

I guess the current deprecation and a changelog entry could be enough

src/LiveComponent/composer.json
Kocal: Fixed

src/LiveComponent/composer.json
```diff
@@ -56,7 +56,9 @@
         "zenstruck/foundry": "^2.0"
     },
     "conflict": {
-        "symfony/config": "<5.4.0"
+        "symfony/config": "<5.4.0",
+        "symfony/type-info": "<7.3",
```
Kocal: Can we remove conflict on `symfony/type-info:<7.3`? It makes some dependencies impossible to install and the CI fail: 
```
  Problem 1
    - Root composer.json requires symfony/form 7.2.* -> satisfiable by symfony/form[v7.2.0-BETA1, ..., 7.2.x-dev].
    - Root composer.json requires symfony/ux-live-component @dev -> satisfiable by symfony/ux-live-component[dev-a0370b125d1542c57866ab89fd64ce80cd316df6].
    - symfony/form[v7.2.0-BETA1, ..., 7.2.x-dev] require symfony/property-access ^6.4|^7.0 -> satisfiable by symfony/property-access[v7.2.0-BETA1, ..., 7.2.x-dev].
    - symfony/property-access[v7.2.0-BETA1, ..., 7.2.x-dev] require symfony/property-info ^6.4|^7.0 -> satisfiable by symfony/property-info[v7.2.0-BETA1, ..., 7.2.x-dev].
    - symfony/property-info v7.2.0-RC1 requires symfony/type-info ^7.2 -> satisfiable by symfony/type-info[v7.2.0-BETA1, ..., 7.2.x-dev].
    - symfony/property-info[v7.2.0-BETA1, ..., v7.2.1] require symfony/type-info ^7.1 -> satisfiable by symfony/type-info[v7.2.0-BETA1, ..., 7.2.x-dev].
    - symfony/property-info[v7.2.2, ..., 7.2.x-dev] require symfony/type-info ~7.1.9|^7.2.2 -> satisfiable by symfony/type-info[v7.2.2, v7.2.4, v7.2.5, 7.2.x-dev].
    - symfony/ux-live-component dev-a0370b125d1542c57866ab89fd64ce80cd316df6 conflicts with symfony/type-info v7.2.5.
    - symfony/ux-live-component dev-a0370b125d1542c57866ab89fd64ce80cd316df6 conflicts with symfony/type-info v7.2.0.
    - symfony/ux-live-component dev-a0370b125d1542c57866ab89fd64ce80cd316df6 conflicts with symfony/type-info v7.2.0-RC1.
    - symfony/ux-live-component dev-a0370b125d1542c57866ab89fd64ce80cd316df6 conflicts with symfony/type-info v7.2.0-BETA1.
```

mtarld: I'll have to update the code a bit but yes!

Kocal: You told me that 7.3 is _only_ for `Type::traverse()`. Since it's used in many places, I think we can introduce a small helper to traverse types if `Type::traverse()` does not exist.

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -270,54 +276,125 @@ public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, objec
...
-                return null;
-            }
+                if (null === $value && $type->isNullable()) {
+                    return null;
+                }
 
-            return $this->serializer->denormalize($value, $type, 'json', $propMetadata->serializationContext());
-        }
+                $isCollection = false;
+                foreach ($type->traverse() as $t) {
```
Kocal: `$t` is unused

src/LiveComponent/src/LiveComponentHydrator.php
```diff
@@ -270,54 +276,125 @@ public function hydrateValue(mixed $value, LivePropMetadata $propMetadata, objec
...
+                    if ($isCollection = $type instanceof CollectionType) {
+                        $isCollection = true;
```
Kocal:
```suggestion
                    if ($type instanceof CollectionType) {
                        $isCollection = true;
```

no?

src/LiveComponent/composer.json
Kocal: (added)

Kocal:
CI is fully green 🤩 (we can ignore fabbot here)

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
Kocal: changelog entry added

Kocal:
I rebased the PR but didn't squash my commits (if it can ease reviews), CI is full green (we can ignore fabbot). 

After reviews, we can merge this PR for UX **2.26**, since the initial PR changed from "only use TypeInfo" to "use TypeInfo is possible" 🎉

Kocal:
I rebased the PR but didn't squash my commits (if it can ease reviews), CI is full green (we can ignore fabbot). 

After reviews, we can merge this PR for UX **2.x** instead of **3.x**, since the initial PR changed from "only use TypeInfo" to "use TypeInfo when possible" 🎉

src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
Kocal: Fixed

kbond (approved):
Wow, nice work @mtarld & @Kocal! I didn't think we'd see the CI green again in 2.x!

smnandre:
Wow so nice!

Kocal:
I've just pushed a new commit for documentation, about upgrading `symfony/property-info:^7.1` to get rid of deprecations

EDIT: reverted, in fact it does not makes sense (the deprecation will be removed simply by upgrading the LiveComponent package).

Kocal:
After discussing with @mtarld, we've just increased `symfony/type-info` conflict to `<7.2`:
- the previous conflict was wrong
- a lot of breaking changes happened in TypeInfo 7.2 (since 7.1 was experimental)

smnandre:
🙇 

Not enough time to make a deep review, but a huge thanks to you guys. 👏

smnandre (approved):
❤️

Kocal:
Thanks Mathias. 🙌🏻

norkunas:
This introduced some sort of BC break.

I have this property on component:
```php
    /**
     * @var array{r: array<string, true>, c: array<string, true>}
     */
    #[LiveProp(url: new UrlMapping('e'))]
    public array $expanded = ['r' => [], 'c' => []];
```

My url is `/tree?expanded[r][C-D]=1` and now it does not properly hydrate because `TypeHelper::accepts($type, $value)` in `QueryStringPropsExtractor::isValueTypeConsistent` return false.

 The problem I see is that both `r` and `c` in `$expanded` are set as `optional: false`:
```php
Symfony\Component\TypeInfo\Type\ArrayShapeType {#5071 ▼
  -type: Symfony\Component\TypeInfo\Type\GenericType {#5157 ▼
    -variableTypes: array:2 [▼
      0 => Symfony\Component\TypeInfo\Type\BuiltinType {#5159 ▼
        -typeIdentifier: Symfony\Component\TypeInfo\TypeIdentifier {#5052 ▶}
      }
      1 => Symfony\Component\TypeInfo\Type\CollectionType {#5068 ▶}
    ]
    -type: Symfony\Component\TypeInfo\Type\BuiltinType {#5158 ▶}
  }
  -isList: false
  -shape: array:2 [▼
    "c" => array:2 [▼
      "type" => Symfony\Component\TypeInfo\Type\CollectionType {#5070 ▶}
      "optional" => false
    ]
    "r" => array:2 [▼
      "type" => Symfony\Component\TypeInfo\Type\CollectionType {#5068 ▶}
      "optional" => false
    ]
  ]
  -extraKeyType: null
  -extraValueType: null
}
```

So when it tries to set only `r` part this shape is not accepted anymore. This was previously working fine.
The functionality is simple:
 - I render a tree;
 - Clicking on first level entries saves state to `$expanded = ['r' => ...]`;
 - Clicking on second level entries save state to `$expanded = ['c' => ...]`;
 - Clicking is handling via two live actions which updates only `r` or `c`.

--

Tried to split into multiple properties:

```php
    /**
     * @var array<string, true>
     */
    #[LiveProp(url: new UrlMapping('r'))]
    public array $expandedRoots = [];

    /**
     * @var array<string, true>
     */
    #[LiveProp(url: new UrlMapping('c'))]
    public array $expandedCategories = [];
```

still `TypeHelper::accepts` returns false on url `/tree?r[A-B]=1&c[01713ef6-eced-7e20-ab12-008458caf697]=1`

Kocal:
To me it looks like the actual behavior may be correct... what about using `array{r?: array<string, true>, c?: array<string, true>}` instead? so `r` and `c` could really be seen as optional.

norkunas:
Well then I'll need to add checks everywhere to account for nullability :) better would be then switch to multiple props, but it still does not work then. I've amended my previous comment with that

PR #2610, opened 2025-02-28
[LiveComponent] Document how to pass a value to an event listener from Twig.

Artusamak:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | -
| License       | MIT

The document is missing an important use case related to Event dispatching and listeners.
There is no documentation about how to pass data from a Twig template to an event listener method within the component. This PR adds an example.


Kocal (approved):

Kocal:
Thank you @Artusamak.

PR #2615, opened 2025-03-02
[CI] Add workflow to automatically publish JS packages on npm

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Follow #2614

### DIFF
--- a/.github/workflows/release-on-npm.yaml
+++ b/.github/workflows/release-on-npm.yaml
@@ -1,0 +1,50 @@
+name: Release on NPM
+
+on:
+  push:
+    tags:
+        - 'v*.*.*'
+
+jobs:
+    release:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-on-npm.yaml
```diff
@@ -0,0 +1,50 @@
+name: Release on NPM
+
+on:
+  push:
+    tags:
+        - 'v*.*.*'
```
stof, line 6: shouldn't this be only for tags `v2.*.*` as you checkout the 2.x branch ?

Kocal: That's true, we can be more precise to prevent any issues

Issue #691, opened 2023-02-18
NPM cannot install, vendor not available in seperate javascript/css building pipeline.


creatiombe:
For the lazy loader image ux plugin (and probably others too, didn't checkout other components):

I use npm standalone without composer and cannot install this package. At least the javascript should be available as a standalone package as i want to be able to run composer and npm in separate docker building runners for each their own, to speed up deployments. Using the vendor path in package.json is not an option for me.

Please give me a solution so i can still use a separate pipeline builder for composer and npm.

If you are really serious about the 'ux' initiative, you should really be serious about splitting up php code to work separately from javascript in my personal opinion.

Using the flex recipe system to adjust the package.json to use a local path should not be promoted, but rather a real npm install package should be available.

creatiombe:
If you are really serious about the 'ux' initiative, you should really be serious about splitting up php code to work seperatly from javascript in my personal opinion.

weaverryan:
> If you are really serious about the 'ux' initiative, you should really be serious about splitting up php code to work separately from javascript in my personal opinion.

I tend to agree, though the current model is dead simple and works quite well (the one downside your mentioned being a legitimate downside, however).

The tricky thing would be matching up the JS and PHP package versions - e.g. if we release 3.0 of the PHP package, it may not work with 2.* of the JS package. There's no mechanism (that I can think of) to warn the user that their JS package version will no longer work. However, in practice, Flex might eliminate this problem because when you run `composer install`, if any of your JS dependencies have too low of a version, the version in `package.json` is updated. So you'd have to either not be using Flex, or willfully ignore that :).

Anyways, I think this is a cool idea.

creatiombe:
Yeah like so many things i am now using the swup plugin from the ux initiative. Next problem: I want to use shared element transitions: 

https://css-tricks.com/still-hoping-for-better-native-page-transitions/

https://github.com/daun/swup-morph-plugin

But again the plugins are controlled in a way that i cannot extend it, controller.json inside VENDOR ux-swup dist folder. So now i have to hack this module completely up in pieces and manually put it inside this symfony site i am working on.

weaverryan:
Hi!

There are events you can use to extend the core behavior - https://symfony.com/bundles/ux-swup/current/index.html#extend-the-default-behavior - the pre connect event is probably what you want so you can hook in custom stuff. But let us know if that extension point won’t work :)

creatiombe:
[...]

PR #2617, opened 2025-03-02
Document about alternative JS assets installation with npm packages

### LINKED ISSUE #2575: [Stimulus] Create NPM Package for @symfony/stimulus-bundle to Decouple from PHP Components

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #2575  
| License       | MIT

Follow #2616

### DIFF
--- a/src/Autocomplete/doc/index.rst
+++ b/src/Autocomplete/doc/index.rst
@@ -30,6 +30,10 @@ needed if you're using AssetMapper):
     $ npm install --force
     $ npm run watch
 
+.. note::
+
+    For more complex installation scenarios, you can install the JavaScript assets through the `@symfony/ux-autocomplete npm package`_
+
 Usage in a Form (without Ajax)
 ------------------------------
 
@@ -288,7 +292,7 @@ Passing Extra Options to the Ajax-powered Autocomplete
 
 Autocomplete field options are **not preserved** when the field is rendered
 on an Ajax call. So, features like exclude some options based on the current
-form data are not possible by default. 
+form data are not possible by default.
 
 To partially avoid this limitation, the ``extra_options`` option was added.
 
@@ -299,7 +303,7 @@ To partially avoid this limitation, the ``extra_options`` option was added.
     can be passed as extra options.
 
 Considering the following example, when the form type is rendered for the first
-time, it will use the ``query_builder`` defined while adding a ``food`` field 
+time, it will use the ``query_builder`` defined while adding a ``food`` field
 to the ``FoodForm``. However, when the Ajax is used to fetch the results, on
 the consequent renders, the default ``query_builder`` will be used::
 
@@ -324,8 +328,8 @@ the consequent renders, the default ``query_builder`` will be used::
         }
     }
 
-If some food can be consisted of other foods, we might want to exclude the 
-"root" food from the list of available foods. To achieve this, we can remove 
+If some food can be consisted of other foods, we might want to exclude the
+"root" food from the list of available foods. To achieve this, we can remove
 the ``query_builder`` option from the above example and pass the ``excluded_foods``
 extra option to the ``FoodAutocompleteField``::
 
@@ -600,13 +604,13 @@ Passing Extra Options to the Autocompleter
 
     The ability to pass extra options was added in Autocomplete 2.14.
 
-If you need to pass extra options to the autocompleter, you can do so by 
-implementing the ``\Symfony\UX\Autocomplete\OptionsAwareEntityAutocompleterInterface`` 
+If you need to pass extra options to the autocompleter, you can do so by
+implementing the ``\Symfony\UX\Autocomplete\OptionsAwareEntityAutocompleterInterface``
 interface.
 
 .. tip::
 
-    If you want to know **why** you might need to use the ``extra_options`` 
+    If you want to know **why** you might need to use the ``extra_options``
     feature, see :ref:`passing-extra-options-to-the-ajax-powered-autocomplete`.
 
 .. code-block:: diff
@@ -757,3 +761,4 @@ the Symfony framework: https://symfony.com/doc/current/contributing/code/bc.html
 .. _`Tom Select Render Templates`: https://tom-select.js.org/docs/#render-templates
 .. _`Tom Select Option Group`: https://tom-select.js.org/examples/optgroups/
 .. _`Symfony Form`: https://symfony.com/doc/current/forms.html
+.. _`@symfony/ux-autocomplete npm package`: https://www.npmjs.com/package/@symfony/ux-autocomplete
--- a/src/Map/doc/index.rst
+++ b/src/Map/doc/index.rst
@@ -16,14 +16,6 @@ Install the bundle using Composer and Symfony Flex:
 
     $ composer require symfony/ux-map
 
-If you're using WebpackEncore, install your assets and restart Encore (not
-needed if you're using AssetMapper):
-
-.. code-block:: terminal
-
-    $ npm install --force
-    $ npm run watch
-
 Configuration
 -------------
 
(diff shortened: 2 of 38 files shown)


src/Map/doc/index.rst
```diff
@@ -16,14 +16,6 @@ Install the bundle using Composer and Symfony Flex:
...
     $ composer require symfony/ux-map
 
-If you're using WebpackEncore, install your assets and restart Encore (not
-needed if you're using AssetMapper):
-
-.. code-block:: terminal
-
-    $ npm install --force
-    $ npm run watch
-
```
Kocal, line 26: There is no assets for `symfony/ux-map`, so let's remove this confusing part.

javiereguiluz (commented):
Teh way I see this is that we now have 3 different ways of installing this:

**(1)** PHP-based and uses AssetMapper

```
.. code-block:: terminal

    $ composer require symfony/ux-cropperjs
```

**(2)** PHP-based and uses WebpackEncore

```
.. code-block:: terminal

    $ composer require symfony/ux-cropperjs
    $ npm install --force
    $ npm run watch
```

**(3)** JavaScript-based

```
.. code-block:: terminal

    $ npm add @symfony/ux-cropperjs package
```

-----

My questions:

* Is (3) totally equivalent to (1) and (2)
* If not, in which scenarios would someone use (3) vs (1) or (2)?

Based on the answers to these questions, let's think about the best possible way of showing this information. Thanks.

Kocal:
(3) alone is not possible, each UX packages on npm (e.g.: https://www.npmjs.com/package/@symfony/ux-chartjs, https://www.npmjs.com/package/@symfony/ux-translator, ...) mention that the npm package **must** also install the php package.

See #2575 an #691 for more information, but to resume, having standalone JavaScript packages would allow users to install their JavaScript dependencies without needing PHP/Composer, which can be pretty useful in a CI (building Docker images, deploying an app, running JS tests, ...).

--- 

I would adapt your examples like this:

**(1)** PHP-based and uses AssetMapper

```shell
$ composer require symfony/ux-cropperjs
```

**(2)** PHP-based and uses WebpackEncore

```shell
$ composer require symfony/ux-cropperjs
$ # if user prefers to keep things separated: npm add -D @symfony/ux-cropperjs  
$ npm install --force
$ npm run watch
```
--- 

What do you think about that?

> Based on the answers to these questions, let's think about the best possible way of showing this information. Thanks.

This is exactly why I added you as a reviewer 🤗

javiereguiluz:
Thanks for the insights. I would expand each installation section as follows:

```rst
Installation
------------

.. caution::

    Before you start, make sure you have `StimulusBundle configured in your app`_.

Run the following command to install this package:

.. code-block:: terminal

    $ composer require symfony/ux-cropperjs

If your Symfony application uses `AssetMapper`_, you don't have to do anything else
and can move on to the next section.

If your Symfony application uses `Webpack Encore`_, run the following commands
to install the assets and restart Encore:

.. code-block:: terminal

    $ npm install --force
    $ npm run watch

.. tip::

    The assets of this package are also provided as the standalone `@symfony/ux-cropperjs`_
    JavaScript package. This is useful, for example, on continuous integration servers
    to build Docker images, run JavaScript tests, etc. without using PHP/Composer.

Usage
-----

[...]

.. _`AssetMapper`: https://symfony.com/doc/current/frontend/asset_mapper.html
.. _`Webpack Encore`: https://symfony.com/doc/current/frontend/encore/index.html
```

Although docs should be concise, I think it's fine to be a bit verbose/detailed in the `Install` section because any error here is very frustrating for readers.

I'd mention the scenarios where the JS package is useful. Experts won't need it, but the rest of folks would need this to decide if this is something for them or not.

----

I have a pending question. In the original text, we say things like:

> _Install the bundle using Composer and Symfony Flex:_

What happens if the reader doesn't have Flex installed? Which steps should be done manually? Thanks

Kocal:
I really like your suggestion, thanks Javier!

--- 

> > _Install the bundle using Composer and Symfony Flex:_
> 
> What happens if the reader doesn't have Flex installed? Which steps should be done manually? Thanks

If Flex is not installed, then the reader must do (unknown) operations manually:
1. Add the bundle in `bundles.php`
2. Do what the recipe does (e.g. with https://github.com/symfony/recipes/blob/main/symfony/ux-vue/2.9/manifest.json (add some lines, copy files, ...)
3. Modify its `package.json` and add the dependency `file:vendor/symfony/ux-<pkg>/assets`
4. and maybe more..

Which is far from being ideal.

smnandre:
> Experts won't need it, but the rest of folks would need this to decide if this is something for them or not.

I would not expose non-experts to manual JS installation for now, as we have not solved the update question as far as I know (@Kocal correct me)

src/Autocomplete/doc/index.rst
```diff
@@ -30,6 +30,10 @@ needed if you're using AssetMapper):
     $ npm install --force
     $ npm run watch
 
+.. note::
+
+    Alternatively, the `@symfony/ux-autocomplete package`_ can be used to install the JavaScript assets without requiring PHP.
```
smnandre: I would explicitely mention "CI" or "containers" here

src/Map/src/Bridge/Google/README.md
```diff
@@ -2,6 +2,25 @@
 
 [Google Maps](https://developers.google.com/maps/documentation/javascript/overview) integration for Symfony UX Map.
 
+## Installation
+
+Install the bundle using Composer and Symfony Flex:
```
smnandre:
```suggestion
Install the bridge using Composer and Symfony Flex:
```

src/Map/src/Bridge/Leaflet/README.md
```diff
@@ -2,6 +2,25 @@
 
 [Leaflet](https://leafletjs.com/) integration for Symfony UX Map.
 
+## Installation
+
+Install the bundle using Composer and Symfony Flex:
```
smnandre:
```suggestion
Install the bridge using Composer and Symfony Flex:
```

src/StimulusBundle/doc/index.rst
```diff
@@ -39,6 +39,10 @@ necessary files. If not, or you're curious, see :ref:`Manual Setup <manual-insta
     If you're using Encore, be sure to install your assets (e.g. ``npm install``)
     and restart Encore.
 
+.. note::
+
+    Alternatively, the `@symfony/stimulus-bundle package`_ can be used to install the JavaScript assets without requiring PHP.
```
smnandre: Are we sure everything works here ? :| This feels suspicious as some path are hard-coded for controller.json

Kocal: I will double check later

Kocal:
@javiereguiluz after some discussions after Simon, I think we will go for the following changes:

1. On `index.rst` documentation files, revert what's been done and add somewhere `for more complex installation scenarios, you can install the assets of symfony/ux-autocomplete from npm` (with a link to https://www.npmjs.com/package/@symfony/ux-autocomplete)
2. On `src/**/assets/README.md` files, the one shown on https://www.npmjs.com/package/@symfony/ux-autocomplete, add more documentation about _when_ you can install this package, how, and with an example about exact versions

Thus, we add documentation where it's necessary, without adding a lot of documentation to already long documents.

Kocal:
After many iterations, I think we are on something really good.

src/StimulusBundle/doc/index.rst
Kocal: Well, you're right, those `dist/` files are only needed when you use the AssetMapper, since we tell users to use npm package `@symfony/stimulus-bridge` when using Encore.

Kocal: I will mark this npm package as private for 2.24, and remove it from npm. I will adapt the `README.md` telling people looks for assets must install `@symfony/stimulus-bridge` if using Encore. 

Let's see in the future if we really need to publish this one.

Thanks!

**EDIT:** the package is not removable anymore, so I marked it as deprecated, that's the best I can do.

smnandre (approved):

PR #2618, opened 2025-03-02
[Turbo] Add Twig Extensions for `meta` tags

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #... 
| License       | MIT

Hi,

I added Twig extensions for `meta` tags:
- `turbo_exempts_page_from_cache`
- `turbo_exempts_page_from_preview`
- `turbo_page_requires_reload`
- `turbo_refreshes_with`
- `turbo_refresh_method`
- `turbo_refresh_scroll`

### DIFF
--- a/src/Turbo/src/Twig/TwigExtension.php
+++ b/src/Turbo/src/Twig/TwigExtension.php
@@ -22,6 +22,12 @@ use Twig\TwigFunction;
  */
 final class TwigExtension extends AbstractExtension
 {
+    private const REFRESH_METHOD_REPLACE = 'replace';
+    private const REFRESH_METHOD_MORPH = 'morph';
+
+    private const REFRESH_SCROLL_RESET = 'reset';
+    private const REFRESH_SCROLL_PRESERVE = 'preserve';
+
     public function __construct(
         private ContainerInterface $turboStreamListenRenderers,
         private string $default,
@@ -32,6 +38,12 @@ final class TwigExtension extends AbstractExtension
     {
         return [
             new TwigFunction('turbo_stream_listen', $this->turboStreamListen(...), ['needs_environment' => true, 'is_safe' => ['html']]),
+            new TwigFunction('turbo_exempts_page_from_cache', $this->turboExemptsPageFromCache(...), ['is_safe' => ['html']]),
+            new TwigFunction('turbo_exempts_page_from_preview', $this->turboExemptsPageFromPreview(...), ['is_safe' => ['html']]),
+            new TwigFunction('turbo_page_requires_reload', $this->turboPageRequiresReload(...), ['is_safe' => ['html']]),
+            new TwigFunction('turbo_refreshes_with', $this->turboRefreshesWith(...), ['is_safe' => ['html']]),
+            new TwigFunction('turbo_refresh_method', $this->turboRefreshMethod(...), ['is_safe' => ['html']]),
+            new TwigFunction('turbo_refresh_scroll', $this->turboRefreshScroll(...), ['is_safe' => ['html']]),
         ];
     }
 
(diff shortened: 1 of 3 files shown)


src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -32,6 +38,9 @@ public function getFunctions(): array
     {
         return [
             new TwigFunction('turbo_stream_listen', $this->turboStreamListen(...), ['needs_environment' => true, 'is_safe' => ['html']]),
+            new TwigFunction('turbo_refreshes_with', $this->turboRefreshesWith(...), ['is_safe' => ['html']]),
```
Kocal, line 44: I'm not fan of `turbo_refreshes_with` name, I think users will prefer to use `turbo_refresh(method='replace', scroll='reset')... What do you think?

```suggestion
            new TwigFunction('turbo_refresh', $this->turboRefresh(...), ['is_safe' => ['html']]),
```

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -52,4 +61,27 @@ public function turboStreamListen(Environment $env, $topic, ?string $transport =
 
         return $this->turboStreamListenRenderers->get($transport)->renderTurboStreamListen($env, $topic);
     }
+
+    public function turboRefreshesWith(string $method = self::REFRESH_METHOD_REPLACE, string $scroll = self::REFRESH_SCROLL_RESET): string
+    {
+        return $this->turboRefreshMethod($method).$this->turboRefreshScroll($scroll);
+    }
+
+    public function turboRefreshMethod(string $method = self::REFRESH_METHOD_REPLACE): string
```
Kocal, line 129: Please add some PHPDoc and a link to the documentation

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -52,4 +61,27 @@ public function turboStreamListen(Environment $env, $topic, ?string $transport =
...
+    {
+        if (!\in_array($method, [self::REFRESH_METHOD_REPLACE, self::REFRESH_METHOD_MORPH], true)) {
+            throw new \InvalidArgumentException(\sprintf('Invalid refresh option "%s".', $method));
+        }
+
+        return \sprintf('<meta name="turbo-refresh-method" content="%s">', $method);
+    }
+
+    public function turboRefreshScroll(string $scroll = self::REFRESH_SCROLL_RESET): string
```
Kocal, line 150: Please add some PHPDoc and a link to the documentation

Kocal (changes requested):
Hi, can you also add some tests and update the CHANGELOG (for 2.24)? 

Thanks!

PS: please for the next time, keep the PR template 🙏🏻

smnandre:
I don't think we need to have 3 methods here. What about keeping just the first two ?

Also i would not set the default values in the methods and force users to be explicit (they do not need this function to set default value... and the day Turbo changes its default value it would be a real mess).

seb-jean:
Hi to you both :),

Thank you for your feedback.

I added the PHPDocs. and updated CHANGELOG.
I will add the documentation later :).

> I don't think we need to have 3 methods here. What about keeping just the first two ?

These <meta> tags are generally used together. [Turbo Rails](https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/drive_helper.rb#L70) and [Turbo Laravel](https://turbo-laravel.com/docs/helpers/) also use this third function.

```rb
# https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/drive_helper.rb

#   turbo_refreshes_with(method: :morph, scroll: :preserve)
def turbo_refreshes_with(method: :replace, scroll: :reset)
  provide :head, turbo_refresh_method_tag(method)
  provide :head, turbo_refresh_scroll_tag(scroll)
end
```

> Also i would not set the default values in the methods and force users to be explicit (they do not need this function to set default value... and the day Turbo changes its default value it would be a real mess).

For [Morphing](https://turbo.hotwired.dev/handbook/page_refreshes#morphing), Turbo says the default is `replace`. Similarly, for [Scroll preservation](https://turbo.hotwired.dev/handbook/page_refreshes#scroll-preservation), Turbo indicates that the default value is `reset`.

These default values ​​are also found in Turbo Rails:

- Method : https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/drive_helper.rb#L76
- Scroll : https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/drive_helper.rb#L82

```rb
# https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/drive_helper.rb

# Configure method to perform page refreshes. See +turbo_refreshes_with+.
def turbo_refresh_method_tag(method = :replace)
  raise ArgumentError, "Invalid refresh option '#{method}'" unless method.in?(%i[ replace morph ])
  tag.meta(name: "turbo-refresh-method", content: method)
end

# Configure scroll strategy for page refreshes. See +turbo_refreshes_with+.
def turbo_refresh_scroll_tag(scroll = :reset)
  raise ArgumentError, "Invalid scroll option '#{scroll}'" unless scroll.in?(%i[ reset preserve ])
  tag.meta(name: "turbo-refresh-scroll", content: scroll)
end
```

Kocal:
Works for me then, thanks!
Maybe it could be a good idea to document (in the PHPDoc) that these new methods are "copied" from Turbo Rails?

seb-jean:
I put a few words (`Inspired by Turbo Rails ...`) per function. Is it better if I put those few words at the class level instead?

Kocal:
That's perfect, thanks!

Kocal (approved):

seb-jean:
I added 3 new `meta` tags and some documentation.

Kocal (commented):
Thanks!

Can you squash/rebase your PR and fix issues found by DoctorRST please?

Kocal:
Status: Needs work

seb-jean:
I have squash and fix issues found by DoctorRST.

Kocal:
Thanks @seb-jean.

PR #2627, opened 2025-03-06
[LiveComponent] Fix default option extraction ignores Translatable #2621

### LINKED ISSUE #2621: [LiveForm] Required select with placeholder of TranslatableMessage has first value set upon rerender
**The bug:**
While having a liveform with required select and placeholder of type TranslatableMessage, instead of leaving this value blank, first not null value from choices is retrieved instead.

This behaviour occurs because of code in `Symfony\UX\LiveComponent\ComponentWithFormTrait` in if on line 276.
https://github.com/symfony/ux/blob/314ca710a16f79ad2d1d3d2c74f5f03dd440fd09/src/LiveComponent/src/ComponentWithFormTrait.php#L276.

In this if, the last condition is `!\is_string($child->vars['placeholder'])` which unfortunately is not satisfied e.g. for option `'placeholder' => new TranslatableMessage('some.value')`.

**Solution**
I would suggest to extending this condition to `(!\is_string($child->vars['placeholder']) && !$child->vars['placeholder'] instanceof TranslatableMessage)` or making it only `!isset($child->vars['placeholder'])` to support more cases. I feel that string only restriction is to harsh when most of the time, we will be creating multilingual applications.

We use TranslatableMessage class for translatable strings so our extractor can generate .yaml files with translations.

stehled:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2621
| License       | MIT

Select placeholder in LiveForm will now accept `string` and `TranslatableInterface` from `symfony/translation-contracts`. Therefore mimicking of default browser behaviour will not happen for `TranslatableInterface`.

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.25.0
 
 -   Add support for [Symfony UID](https://symfony.com/doc/current/components/uid.html) hydration/dehydration
+-   `ComponentWithFormTrait` now correctly checks for a `TranslatableInterface` placeholder for `<select>` elements
 
 ## 2.23.0
 
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -16,6 +16,7 @@ use Symfony\Component\Form\ClearableErrorsInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
+use Symfony\Contracts\Translation\TranslatableInterface;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
 use Symfony\UX\LiveComponent\Attribute\PreReRender;
 use Symfony\UX\LiveComponent\Util\LiveFormUtility;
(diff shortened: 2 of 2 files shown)


smnandre:
Your composer.json change is not needed here, so let's avoid adding a new dependency for this single check :) 

```php
<?php

use Not\Installed\TranslatableInterface;

if ('aa' instanceof TranslatableInterface) {
    echo 'yep';
} else {
    echo 'nope';
}
```

--> "nope" and no error/warning  (see https://3v4l.org/4nHHd )




(on contrary to something like this where an error would occur)

```php
<?php

use Not\Installed\TranslatableInterface;

// This would trigger a Fatal Error Interface "Not\Installed\TranslatableInterface" not found
$foo = new class implements TranslatableInterface {};
```

src/LiveComponent/composer.json
```diff
@@ -31,6 +31,7 @@
         "symfony/property-access": "^5.4.5|^6.0|^7.0",
         "symfony/property-info": "^5.4|^6.0|^7.0",
         "symfony/stimulus-bundle": "^2.9",
+        "symfony/translation-contracts": "^2.5|^3.0",
```
smnandre:
```suggestion
```

smnandre (changes requested):

stehled:
I've committed the suggestion, thank you for this comment, I've learned something new :)

smnandre (approved):

smnandre:
Thank you @stehled !

Kocal:
Thank you @stehled.

PR #2628, opened 2025-03-06
[Live] Preserve file input values after render requests

### LINKED ISSUE #2625: [LiveForm] Files cleared when form error
When submitting a form with a Live Component (live action), if there is an error, we get a 422 response code because the entity is not valid. That’s expected. However, if the form contains a file input, the live_controller.js script clears the submitted files, preventing the user from reselecting them because they are not in the next requests anymore.

We found these three lines (2123-2125), and when commented out, the issue disappears:

` for (const input of Object.values(this.pendingFiles)) {
      input.value = '';
}`

We’re not sure if this is a bug or if it serves another purpose. In any case, resetting the form manually isn't necessary because we already call resetForm() in the Live Action, which clears the files.

akyoscommunication:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2625
| License       | MIT

Remove three lines in live_controller.js to avoid file input to be cleared after re-render. Tests ok, test in real conditions on production projects ok. 

I am not certain if anyone knows the original purpose of these three lines, whether they had a specific reason for being included, or if they should indeed be removed.

As this is my first pull request, I would appreciate any feedback if I've overlooked something or can improve the process.

Thanks

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -2120,9 +2120,6 @@ class Component {
         this.backendRequest.promise.then(async (response) => {
             const backendResponse = new BackendResponse(response);
             const html = await backendResponse.getBody();
-            for (const input of Object.values(this.pendingFiles)) {
-                input.value = '';
-            }
             const headers = backendResponse.response.headers;
             if (!headers.get('Content-Type')?.includes('application/vnd.live-component+html') &&
                 !headers.get('X-Live-Redirect')) {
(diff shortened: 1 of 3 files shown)


smnandre:
There is a problem with your implementation, now files are never removed :) 

For the initial reasons, I believe this thread contains most of it https://github.com/symfony/ux/pull/834#issuecomment-1549438434

I'm just not sure if all what applied then still does today (data-model/name, Morphdom behaviour, etc).

kevinmade:
I think it would be better to reset the file inputs after a successful backend response.
```
if (backendResponse.ok) {
  for (const input of Object.values(this.pendingFiles)) {
    input.value = "";
  }
}
```

Kocal:

Thanks!

Status: Needs Work

PR #2633, opened 2025-03-07
[DX] WDT and strict CSP

DavidPetrasek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #58223
| License       | MIT

Improve DX by enabling Web Debug Toolbar whilst navigating via Turbo Drive and a strict Content-Security-Policy is being used.

### DIFF
--- a/src/Turbo/src/EventSubscriber/WebDebugToolbarTurboDriveFixSubscriber.php
+++ b/src/Turbo/src/EventSubscriber/WebDebugToolbarTurboDriveFixSubscriber.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Symfony\UX\Turbo\EventSubscriber;
+
+use Symfony\Component\DependencyInjection\Attribute\When;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+
+/**
+ * Allows the Web Debug Toolbar (WDT) when navigating via Turbo Drive when a strict Content-Security-Policy is set. 
+ * This is done by reusing WDT nonces.
+ */
+
+#[When(env: 'dev')]
+class WebDebugToolbarTurboDriveFixSubscriber implements EventSubscriberInterface
+{
+    public function onKernelResponse(ResponseEvent $event): void
+    {
+        $request = $event->getRequest();
+        $routeName = $request->get('_route');
+
+        if ($routeName === '_wdt') {return;}
+        if ($request->headers->has('X-Turbo-Request-Id')) {return;}
+        if ('html' !== $request->getRequestFormat()) {return;}
+
+        $response = $event->getResponse();
+        $content = $response->getContent();
+        
+        $scriptContent = <<<'EOD'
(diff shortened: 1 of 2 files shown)


src/Turbo/src/EventSubscriber/WebDebugToolbarTurboDriveFixSubscriber.php
```diff
@@ -0,0 +1,88 @@
...
+                    if (wdtScript.nonce) {event.detail.fetchOptions.headers['X-SymfonyProfiler-Script-Nonce'] = wdtScript.nonce;}
+                }
+            });
+        EOD;
+        $scriptTag = '<script>'.$scriptContent.'</script>';
+
+        $hash = base64_encode(hash('sha256', $scriptContent, true));
+        $hashString = "'sha256-".$hash."'";
+
+        $csp = $response->headers->get('Content-Security-Policy');
```
smnandre, line 63: I would still see this code in the WDT codebase to be honest (there is already some Turbo specific tags)

src/Turbo/src/EventSubscriber/WebDebugToolbarTurboDriveFixSubscriber.php
```diff
@@ -0,0 +1,88 @@
...
+ * Allows the Web Debug Toolbar (WDT) when navigating via Turbo Drive when a strict Content-Security-Policy is set.
+ * This is done by reusing WDT nonces.
+ */
+#[When(env: 'dev')]
+class WebDebugToolbarTurboDriveFixSubscriber implements EventSubscriberInterface
+{
+    public function onKernelResponse(ResponseEvent $event): void
+    {
+        $request = $event->getRequest();
+        $routeName = $request->get('_route');
```
smnandre, line 29: You need to check 

        if (!$this->kernel->isDebug()) {
            return;
        }
        
(see: https://symfony.com/doc/current/profiler.html#updating-the-web-debug-toolbar-after-ajax-requests)

src/Turbo/src/EventSubscriber/WebDebugToolbarTurboDriveFixSubscriber.php
```diff
@@ -0,0 +1,88 @@
...
+
+        if ('_wdt' === $routeName) {
+            return;
+        }
+        if ($request->headers->has('X-Turbo-Request-Id')) {
+            return;
+        }
+        if ('html' !== $request->getRequestFormat()) {
+            return;
+        }
```
smnandre, line 39: Let's check other things here (status code, main request, ... )

src/Turbo/src/EventSubscriber/WebDebugToolbarTurboDriveFixSubscriber.php
smnandre, line 63: Is the script needed when CSP is not set ?

DavidPetrasek: Agree. I'll move it to WDT. I'll finish this and your other comments probably within 48 hours.

DavidPetrasek:
Moved here: https://github.com/symfony/symfony/pull/59945

Issue #2567, opened 2025-02-11
[Dropzone] Multiple uploads only show single file and no possibility to add more files


simonsolutions:
As already discussed and potentially solved in this PR https://github.com/symfony/ux/pull/512:

When set the dropzone to allow multiple files, it works to select more than one file and upload the selected files.

But two things are still not working:
1. The dropzone element only shows one file - not multiple
2. There is no button or element to add other files to the dropzone. It is only possible to remove the file ("x") and then add all files together.

Thanks for yout help :-)

smnandre:
Would you like to work on a implementation ?

rrr63:
Hi, I am facing the same problem here. If this issue is still open for a PR, I can work on it.

simonsolutions:
I haven't managed to find time to look into it. So it would be appreciated of you could work on it :-)

davidnectarestudio:
I need this feature too

davidnectarestudio:
@smnandre could you review this PR, please? https://github.com/symfony/ux/pull/3162

smnandre:
You should be interested by https://github.com/symfony/ux/pull/3769

PR #2642, opened 2025-03-19
[Dropzone] Inform the user of the number of files he uploads

rrr63:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Docs?         |no 
| Issues        | #2567
| License       | MIT

Hi, I was implementing a multiple file dropzone in one of my projects, and i was facing the same problem mentioned in #2567.
I was thinking for a solution, should we make a preview of all files if they're images ?
Sounds an interesting idea first, but if users upload a lot of images, it could be very ugly to manage the preview. Also, i was worried about avoiding regression for projects that are already in production and to break some custom dropzone styles.

The solution i used for my project is to show the number of extra files user uploads. I was inspired by the original file type field, it just shows us the number of files, according to the stylized dropzone of symfony ux, the first element (if it is an image) display a preview, and we write to user the number of extra files he uploads.

Here a screenshot of three fields : 
-dropzone with single file 
-dropzone with 3 files
-standard file type field with 3 files
Capture d’écran 2025-03-19 à 09 11 17

### DIFF
--- a/src/Dropzone/CHANGELOG.md
+++ b/src/Dropzone/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23
+
+-   Support multiple files preview
+
 ## 2.20
 
 -   Enable file replacement via "drag-and-drop"
--- a/src/Dropzone/assets/dist/controller.js
+++ b/src/Dropzone/assets/dist/controller.js
@@ -32,19 +32,25 @@ class default_1 extends Controller {
         this.dispatchEvent('clear');
     }
     onInputChange(event) {
-        const file = event.target.files[0];
-        if (typeof file === 'undefined') {
+        const files = event.target.files;
+        if (!files.length) {
             return;
         }
         this.inputTarget.style.display = 'none';
         this.placeholderTarget.style.display = 'none';
-        this.previewFilenameTarget.textContent = file.name;
+        const firstFile = files[0];
+        let displayText = firstFile.name;
+        if (files.length > 1) {
+            const additionalFiles = files.length - 1;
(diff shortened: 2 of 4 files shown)


Kocal:
Hi, and thanks for working on it!

Unfortunately, I'm not sure this is the best solution:
1. the english "files" is hard-coded and does not follow the browser locale, and it looks out of place for non-english users, e.g. french on your screenshot
2. I think people will expect other files thumbnail to be displayed as well, and not being hidden behind "+N files"

What about displaying the thumbnail + name of each dropped images instead?

src/Dropzone/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.23
```
Kocal, line 3:
```suggestion
## 2.24
```

Kocal (changes requested):

rrr63:
> Hi, and thanks for working on it!
> 
> Unfortunately, I'm not sure this is the best solution:
> 
> 1. the english "files" is hard-coded and does not follow the browser locale, and it looks out of place for non-english users, e.g. french on your screenshot
> 2. I think people will expect other files thumbnail to be displayed as well, and not being hidden behind "+N files"
> 
> What about displaying the thumbnails + names of each dropped images instead?

Hi @Kocal , thank's for your comment

I was thinking about adding thumbnail to every single image dropped. 
But I was worried about breaking custom CSS because the necessary width (or height) could be very large if you have to show 3 or 4 images + filenames in the dropzone section.

Kocal:

Thanks!

Status: Needs Work

PR #2652, opened 2025-03-23
[LiveComponent] add LiveProp name to modifier function

### LINKED ISSUE #2650: [LiveComponent] Include prop name in the url modifier function
# Problem

The current (lack of) DX is best visible when defining abstract re-usable live components.

Let's have a minimal example, using something like a simple list with pagination:
```php
// parent
abstract class AbstractList
{
    use DefaultActionTrait;

    #[LiveProp(writable: true, url: true)]
    public int $page = 1;

    abstract public function getItems(): array;
}

// child 1
#[AsLiveComponent(name: 'FirstCoolList')]
class FirstCoolList extends AbstractList
{
    public function getItems(): array
    {
        // get items from database using $this->page to set offset/limit yadda yadda, not the point of this issue.
    }
}

// child 2
#[AsLiveComponent(name: 'SecondCoolList')]
class SecondCoolList extends AbstractList
{
    public function getItems(): array
    {
        // get items from database using $this->page to set offset/limit yadda yadda, not the point of this issue.
    }
}
```

Now when we render both components, we can see that the page param will be in conflict.. yikes... not good..

```twig
{{ component('FirstCoolList') }}
{{ component('SecondCoolList') }}
```

So.. since the param is declared in the parent, we can't just change the url anymore, we have to use a modifier. We can follow the LiveComponent documentation and it'll allow us to [change the URL parameters based a parameter](https://symfony.com/bundles/ux-live-component/current/index.html#controlling-the-query-parameter-name). Ok cool let's add that.
```php
abstract class AbstractList
{
    // ..

    #[LiveProp(writable: true, url: true, modifier: 'modifyPageQuery')]
    public int $page = 1;

    #[LiveProp]
    public string $pageParamAlias = 'page';

    // ..

    public function modifyPageQuery(LiveProp $prop): LiveProp
    {
        return $prop->withUrl(new UrlMapping($this->pageParamAlias));
    }
}
```

And we use it:
```twig
{{ component('FirstCoolList', { pageParamAlias: 'firstListPage' }) }}
{{ component('SecondCoolList', { pageParamAlias: 'secondListPage' }) }}
```

Now we modify our abstract class, because we don't just need to know the page, we also want to capture some search query the user has entered.

```php
abstract class AbstractList
{
    // ..

    #[LiveProp(writable: true, url: true, modifier: 'modifySearchQuery')]
    public string $query = 1;

    #[LiveProp]
    public string $searchParamAlias = 'q';

    #[LiveProp(writable: true, url: true, modifier: 'modifyPageQuery')]
    public int $page = 1;

    #[LiveProp]
    public string $pageParamAlias = 'page';

    // ..

    public function modifySearchQuery(LiveProp $prop): LiveProp
    {
        return $prop->withUrl(new UrlMapping($this->searchParamAlias));
    }

    public function modifyPageQuery(LiveProp $prop): LiveProp
    {
        return $prop->withUrl(new UrlMapping($this->pageParamAlias));
    }
}
```

To make sure our lists don't collide, we again need to modify both usages as well:

```twig
{{ component('FirstCoolList', { pageParamAlias: 'firstListPage', searchParamAlias: 'firstListSearch' }) }}
{{ component('SecondCoolList', { pageParamAlias: 'secondListPage', searchParamAlias: 'secondListSearch' }) }}
```

Rinse and repeat for 4 or 5 other params, and you can see how the abstract class is starting to get out of hand, and in terms of maintainability, if the usage of the components is spread out over multiple Symfony bundles, hundreds of templates,... this becomes absolutely impossible to maintain.

# Proposal

Pass the name of the property down into the modify function, either in the LiveProp object, or as a second parameter. This way we can configure 1 prefix on the component that can be applied to all props.

```php
abstract class AbstractList
{
    // ..

    #[LiveProp(writable: true, url: true, modifier: 'addPrefix')]
    public int $page = 1;

    #[LiveProp]
    public string $queryPrefix= '';

    // ..

   // PROPOSAL 1:
    public function addPrefix(LiveProp $prop): LiveProp
    {
        $param = $this->queryPrefix . $prop->getName(); // "page"
        return $prop->withUrl(new UrlMapping($param));
    }

    // PROPOSAL 2:
    public function addPrefix(LiveProp $prop, string $propName): LiveProp
    {
        $param = $this->queryPrefix . $propName; // "page"
        return $prop->withUrl(new UrlMapping($param));
    }
}
```

Now we no longer need to modify usages of the component whenever we add new properties that are controllable by the URL.
Both proposals would be completely backwards compatible, since we're just adding another param or new field on LiveProp, we're not changing the behavior of the modifier.

_It's unlikely that multiple lists of the same component are shown on the same page meanwhile it's very possible that multiple lists of different components are shown on 1 page so in my case I would actually set the prefix as a protected field on the implementation of the list, but I didn't want to use this use-case in the example since it's very application dependent._

What are your thoughts?

jannes-io:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2650
| License       | MIT

Implements proposal 2 from the examples in the RFC.

Personally I would just pass the entire liveprop metadata object to this function instead of only the name, but it's marked internal so I'm not sure if we should be exposing it to users like that.


src/LiveComponent/doc/index.rst
```diff
@@ -2617,7 +2617,9 @@ definition::
 
 Then the ``query`` value will appear in the URL like ``https://my.domain/search?q=my+query+string``.
 
-If you need to change the parameter name on a specific page, you can leverage the :ref:`modifier <modifier>` option::
+If you need to change the parameter name on a specific page, you can leverage the :ref:`modifier <modifier>` option
```
smnandre: A couple of remarks here (unrelated to wether we will move forward with this PR or not :) )

First of all: thank you for thinking immediately "documentation". That's a very nice habit to have. 

Now for the changes here: I think we should not _change_ the existing documentation, but instead we may _add_ a dedicated paragraph _after_ this part. 

The current version does it jobs, covers the feature and how most developers use it, while keeping things small / easy to understand. 

Moreover, using a prefix instead of an alias here would offer less possibilities than before, so we should probably not describe this solution in first examples, but only when a prefix _could_ be justified (many identical components in the same page ? still not sure to be honest)

So here I think a dedicated paragraph after this one, for more advanced use case, could be more appropriate.

src/LiveComponent/src/Metadata/LivePropMetadata.php
```diff
@@ -135,7 +135,7 @@ public function withModifier(object $component): self
             throw new \LogicException(\sprintf('Method "%s::%s()" given in LiveProp "modifier" does not exist.', $component::class, $modifier));
         }
 
-        $modifiedLiveProp = $component->{$modifier}($this->liveProp);
+        $modifiedLiveProp = $component->{$modifier}($this->liveProp, $this->getName());
```
smnandre, line 138: For a change like this, a test would be needed. Especially for things related to query, URL, etc.

(The fact tests do pass after your changes does not proves it does work (nor that our current suite of test is good enough, to be fair))

weaverryan:
I personally don't have a big issue with removing the internal. Very thoughtful to realize that issue :).

Alternatively, if there is something you need on the metadata, we could add it to LiveProp

smnandre:
I don't see a problem to pass the second argument  👍 
(we should precise in docs this information may not be the original name of the prop in the class)

Passing the metadata is not something I'd love to do, without an absolute need to do so.
Anything we expose cannot then be changed afterwards.. And it did bring us in situations were we're a bit "stuck" to move forward / improve features.

jannes-io:
> I don't see a problem to pass the second argument 👍 (we should precise in docs this information may not be the original name of the prop in the class)

Okidoki, I can do the requested changes to the docs and add some tests to finish off this PR. Could you give me a case where the property name could be different from the name in the metadata? Or where I could find code that could create this side-effect so I can try to reproduce? Just want to understand what the implications might be there.

> Passing the metadata is not something I'd love to do, without an absolute need to do so. Anything we expose cannot then be changed afterwards.. And it did bring us in situations were we're a bit "stuck" to move forward / improve features.

Agreed 😄

src/LiveComponent/doc/index.rst
```diff
@@ -2651,6 +2651,47 @@ This way you can also use the component multiple times in the same page and avoi
     <twig:SearchModule alias="q1" />
     <twig:SearchModule alias="q2" />
 
+.. versionadded:: 2.25
+
+   The property name is passed into the modifier function since LiveComponents 2.25.
```
jannes-io, line 2656: This'd have to be updated to the actual release version, I just took the next one.

src/LiveComponent/src/Metadata/LivePropMetadata.php
jannes-io: Added to the existing test, I don't see the need to make a separate test for just this param unless there are some edge cases that aren't entirely obvious to me at this time.

jannes-io:
@smnandre , @weaverryan Is there anything I can do to move this along?

src/LiveComponent/doc/index.rst
```diff
@@ -2651,6 +2651,47 @@ This way you can also use the component multiple times in the same page and avoi
...
+    #[AsLiveComponent]
+    class ImportantSearchModule extends AbstractSearchModule
+    {
+    }
+
+    #[AsLiveComponent]
+    class SecondarySearchModule extends AbstractSearchModule
+    {
+        protected string $urlPrefix = 'secondary';
+    }
```
smnandre, line 2686: I'm wondering.. could we find use an example more focused on the feature than the PHP code structure abstract/extends here ?

I was thinking something like "imagine you have a CMS theme editor with a split view allowing you to adjust your design both in light and dark mode. ... 

You could use a ColorTheme LiveComponent for both, but then [describe problem]

```
<twig:ColorSettings prefix="dark" color="..." />

<twig:ColorSettings prefix="light" color="..." />

```

Thanks to the second parameter of YYY, you can dynamically prefix [...].
 
 ```        
     #[LiveProp]
      public string $theme;
 
      #[LiveProp(url: true, modifier: 'prefixParam')]
      public string $color;

        public function prefixParam($liveProp, string $propName): LiveProp
        {
                return $liveProp->withUrl(
                   // Light Theme will use ?light-color
                    new UrlMapping(as: $this->prefix.'-'.$propName)
               );
        }
```

Now your query parameters will be: XXX ZZZ 

It even works with the as: property ..  

```php

     // query param will be ZZZ

      #[LiveProp(url: true, as: 'bg' modifier: 'prefixParam')]
      public string $background;      
```




wdyt ?

src/LiveComponent/src/Metadata/LivePropMetadata.php
smnandre: Let's test a page with two instances of same component, different prefix used in the modifier ... and check that the correct one as the good value and not the other :)

smnandre:
LGTM, and we will improve the docs afterwards

(changelog must be updated, though)

jannes-io:
> (changelog must be updated, though)

Done.

As for the other remarks, apologies for not getting back to you on that.

1. For the test, I didn't find an existing test where actual LiveComponents are rendered on a page. I'm afraid I'm a little out of my depth to be the first to produce such a test. If there is such a test that can be easily copied and modified to test the new parameter, and I just missed it, I can take a look.
2. For the docs, it feels to me like the feature is adequately technically documented, we don't look at docs to read a story, we're there for information on how to use features so we can use them in our own stories. If someone wants to rewrite it with a better example, that's fair 😆

Thanks for your assistance.

jannes-io:
What in tarnation happened here... Fabbot told me not to merge, so I rebased, and now all this crap is in here..

CBA to fiddle with git, remade it here: #2744

PR #2657, opened 2025-03-25
[Autocomplete] Tom Select Options Plugins 

### LINKED ISSUE #2656: [Autocomplete] Tom Select Options Plugins
I create a new autocomplete field with AJAX capabilities using  #[AsEntityAutocompleteField] attribute and extending AbstractType.

When I configure the field, I want to add a plugin on tom-select using tom_select_options -> plugins configuration : 

```

#[AsEntityAutocompleteField]
class AdminCompanyAutocompleteField extends AbstractType
{
    {
        $resolver->setDefaults([
            'class' => Company::class,
            'choice_label' => 'label',
            'searchable_fields' => ['label'],
            'security' => User::ROLE_ADMIN,
            'multiple' => false,
            'tom_select_options' => [
                'plugins' =>  ['dropdown_input']
            ]
        ]);
    }

    {
    }
}
```

The problem is this part overwrite the Tom select plugins configuration done in controller.ts : 
```
           'tom_select_options' => [
                'plugins' =>  ['dropdown_input']
            ]
```

Overwriting at : https://github.com/luchidalgo/ux/blob/ab44abfd74ea319e8c2902af8abd7408e7eb3d5e/src/Autocomplete/assets/src/controller.ts#L217

Maybe we should use deep merge instead of shallow merge ?

luchidalgo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2656  
| License       | MIT

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -150,7 +150,7 @@ export default class extends Controller {
         }
 
         if ('plugins' in this.tomSelectOptionsValue && Array.isArray(this.tomSelectOptionsValue.plugins)) {
-            this.tomSelectOptionsValue.plugins.forEach(pluginName => {
+            this.tomSelectOptionsValue.plugins.forEach((pluginName) => {
                 plugins[pluginName] = {};
             });
         }
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -149,7 +149,7 @@ export default class extends Controller {
             plugins.virtual_scroll = {};
         }
 
-        if('plugins' in this.tomSelectOptionsValue && Array.isArray(this.tomSelectOptionsValue.plugins)) {
+        if ('plugins' in this.tomSelectOptionsValue && Array.isArray(this.tomSelectOptionsValue.plugins)) {
             this.tomSelectOptionsValue.plugins.forEach(pluginName => {
                 plugins[pluginName] = {};
             });
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -149,6 +149,12 @@ export default class extends Controller {
             plugins.virtual_scroll = {};
         }
 
+        if('plugins' in this.tomSelectOptionsValue && Array.isArray(this.tomSelectOptionsValue.plugins)) {
+            this.tomSelectOptionsValue.plugins.forEach(pluginName => {
+                plugins[pluginName] = {};
+            });
+        }
+
         const render: Partial<TomTemplates> = {
             no_results: () => {
                 return `<div class="no-results">${this.noResultsFoundTextValue}</div>`;
@@ -214,7 +220,7 @@ export default class extends Controller {
             config.shouldLoad = () => false;
         }
 
-        return this.#mergeObjects(config, this.tomSelectOptionsValue);
+        return this.#mergeObjects(this.tomSelectOptionsValue, config);
     }
 
     #createAutocomplete(): TomSelect {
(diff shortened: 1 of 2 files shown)


src/Autocomplete/assets/src/controller.ts
```diff
@@ -149,6 +149,12 @@ export default class extends Controller {
             plugins.virtual_scroll = {};
         }
 
+        if ('plugins' in this.tomSelectOptionsValue && Array.isArray(this.tomSelectOptionsValue.plugins)) {
+            this.tomSelectOptionsValue.plugins.forEach((pluginName) => {
+                plugins[pluginName] = {};
+            });
+        }
```
Kocal, line 156: Given examples (https://tom-select.js.org/plugins/caret-position/ and https://tom-select.js.org/plugins/checkbox-options/) it looks like `plugins` can also be a key-value object.

Can't we use a more simple solution like this?

```suggestion
        plugins = this.tomSelectOptionsValue.plugins || [];
```

src/Autocomplete/assets/src/controller.ts
```diff
@@ -214,7 +220,7 @@ export default class extends Controller {
             config.shouldLoad = () => false;
         }
 
-        return this.#mergeObjects(config, this.tomSelectOptionsValue);
+        return this.#mergeObjects(this.tomSelectOptionsValue, config);
```
Kocal, line 223: I'm not really sure about that, it looks like a BC no?

Kocal (changes requested):

src/Autocomplete/assets/src/controller.ts
luchidalgo: By doing this, we prevent user-provided options from overwriting the configuration created within this function, especially for plugins.

src/Autocomplete/assets/src/controller.ts
luchidalgo: This one is treaky : https://tom-select.js.org/docs/plugins/ says we can put just plugin names (simple array) or plugin with options (object in js side). Also with simple assignment we override plugins configuration added line 138 to 150

src/Autocomplete/assets/src/controller.ts
fracsi: As @Kocal said, it is BC break. By looking at the code, the original idea is: if a user provides some config values, these values always win over default config

luchidalgo: Maybe I can revert this change and merge the config with a copy of tomSelectOptionsValue, without the plugins previously merged by a specific process, in order to avoid them being replaced ?

hendrydv:
Hello @luchidalgo,

Did you find any solution in how to use Tom Select plugins with the Symfony UX Autocomplete Bundle? I am struggling with it at this moment.

luchidalgo:
> Hello @luchidalgo,
> 
> Did you find any solution in how to use Tom Select plugins with the Symfony UX Autocomplete Bundle? I am struggling with it at this moment.

Hello 👋

I finally don't change autocomplete code but I add tom_select_options in forms directly depending on specific usage (select or multiselect, clear_button, etc ...) reminding that virtual_scroll plugin is required by remote url autocomplete : 
```
        $resolver->setDefaults([
            'class' => User::class,
            'choice_label' => 'fullName',
            'searchable_fields' => ['firstName', 'lastName'],
            'security' => User::ROLE_ADMIN,
            'query_builder' => function (UserRepository $er) {
                return $er->createQueryBuilder('user')
                    ->where('user.isTrainer = true')
                    ->orderBy('user.lastName', 'ASC')
                    ->addOrderBy('user.firstName', 'ASC');
            },
            'tom_select_options' => [
                'plugins' => [
                    'dropdown_input' => [], // the plugin I want
                    'virtual_scroll' => [], // the plugin required by symfony UX autocomplete
                ]
            ]
        ]);
```

PR #2659, opened 2025-03-25
Drop Symfony PHPUnit bridge for PHPUnit on UX packages PHP >= 8.3

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Inspired from https://github.com/symfony/demo/commit/7f57f38cd878f2f594b67a2e93c7f6072c252f34#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34


Kocal:
Well, that's too early. 

On some UX packages depending on PHP 8.1+, we need to require PHPUnit 10 **but** we can't configure Symfony's PHPUnit Bridge `SymfonyTestsListener` and we don't have access to `$this->expectUserDeprecationMessage()` which is introduced in PHPUnit 11 (PHP 8.2+), making us impossible to expect deprecations. 
Thus, we can't migrate to _raw_ PHPUnit on these packages.

Symfony's PHPUnit Bridge can only be removed on UX packages depending on PHP 8.3+ (only Map and its Bridge)

Kocal:
Well, IDK what's happening with PHPUnit on Map packages, but it's now impossible to run tests, it takes ~3min30 instead of ~700ms, and it's impossible to debug with Blackfire ... 😓  
```
➜  Map git:(drop-phpunit-bridge-for-phpunit) ✗ sfp vendor/bin/phpunit tests/UXMapBundleTest.php   
PHPUnit 12.1-dev by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.19
Configuration: /Users/kocal/workspace-os/symfony-ux/src/Map/phpunit.xml.dist

DD                                                                  2 / 2 (100%)

Time: 03:28.655, Memory: 92.02 MB

OK, but there were issues!
Tests: 2, Assertions: 2, Deprecations: 1.
```

I'm bored, let's retry another time

Issue #2353, opened 2024-11-08
[Icons] Allow to customize the stroke width


javiereguiluz:
Some icons (e.g. `tabler:`) include a hardcoded stroke width:

```xml
<svg width="1em" height="1em" viewBox="0 0 24 24" ...>
    <path stroke-width="2" ... />
</svg>
```

Sometimes, e.g. when using big icons, you want to change that stroke to a smaller value for aesthetic reasons.

Right now I define this CSS class to change the stroke as needed:

```css
svg.icon-stroke-1-5,
svg.icon-stroke-1-5 g {
    stroke-width: 1.5;
}
```

I wonder if it would possible to make the stroke configurable for UX icons. Maybe it's not because not all icon sets define a stroke width to begin with. Thanks!

smnandre:
## What happens here

Iconify make some normalization on the SVG icons to offer a unified format for the icons. 

This does not happen on all icon sets (it does on the ones i use the most -- lucide ❤️ -- so i understand your need very much)

It sometimes add them on the first svg shape... but sometimes it groups the differents shapes in a group and does it only once. 

**Examples:**

lucide:circle
```svg
<svg viewBox="0 0 24 24">
<circle cx="12" cy="12" r="10" fill="none" 
stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"/>
</svg>
```

tabler:circle
```svg
<svg  viewBox="0 0 24 24">
<path fill="none"  d="M3 12a9 9 0 1 0 18 0a9 9 0 1 0-18 0"
stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"/>
</svg>
```

But it does no on some others like prime or material design.

prime:circle
```svg
<svg viewBox="0 0 24 24">
<path fill="currentColor" d="M12  <!-- ... --> 4.5"/>
</svg>
```
mdi:circle
```svg
<svg viewBox="0 0 24 24">
<path fill="currentColor" d="M12 <!-- ... -->  2"/>
</svg>
```

## What we cannot do

We cannot simply remove them inconditionnaly, as we have no idea what has been added here by iconify, what is meant to be in the code, etc etc.

And this would break render on many websites

## What we can do

Currently you can configure attributes you want to define on the root SVG tag. 

I think we can add some configuration entry to define the ones you'd want to filter out from the SVG code.

To me, with lucide icons, that would look something like this : 


```yaml
# config/packages/ux_icons.yaml
ux_icons:

[...]

PR #2661, opened 2025-03-27
[Icons] Add xmlns attribute to svg icons

### LINKED ISSUE #2561: [Icons] Icons are missing xmlns attribute
Reporting this as a bug, even if it may be I use the ux-icons bundle not in the indended fashion. Feel free to change issue type if this is the case.

I installed the UX icons package and use it to just download the icons with like
`console ux:icons:import flowbite:user-solid` and reference them as background images with css.
The icons download just fine but I was surprised the browsers could not render them. It appears the xmlns attribute is missing 
`xmlns="http://www.w3.org/2000/svg"` so the browsers do not render the icons as icons but as xml.
This happens independent of what library I use. I tried akar-icons, devicons and fa6-brands.
While in my case i can add the xmlns manually, the point of using ux-icons for me is not to manually wrangle files but to just download and enjoy.

As stated, maybe I am using it wrong :)

Enz000:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #2561
| License       | MIT

The xmlns attribute was not present in the downloaded SVG icons.
To follow the W3C standard, I added it inside the svg tag.


src/Icons/src/Iconify.php
```diff
@@ -26,7 +26,8 @@
 final class Iconify
 {
     public const API_ENDPOINT = 'https://api.iconify.design';
-
+    public const XMLNS_URL = 'http://www.w3.org/2000/svg';
```
smnandre:
```suggestion
    public const XMLNS_URL = 'https://www.w3.org/2000/svg';
```

smnandre:
Problem is you now add it for everyone, everywhere, even in the inline rendering, where this is not required (and has no realy meaning).

I do believe we should try to implement custom attribute allowing/disallow list

Kocal:
> Problem is you now add it for everyone, everywhere, even in the inline rendering, where this is not required (and has no really meaning).

Is it really an issue? Sure it may be not useful when inlining, but is it breaking the document or the browser? 

Popular icons "libraries" like [HeroIcons](https://heroicons.com) or [FontAwesome](https://fontawesome.com) or [Google's Icons](https://fonts.google.com/icons) provides SVG icons with a `xmlns` attribute, and no one tells you to remove it if you want to inline the svg in your document.

Let's be pragmatic, always adding this `xmlns` attribute is fine

src/Icons/src/Iconify.php
```diff
@@ -90,6 +91,7 @@ public function fetchIcon(string $prefix, string $name): Icon
 
         return new Icon($data['icons'][$name]['body'], [
             'viewBox' => \sprintf('0 0 %s %s', $width ?? $height, $height ?? $width),
+            'xmlns' => self::XMLNS_URL,
```
Kocal: `xmlns` attribute is usually the first attribute of `<svg>`, let's do the same:
```suggestion
            'xmlns' => self::XMLNS_URL,
            'viewBox' => \sprintf('0 0 %s %s', $width ?? $height, $height ?? $width),
```

src/Icons/src/Iconify.php
```diff
@@ -137,6 +139,7 @@ public function fetchIcons(string $prefix, array $names): array
 
             $icons[$iconName] = new Icon($iconData['body'], [
                 'viewBox' => \sprintf('0 0 %d %d', $width ?? $height, $height ?? $width),
+                'xmlns' => self::XMLNS_URL,
```
Kocal: Same comment

src/Icons/src/Iconify.php
Kocal: Let's make it private too, there is no use-case to make it public (and event for tests) 

Also, we can prefix it by `ATTR_` as well

Kocal (changes requested):

norkunas:
> Let's be pragmatic, always adding this xmlns attribute is fine

which will break many tests for us

Kocal:
> > Let's be pragmatic, always adding this xmlns attribute is fine
> 
> 
> 
> which will break many tests for us

Even if the rendered HTML changes, as long as it doesn't break rendering in the browser, that's totally fine.

We understand your tests will _break_, but we do not really cares here, it's not like we are removing some HTML attributes or whatever more impactful.

Also, note that we already did this kind of "breaking changes" on `ComponentAttributes` rendering because of performance issues, and we already stated that these kind of changes were fine.

Thanks for your comprehension!

norkunas:
Sorry,but this will expand html size for no reason for us ... Im not against it, but at least config option would be ok

Enz000:
> Sorry,but this will expand html size for no reason for us ... Im not against it, but at least config option would be ok

The HTML size is not a valid reason for me. The attribute’s length is about 44 characters...

norkunas:
and when you rendering 100 icons in same page it becomes 440 :)

Kocal:
I'm pretty sure you webserver can compress your HTML response...

Enz000:
> and when you rendering 100 icons in same page it becomes 440 :) 

True, but 440 characters represents almost 0,43 ko !
Still nothing but maybe I'm mistaken

norkunas:
i still stand with @smnandre . do as you wish, i will write a script to clean the files

norkunas:
> I'm pretty sure you webserver can compress your HTML response...

but the repository size also grows

Kocal:
Please let's keep the discussion pertinent, I don't think the arguments "my tests will break" nor the "HTML or repo size will grow" are resonable enough to prevent this feature to ship. Otherwise, shouldn't we stop developing things and remove code only? 😬  

To move forwards @Enz000, you still need to update Icons tests and we will be fine. Maybe add one or two lines in the change log about the "breaking" thing.

Thanks everyone!

norkunas:
As I said I'm not against it, just do not understand why such a rare need is pushed without opt-in for all users

smnandre:
I'm sure we can all be a bit more chill here (and respectful).

I think all points are valid here:

UX Icons was made to render SVG icons in Twig. So I do agree this use case does not feel a valid reason to change the HTML generated by everyone else.

On the other side, not having this attribute can be a real pain (at least during development) raising a valid DX problem here. And making it hard to use the icons for anything else than inline rendering. So I do agree this use case does represent a valid reason to change something. 

Ideally, this situation would have require just a tiny bit of goodwill, some people giving a bit of energy or time (like really not much here) to implement something more clever/adapted. 

I already mentioned on another issue (quesstion from Javier i think) the following idea: we could add a include/exclude system of attributes, globally and/or by set and/or by usage (in file / at render time), as this seems to be very opiniated debates. 

So to me forcing this attribute for everyone and storing it in cache is not the best idea, clearly. I suppose this would _maybe_ be something to be done in the commands, via an Input Option. 

But I don't want either to "penalize" someone that took some time to explain its problem, open a PR to try something, etc. 

Have a good week-end you guys

Kocal (approved):

smnandre (approved):
LGTM!

src/Icons/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 -   Add `aliases` configuration option to define icon alternative names.
 -   Add support for `int` and `float` attribute values in `<twig:ux:icon />`.
+-   Add `xmlns` attribute in downloaded svg.
```
Kocal: It must be added in a new section for 2.24:
```suggestion
## 2.24.0

-   Add `xmlns` attribute in downloaded svg.
```

Kocal: Also, as discussed in the PR but also with @smnandre yesterday, about the ""breaking changes"", it is true it can happens when [you didn't lock your icons](https://symfony.com/bundles/ux-icons/current/index.html#locking-on-demand-icons), and your tests/CI always fetch and re-fetch icons for Iconify. 

Maybe we can add few words about the impacts and how to deal with, something like this:
```suggestion
## 2.24.0

-    Add `xmlns` attribute to icons downloaded with Iconify, to correctly render icons browser as an external file, in SVG editors, and in files explorers or text editors previews.
It **may breaks your pipeline** if you assert on `ux_icon()` or `<twig:ux:icon>` output in your tests, and forgot [to lock your icons](https://symfony.com/bundles/ux-icons/current/index.html#locking-on-demand-icons).
We recommend you to **lock** your icons **before** upgrading to UX Icons 2.24. We also suggest you to to **force-lock** your icons **after** upgrading to UX Icons 2.24, to add the attribute `xmlns` to your icons already downloaded from Iconify.
```

Kocal: cc @norkunas

Kocal (changes requested):

src/Icons/CHANGELOG.md
norkunas: Actually I always import them, to avoid useless http requests :) so at least all previously imported icons will be clean

norkunas:
If I will set `ux_icon.default_icon_attributes.xmlns` config option to `false` will it avoid rendering it at least? If no - would it be possible to make this work?

src/Icons/CHANGELOG.md
smnandre: As I see it, we have two options to see it:
- If it’s a breaking change, we don’t release it. 
- If it’s not a breaking change, there’s no need to make a big deal out of it.

But releasing it with BREAKING in the changelog does not feel a good idea.

So which one it is ? 😅

smnandre:
I'd like we don't make a custom option per attribute, and use something more generic / once for all **if possible**.

But a temporary hard-coded option would be acceptable here, until someone has time to make clean things.

Discussion / implementation ideas here: https://github.com/symfony/ux/issues/2353

norkunas:
@smnandre I didn't said about custom option, I am exploring from what's available: `default_icon_attributes`

src/Icons/CHANGELOG.md
chalasr: Trying to cover so strictly the shape of the retval of a public method that isn't even documented code-wise is calling for trouble :) Putting that sentence in the PR body in case anyone want to read it should be more than enough

Kocal: I updated my suggestion about "breaking changes", I believe that's the best of both world.

@Enz000 please apply it with "2.24" change, and we will be good for merging

Thanks!

kbond:
> If I will set ux_icon.default_icon_attributes.xmlns config option to false

I like this. If it doesn't work already, we should make it work.

Kocal (approved):

Kocal:
Thanks @Enz000.

PR #2539, opened 2025-01-31
[Toolkit] Introduce the UX Toolkit ✨

Halleck45:
| Q             | A
| ------------- | ---
| Bug fix?      | no |
| New feature?  | yes |
| Issues        |  |
| License       | MIT

Following numerous discussions with various people, I'm opening this PR as a draft for a potential ux-toolkit component. This is the result of a joint reflection with @Kocal  and @smnandre .

I’m not speaking on their behalf, but summarizing my interpretation of our discussions. Of course, @Kocal  and @smnandre, feel free to correct, adjust, or add anything as needed.

### Saving time

Creating Twig components like Badge or Button seems to be a very common step. The same work appears to be done over and over again, with minor differences from one company to another.

Providing a "packaged" toolkit would allow these companies and developers to save time by starting from ready-made templates.

### Simplifying the process

Today, a beginner coming to Symfony has a lot to learn. Onboarding on a Symfony project can feel overwhelming, so it makes sense to offer them the most pleasant experience possible—including making it as easy as possible to create a clean and visually appealing web page.

### Improving quality and accessibility

By providing a toolkit, we can leverage an entire community to gradually enhance the quality of components, including their digital accessibility. Accessibility is a major challenge, and we can aim to provide components that help companies and developers better support it.

## How?

By providing:

+ Boilerplate for basic, accessible components (button, badge, etc.)
+ A `symfony console ux:toolkit:install-component Button` command that will generate the corresponding Twig component in the user's project. If necessary, this command will also install any required dependencies (if a component relies on another one).

Components will be unique. 

It will be possible for the community to distribute unofficial components: for example, the command could accept `symfony console ux:toolkit:install-component Button --registry github.com/MyRemote/UiKit`

The website [https://ux.symfony.com/](https://ux.symfony.com/) will include:

+ Presentations of each component, with a demo and the code needed to use it.
+ Options for automatic installation (via the command) or manual installation (by copying and pasting Twig code).
+ A few demos showcasing different variants.

## Philosophy 

Since the web is constantly evolving, it would be difficult to commit to maintaining components that require frequent updates.

An approach similar to [shadcn](https://ui.shadcn.com/) seems more suitable: components are generated locally on demand by the developer, who will handle updates if necessary.

This approach likely offers the greatest flexibility: the developer can customize their component as they see fit, add their own variants, and adapt it to their specific needs.

The list of components could be based on the OpenUI initiative, which already provides a solid selection.

[...]

PR #2662, opened 2025-03-28
Introduce a modern, JSX-like syntax for TwigComponents

Halleck45:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | 
| License       | MIT

Hey! 👋  

Following [this PR](https://github.com/symfony/ux/pull/2539), I have a small commit I'd love to discuss — I believe it goes hand-in-hand with the idea of introducing a **Toolkit**.

My proposal is about improving the **HTML syntax** used for TwigComponents, and **introducing a `JSX-like syntax`**

Currently, we write components like this:

```html
<twig:foo xxx=""></twig:foo>
```

My suggestion is to allow a more modern and readable syntax:

```html
<Foo xxx=""></Foo>
```

I strongly believe this new syntax would significantly improve readability.

Here’s a real-world example from one of my projects:

Before:

```html
<twig:PageTitle title="..." />

<twig:Button small  >
    <twig:ux:icon .../>
</twig:Button>

<twig:Section mdCols="2">
    <twig:BasicCard title="...">
    </twig:BasicCard>

    <twig:BasicCard title="...">
    </twig:BasicCard>
</twig:Section>
```

After:

```html
<PageTitle title="..." />

<Button small >
    <Ux:icon .../>
</Button>

<Section mdCols="2">
    <BasicCard title="...">
    </BasicCard>

    <BasicCard title="...">
    </BasicCard>
</Section>
```

To me, this looks cleaner, more readable, and more aligned with modern expectations (especially for developers familiar with React, etc.).

**💡 Backward compatibility note:**

This new syntax is entirely optional and fully backward compatible. The **JSX-like tags are only interpreted as Twig components if the tag name starts with an uppercase letter**. Tags starting with lowercase letters continue to behave as standard HTML or follow the current `<twig:...>` syntax rules.

The commit is super simple — which probably means I'm missing something 😅 — but before digging deeper, I’d love to get your thoughts on this direction.

Thanks!

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.24.0
+
+-  Introduce an experimental Short tags system for TwigComponents, making `twig:` prefix optional #2662
+
 ## 2.20.0
 
 -  Add Anonymous Component support for 3rd-party bundles #2019
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -1719,6 +1719,36 @@ Pass the name of some component as an argument to print its details:
     |                                                   | int $min = 10                     |
     +---------------------------------------------------+-----------------------------------+
 
+Short Tags
+----------
+
+An experimental new short tag system, allowing the omission of the 'twig:' prefix in HTML tags, was introduced in version 2.24.
+
+This mode allows you to omit the `twig:` prefix and reference components directly by their name,
+with the first letter capitalized.
+
+.. code-block:: html+twig
+
+    <Acme:Button type="primary">
+        Click me
+    </Acme:Button>
+
+This is equivalent to:
+
+.. code-block:: html+twig
+
+    <twig:Acme:Button type="primary">
+        Click me
+    </twig:Acme:Button>
+
+To enable this feature, add the following configuration:
+
+.. code-block:: yaml
+
+    # config/packages/twig_component.yaml
+    twig_component:
+        short_tags: true
+
 Contributing
 ------------
 
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -129,10 +129,24 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
         ;
 
         $container->register('ux.twig_component.twig.lexer', ComponentLexer::class);
+        if (true === $config['short_tags']) {
+            $container->getDefinition('ux.twig_component.twig.lexer')
+                ->addMethodCall('enableShortTags');
+        }
 
         $container->register('ux.twig_component.twig.environment_configurator', TwigEnvironmentConfigurator::class)
             ->setDecoratedService(new Reference('twig.configurator.environment'))
             ->setArguments([new Reference('ux.twig_component.twig.environment_configurator.inner')]);
+
+        // Currently, the ComponentLexer is not injected into the TwigEnvironmentConfigurator, but built directly in the
+        // code (with a new ComponentLexer($environment)).
+        // We cannot change this behavior without a major refactoring : environment is currently configured at runtime.
+        // So we add setters for our required options
+        // This should be improved in the future: currently, parameters of the ComponentLexer are not injectables.
+        if (true === $config['short_tags']) {
+            $container->getDefinition('ux.twig_component.twig.environment_configurator')
+                ->addMethodCall('enabledShortTags');
+        }
 
         $container->register('ux.twig_component.command.debug', TwigComponentDebugCommand::class)
             ->setArguments([
@@ -216,6 +230,10 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
                 ->booleanNode('profiler')
                     ->info('Enables the profiler for Twig Component (in debug mode)')
                     ->defaultValue('%kernel.debug%')
+                ->end()
+                ->booleanNode('short_tags')
+                    ->info('Enables the short syntax for Twig Components (the <twig: prefix is optional)')
+                    ->defaultValue(false)
                 ->end()
                 ->scalarNode('controllers_json')
                     ->setDeprecated('symfony/ux-twig-component', '2.18', 'The "twig_component.controllers_json" config option is deprecated, and will be removed in 3.0.')
--- a/src/TwigComponent/src/Twig/ComponentLexer.php
+++ b/src/TwigComponent/src/Twig/ComponentLexer.php
@@ -26,9 +26,11 @@ use Twig\TokenStream;
  */
 class ComponentLexer extends Lexer
 {
+    private bool $withShortTags = false;
+
     public function tokenize(Source $source): TokenStream
     {
-        $preLexer = new TwigPreLexer();
+        $preLexer = new TwigPreLexer(withShortTags: $this->withShortTags);
         $preparsed = $preLexer->preLexComponents($source->getCode());
 
         return parent::tokenize(
@@ -38,5 +40,10 @@ class ComponentLexer extends Lexer
                 $source->getPath()
             )
         );
+    }
+
+    public function enabledShortTags(): void
+    {
+        $this->withShortTags = true;
     }
 }
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -15,7 +15,7 @@ use Twig\Error\SyntaxError;
 use Twig\Lexer;
 
 /**
- * Rewrites <twig:component> syntaxes to {% component %} syntaxes.
+ * Rewrites <twig:component> or <Component> syntaxes to {% component %} syntaxes.
  */
 class TwigPreLexer
 {
@@ -28,15 +28,32 @@ class TwigPreLexer
      */
     private array $currentComponents = [];
 
-    public function __construct(int $startingLine = 1)
+    public function __construct(int $startingLine = 1, private readonly bool $withShortTags = false)
     {
         $this->line = $startingLine;
     }
 
     public function preLexComponents(string $input): string
     {
-        if (!str_contains($input, '<twig:')) {
+        // tag may be:
+        // - prefixed: <twig:componentName>
+        // - short (jsx like): <ComponentName> (with a capital letter)
+
+        $isPrefixedTags = str_contains($input, '<twig:');
+        $isShortTags = $this->withShortTags && preg_match_all('/<([A-Z][a-zA-Z0-9_:-]+)([^>]*)>/', $input, $matches, \PREG_SET_ORDER);
+
+        if (!$isPrefixedTags && !$isShortTags) {
             return $input;
+        }
+
+        if ($isShortTags) {
+            $componentNames = array_map(fn ($match) => $match[1], $matches);
+            $componentNames = array_unique(array_filter($componentNames));
+
+            // To simplify things in the rest of the class, we replace the component name with twig:<componentName>
+            foreach ($componentNames as $componentName) {
+                $input = preg_replace('!<(/?)'.preg_quote($componentName).'!', '<$1twig:'.lcfirst($componentName), $input);
+            }
         }
 
         $this->input = $input = str_replace(["\r\n", "\r"], "\n", $input);
@@ -394,7 +411,7 @@ class TwigPreLexer
         }
         $blockContents = $this->consumeUntilEndBlock();
 
-        $subLexer = new self($this->line);
+        $subLexer = new self($this->line, $this->withShortTags);
         $output .= $subLexer->preLexComponents($blockContents);
 
         $this->consume($closingTag);
--- a/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
+++ b/src/TwigComponent/tests/Unit/TwigPreLexerTest.php
@@ -27,6 +27,15 @@ final class TwigPreLexerTest extends TestCase
     }
 
     /**
+     * @dataProvider getLexTestsWhithShortOptions
+     */
+    public function testPreLexWithShortTags(string $input, string $expectedOutput): void
+    {
+        $lexer = new TwigPreLexer(withShortTags: true);
+        $this->assertSame($expectedOutput, $lexer->preLexComponents($input));
+    }
+
+    /**
      * @dataProvider getInvalidSyntaxTests
      */
     public function testPreLexThrowsExceptionOnInvalidSyntax(string $input, string $expectedMessage): void
@@ -376,6 +385,12 @@ final class TwigPreLexerTest extends TestCase
             '<twig:foobar bar="baz" {{ ...attr }}>content</twig:foobar>',
             '{% component \'foobar\' with { bar: \'baz\', ...attr } %}{% block content %}content{% endblock %}{% endcomponent %}',
         ];
+
+        yield 'jsx_component_simple_component_not_enabled_by_default' => [
+            '<Foo />',
+            '<Foo />',
+        ];
+
         yield 'component_with_comment_line' => [
             "<twig:foo \n   # bar  \n />",
             '{{ component(\'foo\') }}',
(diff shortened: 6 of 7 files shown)


src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -35,10 +35,27 @@ public function __construct(int $startingLine = 1)
...
+        $isClassical = str_contains($input, '<twig:');
+        $isJsx = preg_match_all('/<([A-Z][a-zA-Z0-9_:-]+)([^>]*)>/', $input, $matches, \PREG_SET_ORDER);
+
+        if (!$isClassical && !$isJsx) {
             return $input;
         }
 
+        if ($isJsx) {
+            $componentNames = array_map(fn ($match) => $match[1], $matches);
+            $componentNames = array_unique(array_filter($componentNames));
```
Kocal, line 51: Instead, can we use a single `array_reduce` to loop over `$componentNames` only 1 time instead of 3 times?

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -35,10 +35,27 @@ public function __construct(int $startingLine = 1)
...
+
+            // To simplify things in the rest of the class, we replace the component name with twig:<componentName>
+            foreach ($componentNames as $componentName) {
```
Kocal, line 54: Or maybe better, 1 loop instead of 4?  
```suggestion
            // To simplify things in the rest of the class, we replace the component name with twig:<componentName>
            foreach ($matches as $match) {
                if (!isset($componentName = $match[1])) {
                    continue;
                }
```

smnandre:
As i told you ... it's --personal stand point-- a _total_ no-go for me before 3.0.

Regarding the "new" syntax, am i missing something or is just the prefix ignored ? I mean, are there other features ? 

But I would **very much** use something like it   -- with a couple of maybe "control"  / safeguards but  --  yeah the DX **is** better without twig:)

Kocal (commented):
Cool :eyes:

src/TwigComponent/src/Twig/TwigPreLexer.php
smnandre: I think it would be even easier in the other way.... not sure we need the 2step PATH

smnandre:
> This new syntax is entirely optional and fully backward compatible.

No it is sadly not (at all).

When Twig HTML syntax was added, no one (we can bet on this) ths followinh thing in their templates:

<twig:A 

(Openin tag, followed by "twig"; follwed by colon, followed by letters)

Here it will have incidence on a lot of people, so yeah we cannot push it everywhere for everyone like that.


As you said, we can and should offer a flag/toggler/etc in the first time... But then we must not update already the documentation,  🤷‍♂️ .



So i'm still on the same mood: I know and agree it's cool §told you I already did it..)  But here it's not the part where the questions are. ALmost the oppostie, eve.  not a technical problem in usage, it's in term of impact, in clarity, in future rebustness and

Kheuval:
Hello, I like this idea a lot !

In Vue there is this [eslint rule](https://eslint.vuejs.org/rules/multi-word-component-names), I believe it would be nice to promote multi-words component names to improve readability even more and distinguish at a glance what is HTML and what is user-created.
This could take the form of some kind of warning triggered if a component is named using a single word.
In your example, `Foo` would become `AppFoo` for example.

What do you think ?

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -35,10 +35,27 @@ public function __construct(int $startingLine = 1)
 
     public function preLexComponents(string $input): string
     {
-        if (!str_contains($input, '<twig:')) {
+        // tag may be:
+        // - classical: <twig:componentName>
+        // - jsx: <ComponentName> (with a capital letter)
+
+        $isClassical = str_contains($input, '<twig:');
+        $isJsx = preg_match_all('/<([A-Z][a-zA-Z0-9_:-]+)([^>]*)>/', $input, $matches, \PREG_SET_ORDER);
```
Kocal: I don't think if we can really qualify it as JSX, the properties and spread syntaxes are different and here only the naming case is the same.
What about "$isShort" and "$isLong"?

alexander-schranz: can it be both? Then `has` is better as `is`: https://3v4l.org/Wln8L

alexander-schranz:
Just as here are some discussion about only add this to a new major. It could be possible to optin into this via something like:

```jinja
{% twig_component_mode 'experimental:short' %}
```

Inspired by the [trans_default_domain](https://symfony.com/doc/current/translation.html#using-twig-filters).

weaverryan:
I like the idea of making this "opt in" for now. That would allow people to start playing with this.
Personally I'd rather have a global config vs declaring at the top of the template

src/TwigComponent/src/Twig/TwigPreLexer.php
Halleck45: For me, changing the code might actually make it slower. Your solution helps reduce iterations if all the nodes are unique.

But if a page contains the same node multiple times, it will trigger multiple `preg_replace` calls - which are much more expensive than a simple array iteration.

I think it's better to first reduce the array to a unique list, then perform the minimal number of `preg_replace` calls.

That said, happy to discuss it of course!

Halleck45:
Hi! I’ve made the syntax disabled by default and added a global `short_syntax` option.

I also took the liberty of enhancing the `debug:twig-component` command while I was at it - in case we get any user feedback, I thought it would make it easier to help them debug the issue.

src/TwigComponent/src/Twig/TwigPreLexer.php
Halleck45: Thanks! To me, it doesn't seem relevant to use a `has` naming here. The goal is to detect which syntax is being used, not whether both are present.

If we ever introduce support for mixing syntaxes or handling them differently in the same input, then switching to a `hasShortSyntax` / `hasLongSyntax` style might make sense. But for now, a simple `$isShort` better reflects the current logic and avoids overcomplicating things prematurely.

src/TwigComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.21.0
+
+-  Introduce an experimental JSX-like syntax for TwigComponents, making `<twig:` prefix optional
```
yguedidi:
```suggestion
-  Introduce an experimental JSX-like syntax for TwigComponents, making `twig:` prefix optional
```

src/TwigComponent/doc/index.rst
```diff
@@ -1719,6 +1719,36 @@ Pass the name of some component as an argument to print its details:
     |                                                   | int $min = 10                     |
     +---------------------------------------------------+-----------------------------------+
 
+Short Syntax
+------------
+
+An experimental short syntax is available for components, and has been introduced in 2.21.
+
+This mode allows you to omit the `<twig:` prefix and reference components directly by their name,
```
yguedidi:
```suggestion
This mode allows you to omit the `twig:` prefix and reference components directly by their name,
```

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -85,6 +87,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $components = $this->findComponents();
         $this->displayComponentsTable($io, $components);
+        $this->displayDetailsAboutConfiguration($io);
```
yguedidi: shouldn't the configuration appear first, above components?

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -217,6 +229,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('Enables the profiler for Twig Component (in debug mode)')
                     ->defaultValue('%kernel.debug%')
                 ->end()
+                ->booleanNode('short_syntax')
+                    ->info('Enables the short syntax for Twig Components (the <twig: prefix is optional)')
```
yguedidi, line 235:
```suggestion
                    ->info('Enables the short syntax for Twig Components (the `twig:` prefix is optional)')
```

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -129,18 +129,30 @@ static function (ChildDefinition $definition, AsTwigComponent $attribute) {
         ;
 
         $container->register('ux.twig_component.twig.lexer', ComponentLexer::class);
+        $container->getDefinition('ux.twig_component.twig.lexer')
+            ->addMethodCall('enableShortSyntax', [$config['short_syntax']]);
```
yguedidi: to me:
- either call the method `enableShortSyntax` but with no parameter, and have the `addMethodCall` wrapped by an `if` using the config value
- either call the method `setShortSyntax` and keep the parameter

I'd favor the first personally

src/TwigComponent/src/Twig/ComponentLexer.php
```diff
@@ -26,9 +26,11 @@
  */
 class ComponentLexer extends Lexer
 {
+    private $withShortSyntax = false;
```
yguedidi:
```suggestion
    private $allowingShortSyntax = false;
```
sounds better IMO

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -28,17 +28,34 @@ class TwigPreLexer
...
     {
         $this->line = $startingLine;
     }
 
     public function preLexComponents(string $input): string
     {
-        if (!str_contains($input, '<twig:')) {
+        // tag may be:
+        // - long: <twig:componentName>
+        // - short (jsx like): <ComponentName> (with a capital letter)
```
yguedidi, line 40:
```suggestion
        // - short (JSX-like): <ComponentName> (with a capital first letter)
```

src/TwigComponent/src/Twig/TwigPreLexer.php
```diff
@@ -28,17 +28,34 @@ class TwigPreLexer
...
+
+        if (!$isLongSyntax && !$isShortSyntax) {
             return $input;
         }
 
+        if ($isShortSyntax) {
+            $componentNames = array_map(fn ($match) => $match[1], $matches);
+            $componentNames = array_unique(array_filter($componentNames));
+
+            // To simplify things in the rest of the class, we replace the component name with twig:<componentName>
```
yguedidi, line 53:
```suggestion
            // To simplify things in the rest of the class, we replace the component name with <twig:componentName>
```

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
yguedidi: btw, worth being a separate PR IMO (without the new setting of course), may get merged faster than this one

yguedidi (changes requested):
small remarks from me :)

smnandre:
> I'd rather have a global config vs declaring at the top of the template

So if there is a problem with only one file from a distant bundle resource, it would prevent to use the short syntax anywhere ?

src/TwigComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.21.0
```
```suggestion
## 2.21.0
```

`2.24` i think ?  (feel free to create a new section here)

src/TwigComponent/CHANGELOG.md
```suggestion
-  Add experimental support for prefix-less tags in HTML Syntax (`<FooBar />` instead of `<twig:FooBar />`)
```

src/TwigComponent/doc/index.rst
```diff
@@ -1719,6 +1719,36 @@ Pass the name of some component as an argument to print its details:
     |                                                   | int $min = 10                     |
     +---------------------------------------------------+-----------------------------------+
 
+Short Syntax
+------------
+
+An experimental short syntax is available for components, and has been introduced in 2.21.
```
smnandre: Let's use a "versionadded" for this

src/TwigComponent/doc/index.rst
```diff
@@ -1719,6 +1719,36 @@ Pass the name of some component as an argument to print its details:
     |                                                   | int $min = 10                     |
     +---------------------------------------------------+-----------------------------------+
 
+Short Syntax
```
smnandre: This is **not** a new syntax.

We really should be precise on this :) Having _two_ syntaxes is already something that did create its fair share of frustrations, documetation problems, tests, etc etc ..  let's not frighten or confuse anyone with no reason here :)

src/TwigComponent/doc/index.rst
```diff
@@ -1719,6 +1719,36 @@ Pass the name of some component as an argument to print its details:
...
+        Click me
+    </twig:Acme:Button>
+
+To enable this feature, add the following configuration:
+
+.. code-block:: yaml
+
+    # config/packages/twig_component.yaml
+    twig_component:
+        short_syntax: true
```
smnandre: `prefixless_tags` or `short_tags`  ?

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
smnandre: Yes it will be better if you can split this @Halleck45 thx!

src/TwigComponent/src/Command/TwigComponentDebugCommand.php
```diff
@@ -356,4 +359,17 @@ private function getAnonymousComponentProperties(ComponentMetadata $metadata): a
...
+        $io->section('Configuration of TwigComponent');
+        $io->table(
+            ['Configuration', 'Current value'],
+            [
+                ['anonymous_template_directory', $this->anonymousDirectory],
+                ['short_syntax', $this->withShortSyntax ? 'enabled' : 'disabled'],
+                ['profiler',  $this->withProfiler ? 'enabled' : 'disabled'],
+            ]
+        );
+    }
```
smnandre: Not 100% sure this will bring more value than the `debug:config` command (a bit more complete / with examples, etc...)

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -129,18 +129,30 @@ static function (ChildDefinition $definition, AsTwigComponent $attribute) {
...
             ->setDecoratedService(new Reference('twig.configurator.environment'))
             ->setArguments([new Reference('ux.twig_component.twig.environment_configurator.inner')]);
 
+        // Currently, the ComponentLexer is not injected into the TwigEnvironmentConfigurator, but built directly in the
+        // code (with a new ComponentLexer($environment)).
+        // We cannot change this behavior without a major refactoring : environment is currently configured at runtime.
+        // So we add setters for our required options
+        // This should be improved in the future : currently, some parameters of the ComponentLexer are not settable
+        $container->getDefinition('ux.twig_component.twig.environment_configurator')
+            ->addMethodCall('enabledShortSyntax', [$config['short_syntax']]);
```
smnandre: Configurator should not have `itself` a setter, let's inject value IF we need to.

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -217,6 +229,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('Enables the profiler for Twig Component (in debug mode)')
                     ->defaultValue('%kernel.debug%')
                 ->end()
+                ->booleanNode('short_syntax')
```
smnandre:
```suggestion
                ->booleanNode('short_tags')
```

More accurate than "syntax", which was a bit of a stretch (same thing to do in the otheer changes -- open to another naming idea)

src/TwigComponent/tests/Unit/TwigPreLexerTest.php
```diff
@@ -437,4 +452,79 @@ public static function getLexTests(): iterable
             TWIG,
         ];
     }
+
+    public function getLexTestsWhithShortOptions()
```
smnandre:
```suggestion
    public static function getLexTestsWhithShortOptions()
```

smnandre:
I have real valid reason to not want this to be released right now. Sorry it has absolutely nothing to do with you @Halleck45 (or even the feature in itself).

But let's ignore them as the rest of the team decided. 

---

Focusing on the PR content.. good to me overall...  i'd like to make a final review when suggested changes are done (it will be easier to focus on things) 

Documentation/wording
- remove "JSX" mention (this syntax is in fact.... PascalCase HTML ^^)  
- replace "new syntax"  with one of the alternatives

Clean
- extract debug changes in another PR

And.... 
- add a lot more tests (using both <twig:Foo and <Foo in templates,  usage of ":",   comments,  attribute names, JSON content, etc etc)

Feel free if you need a hand on anything here 👍

Halleck45:
Hi! I believe I've addressed all the feedback.

I have:

- added quite a few tests. My imagination has limits, but I think I’ve covered a good range of cases
- removed the changes to the command
- standardized the naming to "ShortTags" and "shortTags"

I haven’t addressed the comments regarding autowiring. As mentioned in the comments, it’s currently not possible to use autowiring everywhere — the code frequently uses new, which makes it impossible without a major refactoring. In my opinion, rewriting the existing code is out of scope for this PR, which is why I prefer using an `addMethodCall`.
Moreover, many things depend on the environment, so pushing everything through autowiring doesn't always seem necessary for the existing code. In short, I don't think I’m in a position to handle this point in this branch.

Thanks !

nicolas-grekas:
My opinion on this PR:
I fear this is based on a very fragile technical base.

With this PR, the backend-side would be preempting every single tag that starts with an uppercase letter, effectively becoming incompatible with any frontend-side tag that'd use the same convention.

Since you mention JSX: it works with both custom and html5 tags *because* it has full knowledge of the DOM, since its parser is aligned to the HTML5 one.

But here, in twig, we're in confused-land from a parser PoV: twig knows nothing about HTML parsing rules, yet overlaps with some of them.

The `twig:` prefix made this somewhat acceptable: while it prevents writing anything on the frontend that'd use the same prefix, I think we can still consider us safe.

But any uppercase tag? This looks adventurous, to say the least.

A totally different template engine would be needed IMHO. Could be twig, with an HTML5 parser in front. But not just regexps.

I'm likely :-1: here, because of this. (I do understand the motivation, but building on sand wouldn't go very far)

Kocal:
I am closing due to inactivity, and for the reasons explained above. 
Thanks for the contribution!

PR #2663, opened 2025-03-28
[Autocomplete] Escape `querySelector` dynamic selector with `CSS.escape()`

Bartheyrman22:
Properly escape the querySelector for option-values like "["Test"]"/ 
This to mitigate: 
```
SyntaxError: Element.querySelector: ':scope > option[value="["foo"]"]' is not a valid selector
```

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Issues        | 
| License       | MIT

Some remark where I don't have a solution for:
The issue is only reproduceable on Chrome.
In Firefox, the code in the controller.js is not executed properly. So the error get's never triggered.
But no side effects detected so far.

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -192,7 +192,7 @@ export default class extends Controller {
                 for (const [, tomSelectOption] of Object.entries(this.tomSelect.options)) {
                     if (tomSelectOption.$order === optionOrder) {
                         orderedOption = parentElement.querySelector(
-                            `:scope > option[value="${tomSelectOption[this.tomSelect.settings.valueField]}"]`
+                            `:scope > option[value="${CSS.escape(tomSelectOption[this.tomSelect.settings.valueField])}"]`
                         );
 
                         break;
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.25.0
+
+-   Escape `querySelector` dynamic selector with `CSS.escape()` #2663
+
 ## 2.23.0
 
 -   Deprecate `ExtraLazyChoiceLoader` in favor of `Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader`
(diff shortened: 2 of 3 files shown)


Bartheyrman22:
Should I worry about the 3 failed checks?
Thanks in advance!

src/Autocomplete/assets/src/controller.ts
```diff
@@ -325,6 +325,10 @@ export default class extends Controller {
         return string.replace(/(<([^>]+)>)/gi, '');
     }
 
+    #addSlashes(string: string): string {
```
smnandre: I recently learned the existence of CSS.escape()  Would it not be better than a custom implementation here ?

src/Autocomplete/assets/src/controller.ts
```diff
@@ -192,7 +192,7 @@ export default class extends Controller {
                 for (const [, tomSelectOption] of Object.entries(this.tomSelect.options)) {
                     if (tomSelectOption.$order === optionOrder) {
                         orderedOption = parentElement.querySelector(
-                            `:scope > option[value="${tomSelectOption[this.tomSelect.settings.valueField]}"]`
+                            `:scope > option[value="${[this.#addSlashes(this.tomSelect.settings.valueField)]}"]`
```
smnandre: Are we sure here ? What is the goal of providing an array ?

smnandre:
> In Firefox, the code in the controller.js is not executed properly.

Hello @Bartheyrman22 ... do you have a simple reproducer we can test to see what happens here ?

src/Autocomplete/assets/src/controller.ts
Bartheyrman22: CSS.escape escapes more then strictly necessary but the solution you suggest is cleaner indeed.

src/Autocomplete/assets/src/controller.ts
Bartheyrman22: I'll check

Bartheyrman22: Removed the array again.

Bartheyrman22:
> > In Firefox, the code in the controller.js is not executed properly.
> 
> Hello @Bartheyrman22 ... do you have a simple reproducer we can test to see what happens here ?

I believe it is visible as soon as you debug in Firefox-debugger.
But I can try making a reproducer, is there a template to start from?

smnandre:
You _can_ start from the template i started ... it's not much more than a fresh install of symfony webapp + CSS .. but it's already a starting point :) 

You'll still need to install ux-autocomplete and add a new form to show the problem 

(probably composer update to start 😅 )

https://github.com/smnandre/ux-reproducer

smnandre:
> Should I worry about the 3 failed checks? Thanks in advance!

I don't think so... Twig/Live are 🔴  for another reason... and I just opened a PR to fix the JS files false-positive :)

Kocal (approved):

Kocal:
Thank you @Bartheyrman22.

PR #2666, opened 2025-03-30
[Icons] use the composer suggest section for optional dependencies

dbu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | -
| License       | MIT

Make installing user notice that they might want to install the optional dependencies.

### DIFF
--- a/src/Icons/composer.json
+++ b/src/Icons/composer.json
@@ -48,6 +48,10 @@
         "zenstruck/console-test": "^1.5",
         "psr/log": "^2|^3"
     },
+    "suggest": {
+        "symfony/http-client": "To load icons from icon sets",
+        "symfony/ux-twig-component": "To use the <twig:ux:icon> syntax for icons"
+    },
     "config": {
         "sort-packages": true
     },


kbond (approved):
I think in symfony/symfony packages, we don't use suggest but here, if it helps someone, I think it's ok.

Kocal (approved):

dbu:
yeah i wondered and saw symfony/symfony is not using them. imho it would be helpful here, especially the http-client which will probably be needed for a good part of the users.

smnandre:
I'm not sure to remember the reason ... maybe @kbond @weaverryan  or @nicolas-grekas ?

In all manner we won't change this for a single repository, so the question is more "do we want to add composer suggest to UX packages"

We could also add http-client contracts here and throw at runtime if no concrete implementation is available ?

nicolas-grekas:
We don't use that because it's just noise. composer won't display suggestions by default anymore for this reason. Every packages can add stuff there and this is a mess very quickly.

nicolas-grekas:
BTW: https://github.com/composer/composer/pull/12354

dbu:
> composer won't display suggestions by default anymore 

i was not aware of that. then imho lets drop this. i let you close it, in case there is still need for some discussion.

kbond:
Ah ok, I wasn't aware of the reason. TIL

PR #2671, opened 2025-03-30
[Icons] explain svg colors and improve example configuration

dbu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | -
| License       | MIT

I got mislead by the configuration examples setting the `fill` attribute on the `<svg>` element. All icons i looked have `<path>` elements that reference `currentColor` for either fill or outline. and `currentColor` in the context of CSS is the `color` attribute of the element. thus, to change the color of an icon, we need to set the CSS color. changing the `fill` attribute on the top level `<svg>` has no effect.

I am not 100% sure if this is always true, maybe the section i add needs to be less assertive or explain in which cases it is correct.

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -109,6 +109,13 @@ To keep your design consistent and easily adjustable, consider defining a
 :ref:`default attribute <icons_default_attributes>`. This allows you to control
 the size of all your icons from a single place.
 
+Icon Color
+~~~~~~~~~~
+
+Typically, SVG icons use ``fill="currentColor"`` to inherit the color of the containing element.
+You can set the color in CSS on the container or directly on the SVG element/class.
+
+
 Icon Sets
 ~~~~~~~~~
 


src/Icons/doc/index.rst
```diff
@@ -546,7 +549,7 @@ Full Configuration
 
         # Default attributes to add to all icons
         default_icon_attributes:
-            fill: currentColor
+            style: 'color: white;'
```
smnandre: Let's keep the documentation  as it was here (`fill: currentColor` is a default configuration value of this package)

src/Icons/doc/index.rst
```diff
@@ -324,17 +327,17 @@ added to all icons unless overridden by the second argument of the ``ux_icon`` f
     # config/packages/ux_icons.yaml
     ux_icons:
         default_icon_attributes:
-            fill: currentColor
+            style: "color: white;"
```
smnandre: Let's keep the documentation  as it was here (`fill: currentColor` is the default configuration value of this Icon, and I would not recommand seting a color by name on a global scope)

dbu:
ah, i did not realize its a default configuration. i reverted that change.

although i don't understand what that configuration is done for. all icons i look at use `fill=currentColor` or `outline=currentColor` in the elements underneath the `<svg>` root element, and setting the fill on the root svg element has no effect whatsoever. setting the color does have an effect however. so talking about the fill color seems confusing - it was confusing me when i thought i would be able to change the color of the icon using that, and it did not work.

if the default configuration is to work around weird icons that do not define the fill or outline to be currentColor, we could keep that in the doc about default configuration, but the example for adding global settings could show how to set color and size for example, to do something the user actually could want to do.

smnandre:
It is hard to find a solution that pleases everyone, and works for every icon 🤷 

"Styling icons" is indeed a good idea for a new dedicated section, with examples of hard-coded style, CSS classes, or global css styling.

Would you like to try writing a first version of it ?

dbu:
i think a section on styling icons is the right idea to share the relevant knowledge, even if it is not strictly about the symfony bundle per se.

i can't promise if i find time, its gotten pretty busy for me right now. i will try, but if somebody else wants to do it, don't hesitate :-)

smnandre:
Rebased (then removed my merge commit 😅 ) and ready to merge :)

smnandre:
Thank you David.

PR #2673, opened 2025-04-01
[LiveComponent] Add the possibility to map LiveProp as a path parameter

mbuliard:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | No
| License       | MIT

The purpose of this MR is to give the responsability of changing the URL to the backend, allowing to use Symfony Router and to have LiveProps in the path, not only in query parameters :
`http://example.com/content?id=123` vs
`http://example.com/content/123`

To set the LiveProp in the path, a new option, `mapPath` has been added to the UrlMapping option of the LiveProp :
```php
#[LiveProp(writable: true, url: new UrlMapping(mapPath: true))]
public int $id;
```
### WORKFLOW

1. When sending a request to the backend, the frontend add a new header `X-Live-Url` containing the current path and query parameters.
2. On KernelResponse event, the new path and query string is calculated from the header received and the new props and put in the response header `X-Live-Url`.
3. When the frontend receives the response, the current path and query are placed by those received, via `history.replaceState`.

### BACKEND CHANGES
- New **UrlFactory** service to generate URL from the previous one, the path-mapped props and the query-mapped props. The url is first generated by the Symfony Router, using the previous one and the path-mapped props. Then the query-mapped props and the previous query parameters are added.
- **QueryStringPropsExtractor** is renamed to **RequestPropsExtractor** and now extract props from the request attributes and query parameters.
- **UrlMapping** now has a new option `mapPath`, boolean, false by default.
- **LiveComponentMetadata** has new method `getAllUrlMappings` returning urlMappings of all LiveProps.
- **QueryStringInitializeSubscriber** is renamed to **RequestInitializeSubscriber**
- new **LiveUrlSubscriber**, listening to KernelResponse, and setting the `X-Live-Url` of the response with the new URL, generated by the `UrlFactory`. To generate it, the previous location is extracted from the request and the props are extracted from metadata, hydrated with the values of `_live_request_data` and sorted between path-mapped and query-mapped. 
- **LiveComponentSubscriber** now add `responseProps` data to the `_live_request_data` attribute, containing the mounted component data when the action is not the default one. This change is made to take server-side changes into account.
- **LiveComponentMetadata** has new method `getAllUrlMappings` returning urlMappings of all LiveProps.

### FRONTEND CHANGES
- **Backend/RequestBuilder** now add the current pathname and search as `X-Live-Url` header in the request.
- **Backend/BackendResponse** has new property `liveUrl`, populated from the HTTP response `X-Live-Url` header.
- **Component/index.ts** : `performRequest` now check for `X-Live-Url` header in response and, when found, do `history.replace` with the new url and the current hash and origin.
- **url_utils** is removed.
- **Component/plugins/QueryStringPlugin** is removed.

### TODO
Review :-)


smnandre:
(Had a quick chat with @mbuliard  yesterday — he's making a few adjustments before opening the PR. Stay tuned! 😄 )

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -328,6 +329,10 @@ export default class Component {
             }
 
             this.processRerender(html, backendResponse);
+            const liveUrl = await backendResponse.getLiveUrl();
+            if (liveUrl) {
+                HistoryStrategy.replace(new UrlUtils(liveUrl + window.location.hash, window.location.origin));
```
smnandre:
```suggestion
                history.replaceState(history.state, '', liveUrl + window.location.hash);
```

? (not sure)

smnandre: This would allow the remove both queryStringPlugin.ts and url_utils.ts (i think)

smnandre:
Do you want/need any help here @mbuliard ?

mbuliard:
> Do you want/need any help here @mbuliard ?

Thanks ! I needed free time, but the PR is almost done. Related tests are green, but I shall now add my own testing of the backend.

What took me some time was to the handling of custom actions, who may (and often do) update component props. I choose to store the updated component props in `_live_request_data.responseProps`, next to `props` and `updated` values.
I'm thinking to always populate these `responseProps`, even for default action and then simplifying my LiveUrlSubscriber.

smnandre:
Hey @squrious if you find a minute, i'd be happy to have your input (even a quick one) here :)

smnandre:
To me it's 💯  👍 ..

@Kocal @kbond I'd like a second look on this!

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,43 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+    }
+
+
+The if the symfony route is defined like this::
+
+    #[Route('/node/{id}', name: 'node')]
+    public function node(NodeModule $nodeModule): Response
+    {
+        // ...
+    }
```
```suggestion
Then, define the route like this::

    #[AsLiveComponent]
    class NodeModule
    {
        // ... 

        #[Route('/node/{id}', name: 'node')]
        public function node(NodeModule $nodeModule): Response
        {
            // ...
        }

        // ...
    }
```

You'd define the controller in your component, correct? I'm not exactly sure what you'd return here? Is the `NodeModule::$id` still _live_ updated? Could you add the route attribute to the component class itself if you don't require any logic in the controller?

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,43 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+
+Then the ``id`` value will appear in the URL like ``https://my.domain/node/my-node-id``.
+
+If the route parameter name is different from the LiveProp name, the ``as`` option can be used to map the LiveProp.
+
+If the route parameter is not defined, the ``mapPath`` option will be ignored and the LiveProp value will fallback to a query parameter.
```
kbond, line 2778: I'm not quite understanding this - should it be: "if the _route_ is not defined...? I wonder: should an exception be thrown in this case?

kbond (commented):
Looks good! Couple questions in my review but also: Can/should you have multiple routes per component? Or should multiple `LiveProp`'s with `mapPath: true` all be contained in a single route?

I think this should be clarified but also: it looks to me like this fully overrides the route for your component? If so, what about *this* concept being the *feature*?

mbuliard:
@kbond 
Thanks for your review !
Defining the Route in the Component would be an interesting idea but it would be far too big for this MR in my opinion.
For now, the `mapPath` option will only tell the `LiveUrlSubscriber` to try to set the `LiveProp` in the path instead of the query.

> Can/should you have multiple routes per component? Or should multiple `LiveProp`'s with `mapPath: true` all be contained in a single route?

The `Route` and the `LiveProp` are not programmatically linked to each other so there is no restriction : several `LiveProps` can be used for the same `Route` and each `LiveProp` can be used for several `Routes`.

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -0,0 +1,95 @@
...
+
+    public function onKernelResponse(ResponseEvent $event): void
+    {
+        $request = $event->getRequest();
+        if (!$request->attributes->has('_live_component')) {
+            return;
+        }
+        if (!$event->isMainRequest()) {
+            return;
+        }
```
Kocal: Let's put this block at the very top

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $newUrl = null;
+        if ($previousLocation = $request->headers->get(self::URL_HEADER)) {
+            $liveProps = $this->getLivePropsToMap($request);
+            $newUrl = $this->urlFactory->createFromPreviousAndProps(
+                $previousLocation,
+                $liveProps['path'],
+                $liveProps['query']
+            );
```
Kocal: in one line

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        if ($newUrl) {
+            $event->getResponse()->headers->set(
+                self::URL_HEADER,
+                $newUrl
+            );
```
Kocal: in one line

src/LiveComponent/doc/index.rst
kbond: disregard, we should not throw an exception.

src/LiveComponent/doc/index.rst
kbond: disregard, I misunderstood the feature

kbond:
@mbuliard, sorry, I misunderstood the feature. It took me a bit to realize the docs showed the route/controller that _renders_ the component initially.

mbuliard:
> @mbuliard, sorry, I misunderstood the feature. It took me a bit to realize the docs showed the route/controller that _renders_ the component initially.

My example was not clear enough, I added a line to clarify that it was a controller.

smnandre:
What do you think here @Kocal @kbond ? :)

src/LiveComponent/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 2.27.0
```
Kocal:
```suggestion
## 2.28.0
```

src/LiveComponent/assets/src/Backend/BackendResponse.ts
```diff
@@ -13,4 +14,12 @@ export default class {
 
         return this.body;
     }
+
+    async getLiveUrl(): Promise<string | null> {
+        if (undefined === this.liveUrl) {
+            this.liveUrl = await this.response.headers.get('X-Live-Url');
```
Kocal: Can by sync, `Headers.get()` is not sync:
```suggestion
    getLiveUrl(): string | null {
        if (undefined === this.liveUrl) {
            this.liveUrl = this.response.headers.get('X-Live-Url');
```

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
 
 The ``query`` value will appear in the URL like ``/search?query=my+important+query&secondary-query=my+secondary+query``.
 
+Map the parameter to path instead of query
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.27
```
```suggestion
.. versionadded:: 2.28
```

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+
+    The ``mapPath`` option was added in LiveComponents 2.27.
```
```suggestion
    The ``mapPath`` option was added in LiveComponents 2.28.
```

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+
+Instead of setting the LiveProp as a query parameter, you can set it as route parameter.
```
```suggestion
Instead of setting the ``LiveProp`` as a query parameter, you can set it as route parameter.
```

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+definition::
```
Kocal: To remove or re-adapt?

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+    class NodeModule
+    {
+        #[LiveProp(writable: true, url: new UrlMapping(mapPath: true))]
+        public string $id = '';
+
+        // ...
+    }
+
+
+The if the symfony route is defined like this::
```
Kocal: Note sure what you mean here

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+
+    #[Route('/node/{id}', name: 'node')]
+    public function node(NodeModule $nodeModule): Response
+    {
+        // ...
+    }
+
+Then the ``id`` value will appear in the URL like ``https://my.domain/node/my-node-id``.
+
+If the route parameter name is different from the LiveProp name, the ``as`` option can be used to map the LiveProp.
```
```suggestion
If the route parameter name is different from the LiveProp name, the ``as`` option can be used to map the ``LiveProp``.
```

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -255,7 +255,17 @@ public function onKernelView(ViewEvent $event): void
             return;
         }
 
-        $event->setResponse($this->createResponse($request->attributes->get('_mounted_component')));
+        $mountedComponent = $request->attributes->get('_mounted_component');
+        if (!$request->attributes->get('_component_default_action', false)) {
+            // On custom action, props may be updated by the server side
+            // @todo discuss name responseProps
+            // @todo maybe always set in, including default action and use only this, ignoring `props` and `updated` for UrlFactory ?
```
Kocal: Missing todo's?

src/LiveComponent/src/Util/UrlFactory.php
```diff
@@ -0,0 +1,92 @@
...
+    private function replaceQueryString($url, array $props): string
+    {
+        $queryString = http_build_query($props);
+
+        return preg_replace('/[?#].*/', '', $url).
+            ('' !== $queryString ? '?' : '').
+            $queryString;
+    }
+
+    // Keep the query parameters of the previous request
```
Kocal: Use PHPDoc

src/LiveComponent/src/Util/UrlFactory.php
```diff
@@ -0,0 +1,92 @@
...
+    private function getPreviousQueryParameters(string $query): array
+    {
+        parse_str($query, $previousQueryParams);
+
+        return $previousQueryParams;
+    }
+
+    // Symfony router will set props in query if they do not match route parameter
```
Kocal: Use PHPDoc

src/LiveComponent/tests/Unit/Util/UrlFactoryTest.php
```diff
@@ -0,0 +1,179 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
+use Symfony\Component\Routing\Exception\ResourceNotFoundException;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\UX\LiveComponent\Util\UrlFactory;
+
+class UrlFactoryTest extends TestCase
+{
+    public function getData(): \Generator
```
Kocal:
```suggestion
    public static function provideTestCreate(): \Generator
```

src/LiveComponent/tests/Unit/EventListener/LiveUrlSubscriberTest.php
```diff
@@ -0,0 +1,192 @@
...
+            ],
+            'expectedPathProps' => [],
+            'expectedQueryProps' => ['queryMappedProp1' => 2],
+        ];
+    }
+
+    /**
+     * @dataProvider getData
+     */
+    public function testProps(
```
Kocal: I'm not sure to understand what you mean by "testProps", could you improve tests methods names please?

Kocal:
Thanks for working on it! Just few nitpicking, the PR seems pretty mature right now

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
mbuliard: Yes, I was not satisfied with the name and was hoping to find a new one.

src/LiveComponent/doc/index.rst
mbuliard: "if the symfony route is defined like this"
I rewrote the sentence and missed something I guess.

src/LiveComponent/doc/index.rst
mbuliard: I am not sure how to better write it :-/

mbuliard:
> Thanks for working on it! Just few nitpicking, the PR seems pretty mature right now

Thanks for your review ! I made the changes you suggested.

src/LiveComponent/doc/index.rst
Kocal: Sorry I think I've targeted the wrong line

Kocal: Ah, that's the GitHub UI, I'm seeing a `definition::` alone line 2688

 
Image

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+    class NodeModule
+    {
+        #[LiveProp(writable: true, url: new UrlMapping(mapPath: true))]
+        public string $id = '';
+
+        // ...
+    }
+
+
+If the symfony route is defined like this::
```
```suggestion
If the current route is defined like this::
```

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
 
+Map the parameter to path instead of query
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.28
+
+    The ``mapPath`` option was added in LiveComponents 2.28.
+
+Instead of setting the ``LiveProp`` as a query parameter, you can set it as route parameter.
+definition::
```
Kocal: What do we do with `definition::`?

src/LiveComponent/doc/index.rst
```diff
@@ -2677,6 +2677,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+    }
+
+
+If the symfony route is defined like this::
+
+    // src/Controller/NodeController.php
+    // ...
+
+    #[Route('/node/{id}', name: 'node')]
+    public function node(NodeModule $nodeModule): Response
```
Kocal: Wait, you can (or must) inject the LiveComponent instance as a route parameter? 🤔 
It seems weird to me

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+
+    public function onKernelResponse(ResponseEvent $event): void
+    {
+        if (!$event->isMainRequest()) {
+            return;
+        }
+        $request = $event->getRequest();
+        if (!$request->attributes->has('_live_component')) {
+            return;
+        }
```
Kocal:
```suggestion
        if (!$event->isMainRequest()) {
            return;
        }

        $request = $event->getRequest();

        if (!$request->attributes->has('_live_component')) {
            return;
        }
```

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $newUrl = null;
+        if ($previousLocation = $request->headers->get(self::URL_HEADER)) {
+            $liveProps = $this->getLivePropsToMap($request);
+            $newUrl = $this->urlFactory->createFromPreviousAndProps($previousLocation, $liveProps['path'], $liveProps['query']);
+        }
+
+        if ($newUrl) {
+            $event->getResponse()->headers->set(self::URL_HEADER, $newUrl);
+        }
```
Kocal:
```suggestion
        if ($previousLiveUrl = $request->headers->get(self::URL_HEADER)) {
            $liveProps = $this->getLivePropsToMap($request);
            $newLiveUrl = $this->urlFactory->createFromPreviousAndProps($previousLiveUrl, $liveProps['path'], $liveProps['query']);
            $event->getResponse()->headers->set(self::URL_HEADER, $newUrl);
        }
```

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            KernelEvents::RESPONSE => 'onKernelResponse',
+        ];
+    }
+
+    private function getLivePropsToMap(Request $request): array
```
Kocal:
```suggestion
    private function getLivePropsFromRequest(Request $request): array
```

or 

```suggestion
    private function extractLivePropsFromRequest(Request $request): array
```

Kocal: Please add PHPDoc to document the return type

src/LiveComponent/doc/index.rst
kbond: I don't think you need to but indeed this looks weird (and I think led to my confusion on this feature initially).

What about `/search/{mode}/{query}`? This way we can build upon the existing example in the docs.

mbuliard: @Kocal @kbond I can't remember nor figure out why I wrote this in the doc... Of course not, the LiveComponent should not be there, only the `$id` or the fetched $node.
I will rewrite this, following @kbond suggestion to follow the `/search` examples

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
mbuliard: I kept two separated ``if`` to not set the header if the ``$newLiveUrl`` is null.
It can happen if :
* the previous url cannot be parsed
* the previous url does not match any known route
* some mandatory parameter is missing to generate the url of the route

But maybe it would be better to throw exceptions ?
I was worried to break anything doing so

smnandre:
Can you rebase (a last time i hope) @mbuliard ?

Is that ok for you @Kocal ? Anything else to check ?

src/LiveComponent/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 2.29.0
+
+-   Add new `mapPath` options (default `false`) to `UrlMapping` of a `LiveProp`
+    to allow the prop to be mapped to the path instead of the query in the url.
```
Kocal: To move at `2.28.0`

Kocal (approved):

Kocal (approved):

src/LiveComponent/doc/index.rst
```diff
@@ -2737,6 +2737,46 @@ It can be used to perform more generic operations inside of the modifier that ca
...
+
+If the current route is defined like this::
+
+    // src/Controller/SearchController.php
+    // ...
+
+    #[Route('/search/{query}')]
+    public function __invoke(string $query): Response
+    {
+        // ...
```
```suggestion
        // ... render template that uses SearchModule component ...
```

Just to be extra clear.

kbond (changes requested):
Looks good! The docs are much clearer.

kbond (approved):

Kocal:
It took time, but here we go, this is in now. Thank you very much @mbuliard.

PR #2686, opened 2025-04-08
Docs: Update ChartjsController.php: Removing PackageRepository

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | none
| License       | MIT

Page: https://ux.symfony.com/chartjs

It's not used in the template - what was it meant for?

### DIFF
--- a/ux.symfony.com/src/Controller/UxPackage/ChartjsController.php
+++ b/ux.symfony.com/src/Controller/UxPackage/ChartjsController.php
@@ -11,7 +11,6 @@
 
 namespace App\Controller\UxPackage;
 
-use App\Service\UxPackageRepository;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Attribute\Route;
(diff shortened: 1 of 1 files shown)


smnandre:
Checks say it is :) 

https://github.com/symfony/ux/actions/runs/14334467435/job/40177885290?pr=2686

ThomasLandauer:
OK, so these are not just docs but the code is actually executed somewhere?

Looks like the real problem is that the Twig code that's shown in the top right corner is not the complete actual template: https://github.com/symfony/ux/blob/2.x/ux.symfony.com/templates/ux_packages/chartjs.html.twig

Don't know what to do then...

smnandre:
You're right. If you want to test the package, you can ignore anything related to 'package', it's internal to the website.

The website began as a demo, similar to the symfony-demo. We want it to become a "complete" SymfonyUX website, and need to find a way to show simple/raw code that people can test ... but also to support constraints that a bigger website has (directory structure, templates, etc...)

So we will try to isolate the demos more, while offering richer pages to present the packages.

It will take a bit of time, especially as my last weeks have not been fully focused on UX (and that I don't have infinite amount of time after work days, family, etc..)

ThomasLandauer:
Sure. So I'm closing here.

PR #2687, opened 2025-04-09
[CI] Use git diff instead of git status

smnandre:
The CI often fails on the "check dist files" script because some dist file changed .. with no real content change

This PR replace the `git status --porcelain` with a `git diff --quiet` to ignore this situation.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -41,7 +41,7 @@ jobs:
 
             - name: Check if JS dist files are current
               run: |
-                if [[ -n $(git status --porcelain) ]]; then
+                if ! git diff --quiet; then
                   echo "The Git workspace is unclean! Changes detected:"
                   git status --porcelain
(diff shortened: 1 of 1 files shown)


Kocal:
> The CI often fails on the "check dist files" script because some dist file changed .. with no real content change

Do you have a link to a failed job? I would like to understand what you call "with no real content change", thanks 🙏🏻

smnandre:
https://github.com/symfony/ux/actions/runs/14356670896/job/40247624603?pr=2663

Kocal:
Mmmh, why not, in #2663 it happens because lines 4 to 29 have CRLF instead of LF chars we expect (or vice-versa).

Let's merge it, but I will open a new PR to enforce LF chars for text files

Kocal (approved):

Kocal:
Thanks Simon.

PR #2688, opened 2025-04-10
[LiveComponent] WIP: sorting for LiveColletionType

haivala:
| Q             | A
| ------------- | ---
| New feature?  | yes 
| Docs?         | yes/no 
| License       | MIT

`LiveCollectionType` has now new option `allow_sort` that allows sorting
Also added same options than buttons add and delete has for button move_up and move_down

### DIFF
--- a/src/LiveComponent/src/LiveCollectionTrait.php
+++ b/src/LiveComponent/src/LiveCollectionTrait.php
@@ -46,6 +46,46 @@ trait LiveCollectionTrait
         $propertyAccessor->setValue($this->formValues, $propertyPath, $data);
     }
 
+    #[LiveAction]
+    public function moveCollectionItemUp(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name, #[LiveArg] int $index): void
+    {
+        if ($index <= 0) {
+            return;
+        }
+        
+        $propertyPath = $this->fieldNameToPropertyPath($name, $this->formName);
+        $data = $propertyAccessor->getValue($this->formValues, $propertyPath);
+        
+        if (!\is_array($data) || !isset($data[$index]) || !isset($data[$index - 1])) {
+            return;
+        }
+        
+        // Swap the current item with the one above it
+        $temp = $data[$index - 1];
+        $data[$index - 1] = $data[$index];
+        $data[$index] = $temp;
+        
+        $propertyAccessor->setValue($this->formValues, $propertyPath, $data);
+    }
+
+    #[LiveAction]
+    public function moveCollectionItemDown(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name, #[LiveArg] int $index): void
+    {
+        $propertyPath = $this->fieldNameToPropertyPath($name, $this->formName);
+        $data = $propertyAccessor->getValue($this->formValues, $propertyPath);
+        
+        if (!\is_array($data) || !isset($data[$index]) || !isset($data[$index + 1])) {
+            return;
+        }
+        
+        // Swap the current item with the one below it
+        $temp = $data[$index + 1];
+        $data[$index + 1] = $data[$index];
+        $data[$index] = $temp;
+        
+        $propertyAccessor->setValue($this->formValues, $propertyPath, $data);
+    }
+
     private function fieldNameToPropertyPath(string $collectionFieldName, string $rootFormName): string
     {
         $propertyPath = $collectionFieldName;
--- a/src/LiveComponent/src/LiveCollectionTrait.php
+++ b/src/LiveComponent/src/LiveCollectionTrait.php
@@ -63,11 +63,11 @@ trait LiveCollectionTrait
         $formConfig = $this->getForm()->get($name)->getConfig();
         $orderPropertyPath = $formConfig->getOption('order_property_path');
 
-        if ($orderPropertyPath !== null) {
+        if (null !== $orderPropertyPath) {
             // Swap positions using the specified order property
             $prevItemOrder = $propertyAccessor->getValue($data[$index - 1], $orderPropertyPath);
             $currentItemOrder = $propertyAccessor->getValue($data[$index], $orderPropertyPath);
-            
+
             $propertyAccessor->setValue($data[$index], $orderPropertyPath, $prevItemOrder);
             $propertyAccessor->setValue($data[$index - 1], $orderPropertyPath, $currentItemOrder);
         } else {
@@ -93,11 +93,11 @@ trait LiveCollectionTrait
         $formConfig = $this->getForm()->get($name)->getConfig();
         $orderPropertyPath = $formConfig->getOption('order_property_path');
 
-        if ($orderPropertyPath !== null) {
+        if (null !== $orderPropertyPath) {
             // Swap positions using the specified order property
             $nextItemOrder = $propertyAccessor->getValue($data[$index + 1], $orderPropertyPath);
             $currentItemOrder = $propertyAccessor->getValue($data[$index], $orderPropertyPath);
-            
+
             $propertyAccessor->setValue($data[$index], $orderPropertyPath, $nextItemOrder);
             $propertyAccessor->setValue($data[$index + 1], $orderPropertyPath, $currentItemOrder);
         } else {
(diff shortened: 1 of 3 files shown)


haivala:
I would actually move the buttons (delete, move) before the the form row in the template (line 9) so that they are not with the add button as they control the current entry type so they should be first?

Example
2025-04-10_140149-area

Pechynho:
I don't think this is a good way how to order thinks, because this does not respect object references.

Let me give you an example.

Let's say I have entities these categories.

```json
[
  {
    "id": 1,
    "name": "Books"
  },
  {
    "id": 2,
    "name": "Movies"
  }
]
```

Category form type

```php
<?php

namespace App\Form\Type;

use App\Entity\Category;

final class CategoryType extends AbstractType
{
    {
        $builder->add('name', TextType::class, [
            'label' => 'Name',
            'required' => true,
            'empty_data' => '',
        ]);
    }

    {
        $resolver->setDefaults([
            'data_class' => Category::class,
        ]);
    }
}
```

Usage in Live Component in instantiateForm method

```php
{
    return $this->formFactory->createNamed(
        name: 'form',
        type: CollectionType::class,
        data: $this->categories,
        options: [
            'label' => false,
            'entry_type' => CategoryType::class,
            'allow_add' => true,
            'allow_delete' => true,
            'delete_empty' => false,
            'prototype' => false,
        ]
    );
}
```

Initial data in LiveComponent formValues property

```json
[
  {
    "name": "Books"
  },
  {
    "name": "Movies"
  }
]
```

Then I perform swap via your proposed method. This will give me these formValues

```json
[
  {
    "name": "Movies"
  },
  {
    "name": "Books"
  }
]
```

After calling submitForm method in Live Component this is state of my entities

```json
[
  {
    "id": 1,
    "name": "Movies"
  },
  {
    "id": 2,
    "name": "Books"
  }
]
```

As you can see, it just "swapped" data which are exposed to the form. Let say that I had some products in Books category and now they are all in category called Movies!

IMHO this proposed swapping algorithm can be used only in very simple scenarios and DEVs need to be aware of its quirks and limitations.

For true swapping you need at least one extra property, which holds item order.

smnandre:
I tend to agree with @Pechynho here. 

I'm also a bit worried swapping names / ids of form elements may have other side effects on the front, especially for Stimulus. Same for re-rendering 

> For true swapping you need at least one extra property, which holds item order.

Not even sure here, as if you did not have one on the component swapped, i'm not sure they will rerender

smnandre:
@haivala do you want to try another approach ?

@Pechynho what would you recon: add an hidden field on each item dynamically ?

Pechynho:
@smnandre hidden field could work. But there is need to order `child_form` before rendering them in Twig.

Something like this:

```twig
            {% set child_forms = [] %}
            {% for child_form in form.myCollection %}
                {% set child_forms = child_forms|merge({form: child_form, order: child_form.vars.data.order}) %}
            {% endfor %}
            {% for child_form in child_forms|sort((a, b) => a.order <=> b.order)|column('form') %}
                {{ form_row(child_form) }}
            {% endfor %}
```

Ofc, the order property name would have to be configurable.

Pechynho:
Let's say that API of LiveCollectionType would look like this.

```php
{
    return $this->formFactory->createNamed(
        name: 'form',
        type: LiveCollectionType::class,
        data: $this->categories,
        options: [
            'label' => false,
            'entry_type' => CategoryType::class,
            'allow_add' => true,
            'allow_delete' => true,
            'delete_empty' => false,
            'prototype' => false,
            'order_property_name' => 'order'
       ]);
}
```

Are we able to add 'order' hidden field automatically to `CategoryType` ? I don't think that it's possible and it would have to be provided by DEVs.

Other approach is not to modify `CategoryType` at all and store orders value in some internal LiveProp? And then set this value to object via FormEvent?

kbond:
I also think a "position" property is required. I've done this with doctrine-extensions (sortable) and raw js.

smnandre:
Yep, and it would be much easier with JS, because we could change the order visually while not changing the order in HTTP response (and prevent all side effects).

Not sure there is a real possibility here (and people with small arrays or so can always use custom data actions)

haivala:
When using the sorting feature with objects/entities:
 
 1. Simple array sorting (default):
    - Only swaps values in the form data array
    - Can cause data association issues after form submission
    - Should only be used for simple value collections
 
 2. Order property sorting (recommended):
    - Uses a dedicated property to track item order
    - Preserves object identity while changing only the order
    - Example:
      $builder->add('items', LiveCollectionType::class, [
          'entry_type' => ItemType::class,
          'allow_sort' => true,
          'order_property_path' => 'position',
      ]);
    - Your entity needs to have a 'position' property!

src/LiveComponent/templates/form_theme.html.twig
```diff
@@ -6,8 +6,14 @@
...
+    {%- if button_move_up is defined and not button_move_up.rendered -%}
+        {{ form_row(button_move_up) }}
+    {%- endif -%}
+    {%- if button_move_down is defined and not button_move_down.rendered -%}
+        {{ form_row(button_move_down) }}
+    {%- endif -%}
     {%- if button_delete is defined and not button_delete.rendered -%}
         {{ form_row(button_delete) }}
     {%- endif -%}
+    {{ block('form_row') }}
```
smnandre, line 18: This would be a BC break as it would set the delete button before, when is was after for people already using this theme.

src/LiveComponent/src/LiveCollectionTrait.php
```diff
@@ -46,6 +46,70 @@ public function removeCollectionItem(PropertyAccessorInterface $propertyAccessor
...
+            $propertyAccessor->setValue($data[$index], $orderPropertyPath, $prevItemOrder);
+            $propertyAccessor->setValue($data[$index - 1], $orderPropertyPath, $currentItemOrder);
+        } else {
+            // Legacy behavior - simple array swap (with warning about limitations)
+            $temp = $data[$index - 1];
+            $data[$index - 1] = $data[$index];
+            $data[$index] = $temp;
+        }
+
+        $propertyAccessor->setValue($this->formValues, $propertyPath, $data);
```
smnandre, line 80: Did you try it ? I feel like the buttons rendered next would be desynchronised and their key would send incorrect data to the live action ?

src/LiveComponent/src/LiveCollectionTrait.php
```diff
@@ -46,6 +46,70 @@ public function removeCollectionItem(PropertyAccessorInterface $propertyAccessor
...
 
+    #[LiveAction]
+    public function moveCollectionItemUp(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name, #[LiveArg] int $index): void
+    {
+        if ($index <= 0) {
+            return;
+        }
+
+        $propertyPath = $this->fieldNameToPropertyPath($name, $this->formName);
+        $data = $propertyAccessor->getValue($this->formValues, $propertyPath);
```
smnandre, line 57: After some thinking... I wonder if this should be a dedicated trait... 

Because right now someone using the trait will automatically have these methods available, even if they just want the add/delete (and reciprocally)

src/LiveComponent/src/LiveCollectionTrait.php
```diff
@@ -46,6 +46,70 @@ public function removeCollectionItem(PropertyAccessorInterface $propertyAccessor
...
+
+    #[LiveAction]
+    public function moveCollectionItemDown(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name, #[LiveArg] int $index): void
+    {
+        $propertyPath = $this->fieldNameToPropertyPath($name, $this->formName);
+        $data = $propertyAccessor->getValue($this->formValues, $propertyPath);
+
+        if (!\is_array($data) || !isset($data[$index]) || !isset($data[$index + 1])) {
+            return;
+        }
```
smnandre, line 91: This is not compatible with the add/remove behaviour for now..

The add set the new value as the max of the current ones + 1, but the remove simply unset it.

So if you start with

0 A
1 B
2 C

then delete B

0 A
2 C

If you move up A or move down C you will not find the previous / next item.

Kocal:

Thanks!

Status: Needs Work

PR #2689, opened 2025-04-10
[LiveComponent] Release the loading state before any redirect.

sukei:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no 
| Issues        | 
| License       | MIT

This PR attempt to release the loading state before redirecting the user. This one may seems a little weird at first, but it has its use case. 

I just write an export feature that uses a LiveComponent action to build a file. While the export is made, the UI add some classes and/or attributes to convey the "in progress" meaning. When it's done, the user is redirected to a specific controller that initiate the download. Since the redirected page just provide a file attachment, the user stay on the original page (fully expected). But, at this stage, the loading state is never released. By releasing the loading state before the redirect, the UI is made responsive again.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -549,8 +549,8 @@ var HookManager_default = class {
   }
 };
 
-// ../../../node_modules/.pnpm/idiomorph@0.3.0/node_modules/idiomorph/dist/idiomorph.esm.js
-var Idiomorph = function() {
+// ../../../node_modules/idiomorph/dist/idiomorph.esm.js
+var Idiomorph = (function() {
   "use strict";
   let EMPTY_SET = /* @__PURE__ */ new Set();
   let defaults = {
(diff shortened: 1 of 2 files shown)


src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -2152,6 +2152,7 @@ class Component {
         if (!controls.shouldRender) {
             return;
         }
+        this.hooks.triggerHook('loading.state:finished', this.element);
         if (backendResponse.response.headers.get('Location')) {
             if (this.isTurboEnabled()) {
                 Turbo.visit(backendResponse.response.headers.get('Location'));
```
smnandre: I'm not sure we want to trigger the "finished" event _before_ Turbo does its work.

Here you could do something just before the 

```js
window.location.href = backendResponse.response.headers.get('Location') || '';
```

sukei: I use Turbo on this project and I need it to be triggered in that setup too. BTW, I'm actualy using this patch and it didn't seems to create any issues.

EDIT : I disabled Turbo in that case to avoid a double fetch of the URL, that may be the reason I didn't have any issue. So WDYT? Should I move the call?

sukei:
Rebase on 3.x, still apply.

PR #2692, opened 2025-04-11
[Turbo] Minor documentation syntax fixes

seb-jean:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes 
| Issues        | Fix #...
| License       | MIT

Fixed documentation regarding blocks when using broadcast. Blocks are not required.

### DIFF
--- a/src/Turbo/doc/index.rst
+++ b/src/Turbo/doc/index.rst
@@ -901,7 +901,7 @@ The ``Broadcast`` attribute comes with a set of handy options:
 -  ``template`` (``string``): Twig template to render (see above)
 
 The ``Broadcast`` attribute can be repeated (e.g. you can have multiple
-`#[Broadcast]`. This is convenient to render several templates associated with
+``#[Broadcast]``. This is convenient to render several templates associated with
 their own topics for the same change (e.g. the same data is rendered in different
 way in the list and in the detail pages).
 
@@ -914,7 +914,7 @@ are supported:
 -  ``sse_retry`` (``int``): ``retry`` field of the SSE
 
 The Mercure broadcaster also supports `Expression Language`_ in topics
-by starting with `@=`.
+by starting with ``@=``.
 
 Example::
 


src/Turbo/doc/index.rst
```diff
@@ -838,9 +838,8 @@ created, modified or deleted:
     {% endblock %}
 
 By convention, Symfony UX Turbo will look for a template named
-``templates/broadcast/{ClassName}.stream.html.twig``. This template
-**must** contain at least 3 blocks: ``create``, ``update`` and
-``remove`` (they can be empty, but they must exist).
+``templates/broadcast/{ClassName}.stream.html.twig``. This template can contain 3 blocks: ``create`` block for entity creation. ``update`` block for entity update. ``remove`` block for entity deletion.
```
smnandre: When did this change ?

seb-jean: I don't know when it changed but I only tested with the create block in my template and I didn't encounter any errors.

smnandre: Even during an update and a remove broadcast (just to be sure) ?

seb-jean: Exactly, it also works for `update` and `remove`.

seb-jean: And of course, they are mandatory when creating, deleting or updating, otherwise an error is returned.

smnandre: Wait so they "can" or they "must" ?

seb-jean: `create` block must be present if it's a creation.
`update` block must be present if it's an update.
`remove` block must be present if it's a deletion.

For example, for my `Book` entity, I need to render a template on creation and update. I should have the following code:

```twig
{# templates/broadcast/Book.stream.html.twig #}
{% block create %}
  [...]
{% endblock %}

{% block update %}
  [...]
{% endblock %}
```

The `remove` block isn't present, and it works.

With the current documentation, the developer will think it's necessary to do this:

```twig
{# templates/broadcast/Book.stream.html.twig #}
{% block create %}
  [...]
{% endblock %}

{% block update %}
  [...]
{% endblock %}

{% block remove %}
{% endblock %}
```

With an empty `remove` block, but this isn't necessary.

smnandre: I'm not sure i agree with this way of presenting thing. Their entity can be deleted or updated. If this happens and they did not put these blocks here, their app will crash. 

I'd say we need either a way to explain it very clearly, or let's keep the current formulation.

WDYT ?

seb-jean: Yeah ok, let's keep the current formulation then :).

Kocal (approved):

Kocal:
Thank you @seb-jean.

PR #2698, opened 2025-04-21
[Docs] Add step to disable `package.json` synchronization while upgrading PHP package

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

Following https://github.com/symfony/symfony/issues/58678#issuecomment-2817348518

### DIFF
--- a/src/Autocomplete/assets/README.md
+++ b/src/Autocomplete/assets/README.md
@@ -14,6 +14,8 @@ composer require symfony/ux-autocomplete:2.23.0
 npm add @symfony/ux-autocomplete@2.23.0
 ```
 
+**Tip:** Your `package.json` file will be automatically modified by [Flex](https://github.com/symfony/flex) when installing or upgrading a PHP package. To prevent this behavior, ensure to **use at least Flex 1.22.0 or 2.5.0**, and run `composer config extra.symfony.flex.synchronize_package_json false`.
+
 ## Resources
 
 -   [Documentation](https://symfony.com/bundles/ux-autocomplete/current/index.html)
--- a/src/Chartjs/assets/README.md
+++ b/src/Chartjs/assets/README.md
@@ -14,6 +14,8 @@ composer require symfony/ux-chartjs:2.23.0
 npm add @symfony/ux-chartjs@2.23.0
 ```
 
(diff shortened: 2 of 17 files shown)


kbond (approved):

src/Autocomplete/assets/README.md
```diff
@@ -14,6 +14,18 @@ composer require symfony/ux-autocomplete:2.23.0
 npm add @symfony/ux-autocomplete@2.23.0
 ```
 
+To prevent your `package.json` file from being overwritten by Symfony Flex when upgrading the PHP package, you must **use at least Symfony Flex ^1.22.0 or ^2.5.0**, and configure your `composer.json` file like this:
```
94noni: What about using the github markdown note ?

94noni (approved):

smnandre:
I'm not sure here... as always by fear to make newcomers make mistake ... once this is enabled, installing packages will not work anymore, and then no documentation would help :|

src/Autocomplete/assets/README.md
Kocal: Does it work on NPM website?

Kocal:
> I'm not sure here... as always by fear to make newcomers make mistake ... once this is enabled, installing packages will not work anymore, and then no documentation would help :|

I see what you mean, but remember our UX npm packages are reserved for **advanced users** only, not newcomers.

smnandre:
The problem is want to avoid is someone having no idea locally why they cannot install an UX package anymore. 

I do believe this is something we should address asap.. 


In the meantime i understand the need, so can we just add a warning ? Something like "Please be aware NONE of your installed paclages NOR the next you would install .... " (or something like that) ?

smnandre (approved):

Kocal:
> The problem is want to avoid is someone having no idea locally why they cannot install an UX package anymore.

A warning is displayed to the user when `synchronize_package_json` is false: https://github.com/symfony/flex/blob/8ce1acd9842abe0e9b4c4a0bd3f259859516c018/src/Flex.php#L506

It should be enough no?

> In the meantime i understand the need, so can we just add a warning ? Something like "Please be aware NONE of your installed paclages NOR the next you would install .... " (or something like that) ?

I will rewrite the paragraph to be more explicit about that

Kocal:
PR updated

src/Autocomplete/assets/README.md
94noni: I honestly dont know (same for packagist and other sf php repo readme)
But at least Github display info is clearer and denote this specific paragraph

PR #2699, opened 2025-04-27
key is more reliable with child components

### LINKED ISSUE #1643: Error: The data-live-preserve attribute requires an id attribute
Using an `id` prop on a nested LiveComponent causes an error when child model is updated from parent.

### Error:

`Error: The data-live-preserve attribute requires an id attribute to be set on the element at live_controller.js:1353:1`

### How to reproduce
See https://github.com/kachnitel/symfony-live-component-bug/commit/9d28445789bf758be0710059e5549001f8ead3b0

- clone repo and install dependencies
- start server and go to homepage
- click the (small and empty) button to call an action in parent component to update data model
- witness error

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | yes
| Issues        | Fix #1643 
| License       | MIT

I couldn't track down why, but for child components, `key` is more reliable. When `id` is used and a child requires re-rendering (because input props changed), the dummy element returned by https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/EventListener/InterceptChildComponentRenderSubscriber.php will not have an `id`, but with `key` it will. Smells like a bug, but `key` works

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3237,14 +3237,14 @@ Rendering Quirks with List of Elements
 
 If you're rendering a list of elements in your component, to help LiveComponents
 understand which element is which between re-renders (i.e. if something re-orders
-or removes some of those elements), you can add a ``id`` attribute to
+or removes some of those elements), you can add a ``key`` attribute to
 each element
 
 .. code-block:: html+twig
 
     {# templates/components/Invoice.html.twig #}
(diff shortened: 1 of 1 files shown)


smnandre:
Do you have a concrete example/setup so we can track this done later ?

smnandre (approved):

Kocal:
Thank you @weaverryan.

PR #2702, opened 2025-04-29
[StimulusBundle] Skip mapping .ts controller if .js version is available

### LINKED ISSUE #2701: [StimulusBundle] Exception thrown when `.ts` controller transpiled to `.js` in the same directory
The [`loadCustomControllers`](https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php#L64) function iterates over all `.ts` and `.js` files in the project's controllers path to build a controller map.

My project has a build step where TypeScript controllers are transpiled to JavaScript files that are saved alongside the source TypeScript files. My asset mapper config excludes TypeScript files since I'm only interested in serving JavaScript.

When `loadCustomControllers` iterates over all `.ts` and `.js` files in my controllers path, it tries to load mapped assets for my `.ts` controllers. But since these files are excluded, the returned asset is `null` and an exception is thrown when it tries to read the `sourcePath` property on the `null` asset.

Controller files *should* be a mapped asset. Not finding one indicates a problem so an exception *should* be thrown (e.g. see [`null` mapped asset check](https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php#L134) in [`loadUxControllers`](https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php#L92) function). However the `loadCustomControllers` function doesn't realize there are `.js` and `.ts` versions of the same controller and can skip processing the missing version.

I expected that I could exclude `.ts` controller files when `.js` versions are available.

Here's my proposed fix:

- Skip mapping `.ts` controller file if a `.js` version is available
- Add a null asset check and throw an exception similar to one thrown in `loadUxControllers`

Related: #1210

ghost:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2701  
| License       | MIT

Resolves issue #2701 where an exception is thrown when `.ts` controllers are transpiled to `.js` in the same directory and `.ts` files are excluded from mapped assets. See related issue for details.

### DIFF
--- a/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
+++ b/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
@@ -70,6 +70,11 @@ class ControllersMapGenerator
 
         $controllersMap = [];
         foreach ($finder as $file) {
+            // Skip .ts controller if .js version is available
+            if ('ts' === $file->getExtension() && file_exists(substr($file->getRealPath(), 0, -2).'js')) {
+                continue;
+            }
+
             $name = $file->getRelativePathname();
             // use regex to extract 'controller'-postfix including extension
             preg_match(self::FILENAME_REGEX, $name, $matches);
@@ -77,6 +82,10 @@ class ControllersMapGenerator
             $name = str_replace(['_', '/', '\\'], ['-', '--', '--'], $name);
 
             $asset = $this->assetMapper->getAssetFromSourcePath($file->getRealPath());
+            if (!$asset) {
+                throw new \RuntimeException(\sprintf('Could not find an asset mapper path that points to the "%s" controller.', $name));
+            }
+
             $content = file_get_contents($asset->sourcePath);
             $isLazy = preg_match('/\/\*\s*stimulusFetch:\s*\'lazy\'\s*\*\//i', $content);
 
--- a/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php
+++ b/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php
@@ -34,6 +34,8 @@ class ControllersMapGeneratorTest extends TestCase
                     $logicalPath = 'fake-vendor/ux-package1/package-controller-second.js';
                 } elseif (str_ends_with($path, 'package-hello-controller.js')) {
                     $logicalPath = 'fake-vendor/ux-package2/package-hello-controller.js';
+                } elseif (str_ends_with($path, 'other-controller.ts') || str_ends_with($path, 'excluded-controller.js')) {
+                    return null;
                 } else {
                     // replace windows slashes
                     $path = str_replace('\\', '/', $path);
@@ -75,15 +77,18 @@ class ControllersMapGeneratorTest extends TestCase
             $autoImportLocator,
         );
 
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage('Could not find an asset mapper path that points to the "excluded" controller.');
         $map = $generator->getControllersMap();
         // + 3 controller.json UX controllers
         // - 1 controllers.json UX controller is disabled
-        // + 10 custom controllers (1 file is not a controller & 1 is overridden)
-        $this->assertCount(12, $map);
+        // + 11 custom controllers (1 file is not a controller, 1 is overridden)
+        $this->assertCount(13, $map);
         $packageNames = array_keys($map);
         sort($packageNames);
         $this->assertSame([
             'bye',
+            'excluded',
             'fake-vendor--ux-package1--controller-second',
             'fake-vendor--ux-package2--hello-controller',
             'hello',
--- a/src/StimulusBundle/tests/AssetMapper/StimulusControllerLoaderFunctionalTest.php
+++ b/src/StimulusBundle/tests/AssetMapper/StimulusControllerLoaderFunctionalTest.php
@@ -63,7 +63,8 @@ class StimulusControllerLoaderFunctionalTest extends WebTestCase
                 // 2x from UX packages, which are enabled in controllers.json
                 '/assets/fake-vendor/ux-package1/package-controller-second.js',
                 '/assets/fake-vendor/ux-package2/package-hello-controller.js',
-                // 3x from more-controllers
+                // 4x from more-controllers
+                '/assets/more-controllers/excluded-controller.js',
                 '/assets/more-controllers/hello-controller.js',
                 '/assets/more-controllers/minified-controller.js',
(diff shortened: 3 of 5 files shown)


src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
```diff
@@ -70,13 +70,22 @@ private function loadCustomControllers(): array
 
         $controllersMap = [];
         foreach ($finder as $file) {
+            // Skip .ts controller if .js version is available
```
smnandre, line 73: Seen the code just belove, what about we sort files `by name `ascending (`.js` would be before `.ts`)

That way:

```
foo-bar-controller.js
foo-bar-controller.ts
foo_bar-controller.js
foo_bar-controller.ts
```

would only register the first one in the controller map, sa they would all have "foo-bar" as name.


We could
* fix your problem
* avoid open files when not needed
* enforce same order / file content 


```diff
$finder->in($this->controllerPaths)
            ->files()
+            ->sortByName()
            ->name(self::FILENAME_REGEX);

        $controllersMap = [];
        foreach ($finder as $file) {

             $name = $file->getRelativePathname();
             // ... name / preg match / replace / ..
             
+            if (isset($controllersMap[$name]) {
+                continue
+             }
             
             $asset = $this->assetMapper->getAssetFromSourcePath($file->getRealPath());
             // ...
```

ghost: I had a similar idea but later found out that controllers should be overridable. There is a test which checks for this feature:

https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php#L115 

So skipping previously mapped controllers with the same name (`if (isset($controllersMap[$name])) continue;`) breaks that test since it doesn't allow controllers to be overridden.

And since controllers can be overridden, we probably don't want to `sortByName()` as that may break the override order given in the controller paths. 

For example, given the following `config/packages/stimulus.yaml`:

```yaml
# config/packages/stimulus.yaml
stimulus:
    controller_paths:
        - '%kernel.project_dir%/controllers_z'
        - '%kernel.project_dir%/controllers_a'
```

Then `controllers_a/hello-controller.js` would override `controllers_z/hello-controller.js`. But if we call `sortByName()` then `controllers_z/hello-controller.js` would override `controllers_a/hello-controller.js`. So devs could no longer specify controller override order in `stimulus.yaml`. 

When you say:

> * avoid open files when not needed

Are you referring to the `file_exists` call? I haven't looked at the PHP source code but [according to this page](https://flatcoding.com/tutorials/php/php-file-exists-function/#:~:text=The%20file_exists%20doesn't%20actually,is%20there%20where%20you%20specified.): *The file_exists doesn’t actually open the file. It just checks the file system to see if the file or directory is there where you specified.* I'm assuming `file_exists` queries the filesystem instead of trying to allocate a file descriptor/read the file, so it shouldn't lock the file or otherwise consume resources. Also the [`loadCustomControllers`](https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php#L64) function is called once, either when running `console asset-map:compile` or when building the dev/test cache, so any performance impact should be negligible.

Another nice thing about `if ($file->getExtension() === 'ts' && file_exists())` solution is the behaviour is explicit and localized to those few lines, so the `loadCustomControllers` function is easier to understand and refactor.

Kocal: I prefer the solution `if ($file->getExtension() === 'ts' && file_exists())` which is more explicit than a `sortByName()` that works "by miracle" (I mean, without showing the intention behind)

ghost: When you say:

> avoid open files when not needed

Ah, I think I know what you mean now, i.e. skipping previously mapped controllers would bypass

https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php#L80

The `file_exists()` solution will also avoid opening files unecessarily.

src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
```diff
@@ -70,13 +70,22 @@ private function loadCustomControllers(): array
...
             $name = $file->getRelativePathname();
             // use regex to extract 'controller'-postfix including extension
             preg_match(self::FILENAME_REGEX, $name, $matches);
             $name = str_replace(['_'.$matches[1], '-'.$matches[1]], '', $name);
             $name = str_replace(['_', '/', '\\'], ['-', '--', '--'], $name);
 
             $asset = $this->assetMapper->getAssetFromSourcePath($file->getRealPath());
+            if (!$asset) {
+                throw new \RuntimeException(\sprintf('Could not find an asset mapper path that points to the "%s" controller.', $name));
+            }
```
smnandre, line 87: Let's remove this... as there is no other situation it could happen right ? Or let's add a test if possible

src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
smnandre: Ol you guys focused on the TS thing, but i was talking more generally of names vs path, seen the str_replace two lines later.

Nevermind, i'll open a PR after this one :)

smnandre (approved):

src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
ghost: This can happen anytime there are `.js`/`.ts` files in the controllers path that are excluded by the asset mapper config. This exception would be helpful since it indicates a problem with the asset mapper config. There's a similar check in the `loadUxControllers()` function, which I copied for consistency. I've added a test for this situation.

https://github.com/symfony/ux/blob/61819ae6c29f5bc6c159c83bf801cef9c0227c82/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php#L135

Kocal (approved):

Kocal:
Thank you @phasdev.

PR #2704, opened 2025-04-29
[Dropzone] Improve display with multiple files

sblondeau:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Actually, the dropzone will only show one file, even if it allows multiple files to be uploaded. This makes it difficult to see which file has been added to the dropzone.
This PR adds a preview of multiple files (not just the first). There are several previous issues or PRs starting to achieve this (https://github.com/symfony/ux/pull/2642.) but with a different approach.

image

If the dropzone is small or there are many files, they will be displayed on a multiline grid.
image

If the file is NOT an image, a simple file icon will be displayed (to keep the grid aligned and show all file names).
image

The style has also been modernised (effect on hover or drop, new icon for clear button) and some CSS variables added to the style file.


smnandre:
Not seeing the images :/

src/Dropzone/assets/test/controller.test.ts
```diff
@@ -33,7 +33,7 @@ describe('DropzoneController', () => {
 
     beforeEach(() => {
         container = mountDOM(`
-            <div class="dropzone-container" data-controller="check dropzone" data-testid="container"> 
+            <div class="dropzone-container" data-controller="check dropzone" data-dropzone-target="container" data-testid="container">
```
smnandre:
```suggestion
            <div class="dropzone-container" data-controller="check dropzone" data-dropzone-target="container" data-testid="container"> 
```

To revert :)

src/Dropzone/templates/form_theme.html.twig
```diff
@@ -2,23 +2,15 @@
...
+    <div class="dropzone-container" data-controller="{{ dataController }}" data-symfony--ux-dropzone--dropzone-target="container">
         <input type="file" {{ block('widget_attributes') }} data-symfony--ux-dropzone--dropzone-target="input" />
-
         <div class="dropzone-placeholder" data-symfony--ux-dropzone--dropzone-target="placeholder">
             {%- if attr.placeholder is defined and attr.placeholder is not none -%}
                 {{- translation_domain is same as(false) ? attr.placeholder : attr.placeholder|trans({}, translation_domain) -}}
             {%- endif -%}
         </div>
-
+        <button class="dropzone-preview-button" type="button" data-symfony--ux-dropzone--dropzone-target="previewClearButton"></button>
```
smnandre: This is problematic, as people would have a new button that appears on their forms / UI..

I think  we do need to create a new one

src/Dropzone/templates/form_theme.html.twig
```diff
@@ -2,23 +2,15 @@
     {%- set dataController = (attr['data-controller']|default('') ~ ' symfony--ux-dropzone--dropzone')|trim -%}
     {%- set attr = attr|merge({'data-controller': '', class: (attr.class|default('') ~ ' dropzone-input')|trim}) -%}
 
-    <div class="dropzone-container" data-controller="{{ dataController }}">
+    <div class="dropzone-container" data-controller="{{ dataController }}" data-symfony--ux-dropzone--dropzone-target="container">
```
smnandre: To revert

src/Dropzone/assets/src/style.css
```diff
@@ -1,72 +1,130 @@
...
+
+  --dropzone-spacing: 8px;
+  --dropzone-radius: 8px;
+
+  --dropzone-width: cacl(100% - 2 * var(--dropzone-spacing));
+  --dropzone-height: 120px;
+  --dropzone-image-size: 100px;
+
+  --dropzone-file-svg: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill="none" stroke="currentColor" stroke-linejoin="round" stroke-width="2" d="M10 3v4a1 1 0 0 1-1 1H5m14-4v16a1 1 0 0 1-1 1H6a1 1 0 0 1-1-1V7.914a1 1 0 0 1 .293-.707l3.914-3.914A1 1 0 0 1 9.914 3H18a1 1 0 0 1 1 1Z"/></svg>');
+  --dropzone-close-svg: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 24 24"><path fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="m15 9l-6 6m0-6l6 6m6-3a9 9 0 1 1-18 0a9 9 0 0 1 18 0"/></svg>');
```
smnandre: I think this would be better in the theme, not sure everyone want this

src/Dropzone/assets/src/controller.ts
```diff
@@ -117,6 +149,7 @@ export default class extends Controller {
             this.inputTarget.style.display = 'none';
             this.placeholderTarget.style.display = 'none';
             this.previewTarget.style.display = 'block';
+            this.containerTarget.classList.add('dropzone-on-drag-leave');
```
smnandre: Not sur about these class names 🤷‍♂️

src/Dropzone/assets/src/controller.ts
```diff
@@ -16,8 +16,17 @@ export default class extends Controller {
     declare readonly previewClearButtonTarget: HTMLButtonElement;
     declare readonly previewFilenameTarget: HTMLDivElement;
     declare readonly previewImageTarget: HTMLDivElement;
-
-    static targets = ['input', 'placeholder', 'preview', 'previewClearButton', 'previewFilename', 'previewImage'];
+    declare readonly containerTarget: HTMLDivElement;
+
+    static targets = [
+        'container',
```
smnandre: You don't need this... the container is "this.element"

src/Dropzone/assets/src/controller.ts
```diff
@@ -56,38 +65,60 @@ export default class extends Controller {
...
+
+            // Create an image preview element if the file is an image, else a default svg file icon
+            if (file.type) {
+                const imagePreviewElement = document.createElement('div');
+                if (file.type.indexOf('image') !== -1) {
+                    imagePreviewElement.classList.add('dropzone-preview-image');
+                    this._populateImagePreview(file, imagePreviewElement);
+                } else {
+                    imagePreviewElement.classList.add('dropzone-preview-svg');
+                }
```
smnandre: Why svg ?

smnandre (approved):

smnandre (commented):
(sorry misslicked on approve)

src/Dropzone/assets/src/controller.ts
sblondeau: it is the svg "file" icon when file is not an image

src/Dropzone/assets/src/controller.ts
sblondeau: hum, the goal is to allow style on drag leave / drag enter (e.g change color, transition...). I thought it was the most understable name :D but if you have another suggestion no problem ;-)

src/Dropzone/assets/src/style.css
sblondeau: Not sure to understand; actually the "clear" button is a simple `content:-"x"` in CSS (not sure everyone want this :p), IMO, SVG is a better solution, and with the CSS variable you can easily custom it.

src/Dropzone/templates/form_theme.html.twig
sblondeau: But this is the same actual "clear" button, just move it outside the "preview" div. 
Behavior is the same, you have a small X "close" button (but look nicer with a svg) at the top right only when you have file in the dropzone.

smnandre: I understand and I'd be happy to offer a nicer button believe me (you would need to change it's class tbf but that is a detail).

What i'm refering to is big BC break. If they used and customized the CSS, you are changing the HTML of their form doing so, and this can break websites.

So the idea would be to either 
- offer a new theme
- release this new feature in new major version

src/Dropzone/assets/src/style.css
smnandre: Using SVGs as data URLs (data:image/svg+xml;base64,...) is often restricted in enterprise environments. This is mainly due to security policies like CSP (Content Security Policy) that block data: schemes or inline content. Encoded SVGs can also contain JavaScript or CSS, which raises concerns about XSS vulnerabilities. Additionally, data URLs bypass HTTP-level caching and CDN optimization, making them less efficient for large-scale or public-facing applications.

For better compatibility and maintainability, it’s generally recommended to use static SVG files served over HTTP, inline <svg> elements (without embedded scripts or styles), or SVG sprites.

src/Dropzone/assets/src/controller.ts
smnandre: > I thought it was the most understable name :D

Name maybe, but code not so... Did I miss something, or  theses classes are never removed ? If so the animation would only work once.

src/Dropzone/assets/src/controller.ts
smnandre: Ok so here I think you should not expose this as it's not the "semantic" here, and we don't want user to feel they need an svg.

What about dropzone-preview-file or dropzone-no-preview .

You get the idea ?

src/Dropzone/assets/src/style.css
sblondeau: I moved svg to js and html files respectively

src/Dropzone/assets/src/controller.ts
sblondeau: fixed

src/Dropzone/assets/src/style.css
```diff
@@ -1,72 +1,121 @@
...
-    flex-grow: 1;
-    text-align: center;
-    color: #999;
+  flex-grow: 1;
+  text-align: center;
+}
+
+.dropzone-on-drag-leave {
+  background-color: var(--dropzone-background);
+  transition: 0.3s;
```
smnandre, line 120: If this is the only reason to add the 2 classes, maybe it's not 100% required, wdyt ? :) 

(the same thing can be done with `:has` selectors)

src/Dropzone/templates/form_theme.html.twig
```diff
@@ -1,24 +1,17 @@
 {% block dropzone_widget -%}
-    {%- set dataController = (attr['data-controller']|default('') ~ ' symfony--ux-dropzone--dropzone')|trim -%}
-    {%- set attr = attr|merge({'data-controller': '', class: (attr.class|default('') ~ ' dropzone-input')|trim}) -%}
+{%- set dataController = (attr['data-controller']|default('') ~ ' symfony--ux-dropzone--dropzone')|trim -%}
```
smnandre, line 2: Could you revert the indentation as it was ? 😄  

It's to avoid unnecessary changes, and ease the review / maintain process (now and after i mean)

src/Dropzone/assets/src/controller.ts
```diff
@@ -56,48 +56,75 @@ export default class extends Controller {
...
+            // Create an image preview element if the file is an image, else a default svg file icon
+            if (file.type) {
+                const imagePreviewElement = document.createElement('div');
+
+                if (file.type.indexOf('image') !== -1) {
+                    imagePreviewElement.classList.add('dropzone-preview-image');
+                    this._populateImagePreview(file, imagePreviewElement);
+                } else {
+                    const noPreviewSvg =
+                        '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill="currentColor" d="M14 11a3 3 0 0 1-3-3V4H7a2 2 0 0 0-2 2v13a2 2 0 0 0 2 2h9a2 2 0 0 0 2-2v-8zm-2-3a2 2 0 0 0 2 2h3.59L12 4.41zM7 3h5l7 7v9a3 3 0 0 1-3 3H7a3 3 0 0 1-3-3V6a3 3 0 0 1 3-3"/></svg>';
```
smnandre, line 99: Let's keep SVG in HTML :) 

That way use can change them easily with their own form theme

src/Dropzone/tests/Form/DropzoneTypeTest.php
```diff
@@ -38,21 +38,14 @@ public function testRenderForm()
...
-            <button class="dropzone-preview-button" type="button"
-                    data-symfony--ux-dropzone--dropzone-target="previewClearButton"></button>
-
-            <div class="dropzone-preview-image" style="display: none"
-                 data-symfony--ux-dropzone--dropzone-target="previewImage"></div>
-
-            <div data-symfony--ux-dropzone--dropzone-target="previewFilename" class="dropzone-preview-filename"></div>
-        </div>
-    </div></div></div></form>
+'<form name="form" method="post" enctype="multipart/form-data"><div id="form"><div><label for="form_photo" class="required">Photo</label><div class="dropzone-container" data-controller="mydropzone symfony--ux-dropzone--dropzone">
```
smnandre, line 41: This is the proof we cannot do these changes without a dedicated form theme and/or controller js : the structural changes in the HTML are going to break layout/tests of people currently using this component

Kocal: 💯

Changing the DOM and/or styles like in this way is a no go.

smnandre:
Hi @sblondeau!

We're going to deprecate the current Dropzone... So we won't be accept new features/refactor like this for it at the moment.

**BUT**

If you (and others) are interested in working on a new image uploader .. one that would likely combine Dropzone, Cropper, and Uploader I have a lot of specs / ideas / scope

Would you like to be part of it?

sblondeau:
Hi @smnandre; yes I am interested :-)

PR #2705, opened 2025-04-29
[WIP] Update homepage install documenting both assetmapper and webpack

94noni:
WIP

### DIFF
--- a/ux.symfony.com/src/Model/RecipeFileTree.php
+++ b/ux.symfony.com/src/Model/RecipeFileTree.php
@@ -27,7 +27,6 @@ class RecipeFileTree
             ->addDirectory('assets/styles')
             ->addFile('assets/styles/app.css', 'Your main CSS file')
             ->addFile('package.json', 'Holds your node dependencies, most importantly Stimulus & Webpack Encore.')
-            ->addFile('webpack.config.js', 'Configuration file for Webpack Encore: the tool that processes and combines all of your CSS and JS files.')
         ;
     }
 


smnandre:
If we do this we should remove package.json too

94noni:
Would you like I take care of this ?
Boot a new bare project and only install stimulus and assetmapper
Thats mostly the rationale of this pr

Kocal:
> rfc: isnt webpack replaces by asset-mapper ?

No, both tools are equivalent but not identical at all, and fulfill different needs.

Symfony UX also targets Webpack Encore, not only the AssetMapper. 

I'm a little disappointed to see it disappear like this, and I would be in favor to re-add it on the homepage, something like:
```shell
# If you use the Symfony AssetMapper:
$ composer require symfony/asset-mapper symfony/stimulus-bundle

# Or, if you use Symfony Webpack Encore:
$ composer require symfony/webpack-encore symfony/stimulus-bundle
```

Keep files `package.json` and `webpack.config.js` but add a prefix like `When using Webpack Encore, ...`

WDYT?

94noni:
I got you, oki I will see if I can come with a proper rewrite for both usecases/installs 👍🏻
I was just surprise reading only asset mapper on the current composer install and seeing this webpack file :)

smnandre:
It has been removed for a reason: show the simplicity of install

As this page will soon_-ish_ be a relic of the past, i agree whatever you want here :)

94noni:
oki lets not do work for nothing, lets close :)

PR #2707, opened 2025-05-02
[StimulusBundle] Docs, replace `chart` examples by `hello` to avoid confusion with the ChartJS component

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

As ChartJS component exists, the examples using `chart` as controller identifier are confusing. ChartJS component does not expect `name` and `data` as values, but `view`.

Examples has been updated with a dead simple `hello world` style, which can not be confusing.

### DIFF
--- a/src/StimulusBundle/doc/index.rst
+++ b/src/StimulusBundle/doc/index.rst
@@ -174,15 +174,15 @@ For example:
 
 .. code-block:: html+twig
 
-    <div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }) }}>
+    <div {{ stimulus_controller('hello', { 'name': 'World', 'data': [1, 2, 3, 4] }) }}>
         Hello
     </div>
 
     <!-- would render -->
     <div
-       data-controller="chart"
-       data-chart-name-value="Likes"
-       data-chart-data-value="&#x5B;1,2,3,4&#x5D;"
+       data-controller="hello"
+       data-hello-name-value="World"
+       data-hello-data-value="&#x5B;1,2,3,4&#x5D;"
     >
        Hello
     </div>
@@ -191,22 +191,22 @@ If you want to set CSS classes:
 
 .. code-block:: html+twig
 
-    <div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }, { 'loading': 'spinner' }) }}>
+    <div {{ stimulus_controller('hello', { 'name': 'World', 'data': [1, 2, 3, 4] }, { 'loading': 'spinner' }) }}>
         Hello
     </div>
 
     <!-- would render -->
     <div
-       data-controller="chart"
-       data-chart-name-value="Likes"
-       data-chart-data-value="&#x5B;1,2,3,4&#x5D;"
-       data-chart-loading-class="spinner"
+       data-controller="hello"
+       data-hello-name-value="World"
+       data-hello-data-value="&#x5B;1,2,3,4&#x5D;"
+       data-hello-loading-class="spinner"
     >
        Hello
     </div>
 
     <!-- or without values -->
-    <div {{ stimulus_controller('chart', controllerClasses = { 'loading': 'spinner' }) }}>
+    <div {{ stimulus_controller('hello', controllerClasses: { 'loading': 'spinner' }) }}>
         Hello
     </div>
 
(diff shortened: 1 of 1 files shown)


Kocal (approved):
That's totally makes sense, thanks

smnandre:
Thanks for pointing that! I 100% agree for the confusion, but... we try to make examples that do have some sense and avoid the foo bars when possible.

Maybe one of these:
* a cart controller (classic example) with add action and productId ? 
* an alert with close ?
* a notification with link and action ? 

WDYT ?

welcoMattic:
@smnandre I would agree for advanced or "demo" examples (ones on ux.symfony.com). But on pure text documentation, I personally prefer to have easy to understand and dead simple examples like "Hello World". It's totally personal, if you prefer more advanced and real life example, I'm ok with that.

smnandre:
In fact, with its unusual syntax for some, Stimulus isn’t always easy to grasp.

The first examples here should be self-explanatory for most people, and I do believe we need to clearly emphasize what action, target, and controller mean.

If I read this, with the value name, I’m pretty sure it could lead to confusion:

```twig
<div {{ stimulus_controller('hello', { 'name': 'World', 'data': [1, 2, 3, 4] }) }}>
```

That said, your original point is totally valid.  I’ll send you another suggestion later today.

src/StimulusBundle/doc/index.rst
```diff
@@ -191,49 +191,49 @@ If you want to set CSS classes:
...
     </div>
 
     <!-- would render -->
     <div
-       data-controller="chart"
-       data-chart-name-value="Likes"
-       data-chart-data-value="&#x5B;1,2,3,4&#x5D;"
-       data-chart-loading-class="spinner"
+       data-controller="hello"
+       data-hello-name-value="Likes"
```
smnandre:
```suggestion
       data-hello-name-value="World"
```

src/StimulusBundle/doc/index.rst
```diff
@@ -191,49 +191,49 @@ If you want to set CSS classes:
...
 
     <!-- would render -->
     <div
-       data-controller="chart"
-       data-chart-name-value="Likes"
-       data-chart-data-value="&#x5B;1,2,3,4&#x5D;"
-       data-chart-loading-class="spinner"
-       data-chart-other-outlet=".target"
+       data-controller="hello"
+       data-hello-name-value="Likes"
```
smnandre:
```suggestion
       data-hello-name-value="World"
```

src/StimulusBundle/doc/index.rst
```diff
@@ -191,49 +191,49 @@ If you want to set CSS classes:
...
+       data-hello-data-value="&#x5B;1,2,3,4&#x5D;"
+       data-hello-loading-class="spinner"
+       data-hello-other-outlet=".target"
     >
        Hello
     </div>
 
     <!-- or without values/classes -->
-    <div {{ stimulus_controller('chart', controllerOutlets = { 'other': '.target' }) }}>
+    <div {{ stimulus_controller('hello', controllerOutlets = { 'other': '.target' }) }}>
```
smnandre:
```suggestion
    <div {{ stimulus_controller('hello', controllerOutlets: { 'other': '.target' }) }}>
```

src/StimulusBundle/doc/index.rst
```diff
@@ -191,49 +191,49 @@ If you want to set CSS classes:
...
+       data-hello-name-value="Likes"
+       data-hello-data-value="&#x5B;1,2,3,4&#x5D;"
+       data-hello-loading-class="spinner"
     >
        Hello
     </div>
 
     <!-- or without values -->
-    <div {{ stimulus_controller('chart', controllerClasses = { 'loading': 'spinner' }) }}>
+    <div {{ stimulus_controller('hello', controllerClasses = { 'loading': 'spinner' }) }}>
```
smnandre:
```suggestion
    <div {{ stimulus_controller('hello', controllerClasses: { 'loading': 'spinner' }) }}>
```

smnandre (approved):
Well there is a lot to change on this page...

A few very minor details on your PR and LGTM!

Thx @welcoMattic

Kocal:
Thank you @welcoMattic.

Issue #2708, opened 2025-05-03
Stimulus application doesn't start when loading @symfony/stimulus-bundle using version instead of path


amenophis:
Hi 👋 
To allows my renovate update frontend dependencies in sync with backend ones,
I changed my importmap like below, but the stimulus application did not start anymore.

```diff
    '@symfony/stimulus-bundle' => [
-       'path' => './vendor/symfony/stimulus-bundle/assets/dist/loader.js',
+       'version' => '2.24.0',
    ], 
```

I have no error on PHP side, nor in the browser console

Do you have an idea about this issue ?

amenophis:
After thinking, I don't have to specify the version here ... I mixed with another project that use yarn for frontend packaging.
Specifying the version with a package manager is mandatory to use tool like renovate or dependabot otherwise tool fail.

Nevertheless, do you think the issue using the version is normal or not ?

Kocal:
Salut :) 


Maybe I made a mistake when working on the npm-packagization of our UX packages, but I had come to the conclusion that the `@symfony/stimulus-bundle` NPM package shouldn't have been released.

I'll investigate what needs to be done.

Kocal:
Sorry, maybe there was some confusion between this issue and what we've already discussed privately.

Does your project use a `package.json`? Or did you just modify the `importmap.php` to change `path` to `version`, hoping for https://www.npmjs.com/package/@symfony/stimulus-bundle to be installed by the Symfony AssetMapper?

Kocal:
The line `import { isApplicationDebug, eagerControllers, lazyControllers } from './controllers.js';` is inlined / removed from https://cdn.jsdelivr.net/npm/@symfony/stimulus-bundle@2.24.0/dist/loader.min.js, the file downloaded when running `bin/console importmap:require @symfony/stimulus-bundle`. We must find a way to keep this line, and so this file.

The class [`StimulusLoaderJavaScriptCompiler`](https://github.com/symfony/ux/blob/2.x/src/StimulusBundle/src/AssetMapper/StimulusLoaderJavaScriptCompiler.php) must also been updated, in order to support file `%kernel.project_dir%/assets/vendor/@symfony/stimulus-bundle/dist/loader.js` (and also from node_modules)


**EDIT:** Since I've deleted the package from npm, the first link does not work anymore, so here is it:
```js
/**
 * Bundled by jsDelivr using Rollup v2.79.2 and Terser v5.39.0.
[...]

PR #2709, opened 2025-05-03
[StimulusBundle] Make the JS package private

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

This PR follows #2708

The npm package `@symfony/stimulus-bundle` was accidentally published on npm when I was working on the "Symfony UX on NPM" project, but when I realized it was to late to unpublish it.

Today, I was able to unpublish the package from npm. I'm now making the package **private**, so it won't be released anymore (until a solution is implemented)

I would be very happy to re-publish this package again when  https://github.com/symfony/ux/issues/2708#issuecomment-2848816170 will be fixed.

### DIFF
--- a/src/StimulusBundle/assets/README.md
+++ b/src/StimulusBundle/assets/README.md
@@ -2,9 +2,10 @@
 
 JavaScript assets of the [symfony/stimulus-bundle](https://packagist.org/packages/symfony/stimulus-bundle) PHP package.
 
-This package is private and is not intended to be published on npm or installed.
+## Installation
 
-If you're looking for a npm package to integrate Stimulus in your Symfony application, you may be looking for the [`@symfony/stimulus-bridge`](https://www.npmjs.com/package/@symfony/stimulus-bridge) package.
+Due to compatibility issues with JSDelivr causing the package not to work as expected, the package is not yet released on NPM.
+Read more at [symfony/ux#2708](https://github.com/symfony/ux/issues/2708).
 
 ## Resources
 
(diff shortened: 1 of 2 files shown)


smnandre:
Can you explain the scenario where there is a problem ? 

I'm not sure to follow here.

AFAIK, for AssetMapper there is no problem, as it is not concerned with renovate.

For webpack, the bridge deals with all this (on the project i'm working on, renovate works perfectly), and there is no entry for stimulus-bundle in the package.json

Kocal:
It's all explained here https://github.com/symfony/ux/issues/2708#issuecomment-2848816170, where the 
`StimulusLoaderJavascriptCompiler` is not compatible with the `loader.js` comming from the `node_modules`, or worst, when the file does not even exist with the package on jsDelivr. 

It's a temporary solution to prevent more issues.


> For webpack, the bridge deals with all this (on the project i'm working on, renovate works perfectly), and there is no entry for stimulus-bundle in the package.json

Yes, but here we are talking about users that don't use Encore, but the AssetMapper and importmap (with `@symfony/stimulus-bundle` installed from npm or jsDelivr)

smnandre:
> with @symfony/stimulus-bundle installed from npm or jsDelivr

Packages have been published on npm to resolve the path problem in packages.json, preventing renovate/github action to work.

With asset mapper, not using packages.json this is not a problem.

Also, this situation should never happen IRL so let's not take to much time / energy on this :)

PR #2712, opened 2025-05-05
[LiveComponent] Add events assertions in `InteractsWithLiveComponents`

Arkalo2:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | no
| License       | MIT

This MR simplify liveComponent event testing : 
Actually with the [Test Helper](https://symfony.com/bundles/ux-live-component/current/index.html#test-helper), we can't simply test whether an event has been dispatched from a component. 
This PR add 2 asserts : 

```php
$this-> assertComponentEmitEvent($render, 'event-name')->withData(['eventArg1' => $eventArg1]);
$this-> assertComponentEmitEvent($render, 'event-name')
    ->withDataSubset(['eventArg1' => $eventArg1])
    ->withDataSubset(['eventArg2' => $eventArg2])
;
```

```php
$this->assertComponentNotEmitEvent($render, 'event-name');
```

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3778,8 +3778,19 @@ uses Symfony's test client to render and make requests to your components::
             // emit live events
             $testComponent
                 ->emit('increaseEvent')
-                ->emit('increaseEvent', ['amount' => 2]) // emit a live event with arguments
+                ->emit('increaseEvent', ['amount' => 2, 'unit' => 'kg']) // emit a live event with arguments
             ;
+
+            // Assert that the event was emitted
+            $this->assertComponentEmitEvent($testComponent->render(), 'increaseEvent')
+                // optionally, you can assert that the event was emitted with specific data...
+                ->withData(['amount' => 2, 'unit' => 'kg'])
+                // ... or only with a subset of data
+                ->withDataSubset(['amount' => 2])
+            ;
+
+            // Assert that an event was not emitted
+            $this->assertComponentNotEmitEvent($testComponent->render(), 'decreaseEvent');
 
             // set live props
             $testComponent
--- a/src/LiveComponent/src/Test/InteractsWithLiveComponents.php
+++ b/src/LiveComponent/src/Test/InteractsWithLiveComponents.php
@@ -13,6 +13,7 @@ namespace Symfony\UX\LiveComponent\Test;
 
 use Symfony\Bundle\FrameworkBundle\KernelBrowser;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\UX\LiveComponent\Test\Util\AssertEmittedEvent;
 use Symfony\UX\TwigComponent\ComponentFactory;
 
 /**
@@ -43,5 +44,19 @@ trait InteractsWithLiveComponents
             self::getContainer()->get('ux.live_component.metadata_factory'),
             self::getContainer()->get('router'),
         );
+    }
+
+    protected function assertComponentEmitEvent(TestLiveComponent $testLiveComponent, string $expectedEventName): AssertEmittedEvent
+    {
+        $event = $testLiveComponent->getEmittedEvent($testLiveComponent->render(), $expectedEventName);
+
+        self::assertNotNull($event, \sprintf('The component "%s" did not emit event "%s".', $testLiveComponent->getName(), $expectedEventName));
+
+        return new AssertEmittedEvent($this, $event['event'], $event['data']);
+    }
+
+    protected function assertComponentNotEmitEvent(TestLiveComponent $testLiveComponent, string $eventName): void
+    {
+        self::assertNull($testLiveComponent->getEmittedEvent($testLiveComponent->render(), $eventName), \sprintf('The component "%s" did emit event "%s".', $testLiveComponent->getName(), $eventName));
     }
 }
--- a/src/LiveComponent/src/Test/TestLiveComponent.php
+++ b/src/LiveComponent/src/Test/TestLiveComponent.php
@@ -229,4 +229,39 @@ final class TestLiveComponent
 
         return $result;
     }
+
+    /**
+     * @return ?array{data: array<string, int|float|string|bool|null>, event: non-empty-string}
+     */
+    public function getEmittedEvent(RenderedComponent $render, string $eventName): ?array
+    {
+        $events = $this->getEmittedEvents($render);
+
+        foreach ($events as $event) {
+            if ($event['event'] === $eventName) {
+                return $event;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * @return array<array{data: array<string, int|float|string|bool|null>, event: non-empty-string}>
+     */
+    public function getEmittedEvents(RenderedComponent $render): array
+    {
+        $emit = $render->crawler()->filter('[data-live-name-value]')->attr('data-live-events-to-emit-value');
+
+        if (null === $emit) {
+            return [];
+        }
+
+        return json_decode($emit, associative: true, flags: \JSON_THROW_ON_ERROR);
+    }
+
+    public function getName(): string
+    {
+        return $this->metadata->getName();
+    }
 }
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,24 @@
 # CHANGELOG
 
+## 2.27.0
+
+-  Add events assertions in `InteractsWithLiveComponents`:
+```php
+$testComponent = $this->createLiveComponent(name: 'MyComponent');
+
+$renderedComponent = $testComponent->render();
+
+// Assert that the component did emit an event named 'event'
+$this->assertComponentEmitEvent($render, 'event')
+    // optionally, you can assert that the event was emitted with specific data...
+    ->withData(['arg1' => 'foo', 'arg2' => 'bar'])
+    // ... or only with a subset of data
+    ->withDataSubset(['arg1' => 'foo']);
+
+// Assert that the component did not emit an event named 'another-event'
+$this->assertComponentNotEmitEvent($render, 'another-event');
+```
+
 ## 2.26.0
 
 -   `LiveProp`: Pass the property name as second parameter of the `modifier` callable
--- a/src/LiveComponent/src/Test/Util/AssertEmittedEvent.php
+++ b/src/LiveComponent/src/Test/Util/AssertEmittedEvent.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Test\Util;
+
+use PHPUnit\Framework\TestCase;
+
+final class AssertEmittedEvent
+{
+    /**
+     * @param array<string, int|float|string|bool|null> $data
+     */
+    public function __construct(
+        private readonly TestCase $testCase,
+        private readonly string $eventName,
+        private readonly array $data,
+    ) {
+    }
+
+    /**
+     * @return self
+     */
+    public function withDataSubset(array $expectedEventData): object
+    {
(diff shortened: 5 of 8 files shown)


src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -229,4 +229,36 @@ private function flattenFormValues(array $values, string $prefix = ''): array
...
+    }
+
+    /**
+     * @return array<int, array{data: array<string, string>, event: string}>
+     */
+    public function getEmittedEvents(RenderedComponent $render): array
+    {
+        $div = $render->crawler()->filter('div');
+
+        $emit = $div->attr('data-live-events-to-emit-value');
```
Kocal: I may be wrong, but it's work if your component use a `<div>` tag no? What if it uses a `<time>` or `<span>` instead?

Arkalo2: Nice catch ! I rewrite it using  `data-live-name-value` attribute

kbond (approved):
Good idea!

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
```diff
@@ -44,4 +44,24 @@ protected function createLiveComponent(string $name, array $data = [], ?KernelBr
             self::getContainer()->get('router'),
         );
     }
+
+    protected function assertComponentEmitEvent(TestLiveComponent $testLiveComponent, string $eventName, ?array $parameters = null): void
+    {
+        $eventData = $testLiveComponent->getEmittedEvent($testLiveComponent->render(), $eventName);
+
+        $this->assertNotNull($eventData, \sprintf('Event %s not emitted', $eventName));
```
Kocal: Can we have something like that instead?

```suggestion
        $this->assertNotNull($event, \sprintf('The component "%s" did not emit event "%s".', '...', $eventName));
```

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
```diff
@@ -44,4 +44,24 @@ protected function createLiveComponent(string $name, array $data = [], ?KernelBr
...
+
+        if (null === $parameters) {
+            return;
+        }
+
+        foreach ($parameters as $key => $value) {
+            $this->assertSame($value, $eventData['data'][$key] ?? null, \sprintf('EventData (%s) is not valid', $key));
+        }
```
Kocal: It will helps by displaying a nice diff between expected vs actual data:
```suggestion
        $this->assertEquals(expectedEventData, $event['data'], sprintf('The expected event  "%s" data does not match.', $eventName));
```

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
```diff
@@ -44,4 +44,24 @@ protected function createLiveComponent(string $name, array $data = [], ?KernelBr
             self::getContainer()->get('router'),
         );
     }
+
+    protected function assertComponentEmitEvent(TestLiveComponent $testLiveComponent, string $eventName, ?array $parameters = null): void
```
Kocal:
```suggestion
    protected function assertComponentEmitEvent(TestLiveComponent $testLiveComponent, string $expectedEventName, ?array $expectedEventData = null): void
```

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
```diff
@@ -44,4 +44,24 @@ protected function createLiveComponent(string $name, array $data = [], ?KernelBr
...
+    {
+        $eventData = $testLiveComponent->getEmittedEvent($testLiveComponent->render(), $eventName);
```
Kocal: "event data" refers to the data passed when emitting an event, but here you are getting the event (in its arrayable form):
 
```suggestion
        $event = $testLiveComponent->getEmittedEvent($testLiveComponent->render(), $eventName);
```

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
```diff
@@ -44,4 +44,24 @@ protected function createLiveComponent(string $name, array $data = [], ?KernelBr
...
+
+        $this->assertNotNull($eventData, \sprintf('Event %s not emitted', $eventName));
+
+        if (null === $parameters) {
```
Kocal:
```suggestion
        if (null === $expectedEventData) {
```

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
```diff
@@ -44,4 +44,24 @@ protected function createLiveComponent(string $name, array $data = [], ?KernelBr
...
+        }
+
+        foreach ($parameters as $key => $value) {
+            $this->assertSame($value, $eventData['data'][$key] ?? null, \sprintf('EventData (%s) is not valid', $key));
+        }
+    }
+
+    protected function assertComponentNotEmitEvent(TestLiveComponent $testLiveComponent, string $eventName): void
+    {
+        $this->assertNull($testLiveComponent->getEmittedEvent($testLiveComponent->render(), $eventName), \sprintf('Event %s emitted', $eventName));
```
Kocal: I believe it should be more helpful (if you have a better suggestion, add it! :) )
```suggestion
        $this->assertNull($testLiveComponent->getEmittedEvent($testLiveComponent->render(), $eventName), \sprintf('The component "%s" emitted the "%s" event, but it should not have.', '...', $eventName));
```

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -229,4 +229,35 @@ private function flattenFormValues(array $values, string $prefix = ''): array
 
         return $result;
     }
+
+    /**
+     * @return ?array{data: array<string, string>, event: string}
+     */
+    public function getEmittedEvent(RenderedComponent $render, string $eventName): ?array
+    {
+        $eventsData = $this->getEmittedEvents($render);
```
Kocal:
```suggestion
        $events = $this->getEmittedEvents($render);
```

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -229,4 +229,35 @@ private function flattenFormValues(array $values, string $prefix = ''): array
 
         return $result;
     }
+
+    /**
+     * @return ?array{data: array<string, string>, event: string}
```
Kocal:
```suggestion
     * @return ?array{data: array<string, int|float|string|bool|null>, event: non-empty-string}
```

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -229,4 +229,35 @@ private function flattenFormValues(array $values, string $prefix = ''): array
...
+
+        return null;
+    }
+
+    /**
+     * @return array<int, array{data: array<string, string>, event: string}>
+     */
+    public function getEmittedEvents(RenderedComponent $render): array
+    {
+        $div = $render->crawler()->filter('[data-live-name-value]');
```
Kocal:
```suggestion
        $component = $render->crawler()->filter('[data-live-name-value]');
```

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -229,4 +229,35 @@ private function flattenFormValues(array $values, string $prefix = ''): array
...
+            if ($eventData['event'] === $eventName) {
+                return $eventData;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * @return array<int, array{data: array<string, string>, event: string}>
```
Kocal:
```suggestion
     * @return array<array{data: array<string, int|float|string|bool|null>, event: non-empty-string}>
```

src/LiveComponent/src/Test/TestLiveComponent.php
```diff
@@ -229,4 +229,35 @@ private function flattenFormValues(array $values, string $prefix = ''): array
...
+    public function getEmittedEvents(RenderedComponent $render): array
+    {
+        $div = $render->crawler()->filter('[data-live-name-value]');
+        $emit = $div->attr('data-live-events-to-emit-value');
+
+        if (null === $emit) {
+            return [];
+        }
+
+        return json_decode($emit, true);
```
Kocal:
```suggestion
        return json_decode($emit, associative: true, flags: \JSON_THROW_ON_ERROR);
```

src/LiveComponent/doc/index.rst
```diff
@@ -3829,6 +3834,44 @@ uses Symfony's test client to render and make requests to your components::
...
+            $this->assertComponentNotEmitEvent($render, 'unwanted-event');
+        }
+    }
+
+.. note::
+
+    These assertions are only available when using the ``InteractsWithLiveComponents`` trait in a test class that extends
+    ``Symfony\Bundle\FrameworkBundle\Test\KernelTestCase``.
+
+---
```
Kocal: I don't think this whole block is useful as its own, IMHO it could be merged with the previous code example. 

By doing that, we won't repeat twice "When testing live components, you can assert (...) provided by the ``InteractsWithLiveComponents`` trait" nor "These assertions are only available when using (...)".

Kocal (changes requested):
Minor wordings and types, but otherwise that's fine, thanks :)

src/LiveComponent/src/Test/InteractsWithLiveComponents.php
Arkalo2: Yep but with the global assertEquals, we must pass all expectedEventDatas.

If the event contains ['a' => 'b', 'c' => 'd'];
And I expect only ['a' => 'b']; 
It throw an error.

Kocal: That's true, but I think that's the desired behavior. Asserting the whole event data makes more sense to me, but I may be biaised and I understand that sometimes you expect to assert an array subset, but it must be explicit. 

I've seen it tons of time on multiple projects, where sometimes you assert everything, or sometimes only a subset, and every time that's a mess because the method-name / PHPUnit's way of working does not help.

I can suggest you this, so everyone would be happy:
```php
// classic way
$this->assertComponentEmitEvent($component, 'foo'); // do not care about event data
$this->assertComponentEmitEventWithData($component, 'foo', []); // assert whole array
$this->assertComponentEmitEventWithDataSubset($component, 'foo', []); // assert subset array

// or a bit exotic 
$this->assertComponentEmitEvent($component, 'foo'); // do not care about event data
$this->assertComponentEmitEvent($component, 'foo')->withData([]); // assert whole array
$this->assertComponentEmitEvent($component, 'foo')->withDataSubset([]); // assert subset array
```

WDYT?

kbond: > // or a bit exotic 
> $this->assertComponentEmitEvent($component, 'foo'); // do not care about event data
> $this->assertComponentEmitEvent($component, 'foo')->withData([]); // assert whole array
> $this->assertComponentEmitEvent($component, 'foo')->withDataSubset([]); // assert subset array

I personally **_LOVE_** this! 😍

Arkalo2: I tested the `->with` implementation

Arkalo2: An anonym class is enough or I have to separate it ?

The limit of the anonyme class is the dot block : I can not chained the `->with`

Kocal: A separate concrete class please :)

Kocal (approved):

src/LiveComponent/doc/index.rst
```diff
@@ -3781,6 +3781,12 @@ uses Symfony's test client to render and make requests to your components::
                 ->emit('increaseEvent', ['amount' => 2]) // emit a live event with arguments
             ;
 
+            // Assert that the event was emitted
+            $this->componentHasEmittedEvent($testComponent->render(), 'increaseEvent')
```
```suggestion
            $this->assertComponentEmitEvent($testComponent->render(), 'increaseEvent')
```

Kocal:
Rebased, squashed, some rewords, changelog updated and some documentation fixed 💅🏻

Kocal:
Thanks @Arkalo2.

PR #2716, opened 2025-05-06
[Site][Toolkit] Add manual installation steps for Component installation (and some website tweaks) 

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

Another PR started in the plane :D 

IMG_2736 - copie

--- 

This one add manual installation steps in the documentation, letting people see files they will install before (or without) installing the UX Toolkit:

https://github.com/user-attachments/assets/93ec2f2c-1a2b-4637-a9fa-f92fa4273f3d

### DIFF
--- a/ux.symfony.com/assets/styles/components/_Terminal.scss
+++ b/ux.symfony.com/assets/styles/components/_Terminal.scss
@@ -6,6 +6,7 @@
     border-radius: .75rem;
     position: relative;
     font-size: 12px;
+    display: grid; // Ensure the Terminal overflow its parent if "pre" contains a very-long-line of the same highlighted element (e.g.: a long string)
 }
 
 .Terminal_light {
@@ -162,6 +163,7 @@
         overflow: visible;
     }
     scrollbar-width: thin;
+    scrollbar-color: rgba(255, 255, 255, .8) transparent;
     pre {
         background: none;
     }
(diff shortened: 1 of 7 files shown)


ux.symfony.com/assets/styles/components/_Terminal.scss
```diff
@@ -170,15 +172,15 @@
...
-}
+// @media screen and (min-width: 768px) {
+//     .Terminal_content::-webkit-scrollbar {
+//         display: none;
+//     }
+//     .Terminal_content {
+//         --webkit-scrollbar-width: none;
+//         scrollbar-width: none;
+//     }
+// }
```
Kocal, line 183: I'm not sure about this one, but I don't really see why you would want to hide a scrollbar, as it reduce accessibility and UX

smnandre: Is this related to Toolkit ? Any page i can  see the problem ?

smnandre (approved):

ux.symfony.com/assets/styles/components/_Terminal.scss
Kocal: Yes and no, I wanted to display a scroll bar when necessary, but I couldn't understand why this code exists.
You can check the "Alert" component page, by unfolding the files from "manual installation" section, it contains a Terminal with a veryyy long and unbreakable line.

You can also check other parts of the website where a Terminal is displayed, to confirm it didn't break things. 🙏

smnandre: Because it broke things at a time, two scrollbars were displayed. Testing on mac often gives a false impression, as macOS hide scrollbars per default.

But let's see and we will be able to fix if there were any pb 

👍 :

kbond (approved):
Nice!

PR #2717, opened 2025-05-06
[ToolKit] Add missing Interface suffixes

smnandre:
Fix #2539


nicolas-grekas (approved):
Thanks!

Kocal:
I can't verify since I'm on my phone, but do we use any of these interfaces in the Website too?

Kocal (approved):

Kocal:
Thank you @smnandre.

PR #2720, opened 2025-05-08
[Toolkit] Add functional tests to render all Kit components usage codes (from their documentation), with a snapshot system

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

This PR add snapshots testing for each Twig code blocks from `src/Toolkit/kits/**/docs/components/*.md`, to ensure that rendering is always good. 

I don't think we will add visual testing, at least for now. Having snapshots tests and visual previews (on the website) feels convenient enough, but if someone has a super-simple solution, why not!)

I tagged the PR as a feature, not because the snapshots testing itself, but because I introduced a new service `KitContextRunner`. I needed something similar than the "hacks" from the website, but for tests.


.github/workflows/test.yaml
```diff
@@ -116,6 +116,13 @@ jobs:
                 working-directory: "src/${{ matrix.component }}"
                 dependency-versions: ${{ matrix.dependency-version }}
 
+            - name: Install specific packages for Toolkit when PHP >=8.2
+              if: ${{ matrix.component == 'Toolkit' && matrix.php-version != '8.1' }}
+              run:
+                composer require --dev tales-from-a-dev/twig-tailwind-extra
+              with:
+                working-directory: "src/${{ matrix.component }}"
+
```
Kocal: We want to keep PHP 8.1 compatibility for Toolkit, but tales-from-a-dev/twig-tailwind-extra requires PHP 8.2

src/Toolkit/tests/Functional/__snapshots__/ComponentsRenderingTest__testComponentRendering with data set Kit shadcn, component Alert, code 1__1.html
```diff
@@ -0,0 +1,20 @@
...
+<twig:Alert class="max-w-lg">
+    <twig:ux:icon name="tabler:terminal" class="h-4 w-4" />
+    <twig:Alert:Title>Heads up!</twig:Alert:Title>
+    <twig:Alert:Description>
+        You can add components to your app using the cli.
+    </twig:Alert:Description>
+</twig:Alert>
+```
+- Rendered code (prettified for testing purposes, run "php vendor/bin/phpunit -d --update-snapshots" to update snapshots): -->
+<div class="relative w-full rounded-lg border p-4 [&amp;&gt;svg~*]:pl-7 [&amp;&gt;svg+div]:translate-y-[-3px] [&amp;&gt;svg]:absolute [&amp;&gt;svg]:left-4 [&amp;&gt;svg]:top-4 [&amp;&gt;svg]:text-foreground bg-background text-foreground max-w-lg" role="alert">
```
Kocal, line 15: I will see if I should and how I can replace `&amp;` to `&`, etc... later today

src/Toolkit/tests/Fixtures/Kernel.php
```diff
@@ -16,21 +16,34 @@
...
+            new TwigExtraBundle(),
+            new UXIconsBundle(),
             new UXToolkitBundle(),
         ];
+
+        if (class_exists(TalesFromADevTwigExtraTailwindBundle::class)) {
+            $bundles[] = new TalesFromADevTwigExtraTailwindBundle();
+        } elseif (\PHP_VERSION_ID >= 80200) {
+            throw new \RuntimeException('The dependency "tales-from-a-dev/twig-tailwind-extra" must be installed when using PHP 8.2+.');
+        }
```
Kocal: I want to force `tales-from-a-dev/twig-tailwind-extra` installation if we are using PHP 8.2+, in order to not forgot it when running tests

src/Toolkit/tests/Functional/ComponentsRenderingTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+        $kit = $this->instantiateKit($kitName);
+        $template = $twig->createTemplate($code);
+
+        try {
+            $renderedCode = $kitContextRunner->runForKit($kit, fn () => $template->render());
+        } catch (\Twig\Error\SyntaxError $e) {
+            if (\PHP_VERSION < 80200 && str_contains($e->getMessage(), 'Unknown "tailwind_merge" filter')) {
+                $this->markTestSkipped('Filter "tailwind_merge" is not supported on PHP <8.2.');
+            }
```
Kocal: And here, skip the test if the issue is caused by missing `tales-from-a-dev/twig-tailwind-extra` dependency, since we don't want to create "weird" snapshots.

.github/workflows/test.yaml
ker0x: If it helps, I can lower the PHP version to 8.1 for `tales-from-a-dev/twig-tailwind-extra`

Kocal: Oh, that's would be super appreciated 🙏🏻

ker0x: It's done, see https://github.com/tales-from-a-dev/twig-tailwind-extra/releases/tag/v0.4.0 👍

Kocal: Thanks!

src/Toolkit/tests/Functional/__snapshots__/ComponentsRenderingTest__testComponentRendering with data set Kit shadcn, component Alert, code 1__1.html
Kocal: Nothing to do here, `[&amp;&gt;svg~*]:pl-7` correctly matches to `.\[\&\>svg\~\*\]\:pl-7`

PR #2743, opened 2025-05-16
Use singular folder pattern like other PHP components codes

94noni:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no
| Issues        | Fix coherence pattern naming
| License       | MIT

opening this PR mostly as RFC
other code in sf document to create special code (messenger message, event listener|subscriber, etc) inside folders in singular form not plural
does It makes sens for twig component to follow this pattern ?

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -7,8 +7,8 @@ making it easier to render and re-use small template "units" - like an
 
 Every component consists of (1) a class::
 
-    // src/Twig/Components/Alert.php
-    namespace App\Twig\Components;
+    // src/Twig/Component/Alert.php
+    namespace App\Twig\Component;
 
     use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 
@@ -79,8 +79,8 @@ Let's create a reusable "alert" element that we can use to show success
 or error messages across our site. Step 1 is to create a component class
 and give it the ``AsTwigComponent`` attribute::
 
-    // src/Twig/Components/Alert.php
-    namespace App\Twig\Components;
+    // src/Twig/Component/Alert.php
+    namespace App\Twig\Component;
 
     use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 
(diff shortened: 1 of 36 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -23,7 +23,7 @@ And (2) a template:
 
 .. code-block:: html+twig
 
-    {# templates/components/Alert.html.twig #}
+    {# templates/component/Alert.html.twig #}
```
smnandre: templates has a "s" marking plural, similarely here to me it should stay plural

src/TwigComponent/doc/index.rst
```diff
@@ -7,7 +7,7 @@ making it easier to render and re-use small template "units" - like an
 
 Every component consists of (1) a class::
 
-    // src/Twig/Components/Alert.php
+    // src/Twig/Component/Alert.php
```
smnandre, line 10:
```suggestion
    // src/Twig/Component/Alert.php
```

Something we will discuss in 3.0 .. but i totally agree.

To be fair, i even think we should use `src/Component`

smnandre (commented):
👍  for the namespace (but not now)

👎  for the template folder

94noni:
top for php file and ok for twig file makes sens
will rewrite this PR on the php codebase side 👍🏻

src/TwigComponent/doc/index.rst
94noni: i do have "devland codebase" component (relatively to symfony component)
having them at root level with "twig component" seems wrong to me

Kocal:
If we want to be iso everywhere, it also means that we must:
- update the MakerBundle https://github.com/symfony/maker-bundle/blob/1.x/src/Maker/MakeTwigComponent.php#L33
- create a new recipe, because of https://github.com/symfony/recipes/blob/main/symfony/ux-twig-component/2.13/config/packages/twig_component.yaml#L5

94noni:
@Kocal thx ! I can work on it but please before assert to me that its desired so i wont lost my time and yours as well :)

smnandre:
I don't think it is the time

94noni:
copy that, thx for being transparent !
will revisit for 3.x 🤞🏻

Issue #1025, opened 2023-07-27
[Dropzone] Support for multiple files upload


andersmateusz:
As of my knowledge dropzone does not accept multiple files upload and despite nice try of implementing this feature #7 issue was closed. I think multiple file upload is essential for dropzone. I am right that this feature is still not implemented?

weaverryan:
Correct! We just need someone to help carry this forward and think about how it should look and work. I'm going to close - as the issue #7 is still open (the corresponding PR #512 was closed), but yes, still needed 👍

mateuszanders:
@weaverryan  Sure, I will try to implement this

andersmateusz:
@weaverryan Sure I will try to implement this

PR #512, opened 2022-10-24
[Dropzone] Enable multiple file uploads

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Tickets       | Fix #7  
| License       | MIT

This change adds the posibility to upload multiple files using the Dropzone component. This is based on the work by @yassinehamouten.

- [x] Tests
- [x] Make it configurable using the FormType

Issue #7, opened 2020-12-04
Dropzone: handle multiple files


jmsche:
It would be nice if Dropzone could handle multiple files.

I was expecting it would as its parent type is the FileType, that allows setting `multiple` option to `true`.

tgalopin:
Yep, that's a good idea! I'm not sure how to handle it properly yet but that's a good idea :)

AxelBriche:
Probably like https://www.dropzonejs.com, with multiple query in ajax for upload each image and return a valid or error response.

Lenawel:
is Multiple File allowed or not yet ?

connorhu:
I plan to make multiple and chunked file uploads in one of my pilot project. At this moment I don't find any "official" direction about templating of UX modules (at this moment small prototyes). This is problem at my symfony collection type ux module too.

yassinehamouten:
Hello,
I am having the same issue. I did a few things to make it work.
I will try to make it available to everyone as a package

expace-dev:
Hello,

someone have the solution to do multiple upload?

yassinehamouten:
Hello,

You can see how I make it work there https://github.com/yassinehamouten/ux-dropzone-multiple/tree/main/Resources. I tried to do it as a package, but stimulus does not call my controller. If someone has an idea...

I changed:
- in the form template I added "multiple" in the file input https://github.com/yassinehamouten/ux-dropzone-multiple/blob/3d282f520811ae720b0b194b7d28363eba6a1502/Resources/views/form_theme.html.twig#L6
- and in the controller.js https://github.com/yassinehamouten/ux-dropzone-multiple/blob/main/Resources/assets/src/controller.js

yassinehamouten:
It's look like this if you take the css too 
![image](https://user-images.githubusercontent.com/22196574/141649407-ef887b61-0000-4643-b9ea-136cbe628164.png)

JML75:
thank fou for your work, does it work if you use the DropzoneType:: class in your formType ?

tgalopin:
@yassinehamouten this is interesting, I wasn't sure we could do it with the multiple attribute, good news! That would be a great PR if you wish :) !

daFish:
@yassinehamouten Would it be ok for you when I create a PR with your changes?

JML75:
hello

not sure what your are talking about.

best 

JML

Le 24 oct. 2022 à 11:13, Marcus Stöhr ***@***.***> a écrit :


[...]

PR #2747, opened 2025-05-18
[Dropzone] Add multiple file preview

### LINKED ISSUE #2552: [Dropzone] Upload same file empty the drop zone, chrome only
When you upload the same file, Chrome do not trigger change event. Firefox do. I think Chrome have the correct behavior this time.

This behavior introduced an issue in the UX dropzone. In chrome, when you upload the same file, the dropzone is emptied.

https://github.com/user-attachments/assets/4e8753d8-6bb4-4d66-b0f2-e9dce03d3f02

---

Solution :  Listen the drop event and show the drop zone if a input value is set.

What do you think of this solution ? Tested ok in Chrome and Firefox.

in https://github.com/symfony/ux/blob/v2.22.1/src/Dropzone/assets/dist/controller.js

```js
    initialize() {
        // ...
        this.onDrop = this.onDrop.bind(this);
    }

    connect() {
        // ...
        this.element.addEventListener('drop', this.onDrop);
    }

    disconnect() {        // ...
        this.element.removeEventListener('drop', this.onDrop);
    }

    onDrop() {
        if (this.inputTarget.value === '') {
            return;
        }

        this.inputTarget.style.display = 'none';
        this.placeholderTarget.style.display = 'none';
        this.previewTarget.style.display = 'flex';
    }
```

If it's ok for you, I can try to do the PR ;)

### LINKED ISSUE #2567: [Dropzone] Multiple uploads only show single file and no possibility to add more files

ghost:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #2567 Fix #2552
| License       | MIT

Updated Dropzone component to support multiple file preview.
No breaking changes. Backwards compatible by default. User can opt-in to new features.

**Example:**

```php
$form = $this->createFormBuilder()
    ->add('files', DropzoneType::class, [
        'multiple' => true,
        'preview' => [
            'style' => 'block',
            'can_toggle_placeholder' => true,
            'can_open_file_picker' => true,
        ]
    ])
    ->getForm();
```

[Demo repo](https://github.com/phasdev/symfony-ux-dropzone-demo)

---

**Basic support for `legacy` style:**
(File list shown in native browser tooltip by setting filename element's `title` attribute)

**New `inline` style:**

**New `block` style:**

**New `block` style with `preview.can_toggle_placeholder = false`:**

---

- Added `preview.style` option. The (default) `legacy` option reproduces existing behaviour. The `block` and `inline` options enable new features.

- Added `preview.can_toggle_placeholder` option to control whether placeholder remains visible when preview is rendered. Can be `true`, `false` or `auto`. The `auto` option will hide placeholder for single-file uploads; adding additional files will show placeholder again. 

- Added `preview.can_open_file_picker` option to control whether clicking preview container opens input file picker.

- New component handles drop event directly, making cross-browser behaviour consistent [(#2552)](https://github.com/symfony/ux/issues/2552); i.e. it does not rely on file input for drop handling.

- Each selected file has its own preview. Thumbnails are shown for image files and a placeholder for other files. Files can be added/removed individually.

- File preview template, button image and placeholder image markup can be overridden by adding content to corresponding form theme blocks, allowing markup customization without having to override entire form theme.

- File input hidden to improve component style flexibility[[1]](https://stackoverflow.com/questions/572768/styling-an-input-type-file-button#:~:text=Styling%20file%20inputs%20are%20notoriously%20difficult). Input still visible to screen readers.

- File input wrapped within label so label can obtain focus, allowing keyboard/screen reader navigation.

- Component is outlined on drag hover, providing feedback that mouse is detected within dropzone.

- Legacy code clearly marked so it can be removed from component during next major release (if desired).

---
**Related issues:**

https://github.com/symfony/ux/issues/2567
https://github.com/symfony/ux/issues/2552
https://github.com/symfony/ux/issues/1025
https://github.com/symfony/ux/issues/7

**Related PR's:**

https://github.com/symfony/ux/pull/2704
https://github.com/symfony/ux/pull/2642
https://github.com/symfony/ux/pull/512


src/Dropzone/assets/dist/controller.d.ts
```diff
@@ -1,19 +1,48 @@
 import { Controller } from '@hotwired/stimulus';
 export default class extends Controller {
     readonly inputTarget: HTMLInputElement;
-    readonly placeholderTarget: HTMLDivElement;
-    readonly previewTarget: HTMLDivElement;
-    readonly previewClearButtonTarget: HTMLButtonElement;
-    readonly previewFilenameTarget: HTMLDivElement;
-    readonly previewImageTarget: HTMLDivElement;
+    readonly placeholderTarget: HTMLElement;
+    readonly previewTargets: HTMLElement[];
```
ghost, line 5: Possible breaking change? Some controller targets have been removed and I've changed types on `placeholderTarget` and `previewTarget` so they're more generic. For example, the new placeholder is a `<label>` not a `<div>`, so I changed `placeholderTarget` type from `HTMLDivElement` to the more generic `HTMLElement`.

Since the [documentation](https://symfony.com/bundles/ux-dropzone/current/index.html#extend-the-default-behavior) suggests using event listeners to extend behaviour I assumed the controller class wouldn't be extended directly, so perhaps these aren't breaking changes.

ghost:
I can add tests/docs when we're nearly ready to merge. Didn't want to add tests for features that may be dropped during review.

Kocal:
User is a ghost, closing

PR #2763, opened 2025-05-22
[Map] add definition to after create events in controller

mart-insiders:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |
| Docs?         | no
| Issues        | Fix 
| License       | MIT

Hi,
The [documentation](https://symfony.com/bundles/ux-map/current/index.html#advanced-passing-extra-data-from-php-to-the-stimulus-controller )states that the definition should be accessible in the <type>:after-create events in the map controllers.

### DIFF
--- a/src/Map/assets/dist/abstract_map_controller.js
+++ b/src/Map/assets/dist/abstract_map_controller.js
@@ -43,7 +43,7 @@ class default_1 extends Controller {
     createInfoWindow({ definition, element, }) {
         this.dispatchEvent('info-window:before-create', { definition, element });
         const infoWindow = this.doCreateInfoWindow({ definition, element });
-        this.dispatchEvent('info-window:after-create', { infoWindow, element });
+        this.dispatchEvent('info-window:after-create', { infoWindow, definition, element });
         this.infoWindows.push(infoWindow);
         return infoWindow;
     }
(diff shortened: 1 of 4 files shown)


Kocal (approved):
Indeed, thanks for the fix :)

src/Map/assets/dist/abstract_map_controller.js
```diff
@@ -74,7 +74,7 @@ class default_1 extends Controller {
         return ({ definition }) => {
             this.dispatchEvent(eventBefore, { definition });
             const drawing = factory({ definition });
-            this.dispatchEvent(eventAfter, { [type]: drawing });
+            this.dispatchEvent(eventAfter, { [type]: drawing, definition: definition });
```
Kocal: Since I can't push on your repository, you will need to do the following changes:
- This change must be done in the `src/abstract_map_controller.ts` file
- You also need to update the `info-window:after-create` event
- finally, you have to rebuild the dist files (https://github.com/symfony/ux/blob/2.x/CONTRIBUTING.md#working-with-assets)
Thanks!

Kocal (changes requested):

src/Map/assets/dist/abstract_map_controller.js
dannyvw: @Kocal I pushed some changes.

Kocal (approved):

src/Map/assets/dist/abstract_map_controller.js
Kocal: That's perfect, thanks :)

Kocal:
Thanks @mart-insiders and @dannyvw :)

PR #2767, opened 2025-05-23
[LiveComponent] Fix missing _route on live components batch requests

### LINKED ISSUE #1022: BatchActionController isn't passing default route attributes
Let's say we register the `ux_live_component` routes as following:
```
live_component:
    resource: '@LiveComponentBundle/config/routes.php'
    prefix: '/_components'
    defaults:
       _routeScope: ['storefront']
       _loginRequired: true 
```
When the `\Symfony\UX\LiveComponent\Controller\BatchActionController::__invoke` is invoking sub requests, it sets the only certain attributes. So that for example the `_loginRequired` and `_routeScope` defaults are skipped. I want to create a PR to address this issue but not sure where to look. As the defaults are added to the routes at compile time and the batchAction is invoking the requests directly so that the route matching (and default attributes logic) is skipped.

arrabiata-asanz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #1022
| License       | MIT

The `_route` param get overwritten for Live Component Batch Requests, persist this param to the sub requests

`Symfony\UX\LiveComponent\Controller\BatchActionController`
```
$request->duplicate(attributes: [
                '_controller' => [$serviceId, $name],
                '_component_action_args' => $action['args'] ?? [],
                '_mounted_component' => $_mounted_component,
                '_live_component' => $serviceId,
            ]

$response = $this->kernel->handle($subRequest, HttpKernelInterface::SUB_REQUEST, false);

```
Duplicate overwrites all attributes in the sub request, missing one important attribute _route, add it to the array in order to know where the request comes from.

### DIFF
--- a/src/LiveComponent/src/Controller/BatchActionController.php
+++ b/src/LiveComponent/src/Controller/BatchActionController.php
@@ -38,6 +38,7 @@ final class BatchActionController
                 '_component_action_args' => $action['args'] ?? [],
                 '_mounted_component' => $_mounted_component,
                 '_live_component' => $serviceId,
+                '_route' => $request->attributes->get('_route'),
             ]);
 
             $response = $this->kernel->handle($subRequest, HttpKernelInterface::SUB_REQUEST, false);


arrabiata-asanz:
i think the failed tests don't apply to the changes in this PR. (Deprecation warning and Exception Messages Formatting)

Kocal:
Thanks for the PR! I'm totally not familiar with LiveComponent, but:
1. can we add a test preventing any regression?
2. I think we should also pass `_route_params`, no?

Kocal:
> i think the failed tests don't apply to the changes in this PR. (Deprecation warning and Exception Messages Formatting)

Yes you can ignore the 8.4 dev job, but for fabbot, it would be nice to fix it (even if not related :) )

smnandre:
Hmm.. could we also add some tests for this?

Kocal:

Thanks!

Status: Needs Work

PR #2768, opened 2025-05-23
Homogenize PHP-CS-Fixer configuration with symfony/symfony, and run it

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -13,14 +13,19 @@ if (!file_exists(__DIR__.'/src')) {
     exit(0);
 }
 
-$fileHeaderComment = <<<'EOF'
-This file is part of the Symfony package.
+$fileHeaderParts = [
+    <<<'EOF'
+        This file is part of the Symfony package.
 
-(c) Fabien Potencier <fabien@symfony.com>
+        (c) Fabien Potencier <fabien@symfony.com>
 
-For the full copyright and license information, please view the LICENSE
-file that was distributed with this source code.
-EOF;
+        EOF,
+    <<<'EOF'
+
+        For the full copyright and license information, please view the LICENSE
+        file that was distributed with this source code.
+        EOF,
+];
 
 return (new PhpCsFixer\Config())
     ->setParallelConfig(PhpCsFixer\Runner\Parallel\ParallelConfigFactory::detect())
@@ -28,8 +33,17 @@ return (new PhpCsFixer\Config())
         '@PHPUnit75Migration:risky' => true,
         '@Symfony' => true,
         '@Symfony:risky' => true,
-        'header_comment' => ['header' => $fileHeaderComment],
-        'trailing_comma_in_multiline' => ['elements' => ['arrays', 'match', 'parameters']],
+        'protected_to_private' => false,
+        'header_comment' => [
+            'header' => implode('', $fileHeaderParts),
+            'validator' => implode('', [
+                '/',
+                preg_quote($fileHeaderParts[0], '/'),
+                '(?P<EXTRA>.*)??',
+                preg_quote($fileHeaderParts[1], '/'),
+                '/s',
+            ]),
+        ],
     ])
     ->setRiskyAllowed(true)
     ->setFinder(
@@ -38,5 +52,7 @@ return (new PhpCsFixer\Config())
             ->append([__FILE__])
             ->notPath('#/Fixtures/#')
             ->notPath('#/var/#')
+            // does not work well with `fully_qualified_strict_types` rule
+            ->notPath('LiveComponent/tests/Integration/LiveComponentHydratorTest.php')
     )
 ;
--- a/src/Icons/config/twig_component.php
+++ b/src/Icons/config/twig_component.php
@@ -12,7 +12,6 @@
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
 use Symfony\UX\Icons\Twig\UXIconComponent;
-use Symfony\UX\Icons\Twig\UXIconComponentListener;
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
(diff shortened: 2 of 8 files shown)


Kocal:
Interesting, because PHP-CS-Fixer changed:
```
namespace Symfony\UX\LiveComponent\Tests\Integration;

use Symfony\UX\LiveComponent\Tests\Fixtures\Dto\Address;

// ...

/**
 * @var \Symfony\UX\LiveComponent\Tests\Fixtures\Dto\Address[]
 */
#[LiveProp(writable: true, useSerializerForHydration: true)]
```

for:
```
namespace Symfony\UX\LiveComponent\Tests\Integration;

use Symfony\UX\LiveComponent\Tests\Fixtures\Dto\Address;

// ...

/**
 * @var Address[]
 */
#[LiveProp(writable: true, useSerializerForHydration: true)]
```

The following error now happens:
```
1) Symfony\UX\LiveComponent\Tests\Integration\LiveComponentHydratorTest::testCanDehydrateAndHydrateComponentWithTestCases with data set "Array with DTOs: fully writable allows anything to change" (Closure Object (...))
Symfony\Component\Serializer\Exception\NotNormalizableValueException: Could not denormalize object of type "\Address[]", no supporting normalizer found.

/home/runner/work/ux/ux/src/LiveComponent/vendor/symfony/serializer/Serializer.php:222
/home/runner/work/ux/ux/src/LiveComponent/src/LiveComponentHydrator.php:292
/home/runner/work/ux/ux/src/LiveComponent/src/LiveComponentHydrator.php:169
/home/runner/work/ux/ux/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php:128
/home/runner/work/ux/ux/src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php:157
```

That's weird, I would expect `Address` from the PHPDoc to be resolved to `\Symfony\UX\LiveComponent\Tests\Fixtures\Dto\Address`. Maybe that's because of the anonymous class? 

**EDIT:** yes that's because of anonymous classes:
image

.php-cs-fixer.dist.php
```diff
@@ -28,8 +30,20 @@
...
+        'no_superfluous_phpdoc_tags' => [
+            'remove_inheritdoc' => true,
+            'allow_unused_params' => true, // for future-ready params, to be replaced with https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/7377
+        ],
         'header_comment' => ['header' => $fileHeaderComment],
-        'trailing_comma_in_multiline' => ['elements' => ['arrays', 'match', 'parameters']],
+        // TODO: Remove once the "compiler_optimized" set includes "sprintf"
+        'native_function_invocation' => ['include' => ['@compiler_optimized', 'sprintf'], 'scope' => 'namespaced', 'strict' => true],
+        'nullable_type_declaration' => true,
+        'nullable_type_declaration_for_default_null_value' => true,
```
smnandre: Not needed anymore... where do you take your file ? 😅 

https://github.com/symfony/symfony/blob/7.3/.php-cs-fixer.dist.php

.php-cs-fixer.dist.php
```diff
@@ -9,6 +9,8 @@
  * file that was distributed with this source code.
  */
 
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+
```
smnandre:
```suggestion
```

We should avoid this as much as we can, as the class is not needed here, and it can only bring weird things 😅

.php-cs-fixer.dist.php
```diff
@@ -28,8 +30,20 @@
         '@PHPUnit75Migration:risky' => true,
         '@Symfony' => true,
         '@Symfony:risky' => true,
+        'protected_to_private' => false,
+        'no_superfluous_phpdoc_tags' => [
+            'remove_inheritdoc' => true,
+            'allow_unused_params' => true, // for future-ready params, to be replaced with https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/7377
```
smnandre: cf https://github.com/symfony/symfony/pull/54073/files

smnandre (commented):
Yep, always some troubles with anonymous classes in test & php-cs-fixer .. 

For the rest, thanks!

.php-cs-fixer.dist.php
Kocal: Mmmmh...... mmmmmmmmmh.... 🤦🏻 
 definitely not from the last working branch, thanks, I will do the changes ASAP  😅

Kocal:
Thanks for the review @smnandre 🙏🏻 it should be better now!

.php-cs-fixer.dist.php
```diff
@@ -13,23 +13,40 @@
...
+
+        For the full copyright and license information, please view the LICENSE
+        file that was distributed with this source code.
+        EOF,
+];
 
 return (new PhpCsFixer\Config())
     ->setParallelConfig(PhpCsFixer\Runner\Parallel\ParallelConfigFactory::detect())
     ->setRules([
+        '@PHP71Migration' => true,
```
smnandre: Not sure we need this addition ?

.php-cs-fixer.dist.php
```diff
@@ -13,23 +13,40 @@
...
+            'validator' => implode('', [
+                '/',
+                preg_quote($fileHeaderParts[0], '/'),
+                '(?P<EXTRA>.*)??',
+                preg_quote($fileHeaderParts[1], '/'),
+                '/s',
+            ]),
+        ],
+        // It breaks `Symfony\UX\LiveComponent\Tests\Integration\LiveComponentHydratorTest` when modifying the PHPDoc of `#[LiveProp]` from anonymous class properties
+        'fully_qualified_strict_types' => false,
```
smnandre: Let's exclude only this test from csfixer then, wdyt ?

smnandre (approved):

.php-cs-fixer.dist.php
Kocal: That's something do-able yep

Kocal: Done

PR #2780, opened 2025-05-28
[CI] Replace ramsey/composer-install by `composer install` in app-tests workflow

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

In #2607, https://github.com/symfony/ux/actions/runs/15297531864/job/43030093028?pr=2607 is shown as failed because step `Run Encore (dev)`, but in reality it fails before but that's not shown:
image

With this change (which also remove a dependency), I hope the error to be well shown

### DIFF
--- a/.github/workflows/app-tests.yaml
+++ b/.github/workflows/app-tests.yaml
@@ -45,19 +45,15 @@ jobs:
         - uses: shivammathur/setup-php@v2
 
         - name: Install root dependencies
-          uses: ramsey/composer-install@v3
-          with:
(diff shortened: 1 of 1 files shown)


norkunas:
you now lost [caching dependencies](https://github.com/shivammathur/setup-php?tab=readme-ov-file#cache-composer-dependencies) which was abstracted away :)

PR #2783, opened 2025-05-28
[Icons] Docs: Deleting "search" codeblock

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Page: https://symfony.com/bundles/ux-icons/current/index.html#search-icon-sets

Reasons:
* I don't think that somebody will search for icons in a terminal window (i.e. without preview).
* If somebody wants after all, the command is shown at the end of the previous code block.

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -140,36 +140,6 @@ the prefix of a specific icon set:
 
      php bin/console ux:icons:search tabler arrow
 
-Search Icons
-~~~~~~~~~~~~
-
-You can also search for icons within a specific icon set. To search for "arrow"
-icons in the "Tabler Icons" set, use the following command:
-
-.. code-block:: terminal
-
-    $ php bin/console ux:icons:search tabler arrow
-
-    Searching Tabler Icons icons "arrow"...
-    Found 64 icons.
-     ------------------------------------------ ------------------------------------------
-      tabler:archery-arrow                       tabler:arrow-autofit-up
-      tabler:arrow-back                          tabler:arrow-back-up
-      tabler:arrow-badge-down                    tabler:arrow-badge-up
-      tabler:arrow-badge-up-filled               tabler:arrow-bar-both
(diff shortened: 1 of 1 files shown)


smnandre:
You'd be surprised..  :) 

It is a feature, discussed here / on Slack more than once, and I do not see how removing this paragraph would improve the page in any way to be honest.

So it's 👎 for me here.

ThomasLandauer:
Ping @javiereguiluz WDYT?

javiereguiluz:
I don't use this frequently myself ... but I've used it with icon sets that I know well (e.g. `Tabler`). I don't have to look at the icons before downloading them. I just need to find the exact right name for it.

But for icon sets I don't know, I'd prefer to check the icons first. So, I'd say that it's worth mentioning it even if probably it's not a super popular feature.

ThomasLandauer:
My point is that the command itself (`php bin/console ux:icons:search tabler arrow`) is shown in the previous code block anyway. So IMO there's no need to show the **full output** of that command too.

smnandre:
> php bin/console ux:icons:search tabler arrow

This is not the same results / usage.

---

I'm closing here as help is very much welcomed on the UX documentation.. 
... but deleting parts of it  or "un-documenting" features  is not something we have any desire right now. 

Hope you understand.

ThomasLandauer:
> This is not the same results / usage.

What do you mean is not the same?:

Screenshot_20250530_004350

smnandre:
The previous paragraph is called "Search Icon Sets". 

This one is called "Search Icons". People access it while reading (sometimes in diagonal) or .. by clicking on the table of content. So no, the fact the previous command result ends with "php bin/..." is not enough to remove any interest about this one.

It's very clear to me "search icons" is a very decisive feature, for a "icons" package..  so keeping a paragraph on how to do it and what to expect from it is not something we would want to do.

And I still don't see how _removing_ content from the documentation will help _any_ user IRL.

PR #2787, opened 2025-05-28
[Icons] Docs: Adding info on enduser performance

### LINKED ISSUE #2782: [Icons] Explain client side performance
Currently, the docs at https://symfony.com/bundles/ux-icons/current/index.html#performance only explain *server-side* aspects of performance (cache warmup, etc.)

**But what about the end user's performance?**

If include the same icon 200 times on a page, I will get the entire `<svg>...</svg>` string included 200 times in the page's source code, right? Those 200 strings cannot be cached.

So in some cases it probably makes more sense to just create an `.svg` file and include it with ``, cause this will be cached by the browser.

What's your opinion here? Would you say that (since icon files are usually small) it just doesn't make a difference for most users?

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | Closes #2782
| License       | MIT

Page: https://symfony.com/bundles/ux-icons/current/index.html#performance

Info is taken from https://github.com/symfony/ux/issues/2782#issuecomment-2917339106

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -456,7 +456,12 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
 Performance
 -----------
 
-The UX Icons component is designed to be fast. The following are some of
+When including the same icon multiple times on one page, the same string (SVG content)
+is inserted repeatedly in the HTML source. However, since most icons are small (just a few
+hundred bytes), this overhead is negligible in most cases (especiall wthen proper
+transfer encoding like e.g. gzip is enabled).
+
+On the server side, the UX Icons component is designed to be fast. The following are some of
 the optimizations made to ensure the best performance possible.
 
 Caching


src/Icons/doc/index.rst
```diff
@@ -456,7 +456,12 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
 Performance
 -----------
 
-The UX Icons component is designed to be fast. The following are some of
+When including the same icon multiple times on one page, the same string (SVG content)
+is inserted repeatedly in the HTML source. However, since most icons are small (just a few
+hundred bytes), this overhead is negligible in most cases (especiall wthen proper
+transfer encoding like e.g. gzip is enabled).
```
smnandre, line 462: The UX Icons component is a PHP one, that does generates HTML. 

Do you think on twig documentation, you'd write "every time you include a partial, it is included" ? 

:)

smnandre:
I'm personally voting against this change (for the reasons explained [in the original issue](https://github.com/symfony/ux/issues/2782#issuecomment-2917339106) and [here](https://github.com/symfony/ux/pull/2787#discussion_r2113070686))
* this is not something that solves a real need to me
* .... and instead gives importance to something that has none for 99.9% of users

Let's see if other have a different opinion

kbond:
I'm against it also, I'm concerned this could scare users off. With modern browsers/compression/http2, I'm not convinced this is a real issue. We originally (pre-release) had a "deferred" feature to fix this "issue" but I saw numbers that showed it was nothing but a micro-optimization.

ThomasLandauer:

smnandre:

PR #2788, opened 2025-05-28
[Icons] Docs: Removing broken link

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Page: https://symfony.com/bundles/ux-icons/current/index.html#locking-on-demand-icons

The target is just a few paragraphs above, so there's no need for a link IMO.

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -272,7 +272,7 @@ the ``assets/icons/`` directory. You can think of importing an icon as *locking 
 Locking On-Demand Icons
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-You can *lock* (import) all the `*on-demand* <Icons On-Demand>`_ icons you're using in your project by
+You can *lock* (import) all the *on-demand* icons you're using in your project by
 running the following command:
 
 .. code-block:: terminal


src/Icons/doc/index.rst
```diff
@@ -272,7 +272,7 @@ the ``assets/icons/`` directory. You can think of importing an icon as *locking
 Locking On-Demand Icons
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-You can *lock* (import) all the `*on-demand* <Icons On-Demand>`_ icons you're using in your project by
+You can *lock* (import) all the `*on-demand* icons you're using in your project by
```
smnandre:
```suggestion
You can *lock* (import) all the *on-demand* icons you're using in your project by
```

smnandre (changes requested):

ThomasLandauer:
Question: What's the purpose of the "Icon Set" table at https://symfony.com/bundles/ux-icons/current/index.html#icon-sets ?
The message that there are **many** icons available is already included at the very top (second paragraph):
> UX Icons gives you a direct access to over **200,000** vector icons from popular icon sets such as **FontAwesome, Bootstrap Icons, Tabler Icons, Google Material Design Icons,** etc.

smnandre:
"many icons available" is not the same information than "what is the prefix for the most commonly used icon set"

:)

kbond (approved):

ThomasLandauer:
OK, so the purpose of the table is to show some prefixes?
What I find confusing is that from the 200,000 icons, you're listing only a small fraction (~34,000) in the table. This creates the impression that the table is incomplete. On the other hand, having a *complete* table is certainly a maintenance nightmare.
So where do people get the prefixes for the other sets?

smnandre:
> So where do people get the prefixes for the other sets?

image

smnandre (approved):
Still, thank you for the fix!

smnandre:
Thank you Thomas.

PR #2781, opened 2025-05-28
[Icons] Docs: Some typos

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Page: https://symfony.com/bundles/ux-icons/current/index.html

PR #2789, opened 2025-05-28
[Icons] Docs: Fixing links to `symfony/ux-twig-component`

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Page: https://symfony.com/bundles/ux-icons/current/index.html

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -191,7 +191,7 @@ tag if the ``symfony/ux-twig-component`` package is installed:
 
 .. tip::
 
-    To use the HTML syntax, the ``symfony/ux-twig-component`` package must be
+    To use the HTML syntax, the `symfony/ux-twig-component`_ package must be
     installed in your project.
 
 Downloading Icons
@@ -532,7 +532,7 @@ returning the HTML output.
 
 .. warning::
 
-    The <twig:ux:icon> component does not support embedded content.
+    The ``<twig:ux:icon>`` component does not support embedded content.
 
     .. code-block:: html+twig
 
@@ -610,6 +610,7 @@ Learn more
 .. _`ux.symfony.com/icons`: https://ux.symfony.com/icons
 .. _`Iconify`: https://iconify.design
 .. _`symfony/asset-mapper`: https://symfony.com/doc/current/frontend/asset_mapper.html
+.. _`symfony/ux-twig-component`: https://symfony.com/bundles/ux-twig-component/current/index.html
 .. _`W3C guide about SVG icon accessibility`: https://design-system.w3.org/styles/svg-icons.html#svg-accessibility
 .. _`Bootstrap Icons`: https://icons.getbootstrap.com/
 .. _`Boxicons`: https://boxicons.com/


src/Icons/doc/index.rst
```diff
@@ -526,13 +526,13 @@ TwigComponent
...
 The ``ux_icon()`` function is optimized to be as fast as possible. To deliver the
-same level of performance when using the HTML syntax (``<twig:ux:icon name="..." />``),
+same level of performance as if using the HTML syntax (``<twig:ux:icon name="..." />``),
 the TwigComponent overhead is reduced by calling the IconRenderer immediately and
 returning the HTML output.
 
 .. warning::
 
-    The <twig:ux:icon> component does not support embedded content.
+    The `symfony/ux-twig-component`_ does not support embedded content.
```
smnandre: ? I think you're missreading this.

The <twig:ux:icon> component does not support embedded content. 

The symfony/ux-twig-component has nothing to do with it, and I don't want to imply anything else regarding twig component.

src/Icons/doc/index.rst
```diff
@@ -526,13 +526,13 @@ TwigComponent
 ~~~~~~~~~~~~~
 
 The ``ux_icon()`` function is optimized to be as fast as possible. To deliver the
-same level of performance when using the HTML syntax (``<twig:ux:icon name="..." />``),
+same level of performance as if using the HTML syntax (``<twig:ux:icon name="..." />``),
```
smnandre: No, the original meaning is the current one. The HTML syntax is the name of the syntax of TwigComponents that looks like HTML. 

So here it's not "as if" but "when"

src/Icons/doc/index.rst
```diff
@@ -191,7 +191,7 @@ tag if the ``symfony/ux-twig-component`` package is installed:
 
 .. tip::
 
-    To use the HTML syntax, the ``symfony/ux-twig-component`` package must be
+    To use the HTML syntax, the `symfony/ux-twig-component`_ package must be
```
smnandre, line 194: 👍

smnandre (changes requested):
Thanks for the link!

Let's revert the text changes :)

src/Icons/doc/index.rst
ThomasLandauer: > I think you're missreading this.

Maybe, but I'm certainly not *misclicking* it :-)
=> What's currently on the page is a broken link!

src/Icons/doc/index.rst
ThomasLandauer: What you want to say is:
> The twig syntax has the same performance as the HTML syntax.

Right?

Same as in https://github.com/symfony/ux/pull/2781#pullrequestreview-2876147891 - let's switch the parts:

> The TwigComponent overhead is reduced to deliver the same level of performance when using the HTML syntax

=> No. Cause when I use the twig syntax, I cannot use the HTML syntax at the same time.

src/Icons/doc/index.rst
ThomasLandauer: OK, now I see: In fact, there was just the second backtick missing ;-)
The word "component" was misleading me; so I changed that to "tag".

src/Icons/doc/index.rst
smnandre: Not at all.

What we do say here is : using the Twig function is the fastest possible. When using the Twig Component with the HTML Syntax, we have developed tactics.

Said in another way: 

"To deliver the same level of performance when using the HTML syntax... the UX Icon component [...]"

src/Icons/doc/index.rst
```diff
@@ -526,13 +526,13 @@ TwigComponent
...
 The ``ux_icon()`` function is optimized to be as fast as possible. To deliver the
-same level of performance when using the HTML syntax (``<twig:ux:icon name="..." />``),
+same level of performance as if using the HTML syntax (``<twig:ux:icon name="..." />``),
 the TwigComponent overhead is reduced by calling the IconRenderer immediately and
 returning the HTML output.
 
 .. warning::
 
-    The <twig:ux:icon> component does not support embedded content.
+    The ``<twig:ux:icon>`` tag does not support embedded content.
```
smnandre: No. I'm sorry. It's not a question of tag but a question of special Twig Component

smnandre (changes requested):

src/Icons/doc/index.rst
smnandre: => What's currently on the page is a broken link!
 
 Ho i saw that, and thank you for reporting it! 
 
 But you changed the meaning on the way and this is another thing that fixing something :)

src/Icons/doc/index.rst
ThomasLandauer: OK, I finally got what you mean. But this is impossible to get across with the current sentence structure. I'll come up with a better suggestion.
But one question first: Is `ux_icon()` now faster or are they the same?

ThomasLandauer:
Sorry, but I'm starting to lose the overview here. I now brought back the "component"in this sentence:
> The ``<twig:ux:icon>`` component does not support embedded content.

=> But in fact this entire paragraph needs to be removed, as per what you're saying at https://github.com/symfony/ux/issues/2796#issuecomment-2920631290
But I can't do that right now - need to see how the page looks like after all my changes... ;-)

And reduce the text to:

> Due to internal optimizations, there is no real speed difference between the Twig and the HTML syntax.

src/Icons/doc/index.rst
kbond: > But one question first: Is ux_icon() now faster or are they the same?

If they are in fact, the same, what about removing this paragraph entirely?

ThomasLandauer: > what about removing this paragraph entirely?

I didn't dare to ask that :-)

No, seriously - what about shortening this to something like:
> Due to internal optimizations, there is no real speed difference between the Twig and the HTML syntax.

kbond: > Due to internal optimizations, there is no real speed difference between the Twig and the HTML syntax.

Perfect! They don't need to know how the sausage is made!

kbond: But let's keep this PR focused on just the link fix and change this in #2795.

kbond (approved):

src/Icons/doc/index.rst
ThomasLandauer: > know how the sausage is made

A Canadian saying? :-) But you don't eat that much sausage over there, do you?

OK, I reverted it here. But that other PR is a mess already, so I won't add more stuff there...

smnandre (approved):

smnandre:
Thank you Thomas.

PR #2793, opened 2025-05-29
[Icons] Docs: Merging code blocks; language improvements

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Page: https://symfony.com/bundles/ux-icons/current/index.html#installation

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -3,9 +3,9 @@ Symfony UX Icons
 
 The ``symfony/ux-icons`` package offers simple and intuitive ways to render
 SVG icons in your Symfony application. It provides a Twig function to include
-any local or remote icons from your templates.
+both local and remote icons in your templates.
 
-UX Icons gives you a direct access to over 200,000 vector icons from popular
+UX Icons gives you direct access to over 200,000 vector icons from popular
 icon sets such as FontAwesome, Bootstrap Icons, Tabler Icons, Google Material
 Design Icons, etc.
 
@@ -16,14 +16,8 @@ Installation
 
     $ composer require symfony/ux-icons
 
-HTTP Client for On-Demand Icons
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If you plan to use provided icon sets, make sure that you have the HTTP client installed:
-
-.. code-block:: terminal
-
-    $ composer require symfony/http-client
+    # To use provided on-demand icon sets, you also need HTTP client:
+    $ composer require symfony/http-client    
 
 SVG Icons
 ---------
@@ -39,7 +33,7 @@ your own.
 Icon Names
 ~~~~~~~~~~
 
-Icons are referenced using an unique identifier that follows one of the following syntaxes:
+Icons are referenced using a unique identifier using one of the following syntaxes:
 
 * ``prefix:name``  (e.g. ``mdi:check``, ``bi:check``, ``editor:align-left``)
 * ``name`` only (e.g. ``check``, ``close``, ``menu``)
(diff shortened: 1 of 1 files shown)


src/Icons/doc/index.rst
```diff
@@ -1,11 +1,11 @@
 Symfony UX Icons
 ================
 
-The ``symfony/ux-icons`` package offers simple and intuitive ways to render
+The ``symfony/ux-icons`` package offers a simple and intuitive way to render
 SVG icons in your Symfony application. It provides a Twig function to include
```
smnandre: UX Icons offers multiple ways to render your SVG icons. Plural is correct here.

src/Icons/doc/index.rst
```diff
@@ -16,14 +16,8 @@ Installation
...
     $ composer require symfony/ux-icons
 
-HTTP Client for On-Demand Icons
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If you plan to use provided icon sets, make sure that you have the HTTP client installed:
-
-.. code-block:: terminal
-
-    $ composer require symfony/http-client
```
smnandre, line 26: Please do not remove this. It has been discussed already, and this information is usefull.

src/Icons/doc/index.rst
```diff
@@ -50,7 +44,7 @@ The icon ``name`` is the same as the file name without the file extension (e.g.
 
     The name must match a standard ``slug`` format: ``[a-z0-9-]+(-[a-z0-9])+``.
 
-Depending on your configuration, the ``prefix`` can be the name of an icon set, a directory
+Depending on your `configuration`_, the ``prefix`` can be the name of an icon set, a directory
```
smnandre, line 47: 👍  nice addition here

src/Icons/doc/index.rst
ThomasLandauer: I didn't remove it, I just moved it upwards.

kbond: Given that we have improved the DX here, I'm ok removing this as a "doc section". Maybe a tip or note admonition though so it's still obvious?

ThomasLandauer: Well, it's in the very first code block on the page :-)
What would you suggest?

kbond: A tip just below the `composer require symfony/ux-icons` code block.

kbond: Probably a second tip about installing ux-twig-component that links to a single html-syntax section as proposed in https://github.com/symfony/ux/pull/2795#pullrequestreview-2879792790

smnandre:
I really don’t want to clutter the installation instructions for users who just want to display icons.

That’s in line with how many Symfony docs are written:

- https://symfony.com/doc/current/templates.html
- https://symfony.com/doc/current/validation.html
- https://symfony.com/doc/current/notifier.html
- https://symfony.com/doc/current/forms.html

This first install block needs to stay short and intuitive. Adding lines that aren't strictly necessary for most users just adds noise.

That said, I’m totally open to making the rest better..  (optional dependencies and/or how we explain the HTML syntax)

My personal take would be to have a separate “HTML Syntax” section with a code sample and a composer require, and then point to it when relevant. Same thing for the HTTP client.

But placing all of that right after “install” and before showing how it actually works still feels too soon to me.

smnandre (approved):

ThomasLandauer:

smnandre:

Kocal:
Thank you @ThomasLandauer.

PR #2795, opened 2025-05-29
[Icons] Docs: Merging HTML and Twig syntaxes

ThomasLandauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

https://symfony.com/bundles/ux-icons/current/index.html

* The fact that you need `symfony/ux-twig-component` was explained three times.
* The section "HTML Syntax" was included twice: https://symfony.com/bundles/ux-icons/current/index.html#html-syntax and https://symfony.com/bundles/ux-icons/current/index.html#html-syntax-1
* Having the section "Icon Sets" *between* the Twig and the HTML syntax makes no sense IMO, so I started by moving the HTML syntax upwards.
* Then I thought that it's even better to *merge* these two code blocks:
  *  This shows the differences between the syntaxes better
  * And presents the HTML syntax as an *equal* alternative (not some exotic experimental feature ;-)
* The default attribute config is pretty easy to understand, so it doesn't need a fullblown code sample; linking to the config section should be enough IMO.

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -70,7 +70,7 @@ To include an icon in your template, two syntaxes are available.
         $ composer require symfony/ux-twig-component
 
 
-.. code-block:: twig
+.. code-block:: html+twig
 
     {# Includes the contents of 'assets/icons/user-profile.svg' in the template: #}
     {{ ux_icon('user-profile') }}
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -59,29 +59,37 @@ refers to the icons located in the ``header`` directory.
 Loading Icons
 -------------
 
+To include an icon in your template, two syntaxes are available.
+
+.. tip::
+
+    To use the HTML syntax, you need the ``symfony/ux-twig-component`` package:
+
+    .. code-block:: terminal
+
+        $ composer require symfony/ux-twig-component
+
+
 .. code-block:: twig
 
-    {# includes the contents of the 'assets/icons/user-profile.svg' file in the template #}
+    {# Includes the contents of 'assets/icons/user-profile.svg' in the template: #}
     {{ ux_icon('user-profile') }}
+    {# Same in alternative HTML syntax: #}
+    <twig:ux:icon name="user-profile" />
 
-    {# icons stored in subdirectories must use the 'subdirectory_name:file_name' syntax
-       (e.g. this includes 'assets/icons/admin/user-profile.svg') #}
+    {# Includes 'assets/icons/admin/user-profile.svg': #}
     {{ ux_icon('admin:user-profile') }}
+    <twig:ux:icon name="admin:user-profile" />
 
-    {# this downloads the 'user-solid.svg' icon from the 'Flowbite' icon set via ux.symfony.com
-       and embeds the downloaded SVG contents in the template #}
+    {# Adding a CSS class or other attribute to the `<svg>` element: #}
+    {{ ux_icon('user-profile', {class: 'w-4', 'aria-hidden': 'true' }) }}
+    <twig:ux:icon name="user-profile" class="w-4" aria-hidden="true" />
+
+    {# Download the 'user-solid.svg' icon from the 'Flowbite' icon set via ux.symfony.com: #}
     {{ ux_icon('flowbite:user-solid') }}
+    <twig:ux:icon name="flowbite:user-solid" />
 
-The ``ux_icon()`` function defines a second optional argument where you can
-define the HTML attributes added to the ``<svg>`` element:
-
-.. code-block:: html+twig
-
-    {{ ux_icon('user-profile', {class: 'w-4 h-4'}) }}
-    {# renders <svg class="w-4 h-4"> ... </svg> #}
-
-    {{ ux_icon('user-profile', {height: '16px', width: '16px', 'aria-hidden': true}) }}
-    {# renders <svg height="16" width="16" aria-hidden="true"> ... </svg> #}
+You can set default attributes for all icons in your `Configuration`_.
 
 
 Icon Sets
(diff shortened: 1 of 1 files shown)


smnandre (changes requested):
UX Icons is a bundle that does not require Twig Component. 

You suggestion here will delete/remove vital documentation about basic usage of this bundle.

This is not something anyone would benefit. And I must admit I'm not sure to understand your point / end goal on these recent PR. 

I highly value the contributions you make on Symfony, and even more the ones you do on Symfony UX. Here, i don't see the positive / solution-oriented mindset I usually see in your work... is there anything I missread here ? Or anything we could discuss to clear any missunderstandings ?

src/Icons/doc/index.rst
```diff
@@ -314,49 +298,6 @@ Rendering Icons
...
-    <twig:ux:icon name="sub-dir:user-profile" class="w-4 h-4" />
-
-    {# Renders "flowbite:user-solid" from ux.symfony.com #}
-    <twig:ux:icon name="flowbite:user-solid" />
-
-.. note::
-
-    ``symfony/ux-twig-component`` is required to use the HTML syntax.
-
-Default Attributes
```
kbond, line 337: I'd prefer we keep this section.

kbond (commented):
What about having a single, top level "HTML Syntax" section, where this is the only place we talk about/show examples of it?

ThomasLandauer:
> UX Icons is a bundle that does not require Twig Component.

Indeed, that's explained at https://github.com/symfony/ux/blob/b954ff73a6a48e2898a1310a6ea30d1e2395837b/src/Icons/doc/index.rst#L66

> You suggestion here will delete/remove vital documentation about basic usage of this bundle.

I know that reviewing docs changes on GitHub is a nightmare :-)
So just tell me what you think is missing, and I'll tell you where it is now.

> This is not something anyone would benefit. And I must admit I'm not sure to understand your point / end goal on these recent PR.

My main goal is to explain stuff just *once* and not *three times* :-)

ThomasLandauer:
> What about having a single, top level "HTML Syntax" section, where this is the only place we talk about/show examples of it?

Yeah, I thought about that too.
In the end, I went for the "mixed" code block, to present both as **equally good** alternatives.

But if you want, I can split them. This would make *comparing* the syntaxes harder. But later on (after people have decided for one), give them a nicer "reference".

kbond:
I'm thinking of it in terms of "directing someone to the html syntax" - what section do I sent them to currently?

@smnandre wdyt?

src/Icons/doc/index.rst
ThomasLandauer: I shortened it to one sentence here: https://github.com/symfony/ux/blob/b954ff73a6a48e2898a1310a6ea30d1e2395837b/src/Icons/doc/index.rst#L92

Together with the explanation in the config section at https://symfony.com/bundles/ux-icons/current/index.html#full-configuration I think it's clear how this works.

Which part would you say is important to keep?

ThomasLandauer:
You mean for reference on the HTML syntax? I'd say just here: https://github.com/symfony/ux/blob/b954ff73a6a48e2898a1310a6ea30d1e2395837b/src/Icons/doc/index.rst#L59

Or do you mean "directing" them to switch to HTML syntax? (i.e. recommending it)

IMO, the entire syntax is so easy/lightweight, that after having written it 2-3 times, people won't need much reference anymore.

src/Icons/doc/index.rst
kbond: I like the clear example. A single sentence would get lost.

I guess I'm not understanding your motivation here. Why are you suggesting removing this section but not Icon Aliases below?

kbond:
I think mixing syntaxes throughout the docs might be confusing. Better to show a consistent syntax, then "btw, you can use this alternate syntax"

src/Icons/doc/index.rst
ThomasLandauer: > A single sentence would get lost.

Well, yes and no. I you merge everything I'm suggesting, then the page is only half as long ;-)
And if you start expanding the not-so-important parts (like this one), you need to expand/repeat the real important parts too...

And it's still shown *twice*: In the text, and in the config codeblock.

> Why are you suggesting removing this section but not Icon Aliases below?

Mainly cause I haven't looked into it ;-)
And cause I don't understand that feature - so I thought I'd better not touch it...

ThomasLandauer:
Uhm, there is no "throughout" anymore. The (almost) only other place with syntax is https://symfony.com/bundles/ux-icons/current/index.html#icon-aliases - which you just "urged" me to remove too (see https://github.com/symfony/ux/pull/2795#discussion_r2114885611 :-)

src/Icons/doc/index.rst
```diff
@@ -59,29 +59,37 @@ refers to the icons located in the ``header`` directory.
...
+    {{ ux_icon('admin:user-profile') }}
+    <twig:ux:icon name="admin:user-profile" />
+
+    {# Adding a CSS class or other attribute to the `<svg>` element: #}
+    {{ ux_icon('user-profile', {class: 'w-4', 'aria-hidden': 'true' }) }}
+    <twig:ux:icon name="user-profile" class="w-4" aria-hidden="true" />
+
+    {# Download the 'user-solid.svg' icon from the 'Flowbite' icon set via ux.symfony.com: #}
+    {{ ux_icon('flowbite:user-solid') }}
+    <twig:ux:icon name="flowbite:user-solid" />
```
smnandre, line 90: The reason for not mixing them like this is that some users might be confused about which syntax to use, and whether the HTML shown is the generated output

I really don’t think mixing the syntaxes in the first paragraph is a good idea: it doesn’t help clarify anything or solve any actual problem.

smnandre (changes requested):

src/Icons/doc/index.rst
```diff
@@ -59,29 +59,37 @@ refers to the icons located in the ``header`` directory.
...
+    To use the HTML syntax, you need the ``symfony/ux-twig-component`` package:
 
-    {# this downloads the 'user-solid.svg' icon from the 'Flowbite' icon set via ux.symfony.com
-       and embeds the downloaded SVG contents in the template #}
-    {{ ux_icon('flowbite:user-solid') }}
+    .. code-block:: terminal
+
+        $ composer require symfony/ux-twig-component
 
-The ``ux_icon()`` function defines a second optional argument where you can
```
smnandre, line 75: For both the Twig function and the HTML syntax, I think better to explain the first argument properly before jumping into customization.

It follows UX* rules: start simple, then show how to tweak things.. That way, users arent overwhelmed and can follow along more easily.

So I'd move back the CSS / height / etc. examples after the Flowbite one.

* the science field, not the repository^^

src/Icons/doc/index.rst
smnandre: Is the shortest documentation a goal we have? 

Configuring default attributes is a question we had a lot of times, so dedicating some example was decided. It's not only luck or randomness that did drive every choice here 😅

ThomasLandauer:

smnandre:

javiereguiluz:
Folks, if there's interest, I think we could show both syntaxes at the same time, and allow user to select which one they prefer.

Same as we do in Symfony Docs, we could use a `configuration-block` like this:

```rst
.. configuration-block::

    .. code-block:: ux-twig

        {# Includes 'assets/icons/admin/user-profile.svg': #}
        {{ ux_icon('admin:user-profile') }}

    .. code-block:: ux-html

        {# Includes 'assets/icons/admin/user-profile.svg': #}
        <twig:ux:icon name="admin:user-profile" />
```

The `ux-twig` and `ux-html` labels are up to debate, but must be different from the ones used in Symfony (`yaml`, `xml`, `php-attributes`, etc.)

If you like this, tell us and we'd make the (minor) changes required in our RST support to add support for these two new code block labels. Thanks!

kbond:
@javiereguiluz that's a great idea!

smnandre:
@javiereguiluz I'd love this !

Twig vs UX  ?

Kocal:

Thanks!

Status: Needs Work

PR #2074, opened 2024-08-17
chore: run PHP-CS-Fixer

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

Fabbot still misses our PHP-CS-Fixer configuration (tracked in #2014) :(

PR #2803, opened 2025-05-31
[CI] Re-add PHP-CS-Fixer

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Follow #2802, #2074, #2014 and more. 

I feel like Fabbot **only runs on modified files** for a given PR, but sometimes it does not really help (e.g.: if you untouched files contain old references to a deleted class).

This behavior is understandable for symfony/symfony, which contains a tons of code and is super active, but that's not the case symfony/ux, we have very less code and less activity.

**EDIT:** Well, I'm still not sure... PHP-CS-Fixer is showing me fixes for `src/LiveComponent/src/Util/TypeHelper.php`, added in #2778, but it was totally ignored by Fabbot... 😫  

So, I suggest to re-add PHP-CS-Fixer run in the CI to ensure all files are correctly CS-fixed.

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -29,6 +29,17 @@ jobs:
             - run: yarn --immutable
             - run: yarn ci
 
+    coding-style-php:
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: actions/checkout@v4
+            -   uses: shivammathur/setup-php@v2
+                with:
+                    php-version: '8.1'
+            -   run: composer install
+            -   name: php-cs-fixer
+                run: ./vendor/bin/php-cs-fixer fix --dry-run --diff
+
     phpstan:
         name: PHPStan
         runs-on: ubuntu-latest
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -23,10 +23,10 @@ use Symfony\Component\Serializer\Exception\ExceptionInterface as SerializerExcep
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 use Symfony\Component\TypeInfo\Type;
-use Symfony\Component\TypeInfo\TypeIdentifier;
 use Symfony\Component\TypeInfo\Type\CollectionType;
 use Symfony\Component\TypeInfo\Type\ObjectType;
 use Symfony\Component\TypeInfo\Type\WrappingTypeInterface;
+use Symfony\Component\TypeInfo\TypeIdentifier;
 use Symfony\Component\Uid\AbstractUid;
 use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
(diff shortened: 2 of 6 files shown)


kbond (approved):
This works for me. I don't think fabbot ever worked for the website code.

Once merged, I'll disable the webhook?

Kocal:
> Once merged, I'll disable the webhook?

Mmmmh... I don't know, these checks can still be pretty useful (when they are correctly ran): 
image

Can we keep both PHP-CS-Fixer and Fabbot?

kbond:
Yeah no rush to remove it.

smnandre:
100% 👍 

 ... and i'm thinking we should either find a way to exclude false positives in fabbot... or find a way to adapt our code to get green results when it runs

.github/workflows/code-quality.yaml
```diff
@@ -29,6 +29,17 @@ jobs:
             - run: yarn --immutable
             - run: yarn ci
 
+    coding-style-php:
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: actions/checkout@v4
+            -   uses: shivammathur/setup-php@v2
+                with:
+                    php-version: '8.1'
```
smnandre, line 38:
```suggestion
                    php-version: '8.3'
```

I'm thinking tests are here to check our code works, but i'd rather use recent PHP versions on CI, for performance reasons and because it will be what most people will use locally too

wdyt ?

smnandre, line 38: we could use the php-cs-fixer tool provided by shivammathur also

.github/workflows/code-quality.yaml
```diff
@@ -29,6 +29,17 @@ jobs:
...
+            -   run: composer install
+            -   name: php-cs-fixer
+                run: ./vendor/bin/php-cs-fixer fix --dry-run --diff
```
smnandre, line 41:
```suggestion
                run: ./vendor/bin/php-cs-fixer fix --dry-run --diff --show-progress=none
```

I have a doubt to be honest, it may be automatically set

.github/workflows/code-quality.yaml
kbond: I believe you need to run php-cs-fixer on the lowest version of PHP you support. I've been burned in the past but maybe it's not a problem anymore.

smnandre (approved):

.github/workflows/code-quality.yaml
Kocal: That's better to run CS tools to the lowest supported PHP version. 
Some of them infers configuration based on running PHP version, and we don't want them to suggest more modern code that would cause lower PHP versions to break

smnandre: Ok then let's ignore :)

Kocal: But we won't be able to run it locally, see https://github.com/symfony/ux/pull/617

kbond: Yep, easier to use the installed version so you don't have to duplicate version constraints. I forgot that I don't include in my packages - the ci fixes automatically.

PR #2792, opened 2025-05-29
[Map] Add option to configure attribution and zoom control

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

Add option to configure attribution and zoom control for leaflet

Issue #2445, opened 2024-12-11
[Map] Leaflet vector layers


dannyvw:
Is it possible to use vector layers like https://github.com/Esri/esri-leaflet-vector? What is the best way to implement this?

Kocal:
Hi, I think yes, see https://symfony.com/bundles/ux-map/current/index.html#interact-with-the-map
1. create your own Stimulus Controller and use-it to render the map (configure `data-controller`) 
2. in this controller, import `esri-leaflet`
3. in this controller, listen to the event `ux:map:connect`, access `L` and `map` instances from `event.details`

Prometee:
Hi @Kocal!

I tried what you explained with another leaflet plugin (`Leaflet.Editable`) and I ended up retrieving `L` var using `window.L` because `event.details.L` is not the same variable and does not have the plugin instantiated.

```javascript
import {Controller} from "@hotwired/stimulus";
import 'leaflet-editable';

//  listen to the event ux:map:connect and pre-connect to set editable: true

_onConnect(event) {
        const { map, L } = event.detail;
        
        // L.Editable does not exist
}
```

```javascript

import {Controller} from "@hotwired/stimulus";
import 'leaflet-editable';

//  listen to the event ux:map:connect and pre-connect to set editable: true

_onConnect(event) {
        const { map } = event.detail;
        const L = window.L;
        
        // L.Editable exists
}
```

Do you see why it's not the same variable ?

Kocal:
Hi, and sorry for the late reply.

It is not the same variable because:
1. in Symfony UX, we import use `import * as L from 'leaflet'` (https://github.com/symfony/ux/blob/2.x/src/Map/src/Bridge/Leaflet/assets/src/map_controller.ts#L10) which create a new object 
2. but in `leaflet-editabe`, it use `import L from 'leaflet'` (https://cdn.jsdelivr.net/npm/leaflet-editable@1.3.0/+esm)

Which can be equivalent, but not identical:
```js
    _onConnect(event) {
        console.log(event.detail.L); // Module
        console.log(window.L); // Object
        console.log(event.detail.L === window.L);
    }
```
<img width="511" alt="image" src="https://github.com/user-attachments/assets/3920213f-abbb-4a60-a149-5a034056e10b" />

If I edit the Leaflet `map_controller.ts` to use `import L from 'leaflet'`, then it references the same object and `event.detail.L.Editable` exists :
[...]

PR #2810, opened 2025-06-02
[Map] Add possibility to configure extra options

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

TODO

- Add/fix tests
- Add docs
- Rebase when https://github.com/symfony/ux/pull/2792 is merged

Related to https://github.com/symfony/ux/issues/2445#issuecomment-2571609726

Usage:
```php
new LeafletOptions(extra: ['maxZoom' => 1])
```

### DIFF
--- a/src/Map/src/Bridge/Google/assets/dist/map_controller.d.ts
+++ b/src/Map/src/Bridge/Google/assets/dist/map_controller.d.ts
@@ -1,7 +1,9 @@
 import type { LoaderOptions } from '@googlemaps/js-api-loader';
 import AbstractMapController from '@symfony/ux-map';
 import type { Icon, InfoWindowWithoutPositionDefinition, MarkerDefinition, Point, PolygonDefinition, PolylineDefinition } from '@symfony/ux-map';
-type MapOptions = Pick<google.maps.MapOptions, 'mapId' | 'gestureHandling' | 'backgroundColor' | 'disableDoubleClickZoom' | 'zoomControl' | 'zoomControlOptions' | 'mapTypeControl' | 'mapTypeControlOptions' | 'streetViewControl' | 'streetViewControlOptions' | 'fullscreenControl' | 'fullscreenControlOptions'>;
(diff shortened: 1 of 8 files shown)


smnandre:
As this is serialized, I guess we should make some minimal type check on the content of extra, no ? 

cc @Kocal

dannyvw:
@smnandre @Kocal Do you have something in mind?

Kocal (changes requested):
`extra` is already used for `Marker` and cie to let the developper pass extra data from the back to the front, but the difference with the actual system is that `extra` **are not used** by UX Map, these `extra` data are only used in `ux:map:*` events.

Instead, I suggest you to implement `rawOptions` on Map options front-side, nothing should be modified backend-side.

Thanks!

dannyvw:
> `extra` is already used for `Marker` and cie to let the developper pass extra data from the back to the front, but the difference with the actual system is that `extra` **are not used** by UX Map, these `extra` data are only used in `ux:map:*` events.
> 
> Instead, I suggest you to implement `rawOptions` on Map options front-side, nothing should be modified backend-side.
> 
> Thanks!

If we only change the front-side how can we set options in the backend which are passed to the front? We want to set some options to the map configuration based on configuration from Symfony DI. Example: for the vector layers we need to pass a layerName and apiKey.

Kocal:
> If we only change the front-side how can we set options in the backend which are passed to the front? We want to set some options to the map configuration based on configuration from Symfony DI. Example: for the vector layers we need to pass a layerName and apiKey.

This could be done through `extra` property that you will intercept through `ux:map:pre-connect` event, and pass it to `rawOptions` which.. does not exist yet :)

You can't pass `extra` to `new google.maps.Map()` or `L.Map()`. 

I'm not fan how things are done for Map and options, raw options, etc, I will open a new PR asap.

Kocal:
Replaced by #2863, thanks anyway for your contribution :)

PR #2811, opened 2025-06-02
[Map] Do not override `fitBoundsToMarkers` when using LiveComponent

### LINKED ISSUE #2774: [Map] FitBoundsToMarkers
When creating a map you can set zoom/center and fitBoundsToMarkers (true). This works fine because when you don't have markers the zoom/center will be used from the configuration. And when there are markers fitBoundsToMarkers will be activated. 

When using a live component for adding new markers then fitBoundsToMarkers is always set to false when there is a zoom/center configured (https://github.com/symfony/ux/blob/2.x/src/Map/src/Map.php#L181). Both options can exist together. Now you get the situation that as there a no markers you see a grey map instead of the default zoom/center. 

It is possible to use the configured value for fitBoundsToMarkers?

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #2774  
| License       | MIT

### DIFF
--- a/src/Map/CHANGELOG.md
+++ b/src/Map/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 2.27
+
+-   The `fitBoundsToMarkers` option is not overridden anymore when using the `Map` LiveComponent, but now respects the value you defined.
+    You may encounter unwanted behavior when adding/removing elements to the map.
+    To use the previous behavior, you must call `$this->getMap()->fitBoundsToMarkers(false)` in your LiveComponent's live actions
+
 ## 2.26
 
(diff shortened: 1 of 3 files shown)


Kocal (commented):
I was a little skeptical at first, but it does make more sense. 

Can you : 
- modify the changelog accordingly, since it could be interpreted as a breaking-change
- modify the documentation on using Map with LiveComponents, and add an example with `$this->getMap()->fitBoundsToMarkers(false)`? 


Thank you very much!

Kocal (approved):
Modified some documentation

Kocal:
Thank you @dannyvw.

PR #2817, opened 2025-06-04
[Autocomplete] Add `maxOptions: null` to autocomplete with remote data

kevinmade:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

This fixes an issue where the autocomplete component stops loading additional results after 50 due to default limit. The maxOptions value has been set to null to allow loading more than 50 options from the `autocomplete_url`.


smnandre (changes requested):
Hey @kevinmade ! Thank you for this suggestion. 

You need to make your changes on the source code of the controller. 

Here, you changes are done on the `dist` file, that will be re-computed from the source everytime someone make a change, and your suggestion will be lost.

Also, it will be easier to review with a bit of context, some link to documentation or source code impacted, etc.

kevinmade:
### Before
https://github.com/user-attachments/assets/8bd779e1-ce69-4713-a418-7c0dde8317b8

### After
https://github.com/user-attachments/assets/f2b96d95-0762-4315-ae0f-807efac728d9

Kocal:
When seeing your demo, I think it breaks the spirit of "autocompletion". To me, you have to type something and the component autocomplete you some results, but here, it looks like a lazy-loaded `<select>` and it feels wrong to me.

I'm not really against your suggestion, but not in this way. Instead, can we keep it 50 items by default **but** [introduce a new option](https://symfony.com/bundles/ux-autocomplete/current/index.html#form-options-reference)?

Kocal:

Thanks!

Status: Needs Work

Issue #2812, opened 2025-06-02
[TwigComponent] Improve performance for complex use-cases


adri:
Firstly, great work and nice to see a component based way of working for Symfony.

## Scenario
Creating components like [shadcn](https://ui.shadcn.com) with Twig Components is great. However, when using a lot of them, the performance degrades quickly. Especially when adding popovers with buttons in them for each cell.

1. There seems to be overhead with the rendering of each component
2. The response size gets very large (4 MB in the example) even compressed it takes quite a while for the browser to download and parse

### Reproducer 
<img width="1904" alt="Image" src="https://github.com/user-attachments/assets/0ebd0b1b-976d-4573-b7bb-72231d8e98b4" />

1. Checkout the [reproducer](https://github.com/adri/symfony-ux_twig_components_performance)
2. `composer install`
3. `symfony server:start`

## Solutions
I'm probably switching back to simple HTML elements again (`<button class="ghost">`) and use Tailwinds `@apply` for different variants. Of course, I'd prefer to keep using the component approach. 

Phoenix LiveView solved the 2. problem with an optimization, see [Optimization #1: splitting statics from dynamics](https://dashbit.co/blog/latency-rendering-liveview). This way the response size doesn't grow very large.

adri:
@Kocal I see you’re busy with building components like I did. I guess I don’t do something completely wrong 😉
https://github.com/symfony/ux/tree/2.x/src/Toolkit/kits/shadcn/docs/components

Kocal:
Thanks for the reproducer :) I will try to give Blackfire a try when having some free time.

@smnandre already worked on TwigComponent optimization, and IIRC, major performance impacts were found due to how [event listeners](https://symfony.com/bundles/ux-twig-component/current/index.html#events) worked.

smnandre:
Hello, 

Regarding your **2nd** point, TwigComponent only renders the HTML coded in your templates. 

<img width="1287" alt="Image" src="https://github.com/user-attachments/assets/1469429a-edc0-41fd-a1ff-34a7eb4a52b1" />

If you wanted to reduce, I think using CSS here could make you gain at least 50% of the generated HTML size.

Regarding your 1st one, it does not surprise me that much in this situation to be honest. 
- LiveComponent are not made to render tens of thousands or nested components
- Using Component for SVG icon will be less performant than using... UX Icons
- html_cva and Taildwind merge have a real cost too
[...]

PR #2821, opened 2025-06-04
[LiveComponent] Optimize `LiveComponentStack::getCurrentLiveComponent()`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Using Reflection during runtime is costly. Here I suggest using static caching to prevent unnecessary usages of Reflection.

I've re-used reproducer from #2812, but with 50 rows (otherwise I can't create a Blackfire Profile), and it reduced the rendering time by 200ms, https://blackfire.io/profiles/compare/7aa62248-9332-40b3-b6fa-58756cb17232/graph 
image

**EDIT:** see https://github.com/symfony/ux/pull/2821#issuecomment-2943045842

### DIFF
--- a/src/LiveComponent/src/Util/LiveComponentStack.php
+++ b/src/LiveComponent/src/Util/LiveComponentStack.php
@@ -24,6 +24,9 @@ use Symfony\UX\TwigComponent\MountedComponent;
  */
 final class LiveComponentStack extends ComponentStack
 {
+    /** @var array<class-string, bool> */
+    private array $cacheLiveComponents = [];
+
     public function __construct(private readonly ComponentStack $componentStack)
     {
     }
@@ -31,7 +34,8 @@ final class LiveComponentStack extends ComponentStack
     public function getCurrentLiveComponent(): ?MountedComponent
     {
         foreach ($this->componentStack as $mountedComponent) {
-            if ($this->isLiveComponent($mountedComponent->getComponent()::class)) {
+            $componentClass = $mountedComponent->getComponent()::class;
+            if ($this->cacheLiveComponents[$componentClass] ??= $this->isLiveComponent($componentClass)) {
                 return $mountedComponent;
             }
         }


src/LiveComponent/src/Util/LiveComponentStack.php
```diff
@@ -41,7 +41,9 @@ public function getCurrentLiveComponent(): ?MountedComponent
 
     private function isLiveComponent(string $classname): bool
     {
-        return [] !== (new \ReflectionClass($classname))->getAttributes(AsLiveComponent::class);
+        static $cache = [];
```
smnandre: Not sure i would use a static cache here, as --even with an instance prop-- only a fraction of current calls would be made ... wdyt ? 

(splendid quick win btw 👏 )

Kocal: I'm not sure to understand, are you suggesting adding a static property instead?

smnandre: Quite the opposite actually.. using a regular property in this service.. but in the end that does not make much change so nvm

smnandre (approved):

norkunas:
Should this use `kernel.reset`, or you think it will be a small cache and never releasing memory in worker mode will be imperceptible?

Kocal:
> Should this use `kernel.reset`, or you think it will be a small cache and never releasing memory in worker mode will be imperceptible?

Oh, that's a good point, I didn't think about it.

I think we are safe here, given this reproducer:
```php
<?php
$componentsCount = 100;
$cache = [];

echo sprintf('Memory usage before: %d MiB', memory_get_peak_usage(true) / 1024 / 1024 ) . PHP_EOL;

for ($i = 0; $i < $componentsCount; $i++) {
    $cache['\App\Twig\Components\ComponentA' . $i] = $i % 2;
}
echo sprintf('Memory usage after: %d MiB', memory_get_peak_usage(true) / 1024 / 1024 ) . PHP_EOL;

```
1. If you have 100 components (something more realistic) or 1000 components, the memory does not change: 
```
Memory usage before: 2 MiB
Memory usage after: 2 MiB
```

2. With 10k components, +2 MiB:
```
Memory usage before: 2 MiB
Memory usage after: 4 MiB
```

3. And with 100k components (???), only 13 MiB:
```
Memory usage before: 2 MiB
Memory usage after: 13 MiB
```

The memory usage is not an issue.

And `[] !== (new \ReflectionClass($classname))->getAttributes(AsLiveComponent::class)` will **always** return the exact result between requests (unless you modify your components files without redeploying your app 😅) so that's a win-win, the runtime cache is persisted during a request and also for next requests (when using FrankenPHP worker's mode, Swoole, etc...)!

Kocal:
I went for another solution, where we can completely shortcut `isLiveComponent()` call, by storing the static cache in `getCurrentLiveComponent()`. 

It gives better results, https://blackfire.io/profiles/compare/8bbc8e11-fbad-4853-bb08-9cbdcb857e50/graph:
image

smnandre:
We've been using maps of component names -> stuff in various places for some time now, and they've helped winning major optimizations, while I have not received nor read any downsides. 

But more importantly ... 

The class is used here, not the name. That means all the anonymous components in a given project count as one.

Finally, I've never heard of a project with a thousand TwigComponent classes for now... and i'm pretty sure there would be a lot of other problems before 😶

smnandre:
TL;DR; maybe let's see if there is some problem here and we will adapt accordingly :) 

(i'm talking about the kernel.reset here, not this PR that is a wonderful "little change-big effect" illustration of @Kocal's work  ❤️ )

PR #2834, opened 2025-06-12
[LiveComponent] Add support for interfaces while (de)hydrating a Doctrine entity

### LINKED ISSUE #2832: Problem while hydrating a Doctrine Entity when using an interface as alias.
- Using symfony 7.3

With Doctrine ORM, when aliasing a concrete class to an interface, the hydratation process does not work anymore.

**Doctrine configuration**
```
doctrine:
  orm:
    resolve_target_entities:
      MyInterface: App\Entity\MyRealEntity
```
**In the component**
```
    #[LiveProp()]
    public ?MyInterface $initialFormData = null;
```

**The error**
```
An exception has been thrown during the rendering of a template ("Cannot dehydrate value typed as interface
```

**The workaround**

I have a workaround but i'mnot really used about what to do to help, so, guidelines are welcome :)

Digging into DoctrineEntityHydrationExtension.php :
```
  private function objectManagerFor(string $class): ?ObjectManager
    {
        if (!interface_exists($class) && !class_exists($class)) {
            return null;
        }

        // todo cache/warmup an array of classes that are "doctrine objects"
        foreach ($this->managerRegistries as $registry) {
            foreach($registry->getManagers() as $om) {
                // this way, it resolve the interface
                if ($om->getClassMetadata($class)) {
                    return self::ensureManagedObject($om, $class);
                }
            }
            // does not work with interface
            // if ($om = $registry->getManagerForClass($class)) {
            //    return self::ensureManagedObject($om, $class);
            // }
        }

        return null;
    }

```

pink6440:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no
| Issues        | Fix #2832 
| License       | MIT

Doctrine have the feature to use interface aliased to a concrete entity class. 
This PR propose a small modification to ensure that is will also supported by the doctrine hydrator

### DIFF
--- a/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
+++ b/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
@@ -85,9 +85,9 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
             // Keep the standard way for a class
             // todo cache/warmup an array of classes that are "doctrine objects"
             foreach ($this->managerRegistries as $registry) {
-                 if ($om = $registry->getManagerForClass($class)) {
-                    return self::ensureManagedObject($om, $class);
-                 }
+                if ($om = $registry->getManagerForClass($class)) {
+                   return self::ensureManagedObject($om, $class);
+                }
             }
             return null;
         }
--- a/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
+++ b/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
@@ -82,7 +82,7 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
 
     private function objectManagerFor(string $class): ?ObjectManager
     {
-        if (!interface_exists($class) && !class_exists($class)) {
+        if (!class_exists($class) && !interface_exists($class)) {
             return null;
         }
 
--- a/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
+++ b/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
@@ -66,7 +66,7 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
             ->getIdentifierValues($object);
 
         // Dehydrate ID values in case they are other entities
-        $id = array_map(fn($id) => \is_object($id) && $this->supports($id::class) ? $this->dehydrate($id) : $id, $id);
+        $id = array_map(fn ($id) => \is_object($id) && $this->supports($id::class) ? $this->dehydrate($id) : $id, $id);
 
         switch (\count($id)) {
             case 0:
@@ -88,27 +88,23 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
 
         // todo cache/warmup an array of classes that are "doctrine objects"
         foreach ($this->managerRegistries as $registry) {
-
             // The doctrine registry does not resolve aliased interface
             // if ($om = $registry->getManagerForClass($class)) {
             //    return self::ensureManagedObject($om, $class);
             // }
-
             foreach ($registry->getManagers() as $om) {
                 // But we can resolve nicely by trying to ask each manager to get the metadata
                 try {
-                    if ($om->getClassMetadata($class) !== null) {
+                    if (null !== $om->getClassMetadata($class)) {
                         return self::ensureManagedObject($om, $class);
                     }
-
                 } catch (MappingException $e) {
                     // I did not find a nice way to check if it is because the class is really unknown
                     // It is good to check for a specific exception ?
                     // eg: \Doctrine\Persistence\Mapping\MappingException
                     // Maybe not needed, because it does not failed even when the class does not exist at all
-//                    throw $e;
+                    // throw $e;
                 }
-
             }
         }
 
--- a/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
+++ b/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
@@ -106,7 +106,7 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
                     // It is good to check for a specific exception ?
                     // eg: \Doctrine\Persistence\Mapping\MappingException
                     // Maybe not needed, because it does not failed even when the class does not exist at all
-                    throw $e;
+//                    throw $e;
                 }
 
             }
--- a/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
+++ b/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
@@ -82,8 +82,6 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
     private function objectManagerFor(string $class): ?ObjectManager
     {
         if (class_exists($class)) {
-            // Keep the standard way for a class
-            // todo cache/warmup an array of classes that are "doctrine objects"
             foreach ($this->managerRegistries as $registry) {
                 if ($om = $registry->getManagerForClass($class)) {
                     return self::ensureManagedObject($om, $class);
@@ -102,7 +100,6 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
             // Fore more details :
             // @see \Doctrine\ORM\Tools\ResolveTargetEntityListener
 
-            // todo cache/warmup an array of interfaces that are resolved "doctrine objects"
             foreach ($this->managerRegistries as $registry) {
                 foreach ($registry->getManagers() as $om) {
                     // the getClassMetaData can indeed throw an exception
--- a/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
+++ b/src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\LiveComponent\Hydration;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\ObjectManager;
+use Doctrine\Persistence\Mapping\MappingException;
 
 /**
  * Handles hydration of Doctrine entities.
@@ -27,7 +28,8 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
      */
     public function __construct(
         private iterable $managerRegistries,
-    ) {
+    )
+    {
     }
 
     public function supports(string $className): bool
@@ -61,11 +63,10 @@ class DoctrineEntityHydrationExtension implements HydrationExtensionInterface
         $id = $this
             ->objectManagerFor($class = $object::class)
             ->getClassMetadata($class)
-            ->getIdentifierValues($object)
-        ;
+            ->getIdentifierValues($object);
 
         // Dehydrate ID values in case they are other entities
-        $id = array_map(fn ($id) => \is_object($id) && $this->supports($id::class) ? $this->dehydrate($id) : $id, $id);
+        $id = array_map(fn($id) => \is_object($id) && $this->supports($id::class) ? $this->dehydrate($id) : $id, $id);
 
         switch (\count($id)) {
             case 0:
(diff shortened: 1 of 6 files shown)


src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,15 +81,21 @@ public function dehydrate(object $object): mixed
 
     private function objectManagerFor(string $class): ?ObjectManager
     {
-        if (!class_exists($class)) {
+        if (!interface_exist($class) && !class_exists($class)) {
```
Kocal: `interface_exist` does not exist, but `interface_exists` does :D 

However, I bet 99% of the time, users will use a class for a property type and not an interface. 

Meaning we should call `interface_exists` at last, to prevent possible useless calls.

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,15 +81,21 @@ public function dehydrate(object $object): mixed
...
-                return self::ensureManagedObject($om, $class);
+             foreach($registry->getManagers() as $om) {
+                // this way, it resolves the interface
+                if ($om->getClassMetadata($class)) {
+                    return self::ensureManagedObject($om, $class);
+                }
             }
+            // if ($om = $registry->getManagerForClass($class)) {
+            //    return self::ensureManagedObject($om, $class);
+            // }
```
Kocal: This should either be removed or kept

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,15 +81,21 @@ public function dehydrate(object $object): mixed
...
         // todo cache/warmup an array of classes that are "doctrine objects"
         foreach ($this->managerRegistries as $registry) {
-            if ($om = $registry->getManagerForClass($class)) {
-                return self::ensureManagedObject($om, $class);
+             foreach($registry->getManagers() as $om) {
+                // this way, it resolves the interface
+                if ($om->getClassMetadata($class)) {
+                    return self::ensureManagedObject($om, $class);
+                }
             }
```
Kocal: I'm not very sure about the logic, could you add some tests to ensure interfaces are supported and classes still supported? Thanks :)

Kocal (changes requested):
Thanks for your first contribution :)

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
pink6440: You're right, digging into the doctrine ClassMetadataFactory, i realize that it will throw an exception if multiple em ;
I need to catch the MappingException to make it safe ;
I'm will continue digging and watch for produce some tests

pink6440:
The last commit provides some tests.
I put some comments in the code to explain the logic.

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
pink6440: I agree, i will fix it.

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,14 +81,44 @@ public function dehydrate(object $object): mixed
...
-            if ($om = $registry->getManagerForClass($class)) {
-                return self::ensureManagedObject($om, $class);
+        if (interface_exists($class)) {
+            // special handler for interfaces
+            // For use cases : @see https://symfony.com/doc/current/doctrine/resolve_target_entity.html
+            // As today, getManagerForClass don't resolve nicely aliased interfaces
+            // The workaround is to enum over each object manager and trying to get metadata
+            // The metadata are indeed, resolved nicely
+            // Fore more details :
+            // @see \Doctrine\ORM\Tools\ResolveTargetEntityListener
```
smnandre, line 101: Why don't we use the ResolveTargetentityListener  ?

pink6440: Because at that time, i had not digged enough about how it was working under the hood :)

So, you're right, i will refactore it ;

Also, discussing with @Kocal, there was also the possibility to write another specific hydrator ; but it will imply to duplicate the most technical code of the current Doctrine Hydrator so i'm more in favor of updating the current one.

Do you have an opinion about this ?

pink6440: After investigation, the ResolveTargetEntilyListener does not provide something like "resolveRealClassname".

The service is listening for a ClassMetadataNotFound event and eventually respond to it.
To use it, we will have to dispatch a such event for each ObjectManager.

I'm not sure that will made the code simpliest but it will avoid handling exception when calling getClassMetadata.

Any opinion on this ?

smnandre: I'm not sure we should have specific code when it's not our "core" feature in UX component, and there is a bit "too many of this" in LiveComponent for my taste 😅 

What I would love is _**if**_ we could leverage or reuse existing code either in Doctrine, in Serializer, ...

But If we cannot, yep i'd agree with you: updating the current one feels much better. 

Also I suggest we do not implement the cache at the same time because, as we talked once or twice before, as soon with hit 3.0 we will start working "from scrartch" on a brand new metadata/factory system, and the whole hydrator process here is probably the first part of LiveComponent that would use it :)

src/LiveComponent/tests/Fixtures/Kernel.php
```diff
@@ -170,6 +172,8 @@ protected function configureContainer(ContainerConfigurator $c): void
                         'alias' => 'XML',
                     ],
                 ],
+                'resolve_target_entities' => [
+                    AliasedEntityInterface::class => AliasedEntity::class],
```
smnandre:
```suggestion
                'resolve_target_entities' => [
                    AliasedEntityInterface::class => AliasedEntity::class,
                ],
```

src/LiveComponent/tests/Integration/Hydration/DoctrineEntityHydrationExtensionTest.php
```diff
@@ -50,4 +53,34 @@ public function testForeignKeyId(): void
         self::assertSame($foreignKeyIdEntity->id->id, $dehydrated);
         self::assertSame($foreignKeyIdEntity, $extension->hydrate($dehydrated, ForeignKeyIdEntity::class));
     }
+
+    public function testSupportInterface(): void
```
smnandre:
```suggestion
    public function testSupportInterface()
```

src/LiveComponent/tests/Integration/Hydration/DoctrineEntityHydrationExtensionTest.php
```diff
@@ -50,4 +53,34 @@ public function testForeignKeyId(): void
...
+    {
+        /** @var DoctrineEntityHydrationExtension $extension */
+        $extension = self::getContainer()->get('ux.live_component.doctrine_entity_hydration_extension');
+
+        self::assertTrue($extension->supports(AliasedEntityInterface::class), 'AliasedEntityInterface should be supported');
+        self::assertTrue($extension->supports(AliasedEntity::class), 'AliasedEntity should be supported');
+        self::assertFalse($extension->supports('UnknownClass'), 'UnknownClass should not be supported');
+    }
+
+    public function testHydrationFromInterface(): void
```
smnandre:
```suggestion
    public function testHydrationFromInterface()
```

src/LiveComponent/tests/Fixtures/Entity/AliasedEntityInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Entity;
+
+interface AliasedEntityInterface {
+
```
smnandre:
```suggestion
```

src/LiveComponent/tests/Fixtures/Dto/Aliased.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Dto;
+
+class Aliased
+{
+//    public string $address;
```
smnandre:
```suggestion
```

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,14 +81,44 @@ public function dehydrate(object $object): mixed
 
     private function objectManagerFor(string $class): ?ObjectManager
     {
-        if (!class_exists($class)) {
+        if (class_exists($class)) {
+            // Keep the standard way for a class
+            // todo cache/warmup an array of classes that are "doctrine objects"
```
smnandre:
```suggestion
```

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,14 +81,44 @@ public function dehydrate(object $object): mixed
...
+            // The workaround is to enum over each object manager and trying to get metadata
+            // The metadata are indeed, resolved nicely
+            // Fore more details :
+            // @see \Doctrine\ORM\Tools\ResolveTargetEntityListener
+
+            // todo cache/warmup an array of interfaces that are resolved "doctrine objects"
+            foreach ($this->managerRegistries as $registry) {
+                foreach ($registry->getManagers() as $om) {
+                    // the getClassMetaData can indeed throw an exception
+                    // so, we
```
smnandre: ?

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,14 +81,44 @@ public function dehydrate(object $object): mixed
...
+                    try {
+                        if (null !== $om->getClassMetadata($class)) {
+                            return self::ensureManagedObject($om, $class);
+                        }
+                    } catch (MappingException $e) {
+                        // I did not find a nice way to check if it is because the class is really unknown
+                        // It is good to check for a specific exception ?
+                        // eg: \Doctrine\Persistence\Mapping\MappingException
+                        // @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getMetadataFor
+                        // throw $e;
```
smnandre: We 100% can catch a specific exception if that allows another object manager to be selected during following iterations of the loop(s).

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,14 +81,44 @@ public function dehydrate(object $object): mixed
...
+        if (interface_exists($class)) {
+            // special handler for interfaces
+            // For use cases : @see https://symfony.com/doc/current/doctrine/resolve_target_entity.html
+            // As today, getManagerForClass don't resolve nicely aliased interfaces
+            // The workaround is to enum over each object manager and trying to get metadata
+            // The metadata are indeed, resolved nicely
+            // Fore more details :
+            // @see \Doctrine\ORM\Tools\ResolveTargetEntityListener
+
+            // todo cache/warmup an array of interfaces that are resolved "doctrine objects"
```
smnandre:
```suggestion
```

src/LiveComponent/src/Hydration/DoctrineEntityHydrationExtension.php
```diff
@@ -81,14 +81,44 @@ public function dehydrate(object $object): mixed
...
-        foreach ($this->managerRegistries as $registry) {
-            if ($om = $registry->getManagerForClass($class)) {
-                return self::ensureManagedObject($om, $class);
+        if (interface_exists($class)) {
+            // special handler for interfaces
+            // For use cases : @see https://symfony.com/doc/current/doctrine/resolve_target_entity.html
+            // As today, getManagerForClass don't resolve nicely aliased interfaces
+            // The workaround is to enum over each object manager and trying to get metadata
+            // The metadata are indeed, resolved nicely
+            // Fore more details :
```
smnandre, line 100: Can we turn this into two or three lines, insisting more on what we "do" than on what "was" the problem ?

pink6440 (commented):
Thanks for the review !

Kocal:

Thanks!

Status: Needs Work

PR #2836, opened 2025-06-12
[Site] Trim Icon's DataList's `<dd>` to get a better copy value

### LINKED ISSUE #2831: [Icons] Trim Metadata's field to get a better copy value
Hello everyone, 

On https://ux.symfony.com/ when I select an icon I have the following modal that appears:

Image

If I want to copy its `Name` value I double click on it so I can copy it:

Image

But as you can see, there is some space within the selection, and the value that is copied is the following: `     text-align-left ` with some spaces before and after the value.

It's coming from the way the HTML is generated as there is some spaces around the wording:

Image

So I was wandering if there is any chance that you can possibly trim the value:

Image

So when I double click on the value and copy it, I'll have only the value, and not spaces around it.

Thank you and have a nice day!

CyrilKrylatov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2831 
| License       | MIT

On Icon's modal, there is a space at the end of DD's HTML element:

image

So when double-click on it and copy it, we'll copy this space as well. 

This PR fixes that space.

Before:

image

After:

image

### DIFF
--- a/ux.symfony.com/templates/components/DataList.html.twig
+++ b/ux.symfony.com/templates/components/DataList.html.twig
@@ -8,9 +8,9 @@
                 <dd>
                     {%- if value.href|default %}
                         <a href="{{ value.href }}">{{ value.title }}</a>
-                    {% else %}
+                    {% else -%}
                         {{ value.title ?? value }}
-                    {% endif -%}
+                    {%- endif -%}
                 </dd>
             </div>
         {% endif %}


Kocal (approved):

Kocal:
Thank you @CyrilKrylatov.

PR #2843, opened 2025-06-16
[Doc] Update the docs about the mount() method of Twig components

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | -
| License       | MIT

Today I had some issues understanding how the `mount()` method works, so I propose these changes to clarify things.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -447,15 +447,24 @@ need to populate, you can render it with:
 Mounting Data
 -------------
 
-Most of the time, you will create public properties and then pass values
-to those as "props" when rendering. But there are several hooks in case
-you need to do something more complex.
+Most of the time, you will create public properties and pass values to them
+as "props" when rendering the component. However, there are several hooks
+available when you need to perform more complex logic.
 
 The mount() Method
 ~~~~~~~~~~~~~~~~~~
 
-For more control over how your "props" are handled, you can create a method
-called ``mount()``::
+The ``mount()`` method gives you more control over how your "props" are handled.
+It is called once, immediately after the component is instantiated, **but before**
+the component system assigns the props you passed when rendering.
+
+For example, if you call your component like this:
+
+.. code-block:: html+twig
+
+    <twig:Alert type="error" message="..."/>
+
+The following code won't work as expected::
 
     // src/Twig/Components/Alert.php
     // ...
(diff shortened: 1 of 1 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -447,15 +447,25 @@ need to populate, you can render it with:
 Mounting Data
 -------------
 
-Most of the time, you will create public properties and then pass values
-to those as "props" when rendering. But there are several hooks in case
-you need to do something more complex.
+Most of the time, you will create public properties and pass values to them
+as "props" when rendering the component. However, there are several hooks
+available when you need to perform more complex logic.
```
Kocal, line 452: 👍🏻

src/TwigComponent/doc/index.rst
```diff
@@ -447,15 +447,25 @@ need to populate, you can render it with:
...
 
 The mount() Method
 ~~~~~~~~~~~~~~~~~~
 
-For more control over how your "props" are handled, you can create a method
-called ``mount()``::
+The ``mount()`` method gives you more control over how your "props" are handled.
+It is called only once: immediately after your component is instantiated.
```
Kocal: 👍🏻

src/TwigComponent/doc/index.rst
```diff
@@ -447,15 +447,25 @@ need to populate, you can render it with:
...
+The ``mount()`` method **cannot access the values of the public properties**
+passed when rendering the component.
```
Kocal: I see what you mean, but this is _wrong_ if I want to be picky:
1. you can access the public properties values, if they are not typed or have a default values
2. speaking about only "public properties values" can be misleading, I feel like we should speak about the "props" too in priority

src/TwigComponent/doc/index.rst
```diff
@@ -466,29 +476,46 @@ called ``mount()``::
...
+    class Alert
+    {
+        public string $message;
+        public string $type = 'success';
+
+        public function mount(bool $isError = false): void
+        {
+            {# ✅ this works as expected #}
+            if ($isError) {
+                // ...
```
Kocal: Like it was done before, we should keep an example (or document) that we can modify a property here
```suggestion
                $this->type = 'danger';
```

src/TwigComponent/doc/index.rst
```diff
@@ -466,29 +476,46 @@ called ``mount()``::
...
+
+Now, pass the ``mount()`` argument like any other prop:
 
 .. code-block:: html+twig
 
-    <twig:Alert
-        isSuccess="{{ false }}"
-        message="Danger Will Robinson!"
-    />
+    <twig:Alert isError="{{ true }}" message="..."/>
```
Kocal: Can we move it just under `Instead, define the values you need in the ``mount()`` method as arguments::`? I feel like the actual sections order can be misleading

src/TwigComponent/doc/index.rst
```diff
@@ -466,29 +476,46 @@ called ``mount()``::
...
         }
 
         // ...
     }
 
-The ``mount()`` method is called just one time: immediately after your
-component is instantiated. Because the method has an ``$isSuccess``
-argument, if we pass an ``isSuccess`` prop when rendering, it will be
-passed to ``mount()``.
+Instead, define the values you need in the ``mount()`` method as arguments::
```
Kocal: The parameters you define in `mount()` does not only refer to your Component's public properties, but can be any props. For example, if I use `<twig:Alert toto="tata">`, then I can retrieve its value with `public function mount(string $toto)`:

```suggestion
Instead, the props values are available in the ``mount()`` method as arguments::
```

javiereguiluz:
Thanks for the review Hugo! I reworded some parts. Hopefully it's better now.

Kocal (approved):

Kocal:
Thanks for this nice improvment, Javier :)

PR #2866, opened 2025-06-25
Migrate `bin/*.js` to TypeScript, use `tsx`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Types ✨ 

Following #2865, in fact I didn't go with Node & `--experimental-strip-types` flag, but with https://tsx.is/ instead.


package.json
```diff
@@ -1,6 +1,7 @@
 {
     "private": true,
     "packageManager": "yarn@4.5.0",
+    "type": "module",
```
Kocal, line 4: It should not impact `src/**/assets` packages

smnandre:
Trusting you here :)

Kocal:
I noticed using `tsx` for building UX packages slowed down the pipeline by ~17 seconds, so I'm not really sure about this PR... but maybe it can be "compensated" by replacing Rollup with Rolldown in another PR (I still have something to see with React and `react-dom/client` dependency thing)

I may try to give to another chance to `node --experimental-strip-types`, or maybe use Node.js 24 for internal tools instead of 22.6. 🤷🏻

PR #2879, opened 2025-06-29
Remove `@symfony/stimulus-testing` dev dependencies

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

I plan to deprecate `@symfony/stimulus-testing`, the library is too much opinionated by depending on Jest and Babel.

From 837 to 442 (dev)dependencies :tada:


smnandre:
So we only used two DOM-related methods from the `@stimulus/testing` ? 

(the ones you moved in the ../../../helper file)

Are people using it to test their controllers ?

Kocal:
> So we only used two DOM-related methods from the `@stimulus/testing` ?
> 
> (the ones you moved in the ../../../helper file)

Yes, and the `import '@testing-library/jest-dom';` from `test/setup.js` files

> Are people using it to test their controllers ?

It always had few downloads per week ([see stats](https://npmtrends.com/@symfony/stimulus-testing)), but since mid-May it started to grows to ~16k downloads per week. 
But I don't know if it's due to my changes on Symfony UX CI, or if someone else started to use the package.

smnandre:
So maybe we can stop using it ourselves... and see then, depending on the activity, if we deprecate it ?

Kocal:
Let's debate on https://github.com/symfony/stimulus-testing if you want, but for me there is no point to this library, either for maintainers or users.

smnandre:
You cannot say « I suggest we deprecate » then say « Lets not dizcyss it » haha 

IF (and that is a big if) there is a growing usage of the lib, it deserves maybe to suggest them an alternative, and probably some documentation updates, that’s all i’m saying

PR #2884, opened 2025-06-29
[Icons] don't disable commands based on debug

### LINKED ISSUE #2874: [Icons] Command "ux:icons:lock" is not defined
Since upgrading from `symfony/ux-icons` from 2.24 to >=2.25 our CI is failing with  the following error:

```
Command "ux:icons:lock" is not defined
```

This works locally in the `dev` environment, but fails in our CI build because it builds a Docker image which runs some preparatory commands in the `prod` env.

Our use case is that our CI builds the production Docker container, and at this point any icons that can be found are imported into the container. This saves us having all the icons stored in our source repository.

I see there [was a commit](https://github.com/symfony/ux-icons/blame/b1df2c1ae7b5ea9ced184d1dcc71485bfc01069e/src/DependencyInjection/UXIconsExtension.php#L183) which added this command to a condition that removes the service if `kernel.debug` is false. 

@kbond  What was the rationale behind excluding this command from the prod environment?

Is my use-case a bad practice? Thanks!

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2874
| License       | MIT

### DIFF
--- a/src/Icons/src/DependencyInjection/UXIconsExtension.php
+++ b/src/Icons/src/DependencyInjection/UXIconsExtension.php
@@ -176,11 +176,5 @@ final class UXIconsExtension extends ConfigurableExtension implements Configurat
         if (!$mergedConfig['iconify']['on_demand'] || !$mergedConfig['iconify']['enabled']) {
             $container->removeDefinition('.ux_icons.iconify_on_demand_registry');
         }
-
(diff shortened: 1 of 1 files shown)


Kocal (approved):

smnandre:
🤞 this will not have side effects on others 😅

PR #2882, opened 2025-06-29
[CI] Cache Composer dependencies

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Following https://github.com/symfony/ux/pull/2780#issuecomment-2918360074.
Also, sometimes the CI fails because Composer is not able to download files (e.g.: https://github.com/symfony/ux/actions/runs/15955763737/job/45001822817?pr=2880):
```
Chartjs                                                                       8s
+ cd src/Chartjs
+ composer config minimum-stability stable --ansi
+ composer update --no-progress --no-interaction --ansi
Loading composer repositories with package information
Restricting packages listed in "symfony/symfony" to ">=5.4"
Updating dependencies
Lock file operations: 30 installs, 0 updates, 0 removals
  - Locking psr/cache (3.0.0)
  - Locking psr/container (2.0.2)
  - Locking psr/event-dispatcher (1.0.0)
  - Locking psr/log (3.0.2)
  - Locking symfony/cache (v7.3.1)
  - Locking symfony/cache-contracts (v3.6.0)
  - Locking symfony/config (v7.3.0)
  - Locking symfony/dependency-injection (v7.3.1)
  - Locking symfony/deprecation-contracts (v3.6.0)
  - Locking symfony/error-handler (v7.3.1)
  - Locking symfony/event-dispatcher (v7.3.0)
  - Locking symfony/event-dispatcher-contracts (v3.6.0)
  - Locking symfony/filesystem (v7.3.0)
  - Locking symfony/finder (v7.3.0)
  - Locking symfony/framework-bundle (v7.3.1)
  - Locking symfony/http-foundation (v7.3.1)
  - Locking symfony/http-kernel (v7.3.1)
  - Locking symfony/phpunit-bridge (v7.3.1)
  - Locking symfony/polyfill-ctype (v1.32.0)
  - Locking symfony/polyfill-mbstring (v1.32.0)
  - Locking symfony/polyfill-php83 (v1.32.0)
  - Locking symfony/routing (v7.3.0)
  - Locking symfony/service-contracts (v3.6.0)
  - Locking symfony/stimulus-bundle (dev-d0e37bc4d9669c4d60476e89be536b695a5eea7c)
  - Locking symfony/translation-contracts (v3.6.0)
  - Locking symfony/twig-bridge (v7.3.0)
  - Locking symfony/twig-bundle (v7.3.1)
[...]

PR #2885, opened 2025-06-29
[CI] (Re-)use ramsey/composer-install action when possible

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Replace #2882

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -34,10 +34,16 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             -   uses: actions/checkout@v4
+
             -   uses: shivammathur/setup-php@v2
                 with:
                     php-version: '8.1'
-            -   run: composer install
+
+            - name: Install root dependencies
+              uses: ramsey/composer-install@v3
+              with:
+                  working-directory: ${{ github.workspace }}
+
             -   name: php-cs-fixer
                 run: ./vendor/bin/php-cs-fixer fix --dry-run --diff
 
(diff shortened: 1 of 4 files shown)


.github/workflows/code-quality.yaml
```diff
@@ -83,8 +89,23 @@ jobs:
...
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-composer-packages-${{ matrix.php-version }}-${{ matrix.dependency-version }}-${{ matrix.symfony-version }}-${{ matrix.minimum-stability }}-${{ hashFiles('src/**/composer.json') }}
+                  restore-keys: |
+                      ${{ runner.os }}-composer-packages-${{ matrix.php-version }}-${{ matrix.dependency-version }}-${{ matrix.symfony-version }}-${{ matrix.minimum-stability }}
+
             - name: Install root dependencies
-              run: composer install
+              uses: ramsey/composer-install@v3
+              with:
+                  working-directory: ${{ github.workspace }}
```
Kocal, line 108: Manually caching Composer dependencies is still necessary because dependencies are parallelly installed through `composer`, which AFAIK is not possible with `ramsey/composer-install`

.github/workflows/unit-tests.yaml
```diff
@@ -68,8 +68,23 @@ jobs:
...
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-composer-packages-${{ matrix.php-version }}-${{ matrix.dependency-version }}-${{ matrix.symfony-version }}-${{ matrix.minimum-stability }}-${{ hashFiles('src/**/composer.json') }}
+                  restore-keys: |
+                      ${{ runner.os }}-composer-packages-${{ matrix.php-version }}-${{ matrix.dependency-version }}-${{ matrix.symfony-version }}-${{ matrix.minimum-stability }}
+
             - name: Install root dependencies
-              run: composer install
+              uses: ramsey/composer-install@v3
+              with:
+                  working-directory: ${{ github.workspace }}
```
Kocal, line 87: Manually caching Composer dependencies is still necessary because dependencies are parallelly installed through `composer`, which AFAIK is not possible with `ramsey/composer-install`

smnandre: Genuine questiin, do we still need ramsey then here ?

smnandre:
I'm a bit lost and do not understand why we have soo many test jobs... but i trust you here so: 👍

.github/workflows/unit-tests.yaml
Kocal: That's for root dependencies, the one defined here: https://github.com/symfony/ux/blob/2.x/composer.json

Using `ramsey/composer-install` here makes sense, as we install root dependencies across 6 CI workflows/jobs, which share the same cache keys.

Kocal:
> I'm a bit lost and do not understand why we have soo many test jobs... but i trust you here so: 👍

The CI has been modified during the last few months, we have:
- unit tests (for PHP and JavaScript) which require only basic setup
- functional tests (only Turbo) which require more advanced setup (Mercure)
- and "app tests" that ensures UX packages (PHP and JS) can be correctly installed/used in real Symfony apps

kbond (approved):

PR #2816, opened 2025-06-04
[CI] Run unit and functional tests under Windows too

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

I think this is a must have, especially for Toolkit which deals a lot with paths

PR #2890, opened 2025-07-01
[Vue][React][Svelte] Fix components assets compilation on Windows server

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Following #2816, this PR fixes the `controller.js` files generation when the Symfony app is running on Windows server:
```
Testing 
F...                                                                4 / 4 (100%)

Time: 00:01.229, Memory: 30.00 MB

There was 1 failure:

1) Symfony\UX\Vue\Tests\AssetMapper\VueControllerLoaderAssetCompilerTest::testCompileDynamicallyAddsContents
Failed asserting that 'import component_0 from '../../vue/controllers/MyVueController.js';\n
import component_1 from '../../vue/controllers/subdir/DeeperVueController.js';\n
export const components = {"MyVueController": component_0, "subdir\DeeperVueController": component_1};' contains ""subdir/DeeperVueController": component_".

D:\a\ux\ux\src\Vue\tests\AssetMapper\VueControllerLoaderAssetCompilerTest.php:72
```

I was able to set-up a Windows VM to ensure the PR fixes the problem (tip: don't use a remote VM in a train... 🙂) 

Before, you can see that `subfolder\Goodbye` is interpreted as `subfolderGoodbye`, meaning that is not possible to use `subfolder/Goodbye` Vue component (I made a typo in the Twig template, but it has no effect):
Capture d’écran 2025-07-02 à 09 39 00
With the patch, the path now contains `/` instead of `\`, the component `subfolder/Goodbye` is now usable:
Capture d’écran 2025-07-02 à 09 58 07

### DIFF
--- a/src/React/src/AssetMapper/ReactControllerLoaderAssetCompiler.php
+++ b/src/React/src/AssetMapper/ReactControllerLoaderAssetCompiler.php
@@ -57,7 +57,7 @@ class ReactControllerLoaderAssetCompiler implements AssetCompilerInterface
                 $controllerNameForVariable,
                 $relativeImportPath
             );
-            $componentParts[] = \sprintf('"%s": %s', $name, $controllerNameForVariable);
+            $componentParts[] = \sprintf('"%s": %s', Path::normalize($name), $controllerNameForVariable);
         }
 
         $importCode = implode("\n", $importLines);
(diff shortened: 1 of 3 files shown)


Kocal:
I only tested Vue components btw, didn't have time for React and Svelte, the logic is the same

PR #2898, opened 2025-07-08
[LiveComponent] Add `keep-alive` modifier to polling plugin

### LINKED ISSUE #995: [LiveComponent] `in-view` and `tab-active` modifiers for polling
From https://github.com/symfony/ux/issues/102:

> Future D) Polling enhancements: (1) poll less often when the browser tab is not active and (2) don't start polling until visible

```twig
<div
    {{ attributes }}
    data-poll="delay(500)|in-view|tab-active|$render" {# only poll when component is in viewport and browser tab is active #}
>
```

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #995
| License       | MIT

This is a naive implementation, which I think needs to be discussed.

I think the listener somehow somewhere should be detached when poll disconnects.

And the best solution would be to add it once for all polling, but I don't see how it is doable.

### DIFF
--- a/src/LiveComponent/assets/src/Component/plugins/PollingPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/PollingPlugin.ts
@@ -53,6 +53,16 @@ export default class implements PluginInterface {
                         }
 
                         break;
+                    case 'keep-alive':
+                        document.addEventListener('visibilitychange', () => {
+                            if (document.hidden) {
+                                this.pollingDirector.stopAllPolling();
+                            } else {
+                                this.pollingDirector.startAllPolling();
+                            }
+                        });
+
+                        break;
                     default:
                         console.warn(`Unknown modifier "${modifier.name}" in data-poll "${rawPollConfig}".`);
                 }
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.28.0
+
+-   Add `keep-alive` modifier to polling plugin
+
 ## 2.27.0
 
 -  Add events assertions in `InteractsWithLiveComponents`:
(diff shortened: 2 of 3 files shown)


norkunas:
Oh, livewire keep-alive is a reverse, as @smnandre said in the issue.. How we should name it?

smnandre:
I'm not sure this should be a modifier... so maybe name does not matter :)

norkunas:
> I'm not sure this should be a modifier... so maybe name does not matter :)

And if not a modifier, then what it should be?

src/LiveComponent/assets/src/Component/plugins/PollingPlugin.ts
```diff
@@ -52,6 +52,16 @@ export default class implements PluginInterface {
...
 
+                        break;
+                    case 'keep-alive':
+                        document.addEventListener('visibilitychange', () => {
+                            if (document.hidden) {
+                                this.pollingDirector.stopAllPolling();
+                            } else {
+                                this.pollingDirector.startAllPolling();
+                            }
+                        });
```
Kocal, line 63: We should remove this event listener on controller's disconnect

norkunas: i know, first let's agree on direction :) as @smnandre disagree about modifier

smnandre:
It should be the default behaviour. Not sure i see a valid usage for polls while the tab is in background

Kocal:
Could it be considered as a BC?
IMHO we should make it configurable in 2.x (default to `false`) with a deprecation message telling that will be unconfigurable and default to `true` in 3.x. WDYT?

smnandre:
To me it's almost a bug if we send requests at fixed interval when the tab is in background or hidden :| 

--> https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

Problem will happen on every browser (except if we speak about big interval but then.. what is the point? what feature is provided by this behaviour ?)

norkunas:
Maybe it's a game where something should still happen in background :D but it's ok for me doing this without modifier

Kocal:

Thanks!

Status: Needs Work

PR #2904, opened 2025-07-11
[Icons] Add support for <title> and <desc> elements in SVG for accessibility

xDeSwa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

When passing title and/or desc as attributes to ux_icon(), they are now embedded directly into the SVG as <title> and <desc> child elements.

Each <title> and <desc> element is given a unique id and automatically referenced via aria-labelledby, if no such attribute was already set.

This ensures compliance with accessibility best practices for inline SVG content.
See:

    https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

    https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc

Example Usage:

```
{{ ux_icon('bi:plus-square-dotted', {
    width: '16px',
    height: '16px',
    class: 'text-success',
    title: 'Add Stock',
    desc: 'This icon indicates stock entry functionality.'
}) }}
```

Renders:

```
<svg class="text-success" width="16px" height="16px" aria-labelledby="icon-title-abc icon-desc-def">
    <title id="icon-title-abc">Add Stock</title>
    <desc id="icon-desc-def">This icon indicates stock entry functionality.</desc>
    <!-- inner SVG content -->
</svg>
```

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -481,13 +481,13 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
 
         <twig:ux:icon name="user-profile" aria-hidden="false" />
 
-## Accessibility: Descriptive Titles and Descriptions
+**Accessibility: Descriptive Titles and Descriptions**
 
-.. versionadded:: NEXT\_VERSION
+.. versionadded:: 2.28
 
-The `ux_icon()` function and the `<twig:ux:icon>` component now support accessible SVG metadata via the `title` and `desc` attributes.
+The `ux_icon()` function and the `<twig:ux:icon>` component now support accessible SVG metadata via the `title` and `desc` attributes in 2.28.
 
-These are automatically injected into the `<svg>` markup as child elements, and properly referenced using `aria-labelledby` for improved screen reader support.
+These are automatically injected into the ``<svg>`` markup as child elements, and properly referenced using ``aria-labelledby`` for improved screen reader support.
 
 **How it works:**
 
@@ -495,7 +495,6 @@ When you pass a `title` and/or `desc` attribute, they are rendered inside the `<
 
 .. code-block:: html+twig
 
-```
 {{ ux_icon('bi:plus-square-dotted', {
     width: '16px',
     height: '16px',
@@ -503,19 +502,24 @@ When you pass a `title` and/or `desc` attribute, they are rendered inside the `<
     title: 'Add Stock',
     desc: 'This icon indicates stock entry functionality.'
 }) }}
-```
 
 Renders:
 
 .. code-block:: html
 
-```
 <svg class="text-success" width="16px" height="16px" aria-labelledby="icon-title-abc icon-desc-def">
     <title id="icon-title-abc">Add Stock</title>
     <desc id="icon-desc-def">This icon indicates stock entry functionality.</desc>
     <!-- inner SVG content -->
 </svg>
-```
+
+.. note::
+
+    - If ``aria-labelledby`` is already defined in your attributes, it will **not** be overwritten.
+    - ``role="img"`` is **not added automatically**. You may choose to include it if your use case requires.
+    - When neither ``title``, ``desc``, ``aria-label``, nor ``aria-labelledby`` are provided, ``aria-hidden="true"`` will still be automatically applied.
+
+This feature brings UX Icons in line with modern accessibility recommendations and helps developers build more inclusive user interfaces.
 
 To learn more about accessible SVG elements:
 
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -493,7 +493,7 @@ These are automatically injected into the ``<svg>`` markup as child elements, an
 
 When you pass a `title` and/or `desc` attribute, they are rendered inside the `<svg>` as follows:
 
-.. code-block:: html+twig
+.. code-block:: twig
 
 {{ ux_icon('bi:plus-square-dotted', {
     width: '16px',
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -481,6 +481,47 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
 
         <twig:ux:icon name="user-profile" aria-hidden="false" />
 
+## Accessibility: Descriptive Titles and Descriptions
+
+.. versionadded:: NEXT\_VERSION
+
+The `ux_icon()` function and the `<twig:ux:icon>` component now support accessible SVG metadata via the `title` and `desc` attributes.
+
+These are automatically injected into the `<svg>` markup as child elements, and properly referenced using `aria-labelledby` for improved screen reader support.
+
+**How it works:**
+
+When you pass a `title` and/or `desc` attribute, they are rendered inside the `<svg>` as follows:
+
+.. code-block:: html+twig
+
+```
+{{ ux_icon('bi:plus-square-dotted', {
+    width: '16px',
+    height: '16px',
+    class: 'text-success',
+    title: 'Add Stock',
+    desc: 'This icon indicates stock entry functionality.'
+}) }}
+```
+
+Renders:
+
+.. code-block:: html
+
+```
+<svg class="text-success" width="16px" height="16px" aria-labelledby="icon-title-abc icon-desc-def">
+    <title id="icon-title-abc">Add Stock</title>
+    <desc id="icon-desc-def">This icon indicates stock entry functionality.</desc>
+    <!-- inner SVG content -->
+</svg>
+```
+
+To learn more about accessible SVG elements:
+
+- `MDN: <title>`\_ — [https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title)
+- `MDN: <desc>`\_ — [https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc)
+
 Performance
 -----------
 
--- a/src/Icons/src/Icon.php
+++ b/src/Icons/src/Icon.php
@@ -139,27 +139,53 @@ final class Icon implements \Stringable
     public function toHtml(): string
     {
         $htmlAttributes = '';
-        foreach ($this->attributes as $name => $value) {
+        $innerSvg = $this->innerSvg;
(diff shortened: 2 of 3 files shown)


src/Icons/doc/index.rst
```diff
@@ -481,6 +481,47 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
 
         <twig:ux:icon name="user-profile" aria-hidden="false" />
 
+## Accessibility: Descriptive Titles and Descriptions
+
+.. versionadded:: NEXT\_VERSION
```
Kocal:
```suggestion
.. versionadded:: 2.28
```

src/Icons/src/Icon.php
```diff
@@ -139,27 +139,53 @@ public function __construct(
...
             if (true === $value) {
                 continue;
             }
 
-            $value = htmlspecialchars($value, \ENT_QUOTES | \ENT_SUBSTITUTE, 'UTF-8');
+            $value = htmlspecialchars((string) $value, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
             $htmlAttributes .= '="'.$value.'"';
-        }
-
-        return '<svg'.$htmlAttributes.'>'.$this->innerSvg.'</svg>';
```
Kocal, line 162: It should be added back

src/Icons/src/Icon.php
```diff
@@ -139,27 +139,53 @@ public function __construct(
...
+        }
+
+        // Build final attributes string
+        foreach ($attributes as $name => $value) {
             if (false === $value) {
                 continue;
             }
 
-            // Special case for aria-* attributes
-            // https://www.w3.org/TR/wai-aria-1.1/#state_prop_def
```
Kocal: It should be added back

src/Icons/src/Icon.php
```diff
@@ -139,27 +139,53 @@ public function __construct(
...
+        $attributes = $this->attributes;
+
+        // Extract and remove title/desc attributes if present
+        $title = $attributes['title'] ?? null;
+        $desc = $attributes['desc'] ?? null;
+        unset($attributes['title'], $attributes['desc']);
+
+        // Prepare <title> and <desc> elements
+        $labelledByIds = [];
+        $a11yContent = '';
```
Kocal, line 151: `$a11yContent` is not used for rendering

src/Icons/doc/index.rst
Kocal: There is no paragraph under `.. versionadded`, it must looks like:

```
.. versionadded: 2.28

    Support for (...) in 2.28.
```

src/Icons/src/Icon.php
```diff
@@ -139,27 +139,53 @@ public function __construct(
...
+
+        if ($title) {
+            $titleId = 'title-' . bin2hex(random_bytes(4));
+            $labelledByIds[] = $titleId;
+            $a11yContent .= sprintf('<title id="%s">%s</title>', $titleId, htmlspecialchars((string) $title, ENT_QUOTES));
+        }
+
+        if ($desc) {
+            $descId = 'desc-' . bin2hex(random_bytes(4));
```
Kocal: I feel like it's a not a good idea to generate random strings for eacg title/desc, especially when you render a lot of icons in a single page.

src/Icons/doc/index.rst
```diff
@@ -481,6 +481,47 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
 
         <twig:ux:icon name="user-profile" aria-hidden="false" />
 
+## Accessibility: Descriptive Titles and Descriptions
+
```
Kocal: Please use underline chars for headings, this is not in markdown but RST format

src/Icons/doc/index.rst
```diff
@@ -481,6 +481,47 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
...
+.. versionadded:: NEXT\_VERSION
+
+The `ux_icon()` function and the `<twig:ux:icon>` component now support accessible SVG metadata via the `title` and `desc` attributes.
+
+These are automatically injected into the `<svg>` markup as child elements, and properly referenced using `aria-labelledby` for improved screen reader support.
```
Kocal: Same for all others code markups

```suggestion
These are automatically injected into the ``<svg>`` markup as child elements, and properly referenced using ``aria-labelledby`` for improved screen reader support.
```

src/Icons/doc/index.rst
```diff
@@ -481,6 +481,47 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
...
+
+**How it works:**
+
+When you pass a `title` and/or `desc` attribute, they are rendered inside the `<svg>` as follows:
+
+.. code-block:: html+twig
+
+```
```
Kocal: Remove the ```, they are not useful in RST format

Kocal (changes requested):
Thanks for working in this!

Please be sure to add tests

src/Icons/src/Icon.php
xDeSwa: "IDs for the <title> and <desc> elements are generated only when the aria-labelledby attribute is automatically added to reference them."

xDeSwa:
> Thanks for working in this!
> 
> Please be sure to add tests

Thanks for suggestions. can you check again please.

src/Icons/src/Icon.php
Kocal: Yeah I know, but I don't think generating random values like that is a good thing, it can be CPU intensive and it's untestable. 

Instead, either we generate these ids based on the icon (name, title, description, ...), either we use an internal incremental counter or something similar

src/Icons/doc/index.rst
```diff
@@ -481,6 +481,51 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
...
+    desc: 'This icon indicates stock entry functionality.'
+}) }}
+
+Renders:
+
+.. code-block:: html
+
+<svg class="text-success" width="16px" height="16px" aria-labelledby="icon-title-abc icon-desc-def">
+    <title id="icon-title-abc">Add Stock</title>
+    <desc id="icon-desc-def">This icon indicates stock entry functionality.</desc>
```
Kocal, line 512: The `id` attributes do no match the actual behaviour

src/Icons/src/Icon.php
```diff
@@ -139,27 +139,68 @@ public function __construct(
...
+    
             $value = htmlspecialchars($value, \ENT_QUOTES | \ENT_SUBSTITUTE, 'UTF-8');
-            $htmlAttributes .= '="'.$value.'"';
+            $htmlAttributes .= '="' . $value . '"';
         }
-
-        return '<svg'.$htmlAttributes.'>'.$this->innerSvg.'</svg>';
+    
+        // Inject <title> and <desc> before inner content
+        return '<svg' . $htmlAttributes . '>' . $a11yContent . $innerSvg . '</svg>';
```
Kocal: Please revert all changes here, except for `$a11yContent` usage, PR must contains only the necessary changes

Kocal (changes requested):
I'm sorry, but I don't understand the benefits over using `<title>` and `<desc>`, instead of using `aria-label` and `aria-description` attribute. Do you have more information about that?


Tests are still missing, please make sure to add tests for making the CI green, thanks

Kocal:
And apply Fabbot suggestions aswell, thanks

src/Icons/doc/index.rst
```diff
@@ -481,6 +481,51 @@ of the following attributes: ``aria-label``, ``aria-labelledby`` or ``title``.
...
+    <!-- inner SVG content -->
+</svg>
+
+.. note::
+
+    - If ``aria-labelledby`` is already defined in your attributes, it will **not** be overwritten.
+    - ``role="img"`` is **not added automatically**. You may choose to include it if your use case requires.
+    - When neither ``title``, ``desc``, ``aria-label``, nor ``aria-labelledby`` are provided, ``aria-hidden="true"`` will still be automatically applied.
+
+This feature brings UX Icons in line with modern accessibility recommendations and helps developers build more inclusive user interfaces.
```
smnandre, line 522: I'm not so sure about that..

> If an SVG is used in a context where it adds meaning to the content then it is not being used as an icon, and requires a different markup pattern
> ```svg
> <svg role="img" focusable="false">
>    <title>Accessible Name</title>
>     
> </svg>
> ```

https://design-system.w3.org/styles/svg-icons.html#non-decorative-svg-accessibility


I mean, "then it is not being used as an icon" is pretty much outside the scope of this bundle.

Please correct me if i'm outdated on this :)

smnandre: Also

> If an element can be described by visible text, it is recommended to reference that text with an [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby) attribute rather than using the <title> element.

So let's push for this recommendation no ?

src/Icons/src/Icon.php
```diff
@@ -139,7 +139,45 @@ public function __construct(
...
+
+        $labelledByIds = [];
+        $a11yContent = '';
+
+        // Check if aria-labelledby should be added automatically
+        $shouldSetLabelledBy = !isset($attributes['aria-labelledby']) && ($title || $desc);
+
+        if ($title) {
+            if ($shouldSetLabelledBy) {
+                $titleId = 'title-' . bin2hex(random_bytes(4));
```
smnandre, line 158: This will create unpredictable HTML

src/Icons/src/Icon.php
```diff
@@ -139,7 +139,45 @@ public function __construct(
...
+                $labelledByIds[] = $titleId;
+                $a11yContent .= sprintf('<title id="%s">%s</title>', $titleId, htmlspecialchars((string) $title, ENT_QUOTES));
+            } else {
+                $a11yContent .= sprintf('<title>%s</title>', htmlspecialchars((string) $title, ENT_QUOTES));
```
smnandre, line 162: This will add another <title> tag in the SVG if one is already present, generating invalid markup

smnandre (commented):
I'm not really sure of the use case..  Are `<title>` and `<desc>` really something usefull for inlined SVG icons ? It does not match what i read on the topic, but i'd be happy to learn new things here.

If someone uses UX Icons to display images maybe they can add `<title>` and/or `<desc>` directly in the SVG file ?

PR #2922, opened 2025-07-13
[LiveComponent] Fix BC break when using `PropertyTypeExtractorInterface::getType()` on a `#[LiveProp]` property `x` when getter `getX` exists

### LINKED ISSUE #2888: BC break in LiveComponentHydrator
There is a BC break in LiveComponentHydrator as it now throws an exception when dehydrating a User entity class extending UserInterface. 

Image

Image

In `\Symfony\UX\LiveComponent\LiveComponentHydrator::dehydrateObjectValue` the `$classType` parameter value is the interface not the class as it was before (in `LegacyLivePropMetadata`).

The component looks like this:

```php
<?php

declare(strict_types=1);

namespace App\Component;

use App\Controller\BaseController;
use App\Entity\User;
use App\Form\Type\UserAdminType;
use Symfony\Component\Form\FormInterface;
use Symfony\UX\LiveComponent\ComponentWithFormTrait;

#[AsLiveComponent('user_admin_form')]
final class UserAdminForm extends BaseController
{
    use DefaultActionTrait;
    use ComponentWithFormTrait;

    #[LiveProp]
    public User $user;

    {
        return $this->createForm(UserAdminType::class, $this->user, ['is_edit' => true]);
    }
}

```

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2888 
| License       | MIT

Prevent:
```
1) Symfony\UX\LiveComponent\Tests\Functional\Form\ComponentWithFormTest::testFormWithLivePropContainingAnEntityImplementingAnInterface
LogicException: Cannot dehydrate value typed as interface "Symfony\UX\LiveComponent\Tests\Fixtures\Entity\User" on component "Symfony\UX\LiveComponent\Tests\Fixtures\Component\FormWithUserInterfaceComponent". Change this to a concrete type that can be dehydrated. Or set the hydrateWith/dehydrateWith options in LiveProp or set "useSerializerForHydration: true" on the LiveProp to use the serializer.
```

Given the LiveComponent:
```php
<?php

/*
 *
 *
 */

namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;

use Symfony\Component\Form\FormInterface;
use Symfony\UX\LiveComponent\ComponentWithFormTrait;
use Symfony\UX\LiveComponent\Tests\Fixtures\Entity\User;
use Symfony\UX\LiveComponent\Tests\Fixtures\Form\UserFormType;

#[AsLiveComponent('form_with_user_interface', template: 'components/form_with_user_interface.html.twig')]
class FormWithUserInterfaceComponent extends AbstractController
{
    use ComponentWithFormTrait;
    use DefaultActionTrait;

    #[LiveProp]
    public User $user;

    {
        return $this->createForm(UserFormType::class, $this->user);
    }
}

```

Dumping `$propMetadata` in `\Symfony\UX\LiveComponent\LiveComponentHydrator::dehydrateValue` gives: 
```
Symfony\UX\LiveComponent\Metadata\LivePropMetadata {#240
  -name: "user"
  -liveProp: Symfony\UX\LiveComponent\Attribute\LiveProp {#230
    -writable: false
    -hydrateWith: null
    -dehydrateWith: null
    -useSerializerForHydration: false
    -serializationContext: []
    -fieldName: null
    -format: null
    -updateFromParent: false
    -onUpdated: null
    -url: false
    -modifier: null
  }
  -type: Symfony\Component\TypeInfo\Type\NullableType {#4197
    -types: array:2 [
      0 => Symfony\Component\TypeInfo\Type\ObjectType {#5811
        -className: "Symfony\Component\Security\Core\User\UserInterface"
      }
      1 => Symfony\Component\TypeInfo\Type\BuiltinType {#4208
        -typeIdentifier: Symfony\Component\TypeInfo\TypeIdentifier {#5983
          +name: "NULL"
          +value: "null"
        }
      }
    ]
    -type: Symfony\Component\TypeInfo\Type\ObjectType {#5811}
  }
}
```

The class name is the `UserInterface` and not `User` 🤔 

## To tests it

- With PropertyInfo only: `sfcp req symfony/property-info:^6.4 -W && sfp vendor/bin/simple-phpunit tests/Functional/Form/ComponentWithFormTest.php --filter "testFormWithLivePropContainingAnEntityImplementingAnInterface"`
- With PropertyInfo & Type: `sfcp req 'symfony/property-info:7.2.*' 'symfony/type-info:7.2.*' && sfp vendor/bin/simple-phpunit tests/Functional/Form/ComponentWithFormTest.php --filter "testFormWithLivePropContainingAnEntityImplementingAnInterface"`

### DIFF
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -13,9 +13,9 @@ namespace Symfony\UX\LiveComponent\Metadata;
 
 use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
 use Symfony\Component\PropertyInfo\Type as LegacyType;
-use Symfony\Component\TypeInfo\Type\IntersectionType;
-use Symfony\Component\TypeInfo\Type\NullableType;
-use Symfony\Component\TypeInfo\Type\UnionType;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\Component\TypeInfo\Type\CollectionType;
+use Symfony\Component\TypeInfo\TypeResolver\TypeResolver;
 use Symfony\Contracts\Service\ResetInterface;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
 use Symfony\UX\TwigComponent\ComponentFactory;
@@ -33,7 +33,11 @@ class LiveComponentMetadataFactory implements ResetInterface
     public function __construct(
         private ComponentFactory $componentFactory,
         private PropertyTypeExtractorInterface $propertyTypeExtractor,
+        private ?TypeResolver $typeResolver = null,
     ) {
+        if (method_exists($this->propertyTypeExtractor, 'getType') && !$this->typeResolver) {
+            throw new \LogicException('Symfony TypeInfo is required to use LiveProps. Try running "composer require symfony/type-info".');
+        }
     }
 
     public function getMetadata(string $name): LiveComponentMetadata
@@ -77,13 +81,13 @@ class LiveComponentMetadataFactory implements ResetInterface
 
     public function createLivePropMetadata(string $className, string $propertyName, \ReflectionProperty $property, LiveProp $liveProp): LivePropMetadata|LegacyLivePropMetadata
     {
+        $reflectionType = $property->getType();
+        if ($reflectionType instanceof \ReflectionUnionType || $reflectionType instanceof \ReflectionIntersectionType) {
+            throw new \LogicException(\sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
+        }
+
         // BC layer when "symfony/type-info" is not available
         if (!method_exists($this->propertyTypeExtractor, 'getType')) {
-            $type = $property->getType();
-            if ($type instanceof \ReflectionUnionType || $type instanceof \ReflectionIntersectionType) {
-                throw new \LogicException(\sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property %s in %s.', $property->getName(), $property->getDeclaringClass()->getName()));
-            }
-
             $infoTypes = $this->propertyTypeExtractor->getTypes($className, $propertyName) ?? [];
 
             $collectionValueType = null;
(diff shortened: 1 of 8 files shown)


Kocal:
After some investigations, it happens because of `PhpStanExtractor` from Symfony PropertyInfo.

When trying to resolve the type of `FormWithUserInterfaceComponent::$user`, it will instead resolve the type of `AbstractController::getUser()` which returns `?UserInterface`.

Kocal:
Opened https://github.com/symfony/symfony/issues/61113

Kocal:
After more investigations, I don't think it's related to PropertyInfo itself (PropertyInfo 6.4 also returns type for `AbstractController::getUser()`) but because of the extra logic that use the native `ReflectionType` if possible: 
image

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -33,7 +33,11 @@ class LiveComponentMetadataFactory implements ResetInterface
     public function __construct(
         private ComponentFactory $componentFactory,
         private PropertyTypeExtractorInterface $propertyTypeExtractor,
+        private ?TypeResolver $typeResolver = null,
     ) {
+        if (method_exists($this->propertyTypeExtractor, 'getType') && !$this->typeResolver) {
+            throw new \LogicException('Symfony TypeInfo is required to use LiveProps. Try running "composer require symfony/type-info".');
+        }
```
Kocal, line 40: Should not happens since `PropertyTypeInfoExtractor::getType()` exists since 7.1, which also ship TypeInfo, but just in case..

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -115,10 +119,21 @@ public function createLivePropMetadata(string $className, string $propertyName,
...
+
+            $collectionValueType = $infoType instanceof CollectionType ? $infoType->getCollectionValueType() : null;
+
+            if (null !== $collectionValueType && null !== $infoType) {
+                $type = $infoType;
+            } elseif (null !== $reflectionType) {
+                $type = $this->typeResolver->resolve($reflectionType);
+            } else {
+                $type = Type::mixed();
             }
```
Kocal: I'm really not sure about this code, it feels fragile.

I tried to re-use the legacy logic and use `ReflectionType` or the result of `PropertyTypeExtractorInterface::getType()`, and now all tests pass, even the one added in this PR (which was previously breaking)

smnandre (approved):

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -115,10 +119,16 @@ public function createLivePropMetadata(string $className, string $propertyName,
...
             );
         } else {
-            $type = $this->propertyTypeExtractor->getType($className, $property->getName());
+            $infoType = $this->propertyTypeExtractor->getType($className, $property->getName());
+
+            $collectionValueType = $infoType instanceof CollectionType ? $infoType->getCollectionValueType() : null;
 
-            if ($type instanceof UnionType && !$type instanceof NullableType || $type instanceof IntersectionType) {
-                throw new \LogicException(\sprintf('Union or intersection types are not supported for LiveProps. You may want to change the type of property "%s" in "%s".', $propertyName, $className));
+            if (null !== $collectionValueType && null !== $infoType) {
```
mtarld:
```suggestion
            if ($infoType instance of CollectionType) {
```
Seems to be enough no?

Kocal: I think yes, it would ease the readability too

Kocal: Applied, thanks

Kocal:
I've added some comments to improve comprehensibility. 

Since the PR fixed the issue from issue's author (https://github.com/symfony/ux/issues/2888#issuecomment-3068613029), and no tests are failing (either when using `PropertyTypeExtractorInterface::getTypes()` or `PropertyTypeExtractorInterface::getType()`), I'm merging.

Thanks both of you :)

PR #2923, opened 2025-07-14
[TwigComponent]: Add the ability to spread attributes from the `...` attribute

LexAgone:
[TwigComponent]: Add the ability to spread attributes from the `...` attribute  

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

As a front-end developer working primarily with Twig, I frequently create components that require nested attributes. However, there's currently no way to pass an array of attributes with a prefix.

For example, given a component:

```twig
<twig:Card>[...]</twig:Card>
```

I can currently write:

```twig
<twig:Card {{ ...myAttributes }}>[...]</twig:Card>
```

or:

```twig
<twig:Card attributes="{{ myAttributes }}">[...]</twig:Card>
```

But I **can't** do:

```twig
<twig:Card nested:{{ ...myAttributes }}>[...]</twig:Card>
```

or:

```twig
<twig:Card nested:attributes="{{ myAttributes }}">[...]</twig:Card>
```

That's why I'm proposing support for a new spread-like syntax using `...` to allow passing an array of attributes (`array`, `Traversable`, or `StimulusAttributes`) with a prefix.

### Real-world use case:

I have a Twig extension (`App\Twig\Extension\ConfirmExtension`) that provides a `confirm()` function returning a `StimulusAttributes` object.

I also have a component `<twig:TableRow/>` that contains a delete button.

Currently, I **cannot** write:

```twig
<twig:TableRow btn:{{ ...confirm() }} />
```

With this PR, my proposed syntax would allow the following:

```twig
<twig:TableRow btn:...="{{ confirm() }}" />
```

This improves flexibility and developer experience when working with prefixed dynamic attributes in Twig components.

### DIFF
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -26,6 +26,11 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
     private const ALPINE_REGEX = '#^x-([a-z]+):[^:]+$#';
     private const VUE_REGEX = '#^v-([a-z]+):[^:]+$#';
 
+    private const SPREADABLE_ATTRIBUTE = '...';
+
+    /** @var array<string, string|bool> */
+    private array $attributes = [];
+
     /** @var array<string,true> */
     private array $rendered = [];
 
@@ -33,9 +38,10 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
      * @param array<string, string|bool> $attributes
      */
     public function __construct(
-        private array $attributes,
+        array $attributes,
         private readonly EscaperRuntime $escaper,
     ) {
+        $this->attributes = $this->handleAttributes($attributes);
     }
 
     public function __toString(): string
@@ -101,6 +107,31 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
     public function __clone(): void
     {
         $this->rendered = [];
+    }
+
+    private function handleAttributes(array $attributes): array
+    {
+        $spreadAttributes = $attributes[self::SPREADABLE_ATTRIBUTE] ?? null;
+
+        if (!$spreadAttributes) {
+            return $attributes;
+        }
+
+        if ($spreadAttributes instanceof StimulusAttributes) {
+            $spreadAttributes = $spreadAttributes->toArray();
+        }
+
+        if ($spreadAttributes instanceof \Traversable) {
+            $spreadAttributes = iterator_to_array($spreadAttributes);
+        }
+
+        if (!\is_array($spreadAttributes)) {
+            throw new \InvalidArgumentException(\sprintf('The "%s" attribute must be an array, "%s" given.', self::SPREADABLE_ATTRIBUTE, get_debug_type($spreadAttributes)));
+        }
+
+        unset($attributes[self::SPREADABLE_ATTRIBUTE]);
+
+        return [...$attributes, ...$spreadAttributes];
     }
 
     public function render(string $attribute): ?string
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -113,12 +113,6 @@ final class ComponentFactory implements ResetInterface
         $attributes = $data[$attributesVar] ?? [];
         unset($data[$attributesVar]);
 
-        foreach ($data as $key => $value) {
-            if ($value instanceof \Stringable) {
-                $data[$key] = (string) $value;
-            }
-        }
-
         return new MountedComponent(
             $componentMetadata->getName(),
             $component,
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.26.0
+
+-   Add the ability to spread attributes from the `...` attribute
+
 ## 2.25.2
 
 -   Fix `ComponentAttributes` rendering when using `StimulusAttributes` as default attributes


src/TwigComponent/src/ComponentFactory.php
```diff
@@ -113,12 +113,6 @@ public function mountFromObject(object $component, array $data, ComponentMetadat
         $attributes = $data[$attributesVar] ?? [];
         unset($data[$attributesVar]);
 
-        foreach ($data as $key => $value) {
-            if ($value instanceof \Stringable) {
-                $data[$key] = (string) $value;
-            }
-        }
```
smnandre, line 120: why this removal ?

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -103,6 +109,31 @@ public function __clone(): void
         $this->rendered = [];
     }
 
+    private function handleAttributes(array $attributes): array
+    {
+        $spreadAttributes = $attributes[self::SPREADABLE_ATTRIBUTE] ?? null;
```
smnandre, line 114: This check should be done in constructor to avoid method call for everyone not using this feature

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -103,6 +109,31 @@ public function __clone(): void
...
+            $spreadAttributes = iterator_to_array($spreadAttributes);
+        }
+
+        if (!\is_array($spreadAttributes)) {
+            throw new \InvalidArgumentException(\sprintf('The "%s" attribute must be an array, "%s" given.', self::SPREADABLE_ATTRIBUTE, get_debug_type($spreadAttributes)));
+        }
+
+        unset($attributes[self::SPREADABLE_ATTRIBUTE]);
+
+        return [...$attributes, ...$spreadAttributes];
```
smnandre, line 134: This is something that can be done userland, no ? Wether in PHP or Twig

smnandre:
I'm not a bin fan of hard-coding something that uses a Twig construct (the `...`) to create another behaviour. 

```twig
<twig:TableRow btn:...="{{ confirm() }}" />
```

I'm very surprised we allow "..." as attribute name to be honest, and think this is something we should fix.


For your initial need, why don't you prefix your keys in your function ? Or pass it as prop instead as attribute ? Anything not working as expected here ?

src/TwigComponent/src/ComponentFactory.php
LexAgone: Because this loop force attributes to be strings if they are Stringable.  However, StimulusAttributes implements Stringable so I can't pass recursively the object in the ComponentAttributes. Maybe I can exclude StimulusAttributes like this :  
`if (!$value instanceof \StimulusAttributes && $value instanceof \Stringable) {` ?

src/TwigComponent/src/ComponentAttributes.php
LexAgone: I didn't want to break the code guideline but ok, if it's allowed

src/TwigComponent/src/ComponentAttributes.php
LexAgone: What do you mean ? The interest is to be recursive. Currently, for the "primary" attributes it's not that usefull but for nested attributes, we need that

LexAgone:
> I'm very surprised we allow "..." as attribute name to be honest, and think this is something we should fix.

My first thought was to use the "*" attribute but I had to edit the PreLexer to allow this character, but I saw the "." was allowed so I was like... Ok, why not...


> For your initial need, why don't you prefix your keys in your function ? Or pass it as prop instead as attribute ? Anything not working as expected here ?

Here is my ConfirmExtension function :

```php
readonly class ConfirmExtension
{
    public function __construct(
        #[Autowire(service: 'stimulus.helper')]
        private StimulusHelper $stimulusHelper,
    ) {
    }

    #[AsTwigFunction(name: 'confirm', isSafe: ['html_attr'])]
    public function getConfirm(
        string $question = 'Êtes-vous sûr de vouloir effectuer cette action ?',
        string $confirmLabel = 'Confirmer',
        string $cancelLabel = 'Annuler',
        string $title = 'Confirmation',
    ): StimulusAttributes {
        $attributes = $this->stimulusHelper->createStimulusAttributes();
        $attributes->addController('admin--confirm');
        $attributes->addAction(
            'admin--confirm',
            'showConfirm',
            'click',
            [
                'message' => $question,
                'confirm-label' => $confirmLabel,
                'cancel-label' => $cancelLabel,
                'title' => $title,
            ]
        );

        return $attributes;
    }
}
```

There is no way to pass prefix in the StimulusHelper. Maybe I could transform the `StimulusAttributes` to array to add prefix passed in props of the function and prefix all the keys... But It could be so much easier and convenient to be allowed to pass the object directly in the spreadable attribute. I have several exemples in my projects where It would be nice to have this.  

And also, it's possible to have the functionality in the first instance of attributes, why not for the nested ones ?

src/TwigComponent/src/ComponentAttributes.php
smnandre: How do you mean? 

My point was: we would rather not add a method call like this for every ComponentAttributes instanciation, so _**if**_ we were to add this behaviour, we sould probaly do something like

```php
public function __construct(
        array $attributes,
        private readonly EscaperRuntime $escaper,
    ) {
       
           if (array_key_exists(self::SPREADABLE_ATTRIBUTES, $attributes)) {
                // method call here
           }
      }
```

smnandre (commented):
>     private const SPREADABLE_ATTRIBUTE = '...';

I'm not _100% opposed_ to the general idea of “passing down attributes,” but I don't think this is the right approach:

* `...`  is a language construct in both PHP and Twig. Reusing it here as a `string` identifier or `array key` feels confusing and potentially misleading ;
* supporting nested arrays for attributes would be a significant change. Given the escaping issues we've already faced earlier this year, I'm not convinced it's worth the risk at this point.

For those two reasons, I'm 👎 on this proposal.

@kbond @Kocal

src/TwigComponent/src/ComponentFactory.php
smnandre: > Because this loop force attributes to be strings if they are Stringable

This was done very much on purpose :)  As StimulusAttributes accept string|bool attribute values

smnandre:
For now the best solution to implement what you want would be to encapsulate your nested component in a block, and pass the attributes directly to the component you want to set them on

```twig
<twig:Card>
    <twig:block name="card_header">
        <twig:CardHeader {{ ...myAttributes }}  />
    </twig:block>
</twig:Card>
```

Kocal:
I'm not a super fan about nested attributes, but using `...` in an attribute name feels super weird 😕

LexAgone:
> For now the best solution to implement what you want would be to encapsulate your nested component in a block, and pass the attributes directly to the component you want to set them on
> 
> ```
> <twig:Card>
>     <twig:block name="card_header">
>         <twig:CardHeader {{ ...myAttributes }}  />
>     </twig:block>
> </twig:Card>
> ```

Ok, thanks for this review. I understand the unwillingness of this feature. But I really think components miss this passing down to nested attributes capacity. Maybe another approach like allowing something like that :  
  
`<twig:Card {{ ...{ header: stimulus_controller('test') } }}  />`  
  
I had the same "problem" again today

LexAgone:
In the end, for those who expected the same functionality, I created a twig filter that works pretty well :

 ```twig
<twig:Card {{ ...stimulus_controller('test')|stimulus_action('test', 'my_action', 'click')|prefix_keys('header:') }} />
```

Here is my file PrefixKeysExtension.php :  

```php
<?php

declare(strict_types=1);

namespace App\Twig\Extension;

use Twig\Attribute\AsTwigFilter;

readonly class PrefixKeysExtension
{
    #[AsTwigFilter(name: 'prefix_keys')]
    /**
     * @param iterable<string> $data
     */
    public function prefixKeys(iterable $data, string $prefix): array
    {
        if ($data instanceof \Traversable) {
            $data = iterator_to_array($data);
        }

        return array_combine(
            array_map(fn ($key): string => $prefix.$key, array_keys($data)),
            array_values($data)
        );
    }
}

```

Kocal:
Hi, for the reasons explained above and since you find a simple alternative in user-land, I'm closing the feature. 
Thanks for the PR, and thanks for your understanding!

PR #2931, opened 2025-07-20
Explicitly import Vitest APIs instead of relying on globals

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Purely internal.

With Vitest, it is "recommended" to not use globals ([disabled by default](https://vitest.dev/config/#globals)) and instead relying on explicit `import`.

Even if it means to add more code in the codebase, “Explicit is better than implicit”, and global types definitions can be a mess sometimes.

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -7,4 +7,4 @@
  * file that was distributed with this source code.
  */
 
-import '@testing-library/jest-dom';
+import '@testing-library/jest-dom/vitest';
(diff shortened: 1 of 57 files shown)


test/setup.js
```diff
@@ -7,4 +7,4 @@
  * file that was distributed with this source code.
  */
 
-import '@testing-library/jest-dom';
+import '@testing-library/jest-dom/vitest';
```
Kocal, line 10: Needed, because `@testing-library/jest-dom` relied on global `expect`

PR #2222, opened 2024-09-28
Migrate from Yarn Classic to Yarn Berry

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Issues        | Fix #... 
| License       | MIT

[Yarn 1.x](https://github.com/yarnpkg/yarn) is _dead_ since 4 years now:
image
all new features and bug fixes are done on [Yarn Berry](https://github.com/yarnpkg/berry):
image

I've adapted our CI, our building and testing scripts. 
Updates on `src/**/assets/packages.json` are simply dependencies alphabetical sorting. 
The `yarn.lock` has been re-generated to the new structure (but locked versions are still the same).

PR #2221, opened 2024-09-28
[Swup] For Swup's dev dependencies exclusion from our yarn.lock

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Issues        | Fix #... 
| License       | MIT

While working on #2219, I've noticed we had a very old Rollup version that we do not use at all. 
The issue comes from Swup v3, which ship its **building** dependencies as **prod** dependencies, which means we install `microbundle` and all its dependencies:
```
microbundle@^0.15.1:
  version "0.15.1"
  resolved "https://registry.npmjs.org/microbundle/-/microbundle-0.15.1.tgz"
  integrity sha512-aAF+nwFbkSIJGfrJk+HyzmJOq3KFaimH6OIFBU6J2DPjQeg1jXIYlIyEv81Gyisb9moUkudn+wj7zLNYMOv75Q==
  dependencies:
    "@babel/core" "^7.12.10"
    "@babel/plugin-proposal-class-properties" "7.12.1"
    "@babel/plugin-syntax-import-meta" "^7.10.4"
    "@babel/plugin-syntax-jsx" "^7.12.1"
    "@babel/plugin-transform-flow-strip-types" "^7.12.10"
    "@babel/plugin-transform-react-jsx" "^7.12.11"
    "@babel/plugin-transform-regenerator" "^7.12.1"
    "@babel/preset-env" "^7.12.11"
    "@babel/preset-flow" "^7.12.1"
    "@babel/preset-react" "^7.12.10"
    "@rollup/plugin-alias" "^3.1.1"
    "@rollup/plugin-babel" "^5.2.2"
    "@rollup/plugin-commonjs" "^17.0.0"
    "@rollup/plugin-json" "^4.1.0"
    "@rollup/plugin-node-resolve" "^11.0.1"
    "@surma/rollup-plugin-off-main-thread" "^2.2.2"
    asyncro "^3.0.0"
    autoprefixer "^10.1.0"
    babel-plugin-macros "^3.0.1"
    babel-plugin-transform-async-to-promises "^0.8.18"
    babel-plugin-transform-replace-expressions "^0.2.0"
    brotli-size "^4.0.0"
    builtin-modules "^3.1.0"
    camelcase "^6.2.0"
    escape-string-regexp "^4.0.0"
    filesize "^6.1.0"
    gzip-size "^6.0.0"
    kleur "^4.1.3"
    lodash.merge "^4.6.2"
    postcss "^8.2.1"
[...]

PR #2932, opened 2025-07-20
Replace Yarn Berry by PNPM

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Purely internal, this PR replaces Yarn by PNPM for the following reasons:
- when working on https://github.com/symfony/ux/pull/2222, I tried to use PNPM instead but it was not possible to not install [Swup "dev"-dependencies](https://github.com/symfony/ux/pull/2221), this is now possible
- it is [more efficient and performant](https://pnpm.io/benchmarks), and better for your disk
- it has more [features](https://pnpm.io/feature-comparison), better Workspaces support and [Catalogs](https://pnpm.io/catalogs) 🤩 
- commands to release packages on npm are simpler

For developers working on UX, [pnpm can be installed with Corepack](https://pnpm.io/installation#using-corepack)

Note: the ~26k lines changed are mainly about lock files


.github/workflows/release-on-npm.yaml
```diff
@@ -25,29 +25,25 @@ jobs:
...
 
           - name: Commit changes
             run: |
               git add .
               git commit -m "Update versions to ${{ env.VERSION }}"
 
-          - name: Replace local "workspace:*" occurrences
-            run: |
-              yarn workspaces foreach -pA exec "sed -i 's/\"workspace:\*\"/\"${{ env.VERSION }}\"/g' package.json"
-
```
Kocal, line 46: This is automatically done when running `pnpm publish`, see https://pnpm.io/workspaces#publishing-workspace-packages

package.json
```diff
@@ -17,28 +17,14 @@
         "@rollup/plugin-commonjs": "^28.0.6",
         "@rollup/plugin-node-resolve": "^16.0.1",
         "@rollup/plugin-typescript": "^12.1.4",
-        "@testing-library/dom": "^10.4.0",
-        "@testing-library/jest-dom": "^6.6.3",
-        "@testing-library/user-event": "^14.6.1",
+        "@testing-library/dom": "catalog:",
```
Kocal, line 20: [Catalogs](https://pnpm.io/catalogs) are like an alias to a constraint version. 
When defining the catalog version for `@testing-library/dom`, then it will always install the exact same version when installing the same package in different places (if we use the same catalog ofc)

src/Autocomplete/assets/package.json
```diff
@@ -49,7 +49,15 @@
...
         "@hotwired/stimulus": "^3.0.0",
+        "@testing-library/dom": "catalog:",
+        "@testing-library/jest-dom": "catalog:",
+        "@testing-library/user-event": "catalog:",
+        "jsdom": "catalog:",
         "tom-select": "^2.2.2",
+        "tslib": "catalog:",
+        "tsx": "catalog:",
+        "typescript": "catalog:",
+        "vitest": "catalog:",
```
Kocal, line 60: Adding the (dev) dependencies used by the package instead of relying of packages installed from parents directories (that "magically" works with Yarn) is _a good practice_.

By using `catalog:`, we will have the same exact version of the dependency installed everywhere (consistency).

src/React/assets/src/register_controller.ts
```diff
@@ -7,6 +7,8 @@
  * file that was distributed with this source code.
  */
 
+/// <reference types="@types/webpack-env" />
```
Kocal, line 10: Otherwise `__WebpackModuleApi` was not found

src/Map/src/Bridge/Google/assets/src/map_controller.ts
```diff
@@ -7,6 +7,8 @@
  * file that was distributed with this source code.
  */
 
+/// <reference types="google.maps" />
```
Kocal, line 10: Otherwise `google` type was not found

.github/workflows/app-tests.yaml
```diff
@@ -54,33 +54,29 @@ jobs:
...
 
-        - if: matrix.ux-packages-source == 'php-vendor'
-          name: Refresh dependencies from vendor/
-          working-directory: test_apps/encore-app
-          run: yarn
-          env:
-            YARN_ENABLE_HARDENED_MODE: 0
-            YARN_ENABLE_IMMUTABLE_INSTALLS: 0
+        - working-directory: test_apps/encore-app
+          run: pnpm install --ignore-workspace
```
Kocal, line 64: `--ignore-workspace` allows to ignore the `pnpm-workspace.yaml` from parent folders, and install dependencies from `test_apps/encore-app` only (`test_apps/encore-app` is not a workspace on purpose)

smnandre:
What does it mean for the contributors?

smnandre:
Nice job and i’m all in to leave legacy yarn :)

Kocal:
> What does it mean for the contributors?

pnpm will be automatically installed when running `corepack enable`, which we already explain in our CONTRIBUTING.md. And of course, commands like `yarn install` or `yarn build` won't work anymore, in favor of `pnpm install`, `pnpm build`, etc...

kbond (approved):
Good arguments for this change!

PR #2935, opened 2025-07-21
Simplify our building packages tool (drop Rollup for tsdown)

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

This PR simplify a lot our building process by moving from Rollup to [tsdown](https://tsdown.dev/), based on [Rolldown](https://rolldown.rs/) (compatible with Rollup API).

# Why?

Rolldown is much modern and faster than Rollup (note: we do not really see the speed impact here, since our files are pretty small). 

Tsdown is a library bundler based on Rolldown, and it requires far less configuration or manual tweaks than our actual solution:
- it ease CSS bundling a lot, no need for a custom plugin for build & watch support
- we do not need logging code anymore, tsdown handle it for us
- errors are now nicely displayed (see video below)
- super easy to add TypeScript support, with types emiting too (no need for a custom plugin that move types to the good place)
- we can focus on the essential

The file `bin/build_package.ts` becomes smaller and easier to understand, and we don't even need `bin/rollup.ts` file anymore!

# Review

When reviewing the PR, you will notices some modifications on `dist/` files:
1. the code contains `//#region`: these are regions useful when reading the file in an IDE like VSCode, it allows to fold/unfold specific portions of the code. Unfortunately it's not configurable, but we can remove them by minifying them (thanks to `minify` option) 
2. I feel like the code contains more polyfill and other helpers than before: this is because rolldown use the [oxc toolchain](https://github.com/oxc-project/oxc) which generates a similar but different code than Rollup/TypeScript. 
   
   ~~This is fine for the moment, but in a next PR~~ I will upgrade the target to `es2022` or something similar (`static` keyword and private class fields/methods are now supported at 95%)

   **EDIT:** the current polyfill for `static` keyword with target `es2021` break Stimulus controllers, upgrading to `es2022` works fine

4. A lot of files are removed: that's fine, it's only about `.d.ts` files. Their types can now be found in the main dist file (e.g. `controller.d.ts`) 

# Demo 

When building a single package:
```
➜  assets git:(tsdown) pnpm build

> @symfony/ux-autocomplete@2.27.0 build /Users/kocal/workspace-os/symfony-ux/src/Autocomplete/assets
> tsx ../../../bin/build_package.ts .

ℹ entry: src/controller.ts
ℹ target: es2021
ℹ tsconfig: ../../../tsconfig.packages.json
ℹ Build start
ℹ Cleaning 2 files
ℹ dist/controller.js    16.24 kB │ gzip: 4.09 kB
ℹ dist/controller.d.ts   1.91 kB │ gzip: 0.65 kB
ℹ 2 files, total: 18.15 kB
✔ Build complete in 1036ms
```

When watching a single package:

https://github.com/user-attachments/assets/58d2c46a-fd9f-4f48-9d00-74eabeced00a


Kocal:
Converting in draft, I will investigate why Turbo E2E tests do not pass anymore

**EDIT:** fixed when upgrading the target from `es2021` to `es2022`

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -1,3240 +1,2505 @@
...
-        this.updatedModels = updateModels;
-    }
-    containsOneOfActions(targetedActions) {
-        return this.actions.filter((action) => targetedActions.includes(action)).length > 0;
-    }
-    areAnyModelsUpdated(targetedModels) {
-        return this.updatedModels.filter((model) => targetedModels.includes(model)).length > 0;
-    }
+import { Controller } from "@hotwired/stimulus";
+import { Idiomorph } from "idiomorph/dist/idiomorph.esm.js";
```
Kocal: Should be inlined

src/React/assets/dist/render_controller.js
```diff
@@ -1,83 +1,100 @@
-import { Controller } from '@hotwired/stimulus';
-import React from 'react';
-import require$$0 from 'react-dom';
+import { Controller } from "@hotwired/stimulus";
+import React from "react";
+import { createRoot } from "react-dom/client";
```
Kocal, line 3: To revert, we must import `react-dom` only (for AssetMapper support..)

src/LiveComponent/assets/dist/live_controller.js
Kocal: Fixed by moving `idiomorph` from `dependencies` to `devDependencies`, and it makes 100% sense (since this dependency was inlined, it's useless to download it when downloading this package).

See https://tsdown.dev/options/dependencies

Note: it was no present in `importmap` either, that's fine.

src/React/assets/dist/render_controller.js
Kocal: Note: This is the actual solution to deal with the issue with Rollup https://github.com/symfony/ux/blob/2.x/bin/rollup.ts#L123-L130

Kocal: Well, even with the BC promise and cie, I feel it's wrong to try to find hacks to reintroduce the previous behavior that was buggy.

Defining `react-dom/client` and write an changelog entry would be enough I think... WDYT @kbond @smnandre?

src/Map/src/Bridge/Leaflet/assets/dist/map_controller.js
```diff
@@ -1,357 +1,440 @@
-import { Controller } from '@hotwired/stimulus';
-import 'leaflet/dist/leaflet.min.css';
```
Kocal, line 2: The CSS it not imported anymore, it should break Leaflet map display for Webpack users

Kocal: **EDIT:** fixed when upgrading the target from `es2021` to `es2022`

src/React/assets/dist/render_controller.js
kbond: What's the impact on users?

src/LiveComponent/assets/package.json
```diff
@@ -17,7 +17,7 @@
     "main": "dist/live_controller.js",
     "types": "dist/live_controller.d.ts",
     "config": {
-        "css_source": "styles/live.css"
+        "css_source": "src/live.css"
```
Kocal, line 20: I moved the `live.css` file under `src` aside js source code, otherwise tsdown/rolldown create the following files:
- `dist/src/live_controller.js`
- `dist/styles/live.min.css`

Which is not what we want.

Note: other UX packages define their styles in `src/` too

src/React/assets/dist/render_controller.js
Kocal: ~~AssetMapper users will be required to run `php bin/console importmap:require react-dom/client` when upgrading UX React, otherwise it will **break** in the browser since `react-dom/client` asset does not exist (where `react-dom` already exists).~~

~~I was hopping Flex to automatically fetch new importmap dependencies when upgrading a package, but it didn't do anything when running `composer upgrade symfony/ux-react` :/~~

Webpack and/or npm users are not affected, since dependencies are nicely resolved.

Kocal: 🤦🏻 I'm dumb, ofc it won't works if I rely on the `link` script and not a composer repository to link UX React package from this PR to my lab project.

That's fine, I can confirm the dependency `react-dom/client` is correctly added when upgrading UX React: 
image

So, as mentioned in https://github.com/symfony/ux/blob/2.x/bin/rollup.ts#L123-L130, the issue will only affect users using AssetMapper but not Flex. 

I don't think there is many Symfony projects using the AssetMapper, UX React, but not Flex. WDYT?

Kocal: I just pushed some changes for react-dom/client and a changelog entry

smnandre:
> A lot of files are removed: that's fine, it's only about .d.ts files. Their types can now be found in the main dist file (e.g. controller.d.ts)

That’s actually what worries me the most… what happens to users currently importing types directly from those files?

For all the rest I’m very 👍

PS: let’s be very very attentive to details here and consider all potential outcomes — we now have a non-negligible number of users and, as you mentioned, we still have a BC promise to keep 😅

bin/build_package.ts
```diff
@@ -44,107 +43,80 @@ async function main() {
...
-            if (inputStyleFile && id === inputStyleFile) {
-                await buildCss();
+    build({
+        entry: inputFiles,
+        outDir: distDir,
+        clean: true,
+        outputOptions: {
+            cssEntryFileNames: '[name].min.css',
+        },
+        external,
```
smnandre, line 78: no attribute value / key here ?

src/Chartjs/assets/dist/controller.js
```diff
@@ -1,73 +1,67 @@
...
+	disconnect() {
+		this.dispatchEvent("disconnect", { chart: this.chart });
+		if (this.chart) {
+			this.chart.destroy();
+			this.chart = null;
+		}
+	}
+	/**
+	* If the underlying data or options change, let's update the chart!
+	*/
```
smnandre, line 36: Different rule about code comments ?

src/Cropperjs/assets/dist/controller.js
```diff
@@ -1,31 +1,41 @@
...
+		img.classList.add("cropperjs-image");
+		img.src = this.publicUrlValue;
+		const parent = this.element.parentNode;
+		if (!parent) throw new Error("Missing parent node for Cropperjs");
+		parent.appendChild(img);
+		const options = this.optionsValue;
+		this.dispatchEvent("pre-connect", {
+			options,
+			img
+		});
```
smnandre, line 21: Any settings we could touch here ? This seems to add length to the file ?

smnandre: (not that it changes anything in reality, but i'm genuinely curious about this formatting)

smnandre:
How to explain the difference in LiveComponent dist/controller.js ? idiomorph ?

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -1,3240 +1,2991 @@
...
-        this.hooks.set(hookName, hooks);
-    }
-    triggerHook(hookName, ...args) {
-        const hooks = this.hooks.get(hookName) || [];
-        hooks.forEach((callback) => callback(...args));
-    }
-}
-
-// base IIFE to define idiomorph
-var Idiomorph = (function () {
```
smnandre, line 513: Any idea why ?

src/LiveComponent/assets/package.json
```diff
@@ -44,9 +44,6 @@
             "@symfony/ux-live-component": "path:%PACKAGE%/dist/live_controller.js"
         }
     },
-    "dependencies": {
-        "idiomorph": "^0.3.0"
-    },
```
smnandre, line 49: why ?

src/React/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## 2.28.0
+
+-   [BC BREAK] If you are using the Symfony AssetMapper but **not** Symfony Flex,
+    you need to upgrade your `importmap.php` and change the asset `react-dom` to `react-dom/client`,
+    and run `php bin/console importmap:install`.
+
+    Symfony Flex or Webpack Encore users are not affected.
```
smnandre, line 9: I really don't see why we would release this. No feature, no bug, but a big break in userland code.

Let's at least wait to push until we have something to....

src/React/assets/dist/components.js
```diff
@@ -1,3 +1,5 @@
+//#region src/components.ts
 const components = {};
 
-export { components };
+//#endregion
+export { components };
```
smnandre, line 5: Little pb of file ending here right ?

src/React/assets/dist/render_controller.js
smnandre: Ok I was not yet here when i commented previous comment :)  

Thanks for the detailed explanation... I still think that releasing a version that will break people apps, when we bring zero new feature, is neither a good habit nor a good message to send 🤷 

But i understand you want to change everything at once... So to me we need to find a way to "not" release React here.

WDYT ?

tsconfig.packages.json
```diff
@@ -8,7 +8,8 @@
         "rootDir": "src",
         "strict": true,
         "strictPropertyInitialization": false,
-        "target": "es2021",
+        "//": "The target should be kept in sync with `bin/build_packages.ts` file.",
+        "target": "es2022",
```
smnandre, line 12: Can we add polyfills for Object.hasOwn() and Array.at() ?

bin/build_package.ts
Kocal: `{ external: external }` can be shortened to `{ external }`

src/LiveComponent/assets/package.json
Kocal: See https://github.com/symfony/ux/pull/2935#discussion_r2221428147

src/React/CHANGELOG.md
Kocal: I understand your point which is totally legitimate, but the current behavior with Rollup and Rollup TypeScript plugin is buggy.

When building the package, our [`import { createRoot } from 'react-dom/client';`](https://github.com/symfony/ux/blob/2.x/src/React/assets/src/render_controller.ts#L12) is modified to this weird [`import require$$0 from 'react-dom';`](https://github.com/symfony/ux/blob/2.x/src/React/assets/dist/render_controller.js) import, that use `react-dom` instead of `react-dom/client`.
This behaviour was fixed by https://github.com/rollup/plugins/pull/1310, but then I had to find a hack to re-introduce the bug in order to prevent a BC, see https://github.com/symfony/ux/blob/2.x/bin/rollup.ts#L125-L129

I may be biased, but who does not use Flex in 2025? 😛 

More generally, it's a bit unlucky to have to restrict ourselves here because of AssetMapper/Importmap, we lose the freedom to change some pure JavaScript related things where it wouldn't be a problem with in a full Node.js ecosystem (with npm, webpack Encore, ...) 😕 

I will try to investigate if we can re-add this `import require$$0 from 'react-dom';` (or equivalent), but I don't think this will be possible

src/React/assets/dist/components.js
Kocal: It's the dist files, that's fine I guess

tsconfig.packages.json
Kocal: `Array.at()` is available at 93.9% and `Object.hasOwn()` at 93.87%, sure?

src/Cropperjs/assets/dist/controller.js
Kocal: Idk, sometimes I see that a multiline `{ ... }` is inlined and sometimes not. I don't think tsdown/rolldown/oxc has such an option (remember that's a dist file), we can [minify files](https://tsdown.dev/options/minification) but it's still in alpha (and I don't know if we really wants to start minifying our dist files).

src/Chartjs/assets/dist/controller.js
Kocal: There is no such options available, [`legalComments`](https://rolldown.rs/reference/config-options#legalcomments) exists but it only refers to `@license` comments and equivalent

src/React/assets/dist/render_controller.js
Kocal: (let's continue at https://github.com/symfony/ux/pull/2935#discussion_r2224687166 :) )

src/React/CHANGELOG.md
Kocal: > Ok I was not yet here when i commented previous comment :)
> 
> Thanks for the detailed explanation... I still think that releasing a version that will break people apps, when we bring zero new feature, is neither a good habit nor a good message to send 🤷

Yeah I agree with you

> But i understand you want to change everything at once... So to me we need to find a way to "not" release React here.
> 
> WDYT ?

I understand, but I don't think that's good to maintain two similar but different building system. 

... What about releasing UX 3 soon, with this PR and issues from https://github.com/symfony/ux/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A3.x fixed too?

smnandre: I agree with you on everything... i just cannot stress enough that a change in build is not a feature. 

And maintain npm only package is not something i refuse or don't want (they would still work with importmap :) )... 

... i really think without a way to synchronize releases for both php and js vendor... and this is almost impossible to do properly 🤷

tsconfig.packages.json
smnandre: Ok yeah that seems fair (even 97% for both in real usage it seems) ..

.. but after this one, please can we say we do not make changes like this in minor/patch version ? 😆 🙏

src/Chartjs/assets/dist/controller.js
smnandre: Well ... we will have bigger files generated in the end.. 🤷

src/Cropperjs/assets/dist/controller.js
smnandre: Maybe it's a bit early to use such package then but i'm trusting you fully here :)

src/LiveComponent/assets/package.json
smnandre: Yes but it is not inlined anymore, right ? cf comment on top here.... so there is something weird here.

https://github.com/symfony/ux/pull/2935#pullrequestreview-3045144241

smnandre: Nevermind i'm blind

Kocal:
Let's wait a bit until tsdown and rolldown become more mature. 
Still, it's cool to see how many things were simplified, I'll try to apply these changes on 2.x

Kocal:
Opened #2940 with some fixes added here.

PR #2940, opened 2025-07-25
[LiveComponent] Some fixes on JS dependencies assets

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Fixing some incoherent things found in https://github.com/symfony/ux/pull/2935, each commit are separated and contains a link to the comment explaing why

### DIFF
--- a/src/LiveComponent/assets/package.json
+++ b/src/LiveComponent/assets/package.json
@@ -44,9 +44,6 @@
             "@symfony/ux-live-component": "path:%PACKAGE%/dist/live_controller.js"
         }
     },
-    "dependencies": {
-        "idiomorph": "^0.3.0"
-    },
     "peerDependencies": {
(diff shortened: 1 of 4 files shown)


src/LiveComponent/assets/package.json
```diff
@@ -44,9 +44,6 @@
             "@symfony/ux-live-component": "path:%PACKAGE%/dist/live_controller.js"
         }
     },
-    "dependencies": {
-        "idiomorph": "^0.3.0"
-    },
```
Kocal, line 49: To resume, since `idiomorph` **is inlined** in our dist files and there is no more `import ... from 'idiomorh'` in them, then there is no point to make it a prod dependency, it should be a dev dependency instead.

PR #2944, opened 2025-07-26
Modernize and simplify our packages building tools, replace Rollup by tsup

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT

Following #2935. Same goals, but less frictions than with tsdown, [tsup](https://github.com/egoist/tsup): 
- `target` is correctly read from our `tsconfig.packages.json`
- no `//#region` comments
- less polyfills than oxc 
- the `target` **stays** `es2021` without any issues with `static` and Stimulus

Two things:
1. About the `react-dom/client` import, yes it will impact AssetMapper users that don't use Flex, but it will positively impact other users (AssetMapper with Flex, Webpack Encore...) by removing useless code and making the file smaller
2. About the tons of `.d.ts` files removed, that's still fine, there is no point to generate `dist/<file>.d.ts` files when `dist/<file>.js` do not exist, they are not part of the public API. 

The code review must be easier, since less code has been touched than with tsdown.

--- 

# Demo

When building LiveComponent assets:

```
➜  assets git:(tsup) pnpm build

> @symfony/ux-live-component@2.27.0 build /Users/kocal/workspace-os/symfony-ux/src/LiveComponent/assets
> tsx ../../../bin/build_package.ts .

CLI Building entry: src/live.css, src/live_controller.ts
CLI Using tsconfig: ../../../tsconfig.packages.json
CLI tsup v8.5.0
CLI Target: es2021
CLI Cleaning output folder
ESM Build start
[Symfony UX] Minified CSS file: /Users/kocal/workspace-os/symfony-ux/src/LiveComponent/assets/dist/live.css
[Symfony UX] Renamed dist/live.css to dist/live.min.css
ESM dist/live_controller.js 12.25 KB
ESM dist/live.css           74.00 B
ESM ⚡️ Build success in 26ms
DTS Build start
DTS ⚡️ Build success in 1276ms
DTS dist/live_controller.d.ts 7.96 KB

```

When watching LiveComponent assets, the CSS is easily watched too!

https://github.com/user-attachments/assets/a246f278-8bf4-40f6-887e-685be7509af0


tsconfig.json
```diff
@@ -6,7 +6,8 @@
         "module": "esnext",
         "moduleResolution": "node",
         "resolveJsonModule": true,
-        "types": ["@testing-library/jest-dom"]
+        "types": ["node", "@testing-library/jest-dom"],
+        "lib": ["es2022", "dom"]
```
Kocal, line 10: Not really related to this PR, but it helped to get correct types and syntax highlighting when working on `bin/build_package.ts` file

src/Translator/assets/src/translator_controller.ts
```diff
@@ -7,4 +7,168 @@
  * file that was distributed with this source code.
  */
 
-export * from './translator';
```
Kocal, line 10: Conflicted with `options.bundle` from ESBuild, but in reality it didn't bring any value to externalize the translator logic if the final goal is to export everything in the controller. Just define everything in the controller, that's simpler.

src/StimulusBundle/assets/dist/loader.js
```diff
@@ -1,93 +1,98 @@
...
-        registerController(name, eagerControllers[name], application);
-    }
-    const lazyControllerHandler = new StimulusLazyControllerHandler(application, lazyControllers);
-    lazyControllerHandler.start();
+import { Application } from "@hotwired/stimulus";
+import {
+  eagerControllers,
+  isApplicationDebug,
+  lazyControllers
+} from "./controllers.js";
```
Kocal, line 6: Thanks to `options.bundle = false` from ESBuild and `external` from tsup, the file `controllers.js` is **not** inlined, no BC here.

src/React/assets/dist/render_controller.js
```diff
@@ -1,83 +1,47 @@
...
-	    i.usingClientEntryPoint = true;
-	    try {
-	      return m.hydrateRoot(c, h, o);
-	    } finally {
-	      i.usingClientEntryPoint = false;
-	    }
-	  };
-	}
-	return client;
-}
```
Kocal, line 37: This is the code from react-dom about handling things differently between development and prod envs. It requires `react-dom` and not `react-dom/client`, as explained in previous PRs.

Only the production path is handled, less code, smaller file, ESBuild does not need anymore to import `react-dom` with the additional logic, but only `react-dom/client`

Kocal:
Added some comments to help the review

src/StimulusBundle/assets/dist/loader.js
smnandre: Nice

bin/build_package.ts
```diff
@@ -34,117 +33,105 @@ async function main() {
         process.exit(1);
     }
 
-    const packageData = await import(path.join(packageRoot, 'package.json'), {with: { type: 'json'}});
-    const packageName = packageData.name;
-    const srcDir = path.join(packageRoot, 'src');
-    const distDir = path.join(packageRoot, 'dist');
+    const packageData = await readPackageJSON(path.join(packageRoot, 'package.json'));
+    const isStimulusBundle = '@symfony/stimulus-bundle' === packageData.name;
+    const isReactOrVueOrSvelte = ['@symfony/ux-react', '@symfony/ux-vue', '@symfony/ux-svelte'].some(name => packageData.name.startsWith(name));
```
smnandre, line 38:
```suggestion
    const isReactOrVueOrSvelte
```

hahaha i may have use a... another const name

smnandre (approved):
The LiveComponent changes are a mess to review :)

bin/build_package.ts
Kocal: Yeah I wasn't really inspired here... 😅

PR #2946, opened 2025-07-27
[Translator] Add functional test to ensure only enabled locales are dumped

### LINKED ISSUE #1226: [Translator] Filter dumped locales
Currently, the translation dumper dumps all translations for all locales. However, in some projects, we have a list of locales we support and we don't care about others (which might still have keys defined for the third-party translations).

It would be great to filter the list of dumped translations based on the `kernel.enabled_locales` parameter defined in FrameworkBundle.

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Close #1226

### DIFF
--- a/src/Translator/tests/bootstrap.php
+++ b/src/Translator/tests/bootstrap.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\StringInput;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\UX\Translator\Tests\Kernel\FrameworkAppKernel;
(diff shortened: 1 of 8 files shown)


Kocal:
Failures related on lowest, it looks like the `TranslationsCacheWarmer` is not automatically called, when it is with highest deps 🤔

src/Translator/tests/bootstrap.php
```diff
@@ -0,0 +1,25 @@
...
+use Symfony\UX\Translator\Tests\Kernel\FrameworkAppKernel;
+
+require __DIR__.'/../vendor/autoload.php';
+
+(new Filesystem())->remove(__DIR__.'/../var');
+
+$kernel = new FrameworkAppKernel('test', true);
+$application = new Application($kernel);
+
+// Trigger Symfony Translator and UX Translator cache warmers
```
Kocal, line 24: I had some issues in lower tests, where our cache warmer was not called when running the functional test. 
It worked fine in higher tests thanks to https://github.com/symfony/symfony/pull/57553

But to make it easily consistent, we can manually trigger a cache clear here and the functional test will pass for every Symfony version

PR #2952, opened 2025-07-28
Fix package.json files to not use "catalog" feature from PNPM, as it breaks installation from `vendor/` PHP packages

### LINKED ISSUE #2951: npm error Unsupported URL Type "catalog:"
When updating from 2.27 to 2.28, and running `npm i --force`, I get the following output:

```
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "catalog:": catalog:
```

Downgrading to 2.27 fixes the issue.

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2951 
| License       | MIT

Since #2932 

Dropping any fancy stuff from our `package.json` since it can breaks when people install JS packages through the PHP packages 

I will edit the "test apps" jobs to use `npm` to install dependencies, so we will be sure in the future we are not breaking things again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,15 +14,15 @@
     },
     "devDependencies": {
         "@biomejs/biome": "^2.0.4",
-        "@testing-library/dom": "catalog:",
-        "@testing-library/jest-dom": "catalog:",
+        "@testing-library/dom": "^10.4.0",
+        "@testing-library/jest-dom": "^6.6.3",
         "@types/node": "^22.6.0",
         "lightningcss": "^1.28.2",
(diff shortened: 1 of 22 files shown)


kbond (approved):

Kocal:
I can confirm locally it fixes the issue

Kocal:
> I will edit the "test apps" jobs to use npm to install dependencies, so we will be sure in the future we are not breaking things again.

It requires a bit more work than expected, I will open a new PR later

PR #2953, opened 2025-07-28
[CI] Rework `app-tests` job to use `npm` as user's packages manager

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Following https://github.com/symfony/ux/pull/2952, this way we ensure our packages will always be installable with `npm` (related to https://github.com/symfony/ux/issues/2951)


Kocal:
> Post Run actions/setup-node@v4
> Post job cleanup.
> Error: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

Maybe because I removed `test_apps/encore-app/pnpm-lock.yaml`? 

W/e, CI is running fine now, `workspace:*` are now nicely replaced and `npm` is used as main packages manager to install dependencies inside app tests

**EDIT:** re-run the job in debug mode, it tried to cache path `/home/runner/.local/share/pnpm/store/v10` but it does not since we do not run `pnpm install` anymore. I simply removed the cache step, it is not necessary

PR #2955, opened 2025-07-29
[CI] Add Fabbot as a GitHub Actions

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

refs:
- https://github.com/symfony/symfony/pull/61260
- https://github.com/twigphp/Twig/pull/4671
- https://github.com/symfony-tools/fabbot

Since the code formatting process if more transparent, maybe we will remove the job `coding-style-php` that manually execute PHP-CS-Fixer (we had many formatting issues in the past)

### DIFF
--- a/.github/workflows/fabbot.yaml
+++ b/.github/workflows/fabbot.yaml
@@ -1,0 +1,15 @@
+name: Fabbot
+
+on:
+    pull_request:
+
+permissions:
+    contents: read
+
+jobs:
+    call-fabbot:
+        name: Fabbot
+        uses: symfony-tools/fabbot/.github/workflows/fabbot.yml@main
+        with:
+            package: Symfony UX
+            check_license: true
--- a/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
+++ b/src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php
@@ -23,7 +23,7 @@ use Symfony\Component\Form\FormEvents;
  *
  * @internal
  *
- * @deprecated since UX 2.13
+ * @deprecated since Symfony UX 2.13
  */
 final class AutocompleteEntityTypeSubscriber implements EventSubscriberInterface
 {
--- a/src/Autocomplete/src/Form/ChoiceList/Loader/ExtraLazyChoiceLoader.php
+++ b/src/Autocomplete/src/Form/ChoiceList/Loader/ExtraLazyChoiceLoader.php
@@ -18,7 +18,7 @@ use Symfony\Component\Form\ChoiceList\Loader\ChoiceLoaderInterface;
 /**
  * Loads choices on demand only.
  *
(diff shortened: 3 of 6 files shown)


Kocal:
@nicolas-grekas The step `Check PR header` is failing with error message `syntax error near unexpected token |'`, I think that's because the presence of `"` character in the `PR_BODY`

Maybe something like:
```shell
PR_TITLE=$(cat <<EOF
${{ github.event.pull_request.title }}
EOF
)
PR_BODY=$(cat <<EOF
{{ github.event.pull_request.body }}
EOF
)
```
?

.github/workflows/fabbot.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+permissions:
+    contents: read
+
+jobs:
+    call-fabbot:
+        name: Fabbot
+        uses: symfony-tools/fabbot/.github/workflows/fabbot.yml@main
+        with:
+            package: Symfony UX
```
nicolas-grekas, line 14:
```suggestion
            package: Symfony UX
            check_license: true
```

nicolas-grekas, line 14: `package: Symfony UX` this means `@deprecated` should all start with `@deprecated since Symfony UX` that's not the case at the moment, let's fix this also?

Kocal: yes we will need to fix this too, actually we have no conventions at all

nicolas-grekas: fabbot should be fixed
can you fix the `@deprecated` in the same PR?
they should all start with `@deprecated since Symfony UX X.y, etc`

.github/workflows/fabbot.yaml
```diff
@@ -0,0 +1,15 @@
+name: Fabbot
```
nicolas-grekas, line 1:
```suggestion
name: CS
```

.github/workflows/fabbot.yaml
nicolas-grekas: don't forget to push now that this is resolved ;)

.github/workflows/fabbot.yaml
Kocal: I would prefer keeping `Fabbot` or even `PHP CS` since we have other coding styles checks (for JS)

.github/workflows/fabbot.yaml
Kocal: Thanks, I will fix the deprecated annotations

nicolas-grekas (approved):

.github/workflows/fabbot.yaml
OskarStark: Same here, I am also used to tell people "Please make Fabbot happy" :-)

PR #2959, opened 2025-07-29
[Icons] Change `xmlns` attribute of downloaded icons to `http://www.w3.org/2000/svg`

### LINKED ISSUE #2958: [ux-icons] Imported icons have wrong xmlns value
When using `console ux:icons:import icon:name` (ux-icons v2.28.0) the resulting .svg file has "http**s**://www.w3.org/2000/svg" as the xmlns value instead of `http://www.w3.org/2000/svg`.

It looks like the xmlns value was added in #2661 but has a mistyped namespace (https instead of http).

The incorrect namespace causes browsers and image tools to not treat the file as an SVG image when being displayed.

roverwolf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2958
| License       | MIT

Change the xmlns attribute to be `http://www.w3.org/2000/svg` for icons imported using `console ux:icons:import` so that the icons will correctly render in browsers and can be viewed by other image tools.


smnandre:
I did not know that... https://github.com/w3c/svgwg/issues/738  👍 

@roverwolf  Could you also change the UX toolkit affected tests?

roverwolf:
@smnandre Toolkit snapshots have been updated in the PR.  Sorry, I had only checked the tests in the Icons component.

smnandre:
PHP 8.4 failures unrelated (autocomplete)

smnandre:
Thank you @roverwolf!

smnandre (approved):

Kocal (approved):

Kocal:
Interesting, I didn't know this was so subtile

Kocal:
Thank you @roverwolf.

PR #2967, opened 2025-08-01
[LiveComponent] Support JSONResponse for LiveActions

xDeSwa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #
| License       | MIT

## Summary

This PR improves the LiveComponent JavaScript request handling to correctly distinguish between standard HTML LiveComponent responses and pure JSON API responses from LiveAction methods.  
Specifically, it prevents error rendering in the frontend when a LiveAction returns a `JsonResponse` (such as for AJAX-like API interactions) by adding an additional check for the `application/json` content type in the response headers.

## Problem

When using LiveComponent actions that return `JsonResponse` (e.g., for API-like operations), the LiveComponent JS handler previously attempted to render an error if the response did not contain a LiveComponent HTML payload.  
This caused issues in cases where a component intentionally returns JSON (for example, on deletion or API-style actions), leading to a confusing user experience.

## Solution

- Updated the `performRequest` method in the LiveComponent JS handler to also check for `application/json` in the response's `Content-Type` header before triggering error rendering.
- If the response is JSON, the frontend will not try to render a LiveComponent error, enabling custom handling in Stimulus controllers or other client code.

#### Stimulus Controller

```javascript
import {Controller} from "@hotwired/stimulus";
import {getComponent} from '@symfony/ux-live-component';
import Swal from 'sweetalert2';

    static params = {id: Number}

    async initialize() {
    }

    postDelete(event) {
        let id = event.params.id;
        let component = this.component;

        if (id === undefined || id === 0) {
            return;
        }

        Swal.fire({
            title: "Confirmation Box",
            text: "Record will delete, are you sure?!",
            icon: "warning",
            showCancelButton: !0,
            confirmButtonColor: "#28bb4b",
            cancelButtonColor: "#f34e4e",
            confirmButtonText: "Yes, Im Sure",
            cancelButtonText: "Cancel",
        }).then(async (e) => {
            if (e.isConfirmed) {
                try {
                    const result = await component.action('deletePost', {id: id});
                    const response = JSON.parse(result.body);

                    if (response.status === "success") {
                        Swal.fire("Success", response.message, "success");
                    } else {
                        Swal.fire("Error", response.message, "error");
                    }
                } catch (err) {
                    Swal.fire("Error", err, "error");
                }
                this.component.render();
            }
        });
    }
}
```

#### Twig Example

```twig
<div {{ attributes.defaults(stimulus_controller('postactions')) }}>
    <button data-postactions-id-param="{{ post.id }}"
            data-action="click->postactions#postDelete"
            class="btn btn-danger btn-xs d-flex align-items-center justify-content-center"
            title="Delete">
        Delete
    </button>
</div>
```

#### LiveComponent Example

```php
// src/Twig/Components/Posts.php

// ...

class Posts
{
    // ...

    #[LiveAction]
    public function deletePost(#[LiveArg] int $id): JsonResponse
    {
        $post = $this->postRepository->find($id);
        if ($post) {
            try {
                $this->entityManager->remove($post);
                $this->entityManager->flush();
            } catch (Exception) {
                return new JsonResponse(array("status" => "error", "message" => "Post delete failed."));
            }
            return new JsonResponse(array("status" => "success", "message" => "Post deleted successfully."));
        } else {
            return new JsonResponse(array("status" => "error", "message" => "Error: Post was not found."));
        }
    }

    // ...
}
```

### DIFF
--- a/src/LiveComponent/assets/src/Component/index.ts
+++ b/src/LiveComponent/assets/src/Component/index.ts
@@ -317,7 +317,7 @@ export default class Component {
                 this.valueStore.pushPendingPropsBackToDirty();
                 this.hooks.triggerHook('response:error', backendResponse, controls);
 
-                if (controls.displayError) {
+                if (controls.displayError && !headers.get('Content-Type')?.includes('application/json')) {
                     this.renderError(html);
                 }
 
--- a/src/LiveComponent/assets/src/Component/index.ts
+++ b/src/LiveComponent/assets/src/Component/index.ts
@@ -20,6 +20,11 @@ export type ComponentHooks = {
     connect: (component: Component) => MaybePromise;
     disconnect: (component: Component) => MaybePromise;
     'request:started': (requestConfig: any) => MaybePromise;
+    'render:started': (
+        backendResponseBody: string,
+        backendResponse: BackendResponse,
+        controls: { shouldRender: boolean }
+    ) => MaybePromise;
     'render:finished': (component: Component) => MaybePromise;
     'response:error': (backendResponse: BackendResponse, controls: { displayError: boolean }) => MaybePromise;
     'loading.state:started': (element: HTMLElement, request: BackendRequest) => MaybePromise;
@@ -300,34 +305,43 @@ export default class Component {
 
         this.backendRequest.promise.then(async (response) => {
             const backendResponse = new BackendResponse(response);
-            const html = await backendResponse.getBody();
+            const result = await backendResponse.checkResponseType();
 
-            // clear sent files inputs
+            if (result.type === 'json') {
+                this.backendRequest = null;
+                thisPromiseResolve(backendResponse);
+
+                if (this.isRequestPending) {
+                    this.isRequestPending = false;
+                    this.performRequest();
+                }
+                return response;
+            }
+
+            // Clear all file inputs
             for (const input of Object.values(this.pendingFiles)) {
                 input.value = '';
             }
 
+            const backendResponseBody = result.body;
+
             // if the response does not contain a component, render as an error
-            const headers = backendResponse.response.headers;
-            if (
-                !headers.get('Content-Type')?.includes('application/vnd.live-component+html') &&
-                !headers.get('X-Live-Redirect')
-            ) {
+            if (result.type === 'invalid') {
                 const controls = { displayError: true };
                 this.valueStore.pushPendingPropsBackToDirty();
                 this.hooks.triggerHook('response:error', backendResponse, controls);
 
-                if (controls.displayError && !headers.get('Content-Type')?.includes('application/json')) {
-                    this.renderError(html);
+                if (controls.displayError) {
+                    this.renderError(backendResponseBody);
                 }
 
                 this.backendRequest = null;
                 thisPromiseResolve(backendResponse);
-
                 return response;
             }
 
-            this.processRerender(html, backendResponse);
+            // HTML processing
+            this.processRerender(backendResponseBody, backendResponse);
             const liveUrl = backendResponse.getLiveUrl();
             if (liveUrl) {
                 history.replaceState(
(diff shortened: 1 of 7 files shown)


src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -317,7 +317,7 @@ export default class Component {
                 this.valueStore.pushPendingPropsBackToDirty();
                 this.hooks.triggerHook('response:error', backendResponse, controls);
 
-                if (controls.displayError) {
+                if (controls.displayError && !headers.get('Content-Type')?.includes('application/json')) {
```
Kocal: Checking content-type for JSON here is not enough, IMO you should also check it in the `if` condition lines 321-314, IMO we don't want to trigger anything in this block if that's JSON.

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -317,7 +317,7 @@ export default class Component {
...
                     this.renderError(html);
```
Kocal: Would be nice to rename this `html` to `backendResponseBody`, since it can contains something else than HTML, which implies to tweak method `processRerender`:
- rename the argument name of hook `render:started`, won't be a BC but the documentation must be updated
- since we returns JSON and not HTML, all the code about HTML morphing, mutation tracking, etc.. is useless (and broken). We should do that ONLY if we we have HTML

src/LiveComponent/assets/src/Component/index.ts
Kocal: To follow-up, we are doomed if we enter this `if()`, because of the early `return response`. 
It means that the next code about processing the response, trigger hooks, update the URL, etc... won't be executed anymore.

Kocal (changes requested):
Thanks for the suggestion!

Here some majors comments that must be addressed, the actual implementation is partially broken and must be fixed.

Also, please ensure to add tests PHP-side and JS-side, as we want to be sure to not break anything.
Could you also add documentation about returning a custom `Response` and `JsonResponse` as well?

Thanks!

smnandre:
After some time to think, here are my « at this moment » thoughts
- return nothing is something we already discussed and planned but for 3.0 as it would be a major change
- same for délétion, with custom LiveResponse helper and we would implement the délétion and events here, all this features must be easy to use for every users, especially for people not wanting to do additionnal JS
- LiveComponent are not made for API calls and i’m not sure to understand the need here, or the real life use case
- the core concept of LiveComponent is to render HTML and deal with state in DOM with morphing and events, I’m Not sure to see JSON as an additionnal feature but mainly I dont think we offer any interest for people if they want to do all in JS themselves.. over any exposed restish API 

as much I understand the précise need you are dealing with right now, I am really Not sure this is direction we should go to.

LiveComponent is already complex for many, and if I agree we need to open more extension point, they should be only that and we cannot make promise or deal with BC afterward for them implémentation. 

On the contrary I think — and i join you here — we must adress more or the « basic / regular » needs people using LiveComponent have. Including but not limited to component lifecycle (delete, update, refresh), identification, partial rendering or no render, locking or sleep, etc)

But for all this we need, imho, to stop stacking features and more keep a global / easy to understand philosophy here that would ease both maintenance and new feature dev, improve code quality testability and security.

all of this is not directly related to this PR but it is something I wanted to write down as a block for some time, in order to give clear vision of my personal state of mind here 

Again: this is my own personal opinion right now and am more than open to any feedback / critique or exchange on it.

xDeSwa:
It would be great if we could learn about the features planned for version 3.0.

For example, I'm considering implementing something similar to what Livewire provides — like custom error handlers. Here's a use case:

``` javascript
window.LiveComponent.onError = (response, statusCode, component) => {
    if (statusCode === 500) {
        console.error('Server error in component:', component.name);
        return true; // continue with default error handling
    }

    return true;
};
```

Another example relates to dynamic behavior based on responses. While it's possible to attach a Stimulus controller to a component's root element, sometimes we need to react programmatically to the result of a LiveAction — and based on that result, decide whether to re-render the component or not.

In such cases, it would be very useful to have the ability to disable rendering from within a JsonResponse inside a LiveAction method, allowing us to use the returned data on the frontend without triggering a component re-render.

Kocal:

Thanks!

Status: Needs Work

PR #2988, opened 2025-08-12
[LiveComponent] Fix new URL calculation with `LiveProp` using Serializer (and Serializer attributes), and when using custom modifier

### LINKED ISSUE #2971: [LiveComponent] ignored `#[SerializedName(serializedName: 'p')]` after 2.28 upgrade
After upgrading from 2.27 to 2.28.0, the URL mapping of a LiveComponent DTO using the serializer (`useSerializerForHydration`) is not taken into account anymore. I downgraded to fix it for now.

Potentially this could be related:
 * https://github.com/symfony/ux/pull/2673

DTO example:

```php
final class Something
{
    public function __construct(
        #[SerializedName(serializedName: 't')]
        public Test $test,
   ) {}
}
```

Usage:

```php
    #[LiveProp(useSerializerForHydration: true, updateFromParent: true, url: new UrlMapping('e'))]
    public Something $something;
```

URL that gets updated in the browser (Javascript part).

Before:
```
/some/url?e%5Bt%5D%5B...
```

After
```
/some/url?e%5Btest%5D%5B
```

The problem is that when these new URLs are sent to someone else or if someone refreshes, no values can be found as the deserializer looks for `t` but the url has `test`.

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2971, #2991 
| License       | MIT

This PR fixes two bugs introduced #2673:
1. See https://github.com/symfony/ux/issues/2971#issuecomment-3176961430 for explanations
tl;dr: previously, query parameters were updated through `LiveComponentHydrator::dehydrate()` which use the Serializer (if `useSerializerForHydration: true`) and so also used Serializer attributes like `SerializedName`.
2. https://github.com/symfony/ux/issues/2991, when having a `LiveProp(url: new UrlMapping(as: 'foo'))` with a custom modifier that returns a new `LiveProp` with a new `UrlMapping`, ex:
```php
    public function myModifier(LiveProp $liveProp): LiveProp
    {
        $urlMapping = $liveProp->url();
        if (!$urlMapping instanceof UrlMapping) {
            return $liveProp;
        }

        return $liveProp->withUrl(new UrlMapping(as: 'alias_' . $urlMapping->as));
    }
```
then the prop was persisted as `foo` and not `alias_foo` in the URL.


src/LiveComponent/tests/Unit/EventListener/LiveUrlSubscriberTest.php
Kocal, line 1: I don't think this test was adding value, legit everything was mocked and it could not reflect the real behavior.

Instead I went for functional test

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -66,26 +73,26 @@ public static function getSubscribedEvents(): array
...
-            $liveRequestData['responseProps'] ?? []
+        $metadata = $this->metadataFactory->getMetadata($mounted->getName());
+
+        $dehydratedProps = $this->liveComponentHydrator->dehydrate(
+            $mounted->getComponent(),
+            $mounted->getAttributes(),
+            $metadata
         );
 
+        $values = $dehydratedProps->getProps();
```
Kocal, line 86: This is the fix for the first bug

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -1064,12 +1064,43 @@ public function mount()
                 });
         }];
 
+        yield 'Array with DTOs: using serializer, fully writable allows partial changes' => [function () {
```
Kocal, line 1068: Since I added `#[SerializedName(serializedName: 'c')]` on `Address::$city`, our tests failed because we expected props to be dehydrated to `city` instead of `c`.

I duplicated these failing tests and changed the dehydrated property to `c`, and modified the previous tests to not use `useSerializerForHydration`.

Kocal:
cc @mbuliard, can you take a look? Thanks!

src/LiveComponent/tests/Functional/EventListener/LiveUrlSubscriberTest.php
```diff
@@ -45,52 +49,108 @@ public function getTestData(): iterable
...
+                'propName' => 'boundPropWithAlias',
+                'propValue' => 'search',
+            ],
+        ];
+        $address = new Address();
+        $address->address = '123 Main St';
+        $address->city = 'Anytown';
+        yield 'with an object in query, with "useSerializerForHydration: true", keys "address" and "c" must be present' => [
+            'previousLocation' => '/',
+            'expectedLocation' => '/?intProp=3&objectPropWithSerializerForHydration%5Baddress%5D=123+Main+St&objectPropWithSerializerForHydration%5Bc%5D=Anytown',
```
Kocal, line 121: That's not easy to read, but the `c` is present at `5Bc%5D`

Kocal: Btw it would be nice if we could have a "clean" query, like it was done before.

I tried to play with `http_build_query()`'s encoding, but it didn't help

Kocal:
I pushed a second commit to fix https://github.com/symfony/ux/issues/2991, now `modifier` are nicely used to generate the URL

src/LiveComponent/src/Metadata/LiveComponentMetadata.php
```diff
@@ -69,10 +69,10 @@ public function getOnlyPropsThatAcceptUpdatesFromParent(array $inputProps): arra
     /**
      * @return UrlMapping[]
      */
-    public function getAllUrlMappings(): iterable
+    public function getAllUrlMappings(object $component): iterable
     {
         $urlMappings = [];
-        foreach ($this->livePropsMetadata as $livePropMetadata) {
+        foreach ($this->getAllLivePropsMetadata($component) as $livePropMetadata) {
```
Kocal, line 75: This is the fix for the 2nd bug

ruudk:
@Kocal Thanks for the fix 💙 

Would you mind tagging it so we can test it out?

Kocal:
I asked Kevin for a release but I'm not sure when it will happen, but in the meantime I think you can use 2.x-dev constraint version 🙂

Kocal:
@ruudk released in 2.29.2

PR #2997, opened 2025-08-13
[CI] Parallelize Windows Unit tests

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT


Kocal:
Opened https://github.com/aaronriekenberg/rust-parallel/issues/50, I don't find a way to force rust-parallel to use bash on Windows without getting issues.

As alternative, maybe we will have to run multiples jobs in parallel (one per UX package), like before, but only for Windows, sadly

rust-parallel.zip
Kocal, line 1: (to remove)

PR #3010, opened 2025-08-16
 Configure Vitest for unit and browser tests (use @puppeteer/browsers and webdriverio)

### LINKED ISSUE #3008: Configure Vitest for unit and browser tests

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #3008 
| License       | MIT

Moved all existing assets unit tests under `assets/test/unit`, in order to introduce `assets/test/browser` for pure E2E tests using a real Symfony app (in a future PR) and real browsers.

I slightly improved the `test_package.sh` for peerDependencies/importmap versions checks.

Next steps are [#3009](https://github.com/symfony/ux/issues/3009)


.github/workflows/unit-tests.yaml
```diff
@@ -143,21 +143,16 @@ jobs:
                 fi
     js:
         runs-on: ubuntu-latest
-        strategy:
-            fail-fast: false
-            matrix:
-                node-version: [ '18.x', '20.x', '22.x', '24.x' ]
```
Kocal, line 149: There is no need to unit tests in multiple Node.js versions, our `assets/` code is supposed to run in a browser. We are testing the strict necessary

bin/test_package.sh
Kocal, line 1: In the future, it would be nice to shift the 1st argument (package location) and pass all the other arsg to `pnpm exec vitest`

.nvmrc
```diff
@@ -1 +1 @@
-22.11
+22.18
```
Kocal, line 1: So I can use `import.meta.main` (like `__name__ == '__main__'` in Python)

Kocal:
🤦🏻 thought you could use Vitest for "real" E2E testing like, visiting a page, but not at all. 

I will revert some part of this PR and move to raw WebdriverIO, Playwright, or Cypress

PR #3031, opened 2025-08-20
[LiveComponent] Return empty string for `data-value=""` instead of falling back to null

mercuryseries:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | # 
| License       | MIT

## Problem

<details>
    <summary><b>Demo video</b></summary>

https://github.com/user-attachments/assets/31bba6af-8690-4cd9-a5c4-20a561342626
</details>

```html
<input type="text" data-model="post.title">
<button
    type="button"
    data-model="post.title"
    data-value="" <!-- HERE -->
    data-action="live#update"
>
    Clear Title
</button>
```

When using `data-value=""` on an element, the current implementation incorrectly returns `null` instead of the intended empty string value. This happens because the condition `if (element.dataset.value)` evaluates to `false` for empty strings (since empty strings are falsy in JavaScript), causing the function to fall through to the final `return null;` statement.

## Impact

This behavior prevents developers from explicitly setting empty string values via the `data-value` attribute, which is a common use case for clearing form fields or resetting values to empty states rather than `null`.

### Current behavior:

```html
<button data-model="title" data-value="" data-action="live#update">Clear</button>
<!-- Results in: null -->
```

### Expected behavior:

```html
<button data-model="title" data-value="" data-action="live#update">Clear</button>  
<!-- Should result in: "" -->
```

<details>
    <summary><b>After (demo video)</b></summary>

https://github.com/user-attachments/assets/bad5f5f3-1104-4102-a2bc-30264bb3c5bf
</details>

## Changes

Replace the truthiness check `if (element.dataset.value)` with an explicit attribute existence check `if (element.hasAttribute("data-value"))`. This ensures that any `data-value` attribute, including those with empty string values, are properly processed and returned.

## Manual testing done
- ✅ `data-value=""` now returns `""`
- ✅ `data-value="test"` still returns `"test"`
- ✅ Missing `data-value` attribute still falls back to other value sources
- ✅ All existing functionality remains intact

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -377,7 +377,7 @@ function getValueFromElement(element, valueStore) {
     }
     return element.value;
   }
-  if (element.dataset.value) {
+  if (element.hasAttribute("data-value")) {
     return element.dataset.value;
   }
   if ("value" in element) {
--- a/src/LiveComponent/assets/src/dom_utils.ts
+++ b/src/LiveComponent/assets/src/dom_utils.ts
@@ -51,7 +51,7 @@ export function getValueFromElement(element: HTMLElement, valueStore: ValueStore
     }
 
     // element is some other element
-    if (element.dataset.value) {
+    if (element.hasAttribute('data-value')) {
         return element.dataset.value;
     }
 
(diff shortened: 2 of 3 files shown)


smnandre:
👍 

It would indeed be consistent with Stimulus values https://stimulus.hotwired.dev/reference/values#getters

Could you add some tests to prevent any future regression ?

mercuryseries:
@smnandre I added some tests. I see that some tests failed, but seems not related with the changes. Can you confirm please?

smnandre:
It does not feel related to me.. @Kocal  wdyt ?

smnandre (approved):

Kocal:
No that's not related, it should be green after a rebase

Kocal:
Thank you @mercuryseries.

PR #3033, opened 2025-08-20
[Map][Leaflet] Fix InfoWindow auto-opening

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

While writing E2E tests for UX Map, I noticed that I wasn't able to automatically open an InfoWindow (Popup) when using Leaflet, with `opened: true`


Kocal:
wrong branch

PR #3036, opened 2025-08-21
[Map] Deprecate option `title` from `Polygon`, `Polyline`, `Rectangle` and `Circle` in favor of `infoWindow`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

While writing E2E tests for Map, I found that our `title` option from `Polygon`, `Polyline`, `Rectangle` and `Circle` was wrongly used, and does not make sense.

Using `title` won't add an HTML attribute `title` as I expected, it's not something supported by Google nor Leaflet.
Instead, we used it to display a popup, like the `infoWindow` (which is more complete).

I suggest deprecating `title` from 2.x and remove it in 3.0.


src/Map/assets/src/abstract_map_controller.ts
```diff
@@ -65,7 +65,7 @@ export type MarkerDefinition<BridgeMarkerOptions, BridgeInfoWindowOptions> = Wit
     infoWindow?: Omit<InfoWindowDefinition<BridgeInfoWindowOptions>, 'position'>;
     icon?: Icon;
     /**
-     * @deprecated Use "bridgeOptions" instead.
+     * @deprecated since Symfony UX Map 2.27, use "bridgeOptions" instead.
```
Kocal, line 68: I will open a PR on fabbot to valide the `@deprecated since ...` format on JS/TS files too.

PR #3038, opened 2025-08-21
[Map] Add E2E tests, close #3022

### LINKED ISSUE #3022: Add E2E tests for Map

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #3022 
| License       | MIT

I only tested Leaflet map with various features, including custom icons.

Google Maps won't be tested as it requires a **paid** API Key.


test_apps/e2e-app/importmap.php
```diff
@@ -85,6 +85,9 @@
     'react-dom' => [
         'version' => '18.3.1',
     ],
+    'react-dom/client' => [
+        'version' => '18.3.1',
+    ],
```
Kocal, line 90: Related to https://github.com/symfony/ux/pull/3030

src/Map/src/Bridge/Google/assets/test/browser/map.test.ts
```diff
@@ -0,0 +1,13 @@
+import { expect, test } from '@playwright/test';
+
+test('Can render basic map', async ({ page }) => {
+    await page.goto('/ux-map/basic?renderer=google');
+
+    await expect(await page.getByTestId('map')).toBeVisible();
+
+    // Since we can't test Google Maps rendering due to API costs, we only assert that Google Maps API is (wrongly) loaded
```
smnandre, line 8: I would not do this test.

i think this could be a (tiny but still very real) problem in term of privacy, as we are making requests on behalf of the user... (even  if it's on its own local machine)..

If someone want Google Maps fully tested, they probably _already_ pay for GoogleMaps ..... 
so maybe they can invest some coins around here to support the work.... 

In the meantime I clearly would not spend too much time or energy to test a (very) costly (and "not-always" privacy-compliant) solution your our own time and/or money 😅

test/playwright-helpers.ts
```diff
@@ -0,0 +1,16 @@
+import * as path from 'node:path';
+import * as fs from 'fs';
+
+export function getSymfonyKernelVersionId(): number {
+    const kernelPath = path.join(import.meta.dirname, '../test_apps/e2e-app/vendor/symfony/http-kernel/Kernel.php');
```
smnandre, line 5: Would use an ENV_VAR here (as its done for KERNEL_CLASS in a way)

smnandre (approved):
👏

src/Map/src/Bridge/Google/assets/test/browser/map.test.ts
Kocal: Yeah I see what you mean, but since https://github.com/symfony/ux/pull/2146 Map Bridges tests were run inside a browser and so requests were already made to Google. 

I think that's fine if we make it opt-out explicitly:
- `pnpm run test:browser`: run browsers tests on all UX packages with `test:browser` pnpm script **except** UX Google Map
- `pnpm run test:browser:no-privacy` (or something similar): run browsers tests on all UX packages with `test:browser` pnpm script **including** UX Google Map

> If someone want Google Maps fully tested, they probably already pay for GoogleMaps .....
so maybe they can invest some coins around here to support the work....

About writing real tests for Google Map Bridge too, I thought about putting my own API key as a GitHub secret (so no one can see it), but if someone decide to trigger 10k times the CI, I'm bankrupt 🤯 

Running a test that ensure the Google Maps API is loaded is the best of both worlds imo

test/playwright-helpers.ts
Kocal: I thought about it too but re-using `SYMFONY_REQUIRE` env var would makes things counterproductive IMHO:
- it's not very intuitive to pass this env var while running Playwright, I'm sure everyone including me will forgot it
- if you change your Symfony version through `SYMFONY_REQUIRE=x.y.* composer update`,  you must close Playwright and re-open Playwright with `SYMFONY_REQUIRE=x.y.*` 
- there is no native `version_compare()` function in Node.js, we must re-implement the logic ourselves with https://github.com/npm/node-semver for example

The `VERSION_ID` from `test_apps/e2e-app/vendor/symfony/http-kernel/Kernel.php` is the best source of truth we can get when being inside Playwright context

PR #3042, opened 2025-08-22
[Icons] Fallback to default dimensions 16/16 when Icon's dimensions on Iconify are missing

### LINKED ISSUE #3016: [Icons] some pajamas and oui icons throw no height & no width error even when explicitly specified
Hi ! First of all, I don't know whether this should be raised here in Symfony UX or directly to Iconify API but I just thought you ought to know anyway.

When perusing the Symfony UX icons bank website (big fan btw !), I wanted to use the **pajamas:false-positive** icon in my project but I bumped into the following exception : 

> The icon "pajamas:false-positive" does not have a width or height.

This happens both when rendering via : 

```twig
{{ ux_icon('pajamas:false-positive') }}
```

or via : 

```twig
<twig:ux:icon name="pajamas:false-positive" />
```

or even when specifying width & length as the [documentation](https://symfony.com/bundles/ux-icons/current/index.html#icon-sizes) recommends to do: 

```twig
<twig:ux:icon name="pajamas:false-positive" style="height: 1em; width: 1em;" />
```

Which feels quite strange as when explicitly defining height & width on other icons it works just fine (I've tested for instance with this and it works:

```twig
      {{ ux_icon('eos-icons:hourglass', {style: 'height: 2em; width: 2em;'}) }}
```

However, no problem occurs when using the referenced SVG as viewBox, height and width are clearly specified : 

```html

<svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 16 16"><path fill="currentColor" fill-rule="evenodd" d="M6.229.199a8 8 0 0 1 9.727 6.964a.75.75 0 0 1-1.492.157a6.5 6.5 0 1 0-7.132 7.146a.75.75 0 1 1-.154 1.492a8 8 0 0 1-.95-15.76Zm5.051 10.02a.75.75 0 1 0-1.06 1.061L11.94 13l-1.72 1.72a.75.75 0 1 0 1.06 1.06L13 14.06l1.72 1.72a.75.75 0 1 0 1.06-1.06L14.06 13l1.72-1.72a.75.75 0 1 0-1.06-1.06L13 11.94zm.25-3.939a.75.75 0 0 0-1.06-1.06L6.5 9.19L5.28 7.97a.75.75 0 0 0-1.06 1.06l1.75 1.75a.75.75 0 0 0 1.06 0z" clip-rule="evenodd"/></svg>
```

I haven't tested them all but it seems to happen with all the icon sets where the width & height aren't specified (only body property) neither at the whole file level (as I can see some other sets have a  `"height": 24, "category": "UI 24px",` definition at file level) nor at the unitary icon level. I assume when these properties are missing, UX Icons tries to set a default size but fails to in this case ? 

https://github.com/iconify/icon-sets/blob/53c7b77fa3de866110bd1a5162b14b14746e3675/json/pajamas.json#L434

https://github.com/iconify/icon-sets/blob/master/json/oui.json

While not the same issue, I can see width and height are mentioned in this other issue : https://github.com/symfony/ux/issues/2629

Obviously, for my own personal case here, I can simply use the SVG version but I still think it was worth mentioning these inconsistencies of rendering when using `ux_icon()` and `<twig:ux:icon>`, especially as the explicit style rendering seems to make no difference whatsoever, I haven't dug too deep into this issue, only tested it on my Sylius Stack project and on brand new install of Symfony, so please don't hesitate to tell me if I'm missing something obvious here.

stlgaits:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no 
| Issues        | Fix #3016 
| License       | MIT

Set default viewBox size to 0 0 16 16 when not specified in Iconify icon set as per their documentation specifies : https://iconify.design/docs/types/iconify-json.html#icon to avoid exception being thrown when rendering via ux_icon() or <twig:ux:icon />

### DIFF
--- a/src/Icons/src/Iconify.php
+++ b/src/Icons/src/Iconify.php
@@ -34,6 +34,10 @@ final class Iconify
     // -safe margin
     private const MAX_ICONS_QUERY_LENGTH = 400;
 
+    // https://github.com/iconify/iconify/blob/00cc144b040b838bd86474ab83f0e50e6c6a12a1/packages/utils/src/icon/defaults.ts#L23-L30
+    private const DEFAULT_ICON_WIDTH = 16;
+    private const DEFAULT_ICON_HEIGHT = 16;
+
     private HttpClientInterface $http;
     private \ArrayObject $sets;
     private int $maxIconsQueryLength;
@@ -81,13 +85,10 @@ final class Iconify
 
         $height = $data['icons'][$name]['height'] ?? $data['height'] ?? $this->sets()[$prefix]['height'] ?? null;
         $width = $data['icons'][$name]['width'] ?? $data['width'] ?? $this->sets()[$prefix]['width'] ?? null;
-        if (null === $width && null === $height) {
-            throw new \RuntimeException(\sprintf('The icon "%s:%s" does not have a width or height.', $prefix, $nameArg));
-        }
 
         return new Icon($data['icons'][$name]['body'], [
             'xmlns' => self::ATTR_XMLNS_URL,
-            'viewBox' => \sprintf('0 0 %s %s', $width ?? $height, $height ?? $width),
+            'viewBox' => \sprintf('0 0 %s %s', $width ?? $height ?? self::DEFAULT_ICON_WIDTH, $height ?? $width ?? self::DEFAULT_ICON_HEIGHT),
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


smnandre:
I do agree with your messages (here and in the og issue), less so for the "bug" term but nervermind  😄 

Thank you for taking the time to open a PR !

smnandre (approved):

Kocal:
I just did some modifications for adding private const and to ensure a default value is set even if only one of the width or height is null

Kocal:
Thanks @stlgaits.

PR #3043, opened 2025-08-22
[LiveComponent] Fix BC Break when defining `LiveProp` properties with PHPDoc type and no-native type, when using TypeInfo

### LINKED ISSUE #2963: [LiveComponent] BC Break with v2.28
Suddenly all components with live forms and DTO's and without real types started to throw errors.

Like:
```
An exception has been thrown during the rendering of a template ("The "foo" property on component "App\Twig\Component" is missing its property-type. Add the "App\Enum" type so the object can be hydrated later.") in component.html.twig at line 13.
```

Previously everything was working fine.
DTO's has phpdoc typehints, so probably they are not read anymore and instead `Type::mixed()` is used.

norkunas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2963
| License       | MIT

### DIFF
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -13,6 +13,7 @@ namespace Symfony\UX\LiveComponent\Metadata;
 
 use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
 use Symfony\Component\PropertyInfo\Type as LegacyType;
+use Symfony\Component\TypeInfo\Exception\UnsupportedException;
 use Symfony\Component\TypeInfo\Type;
 use Symfony\Component\TypeInfo\Type\CollectionType;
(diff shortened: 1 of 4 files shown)


Kocal:
Hi,

Thanks for working on this, could you add a test case and rebase your PR on latest 2.x to make the CI green? Thanks!

norkunas:
> Hi,
> 
> Thanks for working on this, could you add a test case and rebase your PR on latest 2.x to make the CI green? Thanks!

Testcase added

Kocal:
Thanks for providing a test, I was finally able to reproduce your issue and observe that indeed a small BC was introduced while working on https://github.com/symfony/ux/commit/de302ef36c4f67f1c2e1b378aa31a7b2e41ad5a5

Kocal (approved):

Kocal:
Thank you @norkunas for taking care of this regression.

PR #3045, opened 2025-08-22
 Ensure PHP 8.5 compatibility

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #... 
| License       | MIT


src/Map/src/Bridge/Google/tests/GoogleRendererTest.php
```diff
@@ -71,7 +71,7 @@ public static function provideTestRenderMap(): iterable
                 ->addMarker($marker1)
                 ->addMarker($marker2)
                 ->removeMarker($marker1)
-                ->removeMarker($marker2),
+                ->removeMarker('marker2'),
```
Kocal, line 74: AFAIK we didn't have coverage for `remove(string)`

Issue #2975, opened 2025-08-05
Upgrade to 2.28 breaks live-props as url


cklm:
Hi guys, I found a problem when upgrading `symfony/ux-live-component` from 2.27 to 2.28: I have a live-prop with url true, like this:

    #[LiveProp(writable: true, url: true)]
    public int $page=1;

In 2.27 the page-parameter is in the url and gets updated when the property changes. In 2.28 the page-parameter isn't added to the url, even when the value changes. Maybe it is related to the change in UrlMapping.

Kocal:
Hi, can you try patch #2976?

Issue #2960, opened 2025-07-29
[LiveComponent] 500 errors with MethodNotAllowedException on dev after 2.28 upgrade


jmsche:
Hello,

After upgrading UX Live component from 2.27 to 2.28.1 (but the bug also happens on 2.28.0), I'm getting the following error when I toggle some checkboxes that are marked as `#[LiveProp(writable: true)]` in the related Live component.

After having a little look at the merged PRs between the releases, I guess this is the one making it fail now: https://github.com/symfony/ux/pull/2673

Here's the stack trace:

```
Symfony\Component\Routing\Exception\MethodNotAllowedException:

  at vendor/symfony/routing/Matcher/Dumper/CompiledUrlMatcherTrait.php:42
  at Symfony\Component\Routing\Matcher\CompiledUrlMatcher->match()
     (vendor/symfony/routing/Router.php:177)
  at Symfony\Component\Routing\Router->match()
     (vendor/symfony/ux-live-component/src/Util/UrlFactory.php:63)
  at Symfony\UX\LiveComponent\Util\UrlFactory->createPath()
     (vendor/symfony/ux-live-component/src/Util/UrlFactory.php:45)
  at Symfony\UX\LiveComponent\Util\UrlFactory->createFromPreviousAndProps()
     (vendor/symfony/ux-live-component/src/EventListener/LiveUrlSubscriber.php:48)
  at Symfony\UX\LiveComponent\EventListener\LiveUrlSubscriber->onKernelResponse()
     (vendor/symfony/event-dispatcher/Debug/WrappedListener.php:115)
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke()
     (vendor/symfony/event-dispatcher/EventDispatcher.php:206)
  at Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
     (vendor/symfony/event-dispatcher/EventDispatcher.php:56)
  at Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
     (vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php:126)
  at Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch()
     (vendor/symfony/http-kernel/HttpKernel.php:216)
  at Symfony\Component\HttpKernel\HttpKernel->filterResponse()
     (vendor/symfony/http-kernel/HttpKernel.php:204)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
[...]

PR #3053, opened 2025-08-24
[LiveComponent] Refactor and fix some edgecases on new URL generation with path + query LiveProps

### LINKED ISSUE #2987: [LiveComponent] urlFactory with dynamic routing not working >= 2.28.0
after upgrading to ux-live-component > 2.27.0 I'm getting this error. "Please provide a `path` parameters". It happens when I'm doing $this->redirect('/'); on a live component that extends AbstractController that is on SonataPageBundle front page. It works on 2.27.0

I asked GTP-5 and it gave me this. It explains what is happening and It said that "If you need dynamic slug root and want URL mapping future-proof you should use this"
```
<?php

declare(strict_types=1);

namespace App\LiveComponent;

use Symfony\UX\LiveComponent\Util\UrlFactory;
use Symfony\Component\Routing\RouterInterface;
use Symfony\Component\Routing\Exception\ResourceNotFoundException;
use Symfony\Component\Routing\Exception\MethodNotAllowedException;
use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;

/**
 * Decorator that preserves original matched route parameters (e.g. Sonata PageBundle's
 * required "path" parameter for the dynamic CMS route) when LiveComponents rewrites URLs.
 *
 * The stock UrlFactory matches the previous URL path, extracts only the route name and
 * regenerates with JUST the path-mapped LiveProps. If the underlying route has *required*
 * parameters that were not mapped (like Sonata's page_slug "path"), generation fails.
 *
 * This subclass re-matches the previous path, captures the original parameters, merges
 * them back in (letting path-mapped LiveProps override), then proceeds with query merging.
 */
class PreservingUrlFactory extends UrlFactory
{
    public function __construct(private readonly RouterInterface $router)
    {
        parent::__construct($router);
    }

    /**
     * @param string $previousUrl      The full previous URL (path + optional query)
     * @param array<string,mixed> $pathMappedProps  LiveProps mapped to the path (mapPath = true)
     * @param array<string,mixed> $queryMappedProps LiveProps mapped to the query (default)
     */
    public function createFromPreviousAndProps(
        string $previousUrl,
        array $pathMappedProps,
        array $queryMappedProps,
    ): ?string {
        $parsed = parse_url($previousUrl);
        if ($parsed === false) {
            return null;
        }

        $previousPath = $parsed["path"] ?? "";
        $previousQueryRaw = $parsed["query"] ?? "";

        // Re-match to extract original parameters (needed for routes like Sonata's page_slug).
        [$routeName, $originalParams] = $this->matchRouteAndExtractParams(
            $previousPath,
        );
        if ($routeName === "") {
            return null;
        }

        // Merge original params with new path-mapped props (LiveProps override originals).
        $routeParams = array_merge($originalParams, $pathMappedProps);

        try {
            $newPath = $this->router->generate($routeName, $routeParams);
        } catch (ResourceNotFoundException | MethodNotAllowedException | MissingMandatoryParametersException) {
            return null;
        } catch (\Throwable) {
            return null;
        }

        // Previous query parameters
        $previousQueryParams = $this->parseQueryString($previousQueryRaw);
        // Remnant parameters the router injected into the query string
        $remnantQueryParams = $this->extractQueryParamsFromUrl($newPath);

        // Final query params (later sources override earlier)
        $finalQueryParams = array_merge(
            $previousQueryParams,
            $remnantQueryParams,
            $queryMappedProps,
        );

        return $this->replaceQueryString($newPath, $finalQueryParams);
    }

    /**
     * Match the path and return [routeName, paramsWithoutInternalKeys].
     *
     * @return array{0:string,1:array<string,mixed>}
     */
    private function matchRouteAndExtractParams(string $path): array
    {
        $context = $this->router->getContext();
        $tmpContext = clone $context;
        $tmpContext->setMethod("GET");
        $this->router->setContext($tmpContext);

        try {
            $match = $this->router->match($path);
        } catch (\Throwable) {
            $this->router->setContext($context);
            return ["", []];
        } finally {
            $this->router->setContext($context);
        }

        $routeName = $match["_route"] ?? "";
        unset($match["_route"], $match["_controller"]);

        return [$routeName, $match];
    }

    /**
     * Extract query params from a URL (ignores path & fragment).
     *
     * @return array<string,mixed>
     */
    private function extractQueryParamsFromUrl(string $url): array
    {
        $query = parse_url($url, PHP_URL_QUERY) ?? "";
        return $this->parseQueryString($query);
    }

    /**
     * @return array<string,mixed>
     */
    private function parseQueryString(string $query): array
    {
        if ($query === "") {
            return [];
        }
        $params = [];
        parse_str($query, $params);
        return $params;
    }

    /**
     * Replaces (or appends) the query string on a URL path.
     *
     * @param array<string,mixed> $params
     */
    private function replaceQueryString(string $url, array $params): string
    {
        $base = preg_replace("/[?#].*/", "", $url) ?? $url;
        if (empty($params)) {
            return $base;
        }
        $qs = http_build_query($params);
        return $base . ("" !== $qs ? "?" . $qs : "");
    }
}
```

### LINKED ISSUE #3015: [LiveComponent] LiveProps not changed in url if current route has path parameters
Since 2.28.0 the query string does not update with the new value if the current route contains any path parameters.

Reproducer: https://github.com/fracsi/live-component-url-error
1. On the homepage `/` - the page parameter is updated in the url when you change the param (`?value=2`, etc)
2. On a route that has a parameter (`/dummy/{param}`) the query string is not updated.

2.29.* does not solve the problem

### LINKED ISSUE #3052: [LiveComponent] url: true not updating the URL anymore
Hello, 

I have the following fairly simple LiveComponent with 2 `LiveProp`s with `url: true` (only relevant part kept):

```php
#[AsLiveComponent(template: 'components/date_selector.html.twig')]
final class DateSelector
{
    use DefaultActionTrait;
    use ComponentToolsTrait;

    #[LiveProp(writable: true, format: 'Y-m-d', url: true)]
    public ?CarbonImmutable $start = null;

    #[LiveProp(writable: true, format: 'Y-m-d', url: true)]
    public ?CarbonImmutable $end = null;

    #[LiveAction]
    public function setRange(#[LiveArg] string $preset): void
    {
        $today = CarbonImmutable::today();

        switch ($preset) {
            case 'today':
                $this->start = $today->startOfDay();
                $this->end = $today->endOfDay();
            break;

            case 'this_month':
                $this->start = $today->startOfMonth()->startOfDay();
                $this->end = $today->endOfMonth()->endOfDay();
            break;
        }
    }
}
```

This component is rendered in my `property_show` route (this is important for later), like so:

`<twig:DateSelector></twig:DateSelector>`, again nothing fancy.

The component itself has the following template (relevant parts only):

```twig
<div {{ attributes }}>
        <div class="btn-group mb-3" role="group" aria-label="Date shortcuts">
            <button class="btn {{ this.matchesPreset('today') ? 'btn-primary' : 'btn-outline-primary' }} btn-sm"
                {{ live_action('setRange', { preset: 'today' }) }}>
                Today
            </button>
            <button class="btn {{ this.matchesPreset('this_month') ? 'btn-primary' : 'btn-outline-primary' }} btn-sm"
                {{ live_action('setRange', { preset: 'this_month' }) }}>
                This Month
            </button>
        </div>
</div>
```

Just 2 buttons to select a preset date.

Before upgrading to 2.29 from 2.27 (0 code has been changed on my end between both versions):

Image

This used to work perfectly: the date was correctly set in the component AND the URL was updated to reflect the new values, as expected.

On the latest version though, the date is still correctly set in the component but the URL is never updated.

After doing some digging, I get an exception in `Symfony\UX\LiveComponent\Util\UrlFactory` line 42:

Image

The exception is: 

```
Some mandatory parameters are missing ("property") to generate a URL for route "property_show".
```

It's complaining that it's unable to regenerate the route where the component is rendered in because of a missing parameter.

It is indeed missing that parameter, but that's because the component does not need or care about it. Components should (and used to) be able to change the query part of the URL without regenerating the full URL page.

I know this piece of code has been changed a few times over the latest versions (by @Kocal and @mbuliard most recently), looks like something got broken in the middle.

Let me know if i can help more, for now I have reverted to 2.27

Thanks and keep up the good work, awesome components to work with!

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2987, fix #3015, fix #3052 
| License       | MIT

Follows https://github.com/symfony/ux/pull/2673

This PR aims to fix the issues #2987, #3015, and #3052, by rewriting the logic to compute the new `X-Live-Url` when using `LiveProp()` (either with path paremeters or query parameters), inspired by @smnandre's work


src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -68,34 +65,73 @@ public static function getSubscribedEvents(): array
...
+
+                $route = $matched['_route'];
+                unset($matched['_route']);
+
+                $parameters = $matched;
+                foreach ($pathProps as $name => $value) {
+                    if ($value) {
+                        $parameters[$name] = $value;
+                    }
+                }
```
Kocal: I'm still not sure about this part, I need to double-check if my tests data were wrong or not. If that's the case `$pathProps + $matched` should be enough

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -68,34 +65,73 @@ public static function getSubscribedEvents(): array
...
+                $newUrl = $this->router->generate($route, $parameters);
+            } catch (ResourceNotFoundException) {
```
Kocal, line 115: I removed the `MethodNotAllowedException` since I forced the context to `GET` method

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -68,34 +65,67 @@ public static function getSubscribedEvents(): array
...
-        $dehydratedProps = $this->liveComponentHydrator->dehydrate(
-            $mounted->getComponent(),
-            $mounted->getAttributes(),
-            $metadata
-        );
+        $mountedMetadata = $this->metadataFactory->getMetadata($mounted->getName());
 
-        $values = $dehydratedProps->getProps();
+        $dehydratedProps = $this->liveComponentHydrator->dehydrate($mounted->getComponent(), $mounted->getAttributes(), $mountedMetadata);
+        $props = $dehydratedProps->getProps();
```
Kocal: Unless I'm missing something, we still need to get dehydrated props from the mounted component, see https://github.com/symfony/ux/pull/2988

smnandre: Only if we enter the loop… something we do not need to if component is not using url mapping — probably vast majority of usage out there (forms)

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -68,34 +62,67 @@ public static function getSubscribedEvents(): array
...
-        foreach ($metadata->getAllUrlMappings($mounted->getComponent()) as $name => $urlMapping) {
-            if (isset($values[$name]) && $urlMapping) {
-                $urlLiveProps[$urlMapping->mapPath ? 'path' : 'query'][$urlMapping->as ?? $name] = $values[$name];
+    private function generateNewLiveUrl(string $previousUrl, array $pathProps, array $queryProps): string
+    {
+        $previousUrlParsed = parse_url($previousUrl);
+        $newUrl = $previousUrlParsed['path'];
+        $newQueryString = $previousUrlParsed['query'] ?? '';
+
+        if ([] !== $pathProps) {
```
smnandre, line 97: Nice! A lot more efficient, and this should help fullfill the original « no change for anyone not using this new path feature » promise. :clap:

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -68,34 +62,67 @@ public static function getSubscribedEvents(): array
...
+                $tmpContext = clone $context;
+                $tmpContext->setMethod('GET');
+                $this->router->setContext($tmpContext);
+
+                $matched = $this->router->match($previousUrlParsed['path']);
+
+                $route = $matched['_route'];
+                unset($matched['_route']);
+
+                $newUrl = $this->router->generate($route, $pathProps + $matched);
```
smnandre: Are we sure we want to pass all the $matched array here? 

Also, you need to remove any query string that would be defined here… or maybe we should trigger if you are using path parameter that is not allowed by the route. 

Because if you pass aditional values here urlgeneratorinterface will add them, thus creating a big bug at the end when you concatenate strings

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
```diff
@@ -68,34 +62,67 @@ public static function getSubscribedEvents(): array
...
+    private function generateNewLiveUrl(string $previousUrl, array $pathProps, array $queryProps): string
+    {
+        $previousUrlParsed = parse_url($previousUrl);
+        $newUrl = $previousUrlParsed['path'];
+        $newQueryString = $previousUrlParsed['query'] ?? '';
+
+        if ([] !== $pathProps) {
+            $context = $this->router->getContext();
+            try {
+                // Create a temporary RouterContext pointing to the current controller
```
smnandre: The « current controller » is a call able on a LiveComponent … i’d say we want to create a context similar to the URL of the page hosting the component

smnandre (approved):
Nice tests additions btw.. 

… just a tiny doubt on the path thing but ignore them if you are certain it cannot happen :) 

And big Thank You

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
Kocal: Ah sure, it could be optimized in that way, thanks!

src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
Kocal: Indeed, with the current logic, having another `LiveProp(url: new UrlMapping(mapPath: true)` that is not used as a path parameter, will instead add it as a query parameter. 

Instead, we should use keys from `$matched` but take values from `$pathProps`

Kocal: Just pushed a fix, and added `pathPropForAnotherController` LiveProp that should not be present in the new URL

Kocal:
I've just applied your comments @smnandre, which indeed fixed a case where a `LiveProp(url: new UrlMapping(mapPath: true))` with a default value `!= null` could be present in the URL as a query parameter, if the LiveProp didn't match a path parameter.

Kocal:
I confirm it does not add any regressions for https://github.com/symfony/ux/issues/2960 nor https://github.com/symfony/ux/issues/2975 nor https://github.com/symfony/ux/pull/2988

PR #3063, opened 2025-09-02
[Icons][Doc] Updates for UX Icons docs

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

This PR proposes:

* To demote the installation of `http-client`. Showing it at the beginning of the doc creates unnecessary friction. Better explain it later when it's relaly needed.
* Promote the installation of `ux-twig-component` when using the HTML syntax. It was included in a tip, but now it's shown first and more prominently to avoid user frustation.
* Other minor tweaks

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -16,9 +16,6 @@ Installation
 
     $ composer require symfony/ux-icons
 
-    # To use provided on-demand icon sets, you also need HTTP client:
-    $ composer require symfony/http-client
-
 SVG Icons
 ---------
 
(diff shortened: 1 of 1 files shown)


Kocal:
Thanks for the suggestions!

I added Kevin and Simon as reviewers because the installation of `symfony/http-client` subject that recurs legit every month (😅), and I know they have some opinions on that.

The rest is fine for me

kbond (approved):
Thanks @javiereguiluz, we have better exception messages when http-client is missing, so I think this is perfect now!

kbond:
Thank you Javier.

PR #3064, opened 2025-09-02
[LiveComponent][Doc] Update Live Components docs for reusable bundles

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

I lost a lot of time when trying to make a Live Component inside a reusable bundle work. I could only fix it when looking at how other budnles solved it (e.g. https://github.com/Mezcalito/ux-search/blob/9e34a01707d4c09e7538e35b6e8549f3722e0ca4/config/services.php#L62-L78).

So, I propose to add some docs explaining how to do it.

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -1,13 +1,13 @@
 Live Components
 ===============
 
-Live components builds on top of the `TwigComponent`_ library
+Live components build on top of the `TwigComponent`_ library
 to give you the power to automatically update your Twig components on
 the frontend as the user interacts with them. Inspired by
 `Livewire`_ and `Phoenix LiveView`_.
 
 If you're not familiar with Twig components yet, it's worth taking a few minutes
-to familiarize yourself in the `TwigComponent documentation`_.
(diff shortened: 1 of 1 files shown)


Kocal:
I'm not sure about this one, it feels super hacky since it relies on internals structure. 

Is there a better alternative? note that `Symfony\UX\LiveComponent\Attribute\AsLiveComponent::serviceConfig()` exists but that's still internal. cc @kbond @smnandre

smnandre:
We would like to open this up more, but every step towards opening is met with abuse, so that’s not encouraging…

But, as @Kocal politely said, it’s a no-go to document the internal (and soon-to-change) component-registration mechanism.

The Mezcalito team already knows this can change, and I would strongly advise anyone against using internal definitions to register services (LiveComponent or ... any Symfony component).

Also, as always, we’re not many around here (and none of us is paid to work on Symfony), so things move slowly…

I’m closing this PR to avoid exposing something we don’t want people to feel frustrated about ~if~ when it changes… but rest assured we also have this in mind—just lacking time to implement new ideas/better methods.

PR #3066, opened 2025-09-03
[TwigComponent][Perf] Add option to disable the dump of components

alinceDev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no 
| Issues        | -
| License       | MIT

This PR removes the component object dumping from the Symfony UX Twig Component profiler to address memory consumption issues.

IMHO, the component dump is not necessary in the profiler as it can be quite verbose and memory-intensive. When detailed component inspection is needed, it's preferable to dump the component manually in the specific context where debugging is required.

### DIFF
--- a/src/TwigComponent/config/debug.php
+++ b/src/TwigComponent/config/debug.php
@@ -28,7 +28,7 @@ return static function (ContainerConfigurator $container) {
         ->args([
             service('ux.twig_component.component_logger_listener'),
             service('twig'),
-            abstract_arg('profiler dump components'),
+            abstract_arg('profiler collect components'),
         ])
         ->tag('data_collector', [
             'template' => '@TwigComponent/Collector/twig_component.html.twig',
--- a/src/TwigComponent/config/debug.php
+++ b/src/TwigComponent/config/debug.php
@@ -15,6 +15,7 @@ use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigura
 use Symfony\UX\TwigComponent\DataCollector\TwigComponentDataCollector;
 use Symfony\UX\TwigComponent\EventListener\TwigComponentLoggerListener;
 
+use function Symfony\Component\DependencyInjection\Loader\Configurator\abstract_arg;
 use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
 
 return static function (ContainerConfigurator $container) {
@@ -27,6 +28,7 @@ return static function (ContainerConfigurator $container) {
         ->args([
             service('ux.twig_component.component_logger_listener'),
             service('twig'),
+            abstract_arg('profiler dump components'),
         ])
         ->tag('data_collector', [
             'template' => '@TwigComponent/Collector/twig_component.html.twig',
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -150,10 +150,11 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
         $container->setAlias('console.command.stimulus_component_debug', 'ux.twig_component.command.debug')
             ->setDeprecated('symfony/ux-twig-component', '2.13', '%alias_id%');
 
-        if ($container->getParameter('kernel.debug') && $config['profiler']) {
+        if ($config['profiler']['enabled']) {
             $loader->load('debug.php');
+
             $container->getDefinition('ux.twig_component.data_collector')
-                ->setArgument(2, $config['profiler_dump_components'] ?? true);
+                ->setArgument(2, $config['profiler']['collect_components']);
         }
 
         $loader->load('cache.php');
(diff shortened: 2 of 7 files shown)


miladb (approved):

Kocal (approved):
WDYT @kbond?

smnandre:
@alinceDev

So i checked on the project I work on and .... I'm not a big fan of this idea (but not against the idea of not storing the components ... but)

Multiple data are only visible here in the profiler... and basic scalar data is not something that would take that much memory (at all).

Could you maybe create a repository / share some code (or context) to understand what numver of components (or what amount of inner data) becomes a problem in dev for you ?

I'd rather we change profiling and remove an important feature of it only if we have a concrete example to base this choice on....

Kocal:
Since having components dumps in the profiler is useful but could become a mess on memory usage when having too much components, what about introducing a new option `profile_components` (aside the already existing option `profiler`)?

This option could be `true` by default, but in the recipe it could be configured to `false` with a comment explaining the possible performance issues. 
WDYT?

smnandre:
> Since having components dumps in the profiler is useful but could become a mess on memory usage when having too much components, what about introducing a new option `profile_components` (aside the already existing option `profiler`)?
> 
> This option could be `true` by default, but in the recipe it could be configured to `false` with a comment explaining the possible performance issues. WDYT?

100%, something I suggested [here](https://github.com/symfony/ux/issues/3074#issuecomment-3332448369) but forgot to move back here (my bad)

Kocal:
Status: Needs work

alinceDev:
I updated the PR @smnandre @Kocal

src/TwigComponent/config/debug.php
```diff
@@ -27,6 +28,7 @@
         ->args([
             service('ux.twig_component.component_logger_listener'),
             service('twig'),
+            param('ux.twig_component.profiler_dump_components'),
```
smnandre: I would use an abstract arg here, and define the parameter on the definition after the load(debug.php) in the extension class...  not sure we should create a parameter here

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -69,6 +69,7 @@ public function load(array $configs, ContainerBuilder $container): void
             $defaults = [];
         }
         $container->setParameter('ux.twig_component.component_defaults', $defaults);
+        $container->setParameter('ux.twig_component.profiler_dump_components', $config['profiler_dump_components']);
```
smnandre: I would remove this then :)

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -219,6 +220,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->info('Enables the profiler for Twig Component (in debug mode)')
                     ->defaultValue('%kernel.debug%')
                 ->end()
+                ->booleanNode('profiler_dump_components')
+                    ->info('Enables the dump components for Twig Component (in debug mode)')
+                    ->defaultValue('%kernel.debug%')
+                ->end()
```
smnandre: defaultTrue() seems more logical here .. (its not just in debug mode but... when twig component profiler is enabled) 

Without the parenthesis, probably something like 

"Dump components in the Twig Component profiler panel" 

wdyt ?

smnandre:
Also could you add a tests or two, you can find similar ones in tests/Unit/DependencyInjection/TwigComponentExtensionTest.php and tests/Unit/DataCollector/TwigComponentDataCollectorTest.php

alinceDev:
@smnandre I updated the PR, but I don't know what to test in the TwigComponentDataCollectorTest class.

Kocal:
Hey :) 

Thanks for pushed your changes, I just pushed some modifications too:
- I believe our `twig_components.profiler` configuration should behave like `framework.profiler` from Symfony, thanks to `canBeEnabled()` method:
```yaml
# Below some examples where the profiler is enabled

twig_components:
    profiler: true

twig_components:
    profiler:
        enabled: true

twig_components:
    profiler:
        enabled: true
        collect_components: true # or false

twig_components:
    profiler:
        collect_components: true # or false


# Below some examples where the profiler is disabled

twig_components:
    profiler: false

twig_components:
    profiler: 
        enabled: false

twig_components:
    profiler: 
        enabled: false
        collect_components: true # no effect

```
- I removed the check on `$container->getParameter('kernel.debug')` in `TwigComponentExtension`. AFAIU it does not make sense since we already enable/disable the profiler collection through default value `%kernel.debug%` (from `$config['profiler']['enabled']`, previously `$config['profiler']`). 
  It now behaves like `twig.debug`, `twig.strict_variables`, `doctrine.dbal.profiling_collect_backtrace`, etc...
- The `TwigComponentDataCollectorTest` test has been improved to ensure that no component instances are collected 
- I added an entry in the CHANGELOG

Let's merge when the CI is green

Kocal:
Thank you @alinceDev.

PR #3077, opened 2025-09-11
[LiveComponent] Improve performance for LiveUrlSubscriber

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Improve performance by removing heavy services if not needed.

### DIFF
--- a/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
+++ b/src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
@@ -24,6 +24,7 @@ use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Parameter;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\Routing\RouterInterface;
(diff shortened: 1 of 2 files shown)


src/LiveComponent/src/DependencyInjection/LiveComponentExtension.php
```diff
@@ -138,12 +139,10 @@ function (ChildDefinition $definition, AsLiveComponent $attribute) {
         ;
 
         $container->register('ux.live_component.live_url_subscriber', LiveUrlSubscriber::class)
-            ->setArguments([
-                new Reference('ux.live_component.metadata_factory'),
-                new Reference('ux.live_component.component_hydrator'),
-                new Reference('router'),
-            ])
             ->addTag('kernel.event_subscriber')
+            ->addTag('container.service_subscriber', ['key' => LiveComponentHydrator::class, 'id' => 'ux.live_component.component_hydrator'])
```
94noni, line 143: why not using [lazy](https://symfony.com/doc/current/service_container/lazy_services.html)?

dannyvw: This is the same as the other implementations

Kocal:
Nice catch, thanks!

PR #3079, opened 2025-09-13
[Map] Add "color" support for SVG and UX Icon icons

JeDeveloppe:
Bonjour,

Désolé je ne sais pas trop ou faire ceci et j'espère ne pas tout casser.

Est ce possible de rajouter cette fonction couleur sur l'icon ?

Merci

### DIFF
--- a/src/Map/src/Icon/Icon.php
+++ b/src/Map/src/Icon/Icon.php
@@ -61,6 +61,7 @@ abstract class Icon
         protected IconType $type,
         protected int $width = 24,
         protected int $height = 24,
+        protected ?string $color = null, // Nouvelle propriété pour la couleur
     ) {
     }
 
@@ -89,6 +90,20 @@ abstract class Icon
     }
 
     /**
+     * Sets the color of the icon.
+     * Note: This only has an effect on SvgIcon and UxIcon types when rendered client-side by the JavaScript controller.
+     * The color should be a valid CSS color string (e.g., 'red', '#FF0000', 'rgb(255,0,0)').
+     *
+     * @param non-empty-string $color
+     */
+    public function color(string $color): static
+    {
+        $this->color = $color;
+
+        return $this;
+    }
+
+    /**
      * @internal
      */
     public function toArray(): array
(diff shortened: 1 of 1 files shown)


src/Map/src/Icon/Icon.php
```diff
@@ -61,6 +61,7 @@ protected function __construct(
         protected IconType $type,
         protected int $width = 24,
         protected int $height = 24,
+        protected ?string $color = null, // Nouvelle propriété pour la couleur
```
Kocal, line 64: The comment does not bring any value

```suggestion
        protected ?string $color = null,
```

src/Map/src/Icon/Icon.php
```diff
@@ -88,6 +89,20 @@ public function height(int $height): static
         return $this;
     }
 
+    /**
+     * Sets the color of the icon.
+     * Note: This only has an effect on SvgIcon and UxIcon types when rendered client-side by the JavaScript controller.
+     * The color should be a valid CSS color string (e.g., 'red', '#FF0000', 'rgb(255,0,0)').
```
Kocal, line 95:
```suggestion
     * Must be a valid CSS color (e.g.: 'red', '#FF0000', 'rgb(255,0,0)'). Only affects UX Icon or SVG icons.
```

src/Map/src/Icon/Icon.php
```diff
@@ -97,14 +112,15 @@ public function toArray(): array
             'type' => $this->type->value,
             'width' => $this->width,
             'height' => $this->height,
+            'color' => $this->color, // Ajout de la couleur au tableau de sérialisation
```
Kocal, line 115: The comment does not bring any value

```suggestion
            'color' => $this->color,
```

Kocal (changes requested):
Thanks for the contribution:
1. Please use the default PR template in your description
2. You must add test for this new feature, in order to prevent any regression
3. You also must update existing tests and snapshots, to make the CI checks pass

Thanks

smnandre (changes requested):
I'm sorry but for now we cannot accept this PR.

We cannot add a specific non typed/non-validated field on an central object without tests nor full working usage...

Did you try to use metadata on your markers ? Or to follow the other guidelines / advises given on your issue ?

Kocal:

Thanks!

Status: Needs Work

PR #3082, opened 2025-09-15
[Icons] Add documentation on icons colors

StevenRenaux:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no
| License       | MIT

If we use `fill` or `stroke` attributes to update the color icon, as explained into https://symfony.com/bundles/ux-icons/current/index.html#default-attributes, nothing happens.

Let's explore a default Boostrap SVG together: https://icons.getbootstrap.com/icons/person/
```html
{{ ux_icon('bi:person') }}
```

If we go into Bootstrap to grab that icon, it looks like this
```html
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-person" viewBox="0 0 16 16">
  <path d="M8 8a3 3 0 1 0 0-6 3 3 0 0 0 0 6m2-3a2 2 0 1 1-4 0 2 2 0 0 1 4 0m4 8c0 1-1 1-1 1H3s-1 0-1-1 1-4 6-4 6 3 6 4m-1-.004c-.001-.246-.154-.986-.832-1.664C11.516 10.68 10.289 10 8 10s-3.516.68-4.168 1.332c-.678.678-.83 1.418-.832 1.664z"/>
</svg>
```

UX-icon uses Iconify's icons and this icon is not completely the same https://github.com/iconify/icon-sets/blob/master/json/bi.json#L4472
Iconify returns the `path` tag for the bootsrap icon, this part is incremented with the `fill` attribute or the `stroke`, depend on the need.
```html
<path fill="currentColor" d="M8 8a3 3 0 1 0 0-6a3 3 0 0 0 0 6m2-3a2 2 0 1 1-4 0a2 2 0 0 1 4 0m4 8c0 1-1 1-1 1H3s-1 0-1-1s1-4 6-4s6 3 6 4m-1-.004c-.001-.246-.154-.986-.832-1.664C11.516 10.68 10.289 10 8 10s-3.516.68-4.168 1.332c-.678.678-.83 1.418-.832 1.664z"></path>
```

But if we add `fill` or `stroke` to it will add these attributes to the svg tag and will not update the ones into the `path` tags, and that's the problem.
```html
{{ ux_icon('bi:person', {fill: 'red'}) }}

<!-- Rendered -->
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" fill="red" aria-hidden="true"><path fill="currentColor" d="M8 8a3 3 0 1 0 0-6a3 3 0 0 0 0 6m2-3a2 2 0 1 1-4 0a2 2 0 0 1 4 0m4 8c0 1-1 1-1 1H3s-1 0-1-1s1-4 6-4s6 3 6 4m-1-.004c-.001-.246-.154-.986-.832-1.664C11.516 10.68 10.289 10 8 10s-3.516.68-4.168 1.332c-.678.678-.83 1.418-.832 1.664z"></path></svg>
```

With the modifications the `path` tag also follow the same rules
```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" fill="red" aria-hidden="true"><path fill="red" d="M8 8a3 3 0 1 0 0-6a3 3 0 0 0 0 6m2-3a2 2 0 1 1-4 0a2 2 0 0 1 4 0m4 8c0 1-1 1-1 1H3s-1 0-1-1s1-4 6-4s6 3 6 4m-1-.004c-.001-.246-.154-.986-.832-1.664C11.516 10.68 10.289 10 8 10s-3.516.68-4.168 1.332c-.678.678-.83 1.418-.832 1.664z"></path></svg>
```

### DIFF
--- a/src/Icons/doc/index.rst
+++ b/src/Icons/doc/index.rst
@@ -112,9 +112,31 @@ the size of all your icons from a single place.
 Icon Color
 ~~~~~~~~~~
 
-Typically, SVG icons use ``fill="currentColor"`` to inherit the color of the containing element.
-You can set the color in CSS on the container or directly on the SVG element/class.
+UX Icons renders SVG icons with ``fill="currentColor"`` (see the `default configuration`_).
 
+This means icons take their color from the CSS ``color`` property, which by default is inherited from the text color of
+the parent element.
+
+.. code-block:: html+twig
+
+    <a href="#" class="text-blue-500">
+        {# icon color: blue #}
+        {{ ux_icon('bi:star') }}
+    </a>
+
+You can set the icon's color using CSS classes, either on the parent container or directly on the icon:
+
+.. code-block:: html+twig
+
+    <div class="text-red-500">
+        {# Inherit color from container (red) #}
+        {{ ux_icon('warning') }}
+
+        {# Override parent with CSS class (blue) #}
+        {{ ux_icon('warning', { class: 'text-blue-500' }) }}
+    </div>
+
+To configure a default class for all icons across your application see the `default configuration`_.
 
 Icon Sets
 ~~~~~~~~~
@@ -674,3 +696,4 @@ Learn more
 .. _`Tabler Icons`: https://github.com/tabler/tabler-icons
 .. _`Creating and Using Templates`: https://symfony.com/doc/current/templates.html
 .. _`How to manage CSS and JavaScript assets in Symfony applications`: https://symfony.com/doc/current/frontend.html
+.. _`default configuration`: https://symfony.com/bundles/ux-icons/current/index.html#full-configuration


smnandre:
This is also a massive BC break here, don't you think ? And losing the read-only side of thing ..

DId you try simply setting the color on the icon ?

src/Icons/src/Icon.php
```diff
@@ -155,6 +155,14 @@ public function toHtml(): string
                 continue;
             }
 
+            if ('fill' === $name) {
+                $this->innerSvg = str_replace('fill="currentColor"', "fill=\"{$value}\"", $this->innerSvg);
```
smnandre: Hard BC break here.

src/Icons/src/Icon.php
```diff
@@ -155,6 +155,14 @@ public function toHtml(): string
...
+            }
+
+            if ('stroke' === $name) {
+                $this->innerSvg = str_replace('stroke="currentColor"', "stroke=\"{$value}\"", $this->innerSvg);
```
smnandre: Let's keep an only place we echo value and espace / encode them correctly (just below)

smnandre (changes requested):
I get the idea, but we need to find a better method... 


And i'm not yet convinced of the usage... at least not in the one you provided.. maybe i'm not seeing something

StevenRenaux:
> This is also a massive BC break here, don't you think ? And losing the read-only side of thing ..
> 
> DId you try simply setting the color on the icon ?

@smnandre 
I know, and I agree. It was mainly to show where the problem lies. The documentation suggests this is possible, but it's not with this attribute (https://symfony.com/bundles/ux-icons/current/index.html#default-attributes).

Looking at this again this morning, I just found that this is possible with the color attribute and is even perfect in this case.
https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/color
> The color attribute is used to provide a potential indirect value, currentColor, for the fill, stroke, stop-color, flood-color, 
> and lighting-color attributes.

Perhaps the documentation should be modified accordingly, and mention that if you want to change the color, you can do it via CSS, as Iconify also mentions in its documentation.

image

```html
{{ ux_icon('flowbite:clock-outline', {class: 'size-4 text-gray-500'}) }}
```

Or with this color attribute. And we ultimately avoid the current problem with the fill attribute.

```html
{{ ux_icon('flowbite:clock-outline', {class: 'size-4', color: 'gray'}) }}
```

WDYT?

StevenRenaux:
Already noticed by another PR https://github.com/symfony/ux/pull/2671

smnandre:
Very open if you want to open a PR on the documentation with clear and positive information about this 

(please do not only mention Tailwind that's the only thing I'm asking for haha)

poke me if you need any guidance here :)

src/Icons/doc/index.rst
```diff
@@ -112,9 +112,22 @@ the size of all your icons from a single place.
 Icon Color
 ~~~~~~~~~~
 
-Typically, SVG icons use ``fill="currentColor"`` to inherit the color of the containing element.
-You can set the color in CSS on the container or directly on the SVG element/class.
+By default, SVG icons rendered by UX Icons use ``fill="currentColor"``.
```
smnandre: What about something like 

_We say it's only a default setting and point to where it's configured_

```rst
UX Icons renders SVG icons with ``fill="currentColor"`` (see the :ref:`default configuration <icons_default_attributes>`).
```

_Then we explain what it means_

```rst

This means icons take their color from the CSS ``color`` property. By default, this value is inherited from the text color of the parent element.

.. code-block:: twig
   
    <a href="#">
        {{ ux_icon('bi:star') }}    {# icon color: blue #}
    </a>
```

We then explain how to style one icon (with class)

```rst
To control an **icon's color**, you can set CSS classes on either the containing element or directly on the icon:

.. code-block:: twig

   <div class="text-red-500">

       {# Inherit color from container (red) #}
       {{ ux_icon('warning') }}

        
       {# Override parent with CSS class (blue) #}
       {{ ux_icon('warning', {class: 'text-blue-500'}) }}

   </div>
```

((here the color attribute, see comment below))

_Finally we mention how to style them all / default (add class on default attributes)_

```rst
To configure a default class for all icons across your application, see the :doc:`bundle configuration </configuration>`.
```

src/Icons/doc/index.rst
```diff
@@ -112,9 +112,22 @@ the size of all your icons from a single place.
...
+
+    {{ ux_icon('user-profile', {class: 'my_class'}) }}
+
+Or you can just as easily use the color attribute.
+
+.. code-block:: twig
+
+    {{ ux_icon('user-profile', {color: 'red'}) }}
+
+Both will indirectly apply a potential value for currentColor.
```
smnandre: Nitpicking here, but both would _directly_ set the value of "color".  ;)

src/Icons/doc/index.rst
```diff
@@ -112,9 +112,22 @@ the size of all your icons from a single place.
...
+
+.. code-block:: twig
+
+    {{ ux_icon('user-profile', {class: 'my_class'}) }}
+
+Or you can just as easily use the color attribute.
+
+.. code-block:: twig
+
+    {{ ux_icon('user-profile', {color: 'red'}) }}
```
smnandre: Setting directly the "color attribute" is [not valid in HTML5](https://html.spec.whatwg.org/multipage/indices.html#attributes-3:attr-link-color) (and was already deprecated in XHTML).

Moreover, current best practices tend to avoid any hard-coded color value, as they do not work good with multiple color modes (dark/light, accessibility, etc.)

So i would suggest we either do not mention this technique directly... or we add some contextual information about it.. wdyt ?

StevenRenaux: If we need to avoid it, I prefer to not mentionned it

src/Icons/doc/index.rst
```diff
@@ -112,9 +112,31 @@ the size of all your icons from a single place.
 Icon Color
 ~~~~~~~~~~
 
-Typically, SVG icons use ``fill="currentColor"`` to inherit the color of the containing element.
-You can set the color in CSS on the container or directly on the SVG element/class.
+UX Icons renders SVG icons with fill="currentColor" (see the `default configuration`_).
```
smnandre:
```suggestion
UX Icons renders SVG icons with ``fill="currentColor"`` (see the `default configuration`_).
```

src/Icons/doc/index.rst
```diff
@@ -112,9 +112,31 @@ the size of all your icons from a single place.
...
 
+This means icons take their color from the CSS color property, which by default is inherited from the text color of
```
smnandre:
```suggestion
This means icons take their color from the CSS ``color`` property, which by default is inherited from the text color of
```

smnandre (approved):

smnandre:
(thank you @StevenRenaux ! 😃  )

Kocal:
Thank you @StevenRenaux.

PR #3088, opened 2025-09-18
[LiveComponent] Replace the browser's URL before triggering `render:finished` hook

xDeSwa:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | >
| License       | MIT

After the URL change, performing the **``render:finished``** event will provide smoother control.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -2021,7 +2021,6 @@ var Component = class {
         thisPromiseResolve(backendResponse);
         return response;
       }
-      this.processRerender(html, backendResponse);
       const liveUrl = backendResponse.getLiveUrl();
       if (liveUrl) {
         history.replaceState(
@@ -2030,6 +2029,7 @@ var Component = class {
           new URL(liveUrl + window.location.hash, window.location.origin)
         );
       }
+      this.processRerender(html, backendResponse);
       this.backendRequest = null;
       thisPromiseResolve(backendResponse);
       if (this.isRequestPending) {
--- a/src/LiveComponent/assets/src/Component/index.ts
+++ b/src/LiveComponent/assets/src/Component/index.ts
@@ -327,7 +327,6 @@ export default class Component {
                 return response;
             }
 
-            this.processRerender(html, backendResponse);
             const liveUrl = backendResponse.getLiveUrl();
             if (liveUrl) {
                 history.replaceState(
@@ -336,6 +335,8 @@ export default class Component {
                     new URL(liveUrl + window.location.hash, window.location.origin)
                 );
             }
+
+            this.processRerender(html, backendResponse);
 
             // finally resolve this promise
             this.backendRequest = null;


Kocal (approved):
Makes sense to me

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -2030,6 +2030,7 @@ var Component = class {
           new URL(liveUrl + window.location.hash, window.location.origin)
         );
       }
+      this.hooks.triggerHook("render:finished", this);
```
smnandre: Can we movev the liveUrl update in processRerender instead ?

Kocal: We should be careful on how we do it, because the method contains some early returns that could prevent the URL update

xDeSwa: done. moved into processRerender

Kocal: But now people hooking on `render:started` and passing `shouldRender = false` won't see the URL updated anymore, that's a BC

smnandre: Yep indeed.. but if I get this right, with this PR the hook would be trigger even when the render is skipped or if an error occurs .. and this would be some BC too :|  Or do we consider this a bug ? 

But there are [at least two places](https://github.com/search?q=repo%3AxDeSwa%2Fux+%22render%3Afinished%22+path%3A%2F%5Esrc%5C%2FLiveComponent%5C%2Fassets%5C%2F%2F&type=code) we need to check attentively how it would impact codebases: 

This one (seems without effect but i'd like to be sure):

```js
        component.on('render:finished', () => {
            // re-start polling, in case polling changed
            this.initializePolling();
        });
```

[/src/LiveComponent/assets/src/Component/plugins/PollingPlugin.ts#L21-L24](https://github.com/xDeSwa/ux/blob/6f7e61784218daf5772ae0b15cd55377da5973a2/src/LiveComponent/assets/src/Component/plugins/PollingPlugin.ts#L21-L24)

But this one makes me more hesitant: 

```js
    attachToComponent(component: Component): void {
        this.synchronizeValueOfModelFields(component);
        component.on('render:finished', () => {
            this.synchronizeValueOfModelFields(component);
        });
```

[src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts#L16-L18](https://github.com/xDeSwa/ux/blob/6f7e61784218daf5772ae0b15cd55377da5973a2/src/LiveComponent/assets/src/Component/plugins/SetValueOntoModelFieldsPlugin.ts#L16-L18)

xDeSwa:
@Kocal | @smnandre 
URL change method moved before ReRender function.

Since no backendResponse data changes are made within the ReRender function, changing the URL before the function makes sense. This will avoid any hitches to hook operations and ensure access to the correct URL within the hook.

smnandre:
Seems ok to me! Can you fix fabbot check ? (remove the merge commit and rebase on 2.x) .. or we won't be able to merge 😅

xDeSwa:
@smnandre is it ok now? thanks

smnandre:
Well, i do not see any red in the checks box so .. yeah 😄

smnandre (approved):
Thank you @xDeSwa !!

Kocal:
Thank you @xDeSwa.

PR #3094, opened 2025-09-22
[LiveComponent][Doc] Mention that Live Components don't work in reusable bundles

javiereguiluz:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

This is related to #3064 but instead of explaining how to make live components work in Symfony bundles, the new docs tells that you cannot do it.

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -3946,6 +3946,18 @@ Use the addCollectionItem method from the LiveCollectionTrait to dynamically add
         ]
     ]]);
 
+Using Live Components in Reusable Bundles
+-----------------------------------------
+
+`Symfony Bundles`_ let you reuse features across many different Symfony
+applications. However, you **cannot include live components in reusable
+bundles** to make them available in any application that installs the bundle.
+
+Some projects, such as the `Mezcalito UX File Manager bundle`_, provide live
+components inside reusable bundles. Their implementation relies on unsupported
+hacks that could break at any time, so Symfony does not recommend using this
+approach.
+
 Backward Compatibility promise
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


smnandre:
I'm not very sure why we should document this "non features".. 
- I don't see such block on other docs pages like workflow, mailer, security... 
- this package documentation is already massively too long to read (mainly due to the fact we got a "no" when we asked to split it)
- the "do NOT do it....  _but_ hey look how they do it" seems a bit provocative to me 

The team behind the two bundles you mentioned on this PR, the issue, the other:
- the search bundle uses 100% manual service registration but to do so they indeed used hard-coded internal tag / values  
- the file bundle uses autowire and autoconfigure for all the services

As i see it, none of these can and should be recommended (both being mentioned as "not to do" in best practices and other parts of the Symfony Bundle documentation).

But the main problem here is not on this level... it's a matter of security. Symfony documentation on [Routing](https://symfony.com/doc/current/bundles/override.html#routing) states

> Routing is **never** automatically imported in Symfony. If you want to include the routes from any bundle, then they must be manually imported from somewhere in your application (e.g. config/routes.yaml).
> 
> The easiest way to "override" a bundle's routing is to never import it at all. Instead of importing a third-party bundle's routing, copy that routing file into your application, modify it, and import it instead.

So a Bundle that forcefully enable LiveComponent in userland is in fact opening endpoints to the internet. 

After that, we also have concerns about the naming / overriding stuff .. and finally some problems with Live Components from external 

:| 

Problem is many do not realize how TwigComponent and LiveComponent are intricated, and plenty of recent tentative of changes (even minor ones) have proven it's a very fragile castle.. 

I'd love to work on this (and a thousands other things on UX) but, again, I only have my free time to do so, and had plenty of other things to deal with lately.

Kocal:
I'm closing for the reasons explained above, thanks anyway for the contribution

PR #3096, opened 2025-09-24
 [StimulusBundle] Add missing closing brace in AssetMapper 6.3 example

quentinRogeret34:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | 
| License        | MIT

---

This PR fixes a typo in the documentation.

Changed:

```twig
For AssetMapper 6.3 only, you also need a ``{{ ux_controller_link_tags() }`
``` 
to 

```twig
For AssetMapper 6.3 only, you also need a ``{{ ux_controller_link_tags() }}``
```

### DIFF
--- a/src/StimulusBundle/doc/index.rst
+++ b/src/StimulusBundle/doc/index.rst
@@ -418,7 +418,7 @@ is running in debug mode.
 
 .. tip::
 
-    For AssetMapper 6.3 only, you also need a ``{{ ux_controller_link_tags() }``
+    For AssetMapper 6.3 only, you also need a ``{{ ux_controller_link_tags() }}``
     in ``base.html.twig``. This is not needed in AssetMapper 6.4+.
 
 With WebpackEncoreBundle


smnandre:
Thanks! 

Wondering if we should not remove this mention entirely ? Symfony 6.3 is not maintained anymore and i'm not sure this is usefull for anyone, but it adds "noise" for the other readers.. wdyt ? 

cc @Kocal

smnandre (approved):

quentinRogeret34:
You're right, it does seem very tied to a specific version that shouldn’t be used anymore 🤔 
I can update my PR to remove this block if that makes more sense

Kocal:
I suggest to keep it since we still maintain support for Symfony ^6.3.

However in UX 3.0, we can remove the doc block and the `ux_controller_link_tags()` Twig function (that is already deprecated)

smnandre (approved):

Kocal:
Thank you @quentinRogeret34.

PR #3103, opened 2025-09-24
[Autocomplete] Remove deprecated code for 3.0

smnandre:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no
| Issues         | Fix #...
| License        | MIT

src/Autocomplete/src/Form/ParentEntityAutocompleteType.php:27
```php
        @deprecated since Symfony UX 2.13, use "Symfony\UX\Autocomplete\Form\BaseEntityAutocompleteType" instead
```

src/Autocomplete/src/Form/ChoiceList/Loader/ExtraLazyChoiceLoader.php:21
```php
        @deprecated since Symfony UX 2.23 and will be removed in 3.0, use Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader instead.
```

src/Autocomplete/src/Form/AutocompleteEntityTypeSubscriber.php:26
```php
        @deprecated since Symfony UX 2.13
```        

src/Autocomplete/src/Doctrine/EntityMetadata.php:47
```php
    trigger_deprecation('symfony/ux-autocomplete', '2.15.0', 'Calling EntityMetadata::getPropertyMetadata() is deprecated. You should stop using it, as it will be removed in the future.');
```

src/Autocomplete/src/AutocompleteResultsExecutor.php:38
```php
    trigger_deprecation('symfony/ux-autocomplete', '2.8.0', 'Passing a "%s" instance as the second argument of "%s()" is deprecated, pass a "%s" instance instead.', Security::class, __METHOD__, PropertyAccessorInterface::class);
```


smnandre:
Hmm... the LazyChoiceLoader is only available in Symfony 7.2 :| 

https://symfony.com/blog/new-in-symfony-7-2-lazy-choice-loader

What do we do ? Revert (but then the deprecation is a bit strange to be honest...) or bump requirements ?

Kocal (approved):

Kocal:
> Hmm... the LazyChoiceLoader is only available in Symfony 7.2 :|
> 
> [symfony.com/blog/new-in-symfony-7-2-lazy-choice-loader](https://symfony.com/blog/new-in-symfony-7-2-lazy-choice-loader)
> 
> What do we do ? Revert (but then the deprecation is a bit strange to be honest...) or bump requirements ?

In the deprecation we mentioned to use the `Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader`, which imply to use Symfony Form 7.2. 
Since this version is compatible with Symfony packages 6.4 (https://github.com/symfony/form/blob/7.2/composer.json), we could bump requirements (and add an explicit dependency on `symfony/form`, since the feature is not optional AFAIK?)

Kocal:
> > Hmm... the LazyChoiceLoader is only available in Symfony 7.2 :|
> > [symfony.com/blog/new-in-symfony-7-2-lazy-choice-loader](https://symfony.com/blog/new-in-symfony-7-2-lazy-choice-loader)
> > What do we do ? Revert (but then the deprecation is a bit strange to be honest...) or bump requirements ?
> 
> In the deprecation we mentioned to use the `Symfony\Component\Form\ChoiceList\Loader\LazyChoiceLoader`, which imply to use Symfony Form 7.2. Since this version is compatible with Symfony packages 6.4 ([symfony/form@`7.2`/composer.json](https://github.com/symfony/form/blob/7.2/composer.json?rgh-link-date=2025-09-25T07%3A04%3A51Z)), we could bump requirements (and add an explicit dependency on `symfony/form`, since the feature is not optional AFAIK?)

Finally I went for another solution. Since the `BaseEntityAutocompleteType` is only registered as a service when `symfony/form` is available, I added a check that throws an exception if `LazyChoiceLoader` does not exist, telling the user to install Symfony Form >= 7.2.

Kocal:
Thank you @smnandre.

PR #3111, opened 2025-09-29
[LiveComponent] Better file handling for forms 

florentdestremau:
| Q              | A
| -------------- | ---
| Bug fix?       | no, but this feels like a bugfix to me
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | needed 
| Issues         | Fix #... 
| License        | MIT

These changes allow a better file handling for live component forms with files. Currently, the docs require you to handle files directly from the $request object: "The files will be available in a regular `$request->files` files bag". Si if you have a Dto such as :

```php
class Dto {
  public string $name;
  public ?UploadedFile $file;
}
```
with a corresponding form (text + file inputs), after using $this->submitForm($form) in your live component you would have an empty Dto::file field, and you would be forced to look up inside the $request object. Here I copied and adapted the $form->handleRequest($request) logic you get in a classic controller workflow, wich picks up the files in the request and injects them as data, so now you would do

```php
#[LiveAction]
public function save(Request $request) {
  $this->submitForm(request: $request);
  $dto = $this->getForm()->getData();

  $dto->file->move('public/uploads');   // yay $dto->file is an UploadedFile now !
}
```
And at this point you _will_ have $dto::file as an UploadedFile instance you can freely manipulate.

Because `submitForm` has an existing optional parameter the usage feels weird, but to avoid BC break the Request object is to be injected second like `$this->submitForm(true, $request)` or as a named parameter `$this->submitForm(request: $request)` for proper usage. IMO the two parameters should be swapped in a future major version, I don't think the usage priority of $validateAll precedes the $request handling.

This can maybe break the data workflow if you have touched `$this->formValues` somehow because it takes its data from the request directly. I'm not familiar with _why_ this wasn't done this way in the first place so happy to discuss it!

### DIFF
--- a/src/LiveComponent/src/ComponentWithFormTrait.php
+++ b/src/LiveComponent/src/ComponentWithFormTrait.php
@@ -13,8 +13,12 @@ namespace Symfony\UX\LiveComponent;
 
 use Symfony\Component\Form\ChoiceList\View\ChoiceGroupView;
 use Symfony\Component\Form\ClearableErrorsInterface;
+use Symfony\Component\Form\FormError;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
+use Symfony\Component\Form\Util\FormUtil;
+use Symfony\Component\Form\Util\ServerParams;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException;
 use Symfony\Contracts\Translation\TranslatableInterface;
 use Symfony\UX\LiveComponent\Attribute\LiveProp;
@@ -147,7 +151,7 @@ trait ComponentWithFormTrait
         }
     }
 
-    private function submitForm(bool $validateAll = true): void
+    private function submitForm(bool $validateAll = true, ?Request $request = null): void
     {
         if (null !== $this->formView) {
             // Two scenarios can cause this:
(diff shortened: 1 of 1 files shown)


smnandre:
Can you explain a bit more the problem you're solving ? 

Because this is working right now:

```
#[LiveAction]
public function myAction(Request $request): void
{
    $file = $request->files->get('my_file');
}
```

Obviously a file cannot be per se a LiveProp, as it cannot be in HTML, so I don't 100% get how it improves things ? 

Could you maybe add a test or two, with various situations (one file, multiple files, error on image, error on other field) so we ensure your code change does not impact other behaviour? 

If needed we can help :)

94noni:
to me the problem/issue here is a file bound to a form (via dto or entity mapping), not a direct file send via a live action
i had also once an issue, and I used a workaround to send the file after the form

florentdestremau:
> Can you explain a bit more the problem you're solving ?
> 
> Because this is working right now:
> 
> ```
> #[LiveAction]
> public function myAction(Request $request): void
> {
>     $file = $request->files->get('my_file');
> }
> ```

This is a hack because you can't fetch a file from a dto, you have to go manually digging into the request object. The code example I gave currently does not work, but it does on any classic form usage in a form submission through a controller.

The main thing here is that it allows for instance to upload files through a form class (with the FileType, or VIchUploadType, etc.). Currently this does not work because the request files are not hydrated into the form values.

If we agree on the issue I will gladly add more tests! The number of files does not matter much, it's the very principle behind that is IMO flawed by design if the request is not included in the form submission.

smnandre:
So i'll be 100% honest and transparent here: in **_my eyes_**, one should absolutey _**never**_ use anything but a dedicated endpoint to upload things. In the vast majority of cases, it leads to either security breach, weird UX (reminder: an upload file does not keep its file after submit, and there's a reason). And weird things on both validation, and ... "rollback" (it do believe it would break with your example..) 

BUT... obviously  [httos://github.com/smnandre' 

)# love to see an UX Upload around here (that's even the first one I started to draft / big pictures 18 months ago.)

Would you like to work on this? It does not need to be easy.... at lease can offer something, right now i cannot
 ok-j :)

florentdestremau:
To be fair, this is not really a matter of opinions: the symfony framework allows handling files inside forms, has always done so, and I don't see any plans on stopping. LiveComponents are presented as the better/modern way of doing forms instead of relying on js for some conditional display, etc for forms. Then they should support the same api, such as file management inside a form :)
This PR does just that, I haven't encountered any side effect _yet_ but I haven't really tried to find some. For "easy and typical" use cases (such as conditional forms and live collections) this works.

Kocal:

Thanks!

Status: Needs Work

PR #3115, opened 2025-09-30
[Map][Google] Upgrade @googlemaps/js-api-loader to ^2.0

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

Let's take advantage of the upcoming 3.0 to upgrade `@googlemaps/js-api-loader` to [^2.0.0](https://github.com/googlemaps/js-api-loader/releases/tag/v2.0.0).

It is possible to backport it on UX 2.x, but requires much work for no added-value.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -31,6 +31,12 @@ class MyLiveComponent {
 *  The Twig function `render_map()` has been removed, use `ux_map()` instead
 *  The option `title` from `Polygon`, `Polyline`, `Rectangle` and `Circle`, has been removed, use option `infoWindow` instead
 *  The property `rawOptions` from `ux:map:*:before-create` events has been removed, use `bridgeOptions` instead
+*  The Google Map Bridge has been upgraded to use `@googlemaps/js-api-loader` version `^2.0.0`:
+    * If you use Symfony AssetMapper without Symfony Flex, run `bin/console importmap:require @googlemaps/js-api-loader@^2.0`
+    * Options configurable through `UX_MAP_DSN` query params have changed:
+        * Option `version` has been renamed to `v`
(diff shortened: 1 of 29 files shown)


UPGRADE-3.0.md
```diff
@@ -31,6 +31,8 @@ class MyLiveComponent {
 *  The Twig function `render_map()` has been removed, use `ux_map()` instead
 *  The option `title` from `Polygon`, `Polyline`, `Rectangle` and `Circle`, has been removed, use option `infoWindow` instead
 *  The property `rawOptions` from `ux:map:*:before-create` events has been removed, use `bridgeOptions` instead
+*  The Google Map Bridge has been upgraded to use `@googlemaps/js-api-loader` version `^2.0.0`,
```
smnandre: Maybe we should be a bit more explicit on what user needs to do (or not) here ?

src/Map/src/Bridge/Google/assets/package.json
```diff
@@ -50,7 +50,7 @@
         }
     },
     "devDependencies": {
-        "@googlemaps/js-api-loader": "^1.16.6",
```
smnandre: If only we could do that in composer.json 👼

src/Map/src/Bridge/Google/src/Renderer/GoogleRenderer.php
smnandre, line 23: 👏

smnandre (approved):
Nice!

UPGRADE-3.0.md
Kocal: Done

PR #3117, opened 2025-10-01
[LiveComponent] Support default value for new collection item

IndraGunawan:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix 
| License        | MIT

when adding a new item to a collection of `LiveCollectionType`, it was always an empty value
somehow, defining `prototype_data` is not working

so this introduces a new method to set the default value of a collection item

### DIFF
--- a/src/LiveComponent/tests/Unit/LiveCollectionTraitTest.php
+++ b/src/LiveComponent/tests/Unit/LiveCollectionTraitTest.php
@@ -73,6 +73,23 @@ final class LiveCollectionTraitTest extends TestCase
                 ],
             ],
         ];
+        yield 'named parent, default value' => [
+            [
+                'parentName' => [
+                    'collectionNameDefault' => null,
+                ],
+            ],
+            'parentName[collectionNameDefault]',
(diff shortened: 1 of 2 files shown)


smnandre:
Hi @IndraGunawan ! 

Could you update the CHANGELOG and documentation ?

src/LiveComponent/tests/Unit/LiveCollectionTraitTest.php
```diff
@@ -221,6 +238,17 @@ protected function instantiateForm(): FormInterface
             {
                 return $this->theForm;
             }
+
+            protected function setCollectionItemDefaultValue(string $propertyPath, int $index): iterable
+            {
```
smnandre, line 243: Could you add an integration or functional test here ? 

(To prevent any future regression)

We can guide or help if you want to :)

Kocal:

Thanks!

Status: Needs Work

PR #3118, opened 2025-10-01
[Map] Display warning when trying to define a custom icon for a `Marker` that already has an `Icon`

### LINKED ISSUE #3116: [Map] Unable to customize the icon of a marker with bridgeOptions
In bridge documentations ([Google](https://github.com/symfony/ux-google-map?tab=readme-ov-file#customize-the-marker), [Leaflet](https://github.com/symfony/ux-leaflet-map?tab=readme-ov-file#customize-the-marker)), it state that we can customize the marker icon with `bridgeOptions`. 

However, based on this [line](https://github.com/symfony/ux/blob/2.x/src/Map/src/Bridge/Leaflet/assets/src/map_controller.ts#L156), it would seem that this is not the case.

For now my solution is to delete the `icon` key from the `definition` object:

```js
delete definition.icon;
```

But adding a check would improve DX!

```js
if (icon && !Object.hasOwn(rawOptions, 'icon') && !Object.hasOwn(bridgeOptions, 'icon')) {
    this.doCreateIcon({ definition: icon, element: marker });
}
```

ker0x:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | feature
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3116 
| License        | MIT

Add additional checks before creating the icon for a marker

### DIFF
--- a/src/Map/src/Bridge/Google/CHANGELOG.md
+++ b/src/Map/src/Bridge/Google/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.31
+
+-  Display a warning when trying to define `bridgeOptions.content` for a `Marker` that already has an `Icon`
+
 ## 2.30
 
 -  Ensure compatibility with PHP 8.5
--- a/src/Map/src/Bridge/Leaflet/CHANGELOG.md
+++ b/src/Map/src/Bridge/Leaflet/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.31
+
+-  Display a warning when trying to define `bridgeOptions.icon` for a `Marker` that already has an `Icon`
+
 ## 2.30
 
 -  Ensure compatibility with PHP 8.5
--- a/src/Map/src/Bridge/Google/assets/dist/map_controller.js
+++ b/src/Map/src/Bridge/Google/assets/dist/map_controller.js
@@ -248,6 +248,11 @@ var map_controller_default = class extends abstract_map_controller_default {
       this.createInfoWindow({ definition: infoWindow, element: marker });
     }
     if (icon) {
+      if (Object.prototype.hasOwnProperty.call(bridgeOptions, "content")) {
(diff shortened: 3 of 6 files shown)


Kocal:
Following https://github.com/symfony/ux/issues/3116#issuecomment-3358311885, I suggest instead log a warning message if we saw that the user tried to create an icon from the JavaScript side (through `bridgeOptions`) when `Marker#$icon` is not null

ker0x:
I updated the PR to trigger a warning, as you suggested 👍!

smnandre:
Should this be done on the PHP part ?

Kocal:
No because we want to verify that the user, through a Stimulus Controller, tried to add an icon by itself with `bridgeOptions`

src/Map/src/Bridge/Google/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 2.30
 
 -  Ensure compatibility with PHP 8.5
+-  Display a warning when trying to define `bridgeOptions.content` for a `Marker` that already has an `Icon`
```
ker0x:
```suggestion
## 2.31

-  Display a warning when trying to define `bridgeOptions.content` for a `Marker` that already has an `Icon`

## 2.30

-  Ensure compatibility with PHP 8.5
```
`2.30` has already been release!

src/Map/src/Bridge/Leaflet/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 2.30
 
 -  Ensure compatibility with PHP 8.5
+-  Display a warning when trying to define `bridgeOptions.icon` for a `Marker` that already has an `Icon`
```
ker0x:
```suggestion
## 2.31

-  Display a warning when trying to define `bridgeOptions.icon` for a `Marker` that already has an `Icon`

## 2.30

-  Ensure compatibility with PHP 8.5
```

Kocal:
Thank you @ker0x.

PR #3130, opened 2025-10-09
[TwigComponent] explain that default values are mandatory for props

alexislefebvre:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | -
| License        | MIT

I wanted to write `{% props icon, type %}` but it didn't work, because default values are mandatory for props.

So this PR make it explicit.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -702,7 +702,7 @@ You can also pass a variable (prop) into your template:
     </div>
 
 To tell the system that ``icon`` and ``type`` are props and not attributes, use the
-``{% props %}`` tag at the top of your template.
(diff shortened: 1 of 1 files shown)


Kocal:
This is not true, properties do not requires default values. 
If you define a property without a default value and that you forgot to use it, the exception `<prop> should be defined for component components/<component>.html.twig` is raised.

alexislefebvre:
> This is not true, properties do not requires default values. If you define a property without a default value and that you forgot to use it, the exception `<prop> should be defined for component components/<component>.html.twig` is raised.

Thanks, this is a better way to say it. If I update this PR with this content, would you consider reopening and merging it?

---

Update: I’ll try to take this into account in #3132

PR #3132, opened 2025-10-09
Document how to define mandatory/default props

alexislefebvre:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | -
| License        | MIT

Here is a “trick” to define default values for anonymous components.

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -700,12 +700,14 @@ You can also pass a variable (prop) into your template:
     </div>
 
 To tell the system that ``icon`` and ``type`` are props and not attributes, use the
-``{% props %}`` tag at the top of your template.
+``{% props %}`` tag at the top of your template. Props are required by default, but
+it is possible to set a default value with ``=``:
 
 .. code-block:: html+twig
 
     {# templates/components/Button.html.twig #}
-    {% props icon = null, type = 'primary' %}
+    {# prop "icon" is required, but prop "type" has a default value to "primary" #}
+    {% props icon, type = 'primary' %}
 
     <button {{ attributes.defaults({class: 'btn btn-'~type}) }}>
         {% block content %}{% endblock %}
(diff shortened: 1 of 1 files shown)


src/TwigComponent/doc/index.rst
```diff
@@ -718,6 +718,29 @@ To tell the system that ``icon`` and ``type`` are props and not attributes, use
...
+
+.. code-block:: html+twig
+
+    <twig:Image size="48"/>
+    <twig:Image/>
+    
+    {# renders as: #}
+    <img src="/image.jpg" width="48" height="48" alt="">
+    <img src="/image.jpg" width="24" height="24" alt="">
+
```
Kocal: I don't think we need a new section for default properties values, and the code examples seem wrong to me. What about changing the previous code block?

src/TwigComponent/doc/index.rst
Kocal, line 717: Instead, what about making `icon` prop required:
```suggestion
.. code-block:: html+twig

{# templates/components/Button.html.twig #}
{% props icon, type = 'primary' %}

<button {{ attributes.defaults({class: 'btn btn-'~type}) }}>
{% block content %}{% endblock %}
{% if icon %}
<span class="fa-solid fa-{{ icon }}"></span>
{% endif %}
</button>
```
and add some examples:
```twig
{# property "icon" is missing, an exception is thrown #}
<twig:Button>Share</twig:Button>

{# property "icon" is passed, property "type" use its default value "primary" #}
<twig:Button icon="share">Share</twig:Button> 

{# both properties "icon" and "type" are passed #}
<twig:Button icon="share" type="secondary></twig:Button>
```

Kocal (changes requested):

src/TwigComponent/doc/index.rst
alexisLefebvre: What do you think is wrong in these examples?

The previous code blocks rely on `{{ attributes.defaults({class: 'primary'}) }}`: https://symfony.com/bundles/ux-twig-component/current/index.html#anonymous-components I preferred to separate these 2 different usages of `props`.

src/TwigComponent/doc/index.rst
alexisLefebvre: Thanks, I added it below the current examples because I think that showing `props icon = null` + `{% if icon %}` is interesting too.

smnandre (approved):

Kocal (approved):

Kocal:
Thank you @alexislefebvre.

PR #3143, opened 2025-10-17
[LiveComponent] add controller tag and default metadata values for all service declaration types

FabienSalles:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no 
| Documentation? | no
| License        | MIT

In my project I don't use attributes like `#[AsLiveComponent]` in order to declare services and live components but I use yaml files (not my choice).
It works fine for all my services but live components are not well supported because we have to add extra parameters.

With  `#[AsLiveComponent]` attribute, we have default values for : 
* the route metadata : `'ux_live_component'`
* and the url reference type metadata `UrlGeneratorInterface::ABSOLUTE_PATH`

We also add `controller.service_arguments` tag in `LiveComponentExtension` class.

But if we don't use this attribute and we want to use another type of service declaration, we have to add all these information  : 
```yaml
services:
    MyServiceClass:
        tags:
            -
                name: twig.component
                live: true
                route: ux_live_component
                url_reference_type: 1
            -
                name: controller.service_arguments
```

If we do not, we have errors likes theses : 
```
An exception has been thrown during the rendering of a template ("Symfony\Component\Routing\Router::generate(): Argument #1 ($name) must be of type string, null given, called in /app/vendor/symfony/ux-live-component/src/Util/LiveControllerAttributesCreator.php on line 63")
```
```
An exception has been thrown during the rendering of a template ("Symfony\Component\Routing\Router::generate(): Argument #3 ($referenceType) must be of type int, null given, called in /app/vendor/symfony/ux-live-component/src/Util/LiveControllerAttributesCreator.php on line 63")
```
```
The controller for URI "/_components/my_customer_live_component" is not callable: Controller "MyCustomLiveComponentClass" cannot be fetched from the container because it is private. Did you forget to tag the service with "controller.service_arguments"?
```

These errors do not exist for php attribute declaration so I made this PR to have the same behavior regardless of how the live component is declared

### DIFF
--- a/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
+++ b/src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
@@ -57,7 +57,11 @@ class LiveControllerAttributesCreator
         $attributesCollection = $this->attributeHelper->create();
         $attributesCollection->setLiveController($mounted->getName());
 
-        $url = $this->urlGenerator->generate($metadata->get('route'), ['_live_component' => $mounted->getName()], $metadata->get('url_reference_type'));
+        $url = $this->urlGenerator->generate(
+            $metadata->get('route') ?? 'ux_live_component',
+            ['_live_component' => $mounted->getName()],
+            $metadata->get('url_reference_type') ?? UrlGeneratorInterface::ABSOLUTE_PATH
+        );
         $attributesCollection->setUrl($url);
 
         $liveListeners = AsLiveComponent::liveListeners($mounted->getComponent());
--- a/src/LiveComponent/src/DependencyInjection/Compiler/AddControllerTagToLiveComponentPass.php
+++ b/src/LiveComponent/src/DependencyInjection/Compiler/AddControllerTagToLiveComponentPass.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+final class AddControllerTagToLiveComponentPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        foreach ($container->findTaggedServiceIds('twig.component') as $id => $component) {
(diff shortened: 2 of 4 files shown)


Kocal:
Hi,

For the same reasons explained in https://github.com/symfony/ux/pull/3064, I'm :-1: here. The tag is purely internal and not documented, I don't think we want to encourage users to rely on it.

> In my project I don't use attributes like #[AsLiveComponent] in order to declare services and live components but I use yaml files (not my choice).

Why? Because of "DDD"?

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
```diff
@@ -57,7 +57,11 @@ public function attributesForRendering(MountedComponent $mounted, ComponentMetad
         $attributesCollection = $this->attributeHelper->create();
         $attributesCollection->setLiveController($mounted->getName());
 
-        $url = $this->urlGenerator->generate($metadata->get('route'), ['_live_component' => $mounted->getName()], $metadata->get('url_reference_type'));
+        $url = $this->urlGenerator->generate(
+            $metadata->get('route') ?? 'ux_live_component',
+            ['_live_component' => $mounted->getName()],
+            $metadata->get('url_reference_type') ?? UrlGeneratorInterface::ABSOLUTE_PATH
```
smnandre, line 63: Why this change?

FabienSalles: I'm not sure I understand the question (since I explained why in the PR description).

If we manually declare the service in the YAML file, we have to provide this additional information.

These default internal values ​​are precisely designed to prevent this.

FabienSalles:
Hi @Kocal ,

> Why? Because of "DDD"? 

We use DDD but it's not the reason. We could accept to use attributes in the infrastructure layer but we do not use autowiring and autoconfigure features (my client's rules).

And since this service declaration is still possible, I find it unfortunate that it cannot be used without exposing internal properties as I demonstrate.

src/LiveComponent/src/Util/LiveControllerAttributesCreator.php
smnandre: Sorry i missclicked on the line apparently.. i was asking about the AbsolutePath as default

FabienSalles: Hi,

I use the same defaults values as the attribute, visible here : https://github.com/symfony/ux/blob/2.x/src/LiveComponent/src/Attribute/AsLiveComponent.php#L51

FabienSalles: The `method:POST` is also mandatory in some cases

Kocal:

Thanks!

Status: Needs Work

PR #3145, opened 2025-10-17
[Toolkit][Shadcn] Add Spinner component

bernard-ng:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| License        | MIT

Add Spinner component, An indicator that can be used to show a loading state.
refs: https://ui.shadcn.com/docs/components/spinner

### DIFF
--- a/src/Toolkit/kits/shadcn/spinner/EXAMPLES.md
+++ b/src/Toolkit/kits/shadcn/spinner/EXAMPLES.md
@@ -1,0 +1,32 @@
+# Examples
+
+## Default
+
+```twig {"preview":true}
+<twig:Button disabled>
+    <twig:Spinner /> Please wait
+</twig:Button>
+```
+
+## Size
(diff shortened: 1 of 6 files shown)


Kocal:
Thanks for the contribution, that's much appreciated!

I've pushed on your PR some modifications:
- it was not possible to use `<twig:Spinner class="...">`, the `class` attribute was fully overridden by `{{ ...attributes }}` (looks like the `ComponentAttributes::getIterator()` does not handle `ComponentAttributes::$rendered` property, like `ComponentAttributes::__toString()`) and so the Spinner was not rotating anymore. Using `{{ ...attributes.without('class') }}` fixed the issue
- I added some examples with size and color

Kocal:
Thank you @bernard-ng.

PR #3150, opened 2025-10-28
[LiveComponent] Support for liveProp union type

### LINKED ISSUE #853: [LiveComponent] Support UnionTypes for LiveProps

kevinmade:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #853
| License        | MIT

This PR adds support for liveProp union types in Live Components. Previously, using a union type would throw an exception. With this change `LiveComponentMetadataFactory` will be able to create liveProp metadata for union types, but will only work in combination with a custom hydrate and dehydrate implementation.


Kocal (changes requested):
Hi and thanks for the contribution!

This new feature needs to be covered with some tests, and the CHANGELOG.md must be updated too.

Also, you only support the TypeInfo path and not the legacy PropertyInfo path, but I think that's fine.

src/LiveComponent/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 2.32
+
+- Add support for `LiveProp` union types in Live Components.
+  `LiveComponentMetadataFactory` can now create `liveProp` metadata for union types.
+  Note: This feature only works when using a custom `hydrate` and `dehydrate` implementation.
```
smnandre: Hey @kevinmade ! Could you add a test or two covering the cases "not supposed to work" (ie: when hydrate and/or dehydrate are missing ?)

kevinmade: I thought a hydrator and dehydrator were required for my change, but this had something to do with my own use case, where the union had a type that wasn't a hydratable. I’ve removed the note from the changelog and added a test.

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -1636,6 +1636,23 @@ public function testInterfaceTypedLivePropCannotBeDehydrated()
...
+
+        $component = new $componentClass();
+        $component->object = new class {
+        };
+
+        $this->expectException(\LogicException::class);
+        $this->expectExceptionMessageMatches('/The "object" property on component ".*" is missing its property-type. Add the ".*" type so the object can be hydrated later./');
+
+        $this->hydrator()->dehydrate($component, $this->createComponentAttributes(), $this->createLiveMetadata($component));
+    }
```
Kocal, line 1657: To move in `\Symfony\UX\LiveComponent\Tests\Integration\LiveComponentHydratorTest::provideDehydrationHydrationTests`

src/LiveComponent/tests/Functional/Metadata/LiveComponentMetadataFactoryTest.php
```diff
@@ -42,4 +43,21 @@ public function testQueryStringMapping()
...
+        $propsMetadata = $metadataFactory->createPropMetadatas($class);
+
+        $propsMetadataByName = [];
+        foreach ($propsMetadata as $propMetadata) {
+            $propsMetadataByName[$propMetadata->getName()] = $propMetadata;
+        }
+
+        $this->assertEquals('mixed', (string) $propsMetadataByName['unionProp']->getType());
+        $this->assertEquals('mixed', (string) $propsMetadataByName['intersectionProp']->getType());
+    }
```
Kocal, line 62: I don't think it makes sense to have this test

src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
```diff
@@ -122,6 +123,10 @@ public function createLivePropMetadata(string $className, string $propertyName,
                 $collectionValueType
             );
         } else {
+            if ($reflectionType instanceof \ReflectionUnionType || $reflectionType instanceof \ReflectionIntersectionType) {
+                return new LivePropMetadata($property->getName(), $liveProp, Type::mixed());
+            }
```
Kocal: Why do we use `Type::mixed()` here, instead of `$this->typeResolver->resolve($property)` below?

Kocal (changes requested):
Please add tests in `LiveComponentHydratorTest::provideDehydrationHydrationTests` data provider, so we can test the full dehydration/hydration process. Testing only the metadata factory is not enough.

PR #3155, opened 2025-10-31
[Autocomplete] Fix issue where TomSelect could already been initialized on an element

### LINKED ISSUE #2623: [Autocomplete] Error: Tom Select already initialized on this element
I get the following error message when moving a TomSelect [Autocomplete] in the DOM.
This after upgrading from `2.21.0` to `2.22.0` (and still present in `2.23.0`)

```
Error connecting controller

 Error: Tom Select already initialized on this element
    TomSelect tom-select.ts:104
    _default_1_createTomSelect2 controller.js:372
    _default_1_createAutocomplete2 controller.js:296
```

The  code I use to move the element:
```javascript
    filterModal.querySelector('.modal-body').appendChild(autocompleteElement)
    document.querySelector('body').appendChild(filterModal)
```

daFish:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #2623 
| License        | MIT

**Problem**

When Autocomplete elements are removed and re-added to the DOM (e.g., with Turbo navigation), a race condition causes
"Tom Select already initialized on this element" error.

**Root Cause**

`disconnect()` destroys TomSelect but leaves `this.tomSelect` pointing to the destroyed instance. If `urlValueChanged()`
fires before `connect()` (during re-attachment), it calls `resetTomSelect()` on the stale reference, triggering
double-initialization.

**Impact**

- Affects: Users with Turbo navigation, especially on mobile (Safari iOS)
- Symptoms: "Tom Select already initialized" console errors during navigation
- Fix: Single line addition prevents the race condition
- Risk: None - only affects the disconnect cleanup path

### DIFF
--- a/apps/e2e/src/Form/Type/AutocompleteSelectType.php
+++ b/apps/e2e/src/Form/Type/AutocompleteSelectType.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace App\Form\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class AutocompleteSelectType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder->add(
+            'favorite_fruit',
+            ChoiceType::class,
+            [
+                'choices' => [
+                    'Apple' => 'apple',
+                    'Banana' => 'banana',
+                    'Cherry' => 'cherry',
+                    'Coconut' => 'coconut',
+                    'Grape' => 'grape',
+                    'Kiwi' => 'kiwi',
+                    'Lemon' => 'lemon',
+                    'Mango' => 'mango',
+                    'Orange' => 'orange',
+                    'Papaya' => 'papaya',
+                    'Peach' => 'peach',
+                    'Pineapple' => 'pineapple',
+                    'Pear' => 'pear',
+                    'Pomegranate' => 'pomegranate',
+                    'Pomelo' => 'pomelo',
+                    'Raspberry' => 'raspberry',
+                    'Strawberry' => 'strawberry',
+                    'Watermelon' => 'watermelon',
+                ],
+                'autocomplete' => true,
+                'label' => 'Your favorite fruit:'
+            ]
+        );
+    }
+}
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -46,7 +46,7 @@ export default class extends Controller {
     declare readonly tomSelectOptionsValue: object;
     declare readonly hasPreloadValue: boolean;
     declare readonly preloadValue: string;
-    tomSelect: TomSelect;
+    tomSelect: TomSelect | undefined;
 
     private mutationObserver: MutationObserver;
     private isObserving = false;
@@ -98,6 +98,10 @@ export default class extends Controller {
     disconnect() {
         this.stopMutationObserver();
 
+        if (!this.tomSelect) {
+            return;
+        }
+
         // TomSelect.destroy() resets the element to its original HTML. This
         // causes the selected value to be lost. We store it.
         let currentSelectedValues: string[] = [];
@@ -114,6 +118,7 @@ export default class extends Controller {
         }
 
         this.tomSelect.destroy();
+        this.tomSelect = undefined;
 
         if (this.selectElement) {
             if (this.selectElement.multiple) {
@@ -163,11 +168,15 @@ export default class extends Controller {
             plugins,
             // clear the text input after selecting a value
             onItemAdd: () => {
-                this.tomSelect.setTextboxValue('');
+                this.tomSelect?.setTextboxValue('');
             },
             closeAfterSelect: true,
             // fix positioning (in the dropdown) of options added through addOption()
             onOptionAdd: (value: string, data: { [key: string]: any }) => {
+                if (!this.tomSelect) {
+                    return;
+                }
+
                 let parentElement = this.tomSelect.input as Element;
                 let optgroupData = null;
 
@@ -232,10 +241,10 @@ export default class extends Controller {
         const config = this.#mergeConfigs(commonConfig, {
             maxOptions: this.getMaxOptions(),
             score: (search: string) => {
-                const scoringFunction = this.tomSelect.getScoreFunction(search);
+                const scoringFunction = this.tomSelect?.getScoreFunction(search);
                 return (item: any) => {
                     // strip HTML tags from each option's searchable text
-                    return scoringFunction({ ...item, text: this.#stripTags(item[labelField]) });
+                    return scoringFunction?.({ ...item, text: this.#stripTags(item[labelField]) });
                 };
             },
             render: {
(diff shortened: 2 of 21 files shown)


daFish:
@Kocal I took the chance to add playwright tests. I'll submit another PR with some changes to the documentation to enhance the DX.

src/Autocomplete/assets/test/browser/placeholder.test.ts
```diff
@@ -5,3 +5,48 @@ test('Can see homepage', async ({ page }) => {
 
     await expect(page.getByText("Symfony UX's E2E App")).toBeVisible();
 });
```
Kocal, line 7: Can now be removed

src/Autocomplete/assets/test/browser/placeholder.test.ts
Kocal, line 1: Can now be renamed to `autocomplete.test.ts`

src/Autocomplete/assets/test/browser/placeholder.test.ts
```diff
@@ -5,3 +5,48 @@ test('Can see homepage', async ({ page }) => {
...
+
+test('disconnect/reconnect does not cause double-initialization error (issue #2623)', async ({ page }) => {
```
Kocal: I don't understand how this test could work, the file `apps/e2e/src/Controller/AutocompleteController.php` must be updated in order to render an autocomplete field

Kocal (changes requested):
Thanks Marcus, that's much appreciated! 

I only have some comments about the E2E test, and of course any documentation improvement is welcomed :)

src/Autocomplete/assets/test/browser/placeholder.test.ts
daFish: Yeah, the test works only because it checks if the error is not emited. In this case the error will never be emitted because there isn't even a autocomplete field.
I'll add an autocomplete field on a select-field. Maybe also making it so that the error would be triggered.

daFish: This one is more tricky than initially thought. I was unable to come up with an example which would trigger this behaviour.
I will remove the test because it does not add any value. What will remain are the examples for the E2E-setup.

daFish:
* removed the browser test because I couldn't find a way to reproduce it here
* added a simple non-ajax example for Autocomplete

daFish:
Ok, this is weird. I have force-pushed my changes this morning and the PR is still not updated. My changes are still being processed. 🤨

daFish:
Rebased against 2.x and force-pushed again. Works now.

daFish:
@Kocal After several attempts trying to reproduce the behaviour locally, I was unable to provoce it. How shall we move forward?

apps/e2e/src/Form/Type/AutocompleteSelectType.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
```
smnandre:
```suggestion
```

apps/e2e/src/Form/Type/AutocompleteSelectType.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace App\Form\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class AutocompleteSelectType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
```
smnandre:
```suggestion
    public function buildForm(FormBuilderInterface $builder, array $options): void
```

src/Autocomplete/assets/src/controller.ts
```diff
@@ -163,11 +168,15 @@ export default class extends Controller {
             plugins,
             // clear the text input after selecting a value
             onItemAdd: () => {
-                this.tomSelect.setTextboxValue('');
+                this.tomSelect?.setTextboxValue('');
```
smnandre, line 171: This is not supposed to happen, is it ?

src/Autocomplete/assets/src/controller.ts
```diff
@@ -163,11 +168,15 @@ export default class extends Controller {
...
             },
             closeAfterSelect: true,
             // fix positioning (in the dropdown) of options added through addOption()
             onOptionAdd: (value: string, data: { [key: string]: any }) => {
+                if (!this.tomSelect) {
+                    return;
+                }
```
smnandre, line 178: same thing

src/Autocomplete/assets/src/controller.ts
```diff
@@ -295,7 +304,7 @@ export default class extends Controller {
             },
             optgroupField: 'group_by',
             // avoid extra filtering after results are returned
-            score: (search: string) => (item: any) => 1,
+            score: (_search: string) => (_item: any) => 1,
```
smnandre, line 307: why this change ? genuine question here

src/Autocomplete/assets/src/controller.ts
```diff
@@ -439,6 +448,10 @@ export default class extends Controller {
     }
 
     private changeTomSelectDisabledState(isDisabled: boolean): void {
+        if (!this.tomSelect) {
+            return;
+        }
+
```
smnandre, line 454: Same ting, not supposed to happen here

smnandre (commented):
A couple of comments passing by, thank you for the PR!

src/Autocomplete/assets/src/controller.ts
daFish: The underscore is used to indicate an unused variable while making the linter happy.

smnandre: Was he unhappy ? never observed that .. but ok :)

src/Autocomplete/assets/src/controller.ts
daFish: We can make it more defensive when something like this happens.

src/Autocomplete/assets/src/controller.ts
daFish: See above.

src/Autocomplete/assets/src/controller.ts
daFish: 🤷🏻‍♂️ It's not easy being a linter I guess.

src/Autocomplete/assets/src/controller.ts
daFish: Will be reverted.

daFish:
@smnandre The defensive checks can remain IHMO. We tell other developers that all code paths are safe. Yes, it should not happen but we make it explicit that it will not happen. Especially in the callbacks, e.g. `onItemAdd` or `onOptionAdd`. WDYT?

daFish:
I was able to reproduce this locally and will add an E2E test in the coming days. Let's get this over the finishing line.

daFish:
@Kocal @smnandre This can be reviewed again. However, the tests do obviously pass because the fix is in place.

Kocal (approved):

Kocal:
Thank you @daFish!

PR #3158, opened 2025-11-04
[Toolkit][Shadcn] Update Button CSS class, remove as props

ker0x:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

This PR update the CSS class of the `Button` component according to official Shadcn documentation.

It also removes the `as` props and relies on the presence of an `href` attribute to determine which element to use.


Kocal:
Thanks Romain that's super appreciated! I'll take a good look at all of this tonight :)

Kocal:
Hey, I was finally able to take some time on your contribution, I did few minor changes, but everything is good:
1. I re-introduced the `as` tag, I prefer to make it explicit and this behavior is closer to `asChild` from React
2. I modified some examples to reflect the current Shadcn's Button documentation

Kocal:
Thank you @ker0x.

PR #3173, opened 2025-11-16
[Toolkit][Shadcn] Add `Dialog` component

seb-jean:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes
| Issues         | Fix #... 
| License        | MIT

This pull request introduces the new `Dialog` component to the Shadcn UI kit in the Symfony UX Toolkit, including its templates for all subcomponents (Close, Content, Description, Footer, Header, Title, Trigger), JavaScript controller, documentation, and manifest.

I encountered a display issue with the `Label` component. So I updated this component to reflect [Shadcn](https://ui.shadcn.com/docs/components/label). 

Below is a demonstration of the `Dialog` component:

https://github.com/user-attachments/assets/184036ed-1280-40fe-b126-06a288b257dc

However, I am having a problem with the `lucid-x` icon. In the middle of it, there is what looks like a white line.

lucide-x

The second problem I see is that when using the Tab key while the dialog is open, elements outside the dialog can also be focused. This is unacceptable.

What do you think?
(In any case, this kit system is great, and it's only the beginning.)


Kocal:
Hey, and thanks for the contribution! That's super appreciated 🤗 

I don't see any issue with the closing icon, it's well displayed: 
image

About the "focus trap" thing, I agree. I didn't see native things and the solution used in by Shadcn/RadixUI seems very complex. 
Can we give a try to https://github.com/focus-trap/focus-trap? It looks like super easy to use.

Thanks!

src/Toolkit/kits/shadcn/dialog/templates/components/Dialog/Content.html.twig
```diff
@@ -0,0 +1,48 @@
+{# @prop showCloseButton boolean Whether the close button is shown, default to `true` #}
+{# @block content The default block #}
+{%- props showCloseButton = true -%}
```
Kocal, line 3:
```suggestion
{# @block content The default block #}
```

I don't see any `showCloseButton` prop ShadcnUI side, and more generally I don't think that's a good a11y. Can we remove it?

src/Toolkit/kits/shadcn/dialog/templates/components/Dialog.html.twig
```diff
@@ -0,0 +1,15 @@
+{# @prop open boolean Open (or not) the Dialog at initial rendering, default to `false` #}
```
Kocal, line 1: Missing `@prop` for `id`

seb-jean: I just added it.

src/Toolkit/kits/shadcn/dialog/templates/components/Dialog/Content.html.twig
seb-jean: When I added the component locally, I saw the `showCloseButton` prop. It can also be seen here: https://github.com/shadcn-ui/ui/blob/main/apps/v4/registry/new-york-v4/ui/dialog.tsx#L52

What do you think?

seb-jean:
> Hey, and thanks for the contribution! That's super appreciated 🤗
> 
> I don't see any issue with the closing icon, it's well displayed: image
> 
> About the "focus trap" thing, I agree. I didn't see native things and the solution used in by Shadcn/RadixUI seems very complex. Can we give a try to https://github.com/focus-trap/focus-trap? It looks like super easy to use.
> 
> Thanks!

I'll look into improving the focus section.

src/Toolkit/kits/shadcn/dialog/templates/components/Dialog/Content.html.twig
Kocal: Aaah my bad, I only read the Shadcn documentation and didn't see any hint about `showCloseButton`. That's fine then, let's keep it!

norkunas:
I think you should remove the `open` attribute from the `<dialog>` element itself, and let to control `dialog-controller` with `data-dialog-open-value="boolean"`, this way focus won't escape outside of dialog

seb-jean:
> About the "focus trap" thing, I agree. I didn't see native things and the solution used in by Shadcn/RadixUI seems very complex. Can we give a try to https://github.com/focus-trap/focus-trap? It looks like super easy to use.
@Kocal 


> I think you should remove the open attribute from the <dialog> element itself, and let to control dialog-controller with data-dialog-open-value="boolean", this way focus won't escape outside of **dialog**
> 
> Thanks!
@norkunas 

I've made improvements to enable animations using CSS with [@starting-style](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#starting-style), simplifying the HTML and Stimulus controller. Focus is handled by default with the `dialog` element, but some CSS classes conflicted.

I also noticed that when the dialog box opens, the focusable element is now focused. That's great.

I hope you appreciate this improvement :).

Kocal:
> I think you should remove the `open` attribute from the `<dialog>` element itself, and let to control `dialog-controller` with `data-dialog-open-value="boolean"`, this way focus won't escape outside of dialog

I liked the fact that you can open the dialog by default, but using `*:backdrop:` CSS classes does not work because `<dialog open>` (at boot page) does not display the backdrop anymore, see. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#open

I will this `open` attribute

Kocal:
Thanks @seb-jean, that's super appreciated :)

PR #3196, opened 2025-11-28
[TwigComponent] Fix loading Twig Components by FQCN

### LINKED ISSUE #3195: [TwigComponents] Using FQCN to load component leads to component being loaded anonymously
According to the docs, you can use both, the actual class name and a special `:` syntax to load a twig component. This worked until https://github.com/symfony/ux/commit/da0537d96182b40f60ac9d1dd4b961fa3c605011:

Internally, the `ComponentFactory` would use the `:` syntax to store config per component, but it also has a `$this->classMap`, which maps the actual class name to the `:` version. Before doing anything, the `ComponentFactory` would ask the `classMap` to resolve the loaded component name, only then it would proceed with loading.

This changed in da0537d96182b40f60ac9d1dd4b961fa3c605011: Now, it first tries to load the component with the provided name directly, and if that doesn't work, it will load the component anonymously (which means: the class will be ignored, only the Twig template will be loaded).

There is still code that uses the `classMap`, but that will only be executed if the template can't be loaded anonymously. ~~Which I think will never happen, because any component has a Twig template? At least, it will never happen if the component has a Twig template, because unlike `ComponentFactory` after da0537d96182b40f60ac9d1dd4b961fa3c605011, `ComponentTemplateFinder` is able to resolve FQCNs.~~

It turned out I was wrong about Twig always being able to load a template for an FQCN: It is only able to load a template anonymously for the FQCN if the template path equals the FQCN (with backslashes replaced with slashes and ignoring the file extension). This is the case for all templates in my project, but I think it is not for the Symfony default. Because this issue only occurs in that case, it is not easy to reproduce. I created a reproducer in this repo: https://github.com/janopae/reproduce-symfony-ux-540

So essentially, identifying the component by class name now leads to the class being ignored. Given the tags containing da0537d96182b40f60ac9d1dd4b961fa3c605011, this affects all versions from `2.20.0` to `2.31.0`.

janopae:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3195 
| License        | MIT

When loading Twig Components, code needs a way to identify the component. You can use the identifiers that `ComponentFactory` internally uses (this will usually be an identifier consisting of the namespaces separated by `:`). But as the docs still state, past versions of ux-twig-component could also load them by the fully-qualified PHP class name of the component class.

da0537d96182b40f60ac9d1dd4b961fa3c605011 broke this: From this commit on, `ComponentFactory`, which is responsible for resolving the identifier the user provided to a Twig template and a class, would only attempt to resolve an FQCN if it couldn't provide metadata for loading the component anonymously (which means without a class, just the Twig template). ~~Because loading the component anonymously always works as long as there is a Twig template,~~ In case the FQCN could also be resolved as a Twig template, `ComponentFactory` would not resolve FQCNs to a class component, and instead load only the corresponding Twig template as an anonymous component.

The issue affects all versions from  `2.20.0` to `2.31.0`.

This PR fixes this by trying to resolve the FQCN right at the beginning, makig sure the first attempt to load the component with the class already happens using the right identifier.

### DIFF
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -46,6 +46,8 @@ final class ComponentFactory implements ResetInterface
 
     public function metadataFor(string $name): ComponentMetadata
     {
+        $name = $this->classMap[$name] ?? $name;
+
         if ($config = $this->config[$name] ?? null) {
             return new ComponentMetadata($config);
         }
@@ -57,14 +59,6 @@ final class ComponentFactory implements ResetInterface
             ];
 
             return new ComponentMetadata($this->config[$name]);
-        }
-
-        if ($mappedName = $this->classMap[$name] ?? null) {
-            if ($config = $this->config[$mappedName] ?? null) {
-                return new ComponentMetadata($config);
-            }
-
-            throw new \InvalidArgumentException(\sprintf('Unknown component "%s".', $name));
         }
 
         $this->throwUnknownComponentException($name);
--- a/src/TwigComponent/tests/Unit/ComponentFactoryTest.php
+++ b/src/TwigComponent/tests/Unit/ComponentFactoryTest.php
@@ -93,4 +93,32 @@ class ComponentFactoryTest extends TestCase
         $this->assertSame('foo', $metadata->getName());
         $this->assertSame('foo.html.twig', $metadata->getTemplate());
     }
+
+    /**
+     * @testWith ["bar"]
+     *           ["Foo\\Bar"]
+     */
+    public function testPrefersClassComponentOverAnonymous(string $name)
+    {
+        $templateFinder = $this->createMock(ComponentTemplateFinderInterface::class);
+        $templateFinder->expects($this->never())
+            ->method('findAnonymousComponentTemplate');
+
+        $factory = new ComponentFactory(
+            $templateFinder,
+            $this->createMock(ServiceLocator::class),
+            $this->createMock(PropertyAccessorInterface::class),
+            $this->createMock(EventDispatcherInterface::class),
+            [
+                'bar' => ['key' => 'bar', 'template' => 'bar.html.twig'],
+            ],
+            ['Foo\\Bar' => 'bar'],
+            $this->createMock(Environment::class),
+        );
+
+        $metadata = $factory->metadataFor($name);
+
+        $this->assertSame('bar', $metadata->getName());
+        $this->assertSame('bar.html.twig', $metadata->getTemplate());
+    }
 }


janopae:
I couldn't find a PHP-CS-Fixer config, so I just tried to match the coding style I saw in the rest of the code the best I could

smnandre:
> past versions of ux-twig-component could also load them by the fully-qualified PHP class name of the component

I have no memory of that ... where do we wrote this ?

janopae:
> I have no memory of that ... where do we wrote this ?

The docs mention it for test code, and this is where the bug occured to me – it broke my tests. https://symfony.com/bundles/ux-twig-component/current/index.html#test-helpers

It seems like `ComponentFactory` has a special [`classMap` property](https://github.com/symfony/ux/blob/0c63ac0a3e3aaccf66e20c4aa9b96d680ec6ed88/src/TwigComponent/src/ComponentFactory.php#L42) which gets populated by [`TwigComponentPass`](https://github.com/symfony/ux/blob/0c63ac0a3e3aaccf66e20c4aa9b96d680ec6ed88/src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php#L83) for this sole purpose.

If we come to the conclusion that support for FQCNs for TwigComponents should be dropped, we should remove the population of that property as well as everything in `ComponentFactory::metadataFor` below https://github.com/symfony/ux/blob/0c63ac0a3e3aaccf66e20c4aa9b96d680ec6ed88/src/TwigComponent/src/ComponentFactory.php#L62

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -59,14 +61,6 @@ public function metadataFor(string $name): ComponentMetadata
             return new ComponentMetadata($this->config[$name]);
         }
 
-        if ($mappedName = $this->classMap[$name] ?? null) {
-            if ($config = $this->config[$mappedName] ?? null) {
-                return new ComponentMetadata($config);
-            }
-
-            throw new \InvalidArgumentException(\sprintf('Unknown component "%s".', $name));
-        }
```
WebMamba, line 68: I don't understand if you remove all of this this method only search for anonymous component then? It's mean something is wrong with the current architecture and we don't need to separate the finding of the template of the anonymous component and the normal component

src/TwigComponent/tests/Unit/ComponentFactoryTest.php
```diff
@@ -93,4 +93,32 @@ public function testMetadataForReuseAnonymousConfig()
         $this->assertSame('foo', $metadata->getName());
         $this->assertSame('foo.html.twig', $metadata->getTemplate());
     }
+
+    /**
+     * @testWith ["bar"]
+     *           ["Foo\\Bar"]
+     */
```
WebMamba, line 100: you can remove the comment here

WebMamba (commented):
I need to dig more into this I don't understand how this could work. Looks something is broken somewhere

src/TwigComponent/src/ComponentFactory.php
smnandre: Of course we do! One is compilable once, the other is/may  not be :)

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -59,14 +61,6 @@ public function metadataFor(string $name): ComponentMetadata
             return new ComponentMetadata($this->config[$name]);
         }
 
-        if ($mappedName = $this->classMap[$name] ?? null) {
```
smnandre, line 62: classMap is usefull

janopae: That's why this PR proposes to evaluate it before, not after trying to load the component anonymously

src/TwigComponent/tests/Unit/ComponentFactoryTest.php
janopae: I don't understand. Should I replace [`@testWith`](https://docs.phpunit.de/en/10.5/annotations.html#testwith) with an attribute or a dataprovider?

src/TwigComponent/src/ComponentFactory.php
janopae: > I don't understand if you remove all of this this method only search for anonymous component then? 

No! The first thing it does is trying to create a "normal" component. (see https://github.com/symfony/ux/pull/3196/files#r2591624694) Only if this doesn't work, it tries to create an anonymous component.

Opposed to how it worked before:

First, it tried to create a "normal" component. If that didn't work using the identifier, it tried to load an anonymous component. And only if both of those didn't work, it had the same code for loading a "normal" component again, just this time with the resolved class map.

Now we resolve the class map from the beginning on, so we don't need to repeat the code just to do the same thing again with the name from the resolved class map.

src/TwigComponent/tests/Unit/ComponentFactoryTest.php
Kocal: The comment must be kept for PHPUnit 9 compatibility, but `#[TestWith(...)]` must be added for PHPUnit 10+ (UX 3.x)

smnandre:
-- Discussion ongoing in issue https://github.com/symfony/ux/issues/3195  --

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -59,14 +61,6 @@ public function metadataFor(string $name): ComponentMetadata
             return new ComponentMetadata($this->config[$name]);
```
janopae, line 61: ~~This is the place where "normal" (non-anonymous) components get created. If we resolve the classMap first and save it in `$name`, there's no need for the duplication I removed.~~

smnandre: This is not the place where they are created, as you can see this does only concern anonymous ones.

janopae: I added this comment to the wrong place.

This is the correct one: https://github.com/symfony/ux/pull/3196#discussion_r2591624694

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -46,6 +46,8 @@ public function __construct(
 
     public function metadataFor(string $name): ComponentMetadata
     {
+        $name = $this->classMap[$name] ?? $name;
+
         if ($config = $this->config[$name] ?? null) {
             return new ComponentMetadata($config);
         }
```
janopae, line 53: This is the place where "normal" (non-anonymous) components get created. If we resolve the classMap first and save it in `$name`, there's no need for the duplication I removed.

smnandre:
Let's close here, original issue fixed by a configuration update.

PR #3197, opened 2025-11-29
[Toolkit][Badge] Remove hover style to clearly distinguish from Button

tassilogroeper:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | 
| License        | MIT

Confusion may occur, if the badge shows a different style on mouse hover. As it is not a button, we need to remove it.

Additionally, the badge component on Shadcn does not hold hover styles, when interacted with a mouse. see  https://ui.shadcn.com/docs/components/badge


Kocal:
Thanks @tassilogroeper!

PR #3208, opened 2025-12-03
[Translator] Refactor API to use string-based translation keys instead of generated constants

### LINKED ISSUE #2619: [Translator] Message key syntax and object const for javascript UX Translator
I am delegating translations to the front-end by sending the message through REST API requests to a React component.

The problem I'm facing is that I am forced to create a decoder ring that seems like boilerplate code: and pretty tedious:

```js
import {
    trans,
...
    OTP_FIELD_LABEL,
    OTP_EXCEPTIONS_USERNAME_NOT_FOUND
} from '../../../translator';

const messages = {
    'OTP_EXCEPTIONS_USERNAME_NOT_FOUND' : OTP_EXCEPTIONS_USERNAME_NOT_FOUND
};
```
That decoder ring is `const messages`.

While reading the documentation, I noticed two things:

https://symfony.com/doc/current/translation.html#using-real-or-keyword-messages

The first is that I could refer to this translation in the backend as "otp.exceptions.username_not_found".

The second is that, I wrote the translations in YAML... had I written it in PHP I could send the whole translation file as a JSON object (not efficent obviously.

While importing consts from the translator is usable, it is also an error prone process, and more than once I've found myself using a const that I didn't import.

The feature request is to either or both be able to:

1) Use `trans()` with the key notation and not having to import consts:

```js
trans("otp.exceptions.username_not_found")
```

2) Have a single objects to import, instead of all the consts:

```js
import{
   trans,
   otp, //or OTP, but lowercase would make it consistent
} from '../../../translator';

...

trans(otp.exceptions.username_not_found);

...

trans(otp.field.label);
```

Why I think this is important:

1. As it turned out, the decoder ring is sort of worth the effort because I have REST API unit tests (well, integration tests)... and sending the key notation as the message to delegate translation makes my tests very anti-fragile / not brittle / resilient.

I could be switching around the text and even the variables inside the text (I'm using the ICU format) and as long as the message id / structure remains the same, the tests don't break

2. Either the key notation or the const object adds consistency, and also reduce redundant work. As pretty as the caps are (e.g. "const OTP_EXCEPTIONS_USERNAME_NOT_FOUND"... this is totally different than what one would use in PHP. 

One of the goals of Symfony UX Translator (I assume) is to reuse translations in the front-end to reduce redundant work. The object const notation not only would make it look consistent with the key notation in PHP, but also reduce cognitive load and boilerplate code (multiple imports).

Boilerplate code not only makes the task longer, boring, but is also error prone.

3.  Having the key notation available in Javascript would mean that one wouldn't even have to import the translation consts (could be imported in the background). This would make it super simple to understand the process, help improve API consistency when the translations are being delegated by the backend and make the whole process less error proof.

4. There is a term in the manufacturing world in quality management systems for "error-proofing" called [poka-yoke](https://www.edstellar.com/blog/types-of-poka-yoke-techniques). Error-proofing would essentially translate to Developer Experience :)

Image

## In summary

This feature would lead to more consistent, less error prone, less boilerplate and less cognitive load while working with front-end translation, and also make it very easy to delegate translations from the backend while making enabling more anti-fragile API testing.

I think this feature can be added without compatibility issues by enabling something like this for the key notation:

```js
trans.key("otp.exceptions.username_not_found", ...)
```

And for the const objects, just generating that code in addition to the uppercase consts:

```js

const otp = {
  field : {
    label: OTP_FIELD_LABEL    
  },
  exceptions: {
      username_not_found: OTP_EXCEPTIONS_USERNAME_NOT_FOUND
  }
};

```

```js

import {trans, otp } from '../../../translator';

...

trans(otp.exceptions.username_not_found);

...

trans(otp.field.label);


```

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #2619 
| License        | MIT

This pull request introduces a **breaking change** to the UX Translator component, refactoring its API to use string-based translation keys instead of generated constants. 

The main advantages are:
- You can now use **exactly the same translation keys** as in your Symfony PHP code
- Simpler and more readable code
- No need to memorize generated constant names
- No need to import translation constants: smaller files
- And you can still get autocompletion and type-safety :rocket:

**Before:**
```typescript
import { trans } from '@symfony/ux-translator';
import { SYMFONY_GREAT } from '@app/translations';

trans(SYMFONY_GREAT);
```

**After:**
```typescript
import { createTranslator } from '@symfony/ux-translator';
import { messages } from '../var/translations/index.js';

const { trans } = createTranslator({ messages });
trans('symfony.great');
```

Changing the import from `@app/translations` to `../var/translations/index.js` allows IDEs and other tools to correctly interpret these files, which is required to make `trans()` type-safe.

It also make the configuration easier for AssetMapper users, they must now remove these entries from their `importmap.php`: 
```php
   '@app/translations' => [
        'path' => './var/translations/index.js',
    ],
    '@app/translations/configuration' => [
        'path' => './var/translations/configuration.js',
    ],
```


apps/e2e/importmap.php
```diff
@@ -139,12 +139,6 @@
     '@symfony/ux-translator' => [
         'path' => './vendor/symfony/ux-translator/assets/dist/translator_controller.js',
     ],
-    '@app/translations' => [
-        'path' => './var/translations/index.js',
-    ],
-    '@app/translations/configuration' => [
-        'path' => './var/translations/configuration.js',
-    ],
```
Kocal, line 147: No need for these lines anymore! 🚀

src/Translator/assets/src/translator_controller.ts
```diff
@@ -7,168 +7,176 @@
...
+        return id;
     }
 
-    return message.id;
+    return {
+        setLocale,
+        getLocale,
+        setThrowWhenNotFound,
+        trans,
+    };
```
Kocal, line 181: Here I'm returning an object of methods instead of creating a `Translator` instance (implying this class exists), because:
- I want users being able to use `const { trans } = createTranslator()`, with an instance it does not work because `this` context is lost 
- It looks like a bit to the React Hooks or Vue Use systems, which has been pretty common for years in the JS ecosystem

src/Translator/doc/index.rst
```diff
@@ -27,18 +27,6 @@ Install the bundle using Composer and Symfony Flex:
...
-
-.. code-block:: terminal
-
-    $ npm install --force
-    $ npm run watch
-
-.. note::
-
-    For more complex installation scenarios, you can install the JavaScript assets through the `@symfony/ux-translator npm package`_
-
```
Kocal, line 41: I moved this part, into a dedicated section below

src/Translator/doc/index.rst
```diff
@@ -54,22 +42,52 @@ After installing the bundle, the following file should be created, thanks to the
      * If you use TypeScript, you can rename this file to "translator.ts" to take advantage of types checking.
      */
 
-    import { trans, getLocale, setLocale, setLocaleFallbacks } from '@symfony/ux-translator';
-    import { localeFallbacks } from '../var/translations/configuration';
+    import { createTranslator } from '@symfony/ux-translator';
+    import { messages, localeFallbacks } from '../var/translations/index.js';
```
Kocal, line 46: Btw, the `index.js` here is important, that's for the AssetMapper

src/Translator/doc/index.rst
```diff
@@ -103,86 +122,71 @@ Detecting missing translations
...
 
-These are then imported in your ``assets/translator.js`` file. This setup is
-very similar to working with WebpackEncore. However, the ``var/translations/index.js``
-file contains *every* translation in your app, which is not ideal for production
-and may even leak translations only meant for admin areas. Encore solves this via
-tree-shaking, but the AssetMapper component does not. There is not, yet, a way to
-solve this properly with the AssetMapper component.
+However, modern build tools, caching strategies, and compression techniques (Brotli, gzip)
+make this negligible in 2025. Additionally, a future feature will allow filtering dumped
+translations by pattern for those who need to further reduce bundle size.
```
Kocal, line 189: I thought this part was important, since one of my primary concern when creating UX Translator, was about tree-shaking etc.

src/Translator/src/TranslationsDumper.php
```diff
@@ -182,18 +185,4 @@ private function getLocaleFallbacks(MessageCatalogueInterface ...$catalogues): a
...
-        $prefix = 0;
-        do {
-            $constantName = $translationId.($prefix > 0 ? '_'.$prefix : '');
-            ++$prefix;
-        } while ($this->alreadyGeneratedConstants[$constantName] ?? false);
-
-        $this->alreadyGeneratedConstants[$constantName] = true;
-
-        return $constantName;
-    }
```
Kocal, line 198: I didn't make Blackfire profiles, but removing this code will surely help to improve performances when dealing with a ton of translations.

ux.symfony.com/config/packages/asset_mapper.yaml
```diff
@@ -16,6 +16,12 @@ framework:
             - '*/kits/**/*.md'             # UX Toolkit kit documentation
             - '*/kits/**/manifest.json'    # UX Toolkit kit manifest files
         importmap_polyfill: false
+        missing_import_mode: strict
+
+when@prod:
+    framework:
+        asset_mapper:
+            missing_import_mode: warn
```
Kocal, line 24: Applied https://github.com/symfony/recipes/pull/1347

src/Translator/tests/bootstrap.php
Kocal, line 1: Backported changes from 3.x: https://github.com/symfony/ux/commit/c96987f8e70f8763e6cef8bd42e1ceb3968b7386

Configuring the Application to not auto-exit anymore fixed an issue where PHPUnit exited. Thus, no tests were run.

smnandre:
I’m speechless. Thank you 🙇

PR #3211, opened 2025-12-04
[TwigComponent] Fix InteractsWithTwigComponents not being able to render blocks when variable scope is limited

### LINKED ISSUE #3210: [TwigComponent] Test helper can't render blocks when limiting scope using `{% with … only%}`

```twig

{% endwith %}
```

```twig
<twig:My:Component>
    {{ variableIDefinedMyself }}
</twig:My:Component>
```
this will also be blocked.

```twig
<twig:My:Component>
</twig:My:Component>
```

```twig
```

### LINKED ISSUE #3216: [TwigComponent] Blocks set with TestHelper don't get evaluated
When playing with the reproducer I created for https://github.com/symfony/ux/issues/3210, I realised `InteractsWithTwigComponents::renderTwigComponent` did not at all behave the way I expected:

Twig code inside blocks set using the `$blocks` parameter won't be evaluated. Variables won't be resolved, and Twig functions won't be executed. Instead, the resulting HTML will contain placeholders like `{{ message }}`.

This means that at the current state of the helper, basic interactions of components with their blocks can't be tested: You can't test if variables are being passed correctly or if their content is right.

This is a feature request to allow some form of testing the variables Twig components pass to their blocks.

https://github.com/symfony/ux/pull/3211 would be a way to implement this, as it would enable evaluation of the blocks.

janopae:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3210, Fix https://github.com/symfony/ux/issues/3216
| License        | MIT

When using Blocks, the helper for rendering Twig Components in tests (`InteractsWithTwigComponents::renderTwigComponent`) would create a template like this:

```twig
```

Passing the variables `data` and `blocks` from outside, where `blocks` contains the block content indexed by block name.

If the block however was declared inside a `{% with … only %}` scope, this would not work, because inside the block, you wouldn't have access to `blocks`.

This PR fixes the issue by inlining the block content when creating the template.

This probably means that the template cache creates new entries whenever the block content changes. But I think this is tolerable.

### DIFF
--- a/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
+++ b/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
@@ -62,7 +62,6 @@ trait InteractsWithTwigComponents
             ->createTemplate($template)
             ->render([
                 'data' => $data,
-                'blocks' => $blocks,
             ])
         );
     }
--- a/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
+++ b/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
@@ -51,8 +51,8 @@ trait InteractsWithTwigComponents
 
         $template = \sprintf('{%% component "%s" with data %%}', addslashes($name));
 
-        foreach (array_keys($blocks) as $blockName) {
-            $template .= \sprintf('{%% block %1$s %%}{{ blocks.%1$s|raw }}{%% endblock %%}', $blockName);
+        foreach ($blocks as $blockName => $blockContent) {
+            $template .= \sprintf('{%% block %1$s %%}%2$s{%% endblock %%}', $blockName, $blockContent);
         }
 
         $template .= '{% endcomponent %}';
--- a/src/TwigComponent/tests/Fixtures/Component/WithSlotsAndLimitedScope.php
+++ b/src/TwigComponent/tests/Fixtures/Component/WithSlotsAndLimitedScope.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ */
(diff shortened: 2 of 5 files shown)


janopae:
This probably means that the template cache creates new entries whenever the block content changes. I assume that in most cases, the block content will only change when the Test code changes. Is this tolerable?

Alternatively, I could imagine fixing this by calling `outerScope.blocks` instead of `blocks` in the generated template – this would still require Twig Component authors to allow the `outerScope` variable in blocks they want to test, but at least, `outerScope` is not a name internal to the test helper.

smnandre:
This is problematic to me, as using outerScope really does change the runtime code and makes it in my eyes too different of what would happen in production. Also at best it would not bring added value for most users.. at worst it could be causing problems with outerScope or any change we do around its implementation.

I'm pretty sure we can find a way to help you in that very specific need without impacting every other tests out there .. 

--

Update: comment originally written for the PR with your other implementation idea, but still valid (not for the outerscope runtime, but the cache one indeed is a big change, has even the lines in the template would change, its hash, etc..)

src/TwigComponent/src/Test/InteractsWithTwigComponents.php
```diff
@@ -51,8 +51,8 @@ protected function renderTwigComponent(string $name, array $data = [], ?string $
 
         $template = \sprintf('{%% component "%s" with data %%}', addslashes($name));
 
-        foreach (array_keys($blocks) as $blockName) {
-            $template .= \sprintf('{%% block %1$s %%}{{ blocks.%1$s|raw }}{%% endblock %%}', $blockName);
+        foreach ($blocks as $blockName => $blockContent) {
+            $template .= \sprintf('{%% block %1$s %%}%2$s{%% endblock %%}', $blockName, $blockContent);
```
smnandre, line 55: I don't think this is a good idea to change implementation here.

This will create a new template everytime, risking that two successive calls for the same component end up with different cache files. (what would be the case elsewhen, and something TwigComponent or LiveComponent sometimes uses).

But also because it does change the way it words. Current is runtime block-rendering, when your suggestion is changing that into a pre-compiled static template.

I can understand your desire to make things work the way you'd like, but for the vast majority of users out there, this change will not bring new or value, but would change the way their tests are done...

What about adding new methods in the trait.. could be specialized in testing/working with Components?
What do you feel about looking at what can be done here and how?

smnandre: So i'm 👎  this one

Kocal:

Thanks!

Status: Needs Work

PR #3212, opened 2025-12-04
 Fix InteractsWithTwigComponents not being able to render blocks when variable scope is limited (cache friendlier solution with limitations)

### LINKED ISSUE #3210: [TwigComponent] Test helper can't render blocks when limiting scope using `{% with … only%}`

janopae:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3210
| License        | MIT

Alternative to https://github.com/symfony/ux/pull/3211 which should be more cache-friendly. Has the limitation that it requires components to allow the `outerScope` variable in blocks they define.

### DIFF
--- a/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
+++ b/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
@@ -52,7 +52,7 @@ trait InteractsWithTwigComponents
         $template = \sprintf('{%% component "%s" with data %%}', addslashes($name));
 
         foreach (array_keys($blocks) as $blockName) {
-            $template .= \sprintf('{%% block %1$s %%}{{ blocks.%1$s|raw }}{%% endblock %%}', $blockName);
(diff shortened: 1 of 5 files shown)


smnandre:
This is problematic to me, as using outerScope really does change the runtime code and makes it in my eyes too different of what would happen in production. Also at best it would not bring added value for most users.. at worst it could be causing problems with outerScope or any change we do around its implementation.

I'm pretty sure we can find a way to help you in that very specific need without impacting every other tests out there ..

Kocal:

Thanks!

Status: Needs Work

PR #3218, opened 2025-12-05
[Translator] Add option `ux_translator.dump_typescript` to enable/disable TypeScript types generation

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #... 
| License        | MIT

Recipe PR: https://github.com/symfony/recipes/pull/1499

--- 

When using the UX Translator with the AssetMapper, there is no point to dump TypeScript types in production, since these files will not be used. The deployment will be faster.

The new option `dump_typescript` allows to disable this behavior, default to `true`.

On an app with ~3320 translation messages, a cache clear goes from ~1min to ~10s, see https://blackfire.io/profiles/compare/4d9553ca-7b96-415d-aa0f-2991f721b609/graph:
image

~50 sec for extracting types, especially when using ICU translations, is too much, even for ~3k keys. I think we can have a quick-win by early exiting when no `{` is found in the translation message.


src/Translator/src/DependencyInjection/Configuration.php
```diff
@@ -28,7 +28,14 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode = $treeBuilder->getRootNode();
         $rootNode
             ->children()
-                ->scalarNode('dump_directory')->defaultValue('%kernel.project_dir%/var/translations')->end()
+                ->scalarNode('dump_directory')
+                    ->info('The directory where translations and TypeScript types are dumped.')
+                    ->defaultValue('%kernel.project_dir%/var/translations')
+                ->end()
+                ->booleanNode('dump_typescript')
+                    ->info('Control if TypeScript types should be dumped alongside translations. Can be useful to disable when not using TypeScript (e.g. AssetMapper in production).')
```
smnandre, line 36:
```suggestion
                    ->info('Control whether TypeScript types are dumped alongside translations. Disable this if you do not use TypeScript (e.g. in production when using AssetMapper).')
```

src/Translator/src/TranslationsDumper.php
```diff
@@ -35,6 +35,7 @@ class TranslationsDumper
 
     public function __construct(
         private string $dumpDir,
+        private bool $dumpTypeScript,
         private MessageParametersExtractor $messageParametersExtractor,
```
smnandre, line 39: I now this class is marked @experimental but the component is out there for some times now ...  :/

src/Translator/src/DependencyInjection/Configuration.php
Kocal: See https://github.com/symfony/ux/pull/3221

PR #3244, opened 2025-12-19
[Translator] Refactor `TranslationsDumper` options from `__constructor` and setters, to `dump` method

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

This PR simplifies how `TranslationsDumper` works by moving all configuration options from the constructor/setters into the `dump()` method itself.

Even if using the `TranslationsDumper` directly is not a common use case (most users will the default behavior with the cache warmer), this change improves the overall design of the class:
- no stateful instance
- only one place to configure settings, instead of __construct() and setters
- adding new options will be easier in the future

Promise, this is the last BC for Translator 🙏🏻

### DIFF
--- a/src/Translator/src/CacheWarmer/TranslationsCacheWarmer.php
+++ b/src/Translator/src/CacheWarmer/TranslationsCacheWarmer.php
@@ -18,13 +18,23 @@ use Symfony\UX\Translator\TranslationsDumper;
 /**
  * @author Hugo Alliaume <hugo@alliau.me>
  *
+ * @internal
+ *
  * @experimental
  */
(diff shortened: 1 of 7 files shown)


src/Translator/src/CacheWarmer/TranslationsCacheWarmer.php
```diff
@@ -18,13 +18,19 @@
 /**
  * @author Hugo Alliaume <hugo@alliau.me>
  *
+ * @internal
+ *
```
Kocal, line 22: This was initially missing

PR #3248, opened 2025-12-23
[Translator] Add `keys_patterns` configuration option to filter dumped translations by key patterns

### LINKED ISSUE #3241: [Translator] Add configuration to filter dumped keys by pattern
We already have the option to choose the domain from which we retrieve keys, but adding a filter on key names (via pattern) would allow for more refined filtering.
Since #3208, there is no longer any tree shaking. Adding this feature could help reduce the size of the generated .js files.

```yaml
ux_translator:
   keys_patterns:
      - 'app.**'
      - 'foo.*.bar'
      - 'foo.**.bar' 
      - # ...
```

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #3241 
| License        | MIT

This PR adds a `keys_patterns` configuration option to filter dumped translations by key patterns (e.g., `app.*`, `!*.internal`).
This allows users to reduce the size of the dumped translations by including only the necessary keys.

### DIFF
--- a/src/Translator/src/DependencyInjection/Configuration.php
+++ b/src/Translator/src/DependencyInjection/Configuration.php
@@ -89,6 +89,14 @@ class Configuration implements ConfigurationInterface
                         })
                     ->end()
                 ->end()
+                ->arrayNode('keys_patterns')
+                    ->info('List of translation key patterns to include/exclude from the generated translations. Prefix with a `!` to exclude a pattern. Supports wildcards (e.g., `app.*`, `*.label`).')
+                    ->scalarPrototype()->end()
+                    ->beforeNormalization()
+                        ->ifString()
+                        ->then(fn ($v) => [$v])
+                    ->end()
+                ->end()
             ->end()
         ;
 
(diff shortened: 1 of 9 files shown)


src/Translator/src/DependencyInjection/Configuration.php
```diff
@@ -89,6 +89,14 @@ public function getConfigTreeBuilder(): TreeBuilder
                         })
                     ->end()
                 ->end()
+                ->arrayNode('keys_patterns')
```
94noni, line 92: Why double « s » in config ?

Kocal: My reasoning here was to configure pattern**s** to filter (or not) key**s**, but I may be wrong. Do you have any suggestions in mind?

94noni: Its just the double plural that felt weird
Nothing more in mind ^^

PR #3263, opened 2026-01-03
 PHP CS Fixer: update usage

keradus:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix CS 
| License        | MIT

best to review commit by commit


src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
```diff
@@ -2100,3 +2101,5 @@ class DummyObjectWithObjects
     #[LiveProp]
     public array $products = [];
 }
+
+// @php-cs-fixer-ignore fully_qualified_strict_types Due to https://github.com/symfony/ux/pull/2768#issuecomment-2905091761
```
Kocal: Can we move this line at the very top?

Kocal (approved):
Thanks you for working on this!

Kocal:
Can we also remove the PHP-CS-Fixer configuration file from ux.symfony.com?

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
keradus: i recommend it on bottom, to avoid any placement conflict with license header, declare strict etc - but you can put it anywhere in the file

keradus:
removed as requested

src/LiveComponent/tests/Integration/LiveComponentHydratorTest.php
Kocal: I understand, but I think it would be clearer for a human to have this comment at the beginning of the file, especially since it is quite long here.

keradus: i moved it. if you prefer another place, please locate it at place of your preference

Kocal:
Thanks @keradus, that's super appreciated!

PR #3265, opened 2026-01-03
 Improve tailwind dark mode support for toolkit

sblondeau:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #... 
| License        | MIT

Improve tailwind dark mode support for toolkit. The `prefers-color-scheme` is used by default, unless you specified the .dark / .light classes

ex : if system theme is dark => use dark theme (and can use tailwind `dark:` in classes
BUT if system theme is dark and you define a `class="light"` theme remains light
AND if system them is light and you define a `class="dark"` theme will be dark and `dark:` syntax available

### DIFF
--- a/src/Toolkit/kits/shadcn/INSTALL.md
+++ b/src/Toolkit/kits/shadcn/INSTALL.md
@@ -17,7 +17,16 @@ In your `assets/styles/app.css`, after the TailwindCSS imports, add the followin
 ```css
 @import "tailwindcss";
 
-@custom-variant dark (&:is(.dark *));
+@custom-variant dark {
+  &:where(.dark, .dark *) {
+    @slot;
+  }
+  @media (prefers-color-scheme: dark) {
+    &:where(:not(.light *)) {
+      @slot;
+    }
+  }
+}
 
 :root {
     --background: oklch(1 0 0);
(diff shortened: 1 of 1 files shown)


Kocal (commented):
Thanks for your contribution!

However, these CSS styles are copy-paster from the official documentation https://ui.shadcn.com/docs/installation/manual#configure-styles.
I would like to stay the more sync as possible with Shadcn official doc. 

Do you think you can open such a PR on Shadcn repository? Thanks!

sblondeau:
Hi, 
Thanks for your answer, I understand, however it seems that dark-theme is only managed with JS in Shadcn ? https://ui.shadcn.com/docs/dark-mode
This addition of CSS prefer-color-scheme is not incompatible with shadcn because you can still use the .dark class, but it is true that is make the initial CSS config more complicated.
By the way, to improve DX, I don't know if it is possible, but it could be great to have a way to automatically import a "shadcn.css" and add an @import in app.css (maybe when adding the first shadcn component ?) and/or maybe a cool way to easily use other shadcn [themes](https://ui.shadcn.com/docs/theming) ?

Kocal:
Hi!

I'm sorry but it complexify the CSS too much (the one from `INSTALL.md` and the one generated). Users are free to modify their CSS to supports `prefers-color-scheme` if they want, but as long as it's not something officially "supported" by Shadcn, it's a no for me.  

> By the way, to improve DX, I don't know if it is possible, but it could be great to have a way to automatically import a "shadcn.css" and add an https://github.com/import in app.css (maybe when adding the first shadcn component ?) and/or maybe a cool way to easily use other shadcn [themes](https://ui.shadcn.com/docs/theming) ?

I think yes, still need to investigate what would be the best to do so

Issue #3246, opened 2025-12-21
[Toolkit] Add blocks


tibobaldwin:
Currently, Toolkit manages the components, and it's great!
I saw on Shadcn that they have [blocks](https://ui.shadcn.com/docs/blocks), and I find that interesting.

Here's Shadcn's definition:

> A block can be a single component (e.g., a variation of a UI component) or a complex component (e.g., a dashboard) with multiple components, hooks, and utilities.

Blocks can be downloaded just like components using the command line. For example, on this page (https://ui.shadcn.com/blocks/login) listing the login form blocks, you can run the command `npx shadcn@latest add login-01`, which adds the components included in that block and downloads the block itself.

Kocal:
Hi!

Technically blocks are already supported, Recipes (the ones under `kits/*`) can be components or anything else as long as it follow the conventional structure.

The "recipe type" (e.g.: `component`) is mainly used for documentation purpose

excitedbox:
I think the problem on the one hand is "technically supported" and on the other it is standardization and ease of use. It is usually not good to use non-standard terminology like recipe instead of block because a recipe can be lots of things, while a block has a pretty much accepted definition in the design context.

The other problem is that Twig has kinda become a second class citizen in a world with multiple options used in the rest of the industry. There is very little in the way of reusable building blocks for Twig or Symphony in general, because everything requires a lot more work to get going, which leads to people choosing different options and therefore Symphony also seeing much less integration into projects that could build up an ecosystem.

JS and Wordpress lead not because they are the best in their space but because of the ecosystem around them. Symphony lacks the ecosystem support and lacks any kind of system for that ecosystem to develop.

UX could have become that ecosystem, but it has kind of languished for several years now as the previous ecosystem recipes/bundles age, still barely browsable in some git repository.

smnandre:
Let's keep this **constructive** please.

Also, you cannot even imagine how misplaced "languished for several years now" is here.

excitedbox:
This was being constructive. These issues are cited repeatedly for people choosing other frameworks for projects. Symphony had a bundle directory, but that was discontinued without replacement and little new development has happened since.

It isn't constructive silencing issues or like I was told back when Symphony 5 was released, "no, no, no, Symphony will NEVER, get front-end support."  When the lead developer shows such an adversarial stance against offering a full stack solution, I made the choice to move away from Symphony and mainly keep up because, as part of my work I maintain some projects built on PARTS of Symphony.

PR #3282, opened 2026-01-09
[Toolkit] Prepare base for recipe type `Block`

jeremyFreeAgent:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | #3246
| License        | MIT

This introduce a new`Block` type for the Toolkit recipes as explained in the #3246 issue.

### DIFF
--- a/src/Toolkit/schema-kit-recipe-v1.json
+++ b/src/Toolkit/schema-kit-recipe-v1.json
@@ -5,6 +5,11 @@
     "type": "object",
     "required": ["name", "description"],
     "properties": {
+        "type": {
+            "type": "string",
+            "description": "The type of the Recipe",
+            "enum": ["component", "block"]
+        },
         "name": {
             "type": "string",
             "description": "The name of the Recipe"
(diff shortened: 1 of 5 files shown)


jeremyFreeAgent:
I see that we have several blocks available for the same same, (`login-01`, `login-02`, `login-03`, etc.). Don't know how you want that to be used.

Maybe when we run
```
symfony console ux:install login --kit shadcn
```
it should install all the login blocks?

Kocal:
Hi, and thanks for the contribution :D 

I'm sorry, but I had to undo the changes in `ux.symfony.com`. It didn't match what I had in mind, and reviewing something I'm not 100% sure about myself, would be a pain for both of us... 😅 

I kept only what was necessary and made a few changes (JSON schema and an improved error message).

> I see that we have several blocks available for the same same, (`login-01`, `login-02`, `login-03`, etc.). Don't know how you want that to be used.
> 
> Maybe when we run
> 
> ```
> symfony console ux:install login --kit shadcn
> ```
> 
> it should install all the login blocks?

The command `ux:install login --kit shadcn` should fail because no recipe will be explicitly named `login`. 
If we allowed multiple `login-*` blocks to be installed, we would have conflicts with the files that need to be copied into the Symfony app.

Recipes block like `login-01`, `login-02`, ... must be installed by running `ux:install login-01 --kit shadcn`, etc.

I will open soon a PR with complete `Block` recipe type support and `login-*` blocks.

Kocal:
Thanks @jeremyFreeAgent.

PR #3169, opened 2025-11-14
[TwigComponent] Allow using directory name as component name for anonymous components

andyexeter:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #1998  
| License        | MIT

This pull request introduces support for using `index.html.twig` as a component template when the component is placed in a directory named after itself. This helps avoid repetition and improves organization for nested components. The documentation is updated to explain this feature, and corresponding logic and tests are added to ensure correct template resolution.

PR #3295, opened 2026-01-14
[TwigComponent] Add support for anonymous component name resolution in other bundle

### LINKED ISSUE #1998: [Feature Request]  Anonymous Name Resolution Fallback
It would be cool, if the Anonymous component with the same name as its Directory would be used as a fallback if one with an exact match isn't found

Currently, you would do:
`/components/Menu/Menu.html.twig`
`<twig:Menu:Menu/>`

It would be cool to be able to write:
`<twig:Menu/>`
And have it fallback to
`/components/Menu/Menu.html.twig`
in case that:
`/components/Menu.html.twig`
is not found.

**Reasoning**
| Pros ✅                                                                        | Cons  ❌                                                                     |
|---------------------------------------------------------------------|---------------------------------------------------------------------|
| If Components get more complex, you tend to split them up, and organize them in subdirectories. Having the "Index" component outside its subdir only makes it more cluttered and harder to find stuff.  This would fix that. | This can only be a fallback, to ensure backwards compatibility and because of Name conflicts |
|You Don't Get Duplication in Names anymore, which make it easier to read, and more fun to write templates. |

Phil23:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #1998
| License        | MIT

In a recent PR (#3169), @andyexeter added support for using `index.html.twig` as a component template when the component is placed in a directory named after itself. However, it seems to miss the implementation for external bundles. This PR extends the change so that anonymous component templates in other bundles can also use this naming pattern.

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.33
+
+- Extended support for the `index.html.twig` template fallback when resolving namespaced anonymous components.
+
 ## 2.32
 
 - Add option `profiler.collect_components` to control component data collection
--- a/src/TwigComponent/src/ComponentTemplateFinder.php
+++ b/src/TwigComponent/src/ComponentTemplateFinder.php
@@ -81,6 +81,11 @@ final class ComponentTemplateFinder implements ComponentTemplateFinderInterface
             return $template;
         }
 
+        $template = '@'.$parts[0].'/components/'.$parts[1].'/index.html.twig';
+        if ($loader->exists($template)) {
+            return $template;
+        }
+
         return null;
     }
 }
--- a/src/TwigComponent/tests/Integration/ComponentFactoryTest.php
+++ b/src/TwigComponent/tests/Integration/ComponentFactoryTest.php
@@ -178,6 +178,33 @@ final class ComponentFactoryTest extends KernelTestCase
         $this->assertNull($metadata->get('class'));
     }
 
+    public function testLoadingAnonymousComponentFromBundleWithFallback()
+    {
+        // Component from external bundle with index.html.twig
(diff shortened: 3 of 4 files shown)


smnandre:
Hi @Phil23 

Could you add a few tests for the edge cases... that will mainly be used to decide once and keep the same behaviour afterward. 

```twig
<twig:Acme:Foo:Index /> 

<twig:Acme:Foo:index />

<twig:Acme:Index:index />

```

( depending on the various possibilities (there is or not a template, caps / lower) )

Also let's add a line in the CHANGELOG (the one in the src/TwigComponent directory)


Minor/personal note here: this will work (and thank you for this PR by the way 🤝 ).. 

... I would still not recommand "officially" this trick to bundle maintainers.. or I would recommand them to not abuse it, and keep it "inside" their component structure instead of using this for the "main ones" -- button for instance.. 
... as the differences with how the Twig and Live component are handled could lead to frustrations I guess.

Phil23:
@smnandre Thank you for the feedback.

I added several edge cases to the unit test.

I also added the changelog entry. I wasn't sure about the version number, though.

src/TwigComponent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.33
+
+-  Extended the support for using the directory name as a component name for anonymous components from other bundles.
```
```suggestion
-  Extended support for the `index.html.twig` template fallback when resolving namespaced anonymous components.
```

I am not sure this adds _much_ precision, but it does better reflect the current implementation.

-> Call "Foo:Bar" and we will find the matching "Foo/Bar/index.html.twig" template

smnandre:
Thank you for this nice first PR  @Phil23  🎉

smnandre (approved):

Kocal:
Thank you @Phil23.

PR #3299, opened 2026-01-16
[Toolkit] Rework recipes docs/examples

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #... 
| License        | MIT

> [!NOTE]
> The PR must not be squashed! Some recipes Twig templates were re-adapted from Shadcn 

This PR moved the recipes `EXAMPLES.md` files to the external website repository, but examples are still present in the `examples/` folder of each recipe.

When initially working on Toolkit, I wasn't sure if these files should be readable on GitHub/text editor, or only on the website, and if it was only on the website, how to best maintain them, etc...

Better than nothing, I created these `EXAMPLES.md` files in each recipe to have a place to write down examples in markdown format, which could then be used as a source of truth for the website (and snapshot tests). But it was not convenient because the format was not flexible, and examples were extracted with some weird regex parsing.

Now, examples are split into multiple files in the `examples/` folder of each recipe, they only contain Twig and can correctly be highlighted by IDEs and formatted by Twig-CS-Fixer.

Documentation files are now in the external website repository, see https://github.com/symfony/ux.symfony.com/pull/3

**Before:**
Capture d’écran 2026-01-16 à 08 15 27

**After:**
image


src/Toolkit/kits/shadcn/alert-dialog/templates/components/AlertDialog/Content.html.twig
```diff
@@ -1,7 +1,7 @@
 {# @block content The default block #}
 <dialog
     id="{{ _alert_dialog_id }}"
-    class="{{ 'bg-background fixed top-[50%] left-[50%] z-50 max-w-[calc(100%-2rem)] translate-x-[-50%] translate-y-[-50%] scale-95 gap-4 rounded-lg border p-6 opacity-0 shadow-lg transition-all transition-discrete duration-200 backdrop:transition-discrete backdrop:duration-150 open:grid open:scale-100 open:opacity-100 open:backdrop:bg-black/50 sm:max-w-lg starting:open:scale-95 starting:open:opacity-0 ' ~ attributes.render('class')|tailwind_merge }}"
+    class="{{ 'text-foreground bg-background fixed top-[50%] left-[50%] z-50 max-w-[calc(100%-2rem)] translate-x-[-50%] translate-y-[-50%] scale-95 gap-4 rounded-lg border p-6 opacity-0 shadow-lg transition-all transition-discrete duration-200 backdrop:transition-discrete backdrop:duration-150 open:grid open:scale-100 open:opacity-100 open:backdrop:bg-black/50 sm:max-w-lg starting:open:scale-95 starting:open:opacity-0 ' ~ attributes.render('class')|tailwind_merge }}"
```
Kocal, line 4: This fix the following issue:

Image

We used native `dialog` element in `AlertDialog`, but it defines some color/background styles that can conflict with dark mode. 

This same fix should be added on `Dialog` component.

I will cherry pick this fix.

Kocal: -> https://github.com/symfony/ux/pull/3300

PR #3302, opened 2026-01-16
[LiveComponent] Support DataModel bindings with LiveComponent modifiers

### LINKED ISSUE #555: [TwigComponent] data-model
Hello,

We figured out that there is a [data-model subscriber](https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/EventListener/DataModelPropsSubscriber.php) in twig components which allows you to:

```twig
<div>
    {{ component('textarea_component', {
        'data-model': 'value:someParentComponentProp'
    }) }}
</div>
```

```twig
<textarea>{{ value }}</textarea>
```

I was wondering:  Why don't you just want to pass `{value: someParentComponentProp}` as the textarea_component attributes here?

The problem with the data-model here is that it interferes with the live components syntax of: `norender|registration_form[password][first]` or `debounce(100)|someField`.

This results in a

>  An exception has been thrown during the rendering of a template ("Can't get a way to read the property "norender|registration_form" in class "App\Twig\Component\SomeComponent".").

I currently don't see the value of the data-model subscriber and would suggest to remove it. But I'm sure you have a good counter argument for having it included.

### LINKED ISSUE #3152: [LiveComponent] Debounce not working when data-model is linked to parent data-model
Debouncing does not work when it is used on a child component that is linked to the parent through data-model.

Let's say we have a child component called searchField with an input.
```html
<input type="search" data-model="value" />
```

and a parent that uses this component linking data-model to a prop named query
```html
<twig:searchField data-model="query:value" />
```

If one wishes to delay the re-render by debouncing the input field the only place where this would be accepted is the child component
```html
<input type="search" data-model="debounce(500)|value" />
```

But this has no effect. The parent data-model gets updated immediately triggering the re-render. 

I have notice that many other modifiers work as expected. For instance, `min_length` would prevent  to update the parent if the length requirement is not met. 

Trying to do the same thing on the the parent 
```html
<twig:searchField data-model="debounce(500)|query:value" />
```

triggers an error `Can't get a way to read the property "debounce(500)|query`.

Wertisdk:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3152, Fix #555
| License        | MIT

[LiveComponent] 
### Support DataModel bindings with LiveComponent modifiers

The data-model attribute is used by both the Data Binding (for parent-child prop binding) and LiveComponent (for form input binding with modifiers). This was causing LiveComponent modifiers like 'norender|field' or 'debounce(100)|field' to fail because the property was not extracted from the LiveComponent data-model attribute, but was used directly as a property

This fix adds modifier detection to DataModelPropsSubscriber that checks if data-model contains LiveComponent modifier syntax (pipe '|') and extracts the value from the modifier syntax

Co-developed with @mrbase

### DIFF
--- a/src/LiveComponent/tests/Fixtures/Component/ParentComponentDataModelWithModifiers2.php
+++ b/src/LiveComponent/tests/Fixtures/Component/ParentComponentDataModelWithModifiers2.php
@@ -20,5 +20,6 @@ final class ParentComponentDataModelWithModifiers2
 {
     use DefaultActionTrait;
 
-    #[LiveProp(writable: true)] public string $content;
+    #[LiveProp(writable: true)]
+    public string $content;
 }
--- a/src/LiveComponent/tests/Fixtures/Component/ParentComponentDataModelWithModifiers2.php
+++ b/src/LiveComponent/tests/Fixtures/Component/ParentComponentDataModelWithModifiers2.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent('parent_component_data_model_with_modifiers_2')]
+final class ParentComponentDataModelWithModifiers2
+{
+    use DefaultActionTrait;
+
+    #[LiveProp(writable: true)] public string $content;
+}
--- a/src/LiveComponent/tests/Integration/EventListener/DataModelPropsSubscriberTest.php
+++ b/src/LiveComponent/tests/Integration/EventListener/DataModelPropsSubscriberTest.php
@@ -54,4 +54,39 @@ final class DataModelPropsSubscriberTest extends KernelTestCase
         $this->assertStringContainsString('<textarea data-model="content">default content on mount</textarea>', $html);
         $this->assertStringContainsString('<input data-model="content" value="default content on mount" />', $html);
     }
+
+    public function testDataModelPropsWithModifiersAreSharedToChild()
+    {
+        /** @var ComponentRenderer $renderer */
+        $renderer = self::getContainer()->get('ux.twig_component.component_renderer');
+
+        $html = $renderer->createAndRender('parent_form_component_with_modifiers', [
+            'content' => 'Hello data-model!',
+            'content2' => 'Value for second child',
+            'attributes' => ['id' => 'dummy-live-id'],
+        ]);
+
+        // Verify that the data-model attributes include the "norender" modifier and that values are passed correctly
+        $this->assertStringContainsString('<textarea data-model="norender|content:value">Hello data-model!</textarea>', $html);
+        $this->assertStringContainsString('<textarea data-model="norender|content2:value">Value for second child</textarea>', $html);
+    }
+
+    public function testDataModelPropsWithModifiersAreAvailableInEmbeddedComponents()
+    {
+        $templateName = 'components/parent_component_data_model_with_modifiers.html.twig';
+        $obscuredName = '684c45bf85d3461dbe587407892e59d9';
+        $this->addTemplateMap($obscuredName, $templateName);
+
+        /** @var ComponentRenderer $renderer */
+        $renderer = self::getContainer()->get('ux.twig_component.component_renderer');
+
+        $html = $renderer->createAndRender('parent_component_data_model_with_modifiers', [
+            'attributes' => ['id' => 'dummy-live-id'],
+        ]);
+
+        // Verify that the data-model attributes include the "norender" modifier and that values are passed correctly
+        $this->assertStringContainsString('<textarea data-model="norender|content">default content on mount</textarea>', $html);
+        $this->assertStringContainsString('<input data-model="norender|content" value="default content on mount" />', $html);
+    }
+
 }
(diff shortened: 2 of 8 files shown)


Wertisdk:
I am unsure what the failed test signifies and what could be the cause: 
[Unit Tests / php (8.1, lowest) (pull_request)](https://github.com/symfony/ux/actions/runs/21065838561/job/60582862853?pr=3302)

src/LiveComponent/tests/Fixtures/Component/ParentComponentDataModelWithModifiers2.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent('parent_component_data_model_with_modifiers_2')]
+final class ParentComponentDataModelWithModifiers2
+{
+    use DefaultActionTrait;
+
+    #[LiveProp(writable: true)] public string $content;
```
smnandre:
```suggestion
    #[LiveProp(writable: true)]
    public string $content;
```

src/LiveComponent/tests/Integration/EventListener/DataModelPropsSubscriberTest.php
```diff
@@ -54,4 +54,38 @@ public function testDataModelPropsAreAvailableInEmbeddedComponents()
...
+
+        $html = $renderer->createAndRender('parent_form_component_with_modifiers', [
+            'content' => 'Hello data-model!',
+            'content2' => 'Value for second child',
+            'attributes' => ['id' => 'dummy-live-id'],
+        ]);
+
+        // Verify that the data-model attributes include the "norender" modifier and that values are passed correctly
+        $this->assertStringContainsString('<textarea data-model="norender|content:value">Hello data-model!</textarea>', $html);
+        $this->assertStringContainsString('<textarea data-model="norender|content2:value">Value for second child</textarea>', $html);
```
smnandre, line 71: Hmm.. was this mix of the two syntaxes ever supported by LiveComponent ? I mean, documented etc ?

If not (what i'd say but with not certainety) this is not a bug fix but a feature :) 

And one i fear I would be very critical of... as it creates (another) layer of confusion for people.

Do you think there any way we can solve your original core need, that does not require to add another degree of complexity here ?

mykiwi:
@smnandre @Kocal Any idea when this could be released?

Wertisdk:
@smnandre please let me know if there's anything I can do to address the failed unit test or any other feedback for this to be accepted 😃

smnandre:
Well if I don't send my review..... 🤦‍♂️   sorry for that

image

src/LiveComponent/tests/Integration/EventListener/DataModelPropsSubscriberTest.php
Wertisdk: @smnandre Do you mean 
1) specifically the `norender|content:value` syntax where both '|' and ':' is used together or 
2) the concept in general of supporting modifiers in the data-model attribute when added as an attribute in the component method?

I could fix 1. by checking that ':' is not present before trying to to extract the property name and amending the test to expect an exception thrown because the property is not present (because it is being returned literally and did not match a property).

Wertisdk:
@smnandre 
I see now, that @carsonbot went a bit overboard with labels, because I initially had a formating error in the table for the PR... 
Is there something I can do to make @carsonbot reevaluate the labels, or do someone need to fix them for me? :)

src/LiveComponent/tests/Integration/EventListener/DataModelPropsSubscriberTest.php
smnandre: More "2" here : I had the impression data-model was either used for child component using the xx:yy syntax, or into single components with modifiers, but not both.

Wertisdk: Thanks a lot for taking the time to look at this!

I completely understand the concern about adding complexity, and I also agree that if mixing the two syntaxes was never intended, then treating this as a feature rather than a bugfix is reasonable.

The main reason we experienced this as a bug is not so much the specific behavior, but the inconsistency that emerges from reusing the same data-model attribute in both contexts.

If I write
```twig
<select data-model="norender|myProperty">...</select>
```

this works exactly as documented. But if I wrap that same `<select>` as the root element of a Twig component and use:

```twig
<twig:MySelectComponent data-model="norender|myProperty" />

... or ... 

{{ component('MySelectComponent', {
    dataModel: 'content',
}) }}
```

the exact same syntax suddenly fails, even though from my perspective it feels like the same operation, just composed.

This is where my surprise comes from: not that modifiers exist, but that the same attribute value behaves differently depending on whether it’s applied directly to a DOM element or passed through a Twig component. If these two concepts were never meant to interact, then reusing data-model for both makes that distinction very hard to discover or reason about as a developer.

Since the different behaviour is not explicitly mentioned in the documentation, I believed this was a bug. 

That said, I absolutely defer to the project’s direction here. If it's preferred to not support modifiers in Twig component bindings, I think documenting that limitation (or even failing fast with a clearer error) would already help a lot.
The approach in the PR felt like the least intrusive way to align the behavior with user expectations.

We can work around this issue in a couple of ways, one of them by manually adding modifiers in our Twig component like I describe in https://github.com/symfony/ux/issues/3152#issuecomment-3738647175

Happy to adjust the PR in whatever direction makes the most sense for Symfony UX or contributing to explicitly documenting the behavior, and thanks again for the guidance! :)

Wertisdk:
@smnandre Sorry to mention you, but is there anything else I should do with this pull request after you changed the labels? 
I am new to this project, so I just want to make sure I'm not missing a process I should be aware of :)

Kocal:

Thanks!

Status: Needs Work

Wertisdk:
@Kocal 
Thanks for the heads up!
I've retargeted the PR and everything looks good as far as I can tell 🎉

Issue #3261, opened 2026-01-02
[TwigComponent] Reconsider behaviour for `(bool) false` value for ARIA attributes


polarbirke:
https://github.com/symfony/ux/pull/1805 removed the ability to render ARIA attributes with a `(string) false` value by passing a `(bool) false` value to the `attributes` helper.

Example:
```
{# templates/components/Button.html.twig #}
<button {{ attributes }}>Toggle Foo</button>

{# render component #}
<twig:Button aria-expanded="{{ user.alwaysShowFoo }}" />

{# renders as: #}
user.alwaysShowFoo
? <button aria-expanded="true">Toggle Foo</button>
: <button>Toggle Foo</button>

{# desired output: #}
user.alwaysShowFoo
? <button aria-expanded="true">Toggle Foo</button>
: <button aria-expanded="false">Toggle Foo</button>
```

The reason given in https://github.com/symfony/ux/pull/1797#issuecomment-2083188899 is a prior decision to use `(bool) false` to signal that an attribute should be omitted.

This decision is **generally fine** for HTML attributes. 

I would have preferred to finetune the behaviour **for ARIA attributes** in a slightly more nuanced way than is already the case. As was suggested in https://github.com/symfony/ux/pull/1797#issuecomment-2082614096, there would be a viable alternative to use `null` to not render an ARIA attribute and have `(bool) false` cast to a string value of `"false"`. It feels off to have special behaviour for ARIA attributes for `(bool) true` only.

I would like to submit that the discussion is reopened.

smnandre:
Hi @polarbirke! 

This is a very interesting topic

> https://github.com/symfony/ux/pull/1805 removed the ability to [...]

Well.. to be fair, #1805 fixed a major BC break introduced (by me) two weeks earlier, when this capability was first added.

> It feels off to have special behaviour for ARIA attributes for (bool) true only.

That special case was mostly a DX shortcut to simplify usage, in response to user requests. But I agree it is not perfect right now, and this kind of 'special behaviour' was originally something I wanted to avoid (see my [first reaction](https://github.com/symfony/ux/issues/1701#issuecomment-2043808172)).

## A complex situation

The reality is that we probably will not find a perfect solution here, because HTML is complex:
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-att

[...]

PR #3307, opened 2026-01-19
 Change the deprecation message for `null` attribute values

mpdude:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | First step towards #3261, final fix can only happen in 3.x
| License        | MIT

This PR changes the announcement of what `null` attribute values will mean in 3.x.

The plan towards fixing #3261 is as follows:

In 2.x:
* `null` values are treated as `true`, giving the deprecation notice
* Users need to pass `true` to keep the current behavior and make the notice go away

In 3.x:
* Make a `null` value (as well as `false`) omit the attribute
* `aria-*` attributes with a `false` value will trigger a deprecation notice, saying that it will be `"false"` in 4.0. To omit the `aria-*` attribute, `null` shall be used and will make the notice go away.

In 4.x:
* For `aria-*`, convert `false` to `"false"`, to be in line with `true => "true"`, closing #3261.

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,11 +1,15 @@
 # CHANGELOG
 
+## 2.33
+
+-  Deprecate using `null` to add an attribute. Starting in `3.0`, passing `true` explicitly sets the attribute, while passing `false` or `null` omits it.
+
 ## 2.32
 
 -  Add option `profiler.collect_components` to control component data collection
    in the profiler (enabled in debug mode by default)
 -  Add support for using directory name as component name for anonymous components
--  Deprecate using `null` to add an attribute. Starting in `3.0`, passing `true` explicitly sets the attribute, while passing `false` or `null` omits it.
+
 ## 2.30
 
 -  Ensure compatibility with PHP 8.5
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -5,6 +5,9 @@
 -  Add option `profiler.collect_components` to control component data collection
    in the profiler (enabled in debug mode by default)
 -  Add support for using directory name as component name for anonymous components
+-  Change announced behavior for `null` attribute values: Instead of throwing, 3.0
+   will omit attributes with `null` values. To keep the current result, `true`
+   should be passed instead of `null`.
 
 ## 2.30
 
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -5,10 +5,7 @@
 -  Add option `profiler.collect_components` to control component data collection
    in the profiler (enabled in debug mode by default)
 -  Add support for using directory name as component name for anonymous components
--  Change announced behavior for `null` attribute values: Instead of throwing, 3.0
-   will omit attributes with `null` values. To keep the current result, `true`
-   should be passed instead of `null`.
-
+-  Deprecate using `null` to add an attribute. Starting in `3.0`, passing `true` explicitly sets the attribute, while passing `false` or `null` omits it.
 ## 2.30
 
 -  Ensure compatibility with PHP 8.5
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -61,7 +61,7 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
             }
 
             if (null === $value) {
-                trigger_deprecation('symfony/ux-twig-component', '2.8.0', 'Passing "null" as an attribute value is deprecated and will throw an exception in 3.0.');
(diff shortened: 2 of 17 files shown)


src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -61,7 +61,7 @@ public function __toString(): string
             }
 
             if (null === $value) {
-                trigger_deprecation('symfony/ux-twig-component', '2.8.0', 'Passing "null" as an attribute value is deprecated and will throw an exception in 3.0.');
+                trigger_deprecation('symfony/ux-twig-component', '2.32.0', 'Passing null as value for the %s attribute will omit the attribute in 3.0. To keep the current result, pass true instead.', $key);
                 $value = true;
```
smnandre, line 65: Not sure we should precise the key... I mean maybe we should not suggest passing null is deprecated for _that_ attribute only.

> "To keep the current result, pass true instead."

I think with the various changes we plan, we should avoid using "current" / "previous" / "next" or maintaining (or reading) the docs will become quickly really hard.

What about something like "Use true to add the attribute, null or false to remove it"

src/TwigComponent/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 -  Add option `profiler.collect_components` to control component data collection
    in the profiler (enabled in debug mode by default)
 -  Add support for using directory name as component name for anonymous components
+-  Change announced behavior for `null` attribute values: Instead of throwing, 3.0
+   will omit attributes with `null` values. To keep the current result, `true`
+   should be passed instead of `null`.
```
```suggestion
-  Deprecate using `null` to add an attribute. Starting in `3.0`, passing `true` explicitly sets the attribute, while passing `false` or `null` omits it.
```

wdyt ?

src/TwigComponent/src/ComponentAttributes.php
mpdude: I will try to reword that. 

My initial idea was that it might make it easier for users to understand which attribute to search for. But I also realized that people might be collecting and grouping deprecation messages by type, and so having a different message for each attribute might not be helpful to them either. 

Also, for example the XmlFileReader deprecated in Symfony 7.4 only says XML is deprecated but does not give the file name. So, omitting the attribute name will be in line with that.

mpdude: Now it's

> Passing null as value to set an attribute is deprecated, pass true instead. In 3.0, the null value will remove the attribute.

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -61,7 +61,7 @@ public function __toString(): string
             }
 
             if (null === $value) {
-                trigger_deprecation('symfony/ux-twig-component', '2.8.0', 'Passing "null" as an attribute value is deprecated and will throw an exception in 3.0.');
+                trigger_deprecation('symfony/ux-twig-component', '2.32.0', 'Passing null as value to set an attribute is deprecated, pass true instead. In 3.0, the null value will remove the attribute.', $key);
```
Kocal: To be removed then:
```suggestion
                trigger_deprecation('symfony/ux-twig-component', '2.32.0', 'Passing null as value to set an attribute is deprecated, pass true instead. In 3.0, the null value will remove the attribute.');
```

Kocal (commented):
Thanks you for working on this!

> `aria-*` attributes with a `false` value will trigger a deprecation notice, saying that it will be `"false"` in 4.0 (...)

Couldn't it be done in 3.x?

src/TwigComponent/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 -  Add option `profiler.collect_components` to control component data collection
    in the profiler (enabled in debug mode by default)
 -  Add support for using directory name as component name for anonymous components
-
+-  Deprecate using `null` to add an attribute. Starting in `3.0`, passing `true` explicitly sets the attribute, while passing `false` or `null` omits it.
```
Kocal: To move to a new section `2.33`

mpdude: 🤦🏻  Thanks

mpdude:
> Couldn't it be done in 3.x?

I don't think so, but I haven't tried it, only looking at the code.

In 2.x, `false` means to omit attributes, and `null` (still) is equal to `true` (and should print `"true"` for `aria-*`).

So if users would get a deprecation notice in 2.x, telling them that `false` for `aria-*` would not omit the attribute in 3.0 but print `"false"` instead – there is no other value that they could migrate to (in 2.x) to keep the effect of omitting the attribute. Remember, `null` still results in `true`.

Thus I think we need the change here to give ourselves a little more legroom in 3.0 and something users could switch their `aria-*: false` values to.

Am I missing something?

mpdude:
Not sure about test failures, seem unrelated

Kocal:

I don't think this PR is still necessary, I'm closing it.

Thanks!

PR #3316, opened 2026-01-23
[LiveComponent] Fix race condition with inputs being modified during a request

### LINKED ISSUE #2899: [LiveComponent] Form inputs are sometimes cleared
When working with live forms, we see that sometimes inputs are being cleared. Its probably related to the component rerender while typing. It is hard to reproduce by hand (although our customers sometimes experience this), we see this consistently when using automated browser tests. 

Easy way to reproduce is with the following testcafe script: 
```type=ts
import {Selector} from "testcafe";

fixture `Test`;

test('Test bug', async t =>  {
    await t.navigateTo('https://ux.symfony.com/demos/live-component/auto-validating-form');

    await t.typeText('#registration_form_password', 'verysecretpassword', { replace: true, paste: true });
    await t.typeText('#registration_form_email', 'someemail@symfony.com', { replace: true, paste: true });

    await t.click('#registration_form_terms');
    await t.click('form[name=registration_form] button[type=submit]');

    await t.expect(Selector('div').withText('Welcome someemail@symfony.com!').visible).ok();
});
```

Save this as `livebug.ts`, install testcafe with `npm i -g testcafe`, then run with `testcafe chrome livebug.ts`.
It works fine when using `--speed 0.5`  as extra argument. 

https://github.com/user-attachments/assets/97dd87f1-8685-4aa9-93c2-d97dc3aef7df

tijnema:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #2899
| License        | MIT

As discussed in issue #2899, sometimes form inputs would be cleared. This appears to be a race condition where the input was not sent to the server, but the server response would override the input if it had been modified between request start and morph, even if the server didn't modify anything.

This fixes the issue by checking if an input value is different between now and the value sent to the server, and will then not update the value from the server. Server side changes are still supported, because the code will automatically send a new request for the changed input. 

I tried writing a test for it, but so far couldn't reproduce it with a test

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.d.ts
+++ b/src/LiveComponent/assets/dist/live_controller.d.ts
@@ -69,6 +69,7 @@ declare class export_default{
     set(name: string, value: any): boolean;
     getOriginalProps(): any;
     getDirtyProps(): any;
+    getPendingProps(): any;
     getUpdatedPropsFromParent(): any;
     flushDirtyPropsToPending(): void;
     reinitializeAllProps(props: any): void;
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1119,7 +1119,22 @@ var syncAttributes = (fromEl, toEl) => {
     toEl.setAttribute(attr.name, attr.value);
   }
 };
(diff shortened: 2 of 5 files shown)


Kocal:

Thanks!

Status: Needs Work

PR #3342, opened 2026-02-11
[Translator] Improve performance of dumper under certain condition

mkrauser:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | no
| License        | MIT

This PR moves the fileAppend-Calls out of the foreach to drastically improve performance, close to 500x on my Machine and in my project (with lots of translations). 

```
# current time in my project:
 [INFO] "Symfony\UX\Translator\CacheWarmer\TranslationsCacheWarmer" completed in 296866.54ms.  
 
# with the patch:
 [INFO] "Symfony\UX\Translator\CacheWarmer\TranslationsCacheWarmer" completed in 624.45ms.   
 ```

### DIFF
--- a/src/Translator/src/TranslationsDumper.php
+++ b/src/Translator/src/TranslationsDumper.php
@@ -99,29 +99,31 @@ class TranslationsDumper
             );
         }
 
+        $additions = [];
+        $typescriptAdditions = [];
         foreach ($this->getTranslations($catalogues, $excludedDomains, $includedDomains, $includeKeysRegex, $excludeKeysRegex) as $translationId => $translationsByDomainAndLocale) {
             $translationId = str_replace('"', '\\"', $translationId);
-            $this->filesystem->appendToFile($fileIndexJs, \sprintf(
+
+            $additions[] = \sprintf(
                 '    "%s": %s,%s',
                 $translationId,
                 json_encode(['translations' => $translationsByDomainAndLocale], \JSON_THROW_ON_ERROR),
                 "\n"
-            ));
+            );
 
             if ($dumpTypeScript) {
-                $this->filesystem->appendToFile($fileIndexDts, \sprintf(
+                $typescriptAdditions[] = \sprintf(
                     '    "%s": %s;%s',
                     $translationId,
                     $this->getTranslationsTypeScriptTypeDefinition($translationsByDomainAndLocale),
                     "\n"
-                ));
+                );
             }
         }
 
-        $this->filesystem->appendToFile($fileIndexJs, '};'."\n");
-
+        $this->filesystem->appendToFile($fileIndexJs, implode('', $additions).'};'."\n");
         if ($dumpTypeScript) {
-            $this->filesystem->appendToFile($fileIndexDts, '};'."\n");
+            $this->filesystem->appendToFile($fileIndexDts, implode('', $typescriptAdditions).'};'."\n");
         }
     }
 

### CODE REVIEW

**Comment by Kocal** [comment]
Interesting, I'm not seeing such performance improvements on my side.

Ran `blackfire run symfony console cache:clear` on my test app with 3,130 translations keys for 8 locales:
image

**Comment by mkrauser** [comment]
I dug a little further, but unfortunatelly I don't have blackfire available. 

I have ~4k translations-keys for 2 locales on a very powerful machine (M4 MacBook Pro) with php 8.5. I added a stopwatch to time every iteration of the loop. 
The first ~100 Iterations take at about 5ms each and then it is getting slower and slower. The last iterations take about 150ms each with the occasional 500ms inbetween. It seems that the fileAppend-Call  is getting a bit slower with every iteration. 
I don't know exactly why this happens to that extend. My Team has the same problem, so maybe our dev-setup slows down things.

Nevertheless, IMHO it makes sense to bulk-write here. Every call to fileAppend checks, if the directory exists, sets and resets the error-handler and then opens the file, do the writing and closes the file... just to do it again in the next iteration.

Without this patch it just took 137s vs. 238.60ms (just for the TranslationsCacheWarmer)

**Comment by Kocal** [comment]
Thank you for the additional information, maybe my _benchmark_ was wrong, I will double-check.

> Nevertheless, IMHO it makes sense to bulk-write here. Every call to fileAppend checks, if the directory exists, sets and resets the error-handler and then opens the file, do the writing and closes the file... just to do it again in the next iteration.

TBH I'm not very sure, when working on https://github.com/symfony/ux/pull/3208, I was like "maybe perfs could be improved if I write little by little in the file, and so the memory does not explode", but I may be wrong. Especially if I/O is slow. 

And maybe we shouldn't care about the memory consumption while warming the cache? But I remember that some systems are limited with slow RAM, we should be careful. 

I'm currently creating new Blackfire profiles.

**EDIT:** See the [Blackfire comparaison](https://app.blackfire.io/profiles/compare/a804a697-24d2-45b1-bd97-75c7ef914a6f...b1b5d19b-6a54-4722-aac6-4d635331eb50/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=Symfony%5CUX%5CTranslator%5CTranslationsDumper%3A%3Adump&callname=Symfony%5CUX%5CTranslator%5CTranslationsDumper%3A%3Adump). Dumping my translations on my lab project lasts ~3 seconds less, but that's mainly and randomly due to `getTranslationsTypeScriptTypeDefinition()` - which is not related to your changes: 
image

I'm using a M1 Pro and I won't have access to a M4 Pro until monday. Do you run your `cache:clear` in a Docker container or something equivalent?

**Comment by mkrauser** [comment]
> Do you run your `cache:clear` in a Docker container or something equivalent?

No, I'm doing it directly on the host. But the directory is "bind-mounted" in a docker-container. 
I just checked, and indeed that was slowing down the file-writes, even if the command is executed on the host-system. If I stop the docker containers it runs faster (50s), but still not fast for some reason.  

I'll try to narrow it down further. 

I also checked the Memory-Usage on my System with ~4k translations and 2 locales. 

~2M without the patch
~4M with the patch

**Review by Kocal — APPROVED** [review]

**Comment by Kocal** [comment]
Thank you @mkrauser.

File: src/Translator/src/TranslationsDumper.php [hunk 1]
```diff
@@ -99,29 +99,31 @@ public function dump(
...
                 '    "%s": %s,%s',
                 $translationId,
                 json_encode(['translations' => $translationsByDomainAndLocale], \JSON_THROW_ON_ERROR),
                 "\n"
-            ));
+            );
 
             if ($dumpTypeScript) {
-                $this->filesystem->appendToFile($fileIndexDts, \sprintf(
+                $typescriptAdditions[] = \sprintf(
```
stloyd, line 115: Did you check using simple strings instead of array + implode?
```php
$typescriptAdditions = '';
// ...
$typescriptAdditions .= \sprintf(/**/);

mkrauser: I did not, but I guess this will improve performance only marginally.

PR #3346, opened 2026-02-18
[LiveComponent] Update doc for component initialization

### LINKED ISSUE #2937: Initializing a component with a "Post"
The documentation states that a LiveProp may be an entity however there's no example of how to instantiate the component with the entity. And... There are tricks here!

In my case I had a `mount()` function:

```php
#[LiveProp]
public Post $post;

public function mount(Post $post)
{
    // .. doing something
}
```

It's actually required here to hydrate manually the prop:

```php
#[LiveProp]
public Post $post;

public function mount(Post $post)
{
    $this->post = $post;
    // .. doing something
}
```

More like a doc issue than an actual problem.

MrYamous:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | Fix #2937 
| License        | MIT

Update `LiveComponent` documentation to address #2937

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -525,6 +525,17 @@ Doctrine entity objects are a special case for ``LiveProp``::
         public Post $post;
     }
 
+.. note::
+
+    If you have a ``mount`` method declared in your component, it's required to
+    manually hydrate the prop::
+
+    public function mount(Post $post)
+    {
+        $this->post = $post;
+        // .. doing something
+    }
+
 If the ``Post`` object is persisted, its dehydrated to the entity's ``id`` and then
 hydrated back by querying the database. If the object is unpersisted, it's dehydrated
 to an empty array, then hydrated back by creating an *empty* object
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -527,8 +527,8 @@ Doctrine entity objects are a special case for ``LiveProp``::
 
 .. note::
 
-    If you have a ``mount`` method declared in your component, it's required to
-    manually hydrate the prop::
+    If you're using entities as ``LiveProp`` and having a ``mount`` method declared in
+    your component, it's required to manually hydrate the corresponding prop::
 
     public function mount(Post $post)
     {


src/LiveComponent/doc/index.rst
```diff
@@ -525,6 +525,17 @@ Doctrine entity objects are a special case for ``LiveProp``::
         public Post $post;
     }
 
+.. note::
+
+    If you have a ``mount`` method declared in your component, it's required to
+    manually hydrate the prop::
```
norkunas: While written like this, I'd understood that if you have `mount` method, then you need to set all properties manually, which is not the case - if you don't list them in mount, they are mounted automatically

MrYamous: I see your point, i'll rewrite to be more accurate

src/LiveComponent/doc/index.rst
```diff
@@ -525,6 +525,17 @@ Doctrine entity objects are a special case for ``LiveProp``::
...
+.. note::
+
+    If you're using entities as ``LiveProp`` and having a ``mount`` method declared in
+    your component, it's required to manually hydrate the corresponding prop::
+
+    public function mount(Post $post)
+    {
+        $this->post = $post;
+        // .. doing something
+    }
```
Kocal, line 537: I think it's still not right, or else I didn't understand the issue... 😅 

It is not related to `\Symfony\UX\TwigComponent\ComponentFactory::mount()`, that removes from `$data`, the properties defined in `mount()` arguments, requiring you to manually assign the property?

Given this example
```php
#[AsTwigComponent]
class MyComponent {
    public string $foo;
    public string $bar;

    public function mount(string $foo, string $bar) {
        $this->foo = $foo;
        // missing $this->bar = $bar;
    }
}
```

Calling `<twig:MyComponent foo="foo" bar="bar" />` will only set `foo` property if I'm not wrong.


To me, it is not related to LiveComponent at all.

smnandre: In fact I'm pretty sure it's due to the mount/postHydrate confusion.... our doc is responsible of :)

Kocal:
(any updates? :D)

Kocal:

Thanks!

PR #3347, opened 2026-02-18
[Cropperjs] Add image rotation in php side

### LINKED ISSUE #2930: [CropperJS] Rotation of the image not applied at PHP level
When using CropperJS in the frontend, it provides the ability to rotate images. The selected rotation angle is included in the options object passed by CropperJS. However, the getCroppedImage() method within the symfony/ux-cropperjs component does not apply this rotation to the final cropped image.

The Crop model class in the component currently lacks a direct method to retrieve the rotation information from the options array, other than through the full, JSON-encoded options string. This makes it challenging to properly process the rotation in my own code.

I had considered submitting a Pull Request to directly add the rotation logic to the createCroppedImage() method:

```php
// Rotate (intervention/image rotates counterclockwise)
if (isset($this->options['rotate']) && $this->options['rotate'] !== 0) {
    $image->rotate(-1 * $this->options['rotate']); // Negative because intervention/image rotates counter-clockwise
}
```

However, implementing this directly would introduce a breaking change. Users of the bundle might already have their own logic in place to handle rotation, and this new functionality could lead to double rotations.

The symfony/ux-cropperjs component should provide a way to correctly apply the rotation specified by CropperJS in the frontend to the cropped image on the server side, without disrupting existing implementations.

What is the most suitable approach to add this functionality while maintaining backward compatibility? Here are a few options to consider:

- Add a method to the Crop model to directly retrieve options as PHP array. This would simplify access to the rotation value.
- Introduce a separate, optional method or parameter to getCroppedImage() that specifies whether or not the rotation should be applied. For example, getCroppedImage(..., bool $applyRotation = false). This would give users control and prevent a BC-break.

MrYamous:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | yes
| Documentation? | no
| Issues         | Fix #2930
| License        | MIT

I tweaked a bit the BC layer in comparison to discussion in issue cause to be honest I didn't understand how it was supposed to be done in the described way 😅 
I hope it will be fine

### DIFF
--- a/src/Cropperjs/CHANGELOG.md
+++ b/src/Cropperjs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.34
+
+- Deprecate calling `Crop::getCroppedImage()` and `Crop::getCroppedThumbnail()` without explicitly passing `true` for the $applyRotation parameter. It will always be `true` in 3.0
+
 ## 2.30
 
 - Ensure compatibility with PHP 8.5
--- a/src/Cropperjs/src/Model/Crop.php
+++ b/src/Cropperjs/src/Model/Crop.php
@@ -46,6 +46,7 @@ class Crop
         'y' => 0,
         'width' => null,
         'height' => null,
+        'rotate' => 0,
     ];
 
     public function __construct(ImageManager $imageManager, string $filename)
@@ -54,8 +55,11 @@ class Crop
         $this->filename = $filename;
     }
 
-    public function getCroppedThumbnail(int $maxWidth, int $maxHeight, string $format = 'jpg', int $quality = 80): string
+    public function getCroppedThumbnail(int $maxWidth, int $maxHeight, string $format = 'jpg', int $quality = 80, bool $applyRotation = false): string
     {
+        if (\func_num_args() < 5 || false === $applyRotation) {
+            trigger_deprecation('symfony/ux-cropperjs', '2.34', 'Not passing "true" to the "$applyRotation" argument of "%s()" is deprecated. Rotation will be applied by default in Symfony UX 3.0.', __METHOD__);
+        }
         $image = $this->createCroppedImage();
 
         $image->resize($maxWidth, $maxHeight, static function ($constraint) {
@@ -63,13 +67,20 @@ class Crop
             $constraint->upsize();
         });
 
+        if ($applyRotation && !empty($this->options['rotate'])) {
+            $image->rotate(-1 * $this->options['rotate']);
+        }
+
         $image->encode($format, $quality);
 
         return $image->getEncoded();
     }
 
-    public function getCroppedImage(string $format = 'jpg', int $quality = 80): string
+    public function getCroppedImage(string $format = 'jpg', int $quality = 80, bool $applyRotation = false): string
     {
+        if (\func_num_args() < 3 || false === $applyRotation) {
+            trigger_deprecation('symfony/ux-cropperjs', '2.34', 'Not passing "true" to the "$applyRotation" argument of "%s()" is deprecated. Rotation will be applied by default in Symfony UX 3.0.', __METHOD__);
+        }
         $image = $this->createCroppedImage();
 
         // Max size
@@ -78,6 +89,10 @@ class Crop
                 $constraint->aspectRatio();
                 $constraint->upsize();
             });
+        }
+
+        if ($applyRotation && !empty($this->options['rotate'])) {
+            $image->rotate(-1 * $this->options['rotate']);
         }
 
         $image->encode($format, $quality);
--- a/src/Cropperjs/composer.json
+++ b/src/Cropperjs/composer.json
@@ -32,6 +32,7 @@
         "intervention/image": "^2.5",
         "symfony/config": "^5.4|^6.0|^7.0|^8.0",
         "symfony/dependency-injection": "^5.4|^6.0|^7.0|^8.0",
+        "symfony/deprecation-contracts": "^2.5|^3",
         "symfony/form": "^5.4|^6.0|^7.0|^8.0",
         "symfony/http-kernel": "^5.4|^6.0|^7.0|^8.0",
         "symfony/options-resolver": "^5.4|^6.0|^7.0|^8.0",
--- a/src/Cropperjs/tests/Model/CropTest.php
+++ b/src/Cropperjs/tests/Model/CropTest.php
@@ -1,0 +1,123 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Cropperjs\Tests\Model;
+
+use Intervention\Image\ImageManager;
+use PHPUnit\Framework\TestCase;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
+use Symfony\UX\Cropperjs\Model\Crop;
+
+/**
+ * @internal
+ */
+class CropTest extends TestCase
+{
+    use ExpectDeprecationTrait;
+    private string $testImagePath;
+
+    protected function setUp(): void
+    {
+        $this->testImagePath = tempnam(sys_get_temp_dir(), 'crop_test_').'.jpg';
+        ob_start();
+        imagejpeg(imagecreatetruecolor(200, 100), $this->testImagePath);
+        ob_end_clean();
+    }
+
+    protected function tearDown(): void
+    {
+        if (file_exists($this->testImagePath)) {
+            unlink($this->testImagePath);
+        }
+    }
+
+    private function createCrop(int $rotate = 0): Crop
+    {
+        $imageManager = new ImageManager();
+        $crop = new Crop($imageManager, $this->testImagePath);
+
+        $crop->setOptions(json_encode([
+            'width' => null,
+            'height' => null,
+            'rotate' => $rotate,
+        ]));
+
+        return $crop;
+    }
+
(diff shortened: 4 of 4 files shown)


MrYamous:
Tests fail looks unrelated

src/Cropperjs/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 2.30
 
+- Deprecate call `Crop::getCroppedThumbnail` and `Crop::getCroppedImage()` without `$applyRotation` parameter. Will default to true in 3.0.
```
Kocal:
```suggestion
## 2.33

- Deprecate passing `false` to the `$applyRotation` argument of `Crop::getCroppedThumbnail()` and `Crop::getCroppedImage()`, it will always be `true` in 3.0.
```

src/Cropperjs/src/Model/Crop.php
```diff
@@ -54,22 +60,32 @@ public function __construct(ImageManager $imageManager, string $filename)
         $this->filename = $filename;
     }
 
-    public function getCroppedThumbnail(int $maxWidth, int $maxHeight, string $format = 'jpg', int $quality = 80): string
+    public function getCroppedThumbnail(int $maxWidth, int $maxHeight, string $format = 'jpg', int $quality = 80, bool $applyRotation = false): string
     {
+        if (\func_num_args() < 5) {
+            trigger_deprecation('symfony/ux-cropperjs', '2.0', 'Calling "%s()" without $applyRotation is deprecated and will default to true in 3.0', __METHOD__);
+        }
```
Kocal, line 62: I think it's more something like:

```suggestion
        if (\func_num_args() < 5 || false === $applyRotation) {
            trigger_deprecation('symfony/ux-cropperjs', '2.33', 'Passing false to the $applyRotation argument of "%s()" is deprecated and will always be true in 3.0.', __METHOD__);
        }
```

src/Cropperjs/src/Model/Crop.php
```diff
@@ -54,22 +60,32 @@ public function __construct(ImageManager $imageManager, string $filename)
...
 
         return $image->getEncoded();
     }
 
-    public function getCroppedImage(string $format = 'jpg', int $quality = 80): string
+    public function getCroppedImage(string $format = 'jpg', int $quality = 80, bool $applyRotation = false): string
     {
+        if (\func_num_args() < 3) {
+            trigger_deprecation('symfony/ux-cropperjs', '2.0', 'Calling "%s()" without $applyRotation is deprecated and will default to true in 3.0', __METHOD__);
+        }
```
Kocal:
```suggestion
        if (\func_num_args() < 3 || false === $applyRotation) {
            trigger_deprecation('symfony/ux-cropperjs', '2.33', 'Passing false to the $applyRotation argument of "%s()" is deprecated and will always be true in 3.0.', __METHOD__);
        }
```

Kocal (changes requested):
Thanks for tackling this :)

src/Cropperjs/src/Model/Crop.php
MrYamous: Now I have doubts about the need for `func_num_args()`, isn't enough to check `false === $applyRotation` ? 

With counting args, something like 
```
$crop->getCroppedThumbnail(maxWidth: 100, maxHeight:100, applyRotation:true);
```
will trigger depreciation

Kocal: Yes let's remove the func_num_args()

src/Cropperjs/src/Model/Crop.php
```diff
@@ -54,22 +60,32 @@ public function __construct(ImageManager $imageManager, string $filename)
...
+            trigger_deprecation('symfony/ux-cropperjs', '2.33', 'Passing false to the $applyRotation argument of "%s()" is deprecated and will always be true in 3.0.', __METHOD__);
+        }
         $image = $this->createCroppedImage();
 
         $image->resize($maxWidth, $maxHeight, static function ($constraint) {
             $constraint->aspectRatio();
             $constraint->upsize();
         });
 
+        if ($this->options['rotate'] !== 0 && $applyRotation) {
```
Kocal: Same for the other one
```suggestion
        if ($applyRotation && 0 !== $this->options['rotate']) {
```

MrYamous: Ooops fabbot reported too but I forgot it

src/Cropperjs/src/Model/Crop.php
```diff
@@ -36,6 +36,11 @@ class Crop
      */
     private $maxHeight;
 
+    /**
+     * @var int
+     */
+    private $rotate;
```
Kocal: It's unused right?

src/Cropperjs/src/Model/Crop.php
```diff
@@ -54,22 +60,32 @@ public function __construct(ImageManager $imageManager, string $filename)
...
+            trigger_deprecation('symfony/ux-cropperjs', '2.33', 'Passing false to the $applyRotation argument of "%s()" is deprecated and will always be true in 3.0.', __METHOD__);
+        }
         $image = $this->createCroppedImage();
 
         $image->resize($maxWidth, $maxHeight, static function ($constraint) {
             $constraint->aspectRatio();
             $constraint->upsize();
         });
 
+        if ($applyRotation && 0 !== $this->options['rotate']) {
```
Kocal: The key `rotate` may not exists, right?
```suggestion
        if ($applyRotation && 0 !== $this->options['rotate'] ?? 0) {
```

Kocal (changes requested):
Sorry, I kind of forgot about this PR; I didn't realize you'd made the changes.

Could you add tests for `getCroppedImage(applyRotation: true)` and `getCroppedImage(applyRotation: false)`?

Thanks!

src/Cropperjs/src/Model/Crop.php
MrYamous: Oh, I missed `options` could be defined without the key

MrYamous:
Thanks @Kocal for review, I fixed last points and added tests

src/Cropperjs/src/Model/Crop.php
```diff
@@ -54,22 +55,32 @@ public function __construct(ImageManager $imageManager, string $filename)
         $this->filename = $filename;
     }
 
-    public function getCroppedThumbnail(int $maxWidth, int $maxHeight, string $format = 'jpg', int $quality = 80): string
+    public function getCroppedThumbnail(int $maxWidth, int $maxHeight, string $format = 'jpg', int $quality = 80, bool $applyRotation = false): string
     {
+        if (false === $applyRotation) {
+            trigger_deprecation('symfony/ux-cropperjs', '2.33', 'Passing false to the $applyRotation argument of "%s()" is deprecated and will always be true in 3.0.', __METHOD__);
```
Kocal: Using `trigger_deprecation` requires `"symfony/deprecation-contracts": "^2.5|^3"` dependency

src/Cropperjs/src/Model/Crop.php
Kocal: 💡 TIL!

I'm so sorry, I didn't test it last time, but nope, `func_num_args()` will equals 5 even when calling `$crop->getCroppedThumbnail(maxWidth: 100, maxHeight:100, applyRotation:true)`, see https://3v4l.org/f5DbD#vnull

Please re-add the `func_num_args()` check, so the deprecation will be correctly triggered. Thank you 🙏🏻

src/Cropperjs/src/Model/Crop.php
```diff
@@ -54,22 +55,32 @@ public function __construct(ImageManager $imageManager, string $filename)
...
+
         $image->encode($format, $quality);
 
         return $image->getEncoded();
     }
 
-    public function getCroppedImage(string $format = 'jpg', int $quality = 80): string
+    public function getCroppedImage(string $format = 'jpg', int $quality = 80, bool $applyRotation = false): string
     {
+        if (false === $applyRotation) {
```
Kocal: And.. same here!

src/Cropperjs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.33
+
+- Deprecate passing `false` to the `$applyRotation` argument of `Crop::getCroppedThumbnail()` and `Crop::getCroppedImage()`, it will always be `true` in 3.0.
```
Kocal:
```suggestion
- Deprecate calling `Crop::getCroppedImage()` and `Crop::getCroppedThumbnail()` without explicitly passing `true` for the $applyRotation parameter. It will always be `true` in 3.0
```

Kocal (changes requested):
The final reviews round 🙏🏻

src/Cropperjs/src/Model/Crop.php
Kocal: And for the message:
```php
 trigger_deprecation('symfony/ux-cropperjs', '2.34', 'Not passing "true" to the "$applyRotation" argument of "%s()" is deprecated. Rotation will be applied by default in Symfony UX 3.0.', __METHOD__);
```

MrYamous:
Looks better :)

src/Cropperjs/src/Model/Crop.php
MrYamous: Shame on me for didn't read the `func_num_args()` doc correctly, it's written on php website...

> regard to named arguments, by treating the arguments as if they were all passed positionally, and missing arguments are replaced with their defaults

Kocal: Today we learned :)

Kocal (approved):

Kocal:
Thank you @MrYamous.

PR #3361, opened 2026-02-27
[LiveComponent][TwigComponent] Add support for dynamic template resolution

xDeSwa:
| Q              | A
| -------------- | ---
| Bug fix?      | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| Issues         | -
| License        | MIT

### Description

This PR introduces the ability to dynamically resolve component templates via a dedicated method in the component class. This is particularly useful for building responsive components or multi-theme applications where the same logic (PHP class) needs to render different templates based on its internal state or props.

Instead of relying on a "magic" method, this implementation uses a new `FromMethod` attribute passed to the `template` parameter. This approach ensures explicit intent and maintains compatibility with Symfony's service tag requirements.

**How it works**

- Define a public method in your component that returns the template path as a string.
- `Pass a new FromMethod('yourMethodName')` to the `#[AsTwigComponent]` or `#[AsLiveComponent]` attribute.

**TwigComponent Example**
```php
#[AsTwigComponent('user_profile', template: new FromMethod('getTemplateForRole'))]
class UserProfile
{
    public function getTemplateForRole(): string
    {
        if ($this->isGranted('ROLE_ADMIN')) {
            return 'components/user_profile/admin_dashboard.html.twig';
        }

        return 'components/user_profile/public_card.html.twig';
    }
}
```

**LiveComponent Example**
```php
#[AsLiveComponent('checkout_wizard', template: new FromMethod('getStepTemplate'))]
class CheckoutWizard
{
    #[LiveProp]
    public int $currentStep = 1;

    #[LiveAction]
    public function nextStep(): void
    {
        $this->currentStep++;
    }

    public function getStepTemplate(): string
    {
        return match ($this->currentStep) {
            1 => 'components/checkout/step_1_address.html.twig',
            2 => 'components/checkout/step_2_shipping.html.twig',
            3 => 'components/checkout/step_3_payment.html.twig',
            default => 'components/checkout/summary.html.twig',
        };
    }
}
```

<details>
<summary>First PR</summary>
### Description

When building responsive or multi-theme applications, developers frequently need to reuse the exact same component logic (PHP class) with entirely different Twig templates depending on the context (e.g., a Desktop Search Dropdown vs. a Mobile Search Offcanvas).

Currently, the only ways to achieve this are:
1. **Duplicating the component class** just to change the `#[AsLiveComponent(template: '...')]` attribute.
2. **Creating a global `PreRenderEvent` listener**, which breaks component encapsulation and litters the codebase with boilerplate listeners.

This PR solves this by introducing a magic `getTemplate(): string` method. If a component defines this public method, the `ComponentRenderer` will use its return value as the template, overriding the default attribute metadata. This aligns perfectly with existing UX magic methods like `mount()` and `hydrate()`.

**Additional Note:** P.S. As this relies on the underlying ComponentRenderer, the getTemplate() method also works out-of-the-box for LiveComponents. It might be worth mentioning this in the LiveComponent documentation as well.

### Example Usage

Instead of duplicating components, developers can now encapsulate the logic gracefully:

```php
#[AsLiveComponent('SearchForm')]
class SearchForm extends AbstractController
{
    #[LiveProp]
    public string $theme = 'desktop';

    // The component dynamically decides its own template
    public function getTemplate(): string
    {
        return $this->theme === 'mobile' 
            ? 'components/MobileSearch.html.twig'
            : 'components/DesktopSearch.html.twig';
    }
}


{# Renders the mobile template #}
{{ component('SearchForm', { theme: 'mobile' }) }}
</details>

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.33
 
 - Add `fetch_credentials` option to configure the fetch API credentials mode for cross-origin requests.
-  This is useful when embedding a Live Component from a different domain that requires cookie-based authentication (e.g., JWT stored in cookies).
+  This is useful when embedding a Live Component from a different domain that requires cookie-based authentication (e.g., JWT stored in cookies)
 
     Global configuration in `config/packages/live_component.yaml`:
 
@@ -21,6 +21,8 @@
         // ...
     }
     ```
+
+- Add support for dynamic template resolution with `AsLiveComponent(template: FromMethod('customFunction'))`
 
 ## 2.31
 
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -75,6 +75,28 @@ Want some demos? Check out https://ux.symfony.com/live-component#demo
         {
             // ...
 
+Dynamic Templates
+-----------------
+
+.. versionadded:: 2.33
+
+    Live components support dynamic template resolution using the ``FromMethod`` attribute, just like standard Twig components.
+
+This is particularly useful for complex live components that need to switch views based on user interaction::
+
+    #[AsLiveComponent(template: new FromMethod('getTemplate'))]
+    class SearchResults
+    {
+        #[LiveProp(writable: true)]
+        public string $layout = 'rows';
+
+        public function getTemplate(): string
+        {
+            return 'rows' === $this->layout
+                ? 'components/SearchResults/rows.html.twig'
+                : 'components/SearchResults/grid.html.twig';
+        }
+    }
 
 Installation
 ------------
--- a/src/LiveComponent/src/Attribute/AsLiveComponent.php
+++ b/src/LiveComponent/src/Attribute/AsLiveComponent.php
@@ -13,6 +13,7 @@ namespace Symfony\UX\LiveComponent\Attribute;
 
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+use Symfony\UX\TwigComponent\Attribute\FromMethod;
 
 /**
  * An attribute to register a LiveComponent.
@@ -33,7 +34,7 @@ final class AsLiveComponent extends AsTwigComponent
 
     /**
      * @param string|null              $name              The component name (ie: TodoList)
-     * @param string|null              $template          The template path of the component (ie: components/TodoList.html.twig).
+     * @param string|FromMethod|null   $template          The template path of the component (ie: components/TodoList.html.twig), or a reference to a component's method (ie: FromMethod('getTemplate')
      * @param string|null              $defaultAction     The default action to call when the component is mounted (ie: __invoke)
      * @param bool                     $exposePublicProps Whether to expose every public property as a Twig variable
      * @param string                   $attributesVar     The name of the special "attributes" variable in the template
@@ -44,7 +45,7 @@ final class AsLiveComponent extends AsTwigComponent
      */
     public function __construct(
         ?string $name = null,
-        ?string $template = null,
+        string|FromMethod|null $template = null,
         ?string $defaultAction = null,
         bool $exposePublicProps = true,
         string $attributesVar = 'attributes',
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## 2.33
 
-- Extended support for the `index.html.twig` template fallback when resolving namespaced anonymous components.
+- Extended support for the `index.html.twig` template fallback when resolving namespaced anonymous components
+- Add support for dynamic template resolution with `AsTwigComponent(template: FromMethod('getCustomFuntion'))`
 
 ## 2.32
 
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -382,6 +382,35 @@ use the full path of the template where the macro is defined:
         {{ message_formatter('...') }}
     </twig:Alert>
 
+Dynamic Templates
+-----------------
+
+.. versionadded:: 2.33
+
+    The ability to dynamically resolve templates via the ``FromMethod`` attribute was added.
+
+Sometimes, you need to render a different template based on the component state.
+
+It is possible to reference a component method by passing ``FromMethod`` to the ``template`` option of ``AsTwigComponent``, which will be used to compute the template to use before rendering the component::
+
+    namespace App\Twig\Components;
+
+    use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+    use Symfony\UX\TwigComponent\Attribute\FromMethod;
+
+    #[AsTwigComponent(template: new FromMethod('getTemplate'))]
+    class SearchResults
+    {
+        public string $layout = 'rows';
+
+        public function getTemplate(): string
+        {
+            return 'rows' === $this->layout
+                ? 'components/SearchResults/rows.html.twig'
+                : 'components/SearchResults/grid.html.twig';
+        }
+    }
+
 Fetching Services
 -----------------
 
--- a/src/TwigComponent/src/Attribute/AsTwigComponent.php
+++ b/src/TwigComponent/src/Attribute/AsTwigComponent.php
@@ -36,7 +36,9 @@ class AsTwigComponent
         private ?string $name = null,
 
         /**
-         * The template path of the component (ie: components/Button.html.twig).
+         * The template path of the component (ie: components/Button.html.twig)
+         * or a reference to a component method to resolve the template dynamically
+         * (ie: FromMethod('customFunction')).
          *
          * With the default configuration, the template path is resolved using
          * the component's name.
@@ -46,7 +48,7 @@ class AsTwigComponent
          *
          * @see https://symfony.com/bundles/ux-twig-component#component-template-path
          */
-        private ?string $template = null,
+        private string|FromMethod|null $template = null,
 
         /**
          * Whether to expose every public property as a Twig variable.
@@ -67,12 +69,19 @@ class AsTwigComponent
      */
     public function serviceConfig(): array
     {
-        return [
+        $config = [
             'key' => $this->name,
-            'template' => $this->template,
             'expose_public_props' => $this->exposePublicProps,
             'attributes_var' => $this->attributesVar,
         ];
+
+        if ($this->template instanceof FromMethod) {
+            $config['template_from_method'] = $this->template->method;
+        } else {
+            $config['template'] = $this->template;
+        }
+
+        return $config;
     }
 
     /**
--- a/src/TwigComponent/src/ComponentMetadata.php
+++ b/src/TwigComponent/src/ComponentMetadata.php
@@ -37,6 +37,14 @@ final class ComponentMetadata
     }
 
     /**
+     * @return string|null The method name to fetch the template dynamically
+     */
+    public function getTemplateFromMethod(): ?string
+    {
+        return $this->config['template_from_method'] ?? null;
+    }
+
+    /**
      * @return class-string The Component's FQCN
      */
     public function getClass(): string
--- a/src/TwigComponent/tests/Fixtures/Component/MethodTemplateComponent.php
+++ b/src/TwigComponent/tests/Fixtures/Component/MethodTemplateComponent.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+use Symfony\UX\TwigComponent\Attribute\FromMethod;
+
+#[AsTwigComponent('method_template_component', template: new FromMethod('getDynamicTemplate'))]
+class MethodTemplateComponent
+{
+    public string $type = 'default';
+    
+    public function getDynamicTemplate(): string
+    {
+        return sprintf('components/dynamic_%s.html.twig', $this->type);
+    }
+}
--- a/src/TwigComponent/tests/Integration/DynamicTemplateTest.php
+++ b/src/TwigComponent/tests/Integration/DynamicTemplateTest.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\TwigComponent\Tests\Integration;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\UX\TwigComponent\Test\InteractsWithTwigComponents;
+
+class DynamicTemplateTest extends KernelTestCase
+{
+    use InteractsWithTwigComponents;
+
+    public function testTemplateResolvedAutomaticallyWhenNotProvided()
+    {
+        // When no template is explicitly configured, the default resolution mechanism should be applied.
+        $rendered = (string) $this->renderTwigComponent('default_template_component');
+
+        $this->assertStringContainsString('Default Template Works!', $rendered);
+    }
+
+    public function testTemplateResolvedFromString()
+    {
+        // Ensures that a template defined as a string is properly resolved.
+        $rendered = (string) $this->renderTwigComponent('string_template_component');
+
+        $this->assertStringContainsString('String Template Works!', $rendered);
+    }
+
+    public function testTemplateChangesBasedOnComponentProps()
+    {
+        // 1. Default Type props
+        $html = $this->renderTwigComponent('method_template_component', ['type' => 'default']);
+        $this->assertStringContainsString('Default', $html);
+
+        // 2. Custom Type props
+        $html = $this->renderTwigComponent('method_template_component', ['type' => 'custom']);
+        $this->assertStringContainsString('Custom', $html);
+    }
+
+    public function testTemplateFromMethodThrowsExceptionWhenMethodIsMissing()
+    {
+        // Using FromMethod with a non-existing or non-callable method must throw a LogicException.
+        $this->expectException(\Twig\Error\RuntimeError::class);
+        $this->expectExceptionMessage('does not exist or is not callable');
+
+        $this->renderTwigComponent('missing_method_component');
+    }
+}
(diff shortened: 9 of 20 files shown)

### CODE REVIEW

File: src/TwigComponent/doc/index.rst [hunk 1]
```diff
@@ -382,6 +382,43 @@ use the full path of the template where the macro is defined:
...
+    class SearchForm
+    {
+        public string $theme = 'desktop';
+
+        public function getTemplate(): string
+        {
+            return 'mobile' === $this->theme
+                ? 'components/SearchForm/mobile.html.twig'
+                : 'components/SearchForm/desktop.html.twig';
+        }
```
Kocal, line 411: This is technically a BC break for users that already have a method `getTemplate` in their component.

Kocal (commented):
Thanks for your contribution! 

But I'm afraid that the current code could introduce BC breaks for people already using `getTemplate()` in their component, for a different purpose.

Instead, the behavior should be **explicit** e.g. by tweaking `AsTwigComponent`'s `template` attribute: 
```php
#[AsTwigComponent(template: new FromMethod('getTemplate'))]
class MyComponent
{
    public function getTemplate(): string
    {
        return 'mobile' === $this->theme
            ? 'components/SearchForm/mobile.html.twig'
            : 'components/SearchForm/desktop.html.twig';
    }
}
```

src/TwigComponent/tests/Integration/DynamicTemplateTest.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace Symfony\UX\TwigComponent\Tests\Integration;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\UX\TwigComponent\Test\InteractsWithTwigComponents;
+
+/**
+ * @author Sait KURT <saitkurt41@gmail.com>
+ */
```
Kocal: We do not add `@author` in tests

src/TwigComponent/tests/Integration/DynamicTemplateTest.php
```diff
@@ -0,0 +1,30 @@
...
+class DynamicTemplateTest extends KernelTestCase
+{
+    use InteractsWithTwigComponents;
+
+    public function testDynamicTemplateViaGetTemplateMethod(): void
+    {
+        $rendered = (string) $this->renderTwigComponent('dynamic_template_component');
+
+        $this->assertStringContainsString('Dynamic Template Works!', $rendered);
```
Kocal: Please tests both cases

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -122,6 +122,10 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
             $metadata->getAttributesVar() => $mounted->getAttributes(),
         ]);
 
+        if (method_exists($component, 'getTemplate') && \is_callable([$component, 'getTemplate'])) {
+            $event->setTemplate($component->getTemplate());
+        }
```
Kocal: If `$metadata->getTemplate() instanceof FromMethod && ...`

Kocal (changes requested):

src/TwigComponent/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 2.33
 
 - Extended support for the `index.html.twig` template fallback when resolving namespaced anonymous components.
+- Add support for dynamic template resolution via the `FromMethod` attribute.
```
Kocal: Sorry, looks like a dot was not supposed to be initially here

```suggestion
- Extended support for the `index.html.twig` template fallback when resolving namespaced anonymous components
- Add support for dynamic template resolution via the `FromMethod` attribute
```

src/LiveComponent/src/Attribute/AsLiveComponent.php
Kocal, line 80: What happens if you use LiveComponent 2.33+ with TwigComponent <2.33?

src/LiveComponent/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
     }
     ```
 
+- Add support for dynamic template resolution via the `FromMethod` attribute.
```
Kocal:
```suggestion
- Add support for dynamic template resolution with `AsLiveComponent(template: FromMethod('getTemplate'))`
```

src/LiveComponent/src/Attribute/AsLiveComponent.php
```diff
@@ -33,7 +34,7 @@ final class AsLiveComponent extends AsTwigComponent
 
     /**
      * @param string|null              $name              The component name (ie: TodoList)
-     * @param string|null              $template          The template path of the component (ie: components/TodoList.html.twig).
+     * @param string|FromMethod|null   $template          the template path of the component or a FromMethod instance
```
Kocal:
```suggestion
     * @param string|FromMethod|null   $template          The template path of the component (ie: components/TodoList.html.twig), or a reference to a component's method (ie: FromMethod('getTemplate'))  
```

src/TwigComponent/src/Attribute/AsTwigComponent.php
```diff
@@ -36,17 +36,18 @@ public function __construct(
         private ?string $name = null,
 
         /**
-         * The template path of the component (ie: components/Button.html.twig).
+         * The template path of the component (ie: components/Button.html.twig)
+         * or a FromMethod instance to resolve the template dynamically.
```
Kocal:
```suggestion
         * The template path of the component (ie: components/Button.html.twig), 
         * or a reference to a component method to resolve the template dynamically (ie: FromMethod('getTemplate'))
         
```

src/TwigComponent/src/Attribute/AsTwigComponent.php
```diff
@@ -36,17 +36,18 @@ public function __construct(
...
          *
          * With the default configuration, the template path is resolved using
          * the component's name.
          *
-         *      Button  ->  templates/components/Button.html.twig
-         *      Foo:Bar ->  templates/components/Foo/Bar.html.twig
+         * Button  ->  templates/components/Button.html.twig
+         * Foo:Bar ->  templates/components/Foo/Bar.html.twig
```
Kocal: To revert

src/TwigComponent/src/Attribute/AsTwigComponent.php
```diff
@@ -67,12 +68,20 @@ public function __construct(
...
-        return [
+        $config = [
             'key' => $this->name,
-            'template' => $this->template,
             'expose_public_props' => $this->exposePublicProps,
             'attributes_var' => $this->attributesVar,
         ];
+
+        if ($this->template instanceof FromMethod) {
+            $config['template'] = '';
```
Kocal: Why should we keep it empty?

src/TwigComponent/src/ComponentMetadata.php
```diff
@@ -31,9 +31,17 @@ public function getName(): string
     /**
      * @return string Component's twig template
      */
-    public function getTemplate(): string
+    public function getTemplate(): ?string
     {
-        return $this->config['template'];
+        return $this->config['template'] ?? null;
+    }
```
Kocal: Are we sure about this change?

src/TwigComponent/tests/Fixtures/Component/MethodTemplateComponent.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+use Symfony\UX\TwigComponent\Attribute\FromMethod;
+
+#[AsTwigComponent('method_template_component', template: new FromMethod('getDynamicTemplate'))]
+class MethodTemplateComponent
+{
+    public function getDynamicTemplate(): string
+    {
+        return 'components/custom_method_template.html.twig';
+    }
```
Kocal, line 25: It's not really dynamic anymore, can't you re-add the previous test where the template was changing given a certain props?

src/TwigComponent/src/ComponentRenderer.php
```diff
@@ -122,6 +122,25 @@ private function preRender(MountedComponent $mounted, array $context = []): PreR
...
+            $template = $component->$method();
+
+            if (!\is_string($template)) {
+                throw new \LogicException(\sprintf('The template method "%s" must return a string.', $method));
+            }
+        }
+
+        if ($template) {
+            $event->setTemplate($template);
+        }
```
Kocal: Why not doing this logic in `PreRenderEvent` constructor?

Also, it can be simplified:
```suggestion
        if ($method = $metadata->getTemplateFromMethod()) {
            if (!method_exists($component, $method) || !\is_callable([$component, $method])) {
                throw new \LogicException(\sprintf('The template method "%s" does not exist or is not callable on component "%s".', $method, get_debug_type($component)));
            }

            $template = $component->$method();

            if (!\is_string($template)) {
                throw new \LogicException(\sprintf('The template method "%s" must return a string.', $method));
            }
            $event->setTemplate($template);
            
        }
```

Kocal (changes requested):
Looks like some places will need some love as well
Image

src/LiveComponent/CHANGELOG.md
xDeSwa: - Add support for dynamic template resolution with `AsLiveComponent(template: FromMethod('customFunction'))`
 
 Does it make sense?

Kocal: yes

src/TwigComponent/src/ComponentRenderer.php
xDeSwa: > Why not doing this logic in `PreRenderEvent` constructor?
> 
> Also, it can be simplified:

Moving this logic to the `PreRenderEvent` constructor would make the Event object responsible for executing business logic (calling component methods), which violates the Single Responsibility Principle. Events should remain simple data carriers. Keeping the resolution logic in `ComponentRenderer` ensures that the Event stays 'dumb' and allows the Renderer to handle exceptions and logic flow appropriately.

----------------

The proposed simplification would skip setting the template for standard (static) components. My current implementation first retrieves the static template from metadata and then allows the dynamic method to override it if present. This ensures that both static `template: 'foo.html.twig'` and dynamic `template: new FromMethod('...')` work correctly.

I cannot put `setTemplate` only inside the dynamic method block, because that would prevent static templates (normal usage) from being set. The `ComponentRenderer` must ensure that even if there is no dynamic method, the static template from metadata is still passed to the `PreRenderEvent`.

Gemini did not accept these :)

smnandre:
Well, i'm sorry @xDeSwa but if this is not a BC i'm now 100% I never got what BC mean :(

xDeSwa:
> Well, i'm sorry @xDeSwa but if this is not a BC i'm now 100% I never got what BC mean :( 

I completely understand your concern, and you're right. The initial approach was indeed risky for BC. That's precisely why, based on @Kocal's suggestion, we pivoted to the FromMethod attribute in the current version. This keeps everything fully backward compatible while still adding the dynamic template feature.

smnandre:
@xDeSwa Hey.. I am sorry if it sounded otherwise, this was not about _this_ MR.

My gut tells me we keep stacking features on 2.x. Features that I suspect no one will maintain in a few month/years.
It takes one (great) PR to add a feature... at least two years of maintenance before we can change it.

(please believe me, I am _**thrilled**_ when people open PRs, contribute, bring ideas, or propose new features here)

Imagine this. Maybe, _just maybe_, someone here may soon have some OSS time paid by their employer (and please do note the word _maybe_ here)..

_**If**_ that happens, do you think a full day of work, every week, is better spent on UX 3.x, modern web APIs, incredibly cool things we could build with the Filesystem API, the Invoker API, WebAuth, View Transitions, and globally the entire Baseline 2025 new features _x_ PHP 8.5+ features, recent Symfony DI sugar, AI integrations, etc etc... 

Or....  🥁  ... should it be spent maintaining a major version released five years ago, supporting browsers that companies now consider major security risks, from a time when Yarn was trending and Less was still a thing?

If that ever becomes reality.. well 
Let's say I am not a big fan of the past..... all my mistakes live there 😅

smnandre:
(sorry if it does sound AI... I am the only and very proud author of this last punchline hahaha)

src/LiveComponent/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
     }
     ```
 
+- Add support for dynamic template resolution with AsLiveComponent(template: FromMethod('customFunction')
```
Kocal:
```suggestion
- Add support for dynamic template resolution with `AsLiveComponent(template: FromMethod('customFunction'))`
```

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,46 @@ use the full path of the template where the macro is defined:
...
+    {
+        public string $theme = 'desktop';
+
+        public function getDynamicTemplate(): string
+        {
+            return 'mobile' === $this->theme
+                ? 'components/SearchForm/mobile.html.twig'
+                : 'components/SearchForm/desktop.html.twig';
+        }
+    }
```
Kocal: I feel like this example could be more realistic, usually mobile-vs-desktop happens in the browser through CSS classes:
```suggestion
    #[AsTwigComponent(template: new FromMethod('getTemplate'))]
    class SearchResults
    {
        public string $layout = 'rows';

        public function getTemplate(): string
        {
            return 'rows' === $this->layout
                ? 'components/SearchResults/rows.html.twig'
                : 'components/SearchResults/grid.html.twig';
        }
    }
```

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,46 @@ use the full path of the template where the macro is defined:
...
+
+Now, you can switch the template by passing the ``theme`` prop:
+
+.. code-block:: html+twig
+
+    {# Renders components/SearchForm/mobile.html.twig #}
+    <twig:SearchForm theme="mobile" />
+
+    {# Renders components/SearchForm/desktop.html.twig #}
+    <twig:SearchForm theme="desktop" />
```
Kocal: (to adapt given the previous comment)

src/TwigComponent/src/Attribute/AsTwigComponent.php
Kocal: The comment is not resolved

src/TwigComponent/src/ComponentRenderer.php
Kocal: The comment is not resolved

Kocal (changes requested):
New round, please take care about all comments, not all of them are actually resolved :)

src/LiveComponent/doc/index.rst
```diff
@@ -75,6 +75,44 @@ Want some demos? Check out https://ux.symfony.com/live-component#demo
...
+        public function getStepTemplate(): string
+        {
+            return match ($this->currentStep) {
+                1 => 'components/checkout/step_1_address.html.twig',
+                2 => 'components/checkout/step_2_shipping.html.twig',
+                3 => 'components/checkout/step_3_payment.html.twig',
+                default => 'components/checkout/summary.html.twig',
+            };
+        }
+    }
```
Kocal: Why not re-using the example from https://github.com/symfony/ux/pull/3361#discussion_r2880911533 ?

I don't think this is a good example, since this "multi-step form" feature is now native with Symfony 7.4 https://symfony.com/blog/new-in-symfony-7-4-multi-step-forms

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,36 @@ use the full path of the template where the macro is defined:
...
     </twig:Alert>
 
+Dynamic Templates
+-----------------
+
+.. versionadded:: 2.33
+
+    The ability to dynamically resolve templates via the ``FromMethod`` attribute was added.
+
+Sometimes, you need to render a different template based on the component's state or props.
```
Kocal:
```suggestion
Sometimes, you need to render a different template based on the component state.
```

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,36 @@ use the full path of the template where the macro is defined:
...
+
+Instead of a static string, you can pass a ``FromMethod`` instance to the ``template`` option of the ``AsTwigComponent`` attribute. This tells Symfony to call the specified method to determine the template path at runtime::
```
Kocal:
```suggestion
It is possible to reference a component method by passing ``FromMethod`` to the ``template`` option of ``AsTwigComponent`, which will be used to compute the template to use before rendering the component::
```

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,36 @@ use the full path of the template where the macro is defined:
...
+    {
+        public function getTemplateForRole(): string
+        {
+            if ($this->isGranted('ROLE_ADMIN')) {
+                return 'components/user_profile/admin_dashboard.html.twig';
+            }
+
+            return 'components/user_profile/public_card.html.twig';
+        }
+    }
```
Kocal: Please re-use example from https://github.com/symfony/ux/pull/3361#discussion_r2880911533

Kocal (changes requested):
The last final round 🤞🏻

src/LiveComponent/doc/index.rst
```diff
@@ -75,6 +75,27 @@ Want some demos? Check out https://ux.symfony.com/live-component#demo
...
+.. versionadded:: 2.33
+
+    Live components support dynamic template resolution using the ``FromMethod`` attribute, just like standard Twig components.
+
+This is particularly useful for complex live components that need to switch views based on user interaction::
+
+    #[AsLiveComponent(template: new FromMethod('getTemplate'))]
+    class SearchResults
+    {
+        public string $layout = 'rows';
```
Kocal, line 91: It could be interesting to make it a liveprop

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,34 @@ use the full path of the template where the macro is defined:
...
+Dynamic Templates
+-----------------
+
+.. versionadded:: 2.33
+
+    The ability to dynamically resolve templates via the ``FromMethod`` attribute was added.
+
+Sometimes, you need to render a different template based on the component state.
+
+It is possible to reference a component method by passing ``FromMethod`` to the ``template`` option of ``AsTwigComponent`, which will be used to compute the template to use before rendering the component::
```
Kocal:
```suggestion
It is possible to reference a component method by passing ``FromMethod`` to the ``template`` option of ``AsTwigComponent``, which will be used to compute the template to use before rendering the component::
```

src/TwigComponent/doc/index.rst
```diff
@@ -382,6 +382,34 @@ use the full path of the template where the macro is defined:
...
+
+    use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+    use Symfony\UX\TwigComponent\Attribute\FromMethod;
+
+    #[AsTwigComponent(template: new FromMethod('getTemplate'))]
+    class SearchResults
+    {
+        public string $layout = 'rows';
+        public function getTemplate(): string
+        {
```
Kocal, line 407:
```suggestion
        public string $layout = 'rows';

        public function getTemplate(): string
        {
```

Kocal (changes requested):
The very last final round 🤞🏻

xDeSwa:
@Kocal all done check again please.

Kocal (approved):

Kocal:
Thanks you @xDeSwa, we finally did it 🙌

PR #3364, opened 2026-03-05
[LiveComponent] Add `debug:live-component` command

mbuliard:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | todo 
| Issues         | Fix #... 
| License        | MIT

Hello,

In several applications, I am using multiple events to communicate between LiveComponents.
It sometimes take me a little time to find out mistakes concerning these events.

That's why I propose this small command for debug specific to LiveComponents.

This command is largely copied from @StevenRenaux TwigComponentDebugCommand.
As the previous one, it can be used for listing or to see details of one component :

1) Listing : list all registered LiveComponents. Can filter those listening to an event with `--listening` filter.
image
image

2) Details : providing a LiveComponent name, display its LiveProps and LiveListeners.
image

### DIFF
--- a/src/LiveComponent/src/Command/LiveComponentDebugCommand.php
+++ b/src/LiveComponent/src/Command/LiveComponentDebugCommand.php
@@ -1,0 +1,254 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Command;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\TypeInfo\Type;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveArg;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadata;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+
+#[AsCommand(name: 'debug:live-component', description: 'Display live components and their usage for an application')]
+class LiveComponentDebugCommand extends Command
+{
+    public function __construct(
+        protected readonly LiveComponentMetadataFactory $metadataFactory,
+        protected readonly array $liveComponentList,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setDefinition([
+                new InputArgument(
+                    'name',
+                    InputArgument::OPTIONAL,
+                    'A LiveComponent name or part of the name'
+                ),
+                new InputOption(
+                    name: 'listening',
+                    mode: InputOption::VALUE_REQUIRED,
+                    description: 'Filter list to display only those listening to the given event'
+                ),
+            ])
+            ->setHelp(
+                <<<'EOF'
+                    The <info>%command.name%</info> display all the live components in your application.
+
+                    To list all live components:
+
+                        <info>php %command.full_name%</info>
+
+                    To get specific information about a component, specify its name (or a part of it):
+
+                        <info>php %command.full_name% ProductSearch</info>
+                    EOF
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $name = $input->getArgument('name');
+
+        if (\is_string($name)) {
+            $componentName = $this->findComponentName($io, $name, $input->isInteractive());
+            if (null === $componentName) {
+                $io->error(\sprintf('Unknown component "%s".', $name));
+
+                return Command::FAILURE;
+            }
+
+            $this->displayComponentDetails($io, $componentName);
+
+            return Command::SUCCESS;
+        }
+
+        $components = $this->findComponents($input->getOption('listening'));
+
+        $this->displayComponentsTable($components, $io);
+
+        return Command::SUCCESS;
+    }
+
+    private function findComponentName(SymfonyStyle $io, string $name, bool $interactive): ?string
+    {
+        $components = [];
+        foreach ($this->liveComponentList as $className) {
+            $metadata = $this->metadataFactory->getMetadata($className);
+            $componentName = $metadata->getComponentMetadata()->getName();
+            if ($name === $componentName) {
+                return $name;
+            }
+            if (str_contains($componentName, $name)) {
+                $components[$componentName] = $componentName;
+            }
+        }
+
+        if ($interactive && \count($components)) {
+            return $io->choice('Select one of the following component to display its information', array_values($components), 0);
+        }
+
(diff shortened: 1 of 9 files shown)


src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+                <<<'EOF'
+                    The <info>%command.name%</info> display all the live components in your application.
+
+                    To list all live components:
+
+                        <info>php %command.full_name%</info>
+
+                    To get specific information about a component, specify its name (or a part of it):
+
+                        <info>php %command.full_name% Alert</info>
```
Kocal: Nitpick, let's reuse `ProductSearch` component from the official document, I don't see how a `Alert` can be a live component:
```suggestion
                        <info>php %command.full_name% ProductSearch </info>
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $name = $input->getArgument('name');
+
+        if (\is_string($name)) {
+            $componentName = $this->findComponentName($io, $name, $input->isInteractive());
+            if (null === $componentName) {
+                $io->error(\sprintf('Unknown LiveComponent "%s".', $name));
```
Kocal:
```suggestion
                $io->error(\sprintf('Unknown component "%s".', $name));
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        if ($interactive && \count($components)) {
+            return $io->choice('Select one of the following component to display its information', array_values($components), 0);
+        }
+
+        return null;
+    }
+
+    private function listComponents(?string $eventFilter = null): array
```
Kocal:
```suggestion
     /**
     * @return array<string, LiveComponentMetadata>
     */
    private function findComponents(?string $eventFilter = null): array
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+    {
+        if (null === $eventFilter) {
+            return $this->componentClassMap;
+        }
```
Kocal, line 126:
```suggestion
        $components = [];
        if (null === $eventFilter) {
            foreach ($this->componentClassMap as $class => $name) {
                $components[$name] ??= $this->liveComponentMetadataFactory->getMetadata($name);
            }

            return $components;
        }
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+
+        $filteredComponents = [];
+        foreach ($this->componentClassMap as $name => $component) {
+            foreach (AsLiveComponent::liveListeners($component['class']) as $listener) {
+                if ($listener['event'] === $eventFilter) {
+                    $filteredComponents[$name] = $component;
```
Kocal:
```suggestion
                    $components[$name] ??= $this->liveComponentMetadataFactory->getMetadata($name);
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+        return null;
+    }
+
+    private function listComponents(?string $eventFilter = null): array
+    {
+        if (null === $eventFilter) {
+            return $this->componentClassMap;
+        }
+
+        $filteredComponents = [];
```
Kocal:
```suggestion
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+        foreach ($this->componentClassMap as $name => $component) {
+            foreach (AsLiveComponent::liveListeners($component['class']) as $listener) {
+                if ($listener['event'] === $eventFilter) {
+                    $filteredComponents[$name] = $component;
+                    break;
+                }
+            }
+        }
+
+        return $filteredComponents;
```
Kocal:
```suggestion
        return $components;
```

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
Kocal: Can't we re-use the `LiveComponentMetadataFactory` & `LiveComponentMetadata` instead?

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    private function displayComponentDetails(SymfonyStyle $io, string $name): void
+    {
+        $component = $this->componentClassMap[$name];
```
Kocal: Please do not directly depend on `$componentClassMap` but use the (internal) dedicated classes instead, that would makes things safer instead of relying on array-based and undocumented structures

src/LiveComponent/src/Command/LiveComponentDebugCommand.php
```diff
@@ -0,0 +1,248 @@
...
+                    static fn ($key, $value) => $key.': '.json_encode($value),
+                    array_keys($arguments),
+                    $arguments
+                )).')';
+
+            $propertyDisplay = $type.$propertyName.$defaultValueDisplay.$argumentsDisplay;
+            $properties[$property->name] = $propertyDisplay;
+        }
+
+        return $properties;
```
Kocal: The code cold be super-simplified by relying on `getAllLivePropsMetadata()` from `LiveComponentMetadata` (as suggested earlier)

src/TwigComponent/src/DependencyInjection/Compiler/TwigComponentPass.php
```diff
@@ -74,6 +74,7 @@ public function process(ContainerBuilder $container): void
                 $componentReferences[$tag['key']] = new Reference($id);
                 $componentNames[] = $tag['key'];
                 $componentClassMap[$tag['class']] = $tag['key'];
+                $definition->addTag('twig.component', $tag);
```
Kocal: Why?

Kocal (changes requested):
Thanks for the suggestion, here some comments

mbuliard:
@Kocal Thanks for the review !
I added the `LiveMetadataFactory` in the command as you suggested but I had to tweak the `LiveComponentMetadata->getAllLivePropsMetadata` to work without an actual component.

I also don't see a way to compute the name of a component in the LiveComponentPass without using directly TwigComponent configurations, that's why I added the `$definition->addTag('twig.component', $tag);` in TwigComponentPass, to set the tag with the name computed.

Kocal:
> I also don't see a way to compute the name of a component in the LiveComponentPass without using directly TwigComponent configurations, that's why I added the `$definition->addTag('twig.component', $tag);` in TwigComponentPass, to set the tag with the name computed.

Ah, I see!
Well, I'm not fan about your solution because it adds a second tag to the service definition, and we also try to keep our tags "small" (they already contain a lot of data) 
image

Instead, since LiveComponent requires TwigComponent 2.33 (not released yet), can you rebase and maybe add a new method on `ComponentFactory` to get the components configuration? We use this class in `LiveComponentMetadataFactory`.

WDYT?

Kocal (changes requested):
See previous comment

mbuliard:
@Kocal Thanks ! I figured out a simpler way to find component by name through the LiveComponentMetadataFactory, let me know what you think.

Kocal (approved):

Kocal:
Perfect!

Kocal:
Thank you @mbuliard.

PR #2965, opened 2025-07-30
[LiveComponent] Advanced Polling Features

xDeSwa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

**Added support for poll lifecycle hooks and pollingDirector control in UX LiveComponent**

This PR introduces the following updates to the LiveComponent polling system:

- Added `limit(n)` modifier to limit the number of poll iterations.
- Added new lifecycle events: `poll:started`, `poll:running`, `poll:paused`, `poll:stopped`, and `poll:error`.
- Added the `pollingDirector` API to start, stop, pause, or resume polling via JavaScript.
- Added usage examples using Stimulus controllers.
- Updated documentation with examples and descriptions for all new methods and hooks.

This enhancement introduces support for limiting polling cycles using the `limit` modifier. Once the limit is reached, the polling will automatically stop.

Additionally, new lifecycle hooks (`poll:started`, `poll:running`, `poll:paused`, `poll:stopped`, `poll:error` ) provide full control over the polling process via JavaScript (e.g. Stimulus controllers).

Developers can now track the polling `count` value to build dynamic UI components such as counters, countdown timers, or progress bars.

**Preview:**

https://github.com/user-attachments/assets/0602cb70-1fe1-4fe2-8956-e2efa1808dbc

**Basic usage**

By default, polling runs the ``$render`` action every 2000 milliseconds (2 seconds).

``` html+twig

    {# Unlimited polling every 2 seconds (default) using $render #}
    <div {{ attributes }} data-poll>...</div>

    {# Poll every 5 seconds, up to 20 times, using $render #}
    <div {{ attributes }} data-poll="delay(5000)|limit(20)|$render">...</div>

    {# Poll a custom action (savePost) every 3 seconds, up to 10 times #}
    <div {{ attributes }} data-poll="delay(3000)|limit(10)|savePost">...</div>
```

**Available Modifiers**

- ``delay(ms)`` — The delay between polls in milliseconds (default: ``2000``)
- ``limit(n)`` — Maximum number of times to run the poll (default: unlimited)
- ``actionName`` — The component action to call (default: ``$render``)

**Poll Hooks**

The component emits lifecycle hooks during polling. You can listen to these using JavaScript, for example:

``` javascript

    // controllers/poll_controller.js
    import { Controller } from '@hotwired/stimulus';
    import { getComponent } from '@symfony/ux-live-component';

    export default class extends Controller {
[...]

PR #3365, opened 2026-03-07
[LiveComponent] Enhance polling with limit, visible modifiers and dynamic PHP control

xDeSwa:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| Issues         | -
| License        | MIT

### ✨ Description

This PR introduces lightweight yet powerful controls for the `data-poll` feature to prevent over-fetching, save server resources, and give developers fine-grained control over polling loops. 

While I noticed that PR #2965 proposes a very comprehensive and detailed polling overhaul, I believe this PR covers the most immediate and highly requested polling needs in a simpler, highly stable, and backward-compatible way. (I am completely open to adapting this approach, changing modifier names, or merging efforts based on your feedback!).

### ✨ New Features

* **`limit(n)` Modifier:** Stops the polling loop after exactly `n` successful requests. The counter is securely stored in memory across component re-renders, preventing the limit from infinitely resetting itself.
* **`visible` & `visible(page)` Modifiers:** *Inspired by the visibility discussions in PR #2898.*
    * `visible` (or `visible(component)`): Uses `IntersectionObserver` to automatically pause polling when the component leaves the viewport.
    * `visible(page)`: Pauses polling only when the browser tab is backgrounded/hidden.
    * *Note:* Resuming from a paused state does NOT reset the `limit` counter; it resumes exactly where it left off. Default behavior (without these modifiers) remains unchanged for strict BC.
* **Server-Side Control (`stopPoll`):** Added `stopPoll(string $actionName = '$render')` and `stopAllPolls()` to `ComponentToolsTrait`. This allows developers to stop frontend polling dynamically from PHP. For example, stopping a progress bar poll precisely when a long-running promotional product import finishes.

### 🛠️ Architecture & Stability Improvements

* **Memory Leak Protection:** Added a proper `destroy()` cleanup method to `PollingDirector` that removes global event listeners and disconnects observers when the LiveComponent is disconnected.
* **Over-fetching Prevention:** Shifted from `setInterval` to a safe recursive pattern to prevent overlapping requests (DDoS effect) if the server responds slower than the defined polling delay.

Looking forward to your thoughts and suggestions!

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -22,6 +22,8 @@
     }
     ```
 
+- Add `limit` and `visible` modifiers to `data-poll`, alongside new `stopPoll()` and `stopAllPolls()` methods
+
 ## 2.31
 
 - Add browser events assertions in `InteractsWithLiveComponents`:
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1961,27 +1961,70 @@ var PageUnloadingPlugin_default = class {
 var PollingDirector_default = class {
 	constructor(component) {
 		this.isPollingActive = true;
+		this.polls = [];
 		this.pollingIntervals = [];
+		this.isPageVisible = true;
+		this.isComponentIntersecting = true;
+		this.memoryCounts = {};
 		this.component = component;
+		this.isPageVisible = !document.hidden;
+		this.visibilityChangeListener = () => {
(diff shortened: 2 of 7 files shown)


94noni:
nice one 👍🏻
hope this one or the other make advancement on this topic :)

Kocal:

Thanks!

Status: Needs Work

mmarton:
I need this, feature, can i help somehow?

PR #3367, opened 2026-03-08
[LiveComponent] Allow empty values to bypass model validation modifiers

### LINKED ISSUE #3153: [LiveComponent] min_length should no apply any constraint on 0 lenght
I understand this might be controversial.
But, when `min_lenght` modifier is applied to an input field

```html
<input type="search" data-model="min_lenght(3)|query" />
```

As expected, the re-rendering won't fire until the value is 3 characters long. But, If one resets the field to 0 length, my expectation is that a re-render is fired to take the state of the page/component to the initial state. 
Instead, nothing happens because min_lenght prevents that. That is particularly true if the "X" on the field of type search is not hidden.

Image

If a user clicked on the "X", I think the expectation is to reset to the initial state, especially for a search field. 

Would make more sense to apply the `min_length` constrain on anything above 0, or to give another parameter which allows to exclude 0 from it?

I haven't tried to see if using some workaround, like a live action or a onUpdated handler one can force the component to re-render if the value is changed to an empty string. I guess it can be done, but I wonder if this should not be a baked in the behaviour of the modifier itself.

All of the above, becomes even more complicated if this input field is a child component that is linked to the parent component through data-model `props:props`.

xDeSwa:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   |no 
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3153
| License        | MIT

**The Problem:**
Currently, when using input validation modifiers like `min_length(3)` on a LiveComponent model (e.g., `<input type="search" data-model="min_length(3)|query" />`), the modifier acts as a strict gatekeeper. If a user types "abc", the model updates correctly. However, if the user clears the input (e.g., by clicking the native "X" clear button on a search input or deleting the text), the update is blocked because the length is `0`. 

This creates a frustrating UX: the input appears empty on the screen, but the component fails to re-render or notify the server, leaving the page (and backend state) stuck with the previous search results.

**The Solution:**
This PR updates `LiveController` to bypass `min_length`, `max_length`, `min_value`, and `max_value` checks if the input value is completely empty (`''`, `null`, or `undefined`). 

**Why this is the correct approach (The HTML5 Standard):**
This change aligns LiveComponent's frontend validation strictly with standard HTML5 form validation behavior. In HTML5, attributes like `minlength` and `min` do **not** apply to empty values. If a field is allowed to be empty, it passes validation. If a field *must not* be empty, developers are expected to use the `required` attribute.

By allowing empty strings to pass through these modifiers:
1. It restores the expected behavior for search/filter inputs, allowing the component to naturally reset to its initial/unfiltered state when cleared.
2. It solves complex state synchronization issues between parent/child components when passing models via `props:props`.
3. It prevents numeric inputs (`min_value`) from casting an empty string to `0` and incorrectly failing validation.

### **Demo**
firefox_uET9s3ozhL

### DIFF
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -435,8 +435,11 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
 
         const finalValue = getValueFromElement(element, this.component.valueStore);
 
+        const finalValueIsEmpty = finalValue === '' || finalValue === null || finalValue === undefined;
+
         if (isTextualInputElement(element) || isTextareaElement(element)) {
             if (
+                !finalValueIsEmpty &&
                 modelBinding.minLength !== null &&
                 typeof finalValue === 'string' &&
                 finalValue.length < modelBinding.minLength
@@ -445,6 +448,7 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
             }
 
             if (
+                !finalValueIsEmpty &&
                 modelBinding.maxLength !== null &&
                 typeof finalValue === 'string' &&
                 finalValue.length > modelBinding.maxLength
@@ -454,14 +458,18 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
         }
 
         if (isNumericalInputElement(element)) {
-            const numericValue = Number(finalValue);
+            // An empty input converts to 0 via Number(""), which could incorrectly trigger min_value rules.
+            // Therefore, we bypass validation for numeric fields if they are empty.
+            if (!finalValueIsEmpty) {
+                const numericValue = Number(finalValue);
 
-            if (modelBinding.minValue !== null && numericValue < modelBinding.minValue) {
-                return;
-            }
+                if (modelBinding.minValue !== null && numericValue < modelBinding.minValue) {
+                    return;
+                }
 
-            if (modelBinding.maxValue !== null && numericValue > modelBinding.maxValue) {
-                return;
+                if (modelBinding.maxValue !== null && numericValue > modelBinding.maxValue) {
+                    return;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -435,8 +435,12 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
 
         const finalValue = getValueFromElement(element, this.component.valueStore);
 
+        // Check if the value has been completely cleared by the user
+        const isEmpty = finalValue === '' || finalValue === null || finalValue === undefined;
```
Kocal:
```suggestion
        const finalValueIsEmpty = finalValue === '' || finalValue === null || finalValue === undefined;
```

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -435,8 +435,12 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
...
+
         if (isTextualInputElement(element) || isTextareaElement(element)) {
             if (
+                !isEmpty && // Only check min_length if the value is not empty
```
Kocal: The comment does not bring any value

src/LiveComponent/assets/src/live_controller.ts
```diff
@@ -445,6 +449,7 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
             }
 
             if (
+                !isEmpty && // Only check max_length if the value is not empty
```
Kocal: The comment does not bring any value

Kocal: Also, is this check really useful on `maxLength`?

Kocal (changes requested):

src/LiveComponent/assets/src/live_controller.ts
xDeSwa: That's a good point. You are technically correct that an empty string would never fail a `maxLength` validation. The reason for applying the `!finalValueIsEmpty` check to maxLength as well is for consistency and adherence to standards.
The core principle of this PR is that modifiers like `minLength` and `maxLength` should not act on empty values, perfectly mirroring the native **_HTML5_** validation behavior. By applying this rule universally, we create a single, predictable behavior for all validation modifiers, which simplifies the logic and makes it easier for developers to understand. So, while it's less of a functional fix for `maxLength`, it's a crucial part of making the component's validation logic consistent and standards-compliant.

Kocal (approved):

Kocal:
Thank you @xDeSwa.

PR #3372, opened 2026-03-11
[LiveComponent][TwigComponent] Fix reflection issues for private properties from trait and parent class

kachnitel:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | 
| License        | MIT

## Problem

When a component class *extends* another class that uses a trait with `#[ExposeInTemplate]` or `#[LiveProp]` attributes on **private** properties, those attributes are invisible to the metadata scanners.

PHP's `ReflectionClass::getProperties()` called on a child class does not return private properties from traits used in *ancestor* classes — they are only returned when `getProperties()` is called on the exact class that declares `use TraitName`. This is standard PHP reflection behaviour.

This affects any inheritance pattern where a base component uses `ComponentWithFormTrait` (or any custom trait with private `#[ExposeInTemplate]` properties) and app-level components extend it without re-declaring the trait.

**Symptom in `ux-twig-component`:** `Variable "form" does not exist` in the child component's Twig template.

**Symptom in `ux-live-component`:** `#[LiveProp(fieldName: 'callable()')]` declared in a trait used by a parent class is not registered with the correct `fieldName` on the child component, causing frontend model key mismatches.

### `ux-twig-component` — `ComponentProperties::loadClassMetadata()`

Replace the single `$refClass->getProperties()` call with a loop that walks up `getParentClass()`, calling `getProperties()` at each level. Results are deduplicated by property name (child-class declaration takes priority), giving a complete view of all `#[ExposeInTemplate]` properties across the full hierarchy.

### `ux-live-component` — `LiveComponentMetadataFactory::createPropMetadatas()`

Pass `$property->getDeclaringClass()->getName()` instead of `$class->getName()` to `createLivePropMetadata()`. When a property is declared in a trait used by a parent class, the type extractor must be given the declaring class — not the leaf child class — to resolve the type correctly.

`#[LiveAction]` and `#[LiveListener]` on parent class methods are **not** affected: `getMethods()` already walks the full inheritance chain, so those are discovered correctly on child components.

## Tests

Added `@group trait-inheritance` tests in both packages using a minimal fixture: a trait with a private `#[ExposeInTemplate]` property / `#[LiveProp(fieldName: callable)]`, a parent component that uses the trait, and a child component that extends the parent without re-declaring the trait.

### DIFF
--- a/src/LiveComponent/tests/Integration/LivePropInheritanceTest.php
+++ b/src/LiveComponent/tests/Integration/LivePropInheritanceTest.php
@@ -1,0 +1,146 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\LiveComponent\Tests\Integration;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Metadata\LegacyLivePropMetadata;
+use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\LiveComponent\Metadata\LivePropMetadata;
+use Symfony\UX\LiveComponent\Tests\Fixtures\Component\LivePropInheritanceChild;
+use Symfony\UX\LiveComponent\Tests\Fixtures\Component\LivePropInheritanceParent;
+
+/**
+ * Regression tests for the PHP reflection gap that caused #[LiveProp] attributes
+ * declared in traits used by a *parent* class to be registered without their
+ * fieldName callable on a *child* component, and #[LiveAction]/#[LiveListener]
+ * methods from parent classes to be undiscoverable on child components.
+ *
+ * @group trait-inheritance
+ */
+final class LivePropInheritanceTest extends KernelTestCase
+{
+    /**
+     * Sanity check: the parent component itself must have the LiveProp with
+     * the correct fieldName callable.
+     */
+    public function testParentComponentHasLivePropWithFieldName()
+    {
+        self::bootKernel();
+
+        $prop = $this->findPropMetadata(LivePropInheritanceParent::class, 'formValues');
+
+        $this->assertNotNull($prop, 'formValues LiveProp must be registered on the parent component.');
+
+        $component = new LivePropInheritanceParent();
+        $this->assertSame(
+            'live_prop_inheritance_form',
+            $prop->calculateFieldName($component, 'formValues'),
+            'fieldName callable must resolve correctly on the parent component.'
+        );
+    }
+
+    /**
+     * Regression test for Bug #2: the child component must have the same
+     * LiveProp with the correct fieldName callable, even though it does not
+     * redeclare the trait.
+     *
+     * Before the fix the LiveProp was either not registered at all for the child,
+     * or registered without the fieldName, causing the frontend key mismatch.
+     */
+    public function testChildComponentInheritsLivePropWithFieldName()
+    {
+        self::bootKernel();
+
+        $prop = $this->findPropMetadata(LivePropInheritanceChild::class, 'formValues');
+
+        $this->assertNotNull(
+            $prop,
+            'formValues LiveProp must be registered on the child component.'
+        );
+
+        $component = new LivePropInheritanceChild();
+        $this->assertSame(
+            'live_prop_inheritance_form',
+            $prop->calculateFieldName($component, 'formValues'),
+            'The fieldName callable must be preserved on the child component when '.
+            '#[LiveProp] is declared in a trait used by the parent class.'
+        );
+    }
+
+    /**
+     * Sanity check: the save() action must be allowed on the parent component.
+     */
+    public function testParentComponentHasSaveAction()
+    {
+        self::bootKernel();
+
+        $this->assertTrue(
+            AsLiveComponent::isActionAllowed(LivePropInheritanceParent::class, 'save'),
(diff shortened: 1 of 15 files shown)


kachnitel:
## Example
```php
// A trait that exposes a private property (e.g. ComponentWithFormTrait)
trait HasFormTrait
{
    #[ExposeInTemplate(name: 'form', getter: 'getFormView()')]
    private ?FormView $formView = null;

    #[LiveProp(writable: true, fieldName: 'getFormName()')]
    public array $formValues = [];
}

// A base component that uses the trait directly — works fine before and after
#[AsLiveComponent]
class AdminEntityForm extends AbstractController
{
    use DefaultActionTrait;
    use HasFormTrait;
}

// A child component that extends the base — broken before this fix
#[AsLiveComponent]
class PurchaseOrder extends AdminEntityForm
{
    // Does NOT re-declare: use HasFormTrait;
    // Before: Variable "form" does not exist in PurchaseOrder template
    // Before: formValues LiveProp registered without fieldName callable,
    //         causing "Invalid model name purchase_order_form.status" in JS
    // After:  both work correctly
}
```

src/LiveComponent/tests/Integration/LivePropInheritanceTest.php
```diff
@@ -0,0 +1,158 @@
...
+ * methods from parent classes to be undiscoverable on child components.
+ *
+ * @group trait-inheritance
+ */
+final class LivePropInheritanceTest extends KernelTestCase
+{
+    // -------------------------------------------------------------------------
+    // Bug #2 - #[LiveProp] with fieldName callable on a trait in parent class
+    // -------------------------------------------------------------------------
+
```
Kocal: I don't think we have such comments in Symfony codebase, could you please merge it with PHPDoc from `test*()` methods please? It's fine if it's duplicated.

src/LiveComponent/tests/Integration/LivePropInheritanceTest.php
```diff
@@ -0,0 +1,158 @@
...
+            $prop->calculateFieldName($component, 'formValues'),
+            'The fieldName callable must be preserved on the child component when '.
+            '#[LiveProp] is declared in a trait used by the parent class.'
+        );
+    }
+
+    // -------------------------------------------------------------------------
+    // Bug #3 - #[LiveAction] / #[LiveListener] on parent class method
+    // -------------------------------------------------------------------------
+
```
Kocal: Same here

src/LiveComponent/tests/Integration/LivePropInheritanceTest.php
```diff
@@ -0,0 +1,158 @@
...
+            $listeners,
+            'The save listener must be registered on the child component when '.
+            '#[LiveListener] is declared on the parent class method.'
+        );
+    }
+
+    // -------------------------------------------------------------------------
+    // Helpers
+    // -------------------------------------------------------------------------
+
```
Kocal: Can be removed

src/TwigComponent/src/ComponentProperties.php
```diff
@@ -112,19 +112,38 @@ private function loadClassMetadata(string $class): array
...
+        //
+        // PHP's ReflectionClass::getProperties() called on a child class does
+        // not return private properties from traits used in ancestor classes —
+        // those are only visible when getProperties() is called on the exact
+        // class that declares "use TraitName".  Iterating up via getParentClass()
+        // and deduplicating by property name gives us the complete picture.
+        $seenPropertyNames = [];
+        $currentClass = $refClass;
+        while (false !== $currentClass) {
+            foreach ($currentClass->getProperties() as $property) {
```
Kocal, line 126: Since the issue is about private properties on parent classes, can't we do something like:
- call `getProperties()` on component class // like before
- call `getProperties(\ReflectionProperty::IS_PRIVATE)` on classes parent hierarchy

?

Kocal (changes requested):
Thank you for this PR! I didn't know about `ReflectionClass::getProperties()` behavior about parent private properties and ci.

Just few minor comments :)

src/TwigComponent/src/ComponentProperties.php
kachnitel: Good question. Thinking about that, I don't seem to have a solid answer as I run into asking whether `#[ExposeInTemplate] private ..` should actually be allowed to override, or exist at all?

If another class needs to re-declare it, it's appropriate that it doesn't see the parent's private method/property/..  So one could make an argument that the template shouldn't see it. I'm not sure what the correct answer is, and TBH I haven't looked into how is the attribute actually consumed, but a I have an odd gut feeling about exposing `private` fields.

Not sure how to describe my thought process on this, but essentially I think since it's a private method, I obviously can't use it in my extending class. So should the template? Somehow I can't validate the use case in my head. And while my app exposed this to me, I keep coming to the conclusion that *no private fields should be exposed*, regardless of inheritance.

Ran it by some Claude and figured I'll have at it. 

> #[ExposeInTemplate] on a private property always requires a public getter. The template accesses the getter's return value, not the raw property — so private storage with a public accessor is a deliberate and valid pattern. This fix ensures that pattern works consistently across the full class hierarchy.

src/LiveComponent/tests/Integration/LivePropInheritanceTest.php
kachnitel: yeah meant to turn it into github comment, thanks!

Kocal (approved):

Kocal:
Thank you @kachnitel.

src/TwigComponent/src/ComponentProperties.php
kachnitel: Sorry about the delay, got a little carried away!

Thinking about the `getProperties` question properly, we could structure it that way and extract the logic into a method I think, but then we'd essentially be doing the same thing as if 

```php
$props = $classes[0];

for (1..(array_length($classes) -1) as $class) {
  $props += $class;
}
```

sort of logic, or am I misunderstanding?

PR #3374, opened 2026-03-12
[Toolkit] Fix Shadcn invisible tooltip wrapper on top element

CODEheures:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix
| License        | MIT

Shadcn tooltip wrapper position is on top left of body with a z-50 and it is invisible.
This prevents any action on the elements below it (ex: click on button under this wrapper is impossible).

This fix prevent the wrapper to be on top of others element and preserves the application's initial behavior.


Kocal:
I'm not sure to understand, do you have a video before/after?

CODEheures:
Yes:

# Without fix
the wrapper is over my button, and I can't click it:
2026-03-12_22h19_30

# With this fix
the wrapper is hidden => the size before open is 0x0 => I can click my button
2026-03-12_22h20_43

Kocal:
I'm sorry but even with the gifs I can't really understand what's happening, can you please use **videos** instead and sand a repository?

Thanks!

Kocal:

Thanks!

Status: Needs Work

Kocal:
This got closed back in May in a sweep of pull requests still targeting `2.x`, which we're no longer maintaining, not because the report was off. It wasn't, and thanks for taking the time to file it.

The click interception you found is fixed on `3.x` as of commit `79b974903a9` (2026-07-05): the tooltip wrapper now carries `pointer-events-none open:pointer-events-auto`, so the invisible box stops swallowing clicks meant for whatever's underneath while it's closed. We went with that instead of your `hidden` plus `open:flex` approach because `display: none` can't be transitioned from, and that would have killed the opacity/scale animation the tooltip still relies on. Both approaches solve the click issue, only one keeps the animation.

PR #3376, opened 2026-03-13
[Toolkit][Shadcn] Fix initial "open" state for AlertDialog and Dialog

CODEheures:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix
| License        | MIT

Initilal open state don't work. It is not managed anywhere.
This PR use open attribute to open modal in stimulus dialog controller on connect function if openValue is set to true


Kocal:
Nice catch! I also modified the AlertDialog component since they are nearly identical.

Kocal (approved):

Kocal:
Thank you @CODEheures.

PR #3381, opened 2026-03-14
[TwigComponent] Fix profiling loaded in production

### LINKED ISSUE #3380: [TwigComponent] Profiler data collector always loaded in production when no explicit profiler config is provided

2.x (since 2.32, commit `bd1019`)

### Description

Since commit `bd10191` (*[TwigComponent] Merge profiler and profiler_collect_components config options*), the profiler data collector and logger listener are always loaded in production when no explicit `profiler` configuration is provided in `twig_component.yaml`.

The root cause is in `TwigComponentExtension::getConfigTreeBuilder()`:
```php
->booleanNode('enabled')->defaultValue('%kernel.debug%')->end()
```

`%kernel.debug%` in `defaultValue()` is a raw PHP string.

As a result, `$config['profiler']['enabled']` is always the string `'%kernel.debug%'`, which is always truthy:
```php
// TwigComponentExtension::load()
if ($config['profiler']['enabled']) { // '%kernel.debug%' → always true
    $loader->load('debug.php');       // always loaded, even in production
}
```

### Expected behavior

`debug.php` (profiler data collector and logger listener) should only be loaded when `kernel.debug` is `true`, i.e. in `dev`/`test` environments.

### Actual behavior

`debug.php` is always loaded, even in production, because the TreeBuilder default `'%kernel.debug%'` is a non-empty string that always evaluates to `true`.

Before commit `bd10191`, the code used `$container->getParameter('kernel.debug')` as the primary guard, which correctly returned a boolean.

## How to reproduce
```php
<?php

require __DIR__.'/vendor/autoload.php';

use Symfony\Component\Config\Definition\Processor;
use Symfony\UX\TwigComponent\DependencyInjection\TwigComponentExtension;

$extension = new TwigComponentExtension();
$processor = new Processor();
$config = $processor->processConfiguration($extension, [[
    'defaults' => [],
    'anonymous_template_directory' => 'components/',
]]);

echo 'Type: '.gettype($config['profiler']['enabled'])."\n";
echo 'Value: '.var_export($config['profiler']['enabled'], true)."\n";
echo 'Bool cast: '.var_export((bool) $config['profiler']['enabled'], true)."\n";

// Output:
// Type: string
// Value: '%kernel.debug%'
// Bool cast: true  <-- ALWAYS truthy, even in production!
```

### Possible solution

Replace `booleanNode('enabled')->defaultValue('%kernel.debug%')` with a tri-state `scalarNode('enabled')->defaultNull()` and resolve `kernel.debug` at runtime in the extension's `load()` method:

- `null` -> not configured -> fallback to `$container->getParameter('kernel.debug')`
- `true` -> explicitly enabled
- `false` -> explicitly disabled

### Additional context

The original performance fix (commit `02e3d78`) correctly used `$container->getParameter('kernel.debug')` as the guard. The refactoring commit `bd10191` removed this guard and replaced it with the unresolvable TreeBuilder default.

VincentChataignier:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3380
| License        | MIT

### Summary

Fix profiler debug services being always loaded in production when no explicit `profiler` config is provided.

The `%kernel.debug%` TreeBuilder default was never resolved (TreeBuilder defaults bypass parameter resolution), resulting in a truthy string that always enabled the profiler.

Replace `booleanNode` + `defaultValue('%kernel.debug%')` with a tri-state `scalarNode` (`null`/`true`/`false`) and resolve `kernel.debug` at runtime in the extension.

### Test 

- Add `testDataCollectorNotLoadedInProductionByDefault`, verifies profiler is NOT loaded in production with default config

### DIFF
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -150,7 +150,7 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
         $container->setAlias('console.command.stimulus_component_debug', 'ux.twig_component.command.debug')
             ->setDeprecated('symfony/ux-twig-component', '2.13', '%alias_id%');
 
-        if ($config['profiler']['enabled']) {
+        if ($this->isConfigEnabled($container, $config['profiler'])) {
             $loader->load('debug.php');
 
             $container->getDefinition('ux.twig_component.data_collector')
--- a/src/TwigComponent/tests/Unit/DependencyInjection/TwigComponentExtensionTest.php
+++ b/src/TwigComponent/tests/Unit/DependencyInjection/TwigComponentExtensionTest.php
@@ -58,6 +58,20 @@ class TwigComponentExtensionTest extends TestCase
         $this->assertFalse($container->getDefinition('ux.twig_component.data_collector')->getArgument(2));
     }
 
+    public function testDataCollectorNotLoadedInProductionByDefault()
+    {
+        $container = $this->createContainer();
+        $container->setParameter('kernel.debug', false);
+        $container->registerExtension(new TwigComponentExtension());
+        $container->loadFromExtension('twig_component', [
+            'defaults' => [],
+            'anonymous_template_directory' => 'components/',
+        ]);
+        $this->compileContainer($container);
+
+        $this->assertFalse($container->hasDefinition('ux.twig_component.data_collector'));
+    }
+
     public function testDataCollectorWithDebugModeCanBeDisabled()
     {
         $container = $this->createContainer();


VincentChataignier:
The failing test is in LiveComponent (LiveComponentHydratorTest.php:155), unrelated to this PR. If you're okay with the changes, could you re-run the failed job ? Thanks!

deluxetom (approved):

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -150,7 +150,9 @@ static function (ChildDefinition $definition, AsTwigComponent $attribute) {
         $container->setAlias('console.command.stimulus_component_debug', 'ux.twig_component.command.debug')
             ->setDeprecated('symfony/ux-twig-component', '2.13', '%alias_id%');
 
-        if ($config['profiler']['enabled']) {
+        if ((null === $config['profiler']['enabled'] && true === $container->getParameter('kernel.debug'))
+            || true === $config['profiler']['enabled']
+        ) {
```
Kocal: Let's use `parent::isConfigEnabled($container, $config['profiler'])` instead, it will resolve `$config['profiler']['enabled']` value by interpreting parameters

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -222,7 +224,13 @@ public function getConfigTreeBuilder(): TreeBuilder
...
                     ->canBeEnabled()
                     ->children()
-                        ->booleanNode('enabled')->defaultValue('%kernel.debug%')->end()
+                        ->scalarNode('enabled')
+                            ->defaultNull()
+                            ->validate()
+                                ->ifTrue(static fn ($v) => null !== $v && !\is_bool($v))
+                                ->thenInvalid('The "enabled" option must be a boolean or null.')
+                            ->end()
+                        ->end()
```
Kocal: Let's revert this part

Kocal (changes requested):
Good catch!

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
VincentChataignier: Thanks, I didn't know about this method! I'll fix that.

VincentChataignier:
@Kocal I have amended the commit and force pushed

Kocal (approved):

Kocal:
Thank you @VincentChataignier.

Issue #2737, opened 2025-05-14
[Map] NPM install does not working due to workspace url type


dannyvw:
In https://github.com/symfony/ux/blob/2.x/src/Map/src/Bridge/Leaflet/assets/package.json#L52 is added "workspace:*" url type added but NPM does not recognise this type. Error: npm error Unsupported URL Type "workspace:": workspace:*

package.json
```
"@symfony/ux-leaflet-map": "file:vendor/symfony/ux-leaflet-map/assets",
```

Works "@symfony/ux-leaflet-map": "2.24"

Kocal:
I was able to reproduce it with the following reproducer:
```shell
sf new sf-ux-2737 
cd sf-ux-2737
npm init
composer require symfony/ux-map symfony/ux-leaflet-map
npm i
```
> npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:*
npm error A complete log of this run can be found in: /Users/kocal/.npm/_logs/2025-05-14T13_23_55_268Z-debug-0.log

I didn't expect npm to "read" these **dev dependencies** since they are not supposed to be installed when installing `@symfony/ux-map`... 🤔 

I believe we can modify the npm packages release process to replace `workspace:*` by the actual version...

Kocal:
Fixed with 2.25.0, but first remove the package so you can re-install it

dannyvw:
@Kocal Not sure if it is fixed. With the new release the problem still exists when using the next reference in package.json `"@symfony/ux-leaflet-map": "file:vendor/symfony/ux-leaflet-map/assets",`. I think this is because in the composer vendor package the workspace option still exists.

When using a version (`"@symfony/ux-leaflet-map": "~2.25.1"`) from npm it works.

Kocal:
My bad I think I read your first message too quickly.

Unfortunately, I don't think `@symfony/ux-leaflet-map` with `file:vendor/symfony/ux-leaflet-map/assets` will ever be compatible with `npm`, unless we find a workaround.

The `workspace:*` things help Yarn (that we use in symfony/ux) to manage our workspaces, dependencies, and the **topological order** when running `yarn workspaces foreach ...`. 

[...]

PR #3389, opened 2026-03-19
[React] Fix internal requirement of React 19 (devDependencies)

### LINKED ISSUE #3313: [React] Add React 19 support
When upgrading a project to React 19, `symfony/ux-react` causes "Invalid hook call" errors due to duplicate React instances

The package declares `react@^18.0` as a peer dependency and bundles React 18 in its nested `node_modules/`. This means `render_controller.js` uses React 18 while project components use React 19

milanqtx:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3313 
| License        | MIT

Update react and react-dom devDependencies from ^18.0 to ^18.0 || ^19.0 to match the already-declared peerDependencies and prevent file: installs from forcing React 18 in user projects that require React 19.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -396,10 +396,10 @@ importers:
         specifier: ^26.1.0
         version: 26.1.0
       react:
-        specifier: ^18.0
+        specifier: ^18.0 || ^19.0
         version: 18.3.1
       react-dom:
-        specifier: ^18.0
+        specifier: ^18.0 || ^19.0
         version: 18.3.1(react@18.3.1)
       tslib:
         specifier: ^2.8.1
--- a/src/React/assets/package.json
+++ b/src/React/assets/package.json
@@ -67,8 +67,8 @@
         "@types/webpack-env": "^1.16",
         "@vitejs/plugin-react": "^4.1.0",
         "jsdom": "^26.1.0",
-        "react": "^18.0",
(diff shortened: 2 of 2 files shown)


deluxetom (approved):

Kocal (approved):

Kocal:
> prevent file: installs from forcing React 18 in user projects that require React 19.

This is not correct, devDependencies from a package are purely ignored when installing a package inside a project, users already have the ability to install React 19 if they want. 
Your change only affect the UX repository test suite.

Kocal (approved):

Kocal:
Thank you @milanqtx.

milanqtx:
@Kocal  You're right that `devDependencies` are ignored for registry packages. However, Symfony installs `@symfony/ux-react` via `file:vendor/symfony/ux-react/assets`, and npm v7+ treats `file:` dependencies differently, it installs their `devDependencies`. This is a known npm bug: https://github.com/npm/cli/issues/6405

This was also already encountered in this repository in 
- https://github.com/symfony/ux/issues/2737


The same mechanism is at play here with `@symfony/ux-react`

Kocal:
Oh that's right, I didn't think about that 😅 
What a stupid package manager..

PR #3395, opened 2026-03-20
[TwigComponent] Support component tag names from expressions and component objects

andreybotanic:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no 
| Documentation? | yes
| Issues         |  
| License        | MIT

## Description

Currently, there are three ways to render a component:

1. HTML-like syntax
2. `{% component %}` tag
3. `component()` function

Each approach has its own limitations:

- The `component()` function **supports dynamic component names**, but **does not allow blocks**.
- The `{% component %}` tag **supports blocks**, but **does not allow dynamic component names**.
- The HTML-like syntax behaves similarly to the `{% component %}` tag.

## Proposal

Add support for **dynamic component names** in the `{% component %}` tag.

This would enable a unified syntax that supports both:

- dynamic component resolution
- block usage

The HTML-like syntax is intentionally left unchanged, as introducing dynamic names there would result in awkward and hard-to-read syntax.

## Example Usage

```twig
{% set componentName = 'MyComponent' %}
{% component (componentName) %}
{% endcomponent %}
```

## Extended Idea

Allow passing a **component instance directly**, instead of its name:

```twig
{% component (myComponent) %}
{% endcomponent %}
```

Where `myComponent` is an object of a class marked with:

```php
#[AsTwigComponent]
```

or

```php
#[AsLiveComponent]
```

Both, component name and component object expression must be wrapped in parenthesis.

## Real-world Use Case

This proposal was inspired by a real scenario in my project:

```twig
{% for tab in this.tabs %}
    {% component (tab) with {loading: 'lazy'} %}
        {% block loadingContent %}
            <div class="d-flex align-items-center w-75 mx-auto">
                <strong role="status">Loading...</strong>
                <div class="spinner-border ms-auto" aria-hidden="true"></div>
            </div>
        {% endblock %}
    {% endcomponent %}
{% endfor %}
```

In this case, `tab` is dynamic, and there is a need to:

- pass props (`loading`)
- define custom blocks (`loadingContent`)

Currently, this is not achievable with existing syntax without compromises.

## Benefits

- Eliminates inconsistency between rendering methods
- Enables more flexible and expressive templates
- Improves DX when working with dynamic UI structures (e.g. tabs, registries, dynamic components)

## Notes

- No changes proposed for HTML-like syntax
- Backward compatibility is preserved


Kocal:

Thanks!

Status: Needs Work

andreybotanic:
@Kocal I updated PR.
Some checks failed: I see conflicting checks for code style. Eather PHP-CS-Fixer or "Test method names should not have a return type" fails. I'm not sure how I should fix both but I think it's not blocking for review

src/TwigComponent/doc/index.rst
```diff
@@ -856,6 +856,26 @@ There is also a non-HTML syntax that can be used:
...
+    {% for i in 1..2 %}
+        {% component (prefix ~ i) %}{% endcomponent %}
+    {% endfor %}
+
+If you pass a variable that contains a component object, its class will be
+used to resolve the component name:
+
+.. code-block:: twig
+
+    {% component componentObject %}{% endcomponent %}
```
Kocal: I would suggest that **only** `{% component (componentObject) %}{% endcomponent %}`, between parenthesis could work as a dynamic name, otherwise it could be mis-interpreted.

Like the examples for mapping with dynamic keys ([docs](https://twig.symfony.com/doc/3.x/templates.html#literals)):
```twig
{% set key = 'name' %}
{key: 'Fabien'} {# {key: 'Fabien'} #}
{(key): 'Fabien'} {# {name: 'Fabien'} #}
```

Kocal (changes requested):
About the jobs failing, you can rebase again, they have been fixed in other PRs

src/TwigComponent/doc/index.rst
andreybotanic: @Kocal  You want any expr to be wrapped in parenthesis? E.g.:  
`{% component (componentsArray[i]) %}{% endcomponent %}` instead of  
 `{% component componentsArray[i] %}{% endcomponent %}`?  

`{% component (someObject.component) %}{% endcomponent %}` instead of  
`{% component someObject.component %}{% endcomponent %}`?  

`{% component (componentNameVariable) %}{% endcomponent %}` instead of  
`{% component componentNameVariable %}{% endcomponent %}`?

Kocal: Yeah please, if it's feasible (I'm not a Twig expert), I _just_ want an explicit distinction between `{% component Alert %}` which already try to render the component `Alert`, and `{% component (componentName) %}` which refers to a dynamic component name

andreybotanic: Now parenthesis are required when passing expressions

Kocal: Thank you! I will review your PR ASAP

src/TwigComponent/doc/index.rst
```diff
@@ -856,6 +856,28 @@ There is also a non-HTML syntax that can be used:
...
+The ``{% component %}`` tag also accepts dynamic expressions, but they must be
+wrapped in parentheses. Without parentheses, the value is treated as the
+literal component name.
+
+.. code-block:: twig
+
+    {% set prefix = 'DynamicNameComponent' %}
+    {% for i in 1..2 %}
+        {% component (prefix ~ i) %}{% endcomponent %}
+    {% endfor %}
```
Kocal, line 869: Please add a block `Since 3.1: ...`

src/TwigComponent/doc/index.rst
```diff
@@ -856,6 +856,28 @@ There is also a non-HTML syntax that can be used:
...
+If you pass a variable that contains a component object, its class will be
+used to resolve the component name, but it must also be wrapped in parentheses:
+
+.. code-block:: twig
+
+    {% component (componentObject) %}{% endcomponent %}
+
+Where ``componentObject`` must be an instance of a class-based component (e.g.
+``#[AsTwigComponent]``/``#[AsLiveComponent]``). Anonymous components cannot be
+resolved from an object expression.
```
Kocal: Mmmmh, let's reduce the PR's scope, I don't think we really need to support this case.

Kocal (changes requested):
Let's reduce the PR scope by only keeping the dynamic component tag name please 🙏🏻

Kocal:
The changelog also needs to be updated, for 3.1 please

src/TwigComponent/doc/index.rst
andreybotanic: I can remove support for component objects but I think it's really a **very cool thing**. It allows you to have nested components of different types and still render them in easy way. My example in PR description was not designed specifically to demonstrate the possibility. Instead, it was taken from my actual project and simply rewritten using it.

seb-jean:
@andreybotanic, I created PR https://github.com/symfony/ux/pull/3699, which adds support for the HTML-like syntax based on your PR. What do you think?

Kocal:
I just rebased the PR, fixed conflicts, applied review comments and removed the `componentObject` instance thing

Kocal (approved):

Kocal:
Thank you @andreybotanic.

PR #3404, opened 2026-03-22
[Native] Fix invalid priority for DevServerListener

zairigimad:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | #3405
| License        | MIT

It fixes the devServerListener for local developement

### DIFF
--- a/src/Native/config/debug.php
+++ b/src/Native/config/debug.php
@@ -21,6 +21,6 @@ return static function (ContainerConfigurator $container): void {
         ->args([
             service('.ux_native.configuration_builder'),
         ])
-        ->tag('kernel.event_listener', ['event' => 'kernel.request', 'method' => 'onKernelRequest'])
+        ->tag('kernel.event_listener', ['event' => 'kernel.request', 'method' => 'onKernelRequest', 'priority' => 33])
     ;
 };
--- a/.github/workflows/app-demo-native.yaml
+++ b/.github/workflows/app-demo-native.yaml
@@ -51,10 +51,6 @@ jobs:
               run: php bin/console cache:clear
               working-directory: apps/demo-native
 
-            - name: Dump native configuration files
-              run: php bin/console ux-native:dump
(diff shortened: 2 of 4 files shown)


src/Native/CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 ## 2.33
 
 - Create the component
+- Fix `DevServerEventListener` priority to run before the Symfony Router
```
Kocal: We do not add changelog entry for bug-fixes

src/Native/config/debug.php
```diff
@@ -21,6 +21,6 @@
         ->args([
             service('.ux_native.configuration_builder'),
         ])
-        ->tag('kernel.event_listener', ['event' => 'kernel.request', 'method' => 'onKernelRequest'])
+        ->tag('kernel.event_listener', ['event' => 'kernel.request', 'method' => 'onKernelRequest', 'priority' => 33])
```
Kocal, line 24: Is it possible to add a functional test to prevent regression?  

If it's not, then let's remove the lines from https://github.com/symfony/ux/blob/2.x/.github/workflows/app-demo-native.yaml#L54-L56 :)

Kocal (changes requested):

src/Native/config/debug.php
zairigimad: sure, it needs a functional test

zairigimad: I removed the dump lines from github action workflow, and added  the router in the test to be sure in tests that router doesn't intercept before our DevServerListener

src/Native/CHANGELOG.md
zairigimad: removed

Kocal (approved):

Kocal:
Thank you @zairigimad.

PR #3416, opened 2026-03-28
[LiveComponent] Fix data model radio checkbox

### LINKED ISSUE #3412: [LiveComponent] `data-model` on `radio` and `checkbox` inside a Twig Component overwrites all value attributes with the current prop value
## Problem

Two related bugs affect `data-model` bindings on Twig Components wrapping `<input>` elements.

### Bug 1 — `value` overwritten for `radio` and `checkbox`

When using `data-model` on a Twig Component that renders a `<input type="radio">` or `<input type="checkbox">`, the PHP rendering pipeline overwrites the `value` attribute of every input in the group with the **current LiveProp value**, instead of preserving each input's individual value.

As a result, the JS sync (`setValueOnElement`) computes `element.value == propValue` as `true` for every input, checking all of them simultaneously on re-render.

**Expected:** `value` is preserved per input; only the matching input(s) get `checked`.  
**Actual:** all inputs get `value="{currentPropValue}"`, making them indistinguishable.

### Bug 2 — `[]` array notation crashes with PropertyAccessor

The documentation recommends using `data-model="selected[]"` for checkbox groups bound to an array prop. When used on a **native `<input>`** this works fine (JS handles `[]` transparently). When used on a **Twig Component**, `DataModelPropsSubscriber` passes the full string `selected[]` to `PropertyAccessorInterface::getValue()`, which throws:

```
Could not parse property path "selected[]". Unexpected token "[" at position 8.
```

The `[]` suffix is a JS convention for array models. The PHP layer should strip it before accessing the property.

---

Neither bug affects native `<input>` tags — only when the input is wrapped in a Twig Component.

---

## Affected types

| Type | JS behavior | PHP injection | Result |
|---|---|---|---|
| `radio` | `checked = element.value == value` | overwrites `value` | **broken** |
| `checkbox` (group with explicit `value`) | `checked = value.some(v => v == element.value)` | overwrites `value` | **broken** |
| `checkbox` (boolean, no explicit `value`) | `checked = value` | injects `value="1"` or `value=""` | **broken** — always checked due to `"1" == true` JS loose equality |
| `file` | returns early, ignores `value` | injects into `value` | no visible effect |
| `text`, `email`, `number`, `textarea`… | `element.value = value` | injects into `value` | correct (intended behavior) |

---

### Input Twig Component (UX Toolkit Shadcn UI)
```html
<input
    class="{{ ('flex h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50 ' ~ attributes.render('class'))|tailwind_merge }}"
    {{ attributes }}
>

```

### Bug 1 — Radio group

```php
#[AsLiveComponent]
class MyComponent
{
    use DefaultActionTrait;

    #[LiveProp(writable: true)]
    public string $choice = 'a';
}
```

```twig
{% for option in ['a', 'b', 'c'] %}
    <twig:Input type="radio" data-model="choice" value="{{ option }}" />
{% endfor %}
```

Rendered HTML: all three radios have `value="a"` instead of `a`, `b`, `c`.

### Bug 1 — Checkbox group

```php
#[LiveProp(writable: true)]
public array $selected = ['paris'];
```

```twig
{% for city in ['paris', 'lyon', 'nantes'] %}
    <twig:Input type="checkbox" data-model="selected" value="{{ city }}" />
{% endfor %}
```

Rendered HTML: all checkboxes get `value="Array"` (PHP array cast to string).

### Bug 1 — Boolean checkbox

```php
#[LiveProp(writable: true)]
public bool $active = false;
```

```twig
<twig:Input type="checkbox" data-model="active" />
```

Rendered HTML: `value=""` (PHP `false` → `""`). JS evaluates `"" == false` → true (loose equality) → checkbox is always rendered checked regardless of prop value.

### Bug 2 — `[]` notation crash

```twig
{% for city in ['paris', 'lyon', 'nantes'] %}
    <twig:Input type="checkbox" data-model="selected[]" value="{{ city }}" />
{% endfor %}
```

Throws immediately on render:
```
Could not parse property path "selected[]". Unexpected token "[" at position 8.
```

---

### Bug 1

`DataModelPropsSubscriber::onPreMount()` fires on `PreMountEvent` for every Twig Component with `data-model`. `ModelBindingParser` resolves `data-model="choice"` as `['child' => 'value', 'parent' => 'choice']`, so the subscriber unconditionally injects the current prop into `$data['value']` before the component template renders:

```php
// DataModelPropsSubscriber.php:68
$data[$childModel] = $this->propertyAccessor->getValue($parentMountedComponent->getComponent(), $parentModel);
```

No check for `type="radio"` or `type="checkbox"`. By the time `{{ attributes }}` runs inside the child component template, `value` is already overwritten.

The JS layer (`live_controller.js:290-315`) **does** handle these types correctly — but since all `element.value` attributes are identical after the PHP corruption, the logic produces wrong results:
- For radios: all get `checked = true`
- For checkbox groups: all comparisons fail (`"Array" == "paris"` → `false`) — nothing gets checked
- For boolean checkboxes: always checked due to `"" == false` loose equality

### Bug 2

`ModelBindingParser::parseBinding()` uses `explode(':', $bindingString)` to split the binding string. For `selected[]`, there is no `:`, so it produces `['child' => 'value', 'parent' => 'selected[]']`. The full string `selected[]` is then passed to `PropertyAccessor::getValue()`, which only understands standard Symfony property paths and throws on `[`.

This does not affect native `<input>` tags because those bypass `PreMountEvent` entirely — the `[]` suffix is handled exclusively by the JS layer.

---

## Suggested fix

Both bugs can be fixed in `DataModelPropsSubscriber::onPreMount()`:

```php
foreach ($bindings as $binding) {
    $childModel = $binding['child'];
    // Fix Bug 2: strip the JS array notation suffix before passing to PropertyAccessor.
    $parentModel = rtrim($binding['parent'], '[]');
    $propValue = $this->propertyAccessor->getValue($parentMountedComponent->getComponent(), $parentModel);

    // Fix Bug 1: for radio/checkbox, compute 'checked' instead of overwriting 'value'.
    if ('value' === $childModel && isset($data['type']) && \in_array($data['type'], ['radio', 'checkbox'], true)) {
        if ('checkbox' === $data['type'] && !isset($data['value'])) {
            // Boolean checkbox: no explicit value, prop drives checked directly.
            $data['checked'] = (bool) $propValue;
        } elseif (\is_array($propValue)) {
            // Checkbox group: check if this input's value is included in the array.
            $data['checked'] = \in_array($data['value'] ?? null, $propValue, strict: false);
        } else {
            // Radio or single-value checkbox: strict comparison.
            $data['checked'] = ($data['value'] ?? null) === $propValue;
        }
        continue;
    }

    $data[$childModel] = $propValue;
}
```

This mirrors the existing JS behavior in `setValueOnElement()` (lines 293–301 of `live_controller.js`) and handles all three radio/checkbox use cases consistently.

---

## Environment

- `symfony/ux-live-component`
- PHP 8.x, Symfony 7.x
- Reproducible with any Twig Component that wraps `<input>` and uses `{{ attributes }}`

nullodyssey:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3412
| License        | MIT

## Bug Fix: `data-model` overwrites `value` on radio/checkbox Twig Components

**Issue:** When using `data-model` on a child `<TwigComponent>` that renders a `<input type="radio">` or `<input type="checkbox">`, the subscriber was unconditionally writing the parent prop's value into `$data['value']`. This clobbered the explicit `value="a"` / `value="b"` option values set by the template author, breaking radio groups, checkbox groups, and boolean checkboxes.

**Root cause:** `DataModelPropsSubscriber::onPreMount()` always did `$data[$childModel] = $propValue`. For `data-model="selected"`, the parser resolves `child = 'value'`, so the prop value blindly overwrote whatever `value` the template had already set.

**Fix:** Added three-branch logic in the `foreach` body:

1. **Radio / checkbox group** — when `value` already exists in `$data` (template author passed an explicit option value like `value="b"`), preserve it and instead set `$data['checked']` by comparing the option value against the parent prop. Supports both scalar (radio) and array (multi-checkbox) props.

2. **Boolean checkbox** — when `type="checkbox"` is present in `$data` but no explicit `value` was passed, set `$data['checked']` directly from the boolean prop. Avoids writing a spurious `value` attribute entirely.

3. **All other inputs** — unchanged original behavior (`$data[$childModel] = $propValue`).

**Tests added:** Three integration tests covering radio group, checkbox group, and boolean checkbox scenarios, verifying that `value` attributes are preserved and `checked` is set correctly.

### DIFF
--- a/src/LiveComponent/src/EventListener/DataModelPropsSubscriber.php
+++ b/src/LiveComponent/src/EventListener/DataModelPropsSubscriber.php
@@ -64,8 +64,21 @@ final class DataModelPropsSubscriber implements EventSubscriberInterface
         foreach ($bindings as $binding) {
             $childModel = $binding['child'];
             $parentModel = $binding['parent'];
+            $propValue = $this->propertyAccessor->getValue($parentMountedComponent->getComponent(), $parentModel);
 
-            $data[$childModel] = $this->propertyAccessor->getValue($parentMountedComponent->getComponent(), $parentModel);
+            if ('value' === $childModel && \array_key_exists('value', $data)) {
+                // Radio or checkbox group: an explicit option value was passed (e.g. value="a").
+                // Preserve it and derive the checked state from the parent prop instead.
+                $data['checked'] = \is_array($propValue)
+                    ? \in_array($data['value'], $propValue, false)
+                    : ($data['value'] == $propValue);
+            } elseif ('value' === $childModel && isset($data['type']) && 'checkbox' === $data['type']) {
+                // Boolean checkbox: type="checkbox" declared at call site but no explicit value.
+                // Set checked directly from the boolean prop; do not write a value attribute.
+                $data['checked'] = (bool) $propValue;
+            } else {
+                $data[$childModel] = $propValue;
+            }
         }
 
         $event->setData($data);
--- a/src/LiveComponent/tests/Fixtures/Component/CheckboxInputComponent.php
+++ b/src/LiveComponent/tests/Fixtures/Component/CheckboxInputComponent.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
+
+#[AsTwigComponent('checkbox_input')]
+final class CheckboxInputComponent
+{
+}
--- a/src/LiveComponent/tests/Fixtures/Component/ParentWithBoolCheckboxComponent.php
+++ b/src/LiveComponent/tests/Fixtures/Component/ParentWithBoolCheckboxComponent.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent('parent_with_bool_checkbox')]
+final class ParentWithBoolCheckboxComponent
+{
+    use DefaultActionTrait;
+
+    #[LiveProp(writable: true)]
+    public bool $isActive = true;
+}
--- a/src/LiveComponent/tests/Fixtures/Component/ParentWithCheckboxGroupComponent.php
+++ b/src/LiveComponent/tests/Fixtures/Component/ParentWithCheckboxGroupComponent.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Symfony\UX\LiveComponent\Tests\Fixtures\Component;
+
+use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
+use Symfony\UX\LiveComponent\Attribute\LiveProp;
+use Symfony\UX\LiveComponent\DefaultActionTrait;
+
+#[AsLiveComponent('parent_with_checkbox_group')]
+final class ParentWithCheckboxGroupComponent
+{
+    use DefaultActionTrait;
+
+    #[LiveProp(writable: true)]
+    public array $selected = ['a', 'c'];
+}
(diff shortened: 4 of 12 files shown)


Kocal (changes requested):
Let's discuss in https://github.com/symfony/ux/issues/3412 :)

Kocal:

Thanks!

Status: Needs Work

PR #3422, opened 2026-03-30
[Autocomplete] Add option `clear_on_focus`

### LINKED ISSUE #3415: [AutoComplete] Ajax List Does Not Reset when Cleared
This is a follow-up to #718. 

I'm having the exact same issue as that user. When I use an Ajax search, and I clear the results, it doesn't default back to the default listing order.

Default:
Image

Search Brian:
Image

Select a Brian:
Image

Clear selection:
Image

Only Brian remains:
Image

If I change the search, I get new results via AJAX:

Search Bill:
Image

If I start a start a search and then erase all of the values, it'll load up the original list.

It only does not reset with "Clear" or "Backspace" after selecting an option after a search.

Edit: Black out last names

zairigimad:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3415 
| License        | MIT

It’s not actually a bug, but the default behavior of Tom Select.
When the preload option is set to focus (which is the default), it fetches the data once and then reuses the cache to avoid triggering a request every time the select is opened.

With this PR, a new option is introduced to let you control whether the state should be reset on focus or not.

### DIFF
--- a/src/Autocomplete/assets/src/controller.ts
+++ b/src/Autocomplete/assets/src/controller.ts
@@ -33,6 +33,7 @@ export default class extends Controller {
         minCharacters: Number,
         tomSelectOptions: Object,
         preload: String,
+        resetOnFocus: Boolean,
     };
 
     declare readonly urlValue: string;
@@ -46,6 +47,7 @@ export default class extends Controller {
     declare readonly tomSelectOptionsValue: object;
     declare readonly hasPreloadValue: boolean;
     declare readonly preloadValue: string;
+    declare readonly resetOnFocusValue: boolean;
     tomSelect: TomSelect | undefined;
 
     private mutationObserver: MutationObserver;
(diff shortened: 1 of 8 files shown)


brianfreytag:
Thanks for the clarification on the default functionality and I agree that an additional option here is the best solution.

I do want to clarify that it's not just on "focus" or "change" events, it's when the reset `X` is pressed as well to clear the input. Would it not make sense that the default functionality of "clear" would be to actually reset the select options than maintain the cached version? I guess in my mind, if I'm clearing the input, I'm trying to go back to the original state whether I begin a new search or decide I want to scroll this time. 

This question to ask, I suppose, is, "Should the default value for reset_on_focus be `true` in instances of the "clear" button?" and is this something to take up with the TomSelect folks or a "workaround" on the Symfony side?

src/Autocomplete/assets/src/controller.ts
```diff
@@ -321,6 +323,19 @@ export default class extends Controller {
...
             },
+            onFocus: () => {
+                if (this.resetOnFocusValue && this.tomSelect) {
+                    const query = this.tomSelect.control_input.value.trim();
+                    if (query === '') {
+                        this.tomSelect.clearOptions();
+                        (this.tomSelect as any).loadedSearches = {};
+                        if (typeof (this.tomSelect as any).clearPagination === 'function') {
+                            (this.tomSelect as any).clearPagination();
+                        }
```
Kocal: Please remove the `as any`, it's often a bad practice, and it does not make sense here AFAIK.

Kocal (commented):
Since it's not really a bug but a feature, can you add a new changelog entry for 2.35 please?

Thanks!

src/Autocomplete/assets/src/controller.ts
zairigimad: yes, i just pushed an update for this part

Kocal:
Thank you @zairigimad.

PR #3433, opened 2026-04-04
[Cropper] Always apply rotation in `Crop::getCroppedImage()` and `Crop::getCroppedThumbnail()`

MrYamous:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no
| Documentation? | i'll check if needed
| Issues         | Next part of #2930 
| License        | MIT

I saw some PR about 3.x and tests dedicated to this change were removed, but not yet the code

### DIFF
--- a/src/Cropperjs/CHANGELOG.md
+++ b/src/Cropperjs/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Minimum required Symfony version is now 6.4
 - Minimum required PHP version is now 8.2
+- Always apply rotation in `Crop::getCroppedImage()` and `Crop::getCroppedThumbnail()`
 
 ## 2.34
 
--- a/src/Cropperjs/src/Model/Crop.php
+++ b/src/Cropperjs/src/Model/Crop.php
@@ -55,11 +55,8 @@ class Crop
         $this->filename = $filename;
     }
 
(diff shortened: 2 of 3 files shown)


Kocal (commented):
Oops, nice catch!

Shouldn't we **always** apply the rotation like the deprecation message told us? Meaning, we can remove the `$applyRotation` parameter

MrYamous:
> Oops, nice catch!
> 
> Shouldn't we **always** apply the rotation like the deprecation message told us? Meaning, we can remove the `$applyRotation` parameter

At first look, I didn't like the idea of implicit rotation based on whether or not related options were present.
But in the end, I figure that this parameter doesn't really add much.
I'm going to make the change

src/Cropperjs/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Minimum required Symfony version is now 6.4
 - Minimum required PHP version is now 8.2
+- The `Crop::getCroppedImage()` and `Crop::getCroppedThumbnail()` functions now apply rotation according to provided option
```
Kocal:
```suggestion
- The `Crop::getCroppedImage()` and `Crop::getCroppedThumbnail()` functions now always apply rotation
```

Kocal:
Thank you @MrYamous.

PR #3441, opened 2026-04-07
[Autocomplete] Decouple from Doctrine ORM - support any data source

Prometee:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Issues        | -
| License       | MIT

## Description

This PR decouples the Autocomplete package from Doctrine ORM, allowing **any object source** (API, Elasticsearch, in-memory collections, etc.) to be used as an autocomplete backend while maintaining full backward compatibility.

### New source-agnostic contract
- `AutocompleterInterface` with `fetchResults()` and `isGranted()`
- `OptionsAwareAutocompleterInterface` with `setOptions()`

### New form type infrastructure (no Doctrine ORM required)
- `AutocompleteChoiceType` extending `ChoiceType`
- `#[AsAutocompleteField]` attribute
- `WrappedChoiceTypeAutocompleter`

### New controller and route
- `AutocompleteController` with route `ux_autocomplete` at `/{alias}`
- `ux_entity_autocomplete` is now a deprecated route alias

### Conditional Doctrine ORM registration
- Doctrine services only registered when available via `ContainerBuilder::willBeAvailable()`

### Refactoring (DRY)
- `AutocompleteTypeTrait`, `WrappedAutocompleterTrait`
- `AsEntityAutocompleteField extends AsAutocompleteField`
- Single `processFieldTag()` in compiler pass

### Deprecations
- `EntityAutocompleterInterface`, `OptionsAwareEntityAutocompleterInterface`
- `AsEntityAutocompleteField`, `EntityAutocompleteController`
- Route `ux_entity_autocomplete`

### Maker
- `make:autocomplete-field` now supports Entity and Choice modes

### Documentation
- New sections for non-Doctrine ORM usage with examples

## Backward compatibility
All existing code continues to work without changes.


src/Autocomplete/phpunit.xml.dist
```diff
@@ -11,7 +11,7 @@
     <php>
         <ini name="error_reporting" value="-1"/>
         <env name="SHELL_VERBOSITY" value="-1"/>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0&amp;max[direct]=0"/>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0&amp;max[self]=999999"/>
```
GromNaN: Why did you change this setting?

src/Autocomplete/doc/index.rst
```diff
@@ -99,6 +99,16 @@ have MakerBundle installed, you can run:
...
 
+The command will ask you what type of autocomplete field you want:
+
+- **Entity (Doctrine)**: for autocompleting Doctrine entities (uses ``BaseEntityAutocompleteType``)
+- **Choice (any data source)**: for any other data source (uses ``BaseAutocompleteChoiceType``)
+
+.. versionadded:: 2.24
+
+    The choice between Entity and Choice modes was added in Autocomplete 2.24.
+    If Doctrine is not installed, the command defaults to Choice mode.
```
GromNaN:
```suggestion
    If Doctrine ORM is not installed, the command defaults to Choice mode.
```

src/Autocomplete/doc/index.rst
```diff
@@ -391,6 +401,81 @@ extra option in the default ``query_builder`` of the ``FoodAutocompleteField``::
...
 
+Usage in a Form (Ajax with any data source)
+-------------------------------------------
+
+.. versionadded:: 2.24
+
+    The ``BaseAutocompleteChoiceType`` and ``#[AsAutocompleteField]`` were added
+    in Autocomplete 2.24 to support non-Doctrine data sources.
+
+If your data comes from something other than Doctrine (e.g. an API, Elasticsearch,
```
GromNaN:
```suggestion
If your data comes from something other than Doctrine ORM (e.g. an API, Elasticsearch,
```

src/Autocomplete/src/Form/AutocompleteChoiceType.php
```diff
@@ -0,0 +1,68 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+/**
+ * All choice-based form types that want to expose autocomplete functionality should use this for its getParent().
+ *
+ * This is the non-Doctrine equivalent of BaseEntityAutocompleteType.
+ */
+final class BaseAutocompleteChoiceType extends AbstractType
```
GromNaN: @symfony/ux Is there any point to this `Base` prefix? I usually associate it with abstract classes that are meant to be extended.


```suggestion
final class AutocompleteChoiceType extends AbstractType
```

src/Autocomplete/phpunit.xml.dist
Prometee: The PR adds `trigger_deprecation()` calls in deprecated classes (`AsEntityAutocompleteField`, `EntityAutocompleteController`). Since the existing test fixtures use these deprecated classes, every test that boots the kernel triggers self-deprecation notices (~83 total). Without `max[self]=999999`, the PHPUnit Bridge would fail with exit code 1 on these expected self-deprecations.

src/Autocomplete/src/Form/AutocompleteChoiceType.php
Prometee: @GromNaN The Base prefix was carried over from `BaseEntityAutocompleteType` for consistency, but you're right, since this class is final, the prefix doesn't make sense. I'll rename it to `AutocompleteChoiceType`. Should we also rename `BaseEntityAutocompleteType` to `EntityAutocompleteType` (with a deprecated alias) for consistency, or keep that for a separate PR?

src/Autocomplete/doc/index.rst
Prometee: Applied, thanks!

src/Autocomplete/doc/index.rst
Prometee: Applied, thanks!

src/Autocomplete/phpunit.xml.dist
GromNaN: Tests that trigger the deprecation must be annotated with `@group legacy` (see [PHPUnit Bridge docs](https://symfony.com/doc/current/components/phpunit_bridge.html#mark-tests-as-legacy)).

During normal use, no deprecation notices should be triggered, to ensure that users can use the project without seeing any deprecation notices.

Prometee: Thank you for pointing this out I did not know about that. I'll add `@group legacy` to all test classes that boot the kernel (which registers the deprecated `AsEntityAutocompleteField` fixtures), and restore `phpunit.xml.dist` to `max[direct]=0` without the `max[self]` override.

Kocal:

Thanks!

PS: Please reduce the non-useful changes, like the `: void` return type on `test*` methods, we do not use them in Symfony.

src/Autocomplete/phpunit.xml.dist
GromNaN: Tests marked with the “legacy” group are those that cover deprecation messages; they will be removed in the next major release. We need to find a way to prevent these deprecation warnings from being triggered during normal use.

Prometee:
@Kocal I rebased and adapt, let me know what you need me to modify

Kocal (approved):

Kocal:
Thank you @Prometee.

Issue #2943, opened 2025-07-25
Option to disable `data-model` on a field


FluffyDiscord:
I have form that renders form type from a third party bundle that contains dummy form inputs for selection that are not registered in their form type.

My form uses `data-model="on(change)|*"` to "auto-validate" it against Symfony back-end. Every time the dummy input field changes, live component throws an error that data model with that field name does not exist (obviously).

I need to somehow blacklist that input, either using CSS selector, or by adding attribute to it that stops/ignores it.

```html
<form data-model="on(change)|*">
    <input name="dummy"/> <-- need to somehow exclude it from the change event
</form>
```

xDeSwa:
```twig
<input name="favorite_color" data-live-ignore>
```

https://symfony.com/bundles/ux-live-component/current/index.html#skipping-updating-certain-elements

FluffyDiscord:
@xDeSwa thank you, I somehow missed that in docs

FluffyDiscord:
@xDeSwa just tried it, it does not work

```html
<input type="radio" name="tabs-inquiry" class="hidden" data-live-ignore>
```

this still triggers that thing and throws an error

```
Uncaught Error: Invalid model name "tabs-inquiry".
    at Component.set (live_controller.js:2025:19)
    at extended.updateModelFromElementEvent (live_controller.js:3156:24)
    at extended.handleChangeEvent (live_controller.js:3112:14)
    at HTMLDivElement.callback (live_controller.js:2926:58)
```

xDeSwa:
This example will help you

```php
<?php

namespace App\Twig\Components\Panel\PollTest;

use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\UX\LiveComponent\Attribute\AsLiveComponent;
use Symfony\UX\LiveComponent\Attribute\LiveProp;
use Symfony\UX\LiveComponent\DefaultActionTrait;

#[AsLiveComponent]
class Test8 extends AbstractController
{
    use DefaultActionTrait;

    #[LiveProp(writable: true)]
    public string $meal = 'lunch';


}
```

``` html
<input type="radio" name="meal[]" data-model="meal" value="breakfast" data-live-ignore>
<input type="radio" name="meal[]" data-model="meal" value="lunch" data-live-ignore>
<input type="radio" name="meal[]" data-model="meal" value="dinner" data-live-ignore>
```

FluffyDiscord:
[...]

PR #3448, opened 2026-04-08
[LiveComponent] Prevent automatic model binding for elements with data-live-ignore

xDeSwa:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no 
| Documentation? | no
| Issues         | #2943
| License        | MIT

This PR fixes an issue where elements with a `name` attribute were being automatically bound as models even if they had the `data-live-ignore` attribute, specifically when using form-wide binding like `<form data-model="*">`
The Problem: When a form uses `data-model="*"`, every input with a `name` attribute is automatically treated as a live model. If a UI-only input (e.g., a radio button used for tab switching or a temporary filter) has a name that doesn't exist as a property on the PHP component, it throws an `Invalid model` name error.

This change makes `data-live-ignore` behave more intuitively as a "hands-off" flag for LiveComponent, especially in aggressive binding scenarios like `<form data-model="on(change)|*">`.


Kocal:

Thanks!

PR #3463, opened 2026-04-13
[Toolkit] WIP: [Bootstrap] Add Bootstrap 5

BafS:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| Issues         | no
| License        | MIT

Add bootstrap 5.3 kit with multiple components. There is still more components to do but I think it's worth having feedback before I continue. For disclosure, I used Claude Opus 4.6 to help me.


DcgRG:
Hello, that's great, a new kit!

To start with a quick overview, here are two small points:

- For conditionally applying your classes, I recommend using [html_cva](https://twig.symfony.com/doc/3.x/functions/html_cva.html), which will make the templates more readable.
- For interactive components, I think it might be beneficial to keep the trigger pattern from other kits so you have full control over the elements that trigger interactions.

Bonus and purely personal opinion: for interactive elements like collapse, rather than using data attributes, I think it might be more efficient to use a stimulus controller that then uses a Bootstrap object (e.g., [new bootstrap.Collapse(collapseEl)](https://getbootstrap.com/docs/5.3/components/collapse/#methods)).

BafS:
Thanks for the feedback @DcgRG, I tried to address those

src/Toolkit/kits/bootstrap/collapse/assets/controllers/collapse_controller.js
```diff
@@ -0,0 +1,52 @@
+import { Controller } from '@hotwired/stimulus';
+import { Collapse } from 'bootstrap';
+
+export default class extends Controller {
+    static targets = ['trigger', 'content'];
+
+    connect() {
+        this._collapse = new Collapse(this.contentTarget, {
+            toggle: false,
```
DcgRG, line 13: You can add "parent" option here

DcgRG:
Ultimately, is an Accordion component simply a group of Collapse elements?
You can use your generic collapse element to create your Accordion component; this could save you from creating an additional controller for Accordion.

Nope ?

BafS:
Indeed, the accordion is a map of collapse. I removed the accordion controller. It's just a little bit confusing to have things like `data-action="click->collapse#toggle"` in the accordion I feel, but it's probably a detail.

DcgRG:
> Indeed, the accordion is a map of collapse. I removed the accordion controller. It's just a little bit confusing to have things like `data-action="click->collapse#toggle"` in the accordion I feel, but it's probably a detail.

You can nested components like in this examples with a sub dropdown un main dropdown

https://github.com/symfony/ux/blob/3.x/src/Toolkit/kits/flowbite-4/dropdown/templates/components/Dropdown/Sub.html.twig

DcgRG:
I think you do name your kit with name and version in case an another version is release

`bootstrap-5`

sbolch:
@BafS I was unaware of this PR and started a [separate repo](https://github.com/sbolch/symfony-ux-toolkit-bootstrap) for Bootstrap 5.3 Twig components with more components ready to use and I'm currently actively working on it. I see this PR and your fork haven't been updated since last month. Should we somehow merge our progress?

BafS:
Sorry for the very late answer. It would be good to merge progress, I think the best would be to be able to agree on the convention and to merge the base, like that we could focus on individual components. I don't know if you have a better idea?

Kocal:
Hi everyone, I'm sorry but we merged https://github.com/symfony/ux/pull/3713 instead.

I know it can be frustrating, but Simon had been working on it for months and month, but never had time to open a PR here. 

Thanks anyway for your contributions, and we look forward to seeing them in the new Bootstrap Kit!

smnandre:
I'm really sorry. I did not even checked if someone started one during my long time "away" UX.. I started this a long time ago (more than a year I think), and as @Kocal said I never found time to finish and open the PR ...  but this is in no way a valid reason not to check other PR and collaborate with existing, ongoing contribution. 

Again, very sorry @BafS

BafS:
Don't worry, I'm glad there is something merged! Thanks for your kind words @smnandre and @Kocal.

PR #3465, opened 2026-04-14
[LiveComponent] Prevent LiveCollectionTrait from reusing removed indexes

### LINKED ISSUE #3371: [LiveCollection] LiveCollectionType may reuse removed Doctrine entity when adding a new item due to index reuse
## Description

I'm encountering an issue when using **LiveComponent + LiveCollectionType with a Doctrine Collection**.

The problem seems related to how `$formValues` indexes are handled when removing and adding items.

### Context

Symfony UX LiveComponent version: **2.31**
Symfony version: **7.4.1**
Doctrine/ORM version: **3.5.7**
PHP version: **8.4**

I'm using:

- Symfony UX LiveComponent
- `LiveCollectionType`
- A Doctrine collection inside a form.

### Current behaviour

Suppose the collection initially contains 3 elements.

Indexes in `$formValues` are:

```
0, 1, 2
```

If I remove the last element using the `remove` action from `LiveCollectionTrait`, index `2` is removed and the array  keys become:

```
0, 1
```

Then, when adding a new element via the `add` action, the following code is executed in LiveCollectionTrait -> addCollectionItem (line 36) :

```php
$index = [] !== $data ? max(array_keys($data)) + 1 : 0;
```

In this case:

```
max index = 1
new index = 2
```

So the new item is inserted at index `2`.

### The issue

Because index `2` previously existed and was mapped to a Doctrine entity, **the newly added entry ends up being mapped back to the previously removed entity instead of creating a new one.**

This leads to unexpected behavior where:

- the removed entity appears again
- a new object is not actually created

### Expected behaviour

When adding a new element after a removal, a **new entry should always be created**, without reusing a previously removed reference.

### Possible cause

The issue seems to be that the index used in `$formValues` is derived from the **current max key**, which may correspond to a previously removed entry.

This becomes problematic when working with Doctrine collections and form data mapping.

### Possible solution ideas

The workaround I've found is : 

```php
#[LiveProp]
public int $maxInitialKey = 0;

#[PostMount]
public function postMount(): void
{
    $this->maxInitialKey = max($this->entity->collection->getKeys());
}

#[LiveAction]
public function addCollectionItem(PropertyAccessorInterface $propertyAccessor, #[LiveArg] string $name): void
{
    $key = $this->maxInitialKey + 1;
    if (array_key_exists($key, $this->formValues['collection'])) {
        $key = max(array_keys($this->formValues['collection'])) + 1;
    }

    $this->formValues['collection'][$key] = [];
}
```

### Question

Is this the expected behavior of `LiveCollectionType`, or could this be considered a bug when working with Doctrine collections?

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3371
| License        | MIT

### Problem

`LiveCollectionTrait::addCollectionItem()` picks the next index as:

```php
$index = [] !== $data ? max(array_keys($data)) + 1 : 0;
```

This only looks at the *current* keys. If a user removes the last item and then adds a new one, the new item reuses the index that was just freed. When the form is bound to a Doctrine collection, the form machinery matches the new entry back to the (previously removed) entity at that key, so the removed entity silently reappears and no new object is created.

Repro from the issue:
1. Collection starts with entities at keys `[0, 1, 2]`.
2. `removeCollectionItem` on index `2` → `[0, 1]`.
3. `addCollectionItem` → `max([0,1]) + 1 = 2` → form rebinds to the just-removed entity.

### Fix

Track indexes freed by `removeCollectionItem()` in a `#[LiveProp]` tombstone keyed by collection field name, and include them in the `max()` computation of `addCollectionItem()`. After the scenario above, the next index becomes `3` as expected.

- Added a regression test covering the remove-then-add scenario.
- The new `$liveCollectionRemovedKeys` prop is additive and non-writable; existing serialized state defaults to `[]` for older clients.

### Scope

- Only fixes the integer-index case (which is what `LiveCollectionType` generates). String-keyed collections are already unsupported by the existing `max(array_keys(...))` logic and are out of scope.

Thanks for the Live Component work — it's been a pleasure to dig into. 🙏


src/LiveComponent/src/LiveCollectionTrait.php
```diff
@@ -22,6 +23,18 @@ trait LiveCollectionTrait
...
+     * Tracks indexes removed from each collection so that newly added items
+     * never reuse a previously used index, which would make the form re-bind
+     * to a stale entity from the initial collection.
+     *
+     * Keyed by the collection field name passed to the add/remove actions.
+     *
+     * @var array<string, list<int>>
+     */
+    #[LiveProp]
+    public array $liveCollectionRemovedKeys = [];
```
JacquesMougin, line 36: @Amoifr Do you think this could be "private" instead of "public"? I can’t think of a case where it would be useful outside this context

Amoifr: Good catch — I'd love to make it `private`, but Symfony's PropertyAccess (used by `LiveComponentHydrator`) can't read or write private properties without an accessor, so the dehydration/rehydration cycle would silently drop the tombstone. I verified it with a tiny repro:

```php
class T {
    #[LiveProp] private array $privateArray = ['init' => 1];
    #[LiveProp] public  array $publicArray  = ['init' => 1];
}

$pa = PropertyAccess::createPropertyAccessor();
$pa->getValue(new T(), 'privateArray'); // ❌ "Can't get a way to read the property "privateArray"…"
$pa->getValue(new T(), 'publicArray');  // ✅
```

Every existing `#[LiveProp]` in the codebase (`ComponentWithFormTrait::$formName`, `$formValues`, `$isValidated`, `$validatedFields`, `ValidatableComponentTrait::$isValidated`, …) is `public` for the same reason. I'll keep `public` here to match that convention — let me know if you'd rather I add a private + getter/setter pair instead, happy to rework.

JacquesMougin: @Amoifr Thanks a lot for the explanation, I didn’t dig deep enough to understand the full context !

Amoifr: @JacquesMougin you're welcome ! I get to much pain in the first hours of the live component with this kind of things to remember about that :smile:

PR #3466, opened 2026-04-14
[Toolkit][Shadcn] Add native-select, slider, scroll-area and hover-card recipes

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Part of #3233
| License        | MIT

Hi 👋

Second batch of recipes from the unchecked list in #3233 (see #3464 for the first batch). All four recipes are JS-free and built on native HTML elements where possible:

- **native-select** — styled native `<select>`; `optgroup`/`option` are passed through as the block content.
- **slider** — styled native `<input type="range">` using `accent-primary`.
- **scroll-area** — scrollable container that keeps browser-native scrolling and styles the WebKit scrollbar track/thumb with Tailwind.
- **hover-card** — pure-CSS popup using `group/hover-card` + `group-hover` / `group-focus-within`, with `HoverCard:Trigger` and `HoverCard:Content` sub-components. Focus-within gives keyboard users the same reveal as hover.

Each recipe ships with `Usage.html.twig`, `Demo.html.twig`, and passing snapshot tests (`256 tests, 512 assertions`).

Happy to split into four PRs if that's easier to review — each recipe is a single commit.

Thanks again for the Toolkit work! 🙏

Issue #3233, opened 2025-12-13
[Toolkit][Shadcn] Component Tracking Metaissue


daFish:
Tracking Issue for Toolkit components from the shadcn/ui kit:

- [x] Accordion (https://github.com/symfony/ux/pull/3292)
- [x] Alert Dialog
- [x] Alert
- [x] Aspect Ratio
- [x] Avatar
- [x] Badge
- [x] Breadcrumb
- [x] Button Group
- [x] Button
- [ ] Calendar
- [x] Card
- [ ] Carousel
- [ ] Chart
- [x] Checkbox
- [x] Collapsible
- [ ] Combobox
- [ ] Command
- [ ] Context Menu
- [ ] Data Table
- [ ] Date Picker
- [x] Dialog
- [ ] Drawer
- [ ] Dropdown Menu
- [x] Empty
- [x] Field
- [ ] Form
- [ ] Hover Card
- [x] Input Group
- [ ] Input OTP
- [x] Input
- [x] Item
- [x] Kbd
- [x] Label
- [ ] Menubar
- [ ] Native Select
- [ ] Navigation Menu
- [x] Pagination
- [ ] Popover
- [x] Progess
- [x] Radio Group
- [ ] Resizable
- [ ] Scroll Area
- [x] Select
- [x] Separator
- [ ] Sheet
- [ ] Sidebar
- [x] Skeleton
- [ ] Slider
- [ ] Sonner
- [x] Spinner
- [x] Switch
- [x] Table
- [x] Tabs
- [x] Textarea
- [ ] Toast
- [x] Toggle Group
- [ ] Toggle
- [x] Tooltip https://github.com/symfony/ux/pull/3321
- [x] Typography

List is based on https://ui.shadcn.com/docs/components

Edgarruiz856:
edgarplasticos3@hotmail.com

benjaminroche4:
Do you know when is the next release ? Especially for dropdown ?

Kocal:
There is an open PR for Dropdown, see https://github.com/symfony/ux/pull/3232

Amoifr:
Hey @Kocal 👋 I picked a few of the unchecked items and started with `radio-group` → #3464. I have three more ready locally (`collapsible`, `typography`, `toggle-group`) that I'll send as separate PRs once I know the shape is right.

Two quick questions I'd love your take on before I push the rest:

[...]

PR #3478, opened 2026-04-18
[Toolkit][Shadcn] Add hover-card recipe

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| License        | MIT

Adds the `hover-card` recipe to the Shadcn kit.

Part of the split of #3466 into one PR per component, tracking #3233.

Snapshots will be regenerated after rework.

### DIFF
--- a/src/Toolkit/CHANGELOG.md
+++ b/src/Toolkit/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [Flowbite] Add `avatar` recipe
 - [Flowbite] Add `dropdown` recipe
+- [Shadcn] Add `hover-card` recipe
 - [Shadcn] Add `resizable` recipe
 
 ## 3.0.0
--- a/src/Toolkit/kits/shadcn/hover-card/examples/Demo.html.twig
+++ b/src/Toolkit/kits/shadcn/hover-card/examples/Demo.html.twig
@@ -1,0 +1,10 @@
+<twig:HoverCard>
+    <twig:HoverCard:Trigger>
+        <twig:Button variant="link" {{ ...hover_card_trigger_attrs }}>Hover Here</twig:Button>
+    </twig:HoverCard:Trigger>
+    <twig:HoverCard:Content class="w-64 flex flex-col gap-0.5">
+        <div class="font-semibold">@symfony</div>
+        <div>The PHP framework for web applications — created by @fabpot.</div>
+        <div class="mt-1 text-xs text-muted-foreground">Joined October 2010</div>
+    </twig:HoverCard:Content>
+</twig:HoverCard>
--- a/src/Toolkit/kits/shadcn/hover-card/examples/Usage.html.twig
+++ b/src/Toolkit/kits/shadcn/hover-card/examples/Usage.html.twig
@@ -1,0 +1,8 @@
+<twig:HoverCard>
+    <twig:HoverCard:Trigger>
+        <a href="#" class="underline" {{ ...hover_card_trigger_attrs }}>@symfony</a>
+    </twig:HoverCard:Trigger>
+    <twig:HoverCard:Content>
+        The Symfony PHP framework — official organization on GitHub.
+    </twig:HoverCard:Content>
+</twig:HoverCard>
--- a/src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Content.html.twig
+++ b/src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Content.html.twig
@@ -1,0 +1,10 @@
+{# @block content The content revealed on hover #}
+<span
+    role="tooltip"
+    class="{{ ('invisible opacity-0 in-data-[state=open]:visible in-data-[state=open]:opacity-100 transition-opacity absolute left-1/2 top-full z-50 mt-2 w-64 -translate-x-1/2 rounded-md border bg-popover p-4 text-sm text-popover-foreground shadow-md outline-none ' ~ attributes.render('class'))|tailwind_merge }}"
+    {{ attributes.defaults({
+        'data-slot': 'hover-card-content',
+    }) }}
+>
+    {%- block content %}{% endblock -%}
+</span>
--- a/src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Trigger.html.twig
+++ b/src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Trigger.html.twig
@@ -1,0 +1,7 @@
+{# @block content The element that reveals the hover card on hover or focus #}
+{%- set hover_card_trigger_attrs = {
+    'data-slot': 'hover-card-trigger',
+    tabindex: 0,
+    'data-action': 'focus->hover-card#show blur->hover-card#hide'|html_attr_type('sst'),
+} -%}
+{%- block content %}{% endblock -%}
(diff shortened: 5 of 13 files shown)


Amoifr:
Main exemple 

https://github.com/user-attachments/assets/680de98c-44fb-4012-9f5a-83f4ebbd474e

Amoifr:
Slides example 


https://github.com/user-attachments/assets/733203e2-969b-498c-b54f-a7cb5e9e814c

Amoifr:
RTL example 


https://github.com/user-attachments/assets/a60ddf4c-e1a5-4367-bae2-a9ce34ca7ea6

src/Toolkit/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - [Shadcn] Add `collapsible` recipe
 - [Shadcn] Add `typography` recipe
 - [Shadcn] Add `toggle-group` recipe
+- [Shadcn] Add `hover-card` recipe
```
Kocal: For 3.1

src/Toolkit/kits/shadcn/hover-card/examples/Demo.html.twig
```diff
@@ -0,0 +1,10 @@
+<twig:HoverCard>
+    <twig:HoverCard:Trigger>
+        <twig:Button variant="link">Hover Here</twig:Button>
+    </twig:HoverCard:Trigger>
+    <twig:HoverCard:Content class="w-64 flex flex-col gap-0.5">
+        <div class="font-semibold">@nextjs</div>
+        <div>The React Framework — created and maintained by @vercel.</div>
+        <div class="mt-1 text-xs text-muted-foreground">Joined December 2021</div>
```
Kocal: Can we make it more Symfony? 🙏🏻

src/Toolkit/kits/shadcn/hover-card/examples/Usage.html.twig
```diff
@@ -0,0 +1,6 @@
+<twig:HoverCard>
+    <twig:HoverCard:Trigger class="underline">@symfony</twig:HoverCard:Trigger>
+    <twig:HoverCard:Content>
+        The Symfony framework — official account on GitHub.
```
Kocal: Technically, @symfony is not a GitHub account 🤓

Kocal (approved):
with some comments

src/Toolkit/CHANGELOG.md
Amoifr: Done — moved to 3.1 (and applied the same fix across my 10 other open Shadcn recipe PRs to save you the round-trip 😄).

src/Toolkit/kits/shadcn/hover-card/examples/Demo.html.twig
Amoifr: Done — swapped Next.js/Vercel for Symfony/@fabpot. 🎉

src/Toolkit/kits/shadcn/hover-card/examples/Usage.html.twig
Amoifr: Good catch — rephrased as "official organization on GitHub". 🤓

src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Trigger.html.twig
Kocal, line 1: I just noticed, but the comment https://github.com/symfony/ux/pull/3487#discussion_r3129054070 also applies here, `*Trigger.html.twig` files only expose a Twig variable and a `block content`

Can you rework this part please?
Thanks!

Kocal (changes requested):

src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Trigger.html.twig
Amoifr: Done in 71857370 — the Trigger now only exposes a `_trigger_attrs` variable and a `block content`, letting the consumer pick their own element (button, link, span…). Also updated the examples (Demo, Usage, Sides, RTL) and regenerated the snapshots accordingly.

Thanks for the nudge! 🙏

src/Toolkit/kits/shadcn/hover-card/templates/components/HoverCard/Content.html.twig
```diff
@@ -0,0 +1,10 @@
+{# @block content The content revealed on hover #}
+<span
+    role="tooltip"
+    class="{{ ('invisible opacity-0 group-hover/hover-card:visible group-hover/hover-card:opacity-100 group-focus-within/hover-card:visible group-focus-within/hover-card:opacity-100 transition-opacity absolute left-1/2 top-full z-50 mt-2 w-64 -translate-x-1/2 rounded-md border bg-popover p-4 text-sm text-popover-foreground shadow-md outline-none ' ~ attributes.render('class'))|tailwind_merge }}"
```
Kocal: I noticed that the hover-card 's content is opened when you click on the trigger element, but it stays opened when leaving the mouse out. Shouldn't be: 

```suggestion
    class="{{ ('invisible opacity-0 group-hover/hover-card:visible group-hover/hover-card:opacity-100 group-hover/hover-card:visible group-hover/hover-card:opacity-100 transition-opacity absolute left-1/2 top-full z-50 mt-2 w-64 -translate-x-1/2 rounded-md border bg-popover p-4 text-sm text-popover-foreground shadow-md outline-none ' ~ attributes.render('class'))|tailwind_merge }}"
```

?

Amoifr:
Hi @Kocal! Applied your suggestion in b4851401 (with the duplicated classes deduped, that was just a copy-paste artefact in the snippet — no worries 🙂).

Quick disclaimer though: thinking about it after the fact, I'm wondering if a slightly different fix would have been better. The reason `group-focus-within` was there in the first place is keyboard accessibility — combined with the `tabindex="0"` on the Trigger, it lets keyboard-only users tab onto the trigger and reveal the content. Removing it (without also removing the `tabindex`) means:

- ✅ click + mouse-away no longer leaves the card stuck open (your bug, fixed)
- ❌ keyboard users tabbing onto the trigger no longer see the content at all

If you'd rather, I can switch to one of:
- drop the `tabindex="0"` from the Trigger too, making this a hover/touch-only component (consistent, no half-broken a11y)
- or keep things as they are now and address keyboard support in a follow-up with a small Stimulus controller (the same pattern used by the other Shadcn toggle components)

Happy to do whichever you prefer! 🙏

Kocal:
Ah you're right, I forgot about a11y when focusing the trigger with tab key. 
Indeed, I also think having a small Stimulus controller for showing the content at hovering/focusing the trigger is required

Amoifr:
Hi @Kocal! Done in 1c48c8ab. The new `hover_card_controller.js` toggles `data-state` on the root on `mouseenter`/`mouseleave` and the trigger wires `focus`/`blur` in addition. Both keyboard accessibility and the original click-then-leave bug are now handled. I also dropped the `group-hover/hover-card` CSS hack in favor of `in-data-[state=open]` selectors so the JS is the single source of truth.

Bonus: I added optional `openDelay` / `closeDelay` props on `<twig:HoverCard>` so consumers can tune the show/hide timing if they want — defaults to `0` for no behavior change.

Let me know what you think! 🙏

Kocal:
That's a lot better, thank you! 

Merging when it's green

Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3481, opened 2026-04-18
[Toolkit][Shadcn] Add slider recipe

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| License        | MIT

Adds the `slider` recipe to the Shadcn kit.

It's a fully interactive component built with Twig and Stimulus only, no external JS slider library. It supports keyboard navigation (arrows, Home, End, PageUp, PageDown), pointer drag, multiple thumbs, vertical orientation and RTL.

Examples are synced with the current shadcn base slider reference (https://ui.shadcn.com/docs/components/base/slider): default/hero, Range, Multiple Thumbs, Vertical, Disabled, Controlled and RTL.

Follows the current kit conventions, with examples inlined directly in the recipe's `README.md`. Part of splitting #3466 into one PR per component, tracked by #3233.

### DIFF
--- a/src/Toolkit/kits/shadcn/slider/assets/controllers/slider_controller.js
+++ b/src/Toolkit/kits/shadcn/slider/assets/controllers/slider_controller.js
@@ -1,0 +1,197 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+    static targets = ['track', 'range', 'thumb', 'input'];
+    static values = {
+        min: { type: Number, default: 0 },
+        max: { type: Number, default: 100 },
+        step: { type: Number, default: 1 },
+        orientation: { type: String, default: 'horizontal' },
+    };
+
+    connect() {
+        this._activeThumb = null;
+        this._abortController = new AbortController();
+        const { signal } = this._abortController;
+
+        this.thumbTargets.forEach((thumb, index) => {
+            thumb.addEventListener('pointerdown', (e) => this._onThumbDown(e, index), { signal });
+            thumb.addEventListener('keydown', (e) => this._onThumbKey(e, index), { signal });
+        });
+        this.trackTarget.addEventListener('pointerdown', (e) => this._onTrackDown(e), { signal });
+
+        this.update();
+    }
+
+    disconnect() {
+        this._abortController?.abort();
+        this._abortController = null;
+        this._dragAbortController?.abort();
+        this._dragAbortController = null;
+        this._activeThumb = null;
+    }
+
+    _startDrag() {
+        this._dragAbortController?.abort();
+        this._dragAbortController = new AbortController();
+        const { signal } = this._dragAbortController;
+        window.addEventListener('pointermove', (e) => this._onPointerMove(e), { signal });
+        window.addEventListener('pointerup', () => this._onPointerUp(), { signal });
+    }
+
+    _onThumbDown(event, index) {
+        if (this._isDisabled()) return;
+        event.preventDefault();
+        this._activeThumb = index;
+        this.thumbTargets[index].focus();
+        this._startDrag();
+    }
+
+    _onTrackDown(event) {
+        if (this._isDisabled()) return;
+        const ratio = this._pointerRatio(event);
+        const value = this._ratioToValue(ratio);
+        const index = this._nearestThumbIndex(value);
+        this._setThumbValue(index, value);
+        this._activeThumb = index;
+        this.thumbTargets[index].focus();
+        this._startDrag();
+    }
+
+    _onPointerMove(event) {
+        if (this._activeThumb === null) return;
+        const ratio = this._pointerRatio(event);
+        this._setThumbValue(this._activeThumb, this._ratioToValue(ratio));
+    }
+
+    _onPointerUp() {
+        this._activeThumb = null;
+        this._dragAbortController?.abort();
+        this._dragAbortController = null;
+    }
+
+    _onThumbKey(event, index) {
+        if (this._isDisabled()) return;
+        const current = Number(this.inputTargets[index].value);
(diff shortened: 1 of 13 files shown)


Amoifr:
Main example 


https://github.com/user-attachments/assets/1265c31e-d520-4b99-8829-995d808a527e

Amoifr:
Range example 

https://github.com/user-attachments/assets/265012c7-d0c5-41ad-aad2-bf344a5b6ff1

Amoifr:
Multiple Thumbs example

https://github.com/user-attachments/assets/f2b86814-c7c6-490b-8178-5ce198cb2573

Amoifr:
Vertical example


https://github.com/user-attachments/assets/54d7d0da-acda-497d-8cbe-881e9f4a1f93

Amoifr:
Disabled example 

https://github.com/user-attachments/assets/5ed1e34b-c3ff-49d0-ba5c-0f0aafe61944

Amoifr:
Controlled example 


https://github.com/user-attachments/assets/4fea1d34-f1d9-4909-861d-32cb255c5b6d

Amoifr:
RTL example


https://github.com/user-attachments/assets/df0bc59c-5698-4adc-827a-dda96ef41643

src/Toolkit/kits/shadcn/slider/assets/controllers/slider_controller.js
```diff
@@ -0,0 +1,194 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+    static targets = ['track', 'range', 'thumb', 'input'];
```
zairigimad, line 4: Can we find another naming for thumb, they call this thumbs in shadcns doc but not in their code, maybe we can call it handle like popular UI kits ( Radix, chakra ui ... )
Wdyt ?

src/Toolkit/kits/shadcn/slider/assets/controllers/slider_controller.js
```diff
@@ -0,0 +1,194 @@
...
+            this.rangeTarget.style.right = `${100 - hi}%`;
+            this.rangeTarget.style.top = '';
+            this.rangeTarget.style.bottom = '';
+        }
+    }
+
+    _isDisabled() {
+        return this.element.getAttribute('aria-disabled') === 'true' || this.element.matches('[data-disabled]');
+    }
+}
```
zairigimad, line 208: Maybe we should clear the event listeners in the disconnect.

src/Toolkit/kits/shadcn/slider/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "$schema": "../../../schema-kit-recipe-v1.json",
+    "type": "component",
+    "name": "Slider",
+    "description": "An input that allows the user to select a value from a given range.",
```
zairigimad: An input where the user selects a value from within a given range.

zairigimad:
From the videos, it looks great. I left some minor comments

src/Toolkit/kits/shadcn/slider/assets/controllers/slider_controller.js
Amoifr: Thanks for raising this! I went back and forth on the naming and landed on `thumb` to stay aligned with shadcn upstream, which uses `SliderPrimitive.Thumb` and `data-slot="slider-thumb"`. The ARIA pattern itself puts `role="slider"` on that element, so "thumb" is also the accessible name it inherits in assistive tech docs. Since this kit mirrors shadcn's vocabulary, I'd rather keep `thumb` here to avoid a one-recipe divergence. If the broader kit leans towards Radix/Chakra-style `handle` in the future, happy to revisit it kit-wide in one pass. WDYT? 🙏

src/Toolkit/kits/shadcn/slider/assets/controllers/slider_controller.js
Amoifr: Good catch, thanks! Fixed in e1fc1e17 — using a single `AbortController` for the static thumb/track listeners, plus a second one scoped to the drag that's also aborted in `disconnect()`. Bonus: the inline arrow functions are now safely removable, and interrupting a drag mid-Turbo-nav no longer leaks `window` listeners.

src/Toolkit/kits/shadcn/slider/manifest.json
Amoifr: Much nicer wording, adopted in e1fc1e17. Thanks!

Amoifr:
While reviewing #3483 (input-otp), @Kocal pointed out a missing per-input \`aria-label\`, and the same gap was hiding here on the slider thumbs: each \`<span role="slider">\` had \`aria-valuenow\` but no label, which makes range/multi-thumb sliders confusing for screen-reader users (you hear "slider 25 / slider 50" with no way to know which is min and which is max).

Just pushed a tiny a11y patch to fix that:

- Each thumb now gets a default \`aria-label\` — \`Slider\` for a single thumb, \`Slider thumb X of Y\` for multi-thumb.
- New optional \`labels\` prop (comma-separated string or array) lets you provide friendlier labels:
  \`\`\`twig
  <twig:Slider value="25,50" labels="Min,Max" />
  <twig:Slider value="20,50,80" labels="Low,Medium,High" />
  \`\`\`
- Range / Multiple examples updated, snapshots regenerated.

Force-pushed on top of the existing branch. Happy to adjust the wording or the prop name if you'd prefer something else! 🙏

Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3467, opened 2026-04-14
[Toolkit][Shadcn] Add input-otp, resizable and navigation-menu recipes

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Part of #3233
| License        | MIT

Hi 👋

Third batch of recipes from the unchecked list in #3233 (previous batches: #3464, #3466). Still JS-free and leaning on native HTML / CSS:

- **input-otp** — a group of single-character inputs (`inputmode=numeric`, `maxlength=1`), with `InputOtp:Slot` and an optional `InputOtp:Separator`.
- **resizable** — a panel resizable via the browser-native `resize` CSS property. Accepts a `direction` prop (`both` / `horizontal` / `vertical`). Note: shadcn's official Resizable is a panel group with draggable split bars (Radix primitive), which requires JS; this recipe ships the pragmatic CSS-only variant — happy to drop it if you'd prefer to wait for the draggable version.
- **navigation-menu** — horizontal nav with optional hover/focus submenus, using `group/nav-item` + `group-hover/nav-item` / `group-focus-within/nav-item` (same pattern as the `hover-card` recipe in #3466). Comes with `List`, `Item`, `Link`, `Trigger`, and `Content` sub-components.

Each recipe has `Usage.html.twig`, `Demo.html.twig`, and passing snapshot tests (`254 tests, 508 assertions`).

Happy to split if easier — each recipe is a single commit.

Thanks again! 🙏

Issue #2522, opened 2025-01-23
[TwigComponent] Add `aware` tag


seb-jean:
Hi,

This issue is a continuation of https://github.com/symfony/ux/issues/2514.

I looked at how this was done for [Laravel components](https://laravel.com/docs/11.x/blade#accessing-parent-data). It uses an `@aware` directive to pass props from a parent component to a child component. This also allows us to say which props will be accessible or not for children.  This works for self closing tags but also for opening and closing tags.

WebMamba:
I think every component should be isolated as much as possible. You should not have components depending on the external context. Why not just use props?

Kocal:
100% agree here, _good_ components must stay isolated the more as possible.

seb-jean:
The idea behind this is to avoid repeating the props on each child component like this:

```twig
<twig:Form:Row disabled>
    <twig:Form:Label disabled>Description</twig:Form:Label>
    <twig:Form:Textarea disabled />
    <twig:Form:Description disabled>Symfony UX Core Team is awesome!</twig:Form:Description>
</twig:Form:Row>
```

smnandre:
So a sort of temporary data-store, during the render. 

That would not be "passed down" but more...  "available anywhere while the parent component is rendering" .. right ?

smnandre:
@seb-jean : is the sample you provide a real-world one ? Like here, why would you need to "pass down" the `disabled` attribute ? (genuine question i'm looking for examples)

seb-jean:
Here are some examples that inspired me:
- https://catalyst.tailwindui.com/docs/input#disabled-state
- https://catalyst.tailwindui.com/docs/checkbox#disabled-state

smnandre:
Disabled can be set on the highest form element to disable everything below, not sure if this is a good example (checkbox disabled would need to be identified)

For the rest, have I understood correctly for the "inside share context"  ?

seb-jean:
Yes, you understood correctly :).

Kocal:
So, something like https://vuejs.org/guide/components/provide-inject or https://react.dev/reference/react/useContext#passing-data-deeply-into-the-tree

smnandre:
Hmm.. not exactly. I mean what i had in mind, but we could thing anything. Just be aware (pun intended) this will hurt our brain the second we need to prioritize things.

So what i had in mind was more the context for the content (so anything in Twig)

If i look at your examples and the one @Kocal shared, we are then speaking about mounting data as props.

[...]

PR #3483, opened 2026-04-18
[Toolkit][Shadcn] Add input-otp recipe

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| License        | MIT

Adds the `input-otp` recipe to the Shadcn kit.

Part of the split of #3467 into one PR per component, tracking #3233.

Snapshots will be regenerated after rework.

### DIFF
--- a/src/Toolkit/.gitignore
+++ b/src/Toolkit/.gitignore
@@ -1,5 +1,6 @@
 /vendor/
 /config/reference.php
+/config/schema.json
 /composer.lock
 /phpunit.xml
 /.phpunit.cache
--- a/src/Toolkit/CHANGELOG.md
+++ b/src/Toolkit/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add support for blocks: ready-to-use page sections built from a kit's components, installed and previewed as a unit
 - [Shadcn] Add the `login-01` and `login-02` login blocks
+- [Shadcn] Add `input-otp` recipe
 - [Shadcn][Flowbite] Merge component classes with `attributes.defaults({ class: '...'|tailwind_classes })` instead of `tailwind_merge` (consumer classes now override component variants)
 - [Shadcn] Fix Field/Label, InputGroup and Pagination silently dropping their own base classes when forwarding to a child component (their checked/disabled/nested/dark styles now apply)
 - Remove the now-obsolete `ClassMergeSpacingChecker` linter
--- a/src/Toolkit/kits/shadcn/input-otp/README.md
+++ b/src/Toolkit/kits/shadcn/input-otp/README.md
@@ -1,0 +1,254 @@
+# Input OTP
+
+A group of single-character inputs for one-time passwords and verification codes.
+
+```twig {"preview":true}
+<twig:InputOtp inputs="6">
+    <twig:InputOtp:Group>
+        {% for i in 1..6 %}
+            <twig:InputOtp:Slot input="{{ i }}" value="{{ i }}" />
+        {% endfor %}
+    </twig:InputOtp:Group>
+</twig:InputOtp>
+```
+
+## Installation
+
+::: installation
+
+## Usage
+
+```twig
+<twig:InputOtp inputs="6">
+    <twig:InputOtp:Slot input="1" />
+    <twig:InputOtp:Slot input="2" />
+    <twig:InputOtp:Slot input="3" />
+    <twig:InputOtp:Slot input="4" />
+    <twig:InputOtp:Slot input="5" />
+    <twig:InputOtp:Slot input="6" />
+</twig:InputOtp>
+```
+
+## Examples
+
+### Four Digits
+
+Use the `inputs` prop on `InputOtp` to control the number of slots; each `InputOtp:Slot` derives its accessible label and `name` from it and its own `input` position.
+
+```twig {"preview":true}
+<twig:InputOtp inputs="4">
+    <twig:InputOtp:Group>
+        {% for i in 1..4 %}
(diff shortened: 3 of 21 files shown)


Amoifr:
Pattern example :


https://github.com/user-attachments/assets/d920e9e0-8149-48d4-b47c-655d19d0c2c2

Amoifr:
Separator example 


https://github.com/user-attachments/assets/3fffae25-cefa-4535-93c8-d43e5baab24b

Amoifr:
Disabled example 


https://github.com/user-attachments/assets/0ba25834-6e66-40be-96a2-fbf0c5f963fd

Amoifr:
Controlled example : 


https://github.com/user-attachments/assets/64b6bba4-4ec8-401f-8887-9ce23e9f916f

Amoifr:
Invalid example 

Capture d’écran du 2026-04-19 15-27-49

Amoifr:
Four digits example :


https://github.com/user-attachments/assets/b5eae5fb-5913-4c90-9493-8e2ac590556c

Amoifr:
Alphanumeric example :


https://github.com/user-attachments/assets/b8ee6e04-53ec-43ee-9082-645e6c7d5db9

Amoifr:
Form example : 

https://github.com/user-attachments/assets/9fbf8441-a6ef-4a9d-a1c8-70265f3c066c

Amoifr:
RTL example : 


https://github.com/user-attachments/assets/8cf3d8a1-62ad-4310-94d6-4f1124d37c20

Kocal (commented):
Thank you for your work, it's almost perfect! 👏🏻  
I just have two comments: 
1. The absence of the `autocomplete=“one-time-code”` attribute, which prevents autocompletion from a password manager, SMS, etc...
2. The absence of a title/aria-label `Input X of Y` for each `<twig:InputOtp:Slot>`, which prevents proper accessibility.

As suggestions:
1. I think we can add the attribute to all `<twig:InputOtp:Slot>` elements, not just the first one, but this needs to be double-checked
2. Since TwigComponents do not have the concept of `children` components (like React or Vue), we'll need to make this explicit using `<twig:InputOtp inputs="6">` and `<twig:InputOtp:Slot input="1">`, for example (if you have a better name than `input`, go for it!) 

PS: I force pushed your branch, but I didn't commit anything, I just rebased it "on reflex" 😅 you can force push again without worries

Amoifr:
Thanks a ton for the careful review and the helpful suggestions! 🙏 Both points addressed:

1. **`autocomplete="one-time-code"`** — added by default on every `<twig:InputOtp:Slot>`, users can still override per-slot (e.g. `autocomplete="off"`) via the merged attributes.

2. **Accessible `Input X of Y` label** — went the explicit route as you suggested, with one twist: I tried first to forward `inputs="6"` from `<twig:InputOtp>` down to the `Slot` via the `{% set _kit_var %}` pattern (the one Accordion uses), but since slots live inside a `<twig:InputOtp:Group>` wrapper, the variable doesn't survive the extra component level (got a clean `Variable "_input_otp_inputs" does not exist`). So I exposed two props directly on the slot:

   ```twig
   <twig:InputOtp:Slot input="1" inputs="6" name="otp[0]" />
   ```

   `input` alone yields `aria-label="Input 1"`, and `input` + `inputs` yields `aria-label="Input 1 of 6"`. A bit more verbose to write, but rock-solid and matches the spirit of your suggestion. All 11 examples updated + snapshots regenerated.

And no worries about the rebase — that "force-push reflex" is a feature, not a bug 😄. Force-pushed the new commits, ready for another pass whenever you have a minute!

Kocal:
About 2., this is related to https://github.com/symfony/ux/issues/2522, I will try to work on it, I really don't like the current workaround 😅 🤞🏻

Kocal:
https://github.com/symfony/ux/pull/3512 has been merged, I didn't tweak our skill yet, but I believe you can already starts to make this recipe better ;)

Amoifr:
Done! The recipe now builds on `provide()`/`inject()`, thanks for the nudge (and for #3512, it fits this component like a glove).

`InputOtp` provides `inputs` and `name`, and each `Slot` only declares its `input` position: the accessible "Input X of Y" label and the default `name` are derived from the context, so the `inputs="6" name="otp[...]"` repetition is gone from every example.

The branch predated the new Toolkit conventions, so I also migrated the recipe while I was at it: README-based examples, `tailwind_classes` merging, `@prop` docblocks, and `ux:toolkit:lint-kit` is green. Snapshots regenerated, 610 rendering tests pass locally.

Kocal:
Thanks for making some changes, I just did theses:
  - Enforced the OTP `pattern` while typing/pasting, not just on submit. The Stimulus controller now checks each keystroke and paste against the slot's `pattern`.
  - Swapped the plain text `-` separator for the Lucide `minus` icon (`<twig:ux:icon name="lucide:minus">`) to match the shadcn reference, and added `symfony/ux-icons` to the recipe manifest.
  - Rewired the controller to the kit's usual Stimulus convention (`data-action` + `static targets`) instead of manual `addEventListener`/`bind` in `connect()`. Matches the other shadcn controllers and cuts ~20 lines. Also memoized the compiled `pattern` regex instead of recompiling it on every keystroke.
  - Removed a stray `src/Toolkit/config/schema.json` and added  it to `.gitignore` alongside `config/reference.php`.

Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3464, opened 2026-04-14
[Toolkit][Shadcn] Add radio-group, collapsible, typography and toggle-group recipes

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Part of #3233
| License        | MIT

Hi 👋

This PR adds four recipes from the unchecked list in #3233:

- **radio-group** — form control mirroring the Checkbox pattern (native `<input type=radio>` + `accent-primary`), with a `RadioGroup:Item` sub-component.
- **collapsible** — native `<details>` / `<summary>` based, no JS.
- **typography** — H1–H4, P, Blockquote, List, InlineCode sub-components.
- **toggle-group** — container for Toggle items, with `variant` / `size` propagated to items through outer-context variables (`_toggle_group_variant` / `_toggle_group_size`).

Each recipe has a `Usage.html.twig` and a `Demo.html.twig` example and passing snapshot tests.

I initially planned to send these as separate PRs and asked for a sanity check on #3233 — happy to split them back up if that is easier to review. Each recipe is a single commit so rebasing into 4 PRs is straightforward.

### Note on sub-component context propagation

While working on `radio-group`, I noticed that variables set in a parent component (`{% set _radio_group_name = ... %}`) do not propagate to **self-closing** children, because `<twig:RadioGroup:Item .../>` compiles to a `component()` function call that does not carry outer context. It works fine for non-self-closing children (e.g. `<twig:ToggleGroup:Item>...</twig:ToggleGroup:Item>`, which uses the `{% component %}` tag — this pattern is used in `toggle-group` to inherit `variant`/`size`).

For `radio-group`, `Item` is always self-closing, so I kept `name` as an explicit prop on each `Item`. Curious if you would prefer a different pattern here (e.g. `outerScope`, forcing a block body, a helper).

Thanks a lot for the Toolkit work! 🙏

PR #3490, opened 2026-04-19
[Toolkit][Shadcn] Align toggle-group examples with shadcn reference

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| License        | MIT

Aligns the `toggle-group` examples with the shadcn reference:

- **Sizes** (`apps/v4/examples/radix/toggle-group-sizes.tsx`) — both groups now use `type="single"` (shadcn uses single, ours defaulted to multiple so all items could be pressed at once).
- **Spacing** (`apps/v4/examples/radix/toggle-group-spacing.tsx`) — same fix: `type="single"`.
- **RTL** (`apps/v4/examples/radix/toggle-group-rtl.tsx`) — adds the Hebrew version stacked below the Arabic one. Shadcn's reference switches languages via the shared `language-selector` component (React state + `<Select>`); the Twig toolkit has no equivalent, so the example stacks both variants like the `collapsible` RTL example does.

Follow-up to #3464.

Snapshots regenerated.


Kocal:
I just found out the `single` type is not working correctly. I'm merging, and will fix it in another PR

**EDIT:** I didn't see https://github.com/symfony/ux.symfony.com/pull/76, it's working fine 👍🏻

Kocal:
Thank you @Amoifr.

Issue #2786, opened 2025-05-28
[Map] Leaftlet 2.0 migration


smnandre:
Some changes to support here. We need to discuss how we can (if we can) handle both versions

https://github.com/Leaflet/Leaflet/blob/main/CHANGELOG.md

Kocal:
Didn't even know Leaflet 2.0 was a thing! 

I quickly read the breaking changes and I don't see anything preventing us to add Leaflet 2 support

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #3492, opened 2026-04-20
[Map][Leaflet] Add Leaflet 2.x support via runtime adapter

makraz:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | #2786
| License        | MIT

handle both Leaftlet versions

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -303,6 +303,9 @@ importers:
       typescript:
         specifier: ^5.8.3
         version: 5.8.3
+      vitest:
+        specifier: ^4.1.0
+        version: 4.1.0(@types/node@25.3.0)(jsdom@26.1.0)(msw@2.10.4(@types/node@25.3.0)(typescript@5.8.3))(vite@7.3.1(@types/node@25.3.0)(lightningcss@1.32.0)(terser@5.43.1)(tsx@4.20.3))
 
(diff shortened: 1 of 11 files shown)


Kocal:
Leaflet 2 is still alpha, and it's stalled: only `2.0.0-alpha` (May 2025) and `2.0.0-alpha.1` (August 2025) have ever been published, nothing in the 13 months since, and npm's `latest` tag is still `1.9.4`. The peer dependency range this PR declares, `^1.9.4 || ^2.0.0`, doesn't even match `2.0.0-alpha.1`, since caret ranges exclude prereleases, so nobody can actually install the combination it advertises.

On top of that, the v2 code path is never exercised against real Leaflet 2: devDependencies stay pinned to `leaflet: ^1.9.4`, and the new vitest suite mocks the v2 API instead of importing it, so the green CI here only covers the v1 path. Writing an adapter against an alpha that can still change underneath it is likely to be wrong by the time Leaflet 2 actually ships, so I'd rather close this now than merge something we can't verify.

Thanks for the work here, @makraz. I'm closing this rather than rejecting the idea: the branch stays available, and we'd welcome it back once Leaflet 2 reaches beta or stable and there's a real version to test against. #2786 stays open to track it.

PR #3493, opened 2026-04-20
Remove old "versionadded: 2.x" directives 

MrYamous:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | 
| License        | MIT

Update doctor rst config and documentation for 3.x

PR #3505, opened 2026-04-26
[Turbo] Add `<turbo-mercure-stream-source>` custom element

### LINKED ISSUE #2460: [Turbo] Improving Turbo Stream Sources and Listeners (proposal)
**UPDATE 2025-02-23**

> [!TIP]
> If you want to have a good example supporting this suggestion : https://github.com/symfony/ux/pull/2447#issuecomment-2676467072 

Original post below

-----

## TL;DR; - Turbo Stream

Proposal to introduce new helper / classes
* StreamSourceUrlGenerator
* turbo-stream-source-controller.js
* turbo_stream_source_url
* <twig:turbo:stream:source />

With all implementation details being hidden behind the StreamSourceUrlGenerator.

This would ease turbo stream usage, and open features to custom implementations while keeping things seamless for users.

Open to feedback, ideas, comments !

## 1. Main Goal

We aim to improve the developer experience (DX) offered by UX Turbo when using Turbo, especially Turbo Streams.

Several recent pull requests have helped us move forward:
- New `TurboStreamResponse` and `TurboStream` helper classes in [#2196](https://github.com/symfony/ux/pull/2196), completed in [#2298](https://github.com/symfony/ux/pull/2298)
- New `<twig:Turbo:Stream />` components in [#2227](https://github.com/symfony/ux/pull/2227), completed in [#2302](https://github.com/symfony/ux/pull/2302)
- New `<twig:Turbo:Frame />` in [#2303](https://github.com/symfony/ux/pull/2303)

*(Apologies if I missed one—please correct me! And thank you to everyone involved.)*

## 2. Current Situation

@Norkunas recently had to find clever internal tricks ([#2407](https://github.com/symfony/ux/pull/2407)) to allow registration on multiple topics at once.

In [#2447](https://github.com/symfony/ux/pull/2447), @Fan2Shrek tried to adapt the current Twig function to pass Mercure options to handle authorization.

This, however, raised questions about implementation, coupling, current limitations, and DX.

### Challenges

As demonstrated by @Norkunas and @Fan2Shrek, we must ensure compliance with Symfony's BC (backward compatibility) promise.

For a full explanation of what this entails, refer to [Symfony’s BC guidelines](https://symfony.com/doc/current/contributing/code/bc.html).

In this context, our actions are mostly constrained by:
- The `TurboStreamListenRendererInterface`
- The `TurboStreamListenRenderer`

### Limitations

But there is also several issues with the current DX.

The `turbo_render_listen` function does not render a full HTML element, only attributes. 

This requires users to create the HTML tag themselves, which is not intuitive and slightly inconsistent with other UX features. 

Additionally:
- Stimulus attributes cannot be chained, making it impossible to add another Stimulus controller to the same tag.
- Directly rendering attributes as escaped strings makes the function incompatible with Twig components, custom Stimulus attributes, or JavaScript. This also makes it unsafe for use with includes, embeds, etc., without requiring a `|raw` filter.

### Coupling

The PHP code utilizes interfaces, so the coupling with the Mercure bridge is currently low. 

However, on the JavaScript/controller side, the situation is different. The `turbo_stream_controller.js` is exclusively used as a Mercure stream controller, making it currently impossible to use it with other types of event sources. The controller requires multiple inputs/values (hub, topics, etc.), but ultimately only needs a URL for the EventSource required by Turbo. We should construct this URL in PHP and pass it directly.

Less important, the `Twig function renderer interface` depends on the `Environment` class. Yet, this dependency is not currently used in `TurboStreamListenRenderer`, as it is injected into the `StimulusHelper`.

### Mercure Internals

We should avoid implementing anything in the Turbo Bundle that duplicates functionality from the Mercure Bundle. Instead, it should be clear that the Mercure Bundle is a requirement for using Mercure streams with Turbo. All aspects related to security, authorization, etc., should be delegated to the Mercure Bundle.

Simultaneously, any direct usage of Mercure Bundle code (or assumptions about its existence) should include safeguards to avoid issues arising from future updates to the bundle.

### Turbo Needs

Taken from the Turbo documentation:
> Turbo's `<turbo-stream-source>` custom element connects to a stream source through its `[src]` attribute. When declared with a `ws://` or `wss://` URL, the underlying stream source will be a WebSocket instance. Otherwise, the connection uses an EventSource.

Ultimately, we only need a URL and possibly some options. Constructing this (including creating cookies or JWTs if necessary) should be handled internally.

## 4. Proposal

- Keep existing methods/interfaces (?)
- Introduce a generic StreamSourceUrlGenerator
- Introduce a new generic function/component to render Turbo Stream sources
- Create a new `turbo_stream_source_controller` to handle EventSource 

#### Interface: `TurboStreamUrlGenerator`

Introduce an interface:

```php
interface UX\Turbo\StreamSourceUrlGenerator {
    public function generateUrl(): string;
}
```

This interface would function similarly to the Router component, allowing for implementation flexibility. 

We may need to pass additional parameters or introduce a Value Object (VO). I'm open to ideas as long as it remains generic and future-proof.

This allows users to use the assembled URL for any purpose.

### Stream Source Controller

A new Stimulus Controller could be introduced, doing exactly what does the existing one, but manipulating only one value (the URL).

```html
<div 
    class="foo" 
    data-controller="[...]-stream-source" 
    data-...-stream-source-url-value="{{ event_source }}" 
>
   ...
</div>
```

### Stream Source Tag

Rather than rendering and escaping HTML attributes (which requires a Stimulus controller), we could generate a simple HTML element:

```html
<turbo-stream-source src="..." />
```

This approach provides a clean DX for most users.

### Twig Function: `turbo_stream_source()`

A new `turbo_stream_source()` function could be introduced. This function would:
- Use the same signature as `turbo_stream_url()` (which it would call internally) ? 
- ...Or maybe accept only an url parameter ? 
- Handle escaping and other concerns

```twig
{{ turbo_stream_source() }}
```

```twig
{{ turbo_stream_source(url: ...) }}
```

### Twig Component: `<twig:turbo:stream:source />`

The same functionality could be offered as a Twig component. The exact name is up for discussion (`<twig:turbo:source />` might be simpler but less descriptive).

Example:

```twig
<twig:turbo:stream:source foo="..." topic="{{ [topicA, topicB] }}" ... />
```

or

```twig
<twig:turbo:stream:source url="{{ urlParams }}" />
```

## Next Steps

I’d love to hear your feedback, ideas, and desires on this topic (as long as they stay within scope, haha).  

I’m happy to start working on this or assist anyone interested in contributing!

### Mercure
- [Mercure website](https://mercure.rocks/)
- [Symfony Mercure Bundle](https://symfony.com/doc/current/mercure.html)

### Turbo
- [Turbo](https://turbo.hotwired.dev)
- [Turbo Streams](https://turbo.hotwired.dev/handbook/streams)
- [Integrating Turbo Streams with the Backend](https://turbo.hotwired.dev/handbook/streams#integration-with-server-side-frameworks)
- [Processing Stream Elements](https://turbo.hotwired.dev/reference/streams#processing-stream-elements)

### UX Turbo
- [Website & Demo](https://ux.symfony.com/turbo)
- [Documentation](https://symfony.com/bundles/ux-turbo/current/index.html)

### Web Events
- [WebSockets on MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API)
- [EventSource on MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventSource)
- [EventSource HTML Spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface)

seb-jean:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | yes 
| Documentation? | yes 
| Issues         | Fix #2460 
| License        | MIT

**Summary**

This PR introduces a native HTML custom element `<turbo-mercure-stream-source>` to replace the `mercure-turbo-stream` Stimulus controller for subscribing to Mercure topics and receiving Turbo Stream updates.

**What's new**

- `<turbo-mercure-stream-source>` custom element (`mercure_stream_source_element.js`): manages the `EventSource` lifecycle (connect, disconnect, reconnect on attribute change) and sets a `connected` attribute when the SSE connection is established 
- `StreamSourceUrlGenerator` interface: decouples URL generation from rendering, making custom transports easier to implement
- `MercureStreamSourceUrlGenerator`: Mercure implementation, delegates authorization cookie handling to the Mercure Bundle for private topics
- `turbo_stream_from()` Twig function and `<twig:Turbo:Stream:From>` Twig component: the new way to subscribe to Mercure topics in templates

```twig
{# Before #}
<div id="books" {{ turbo_stream_listen('App\\Entity\\Book') }}></div>
```

```twig
{# After - Twig component #}
<twig:Turbo:Stream:From topics="App\\Entity\\Book" />
<div id="books"></div>
```

```twig
{# After - Twig function #}
{{ turbo_stream_from('App\\Entity\\Book') }}
<div id="books"></div>
```

```twig
{# Private topic #}
<twig:Turbo:Stream:From topics="App\\Entity\\Book" private />
{{ turbo_stream_from('App\\Entity\\Book', private=true) }}
```

```twig
{# Specific transport #}
<twig:Turbo:Stream:From topics="App\\Entity\\Book" transport="hub2" />
{{ turbo_stream_from('App\\Entity\\Book', transport='hub2') }}
```

```twig
{# Multiple topics #}
<twig:Turbo:Stream:From :topics="['topic_a', 'topic_b']" />
{{ turbo_stream_from(['topic_a', 'topic_b']) }}
```

**Deprecations**

- `turbo_stream_listen()` Twig function → use `turbo_stream_from()` or `<twig:Turbo:Stream:From>`
- `TurboStreamListenRendererInterface` → implement `StreamSourceUrlGenerator` instead
- `TurboStreamListenRenderer` → use `MercureStreamSourceUrlGenerator`
- `TopicSet`
- `mercure-turbo-stream` Stimulus controller → replaced by the `<turbo-mercure-stream-source>` custom element, auto-imported via `turbo-core`

Thanks to @smnandre for opening issue #2460, which inspired this PR. 🙏


apps/e2e/templates/ux_turbo/broadcast/artist.html.twig
```diff
@@ -9,7 +11,7 @@
     <div class="card mt-3">
         <div class="card-header">Songs</div>
         <div class="card-body">
-            <div id="songs-list" {{ turbo_stream_listen('songs_by_artist_' ~ artist.id) }}></div>
```
Kocal, line 12: You sure have more experience than me with Turbo/UX Turbo, but does it mean that since the beginning, it was possible to use `{{ turbo_stream_listen('songs_by_artist_' ~ artist.id) }}` on another element? (I mean, an element that will never ever be updated by a Turbo Frame?)

bin/build_package.ts
```diff
@@ -41,6 +41,7 @@ async function main() {
     const inputCssFile = packageData?.config?.css_source;
     const inputFiles = [
         ...globSync('src/*controller.ts'),
+        ...globSync('src/*element.ts'),
```
Kocal: I would suggest `...(isTurbo ? ['src/...'] : [])` instead, since it's ultra specific

src/Turbo/assets/dist/turbo_stream_controller.js
```diff
@@ -1,7 +1,12 @@
 import { Controller } from "@hotwired/stimulus";
 import { connectStreamSource, disconnectStreamSource } from "@hotwired/turbo";
+/**
+* @author Kévin Dunglas <kevin@dunglas.fr>
+* @deprecated Use turbo_stream_from() or the <twig:Turbo:Stream:From> Twig component instead.
+*/
 var _Class = class extends Controller {
 	initialize() {
+		console.warn("[Symfony UX] The \"mercure-turbo-stream\" Stimulus controller is deprecated since Symfony UX 3.0 and will be removed in 4.0. Use turbo_stream_from() or the <twig:Turbo:Stream:From> Twig component instead.");
```
Kocal:
```suggestion
		console.warn("[Symfony UX] The \"mercure-turbo-stream\" Stimulus controller is deprecated since Symfony UX 3.1 and will be removed in 4.0. Use turbo_stream_from() or the <twig:Turbo:Stream:From> Twig component instead.");
```

Same comment for other deprecation messages

src/Turbo/doc/index.rst
```diff
@@ -707,18 +707,6 @@ Start by installing `the Mercure support`_ on your project:
...
-.. code-block:: diff
-
-    "@symfony/ux-turbo": {
-        "mercure-turbo-stream": {
-    +         "enabled": true,
-    -         "enabled": false,
-            "fetch": "eager"
-        }
-    },
-
```
Kocal, line 721: I don't think we should remove this block of code. This code still works in 3.0 and 3.x overall, removing such documentation could lead to a bad DX IMHO.
Can't we add a comment or a info box telling the user "Since 3.1, manually enabling `mercure-turbo-stream` is not necessary when using `<twig:Turbo:Stream:From>` ..." ?

src/Turbo/src/StreamSourceUrlGenerator.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Symfony\UX\Turbo;
+
+/**
+ * Generates the URL for a Turbo stream source element.
+ *
+ * @author Sébastien Jean <sebastien.jean76@gmail.com>
+ */
+interface StreamSourceUrlGenerator
```
Kocal:
```suggestion
interface StreamSourceUrlGeneratorInterface
```

src/Turbo/src/Twig/TurboRuntime.php
```diff
@@ -29,15 +29,39 @@ public function __construct(
...
+        $transport ??= $this->defaultTransport;
+
+        if (!$this->urlGenerators->has($transport)) {
+            throw new \InvalidArgumentException(\sprintf('The Turbo stream transport "%s" does not exist.', $transport));
+        }
+
+        $url = $this->urlGenerators->get($transport)->generateUrl($topics, ['private' => $private]);
+
+        return \sprintf(
+            '<turbo-mercure-stream-source src="%s"%s></turbo-mercure-stream-source>',
```
Kocal: `<twig:Turbo:Stream:From>` is generic, but we always render a "Mercure-specific" element, can't we decouple things?

Kocal (commented):
Few (and maybe stupid) comments, I don't have such experience with UX Turbo, I trust you here 🤞🏻

src/Turbo/doc/index.rst
```diff
@@ -830,6 +819,38 @@ Keep in mind that you can use all features provided by Symfony Mercure,
 including `private updates`_ (to ensure that only authorized users will
 receive the updates) and `async dispatching with Symfony Messenger`_.
 
+<twig:Turbo:Stream:From> Twig Component
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
Kocal, line 841: Please add a block "Since 3.1: ..."

src/Turbo/src/Bridge/Mercure/MercureStreamSourceUrlGenerator.php
```diff
@@ -0,0 +1,83 @@
...
+        );
+
+        $mercureOptions = ['hub' => $this->hubName];
+        if ($options['private'] ?? false) {
+            $mercureOptions['withCredentials'] = true;
+            $mercureOptions['subscribe'] = $topicStrings;
+        }
+
+        try {
+            return $this->twig->getExtension(MercureExtension::class)->mercure($topicStrings, $mercureOptions);
```
Kocal: This break compatibility with Mercure 0.7+. Instead, it should be something like that:
```suggestion
                // Mercure >= 0.7: https://github.com/symfony/mercure/pull/123
                /* @phpstan-ignore-next-line function.alreadyNarrowedType */
                $mercure = is_subclass_of(MercureExtension::class, AbstractExtension::class)
                    ? $this->twig->getExtension(MercureExtension::class) /* @phpstan-ignore argument.templateType */
                    : $this->twig->getRuntime(MercureExtension::class);
                    
            return $mercure->mercure($topicStrings, $mercureOptions);
```

src/Turbo/src/Bridge/Mercure/MercureStreamSourceUrlGenerator.php
```diff
@@ -0,0 +1,83 @@
...
+        } catch (RuntimeError) {
+            $hubUrl = $this->hub->getPublicUrl();
+            $separator = str_contains($hubUrl, '?') ? '&' : '?';
+
+            return $hubUrl.$separator.implode('&', array_map(
+                static fn (string $t) => 'topic='.rawurlencode($t),
+                $topicStrings,
+            ));
```
Kocal: In this case, I think it's better to not trying to manually build the URL, but just remove the try/catch

src/Turbo/src/Bridge/Mercure/TopicSet.php
```diff
@@ -13,6 +13,8 @@
 
 /**
  * @internal
+ *
+ * @deprecated since Symfony UX 3.0, will be removed in 4.0.
```
Kocal: Why deprecate it?

src/Turbo/src/DependencyInjection/TurboExtension.php
```diff
@@ -90,20 +96,29 @@ private function registerBroadcast(array $config, ContainerBuilder $container, L
...
+                'defaults' => [
+                    'Symfony\\UX\\Turbo\\Twig\\' => [
+                        'template_directory' => '@Turbo/components/',
+                        'name_prefix' => 'Turbo',
+                    ],
                 ],
-            ],
-        ]);
+            ]);
+        }
```
Kocal: Why do we have to do that? AFAIK we can already use existing components like `<twig:Turbo:Stream>` or `<twig:Turbo:Frame>`, right?

src/Turbo/src/Twig/TurboRuntime.php
```diff
@@ -29,15 +29,39 @@ public function __construct(
...
+
     /**
      * @param object|string|array<object|string> $topic
      * @param array<string, mixed>               $options
+     *
+     * @deprecated since Symfony UX 3.0, use renderTurboStreamFrom() or the <twig:Turbo:Stream:From> Twig component instead. Will be removed in 4.0.
      */
     public function renderTurboStreamListen(Environment $env, $topic, ?string $transport = null, array $options = []): string
     {
+        trigger_deprecation('symfony/ux-turbo', '3.0', 'The "%s()" method is deprecated since Symfony UX 3.0, use "%s()" or the <twig:Turbo:Stream:From> Twig component instead. It will be removed in 4.0.', __METHOD__, __CLASS__.'::renderTurboStreamFrom');
```
Kocal: AFAIK there is no need to add both `@deprecated` and `trigger_deprecation()`, you must only use `@deprecated`

src/Turbo/doc/index.rst
```diff
@@ -813,8 +803,7 @@ Let's create our chat::
         </turbo-frame>
     {% endblock %}
 
-If you're using a private hub, you can add ``{ withCredentials: true }``
-as ``turbo_stream_listen()`` third argument to authenticate with the hub
+If you're using a private hub, pass ``private`` to authenticate with the hub
```
Kocal:
```suggestion
If you're using a private hub, pass ``private`` property to ``<twig:Turbo:Stream:From>`` authenticate with the hub
```

src/Turbo/doc/index.rst
```diff
@@ -830,6 +819,38 @@ Keep in mind that you can use all features provided by Symfony Mercure,
...
+    <twig:Turbo:Stream:From topics="App\\Entity\\Book" />
+
+    {# subscribe to a private topic (sets withCredentials on the EventSource) #}
+    <twig:Turbo:Stream:From topics="chat" private />
+
+    {# subscribe using a specific transport #}
+    <twig:Turbo:Stream:From topics="App\\Entity\\Book" transport="hub2" />
+
+    {# subscribe to multiple topics #}
+    <twig:Turbo:Stream:From :topics="['topic_a', 'topic_b']" />
```
Kocal:
```suggestion
    <twig:Turbo:Stream:From topics="{{ ['topic_a', 'topic_b'] }}" />
```

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
```diff
@@ -37,6 +39,7 @@ public function __construct(
 
     public function renderTurboStreamListen(Environment $env, $topic, array $eventSourceOptions = []): string
     {
+        trigger_deprecation('symfony/ux-turbo', '3.0', 'The "%s" class is deprecated since Symfony UX 3.0, use "%s" with turbo_stream_from() or the <twig:Turbo:Stream:From> Twig component instead. It will be removed in 4.0.', self::class, MercureStreamSourceUrlGenerator::class);
```
Kocal: We already have the `@deprecated` annotation on the class itself, is this needed here?

apps/e2e/templates/ux_turbo/broadcast/artist.html.twig
seb-jean: Yes, `turbo_stream_listen()` can be placed on any element.
The incoming Turbo Stream messages then target other elements on the page via their action and targets attributes, independently of the element that holds the subscription.

bin/build_package.ts
seb-jean: Done.

src/Turbo/assets/dist/turbo_stream_controller.js
seb-jean: Done.

src/Turbo/doc/index.rst
seb-jean: Done.

src/Turbo/src/StreamSourceUrlGenerator.php
seb-jean: Done.

src/Turbo/src/Twig/TurboRuntime.php
seb-jean: Done.

src/Turbo/doc/index.rst
seb-jean: Done.

src/Turbo/src/Bridge/Mercure/MercureStreamSourceUrlGenerator.php
seb-jean: Done.

src/Turbo/src/Bridge/Mercure/MercureStreamSourceUrlGenerator.php
seb-jean: Done.

src/Turbo/src/Bridge/Mercure/TopicSet.php
seb-jean: `TopicSet` is only used internally by the deprecated `renderTurboStreamListen()` path. Since that entire path will be removed in 4.0, `TopicSet` will be removed along with it.

src/Turbo/src/DependencyInjection/TurboExtension.php
seb-jean: Oh, my bad.

src/Turbo/src/Twig/TurboRuntime.php
seb-jean: The Symfony conventions require both `@deprecated` and `trigger_deprecation()`: https://symfony.com/doc/current/contributing/code/conventions.html#deprecating-code

src/Turbo/doc/index.rst
seb-jean: Done.

src/Turbo/doc/index.rst
seb-jean: Done.

src/Turbo/src/Bridge/Mercure/TurboStreamListenRenderer.php
seb-jean: Done.

seb-jean:
I'm getting an error related to the `.doctor-rst.yaml` file.

```
src/Turbo/doc/index.rst ✘

722: You are not allowed to use version "3.1". Only major version "2" is allowed.

-> .. deprecated:: 3.1
```

Should I modify `src/Turbo/doc/index.rst` to remove `.. deprecated:: 3.1`, or should I modify the `.doctor-rst.yaml` file?

Kocal:
Following @MrYamous's work in https://github.com/symfony/ux/pull/3493, I think we must update the following rules:
```yaml
    deprecated_directive_major_version:
        major_version: 3
    deprecated_directive_min_version:
        min_version: '3.0'
```

src/Turbo/src/Twig/TurboRuntime.php
Kocal: I don't see where it requires to have both `@deprecated` and `trigger_deprecation()` on the same **method**, it only show an example about deprecating a whole class 🤔 

But it looks like you're right, I found one example on current Symfony version: https://github.com/symfony/symfony/blob/8e8f87c6fa5f47a431fc2f49bdbe601f6769e19d/src/Symfony/Component/Serializer/Exception/PartialDenormalizationException.php#L47-L52

Maybe our documentation could be improved

Kocal:
I just merged https://github.com/symfony/ux/pull/3515, you can rebase this one 🙏🏻

src/Turbo/src/Twig/TurboRuntime.php
seb-jean: For me, the documentation refers to both classes and methods. But as you pointed out, the documentation isn't very clear.
I think that with the example you cited, we can be sure we can include both.

seb-jean:
I just rebased

src/Turbo/src/StreamSourceUrlGeneratorInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+interface StreamSourceUrlGeneratorInterface
+{
+    /**
+     * @param string|object|array<string|object> $topics
+     * @param array<string, mixed>               $options
+     */
+    public function generateUrl(string|object|array $topics, array $options = []): string;
+
+    public function getElement(): string;
```
Kocal: But now, this method does not make any sense in this interface. 

Can't we rename the interface to `StreamSourceRendererInterface` with a single method `render()`, similarly to `TurboStreamListenRendererInterface`?

Kocal (changes requested):

src/Turbo/src/StreamSourceUrlGeneratorInterface.php
seb-jean: Ok, I renamed.

src/Turbo/src/Twig/TwigExtension.php
```diff
@@ -28,9 +29,10 @@ final class TwigExtension extends AbstractExtension
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('turbo_stream_listen', [TurboRuntime::class, 'renderTurboStreamListen'], ['needs_environment' => true, 'is_safe' => ['html']]),
             new TwigFunction('turbo_is_frame_request', [TurboRuntime::class, 'isTurboFrameRequest']),
             new TwigFunction('turbo_frame_request_id', [TurboRuntime::class, 'getTurboFrameRequestId']),
+            new TwigFunction('turbo_stream_from', [TurboRuntime::class, 'renderTurboStreamFrom'], ['needs_environment' => true, 'is_safe' => ['html']]),
+            new TwigFunction('turbo_stream_listen', [TurboRuntime::class, 'renderTurboStreamListen'], ['needs_environment' => true, 'is_safe' => ['html'], 'deprecation_info' => new DeprecatedCallableInfo('symfony/ux-turbo', '3.1', 'turbo_stream_from')]),
```
Kocal: `\Twig\DeprecatedCallableInfo` has been introduced in Twig 3.15 and we don't have requirement on it. Please use something like:
```suggestion
            new TwigFunction('turbo_stream_listen', [TurboRuntime::class, 'renderTurboStreamListen'], [
                'needs_environment' => true, 
                'is_safe' => ['html']
                ..(class_exists(DeprecatedCallableInfo::class)
                    ? ['deprecation_info' => new DeprecatedCallableInfo(/* ... */)]
                    : ['deprecated' => '3.1', 'deprecating_package' => 'symfony/ux-turbo', 'alternative' => 'turbo_stream_from']),
            ]),
```

src/Turbo/src/Bridge/Mercure/TopicSet.php
```diff
@@ -13,6 +13,8 @@
 
 /**
  * @internal
+ *
+ * @deprecated since Symfony UX 3.1, will be removed in 4.0.
```
Kocal, line 19: Missing `trigger_deprecation()`

src/Turbo/assets/src/mercure_stream_source_element.ts
```diff
@@ -0,0 +1,36 @@
+import { connectStreamSource, disconnectStreamSource } from '@hotwired/turbo';
+
+class TurboMercureStreamSourceElement extends HTMLElement {
+    static observedAttributes = ['src', 'private'];
+
+    private es: EventSource | undefined;
+
+    connectedCallback() {
+        this.es = new EventSource(this.getAttribute('src')!, {
```
Kocal: Let's do thing correctly and do not use `!`, but check if the attribut `src` exist on the element, otherwise throw an exception

src/Turbo/CHANGELOG.md
Kocal, line 1: Deprecated classes/methods must be listed entry per entry

src/Turbo/doc/index.rst
Kocal, line 1: The comment https://github.com/symfony/ux/pull/3505/changes#r3176402256 applies for all the file IMHO, since the Symfony website only render `3.x` branch.

apps/e2e/templates/ux_turbo/broadcast/artist.html.twig
```diff
@@ -1,6 +1,8 @@
 {% extends 'example.html.twig' %}
 
 {% block example %}
+    <twig:Turbo:Stream:From :topics="'songs_by_artist_' ~ artist.id" />
```
Kocal:
```suggestion
    <twig:Turbo:Stream:From topics="{{ 'songs_by_artist_' ~ artist.id }}" />
```

Kocal (changes requested):
Final round

src/Turbo/doc/index.rst
seb-jean: You're right, I'll do the migrations from `turbo_stream_listen()` to `<twig:Turbo:Stream:From />` when we are on 4.x.

seb-jean:
> Final round

I made the corrections.

Kocal (approved):

Kocal:
Thank you @seb-jean.

PR #3506, opened 2026-04-26
[Toolkit] Add skill for Toolkit recipe development/review

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

cc @Amoifr @DcgRG @zairigimad @ker0x @seb-jean @CODEheures since you have _recently_ contributed on UX Toolkit.

I didn't have the change to try it yet, and I don't know if you all use LLMs, but any suggestion welcome :)

Generated with the following prompt (and [caveman](https://github.com/JuliusBrussee/caveman/)):
```
Génère un skill pour UX Toolkit, basé sur : 
- le code des Kits existants @src/Toolkit/kits/ 
- le @src/Toolkit/CONTRIBUTING.md 
- les commentaires de reviews dans les PR : 
  - https://github.com/symfony/ux/pull/3464
  - https://github.com/symfony/ux/pull/3488
  - https://github.com/symfony/ux/pull/3490
  - https://github.com/symfony/ux/pull/3485

Il aura pour but de faciliter et normaliser la génération de code, et faciliter la code review. Identifie les différents patterns (twig doc, variable `<component>_trigger_attrs` pour Trigger/Close.html.twig, etc...)
```

Thanks!

### DIFF
--- a/.agents/skills/symfony-ux-toolkit-kit/SKILL.md
+++ b/.agents/skills/symfony-ux-toolkit-kit/SKILL.md
@@ -1,0 +1,438 @@
+---
+name: symfony-ux-toolkit-kit
+description: >
+    Generate, modify, or review Symfony UX Toolkit kit recipes (Shadcn, Flowbite, future kits).
+    Enforces conventions for manifest, Twig component docblocks, sub-components, asChild
+    `<recipe>_<role>_attrs` pattern, outer-scope propagation, Stimulus controllers, examples,
+    snapshots, and PR hygiene. Use when adding/editing files under `src/Toolkit/kits/` or
+    reviewing PRs touching the Toolkit.
+---
+
+# Symfony UX Toolkit — Kit Recipe Skill
+
+Author/review recipes for UX Toolkit. Recipes = unit shipped to users (Twig components + optional Stimulus controllers + examples).
+
+## When to Activate
+
+- User says "add recipe", "new kit recipe", "Toolkit component", "port shadcn X", or similar.
+- Any file change under `src/Toolkit/kits/<kit>/<recipe>/`.
+- Reviewing PR titled `[Toolkit][...]`.
+
+---
+
+## Core Rules
+
+1. **One PR per recipe.** Never batch. PR title: `[Toolkit][<Kit>] Add <recipe> recipe` or `[Toolkit][<Kit>] Align <recipe> with <upstream> reference`.
+2. **Target `3.x`.** CHANGELOG entry under active `3.x` section in `src/Toolkit/CHANGELOG.md`.
+3. **Visual + behavioral parity** with upstream (Shadcn UI / Flowbite). Verify manually; attach screenshot/video for animated/interactive components.
+4. **Reuse all upstream examples.** No subsets. Read component source **and** every example file (see [Shadcn UI](#shadcn-ui) / [Flowbite v4](#flowbite-v4)).
+5. **Companion PR on `symfony/ux.symfony.com`** for visual preview/docs. Link in recipe PR body.
+6. **Regenerate snapshots** after every recipe change, commit them. CI/reviewers reject stale snapshots.
+7. **Use GitHub PR template** (Bug fix / Feature / License: MIT / Issues: Part of #3233). Fabbot fails otherwise.
+8. **Prefer Stimulus controller** over native browser features when parity needs animations, ARIA sync, or coordinated state. Native OK only when it matches upstream UX exactly.
+
+---
+
+## Recipe Directory Layout
+
+```
+src/Toolkit/kits/<kit>/<recipe>/
+├── manifest.json
+├── examples/
+│   ├── Usage.html.twig                    # mandatory, minimal API showcase
+│   ├── Demo.html.twig                     # mandatory, rich preview (used on ux.symfony.com)
+│   └── <Variant Name>.html.twig           # one per upstream example, Title Case with spaces
+├── templates/components/
+│   ├── <Component>.html.twig              # root component
+│   └── <Component>/<SubName>.html.twig    # e.g. Trigger, Close, Header, Item, Content
+└── assets/controllers/                    # optional, only if interactive behavior is needed
+    └── <recipe>_controller.js
+```
+
+Sub-component path `Component/SubName.html.twig` consumed as `<twig:Component:SubName>`.
+
+---
+
+## Shadcn UI
+
+Always emit `data-slot="<recipe-name>"` on root and `data-slot="<recipe-name>-<sub>"` on every sub-component root. This is a Shadcn-specific convention driven by its CSS selectors.
+
+### Upstream sources
+
+Read **both** files per recipe — component source carries canonical classes and `data-*` surface; examples show usage patterns.
+
+| File | Purpose |
+| --- | --- |
+| `apps/v4/styles/radix-nova/ui/<recipe>.tsx` | **Component source** — canonical Tailwind classes, sub-component structure, `data-slot`/`data-state` surface |
+| `apps/v4/examples/radix/<recipe>-*.tsx` | **Usage examples** — one file per variant, drives examples list |
+
+Enumerate every example file for a recipe:
+
+```bash
+gh api "repos/shadcn-ui/ui/git/trees/main?recursive=1" --jq '.tree[].path' \
+  | grep "apps/v4/examples/radix/<recipe>"
+```
+
+Then fetch each:
+```
+https://raw.githubusercontent.com/shadcn-ui/ui/refs/heads/main/apps/v4/examples/radix/<example>.tsx
+```
+
+---
+
+## Flowbite v4
+
+Kit identifier: `flowbite-4`.
+
+### Upstream sources
+
+| Source | Purpose |
+| --- | --- |
+| `https://flowbite.com/docs/components/<recipe>/` | **Reference page** — canonical markup, variants, accessibility notes |
+| `https://github.com/themesberg/flowbite/blob/main/src/components/<recipe>/index.ts` | **JS source** — behavior, state, options (when Stimulus controller needed) |
+
+Flowbite docs page = primary source: copy-pasteable HTML with Tailwind classes + every variant. Read full page before writing any template.
+
+---
+
+## Local Visual Testing
+
+`ux` and `ux.symfony.com` **must be on matching branches**; mismatch causes assetmap failures:
+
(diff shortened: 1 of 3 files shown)


seb-jean:
I'll take a look at it in the next few days.

DcgRG:
@Kocal Same, i'll check ! :)

.agents/skills/symfony-ux-toolkit-kit/SKILL.md
```diff
@@ -0,0 +1,346 @@
...
+        'aria-labelledby': _<recipe>_title_id,
+    }) }}
+>
+    {%- block content %}{% endblock -%}
+</div>
+```
+
+- Always use `attributes.defaults({...})` (NOT raw `{{ attributes }}`). Consumers must be able to override.
+- Class merging is **mandatory**: `('<base> ' ~ attributes.render('class'))|tailwind_merge`.
+- Always emit `data-slot="<recipe-name>"` on the root and `data-slot="<recipe-name>-<sub>"` on sub-component roots.
```
DcgRG: This is a ShadCN convention, do you think it should be extended to other kits?

.agents/skills/symfony-ux-toolkit-kit/SKILL.md
```diff
@@ -0,0 +1,346 @@
...
+
+```
+src/Toolkit/kits/<kit>/<recipe>/
+├── manifest.json
+├── examples/
+│   ├── Usage.html.twig                    # mandatory, minimal API showcase
+│   ├── Demo.html.twig                     # mandatory, rich preview (used on ux.symfony.com)
+│   └── <Variant Name>.html.twig           # one per upstream example, Title Case with spaces
+├── templates/components/
+│   ├── <Component>.html.twig              # root component
```
DcgRG, line 46: What do you think about placing the main component in the `<Component>` folder with `<Component>/index.html.twig` in order to group all the templates in the same place?

.agents/skills/symfony-ux-toolkit-kit/SKILL.md
```diff
@@ -0,0 +1,346 @@
...
+    └── <recipe>_controller.js
+```
+
+Sub-component file path `Component/SubName.html.twig` is consumed as `<twig:Component:SubName>`.
+
+---
+
+## `manifest.json`
+
+### Kit-level (`src/Toolkit/kits/<kit>/manifest.json`)
```
DcgRG, line 131: If a kit exists in multiple versions, specify the version `<kit>-<major_version_identifier>` like "flowbite-4"

DcgRG (commented):
For keyboard actions, use the stimulus convention with data-actions whenever possible.

Example: `data-action="keydown.enter-><recipe_controller>#<recipe_action>"`

.agents/skills/symfony-ux-toolkit-kit/SKILL.md
```diff
@@ -0,0 +1,346 @@
...
+
+---
+
+## Core Rules
+
+1. **One PR per recipe.** Never batch multiple recipes in a single PR. PR title:
+   `[Toolkit][<Kit>] Add <recipe> recipe` or `[Toolkit][<Kit>] Align <recipe> with <upstream> reference`.
+2. **Target `3.x`.** CHANGELOG entry goes under the active `3.x` section in `src/Toolkit/CHANGELOG.md`.
+3. **Visual + behavioral parity** with the upstream reference (Shadcn UI / Flowbite). Verify manually; attach a screenshot or video to the PR body for animated/interactive components.
+4. **Reuse all upstream examples.** Do not pick a subset. For Shadcn, mirror `apps/v4/examples/radix/<recipe>-*.tsx`.
```
seb-jean: Perhaps it would be better to say `https://github.com/shadcn-ui/ui/tree/main/apps/v4/examples/radix-<recipe>-*.tsx` instead of `apps/v4/examples/radix/<recipe>-*.tsx`.

seb-jean:
Regarding the example files, I saw that you suggested using the Shadcn/Flowbite examples and specifying a path, so that's good.

However, I don't see where you indicate using the component/subcomponent files. For Shadcn, it would be in `https://github.com/shadcn-ui/ui/blob/main/apps/v4/styles/radix-nova/ui/<recipe>.tsx`

Shouldn't Flowbite and Shadcn be distinguished? And therefore, separate files created?

Amoifr:
Thanks @Kocal for kicking this off — really nice base, especially the anti-patterns table and the `<recipe>_<role>_attrs` formalization 🙌

I've been maintaining a similar skill on my side (FR, Shadcn-focused), built up incrementally from the review feedback I got across recent recipe PRs. A lot overlaps, but there are a few areas where mine goes deeper because they bit me in real PRs — sharing in case it's useful to fold in:

### 1. Upstream sourcing — pick up @seb-jean's point

Neither of our skills points to the **component source** (`apps/v4/styles/radix-nova/ui/<recipe>.tsx`), only to the examples. The component file is what carries the canonical classes / structure / `data-*` surface — without it, you only see how examples *use* the component, not the component itself. I'd add it as the first source to read.

Also, listing examples is non-trivial — I have a concrete recipe for it:

```bash
# Find the MDX page that lists every example slug
gh api repos/shadcn-ui/ui/git/trees/main?recursive=1 --jq '.tree[].path' | grep '<recipe>'
# Then grab each .tsx referenced in the MDX:
# https://raw.githubusercontent.com/shadcn-ui/ui/refs/heads/main/apps/v4/examples/radix/<example>.tsx
```

### 2. Local visual testing — the dual-repo gotcha

The biggest pain when validating a recipe locally is that **`ux` and `ux.symfony.com` must be on matching branches**, otherwise the assetmap explodes:

> The asset "./vendor/symfony/ux-toolkit/kits/shadcn/<other>/assets/controllers/<other>_controller.js" cannot be found in any asset map paths.

Worth a section like:

```bash
cd ux             && git checkout feat/toolkit-shadcn-<recipe>
cd ux.symfony.com && git checkout docs/shadcn-<recipe>
symfony php bin/console tailwind:build
```

### 3. ux.symfony.com companion PR — what actually goes in it

The skill says "open a companion PR" but doesn't spell out what it contains. From experience:

- `templates/toolkit/docs/shadcn/<recipe>.md.twig` extending `_base_component.md.twig`
- One `{{ toolkit_code_example(...) }}` per example file shipped (otherwise they don't render — silent failure)
- If JS: register the Stimulus controller in `assets/toolkit-shadcn.js` + `importmap.php`
- Rebuild Tailwind: `symfony php bin/console tailwind:build`

### 4. Hover/focus components — don't go pure CSS

Worth flagging explicitly: for `hover-card`, `tooltip`, `dropdown-menu`, `menubar`, **don't try `group-hover` + `group-focus-within` + `tabindex=0`**. It looks elegant but breaks: clicking the trigger leaves the content open after mouseleave, and removing `group-focus-within` kills keyboard a11y. Always go straight to a Stimulus controller with `openDelay`/`closeDelay` values.

Minimal pattern:

```js
static values = {
    openDelay: { type: Number, default: 0 },
    closeDelay: { type: Number, default: 0 },
};
show() { /* clear timeouts, set data-state=open after openDelay */ }
hide() { /* clear timeouts, set data-state=closed after closeDelay */ }
```

### 5. Named groups for nested open-state

For components with a nested structure (`menubar` has Menu+Sub, `dropdown-menu` has Root+Sub), **`in-data-[state=open]:visible` matches any ancestor**, so opening the parent also opens every child at once. Use named groups instead:

```
.group/<recipe>-menu          (parent with data-controller)
.group-data-[state=open]/<recipe>-menu:visible  (parent's content)
.group/<recipe>-sub           (sub with its own data-controller)
.group-data-[state=open]/<recipe>-sub:visible   (sub's content)
```

### 6. Animation: grid-template-rows, never `hidden`

For expand/collapse (accordion, collapsible, content panels):

```twig
<div class="grid overflow-hidden transition-[grid-template-rows] duration-300 ease-out"
     style="grid-template-rows: {{ open ? '1fr' : '0fr' }};">
    <div class="min-h-0 min-w-0 overflow-hidden">{%- block content %}{% endblock -%}</div>
</div>
```

`hidden` (display:none) causes layout jumps — only `grid-template-rows: 0fr` + `overflow:hidden` gives a clean transition.

### 7. Snapshot gotchas

Two things reviewers consistently flag and that the skill doesn't cover:

- **Orphan snapshots after a rework.** When you rename or convert `<details>` to a Stimulus version, you often end up with stale `Demo.html__1.html` (no `.twig` suffix) sitting next to the new `Demo.html.twig__1.html`. They are never regenerated — `git rm` them.
- **Re-regenerate after rebase on `3.x`.** The snapshot formatter evolves upstream; running `UPDATE_SNAPSHOTS=true vendor/bin/phpunit --filter "<recipe>"` once more after the final rebase is what avoids "diff in snapshots" CI failures.

### 8. Trailing newlines on every shipped file

The skill mentions snapshots/lint but not this: **every shipped file must end with a newline** — `.html.twig`, `.json`, `.js`, `.css`, `.md`. Maintainers add them by hand at merge if missing, so worth doing it upfront. One-liner:

```bash
for f in <files>; do tail -c1 "$f" | od -c | grep -q '\\n' || printf '\n' >> "$f"; done
```

---

Happy to open a follow-up PR pulling these into your skill if you're open to it — or if you'd rather keep the skill lean for now, totally fair, just leaving them here for visibility 😊

Amoifr:
@seb-jean +1, fully agree on both points 👍

For **(1) component source files**: the examples (`apps/v4/examples/radix/<recipe>-*.tsx`) only show how a consumer *uses* the component — they don't carry the canonical class names, `data-*` surface, or sub-component structure. Those live in `apps/v4/styles/radix-nova/ui/<recipe>.tsx`, and that's what you actually need to mirror for visual + behavioral parity. Both files should be listed as inputs.

For **(2) Flowbite vs Shadcn**: I think the *conventions* (manifest, `html_cva`, `tailwind_merge`, `data-slot`, asChild `<recipe>_<role>_attrs`, Stimulus patterns…) are kit-agnostic and worth keeping in one skill. But the *upstream sources* are kit-specific:

- **Shadcn** → `apps/v4/styles/radix-nova/ui/<recipe>.tsx` (component) + `apps/v4/examples/radix/<recipe>-*.tsx` (examples)
- **Flowbite** → component & examples live in `flowbite-react/packages/ui/src/components/<Component>/` and on the docs site

So rather than splitting the skill, maybe a small per-kit "Upstream sources" section (a table or a switch on the kit name) does the job? Curious what @Kocal thinks.

Kocal:
I just applied your comments and compressed the SKILL with caveman, thanks everyone!

PR #3508, opened 2026-04-27
[LiveComponent] Fix dynamic template resolution when using "loading" attribute

xDeSwa:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

### Bug Description
When using a dynamic template in a `LiveComponent` (via `FromMethod`) along with a `loading` attribute (e.g., `loading="defer"` or `loading="lazy"`), the dynamic template path was being ignored and replaced by the default static template defined in the metadata.

### Root Cause
In `DeferLiveComponentSubscriber::onPreRender()`, the subscriber was fetching the template path directly from the component's metadata using `$event->getMetadata()->getTemplate()`. 

However, `PreRenderEvent` already resolves the dynamic template in its constructor and stores it. By calling the metadata directly, the subscriber was bypassing the resolved dynamic template and falling back to the static one.

### Solution
Updated `DeferLiveComponentSubscriber` to use `$event->getTemplate()` instead of `$event->getMetadata()->getTemplate()`. This ensures that if a dynamic template is resolved via a method, it is correctly passed to the deferred template variables.

This fix can also be applied to the previous `2.x version.`

### DIFF
--- a/src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
+++ b/src/LiveComponent/src/EventListener/DeferLiveComponentSubscriber.php
@@ -72,7 +72,7 @@ final class DeferLiveComponentSubscriber implements EventSubscriberInterface
             return;
         }
 
-        $componentTemplate = $event->getMetadata()->getTemplate();
+        $componentTemplate = $event->getTemplate();
         $event->setTemplate('@LiveComponent/deferred.html.twig');
 
         $variables = $event->getVariables();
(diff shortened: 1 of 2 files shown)


Kocal (commented):
Can you add a test? Is it only specific to `loading` attribute?

xDeSwa:
> Can you add a test? Is it only specific to `loading` attribute?

added test. yes, it is only specific about LiveComponent `loading` attribute.

Kocal (approved):

Kocal:
Thank you @xDeSwa.

PR #3510, opened 2026-05-01
 feat(toolkit): add MultiSelect component for shadcn kit

KylianCodes:
Adds a multi-select component that displays selected options as removable badges, backed by a Stimulus controller and a hidden select multiple for standard form submission.

Relates to #3233

| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | #3233
| License        | MIT

Adds a `MultiSelect` component to the shadcn kit. Selected options are displayed as removable badges inside the trigger. A hidden `<select multiple>` handles standard Symfony form submission.

**Usage**                                                                               
```twig
<twig:MultiSelect name="fruits" placeholder="Select fruits...">
    <twig:MultiSelect:Option value="apple" label="Apple" />
    <twig:MultiSelect:Option value="banana" label="Banana" />
</twig:MultiSelect>

### DIFF
--- a/src/Toolkit/CHANGELOG.md
+++ b/src/Toolkit/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 3.1
 
+- [Shadcn] Add `multi-select` recipe
 - [Flowbite] Add `avatar` recipe
 - [Flowbite] Add `dropdown` recipe
 - [Shadcn] Add `hover-card` recipe
--- a/src/Toolkit/kits/shadcn/multi-select/assets/controllers/multi_select_controller.js
+++ b/src/Toolkit/kits/shadcn/multi-select/assets/controllers/multi_select_controller.js
@@ -1,0 +1,109 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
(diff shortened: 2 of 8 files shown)


MrYamous:
AFAIK Shadcn isn't providing a MultiSelect component
I fully understand how useful it can be but I'm wondering how to integrate this kind of components

Kocal:
Closing this one. @MrYamous was right: I went through the full shadcn-ui/ui file tree and there's no multi-select component upstream, just `select` and `native-select`, plus a few `combobox-multiple` examples. The rule for the Toolkit's shadcn kit is parity with the upstream reference, so a component with nothing to check against would just drift over time with no way to keep it in sync.

On top of that, the recipe structure changed since you opened this in May: examples used to live in an `examples/` directory, they're now inline in each recipe's `README.md` as preview blocks, and that directory doesn't exist anymore in the Toolkit. So the docs part here would need a full rewrite regardless of the multi-select question.

Thanks for the work you put into this, it's appreciated. The door stays open though: a `combobox` recipe with multiple selection does have an upstream reference and would be very welcome if you (or anyone) want to pick that up instead.

PR #3512, opened 2026-05-01
[TwigComponent] Add `provide()` and `inject()` Twig functions

### LINKED ISSUE #2522: [TwigComponent] Add `aware` tag

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #2522 
| License        | MIT

Suggestion for #2522, with a behavior similar to Vue.js.

Some components from Toolkit kits could be improved with this feature (will do it in another PR)

### DIFF
--- a/src/TwigComponent/doc/index.rst
+++ b/src/TwigComponent/doc/index.rst
@@ -1327,6 +1327,76 @@ to resolve conflicts:
         {% block content %}{% endblock %}
     </div>
 
+Sharing State With ``provide`` / ``inject``
+-------------------------------------------
+
+When a component is made of several pieces, descendants often need values
+declared on the root. Forwarding those values through every level as props
+gets tedious fast. Use ``provide()`` in the parent to publish a value, and
+``inject()`` in any descendant to read it.
+
+In the parent, publish values with ``provide()``:
+
+.. code-block:: html+twig
+
+    {# templates/components/InputOtp.html.twig #}
+    {% props maxLength = 6 %}
+
+    {% do provide('inputOtp.maxLength', maxLength) %}
+
+    <div class="input-otp">
+        {% block content %}{% endblock %}
+    </div>
+
+In any descendant (at any depth), read them with ``inject()```:
+
+.. code-block:: html+twig
+
+    {# templates/components/InputOtp/Slot.html.twig #}
+    {% props input %}
+
+    {# Fallback to 4 if the parent doesn't provide a value for some reason #}
+    {% set maxLength = inject('inputOtp.maxLength', 4) %}
+
+    <input type="text" maxlength="1" data-index="{{ input }}" data-max-length="{{ maxLength }}" />
+
+Here, ``Slot`` reads ``maxLength`` from ``InputOtp`` even though ``Group``
+sits in between and does not forward anything:
+
+.. code-block:: html+twig
+
+    <twig:InputOtp maxLength="6">
+        <twig:InputOtp:Group>
+            {% for i in 0..5 %}
+                <twig:InputOtp:Slot input="{{ i + 1 }}" />
+            {% endfor %}
+        </twig:InputOtp:Group>
+    </twig:InputOtp>
+
+Keys are arbitrary strings; values can be any type, including ``null``.
+Prefix keys with the component name (``'inputOtp.maxLength'``,
+``'tabs.active'``) to avoid collisions across unrelated components.
+
+Rules:
+
+- Values flow top-down only. A child cannot push values back up.
+- ``inject()`` walks ancestors nearest-first, and skips the current
+  component. Read your own state through local variables.
+- Once a parent finishes rendering, its provides are dropped. Sibling
+  components never share state.
+- Calling ``provide()`` outside a component template throws.
+
+.. warning::
+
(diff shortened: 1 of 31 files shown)


seb-jean:
I will try to review it as soon as I have time, if possible.

src/TwigComponent/src/MountedComponent.php
```diff
@@ -74,4 +79,23 @@ public function getExtraMetadata(string $key): mixed
...
+
+    public function hasProvided(string $key): bool
+    {
+        return \array_key_exists($key, $this->providedContext);
+    }
+
+    public function getProvided(string $key): mixed
+    {
+        if (!$this->hasProvided($key)) {
+            throw new \InvalidArgumentException(\sprintf('No provided context for key "%s" found.', $key));
```
seb-jean: Is this exception reachable because `getProvided()` is always preceded by `hasProvided()`?

src/TwigComponent/doc/index.rst
```diff
@@ -1327,6 +1327,67 @@ to resolve conflicts:
...
+  parent finishes rendering, its provides are dropped, so two components
+  rendered side-by-side never share state.
+- ``inject`` skips the current component. A component reads its own state
+  through local variables, not through ``inject``.
+- ``provide(key, null)`` is not the same as "missing": ``inject(key, 'x')``
+  returns ``null`` here, not ``'x'``.
+- Namespace your keys (``'inputOtp.maxLength'``, ``'tabs.active'``) to avoid
+  collisions.
+- Calling ``provide()`` outside a component template throws.
+
```
seb-jean, line 1388: You don't mention the second argument of `inject()` in the documentation.

src/TwigComponent/doc/index.rst
```diff
@@ -1327,6 +1327,67 @@ to resolve conflicts:
...
+This complements :ref:`outerScope <embedded-components-outerScope>`, which
+only exposes the immediate parent's local context.
+
```
seb-jean, line 1399: You don't mention the "overwrite" behavior, but it's well tested :D.

seb-jean:
You had proposed two syntaxes (https://github.com/symfony/ux/issues/2522#issuecomment-2802225212):
- [Vue-style](https://vuejs.org/guide/components/provide-inject), with `provide()` and `inject()`
- [React-style](https://react.dev/learn/passing-data-deeply-with-context), with `<twig:ux:context>` and context()`

I had a preference for React's syntax :D.

What made you lean more toward Vue.js syntax rather than React?

---

I tested with this example:

- `templates/components/Input.html.twig`
- `templates/components/Fieldset/Description.html.twig`
- `templates/components/Fieldset/Field.html.twig`
- `templates/components/Fieldset/Label.html.twig`

```twig
<twig:Fieldset:Field disabled> 
    <twig:Fieldset:Label>Full name</twig:Fieldset:Label> 
    <twig:Input name="full_name" /> 
    <twig:Fieldset:Description>If you have a tiger, we'd like to know about it.</twig:Fieldset:Description>
</twig:Fieldset:Field>
```

This is interesting because the `Input` component is a separate component of `Fieldset`, not even a sub-component compared to `Label` and `Description`.

It works well, I get the `disabled` prop on each element using `provide()` / `inject()`.

---

The rendering of Twig functions on this page https://symfony.com/doc/current/reference/twig_reference.html is interesting. Perhaps the same could be done for `inject()` and `provide()`.

Kocal:
Thanks for the reviews :) 

> What made you lean more toward Vue.js syntax rather than React?

For two main reasons:
- The way you use one or more contexts in React is, in my opinion, an abomination. You have to nest components, and you quickly find yourself in [React Context Hell](https://medium.com/@ambrosekibet576/navigating-reacts-context-hell-f6f4fe5dc23a)
- More generally, Twig components are cool, but if we can avoid overusing them in favor of more "native" and more efficient solutions (in this case, two Twig functions), that's great

seb-jean:
Do we also need to modify the `.agents/skills/symfony-ux-toolkit-kit/SKILL.md` file to add these 2 Twig functions?

Kocal:
> Do we also need to modify the `.agents/skills/symfony-ux-toolkit-kit/SKILL.md` file to add these 2 Twig functions?

Yes, but I planned to do it in a dedicated PR

src/TwigComponent/doc/index.rst
Kocal: Done

src/TwigComponent/doc/index.rst
Kocal: Done

src/TwigComponent/src/MountedComponent.php
Kocal: It could if someone tries to manually call the method, like the `getExtraMetadata`

PR #3516, opened 2026-05-04
 Stop publishing extra package minor versions with no changes

fabpot:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | n/a
| License        | MIT

Everything is in the title :)

### DIFF
--- a/splitsh.json
+++ b/splitsh.json
@@ -25,5 +25,8 @@
         "ux-twig-component": "src/TwigComponent",
         "ux-typed": "src/Typed",
         "ux-vue": "src/Vue"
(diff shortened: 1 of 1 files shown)


Kocal:
The [release-on-npm workflow](https://github.com/symfony/ux/blob/3.x/.github/workflows/release-on-npm.yaml) must be updated, because it currently use the new tag from `symfony/ux` repository, and will always create a new release on NPM even if no tag is created for a specific UX package.

PR #3518, opened 2026-05-04
[TwigComponent] Fix PreLexer confusing `{#` inside output expressions with comments

### LINKED ISSUE #3514: [TwigComponent] Twig PreLexer gets confused by ICU MessageFormat plural syntax
I have run into an edge case where I’m using ICU messages in a translatable string, and the message has plural cases in it. [See here for a more complete documentation](https://simplelocalize.io/blog/posts/what-is-icu/), but here’s an example string: 

```
{count, plural, 
   one {# item}
   other {# items}
}
```

This trips up the Twig PreLexer and makes it look for an end-of-comment `#}` that it may or may not find, and ignore all `<twig:Blocks>` until it finds one.

Here's a minimal template as example:

```twig
There are {{ '{count, plural, one {# item} other {# items}}'|trans({'count': 42}) }}!

<twig:Test />
```

It outputs:

```html
There are {42, plural, one {# item} other {# items}}!

<twig:Test />#}
```

The ICU syntax isn’t processed in the raw string, but this doesn’t matter in practice since it’s only here as a placeholder for translators. Real strings from XLIFF files are handled correctly. Most importantly, the HTML tag is present in the output… Now, if I modify the template to add a “fake” comment:

```twig
There are {{ '{count, plural, one {# item} other {# items}}'|trans({'count': 42}) }}!
{# Comment #}
<twig:Test />
```

The component is now rendered:

```html
There are {42, plural, one {# item} other {# items}}!
TEST COMPONENT
```

I understand this is niche/edge case, and in any case thanks a lot for your hard work on this project and thanks for your time!

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #3514
| License       | MIT

The `TwigPreLexer` was treating any `{#` it encountered as the opening of a Twig comment, even when that `{#` actually lived inside a Twig output expression (e.g. inside a string literal passed to `|trans`). When the matching `#}` was missing — typically the case for ICU MessageFormat plural strings like `{count, plural, one {# item} other {# items}}` — the lexer consumed the rest of the template as a fake comment, swallowing any `<twig:…>` tags after it.

This PR adds a top-level handler for `{{ … }}` in the PreLexer's main loop. The new helper `consumeTwigExpressionContent()` walks the expression body with string-literal awareness (single quotes, double quotes, `\\`-escapes), so that `}}` or `{#` appearing inside a string don't end the expression early or trigger comment mode. The implicit `{% block content %}` injection still kicks in when the expression appears as content inside a component.

5 regression tests added covering:
- the original ICU plural example from the issue
- `}}` inside a string literal in `{{ … }}`
- `{#…#}` inside a double-quoted string
- escaped quotes inside the expression
- `{{ … }}` inside a component (default-block injection)

Marking as draft pending @Kocal's input on the scope question raised in #3514: should this same protection be extended to `{% … %}` (where `{% set foo = '{# bar #}' %}` has the same latent issue), or kept focused on `{{ … }}` ?

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add support for dynamic component names in the HTML syntax using `<twig:component is="myComponent" ...`
 - Capture `## <description>` documentation comments written above props inside `{% props %}` (Twig 3.29+), exposed per prop via `PropsNode::getPropDocumentation()`
 - Fix `{% props %}` treating a prop explicitly passed as `null` as missing: a required prop no longer throws, and a prop declaring a default value now keeps the `null` it was given
+- Fix the pre-lexer treating a `{#` inside a `{{ ... }}` expression as the start of a Twig comment, which silently stopped every component after it from being rendered
 - Fix HTML/`{% component %}` syntax failing when a prop uses the null-safe operator (`?.`)
 
 ## 3.4.0
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -70,6 +70,31 @@ class TwigPreLexer
                 }
             }
 
+            // ignore content inside twig output expressions, see #3514
+            // (so that e.g. `{{ '{# foo #}' }}` doesn't trip the comment handler above)
+            if ($this->check('{{')) {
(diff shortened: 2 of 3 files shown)


Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3528, opened 2026-05-05
[Toolkit][Shadcn] Rework `toggle-group` recipe to use `provide()`/`inject()`

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

Following #3512

### DIFF
--- a/src/Toolkit/kits/shadcn/toggle-group/templates/components/ToggleGroup/Item.html.twig
+++ b/src/Toolkit/kits/shadcn/toggle-group/templates/components/ToggleGroup/Item.html.twig
@@ -1,8 +1,8 @@
 {# @prop pressed boolean Whether the item is initially pressed. Defaults to `false` #}
 {# @block content The toggle item label and/or icon #}
 {%- props pressed = false -%}
-{%- set _item_variant = _toggle_group_variant ?? 'default' -%}
-{%- set _item_size = _toggle_group_size ?? 'default' -%}
+{%- set _toggleGroup_variant = inject('toggleGroup.variant', 'default') -%}
+{%- set _toggleGroup_size = inject('toggleGroup.size', 'default') -%}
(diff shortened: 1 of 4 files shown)


src/Toolkit/kits/shadcn/toggle-group/templates/components/ToggleGroup/Item.html.twig
```diff
@@ -1,8 +1,8 @@
 {# @prop pressed boolean Whether the item is initially pressed. Defaults to `false` #}
 {# @block content The toggle item label and/or icon #}
 {%- props pressed = false -%}
-{%- set _item_variant = _toggle_group_variant ?? 'default' -%}
-{%- set _item_size = _toggle_group_size ?? 'default' -%}
+{%- set _toggleGroup_variant = inject('toggleGroup.variant', 'default') -%}
```
seb-jean, line 4: Twig variables should be in snake case.
https://twig.symfony.com/doc/3.x/coding_standards.html

Kocal: You're right, but I did it on purpose in order to reduce the mental load. 

This line can be interpreted like that:
```twig
{% set _<component>_<prop> = inject('<component>.<prop>', '...') %}
```

Here, `<prop>` represents the props name from a parent component, and props are named in **camelCase**. Having `<prop>` in snake_case would decrease the DX when searching/replacing (yeah we have IA, but that's overkill and not free).
Finally, to prevent mixing different case namings in `_<component>_<prop>` or `'<component>.<prop>'`, `<component>` is also in **camelCase**

Issue #3509, opened 2026-04-28
Documentation mentions kit which doesn't exist


tomazartack:
URL: https://ux.symfony.com/toolkit/kits/flowbite-4/components/modal

To install the modal, I'm supposed to run:

bin/console ux:install modal --kit flowbite-4

However, it fails with an error message:

10:08:44 WARNING   [app] Command `ux:install` exited with status code 1
In LocalRegistry.php line 48:
  Kit "flowbite-4" does not exist.  
ux:install [-k|--kit [KIT]] [-d|--destination [DESTINATION]] [-f|--force] [--] [<recipe>]

Kocal:
What version of UX Toolkit do you use?

tomazartack:
Thank you the version was the problem. I had to upgrade from 2.32 to 2.35. The documentation should mention from which version on, this toolkit is available

Kocal:
You're right, that could be a nice improvement. 

I think we could have a new key `added-in-version` (or something similar) in the `manifest.json` for each Kit recipes. Then, we could use it in UX website to display which version of UX Toolkit should be used. 

Anyone up for a PR? The _hardest_ (maybe, the less funny) part is about identifying the versions in which the recipes were released. 

If anyone has a better idea, let us know! :D

MrYamous:
> You're right, that could be a nice improvement.
> 
> I think we could have a new key `added-in-version` (or something similar) in the `manifest.json` for each Kit recipes. Then, we could use it in UX website to display which version of UX Toolkit should be used.
> 
> Anyone up for a PR? The _hardest_ (maybe, the less funny) part is about identifying the versions in which the recipes were released.
> 
> If anyone has a better idea, let us know! :D

OMW, I'll give a first shot with `versionadded`, I'm used to it with framework documentation ;)

Amoifr:
Hi @Kocal! 👋

Took a swing at the foundation — opened #3532. It adds the optional `added-in-version` property to the Recipe manifest schema and exposes it on `RecipeManifest::$addedInVersion`. Existing manifests are left untouched so the change is purely additive and they keep validating.

The "less funny part" (populating the 51 existing recipes with their actual introduction version) is left to follow-up, since several recipes were restructured (e.g. #3107) and git creation dates don't always match the release version. I'd happily tackle that once this lands, but figured it was cleaner to split rather than block the schema/parsing piece on the archaeology.

Let me know if you'd rather see a different split.

Kocal:
[...]

PR #3107, opened 2025-09-27
[Toolkit] Rework recipe name

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

This PR introduce a new field "name" for Recipe (based on Recipe's folder name), and change the case of Recipe folders from PascalCase to kebab-case, in order to make things more friendly.

With this new field:
- `bin/console ux:install "Alert Dialog" --kit shadcn` becomes `bin/console ux:install alert-dialog --kit shadcn`
- `ux.symfony.com/toolkit/kits/shadcn/components/Alert%20Dialog` becomes `ux.symfony.com/toolkit/kits/shadcn/components/alert-dialog`

PR #3531, opened 2026-05-10
[Toolkit] Add `version-added` key in toolkit manifest

MrYamous:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | Fix #3509 
| License        | MIT

Allow a new `version-added` key in Toolkit manifest to add it in documentation
I make it `nullable` cause I think for components which exists for a long time this information is less important. And it allow us to not add it for all components right now

PR #3532, opened 2026-05-11
[Toolkit] Add `added-in-version` property to Recipe manifests

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | Refs #3509
| License       | MIT

Following @Kocal's suggestion in #3509, this PR introduces an optional `added-in-version` property in the Recipe manifest schema. It exposes the value as `RecipeManifest::$addedInVersion`, ready to be consumed by ux.symfony.com to advertise the minimum UX Toolkit version required for each recipe.

## Changes

- `schema-kit-recipe-v1.json`: declare the optional `added-in-version` property.
- `Recipe/RecipeManifest.php`: new constructor argument + parsing with validation (`null` or non-empty string).
- `Tests/Recipe/RecipeManifestTest.php`: three new tests (with value, with empty string, with non-string).

## Scope

Intentionally minimal. **Existing manifests are not touched** — the field is purely additive and they remain valid. The follow-up effort (populating `added-in-version` on the 51 existing recipes) is more delicate, since several recipes were renamed/restructured (e.g. #3107) and the introduction version doesn't always match the file's git creation date. That part is better driven by the maintainers who know the history, or split into smaller PRs per kit.

cc @Kocal — happy to take a stab at the manifest-population follow-up too once this foundation lands, if you'd like.

### DIFF
--- a/src/Toolkit/schema-kit-recipe-v1.json
+++ b/src/Toolkit/schema-kit-recipe-v1.json
@@ -18,6 +18,10 @@
             "description": "A brief description of the Recipe",
             "type": "string"
         },
+        "added-in-version": {
+            "type": "string",
+            "description": "The first UX Toolkit version in which this Recipe was made available, e.g., '2.35'"
+        },
         "copy-files": {
             "description": "Relative files or directories to copy into the project",
             "type": "object"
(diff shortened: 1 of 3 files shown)


Kocal:
Thanks for your contribution, but @MrYamous already opened https://github.com/symfony/ux/pull/3531 ~16 hours ago and, naturally, I'd prefer to merge his PR (first come, first served 😄)

Thank you anyway!

PR #3533, opened 2026-05-11
[LiveComponent] Return StimulusAttributes from live_action() so it can be spread in a Twig component tag

### LINKED ISSUE #3413: [TwigComponent] How to use `live_action()` twig function in Twig Components tag?
In a Symfony Live Component, I have something like this (simplified example):

```twig
<div {{ attributes }}>
    {% set parent = this %} {# I do not know if there is a better alternative in nested twig components #}

    <table>
        <thead>
            <tr>
                {# twig:Table:Head comes from https://ux.symfony.com/toolkit/kits/shadcn/components/table, it's an anonymous Twig Component #}
                <twig:Table:Head
                    data-action="click->live#action"
                    {{ live_action('sort', {sortBy: 'id'}) }}
                >
                    {{ 'ID'|trans }}
                </twig:Table:Head>
            </tr>
        </thead>
    </table>
</div>
```

But I have this error:

> Expected attribute name when parsing the "<twig:Table:Head" syntax in components/FactorTypeSearch.html.twig at line 9.

The error comes from `{{ live_action('sort', {sortBy: 'id'}) }}`.

I know I can use the attributes syntax:

```html
<twig:Table:Head
    data-action="click->live#action"
    data-live-action-param="sort"
    data-live-sort-by-param="id"
>
    {{ 'ID'|trans }}
</twig:Table:Head>
```

But I want to use `live_action` because the syntax is cleaner. And since all of this comes from Symfony UX, we expect it to work naturally.

Is there a way to use `{{ live_action('sort', {sortBy: 'id'}) }}`? or will it be possible in the future?

Versions:
- Symfony: 8.0.7
- PHP: 8.5.3
- UX Live Component: 2.34.0
- UX Turbo: 2.34.0
- Twig: 3.24.0

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #3413
| License       | MIT

Today, `live_action()` casts its return value to a `string`, which prevents it from being spread into a Twig component tag:

```twig
{# ❌ doesn't work: the spread receives a plain string #}
<twig:Table:Head {{ ...live_action('sort', {sortBy: 'id'}) }}>
```

`stimulus_action()` already returns a `StimulusAttributes` object (which is both `\Stringable` and `\IteratorAggregate`), so the same call written against it does work. The asymmetry was [pointed out by @squrious](https://github.com/symfony/ux/issues/3413#issuecomment-4379725306) in the issue discussion.

This PR aligns `live_action()` with `stimulus_action()` by returning the same `StimulusAttributes` object directly. Existing usages remain unaffected because Twig calls `__toString()` automatically when rendering the value (e.g. `<button {{ live_action('save') }}>`), and the helper is marked `@internal` on the PHP side.

The existing `LiveComponentRuntimeTest::testGetLiveAction` is adapted to cast the value to string for the equality assertion and gains a new `assertInstanceOf` to lock the return type.

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 3.1
 
 - Use `aria-busy` attribute during component re-render
+- Return a `StimulusAttributes` object from the `live_action()` Twig function so it can be spread into a Twig component tag (`<twig:Foo {{ ...live_action('bar') }}>`), aligning its behavior with `stimulus_action()`.
 
 ## 3.0.0
 
--- a/src/LiveComponent/src/Twig/LiveComponentRuntime.php
+++ b/src/LiveComponent/src/Twig/LiveComponentRuntime.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\LiveComponent\Twig;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\UX\LiveComponent\LiveComponentHydrator;
 use Symfony\UX\LiveComponent\Metadata\LiveComponentMetadataFactory;
+use Symfony\UX\StimulusBundle\Dto\StimulusAttributes;
 use Symfony\UX\StimulusBundle\Helper\StimulusHelper;
 use Symfony\UX\TwigComponent\ComponentFactory;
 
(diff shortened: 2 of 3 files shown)


Kocal:
For 3.x, 2.x is not maintained anymore

Amoifr:
Done, rebased onto `3.x` and re-targeted the PR. Thanks @Kocal! 🙏

smnandre:
Will the following code continue to work identically?

```twig
<tag {{ 'data-controller="foo bar" ' ~ live_action(...) }}> 
```

Amoifr:
Yes, identically. 👍

Twig's `~` operator stringifies its operands, and `StimulusAttributes` is `\Stringable`. The previous code already did `return (string) $attributes` right before returning, so `__toString()` produces the exact same string that was returned before. Concatenation (`~`), interpolation and direct rendering (`{{ live_action(...) }}`) are all unaffected; only the spread case (`{{ ...live_action(...) }}`), which used to fail, now works.

Happy to add a test asserting the concatenation output if you'd like it locked down. 🙂

PR #3537, opened 2026-05-12
[LiveComponent] Rebuild Component on reconnect when props changed in between

### LINKED ISSUE #3424: [UX Live Component] Bug: Stale ValueStore after component reconnect (disconnect → connect without initialize)

2.34.0

### Description

  When a live component is disconnected and later reconnected to the **same DOM element** (Stimulus reuses the controller instance), the `ValueStore` retains stale state from the previous component
  lifecycle. This causes `Invalid model name` errors because the `ValueStore` props no longer match the server-rendered HTML.

  The root cause: Stimulus calls `initialize()` only once per controller instance. On reconnect, only `connect()` is called — but `connect()` does not call `createComponent()`, so the old
  `Component` (and its `ValueStore`) is reused with outdated props.

### How to reproduce

 **Step 1 — Initial render**

  Stimulus calls `initialize()` → `createComponent()` → `connect()` on `#child_component`.
  `ValueStore` is initialized with props containing form name `abc`.

  ```html
  <div id="parent_component" data-controller="live">
      <div id="child_component" data-controller="live">
          <form name="abc"></form>
      </div>
  </div>
```

  Step 2 — Parent action removes child component (div is reused)

  Stimulus calls disconnect() on #child_component. The controller instance is kept alive by Stimulus (tied to the DOM element reference).

  ```html
  <div id="parent_component" data-controller="live">
      <div">new content without live controller</div>
  </div>
```

  Step 3 — Parent action re-adds child component (div is reused, form name changed)

  Stimulus calls connect() (NOT initialize()) on #child_component — the previous controller instance is reused. createComponent() is never called, so the old Component with old ValueStore remains
  active.

  ```html
  <div id="parent_component" data-controller="live">
      <div id="child_component" data-controller="live">
          <form name="xyz"></form>
      </div>
  </div>
```

 Expected behavior

  The ValueStore should reflect the current server-rendered props (form name="xyz"). User input should work without errors.

  Actual behavior

  The ValueStore still holds props from step 1 (form name="abc"). When the user types into an input field, Component.set() is called with the new model name (based on xyz), but ValueStore.has()
  returns false because it only knows about abc:

  Uncaught Error: Invalid model name "xyz.content".
      at Component.set (live_controller.js:1926:1)
      at extended.updateModelFromElementEvent (live_controller.js:3131:1)
      at extended.handleInputEvent (live_controller.js:3045:1)

### Possible Solution

  In the live controller's connect() method, detect when props have changed since the Component was created, and recreate it:

```ts
  connect() {
      const currentProps = this.propsValue;
      const storedProps = this.component.valueStore.getOriginalProps();
      if (JSON.stringify(currentProps) !== JSON.stringify(storedProps)) {
          this.createComponent();
      }
      this.connectComponent();
      this.mutationObserver.observe(this.element, { attributes: true });
  }
```

### Additional Context

_No response_

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3424
| License        | MIT

## Summary

When a parent live component action removes and re-adds a child live component on the **same DOM element**, Stimulus keeps the controller instance alive — it only calls `disconnect()` then `connect()` again, not `initialize()`. Because `createComponent()` only runs in `initialize()`, the `Component` (and its `ValueStore`) keep the **stale props** from before the parent action, even though the server-rendered HTML now exposes a fresh set.

The next user input then throws `Uncaught Error: Invalid model name "xyz.content"` — the model name has changed, but `ValueStore.has()` still looks up against the old props.

### Fix

In `connect()`, compare the current `propsValue` (which Stimulus has refreshed from the live `data-live-props-value` attribute) against `component.valueStore.getOriginalProps()`. When they differ, rebuild the `Component`.

- First `connect()` after `initialize()`: props match by construction → no rebuild.
- Plain disconnect/reconnect without any prop change: props match → no rebuild.
- Parent morph that swaps props on the same element: props diverge → rebuild.

The diagnosis and proposed fix are from @Pechynho in #3424 (confirmed working by another reporter in the thread).

## Test plan

- [x] New unit test `rebuilds the Component on reconnect when props changed in between` exercises the bug scenario by calling `controller.disconnect()` / mutating `propsValue` / `controller.connect()`, and asserts the `Component` instance has been swapped and the new `ValueStore` reflects the fresh props.
- [x] New unit test `keeps the existing Component on reconnect when props are unchanged` asserts that a no-op reconnect cycle does **not** recreate the `Component`.
- [x] Full LiveComponent controller test suite still green (109 tests).
- [x] `oxlint` + `oxfmt --check` clean.

Fixes #3424

### DIFF
--- a/src/LiveComponent/assets/src/live_controller.ts
+++ b/src/LiveComponent/assets/src/live_controller.ts
@@ -93,6 +93,16 @@ export default class LiveControllerDefault extends Controller<HTMLElement> imple
     }
 
     connect() {
+        // Stimulus only calls initialize() once per controller instance. When
+        // the same DOM element is disconnected then reconnected (e.g. a parent
+        // morphs it back in with new props), connect() is called again on the
+        // existing controller. If the props now differ from what the Component
+        // was built with, the ValueStore is stale and would reject any model
+        // matching the new server-rendered HTML — so rebuild the Component.
+        if (JSON.stringify(this.propsValue) !== JSON.stringify(this.component.valueStore.getOriginalProps())) {
+            this.createComponent();
+        }
+
         this.connectComponent();
 
         this.mutationObserver.observe(this.element, {
--- a/src/LiveComponent/assets/test/unit/controller/basic.test.ts
+++ b/src/LiveComponent/assets/test/unit/controller/basic.test.ts
@@ -12,7 +12,7 @@ import Component from '../../../src/Component';
 import { findComponents } from '../../../src/ComponentRegistry';
 import { htmlToElement } from '../../../src/dom_utils';
 import { getComponent } from '../../../src/live_controller';
-import { createTest, initComponent, shutdownTests, startStimulus } from '../../tools';
+import { createTest, getStimulusApplication, initComponent, shutdownTests, startStimulus } from '../../tools';
 
 describe('LiveController Basic Tests', () => {
     afterEach(() => {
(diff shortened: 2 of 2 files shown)


Kocal:
For 3.x, 2.x is not maintained anymore, and please follow the PR template (it's not the first time 😬)

Amoifr:
Sorry for the slip @Kocal — rebased onto `3.x` and prepended the Q/A table to the body. I've also saved this as a checklist note on my side so the next PR opens with both right from the start. 🙏

Amoifr:
@Kocal in my defense — I catch on quick, you just have to explain things to me at length 😅

PR #3550, opened 2026-05-19
[LiveComponent] Add push history state support to LiveComponent

### LINKED ISSUE #3149: [LiveComponent] URL linked LiveProps don't seem to honour the browser back button
We have a live component that maps it's #[LiveProp] to a URL parameter, which we use to determine which child components to display on the page.

This works find, but when we use the browser back button to navigate back to the previous "page", it doesn't navigate back, and pressing a second time returns to the page prior to the previous one.

Is this expected bahaviour, or is this a bug?  I'm happy to create a reproducer if this is not expected behaviour.

Thanks.

ajgarlag:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes
| Issues         | Fix #3149
| License        | MIT

This PR adds support for pushing browser history entries when a LiveComponent updates URL-bound LiveProp values.

Until now, URL updates were always done with `history.replaceState()`, which overwrote the current entry. With this change, components can opt in to `history.pushState()` so users can navigate previous component states via the browser back button (for example, previous pages in pagination or prior filter states in search UIs).

Included changes:
- Added a new `pushHistoryState` parameter to `#[AsLiveComponent(...)]` (default: false).
- Exposed this option in component metadata as `push_history_state`.
- Added `LiveComponentMetadata::hasPushHistoryStateEnabled()` to read the flag.
- Updated LiveUrlSubscriber to include a new response header:
  - `X-Live-Url-Push-History-State: 1|0`
- Updated frontend `BackendResponse` to parse the new header.
- Updated frontend URL update logic to choose:
  - `history.pushState(...)` when enabled
  - `history.replaceState(...)` otherwise (current behavior preserved)

### DIFF
--- a/src/LiveComponent/CHANGELOG.md
+++ b/src/LiveComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 3.1
 
 - Use `aria-busy` attribute during component re-render
+- Add a new `pushHistoryState` parameter to the `AsLiveComponent` attribute, which allows pushing a new history state on URL change after a re-render.
 
 ## 3.0.0
 
--- a/src/LiveComponent/assets/dist/live_controller.d.ts
+++ b/src/LiveComponent/assets/dist/live_controller.d.ts
@@ -3,9 +3,11 @@ declare class export_default$2 {
   response: Response;
   private body;
   private liveUrl;
+  private pushHistoryState;
   constructor(response: Response);
   getBody(): Promise<string>;
   getLiveUrl(): string | null;
+  hasPushHistoryStateEnabled(): boolean;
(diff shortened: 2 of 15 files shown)


xDeSwa:
I hope it also fixes the bug in back and forward (hotwire turbo) actions in the browser.

smnandre (commented):
I'm not really eager to use and promote push state now navigation API is [baseline](https://web.dev/blog/baseline-navigation-api).

Could we maybe use X-Live-Url-Navigation === "navigate"

PR #3559, opened 2026-05-20
[Toolkit][Shadcn] Align `item` with shadcn reference

seb-jean:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | Part of https://github.com/symfony/ux/issues/3233
| License        | MIT

### DIFF
--- a/src/Toolkit/kits/shadcn/item/templates/components/Item.html.twig
+++ b/src/Toolkit/kits/shadcn/item/templates/components/Item.html.twig
@@ -1,19 +1,20 @@
 {# @prop variant 'default'|'outline'|'muted' The visual style variant. Defaults to `default` #}
-{# @prop size 'default'|'sm' The item size. Defaults to `default` #}
+{# @prop size 'default'|'sm'|'xs' The item size. Defaults to `default` #}
 {# @prop as 'div' The HTML tag to render. Defaults to `div` #}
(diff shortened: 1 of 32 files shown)


src/Toolkit/kits/shadcn/item/templates/components/Item.html.twig
```diff
@@ -1,19 +1,20 @@
...
+            default: 'border-transparent',
             outline: 'border-border',
-            muted: 'bg-muted/50',
+            muted: 'border-transparent bg-muted/50',
         },
         size: {
-            default: 'p-4 gap-4',
-            sm: 'py-3 px-4 gap-2.5',
+            default: 'gap-2.5 px-3 py-2.5',
+            sm: 'gap-2.5 px-3 py-2.5',
```
Kocal, line 16: There is no difference between default and sm sizes, is it normal?

seb-jean: Good point, there is no difference on Shadcn regarding `default` and `sm`.

Kocal (approved):

Kocal:
Thank you @seb-jean.

PR #3565, opened 2026-05-20
[Autocomplete] Use `hash_equals()` to compare the `extra_options` checksum

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

`validateChecksum() `compared the client-supplied checksum against the
server-computed HMAC with `!==`, which short-circuits on the first
differing byte and is therefore not constant-time (CWE-208).

The checksum guards the integrity of `extra_options`, which the
controller feeds into the autocompleter as form-creation options, so the
comparison should be timing-safe. Switch to hash_equals() with the
known value as the first argument. LiveComponent already does this in
LiveComponentHydrator::verifyChecksum() — this aligns Autocomplete with
it.

### DIFF
--- a/src/Autocomplete/src/Controller/EntityAutocompleteController.php
+++ b/src/Autocomplete/src/Controller/EntityAutocompleteController.php
@@ -110,7 +110,7 @@ final class EntityAutocompleteController
             \ARRAY_FILTER_USE_KEY,
         );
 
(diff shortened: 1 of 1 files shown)


Kocal:
Thank you @Amoifr.

PR #3588, opened 2026-05-22
[Toolkit][Shadcn] Align `typography` with shadcn reference

seb-jean:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | Part of https://github.com/symfony/ux/issues/3233
| License        | MIT


src/Toolkit/kits/shadcn/typography/examples/h1.html.twig
Kocal, line 1: I think there is an issue here and with some other examples, why do we have `H1`, `H2`, `H3` ... components, if we do not use them in the examples?

Kocal (changes requested):

src/Toolkit/kits/shadcn/typography/examples/h1.html.twig
seb-jean: In React, Typography in `shadcn/ui` isn't actually a component.
There's nothing to install, no `button.tsx`, no `npx shadcn add typography`, no import. The docs page just gives you a catalog of Tailwind utility classes to copy directly onto plain HTML tags.

That's exactly why shadcn says "We do not ship any typography styles by default." Unlike Button (a real file with variants and an API) Typography is just a set of class recipes you apply by hand.

There are no components to create here, just examples.
The page needs to be reviewed.

src/Toolkit/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 - [Shadcn] Align `textarea` with shadcn reference
 - [Shadcn] Align `toggle` with shadcn reference
 - [Shadcn] Align `tooltip` with shadcn reference
+- [Shadcn] Align `typography` with shadcn reference
```
Kocal: For 3.2.0

Kocal (changes requested):

seb-jean:
I made the corrections and no files appear in templates/components/ anymore - `Typography` is not an installable component, it's just a collection of Tailwind utility class examples, unlike other components that ship actual templates.

Kocal:
Hey!

By curiosity I just checked https://ui.shadcn.com/docs/components/base/typography but it redirects to https://ui.shadcn.com/docs/typeset. 

Given this, and following https://github.com/symfony/ux/pull/3588#discussion_r3322577855, I think the best thing to do here is to remove this recipe, I feel like there's no point in keeping it, at least not in this condition.

I'm closing this PR and will re-open a new one to remove the recipe.

Thanks anyway!

PR #1247, opened 2023-11-03
[TwigComponents] Fixing inconsistency with how {% component rendered vs {{ component()

weaverryan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #1205 (?) and maybe others?
| License       | MIT

tl;dr `{% component` was rendered slightly differently than `{{ component`. Additionally, for `{% component`, the template was determined too early, which didn't allow for the template to be overridden via an event.

**Longer Explanation**

This cleans up `ComponentTokenParser`, removing some unnecessary parts. We now always set the "embedded" component parent to a magic `__parent__` variable. This allows us to delay the determination of the component template until runtime, after the `PreRenderEvent` has been called.

TODO:

* [ ] Bring in tests from #1236 and #1237 and check other tests
* [ ] Check #1205 to verify this fixes
* [ ] Check debug:twig-component command for any changes needed

Cheers!

PR #3600, opened 2026-05-27
[TwigComponent] Expose pre/embedded render methods on `ComponentRendererInterface`

treztreiz:
| Q              | A   |
|----------------|-----|
| Bug fix?       | yes |
| New feature?   | yes |
| Deprecations?  | no  |
| Documentation? | no  |
| Issues         | n/a |
| License        | MIT |

## Problem

`ComponentRuntime` depends on the concrete `ComponentRenderer` class (which is `final`):

```php
public function __construct(
    private readonly ComponentRenderer $renderer,
    // ...
) {}
```

`ComponentRendererInterface` exists and is already aliased to the renderer service in the DI container, but it only declares `createAndRender()`. The three other methods actually called by `ComponentRuntime` — `preCreateForRender()`, `startEmbeddedComponentRender()`, and `finishEmbeddedComponentRender()` — are missing from the interface.

Projects that need to intercept or augment the render pipeline (e.g. for server-side pre-loading of component instances) currently have to either fork `ComponentRuntime` and replace the service via a compiler pass, or use `\ReflectionProperty` to access private state on `ComponentRenderer`. Neither is acceptable for production code that needs to survive upstream upgrades.

## Solution

- Add the three missing methods to `ComponentRendererInterface`
- Switch `ComponentRuntime::$renderer` from `ComponentRenderer` to `ComponentRendererInterface`

No DI change is needed: the container already injects the renderer via its interface alias (`ComponentRendererInterface::class → ux.twig_component.component_renderer`).

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 3.2.0
+
+- Add `ComponentRendererInterface::preCreateForRender()`, `ComponentRendererInterface::startEmbeddedComponentRender()`, and `ComponentRendererInterface::finishEmbeddedComponentRender()` methods
+
 ## 3.1.0
 
 - Add `provide()` and `inject()` Twig functions to share state from a parent
--- a/src/TwigComponent/src/ComponentRendererInterface.php
+++ b/src/TwigComponent/src/ComponentRendererInterface.php
@@ -11,6 +11,13 @@
 
 namespace Symfony\UX\TwigComponent;
 
+use Symfony\UX\TwigComponent\Event\PreRenderEvent;
+
+/**
+ * @method ?string        preCreateForRender(string $name, array $props = [])
+ * @method PreRenderEvent startEmbeddedComponentRender(string $name, array $props, array $context, string $hostTemplateName, int $index)
+ * @method void           finishEmbeddedComponentRender()
+ */
 interface ComponentRendererInterface
 {
     /**
--- a/src/TwigComponent/src/Twig/ComponentRuntime.php
+++ b/src/TwigComponent/src/Twig/ComponentRuntime.php
@@ -12,7 +12,7 @@
 namespace Symfony\UX\TwigComponent\Twig;
 
 use Symfony\Component\DependencyInjection\ServiceLocator;
-use Symfony\UX\TwigComponent\ComponentRenderer;
+use Symfony\UX\TwigComponent\ComponentRendererInterface;
 use Symfony\UX\TwigComponent\ComponentStack;
 use Symfony\UX\TwigComponent\Event\PreRenderEvent;
 
@@ -25,7 +25,7 @@ use Symfony\UX\TwigComponent\Event\PreRenderEvent;
 final class ComponentRuntime
 {
     public function __construct(
-        private readonly ComponentRenderer $renderer,
+        private readonly ComponentRendererInterface $renderer,
         private readonly ServiceLocator $renderers,
         private readonly ComponentStack $componentStack,
     ) {


src/TwigComponent/src/ComponentRendererInterface.php
```diff
@@ -11,10 +11,25 @@
...
      */
     public function createAndRender(string $name, array $props = []): string;
+
+    /**
+     * @param array<string, mixed> $props
+     * @param array<string, mixed> $context
+     */
+    public function startEmbeddedComponentRender(string $name, array $props, array $context, string $hostTemplateName, int $index): PreRenderEvent;
+
+    public function finishEmbeddedComponentRender(): void;
```
Kocal: Adding these new methods on a non `@internal` and already released interface is a BC break, see https://symfony.com/doc/current/contributing/code/bc.html#changing-interfaces

IIRC, methods must be added in PHPDoc with `@method` annotations in 3.x branch (the current maintained branch), and in the future in 4.x branch, migrate these `@method` annotations to real methods declaration. So we are not breaking user's code.

Friendly ping @kbond @smnandre, do we really want to expose these methods? I feel like it was done on purpose, but I'm not sure 🤔

Kocal (changes requested):

src/TwigComponent/src/ComponentRendererInterface.php
kbond: Yes, using `@method` would be needed.

Could be these extra methods were added without thinking about the interface. I'm digging a bit to find out.

kbond:
@treztreiz this makes sense to me, I'm just going to dig a bit see if I can understand why they weren't added to the interface originally.

Why did you not add `render()` method?

kbond:
Hmm, `render()` is only used in `ComponentRenderer`. Since the class is internal, we should make this private.

kbond:
- `preCreateForRender()` was added in #913. It was a pretty complex PR. I don't see any discussion about the interface - maybe just missed?
- `finishEmbeddedComponentRender()` was added in #1093 - no discussion on the interface
- `startEmbeddedComponentRender()` was added in #317 but it was called `embeddedContext()`. The interface didn't exist then. It was refactored and renamed in #1247.

I honestly think updating the interface was just missed - or at least the discussion about it.

treztreiz:
@kbond @Kocal  Thanks for the feedback! Updated to use `@method` annotations to avoid the BC break. And good catch on `render() `, agreed it should stay off the interface since it's only used internally within `ComponentRenderer`.

The php (8.5, stable) failure is a transient CI infrastructure issue: a corrupted twig/twig zip in the GitHub Actions cache caused LiveComponent's composer update to fail. All other packages including TwigComponent pass. Re-running the job should resolve it.

kbond (changes requested):
Looks good, would you mind marking render as private in this PR?

kbond (approved):
👍

treztreiz:
@kbond `LiveComponentSubscriber::createResponse()` (line 339) calls `$renderer->render($mounted)` directly on the concrete class, so making it private breaks LiveComponent. It would need a separate, larger refactor of `LiveComponentSubscriber` to stop reaching into ComponentRenderer's internals, which seems out of scope for this PR ☺️

kbond:
Hah! Good catch. For sure, we need to look into that at some point.

kbond (dismissed):

Kocal (dismissed):

treztreiz:
All remaining CI failures are pre-existing on 3.x itself and not introduced by this PR. Also resolved the changelog conflict, seems this is ready to merge! ☺️ @kbond @Kocal

src/TwigComponent/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - Add `provide()` and `inject()` Twig functions to share state from a parent
   component down to deeply-nested descendants without prop drilling
 - Include the attribute name in the error message thrown when an attribute value is `null`
+- Add `preCreateForRender()`, `startEmbeddedComponentRender()`, and `finishEmbeddedComponentRender()`
+  on `ComponentRendererInterface` via `@method` annotations
+- Use `ComponentRendererInterface` in `ComponentRuntime` instead of the concrete `ComponentRenderer` class
```
Kocal: 3.1.0 is already released, for 3.2.0 🙏🏻

Kocal (approved):

Kocal:
Thank you @treztreiz.

PR #3604, opened 2026-05-29
 Replace npm release workflow with a local interactive PHP script

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

**I'm open to discussion, but I strongly believe automated releases on npm must not be a thing anymore.**

The `release-on-npm.yaml` workflow has been failing repeatedly when publishing `@symfony/ux-native` (see [run 26623506573](https://github.com/symfony/ux/actions/runs/26623506573/job/78454667136) and #3522) **EDIT: I forgot to enable OIDC on `@symfony/ux-native` npm package 🙄**.

The combination of OIDC (#3142), workflow permissions tweaks, and an automated bump+commit+push from CI is hard to debug and adds a real security surface for a step that runs once per release.

This PR drops the workflow and adds `bin/release-on-npm.php`, an interactive script run manually after pushing a release tag.

All UX packages have been configured on npm to only allow manual releases with 2FA enabled
image

After merging and upmerge, the OIDC (trusted publisher) will be removed from npm packages, since it won't be useful anymore.

## Usage

```bash
php bin/release-on-npm.php 2.37.0 # default remote: upstream
php bin/release-on-npm.php 2.37.0 --remote=origin
```

PR #3605, opened 2026-05-29
[CI] Harden the release-on-npm workflow

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

Following #3604

## Summary

Following an audit of `.github/workflows/release-on-npm.yaml` with my super friend Claude, 
this PR tightens several aspects of the npm release pipeline. 
OIDC trusted publisher protects authentication on the npm side, but says nothing about the integrity of the code being published. 
The changes below close that gap.

Each commit addresses one finding from the audit, tagged `(F<n>)` in its title.

## Commits

- **[CI] Publish from the tagged commit, not the branch HEAD (F1)** — checkout `${{ github.ref }}` (the tag) instead
of the hardcoded `2.x` branch, and verify the tagged commit is an ancestor of the matching release branch
(`2.x` for `v2.*.*`, `3.x` for `v3.*.*`). Closes a TOCTOU where `2.x` HEAD could drift between tag creation and job execution.

- **[CI] Stop touching package.json versions during release (F8)** — drop the `pnpm version` + commit + push back to
`2.x` block. The Git tag is the single source of truth; version bumps are now done manually before tagging. The workflow token is downgraded from `contents: write` to `contents: read`, and **the new `MAINTAINERS.md` documents the bump procedure.** 
Code can not be commited/pushed to the repo anymore ✨ 

- **[CI] Pin npm version used by the release workflow (F6)** — replace `npm@latest` with `npm@11.16.0` so a compromised npm release cannot ship through the release path without an explicit, reviewable bump PR.

- **[CI] Generate npm provenance attestations on publish (F7)** — add `--provenance` to `pnpm publish`. Each tarball now carries a signed SLSA attestation pointing to the repo, commit and workflow run, verifiable via `npm audit signatures`.

- **[CI] Serialize release runs per tag (F9)** — add a concurrency group keyed on `github.ref_name`, so two runs on the same tag queue instead of overlapping. Distinct tags (`v2.x.y` vs `v3.0.0`) still run in parallel.

- **[CI] Build assets and refuse to publish if dist files drift (F2)** — run `pnpm build` on the tagged commit and fail if `git diff --quiet` reports anything. Mirrors the existing `dist-files-unbuilt.yaml` PR check, brought into the release path so tampered dist/source pairs cannot publish even if the branch CI was bypassed.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,26 @@
+# Maintainers' guide
+
+This document is for Symfony UX maintainers. It collects procedures that are not
+needed by regular contributors.
+
+## Releasing UX packages on NPM
+
+The Git tag is the source of truth for the released version. The
+`release-on-npm.yaml` workflow refuses to publish if any workspace
+`package.json` does not match the tag — bump and commit **before** tagging.
+
+From the release branch (`2.x` or `3.x`), with `upstream` pointing to `symfony/ux`:
+
+```shell
+$ git checkout 2.x # or 3.x
+$ VERSION=2.36.0 && \
+  pnpm install --frozen-lockfile && \
+  pnpm build && \
+  pnpm version $VERSION --no-git-tag-version --workspaces --no-workspaces-update && \
+  git add . && \
+  git commit -m "Bump npm packages to v$VERSION"
+$ git push upstream HEAD
+```
+
+The tag is created and pushed afterwards by following the Symfony's release process.
+The `release-on-npm.yaml` workflow then publishes via OIDC trusted publisher.
--- a/.github/workflows/release-on-npm.yaml
+++ b/.github/workflows/release-on-npm.yaml
@@ -53,5 +53,18 @@ jobs:
             - name: Install root JS dependencies
               run: pnpm install --frozen-lockfile
 
+            - name: Build JS assets
+              run: pnpm run build
+
+            - name: Verify dist files match committed sources
+              run: |
+                  if ! git diff --quiet; then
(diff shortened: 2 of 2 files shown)


MAINTAINERS.md
```diff
@@ -0,0 +1,26 @@
...
+From the release branch (`2.x` or `3.x`), with `upstream` pointing to `symfony/ux`:
+
+```shell
+$ git checkout 2.x # or 3.x
+$ VERSION=2.36.0 && \
+  pnpm install --frozen-lockfile && \
+  pnpm build && \
+  pnpm version $VERSION --no-git-tag-version --workspaces --no-workspaces-update && \
+  git add . && \
+  git commit -m "Update versions to $VERSION"
```
Kocal: I think we can remove the `git add` and `git commit` commands, I believe `pnpm version` can already handle this (at least, yarn could)

Kocal: Well, nope

kbond (commented):
Does/should this address the packages.json version update?

Kocal:
> Does/should this address the packages.json version update?

Yes, the `release-on-npm.yaml` workflow previously commited/push all new files, and had 
`write` permissions on contents.

With this potential security vulnerability, and the fact that `version` field from `package.json` files were updated **after** the creation of the release tag (by you or Fabien), leading to inconsistency etc., ATM I think the best option is to **manually** bump npm versions **before** the release tag.

Kocal:
As discussed, I'm will add a new dispatchable workflow (in another PR) for opening a PR that automatically modifies the `version` from `package.json` files

Issue #469, opened 2022-09-20
[LiveComponent] Select list with no empty choice considered empty


jcrombez:
When using a FormType in a LiveComponent, if your form contains a select field with no empty option, the browser autoselect the first one by default.
But since the form itself had no value for this field, if the user keep the default value, the component data is never updated with it and the user get a validation error for missing value on a field he sees with a value.

weaverryan:
This looks legitimate to me. In your Symfony form, the field (e.g. `country`) has no value. But since there is no blank value (e.g. "choose a country"), then in practice, the first country in the list is auto-selected. If the user never changes this field, then the first country is never set onto the "live components" data.

I've purposely avoided doing this until now, but the only solution I can think of is:

A) When the live controller loads, find all "model" inputs
B) Read the "value" of the model input and "set" it back into the client-side component data.

So basically, on load, we look for all model fields and "synchronize" their value back to the component data.

jcrombez:
I don't know if it's related, i suspect it is but, if some javascript modify a field value, it doesn't update the component state and when the user manually trigger a change, the field is back to its value before any javascript modificaiton.

weaverryan:
That is a separate issue. If some external JavaScript modifies the value of a field directly, as far as I know, there is no way for live component to "hook into that" (e.g. via a listener) and update the model. That external JavaScript needs to trigger an `input` or `change` event on that element after setting the value. This is something that should be documented

weaverryan:
I'm working on this in #472 - but the documentation for changing fields via JavaScript is already there :)

jcrombez:
Thank you :+1:

PR #3613, opened 2026-06-01
[LiveComponent] Sync auto-selected <select> value in forms

### LINKED ISSUE #767: [LiveComponent] No selected <option> treated differently when submitted from a Live Component
Using `2.x.dev` which includes #728, and latest Stimulus Bridge `3.2.2`.

This is a bit of a weird one. If I for example use an `EntityType` or a `ChoiceType` to generate a `<select>` element, and I do not use a `placeholder` value, we will get HTML such as:

```html
<select>
  <option value="1">Option One</option>
  <option value="2">Option Two</option>
</select>
```

When this is rendered by the browser, if no option is selected, and there is no placeholder option, the first option "Option One" will be shown, but without a `selected` attribute.

In a normal Symfony form in a standard POST submission, the browser seems to send the "Option One" as though the user had physically selected it. For example, a `NotBlank()` validation would PASS.

### The Problem

If you then move this form into a LiveComponent, the behaviour changes. 

Submitting the form without touching the select element does NOT assume that the first element is implicitly selected, and `null` is sent as the select field's value. 

This means any `NotBlank()` validation FAILS, even though as far as the user is concerned, the "Option One" was selected (it was shown as the selected entry in the form field, even though they didn't physically click on it).

This then makes a mess when submitting the form, as you get a validation error, and you have to go and click and select the option that is already, as far as you the user is concerned, selected.

I'm guessing this is to do with the way that the Live Component's JS submits the form, and that this is differing in this case with a standard browser form submit.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #767
| License       | MIT

A single `<select>` without a placeholder/empty option is auto-selected by the browser on its first `<option>`. In a native form POST, that first option is submitted as if picked. Inside a LiveComponent form, however, the model stayed `null` until the user actually interacted with the field, so a `NotBlank()` constraint failed even though the option appeared selected.

`SetValueOntoModelFieldsPlugin` already handles this for selects carrying a `data-model` (#469), but it only iterated over `[data-model]` elements. Form fields inside `<form data-model="*">` are bound through their `name` attribute and have no `data-model` of their own, so they were skipped.

This widens the selector to also pick up `select[name]` elements and resolves their model through the existing `getModelDirectiveFromElement()` (which already knows how to derive the model name from `name` within a `data-model` form). The existing single-select sync logic then reflects the auto-selected option into the model — matching native browser behavior.

Added a unit test mirroring the existing `#469` one, but for a select living inside a `<form data-model="*">`.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -623,9 +623,12 @@ var Idiomorph = (function() {
 			} else if (ctx.head.shouldRemove(currentHeadElt) !== false) removed.push(currentHeadElt);
 		}
 		nodesToAppend.push(...srcToNewHeadNodes.values());
+		log("to append: ", nodesToAppend);
 		let promises = [];
 		for (const newNode of nodesToAppend) {
+			log("adding: ", newNode);
 			let newElt = document.createRange().createContextualFragment(newNode.outerHTML).firstChild;
+			log(newElt);
 			if (ctx.callbacks.beforeNodeAdded(newElt) !== false) {
 				if (newElt.href || newElt.src) {
 					let resolve = null;
@@ -653,6 +656,7 @@ var Idiomorph = (function() {
 		});
 		return promises;
 	}
+	function log() {}
 	function noOp() {}
 	function mergeDefaults(config) {
 		let finalConfig = {};
@@ -1748,8 +1752,7 @@ var ChildComponentPlugin_default = class {
 	constructor(component) {
 		this.parentModelBindings = [];
 		this.component = component;
-		const modelDirectives = getAllModelDirectiveFromElements(this.component.element);
-		this.parentModelBindings = modelDirectives.map(get_model_binding_default);
+		this.parentModelBindings = getAllModelDirectiveFromElements(this.component.element).map(get_model_binding_default);
 	}
 	attachToComponent(component) {
 		component.on("request:started", (requestData) => {
@@ -2049,11 +2052,11 @@ var SetValueOntoModelFieldsPlugin_default = class {
 		});
 	}
 	synchronizeValueOfModelFields(component) {
-		component.element.querySelectorAll("[data-model]").forEach((element) => {
+		component.element.querySelectorAll("[data-model], select[name]").forEach((element) => {
 			if (!(element instanceof HTMLElement)) throw new Error("Invalid element using data-model.");
 			if (element instanceof HTMLFormElement) return;
 			if (!elementBelongsToThisComponent(element, component)) return;
-			const modelDirective = getModelDirectiveFromElement(element);
+			const modelDirective = getModelDirectiveFromElement(element, false);
 			if (!modelDirective) return;
 			const modelName = modelDirective.action;
 			if (component.getUnsyncedModels().includes(modelName)) return;
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -623,12 +623,9 @@ var Idiomorph = (function() {
 			} else if (ctx.head.shouldRemove(currentHeadElt) !== false) removed.push(currentHeadElt);
 		}
 		nodesToAppend.push(...srcToNewHeadNodes.values());
-		log("to append: ", nodesToAppend);
 		let promises = [];
 		for (const newNode of nodesToAppend) {
-			log("adding: ", newNode);
 			let newElt = document.createRange().createContextualFragment(newNode.outerHTML).firstChild;
-			log(newElt);
 			if (ctx.callbacks.beforeNodeAdded(newElt) !== false) {
 				if (newElt.href || newElt.src) {
 					let resolve = null;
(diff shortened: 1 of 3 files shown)


smnandre:
> however, the model stayed null until the user actually interacted with the field

For now, determining "which value is selected" is handled on the PHP side (and has cost us a few hairs 😅).

I believed this was fixed, could you assemble a minimal reproducer so we can look into it before making changes in the JS part ?

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -623,9 +623,12 @@ var Idiomorph = (function() {
 			} else if (ctx.head.shouldRemove(currentHeadElt) !== false) removed.push(currentHeadElt);
 		}
 		nodesToAppend.push(...srcToNewHeadNodes.values());
+		log("to append: ", nodesToAppend);
```
smnandre, line 626:
```suggestion
```

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -623,9 +623,12 @@ var Idiomorph = (function() {
...
 		let promises = [];
 		for (const newNode of nodesToAppend) {
+			log("adding: ", newNode);
```
smnandre, line 629:
```suggestion
```

smnandre: And the other ones too :)

Amoifr:
Thanks for looking into this, and you're right to challenge the premise. 🙏

I dug in and you're spot on: the `required` case is already handled on the PHP side by `ComponentWithFormTrait::extractFormValues()` (the browser-simulation block gated on `$child->vars['required']`, added in #2621 / your edge-case fix). So the original #767 scenario (required select, no placeholder, `NotBlank`) is **already fixed** on `3.x`. My PR's "the model stays null" claim doesn't hold for required fields. Sorry for the noise!

Here's a minimal reproducer (drop it into `src/LiveComponent/tests/Unit/Form/` and run with `phpunit --filter testInitialSelectValues`):

```php
<?php

namespace Symfony\UX\LiveComponent\Tests\Unit\Form;

use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
use Symfony\Component\Form\FormFactoryInterface;
use Symfony\Component\Form\FormInterface;
use Symfony\UX\LiveComponent\ComponentWithFormTrait;

final class Repro767SelectTest extends KernelTestCase
{
    public function testInitialSelectValues(): void
    {
        $formFactory = self::getContainer()->get('form.factory');
        $component = new Repro767Component($formFactory);
        $component->initializeForm([]);

        // REQUIRED select, no placeholder: already synced to the first option (#767 fixed)
        $this->assertSame('a', $component->formValues['required_select']);
        // OPTIONAL select, no placeholder: PHP deliberately leaves it empty
        $this->assertSame('', $component->formValues['optional_select']);
    }
}

class Repro767Component extends AbstractController
{
    use ComponentWithFormTrait;
    use DefaultActionTrait;

    public function __construct(private FormFactoryInterface $formFactory) {}

    {
        return $this->formFactory->createNamed('form', Repro767FormType::class);
    }
}

class Repro767FormType extends AbstractType
{
    {
        $builder
            ->add('required_select', ChoiceType::class, ['choices' => ['A' => 'a', 'B' => 'b'], 'required' => true])
            ->add('optional_select', ChoiceType::class, ['choices' => ['A' => 'a', 'B' => 'b'], 'required' => false]);
    }
}
```

Both assertions pass on `3.x` today. So the **only** remaining discrepancy is the `required => false` select: a browser still auto-submits its first option in a native POST, but the live component keeps it empty. Pushing that into the JS layer (as this PR does) would actually create a PHP/JS mismatch (server renders `''`, client syncs `"a"`), which is probably not what we want.

Given that, I'm inclined to say #767 itself is resolved. Happy to close this PR unless you consider the optional-select case a bug worth aligning too, in which case it likely belongs on the PHP side next to the existing `required` logic rather than in the plugin. Your call. 🙂

src/LiveComponent/assets/dist/live_controller.js
Amoifr: Gone, and sorry for the long silence on this one: I answered you in a top-level comment back in July and never came back to the threads themselves, so you had no way to close them.

Those `log()` calls were not mine to add. They come from the vendored Idiomorph inside the bundle and my local rebuild pulled in a version that had them, which then landed in `dist/`. 695d1782 rebuilds it cleanly.

src/LiveComponent/assets/dist/live_controller.js
Amoifr: Done, all of them. The `dist/` diff is now down to the two lines this PR is actually about:

```diff
-		component.element.querySelectorAll("[data-model]").forEach((element) => {
+		component.element.querySelectorAll("[data-model], select[name]").forEach((element) => {
-			const modelDirective = getModelDirectiveFromElement(element);
+			const modelDirective = getModelDirectiveFromElement(element, false);
```

Nothing else from Idiomorph or anywhere else in the built file.

PR #3655, opened 2026-06-04
[Toolkit] Add lint kit command and CI lint workflow

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

Add a quality-control layer for Toolkit kits, so structural and dependency-coherence regressions are caught before they ship.

A new `ux:toolkit:lint-kit <kit-path>` command runs a pipeline of small checkers against a kit and reports issues per recipe.
The command is exposed as a standalone `bin/ux-toolkit-kit-lint` only, no DI wiring, since it has no value inside a user's app: it's a tool for kit maintainers, used locally or from CI.

Checks performed:
- **structural**: missing recipe manifests, broken copy-files entries, unknown references in `dependencies.recipe[]` (errors);
- **coherence**: Stimulus controllers used in Twig without a matching JS file, bare JS imports not declared in `dependencies.npm` / `.importmap`, and a curated mapping between Composer packages and the Twig symbols they provide.

A dedicated GitHub Actions workflow runs the linter on every kit under`src/Toolkit/kits/*` via a dynamically built matrix, triggered only when`src/Toolkit/**` changes.
Each kit gets its own job (`fail-fast: false`) so output stays readable and failures are isolated; adding a new kit needs no workflow change.

### DIFF
--- a/.github/workflows/toolkit-lint-kits.yaml
+++ b/.github/workflows/toolkit-lint-kits.yaml
@@ -1,0 +1,71 @@
+name: Toolkit Lint Kits
+
+defaults:
+    run:
+        shell: bash
+
+on:
+    push:
+        branches-ignore:
+            - 'dependabot/**'
+        paths:
+            - 'src/Toolkit/**'
+            - '.github/workflows/toolkit-lint-kits.yaml'
+    pull_request:
+        paths:
+            - 'src/Toolkit/**'
+            - '.github/workflows/toolkit-lint-kits.yaml'
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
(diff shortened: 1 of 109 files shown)


seb-jean:
This lint kit command is great, thank you.

Would it be difficult to add an option to fix errors?
Like the ones already available for PHP CS Fixer or Twig CS Fixer.

Kocal:
(warnings found in the two jobs will be fixed in another PR)

> Would it be difficult to add an option to fix errors?
> Like the ones already available for PHP CS Fixer or Twig CS Fixer.

That's a good question, it's surely possible, maybe in a next iteration, if identified errors could be automatically fixed.

PR #3661, opened 2026-06-05
[TwigComponent] Fix backslashes being stripped from literal component attribute values

fabpot:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | n/a
| License        | MIT

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 3.2.0
 
+- Fix backslashes being stripped from literal component attribute values
 - Add `ComponentRendererInterface::preCreateForRender()`, `ComponentRendererInterface::startEmbeddedComponentRender()`, and `ComponentRendererInterface::finishEmbeddedComponentRender()` methods
 
 ## 3.1.0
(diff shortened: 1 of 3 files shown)


Kocal:
Browser tests are failing because http://127.0.0.1:9876/ux-turbo/broadcast/books now renders:
> An exception has been thrown during the rendering of a template ("The autoloader expected class "App\\Entity\\Turbo\\Book" to be defined in file "/Users/kocal/workspace/symfony/ux/apps/e2e/vendor/composer/../../src//Entity//Turbo//Book.php". The file was found but the class was not in it, the class name or namespace probably has a typo.") in @Turbo/components/Stream/From.html.twig at line 2.

The code `<twig:Turbo:Stream:From topics="App\\Entity\\Turbo\\Book" />` must be changed to `<twig:Turbo:Stream:From topics="App\Entity\Turbo\Book" />`. 
The other references must be fixed in the following files:
- `apps/e2e/templates/ux_turbo/broadcast/artist_from_song.html.twig`
- `apps/e2e/templates/ux_turbo/broadcast/artists.html.twig`
- `apps/e2e/templates/ux_turbo/broadcast/book.html.twig`
- `src/Turbo/doc/index.rst`

Btw, it currently looks like a BC break right? Should we replace `\\` in topic(s) inside `\Symfony\UX\Turbo\Twig\TurboRuntime::renderTurboStreamFrom()` or something equivalent?

PR #3664, opened 2026-06-06
[LiveComponent] Add robust support for async action confirmations (Turbo, SweetAlert2, Native)

xDeSwa:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| Issues         | #
| License        | MIT

**Preview** **(This preview is outdated and is not valid for the new system.)**
Animation

### Description

This PR introduces a robust and flexible way to add confirmation dialogs to LiveComponent actions using the new `live#confirmAction` mechanism. 

Previously, trying to intercept the core `live#action` asynchronously caused issues with `event.currentTarget` nullification and disrupted the synchronous action queue. By separating this into its own `live#confirmAction`, we ensure a flawless integration that safely resolves the confirmation before safely triggering the original action logic.

**Key Features:**
1. **Zero-Config Native Fallback**: Out of the box, it falls back to the browser's native `window.confirm()`.
2. **Automatic Turbo Support**: If the project uses `@hotwired/turbo` and has a globally configured confirm method (`Turbo.config.forms.confirm` or `Turbo.navigator.confirmMethod`), it detects and uses it automatically! No extra code needed.
3. **Promise-based Custom Modals**: Developers can easily hook into the `live:confirmAction` event to integrate any UI library (SweetAlert2, Bootstrap Modals, etc.) simply by passing a Promise.
4. **Form Interception**: Fully supports the `:prevent` modifier to stop native `<form>` submissions until confirmed.

### Examples

**Basic Usage:**
```html
<button
    data-action="live#confirmAction"
    data-live-action-param="delete"
    data-live-confirm-message-param="Are you sure?"
>Delete</button>
```

**Intercepting a Form:**
```html
<form data-action="live#confirmAction:prevent"
      data-live-action-param="save"
      data-live-confirm-message-param="Save all changes?">
    <button type="submit">Save Settings</button>
</form>
```

**Custom Modal Integration (e.g. SweetAlert2):**

```javascript
document.addEventListener('live:confirmAction', (event) => {
    event.preventDefault();
    event.detail.promise = Swal.fire({
        title: event.detail.message,
        icon: 'warning',
        showCancelButton: true
    }).then(result => result.isConfirmed);
});
```

**Customizing the Confirmation Dialog with Turbo (Optional)**

```javascript
import * as Turbo from '@hotwired/turbo';
    import Swal from 'sweetalert2';

    // Turbo 7 usage:
    // Turbo.setConfirmMethod((message, element) => { ... });

    // Turbo 8 usage:
    Turbo.config.forms.confirm = (message, formElement, submitter) => {
        return Swal.fire({
            title: message,
            icon: 'warning',
            showCancelButton: true
        }).then(result => result.isConfirmed);
    };
```

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -1282,26 +1282,69 @@ the ``#[LiveArg]`` attribute::
 Confirming Actions
 ~~~~~~~~~~~~~~~~~~
 
-If you want to ask the user for confirmation before executing a LiveComponent action,
-you have three ways to do it.
+To ask the user for confirmation before executing a LiveComponent action,
+use the ``live#confirmAction`` action instead of the standard ``live#action``.
 
-**1. Using standard Turbo Confirm (If Turbo is installed)**
+This dedicated action ensures your confirmations are handled asynchronously without disrupting
+the synchronous execution queue of standard actions.
 
-If you are already using ``@hotwired/turbo`` and have configured a custom confirm method
-(like SweetAlert) via ``Turbo.config.forms.confirm`` or ``Turbo.setConfirmMethod``,
-LiveComponent will automatically respect it. Just add the ``data-turbo-confirm`` attribute:
+**1. Basic Usage**
+
+Use ``data-action="live#confirmAction"`` and provide the action name and message:
 
 .. code-block:: html+twig
 
     <button
-        data-action="live#action"
+        data-action="live#confirmAction"
         data-live-action-param="delete"
-        data-turbo-confirm="Are you sure you want to delete this item?"
+        data-live-confirm-message-param="Are you sure you want to delete this item?"
     >Delete</button>
 
-Here is an example of how you might configure Turbo with SweetAlert2 in your JavaScript
+.. note::
+
+    If you are used to Turbo, you can also use the ``data-turbo-confirm`` attribute instead
+    of ``data-live-confirm-message-param``. Both will work seamlessly.
+
+By default, this triggers the browser's native ``window.confirm()``.
+
+**2. Passing Extra Parameters & Modifiers**
+
+Just like standard actions, you can pass additional parameters, use modifiers (like ``:prevent``),
+and pipe multiple actions together:
+
+.. code-block:: html+twig
+
+    <button
+        data-action="live#confirmAction"
+        data-live-action-param="debounce(500)|delete"
+        data-live-confirm-message-param="Are you sure?"
+        
+        data-live-id-param="{{ item.id }}"
+    >Delete</button>
+
+You can also use it on forms to intercept the ``submit`` event and prevent page reloads:
+
+.. code-block:: html+twig
+
+    <form data-action="live#confirmAction:prevent"
+          data-live-action-param="saveSettings"
+          data-live-confirm-message-param="This will reboot the system. Are you sure?">
+        
+        <!-- form fields... -->
+        
+        <button type="submit">Save Settings</button>
+    </form>
+
+**3. Automatic Turbo Integration**
+
+If your project uses ``@hotwired/turbo`` and you have a custom confirm method configured
+globally via ``Turbo.config.forms.confirm`` or ``Turbo.setConfirmMethod``,
+LiveComponent will automatically detect and use it! No extra configuration is needed.
+
+Here is an example of how you might configure Turbo in your JavaScript
 (e.g., in ``app.js``). While this example uses SweetAlert2, you can seamlessly integrate
-any other library or your own custom UI design as long as it returns a Promise:
+any other library (like Bootstrap Modals, Tailwind UI) or your own custom UI design
+as long as it returns a Promise:
 
 .. code-block:: javascript
 
@@ -1312,7 +1355,7 @@ any other library or your own custom UI design as long as it returns a Promise:
     // Turbo.setConfirmMethod((message, element) => { ... });
 
     // Turbo 8 usage:
-    Turbo.config.forms.confirm = (message, element) => {
+    Turbo.config.forms.confirm = (message, formElement, submitter) => {
         return Swal.fire({
             title: message,
             icon: 'warning',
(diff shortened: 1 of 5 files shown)


xDeSwa:
Status: Needs Work

xDeSwa:
### Update: Fixed Stimulus `:prevent` Modifier & Native `onclick` BC Compatibility

During testing, we identified and resolved a critical edge case regarding the event propagation cycle and Backward Compatibility (BC) with native HTML handlers.

**The Issue:**
Previously, an `if (event.defaultPrevented) { return; }` check was added at the top of the `action(event)` method in `live_controller.ts` to abort the action if another listener canceled the event.
However, this inadvertently **broke the core Stimulus `:prevent` modifier** (e.g., `data-action="live#action:prevent"`). Stimulus calls `event.preventDefault()` *before* invoking the action, which meant the action was silently aborting itself.

Additionally, removing the check entirely would break backward compatibility for existing legacy codebases that still rely on native `onclick="return confirm('...')"` or `onclick="return false;"` to abort actions.

**The Fix:**
We introduced a smart heuristic in `live_controller.ts` to distinguish between an event prevented by Stimulus versus one prevented by external/legacy code:

```typescript
const actionAttr = target.getAttribute('data-action') || '';
// Only abort if the event was prevented natively AND it wasn't caused by the Stimulus :prevent modifier
if (event.defaultPrevented && !actionAttr.includes(':prevent')) {
    return;
}
```
Status: Needs Review

src/LiveComponent/doc/index.rst
```diff
@@ -1279,6 +1279,96 @@ the ``#[LiveArg]`` attribute::
...
+
+        // Pass your Promise to event.detail.promise
+        event.detail.promise = Swal.fire({
+            title: event.detail.message,
+            icon: 'warning',
+            showCancelButton: true
+        }).then(result => result.isConfirmed);
+    });
+
+**3. Using Native Inline Confirm**
```
MrYamous, line 1358: Is this behavior different than the one exposed in second point without using any library while listening `live:confirm` event ? IIUC both rely on native `window.confirm()`, did I miss something ?

xDeSwa: `data-live-confirm` falls back to the browser’s native window.confirm() by default, but you can easily customize the dialog (SweetAlert2, Bootbox, AlertifyJS, Notiflix, MicroModal, etc.).

Two supported ways to customize:

- If you use Turbo (7/8), set the global confirm method to return a Promise<boolean>:
```javascript
import * as Turbo from '@hotwired/turbo';
import Swal from 'sweetalert2';

Turbo.config.forms.confirm = (message, element) => {
  return Swal.fire({
    title: message,
    icon: 'warning',
    showCancelButton: true
  }).then(result => result.isConfirmed);
};
```

- If you don’t use Turbo, listen for the live:confirm event, call event.preventDefault() and assign a Promise<boolean> to event.detail.promise:
```javascript
import Swal from 'sweetalert2';

document.addEventListener('live:confirm', (event) => {
  event.preventDefault(); // stop native fallback
  event.detail.promise = Swal.fire({
    title: event.detail.message,
    icon: 'warning',
    showCancelButton: true
  }).then(result => result.isConfirmed);
});
```

Libraries that work well for async confirms: SweetAlert2 (recommended — Promise-based), Bootbox (Bootstrap), AlertifyJS, Notiflix, MicroModal, Tingle.js, etc. Callback-only libs can be wrapped in a Promise as shown above.

NikkieDev (approved):

smnandre:
Does it work with multiple actions ? Can one select the one they want to confirm ? 

If i'm not mistaken this freeze the handling, what happens if one action react to an event, but other are not expecting any confirm ? Could we add a test ? 

All this is not fresh enough for me to review quickly, i'll give a deeper look later this week!

smnandre:
I think the asynchronous confirmation needs a different boundary before it can be added to LiveComponent.

Stimulus dispatches the action descriptors for an event synchronously and in template order. Its [event listener loops through the bindings synchronously](https://github.com/hotwired/stimulus/blob/v3.2.2/src/core/event_listener.ts#L34-L43), and [invokes each controller method without awaiting its return value](https://github.com/hotwired/stimulus/blob/v3.2.2/src/core/binding.ts#L74-L82).

An asynchronous Stimulus action is fine when it is independent. However, it cannot pause and later resume the original event dispatch. It runs until the first `await`, returns a Promise that Stimulus does not observe or await, then resumes later.

The current implementation calls [`stopImmediatePropagation()` before the first `await`](https://github.com/symfony/ux/blob/08329850b91a4f3e72c2495aa6e4126e18432922/src/LiveComponent/assets/src/live_controller.ts#L146-L161). This produces different behavior depending on descriptor order:

```text
live#action other#action
-> Live starts the confirmation
-> other#action never runs, even when confirmation succeeds

other#action live#action
-> other#action runs before the confirmation result
-> canceling the confirmation cannot undo it
```

Removing `stopImmediatePropagation()` would not solve the ordering issue. The following actions would run while Live is waiting, then the LiveAction would resume later.

Ordering is normally a choice made by the template author, but neither order provides coherent confirmation semantics here. One order suppresses unrelated actions, while the other allows them to run before confirmation.

There are a few related issues:

1. [`defaultPrevented` and the `:prevent` detection](https://github.com/symfony/ux/blob/08329850b91a4f3e72c2495aa6e4126e18432922/src/LiveComponent/assets/src/live_controller.ts#L131-L136) do not identify why an event was canceled. `defaultPrevented` is shared event state, and `actionAttr.includes(':prevent')` searches every descriptor on the element. An unrelated `other#action:prevent` can therefore make Live ignore a native cancellation. The same problem occurs with `live#action:prevent`: if native or external code canceled the event first, the presence of the Stimulus modifier still makes the LiveAction execute. A local regression test reproduced the unexpected request.

2. The Turbo confirmation contract is not preserved. [Turbo 8 resolves the attribute from the submitter and form, then calls](https://github.com/hotwired/turbo/blob/v8.0.20/src/core/drive/form_submission.js#L77-L90):

   ```js
   confirmMethod(message, formElement, submitter)
   ```

   The PR reads only `event.currentTarget` and calls `confirmMethod(message, target)`. A configured method that uses the form or submitter receives different arguments. The current unit test expects `('Turbo?', expect.any(HTMLElement))`, so it encodes the two-argument behavior instead of testing Turbo's contract.

   The documented Turbo 7 support also seems incomplete. [`Turbo.setConfirmMethod()` assigns `FormSubmission.confirmMethod`](https://github.com/hotwired/turbo/blob/v7.3.0/src/core/index.ts#L127-L131), not `Turbo.navigator.confirmMethod`.

3. Making the public Stimulus [`action()` method async](https://github.com/symfony/ux/blob/08329850b91a4f3e72c2495aa6e4126e18432922/src/LiveComponent/assets/src/live_controller.ts#L131-L143) changes its error channel. Existing synchronous errors now become rejected Promises. Since Stimulus does not await the return value, these errors no longer reach its synchronous error handler. A rejected Turbo confirmation Promise has the same problem.

My preference would be to separate the existing `defaultPrevented` bugfix from the asynchronous confirmation feature.

Asynchronous confirmation can already be implemented in userland through the documented [`getComponent()` and `component.action()` API](https://github.com/symfony/ux/blob/08329850b91a4f3e72c2495aa6e4126e18432922/src/LiveComponent/doc/index.rst#L885-L921):

```js

    async initialize() {
    }

    async action(event) {
        const confirmed = await showConfirmation(event.params.message);

        if (confirmed) {
            await this.component.action(event.params.action);
        }
    }
}
```

The template would use this orchestration controller instead of combining it with `live#action`:

```html
<button
    data-action="confirm#action"
    data-confirm-action-param="delete"
    data-confirm-message-param="Are you sure?"
>
    Delete
</button>
```

This makes the confirmation, cancellation and eventual LiveAction part of one explicit operation.

If asynchronous confirmation should become a core LiveComponent feature, I think it needs an explicit awaited pre-action pipeline or hook. Integrating it inside the generic Stimulus action cannot preserve the ordering semantics of the original event.

The tests currently cover one `live#action` descriptor at a time and use a hand-written `window.Turbo` shape. They do not load Turbo or exercise its public API.

Tests covering both descriptor orders, `live#action:prevent` after an external cancellation, unrelated `:prevent` modifiers, following native listeners, Turbo 7, Turbo's three callback arguments, rejected confirmation Promises and controller disconnection while a confirmation is pending would help define the intended behavior.

While I completely understand the idea behind this PR, I think we need to find an approach that does not involve such a significant change for every application using LiveComponent.

xDeSwa:
Thank you so much for this detailed and insightful analysis @smnandre , You are absolutely right. I hadn't fully considered how Stimulus synchronously loops through event bindings and how intercepting the generic `live#action` disrupts the natural execution order and error handling. i will rework on it.

Status: Needs Work

xDeSwa:
Hi @smnandre,

You were absolutely right about the concerns of intercepting the core `live#action` directly. Making it asynchronous caused issues with `event.currentTarget` nullification and risked disrupting the synchronous action queue and existing parsers. 

To solve this cleanly, I rewrote the approach and introduced a dedicated `live#confirmAction` mechanism instead of modifying the core `action()` method.

**How the new architecture works:**
- It is triggered via `data-action="live#confirmAction"`.
- It pauses execution to resolve a confirmation (Promise-based).
- Once confirmed, it internally mocks the event and safely triggers the original `this.action(fakeEvent)` without any disruption.

**Key Benefits:**
1. **Zero-Config Native Fallback**: Uses `window.confirm()` by default if no library is present.
2. **Automatic Turbo Support**: Automatically detects and uses `Turbo.config.forms.confirm` (or Turbo 7's equivalent) if the project uses it.
3. **Custom Modal Support**: Developers can intercept the `live:confirmAction` event to pass a custom Promise (e.g., SweetAlert2, Bootstrap) easily.

I've pushed the new commit which includes comprehensive unit tests for all these scenarios and updated the documentation with detailed examples (including `<form>` interceptions). 

Let me know what you think of this new approach

PR #3676, opened 2026-06-11
[LiveComponent] Avoid 500 errors on malformed hydration payloads

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | -
| License       | MIT

Hydrating a LiveComponent from a tampered `updated` payload could raise an uncaught native exception and surface as a 500 instead of being ignored:

- an array sent for a writable backed-enum prop → `TypeError` in `BackedEnum::tryFrom()`
- an unknown property sent for a writable object prop → `ReflectionException`

These are only reachable by manually crafting the payload (a regular user never hits them). Each case is now turned into a `HydrationException`, so the malformed writable value is ignored like any other bad data and the original value is kept.

Discussed with @Kocal.

### DIFF
--- a/src/LiveComponent/src/LiveComponentHydrator.php
+++ b/src/LiveComponent/src/LiveComponentHydrator.php
@@ -619,6 +619,10 @@ final class LiveComponentHydrator
                 return null;
             }
 
+            if (!\is_int($value) && !\is_string($value)) {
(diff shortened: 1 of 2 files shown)


Kocal (approved):

Kocal:
Thank you @Amoifr.

Issue #3678, opened 2026-06-11
[Cropperjs] Upgrade to Intervention Image v4 and make the image driver configurable


deluxetom:
Follow-up to #3394.

### Context

`symfony/ux-cropperjs` currently requires `intervention/image: ^2.5`. The v2 line is legacy (PHP 5.4-era) and triggers the indirect deprecations reported in #3394. As noted there, upgrading isn't a drop-in bump: from Intervention **v3+**, `ImageManager` no longer accepts an implicit driver-name string — it **requires a `DriverInterface`** — which left an open design question: adopt Intervention's official Symfony integration, or let the bundle expose driver selection itself?

### Proposal

Upgrade straight to **Intervention Image v4** (`^4.0`, dropping v2) and let users choose the driver via bundle config:

```yaml
# config/packages/cropperjs.yaml
cropperjs:
    driver: gd            # gd (default), imagick or vips
    driver_service: ~     # optional: a custom Intervention DriverInterface service (takes precedence)
```

- The `ImageManager` service is built via `ImageManager::usingDriver(...)`.
- `gd`/`imagick` ship with `intervention/image`; `vips` uses the official `intervention/image-driver-vips` package, listed under composer **`suggest`** (it needs libvips + `ext-ffi`). Selecting `vips` without that package **fails fast** with an actionable message.
- `driver_service` is an escape hatch for a fully custom / pre-configured `DriverInterface`.

**Why v4 (not v3):** a single major to maintain, the vips driver is v4-only, and the bundle already requires PHP >= 8.4 (v4 needs >= 8.3). The `Crop` API maps cleanly: `make() -> decodeBinary()`, `resize(closure) -> scaleDown()`, `encode() + getEncoded() -> encodeUsingFileExtension()`.

**Backward compatibility:** the bundle's public API is unchanged and **`gd` stays the zero-config default**, so existing apps are unaffected.

### Reference implementation

A working branch (migration + config + docs + tests parameterized over gd & imagick): https://github.com/deluxetom/ux/tree/cropperjs-intervention-v4

Happy to open a PR if this direction looks good — or to adjust if the official-integration route is preferred instead.

PR #3679, opened 2026-06-11
[Cropperjs] Upgrade to Intervention Image v4 and make the image driver configurable

### LINKED ISSUE #3394: [Cropperjs] Indirect deprecations with `intervention/image`
While working on #3347 I noticed few deprecations related to `intervention/image`
That can be [checked on CI](https://github.com/symfony/ux/actions/runs/23342065874/job/67897848646?pr=3347)

This dependency may need to be upgraded to v3

deluxetom:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| Issues         | Fix #3394
| License        | MIT

Resolves the indirect `intervention/image` v2 deprecations reported in #3394 by upgrading the dependency to v4, and adds the ability to choose the image driver (proposal/discussion in #3678).

### What changes

- **`intervention/image`: `^2.5` → `^4.0`** (drops v2). The `Crop` model migrates to the v4 API: `make()` → `decodeBinary()`, the `resize(closure)` aspect-ratio + upsize calls → `scaleDown()`, `encode()` + `getEncoded()` → `encodeUsingFileExtension()`, and the type-hints move to `ImageManagerInterface` / `ImageInterface`.
- **Configurable image driver** via a new bundle configuration:

```yaml
# config/packages/cropperjs.yaml
cropperjs:
    driver: gd            # gd (default), imagick or vips
    driver_service: ~     # optional: a custom Intervention DriverInterface service (takes precedence)
```

  The `ImageManager` is built through `ImageManager::usingDriver()`. `gd` / `imagick` are bundled with `intervention/image`; `vips` uses the official `intervention/image-driver-vips` package, declared under composer `suggest` (it needs libvips + `ext-ffi`). Selecting `vips` without that package fails fast at container compile time with an actionable message. `driver_service` lets advanced users inject a fully custom / pre-configured `DriverInterface`.

### Backward compatibility

The bundle's public API is unchanged and **`gd` remains the zero-config default**, so existing apps behave exactly as before. The only requirement change is `intervention/image` v4 (PHP >= 8.3, already satisfied by the bundle's PHP >= 8.4 floor).

### Tests

`CropTest` is parameterized over the **gd and imagick** drivers and now covers the crop region (locking the `crop()` argument order), `setCroppedMaxSize()` downscaling, thumbnail downscaling and output format. New `ConfigurationTest` and `CropperjsExtensionTest` cover the driver configuration, the custom-driver precedence and the fail-fast `vips` guard.

### DIFF
--- a/src/Cropperjs/src/Model/Crop.php
+++ b/src/Cropperjs/src/Model/Crop.php
@@ -61,7 +61,7 @@ class Crop
         $image->scaleDown($maxWidth, $maxHeight);
 
         if (!empty($this->options['rotate'])) {
-            $image->rotate(-1 * $this->options['rotate'], 'ffffff');
+            $image->rotate($this->options['rotate'], 'ffffff');
         }
 
         return (string) $image->encodeUsingFileExtension($format, quality: $quality);
@@ -77,7 +77,7 @@ class Crop
         }
 
         if (!empty($this->options['rotate'])) {
-            $image->rotate(-1 * $this->options['rotate'], 'ffffff');
+            $image->rotate($this->options['rotate'], 'ffffff');
         }
 
         return (string) $image->encodeUsingFileExtension($format, quality: $quality);
--- a/src/Cropperjs/src/Model/Crop.php
+++ b/src/Cropperjs/src/Model/Crop.php
@@ -61,7 +61,7 @@ class Crop
         $image->scaleDown($maxWidth, $maxHeight);
 
         if (!empty($this->options['rotate'])) {
-            $image->rotate($this->options['rotate'], 'ffffff');
+            $image->rotate($this->options['rotate']);
         }
 
         return (string) $image->encodeUsingFileExtension($format, quality: $quality);
@@ -77,7 +77,7 @@ class Crop
         }
 
         if (!empty($this->options['rotate'])) {
-            $image->rotate($this->options['rotate'], 'ffffff');
+            $image->rotate($this->options['rotate']);
         }
 
         return (string) $image->encodeUsingFileExtension($format, quality: $quality);
--- a/src/Cropperjs/src/Model/Crop.php
+++ b/src/Cropperjs/src/Model/Crop.php
@@ -11,9 +11,8 @@
 
 namespace Symfony\UX\Cropperjs\Model;
 
-use Intervention\Image\Constraint;
-use Intervention\Image\Image;
-use Intervention\Image\ImageManager;
+use Intervention\Image\Interfaces\ImageInterface;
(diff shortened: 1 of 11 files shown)


nikola-jovanovic-php (approved):

deluxetom:
@Kocal @MrYamous let me know if you see anything that needs to be updated

src/Cropperjs/src/Model/Crop.php
```diff
@@ -79,24 +73,19 @@ public function getCroppedImage(string $format = 'jpg', int $quality = 80): stri
...
-            $image->resize($this->maxWidth, $this->maxHeight, static function (Constraint $constraint) {
-                $constraint->aspectRatio();
-                $constraint->upsize();
-            });
+            $image->scaleDown($this->maxWidth, $this->maxHeight);
         }
 
         if (!empty($this->options['rotate'])) {
-            $image->rotate(-1 * $this->options['rotate']);
+            $image->rotate(-1 * $this->options['rotate'], 'ffffff');
```
MrYamous: In V4 the `rotate()` rotates clockwise by default (from [upgrade guide](https://image.intervention.io/v4/getting-started/upgrade)), i'm not sure the `-1` is still appropriated with this change. Tests are done with a 90 degrees rotation so I don't think it's noticeable. Maybe we need to add a new test for that ?

deluxetom: Verified: in v2 `rotate()` turned counter-clockwise for positive angles, so the `-1 *` matched cropper.js (which rotates clockwise for positive degrees). In v4 `rotate()` turns clockwise for positive angles (RotateModifier docblock says "Clockwise rotation angle", and the GD modifier negates internally for `imagerotate`), so the negation now inverts the output relative to the cropper preview.

Dropped the `-1` and added a directional test (`testGetCroppedImageRotatesClockwise`): a left-red/right-blue image rotated 90° should land red on top and blue on the bottom. The existing 90° tests only asserted dimensions, which are direction-agnostic, so they couldn't catch this. Fixed in 72f9563b.

MrYamous: Thx, this new test LGTM

Kocal:
Dropping support Intervention Image v2  like that is a big BC break and can't be done on 3.x. 
Instead, I reverted Intervention Image v2 support and added support for v3 and v4, so upgrading steps will be smooth.

Kocal (approved):

Kocal:
Thank you @deluxetom.

PR #3684, opened 2026-06-19
[Dropzone] Allow for multiple file uploads at once

### LINKED ISSUE #7: Dropzone: handle multiple files

### LINKED ISSUE #2567: [Dropzone] Multiple uploads only show single file and no possibility to add more files

Dooij:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no 
| Documentation? | yes
| Issues         | Fix https://github.com/symfony/ux/issues/7 Fix  https://github.com/symfony/ux/issues/2567
| License        | MIT

Adds proper `multiple` support to UX Dropzone.

A native `<input type="file" multiple>` *replaces* its selection on every pick and offers no way to remove a single file, so the Stimulus controller now owns the selection in a `DataTransfer`: it accumulates files across successive picks (files with the same name, size and updatedAt are seen as the same file and not added more than once), writes the result back to `input.files` so the form still submits everything, previews each file, and lets the user remove files individually.

Behavior is triggered by enabling the Types`multiple` option (from FileType).

Usage:

```php
$builder->add('photos', DropzoneType::class, [
    'multiple' => true,
]);
```

A `dropzone:remove` event (the removed `File` in `event.detail`) has been added.

Tests: vitest (accumulate / dedupe / remove), a Playwright browser test using real `setInputFiles`, and a PHP render test for the multiple markup.


smnandre:
I'm opening tomorrow a PR for a brand new UX Upload component, I'm hoping it will match your needs !

Kocal (approved):

Kocal:
Thank you @Dooij.

PR #3685, opened 2026-06-20
[Autocomplete] Fix LIKE ESCAPE clause breaking search on PostgreSQL

### LINKED ISSUE #3668: [ux-autocomplete] LIKE ESCAPE clause breaks search on PostgreSQL since v2.33
## Bug report

**Summary**
Since `symfony/ux-autocomplete` v2.33+, using `searchable_fields` with a PostgreSQL database throws an exception when typing in an autocomplete field.

**Versions**
- `symfony/ux-autocomplete`: v2.32.0 (working) → v2.36.0 (broken)
- Database: PostgreSQL
- PHP: 8.x

**Steps to reproduce**
1. Define an `AsEntityAutocompleteField` with `searchable_fields` containing `string` properties
2. Use a PostgreSQL database
3. Type any character in the autocomplete field

**Expected behavior**
Search works as it did in v2.32.0.

**Actual behavior**
One of the following exceptions is thrown depending on the PDO driver version:
- `SQLSTATE[22025]: Invalid escape sequence: ERROR: invalid escape string — HINT: Escape string must be empty or one character.`
- `SQLSTATE[HY093]: Invalid parameter number: parameter was not defined`

**Root cause**
In `src/Doctrine/EntitySearchUtil.php`, the search clause for text properties is built as follows:

```php
$expressions[] = sprintf("LOWER(%s.%s) LIKE :query_for_text ESCAPE '\\'", $entityName, $propertyName);
```

This generates the following SQL:

```sql
LOWER(p0_.libelle) LIKE :query_for_text ESCAPE '\'
```

PostgreSQL rejects this because the `ESCAPE` clause must be either empty or a single character — a backslash followed by a quote is not valid.

Furthermore, the `ESCAPE` clause is unnecessary here: the wildcards (`%`, `_`) and the backslash are already escaped upstream by `addcslashes($lowercaseQuery, '\\%_')`, so PostgreSQL will treat them literally without any `ESCAPE` clause.

**Suggested fix**
Remove the `ESCAPE` clause entirely for the `LIKE` expression:

```php
$expressions[] = sprintf('LOWER(%s.%s) LIKE :query_for_text', $entityName, $propertyName);
```

This restores the behavior from v2.32.0 and is safe since wildcard escaping is already handled by `addcslashes`.

Note: removing the ESCAPE clause solved the issue in my case (PostgreSQL). However, this may break MySQL/MariaDB where the backslash is natively treated as an escape character in LIKE expressions. A driver-aware fix might be more appropriate.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Tickets       | Fix #3668
| License       | MIT

Since 2.36, `EntitySearchUtil` escapes the LIKE wildcards (`%`, `_`) in the
autocomplete search query and pairs that with an `ESCAPE '\'` clause so the
escaped wildcards are treated literally.

That backslash breaks on **PostgreSQL**: with `standard_conforming_strings = off`
(older setups / some PDO driver versions) the backslash escapes the closing
quote of the SQL string literal, producing invalid SQL. It surfaces as
`SQLSTATE[HY093] Invalid parameter number` or
`SQLSTATE[22025] invalid escape string` and makes every text search throw.

### Fix

Use a **backslash-free** LIKE escape character (`!`) via the platform's
`escapeStringForLike()`. The explicit `ESCAPE` clause is kept on purpose:
SQLite has **no default** LIKE escape character, so dropping it would silently
re-open the wildcard-injection the 2.36 hardening closed.

> Note: the workaround suggested in the issue (`escapeStringForLike($q, '\')` +
> removing the `ESCAPE` clause) wouldn't fully work — `escapeStringForLike($q, '\')`
> produces byte-identical output to the current `addcslashes($q, '\\%_')`, so the
> only effective change there is dropping `ESCAPE`, which regresses wildcard
> escaping on SQLite. Switching the escape *character* is what actually fixes
> PostgreSQL without that regression.

### Tests

- SQLite (CI): the existing `testItEscapesLikeWildcardsInTheQuery` is extended
  with a literal `!` case and stays green — it can only pass if the `ESCAPE`
  clause works.
- PostgreSQL (verified locally against a real server): the old `ESCAPE '\'`
  errors with `standard_conforming_strings=off`; the new `ESCAPE '!'` works
  under both settings and correctly matches literal `%`, `_` and `!`.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 2.37
+
+- Fix the autocomplete search query throwing an exception on PostgreSQL because of
+  the `ESCAPE '\'` clause introduced in 2.36; a backslash-free LIKE escape character is now used.
+
 ## 2.36
 
 - Escape LIKE wildcards (`%` and `_`) in the autocomplete search query so user
--- a/src/Autocomplete/src/Doctrine/EntitySearchUtil.php
+++ b/src/Autocomplete/src/Doctrine/EntitySearchUtil.php
@@ -20,6 +20,13 @@ use Symfony\Component\Uid\Uuid;
  */
 class EntitySearchUtil
 {
+    /**
+     * Custom LIKE escape character. A backslash cannot be used here: the
+     * resulting "ESCAPE '\'" clause produces invalid SQL on PostgreSQL, while
+     * "!" is safe to embed in the ESCAPE clause on every supported platform.
+     */
+    private const LIKE_ESCAPE_CHARACTER = '!';
+
     public function __construct(private EntityMetadataFactory $metadataFactory)
     {
     }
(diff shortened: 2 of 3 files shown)


Kocal:
Can you target branch 2.x instead? Since the security fix was also applied on 2.x, it means that it's also broken there.

Thanks!

Amoifr:
Good catch — you're right, the `ESCAPE '\'` clause landed on 2.x too, so the PostgreSQL breakage is there as well. 🙏

Rebased onto `2.x` and retargeted the PR. CHANGELOG entry moved to a `## 2.37` section. SQLite integration tests still green; the PostgreSQL fix is unchanged. Thanks!

Kocal:
Thank you @Amoifr.

PR #3688, opened 2026-06-25
[LiveComponent] Fix data-loading scanner ignoring data-live-ignore subtrees

### LINKED ISSUE #3665: [LiveComponent] `data-loading` scanner ignores `data-live-ignore`
`LoadingPlugin.getLoadingDirectives()` scans a component's entire subtree for `[data-loading]` and interprets the attribute value as a loading directive, throwing on anything it does not recognise. However, it does not appear to skip `data-live-ignore` subtrees.

As a result, a third-party web component with `data-live-ignore` placed inside a Live Component that manages its own `data-loading` attribute internally will crash the component with a `Unknown data-loading action "<value>"`. And since the scan runs in `attachToComponent()` (`finishLoading(component, component.element)`), the component fails to initialise on page load, and every subsequent live request is broken too.

I came across this while trying to use Altcha (via [`tito10047/altcha-bundle`](https://github.com/tito10047/altcha-bundle)).

---

Short reproducer: a Live Component template containing an embedded widget that sets its own `data-loading`:

```twig
<div {{ attributes }}>
    {# e.g. <altcha-widget>, which sets data-loading="true"/"false" on its internals #}
    <div data-live-ignore>
        <span data-loading="false">widget internals</span>
    </div>

    <button data-action="live#$render">Re-render</button>
</div>
```

On mount, the console throws:

```
Error: Unknown data-loading action "false"
```

With Altcha specifically: `tito10047/altcha-bundle` renders `<altcha-widget data-live-ignore>`, and altcha core sets `data-loading="true"/"false"` on `.altcha-checkbox` during the
proof-of-work. The widget has `data-live-ignore`, yet the loading scanner still reads its internal attribute and throws an error.

---

I am happy to provide a PR, but I would like some guidance on what is an acceptable fix. My preferred solution would be to honour `data-live-ignore`, such that

https://github.com/symfony/ux/blob/4474e2fd7c1acd963492e22c24d57fe10d224780/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts#L179

would become

```ts
matchingElements = matchingElements
    .filter((elt) => elementBelongsToThisComponent(elt, component))
    .filter((elt) => !elt.closest('[data-live-ignore]'));
```

Using `symfony/ux-live-component` v3.1.0, with `tito10047/altcha-bundle` 3.1.0. Somewhat related to GH-3448.

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3665
| License        | MIT

The `data-loading` scanner walks a component's entire subtree, including `data-live-ignore` regions. A third-party widget placed inside such a region (e.g. Altcha) that manages its own non-directive `data-loading` value crashes the component on mount with `Unknown data-loading action "..."`.

Since the scan runs in `attachToComponent()` (`finishLoading(component, component.element)`), the component fails to initialise on page load and every subsequent live request is broken too.

This skips `data-loading` elements living inside a `data-live-ignore` subtree, the same way nested live-component elements are already skipped just above.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1895,6 +1895,7 @@ var LoadingPlugin_default = class {
 		const loadingDirectives = [];
 		let matchingElements = Array.from(element.querySelectorAll("[data-loading]"));
 		matchingElements = matchingElements.filter((elt) => elementBelongsToThisComponent(elt, component));
+		matchingElements = matchingElements.filter((elt) => !elt.closest("[data-live-ignore]"));
 		if (element.hasAttribute("data-loading")) matchingElements = [element, ...matchingElements];
 		matchingElements.forEach((element) => {
 			if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) throw new Error("Invalid Element Type");
--- a/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
+++ b/src/LiveComponent/assets/src/Component/plugins/LoadingPlugin.ts
@@ -178,6 +178,10 @@ export default class implements PluginInterface {
         // ignore elements which are inside a nested "live" component
         matchingElements = matchingElements.filter((elt) => elementBelongsToThisComponent(elt, component));
 
+        // ignore elements inside a "data-live-ignore" subtree: those are managed
+        // outside of Live (e.g. third-party widgets) and may set their own "data-loading"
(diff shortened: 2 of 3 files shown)


tomudding (approved):
Thanks! Changes work as expected.

Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3689, opened 2026-06-28
[TwigComponent] Fix test helper failing to render blocks in scoped templates

### LINKED ISSUE #3210: [TwigComponent] Test helper can't render blocks when limiting scope using `{% with … only%}`

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3210
| License        | MIT

When a component template restricts its scope with `{% with ... only %}`, `renderTwigComponent()` from `InteractsWithTwigComponents` fails with:

```
Twig\Error\RuntimeError: Variable "blocks" does not exist
```

The helper builds a wrapper template that injects the mock block content through a context variable:

```twig
{% component "My:Component" with data %}{% block slot1 %}{{ blocks.slot1|raw }}{% endblock %}{% endcomponent %}
```

But `{% with ... only %}` resets the context (`$context = $vars + [] + globals`, see `WithNode`), so the `blocks` variable is filtered out before the overridden block is rendered.

This PR inlines the mock content as a Twig string literal instead of passing it through a context variable, so it no longer depends on the rendering context and survives the scope reset. Single-quoted Twig strings don't interpolate `{{ ... }}` or `#{...}`, and `\` / `'` are escaped, so the content is rendered verbatim.

A reproducer was provided by the reporter: https://github.com/janopae/reproduce-symfony-ux-blocks-cant-be-rendered-in-tests

### DIFF
--- a/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
+++ b/src/TwigComponent/src/Test/InteractsWithTwigComponents.php
@@ -51,8 +51,15 @@ trait InteractsWithTwigComponents
 
         $template = \sprintf('{%% component "%s" with data %%}', addslashes($name));
 
-        foreach (array_keys($blocks) as $blockName) {
-            $template .= \sprintf('{%% block %1$s %%}{{ blocks.%1$s|raw }}{%% endblock %%}', $blockName);
+        foreach ($blocks as $blockName => $blockContent) {
+            // Inline the content as a Twig string literal instead of passing it
+            // through a context variable: a variable would be filtered out when
+            // the component template restricts the scope with `{% with ... only %}`.
+            $template .= \sprintf(
+                '{%% block %s %%}{{ \'%s\'|raw }}{%% endblock %%}',
+                $blockName,
+                strtr((string) $blockContent, ['\\' => '\\\\', '\'' => '\\\'']),
+            );
         }
 
         $template .= '{% endcomponent %}';
@@ -62,7 +69,6 @@ trait InteractsWithTwigComponents
             ->createTemplate($template)
             ->render([
                 'data' => $data,
-                'blocks' => $blocks,
             ])
         );
     }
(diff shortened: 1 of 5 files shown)


Kocal:
Without repeating myself, the 2.x version is in security maintenance only, you need to target the 3.x branch.

Kocal (commented):
Seems ok, @smnandre  is it fine to you?

smnandre:
I still don't think it's a good idea to change this implementation for everyone for this single use case. Not saying it's not legit, but it generates weird code, and does not provide any value (again, except this very particular case).

I'd still be inclined to suggest a dedicated way for this use case, but that's just my two cents.

Previous discussions here: https://github.com/symfony/ux/pull/3212 and https://github.com/symfony/ux/pull/3211

Amoifr:
Totally fair, and thanks for the patience across all three attempts. 🙏 You're right that inlining for everyone changes the shared path (and the cache behavior you flagged on #3211).

What about keeping the current implementation untouched as the default, and adding an opt-in dedicated to this case? Something like:

```php
$this->renderTwigComponent(name: 'MyComponent', blocks: [...], inlineBlocks: true);
```

`inlineBlocks: false` (default) keeps today's behavior verbatim, so no impact and no cache churn for existing tests. `true` switches to inlining, which is the only thing that survives a `{% with … only %}` scope. That way the "weird" code only ever runs for the specific case that needs it.

Does that match the dedicated way you had in mind, or would you rather shape the API differently? Happy to rework the PR along those lines. 🙂

Amoifr:
Closing this one: I agree the shared-path change isn't the right trade-off given the concerns raised here and on #3211/#3212. Thanks a lot @smnandre and @Kocal for the thoughtful reviews. 🙏

If the dedicated opt-in idea sounds worthwhile down the line, I'm happy to open a fresh PR scoped to just that. Appreciate the open source work as always! 🙂

PR #3690, opened 2026-07-01
[LiveComponent] Fix support for nullable collection properties on component

rogierknoester:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | - 
| License        | MIT

Currently it is not possible to have a nullable array liveprop on a live component. During dehydration it fails because it doesn't inspect the inner type, e.g. the following component:

```php
class Foo {
    /**
     * @var null|list<string>
     */
    #[LiveProp(useSerializerForHydration: true)]
    public ?array $items = null;
}
```

This is because in the [LiveComponentMetadataFactory](https://github.com/symfony/ux/blob/3.x/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php#L89) the if-statement falls through and starts using the reflection type. This resolves in a `mixed[]` instead of `null|list<string>`. This PR fixes that by checking if the inner type of a nullable type might be a collection type.

### DIFF
--- a/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
+++ b/src/LiveComponent/src/Metadata/LiveComponentMetadataFactory.php
@@ -86,7 +86,7 @@ class LiveComponentMetadataFactory implements ResetInterface
 
         $infoType = $this->propertyTypeExtractor->getType($className, $property->getName());
 
-        if ($infoType instanceof CollectionType) {
+        if ($infoType?->isSatisfiedBy(static fn ($t): bool => $t instanceof CollectionType)) {
             // If it's an "advanced" type (like CollectionType), let's use the PropertyTypeExtractor to get the Type
             $type = $infoType;
         } elseif (null !== $reflectionType) {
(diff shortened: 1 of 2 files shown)


SHANHE2024 (approved):

sinaahmadpour-ticketswap (approved):

Kocal (approved):
Seems nice, @mtarld can you confirm this is the way to go to check for a nullable collection type?

Thank you

mtarld:
@Kocal, I'd have done this instead:
```php
if ($infoType->isSatisfiedBy(static fn ($t): bool => $t instanceof CollectionType)) {
```

rogierknoester:
> @Kocal, I'd have done this instead:
> 
> ```
> if ($infoType->isSatisfiedBy(static fn ($t): bool => $t instanceof CollectionType)) {
> ```

I like that more, also covers more scenarios, let me update it to that :)

mtarld (approved):

Kocal:
Thanks @mtarld!

Kocal:
Thank you @rogierknoester.

PR #3692, opened 2026-07-04
 [TwigComponent] Reduce per-render CPU cost of static components

javiereguiluz:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

Pages that render hundreds of small components (e.g. an EasyAdmin index page renders ~630 components of 15 types: icons, table cells, action-menu items, ...) spend a significant share of their CPU time inside TwigComponent's own render pipeline rather than in the templates. Profiling such a workload showed three recurring per-render costs that this PR removes, without changing any observable behavior: the rendered HTML is byte-identical before/after each commit.

This PR contains three independent commits:

**1. Cache `ComponentMetadata` instances in `ComponentFactory`**

`metadataFor()` constructed a new `ComponentMetadata` on every call, and it is called twice per render (`create()` and `ComponentRenderer::preRender()`). The class is immutable, so instances are now cached per component name (~1,260 fewer allocations on the page above). The cache is cleared in `reset()` like the other per-request memo caches.

**2. Skip event allocation/dispatch when nobody listens**

Every render dispatched 5 freshly-allocated events (`PreCreateForRender`, `PreMount`, `PostMount`, `PreRender`, `PostRender`) even with zero listeners registered, the common case in production when the LiveComponent bridge is not installed (~3,150 event allocations + dispatches per request on the page above).

When the injected dispatcher implements the introspectable `Symfony\Component\EventDispatcher\EventDispatcherInterface`, `hasListeners()` is now checked (a cheap `!empty()` array lookup) before creating and dispatching each event:

- the check is done per dispatch call, so listeners registered at runtime keep working;
- any other PSR-14 dispatcher keeps the previous unconditional behavior;
- `PreRenderEvent` is still always created, because it carries the resolved template and variables to the compiled code; only its dispatch is skipped;
- when `PostMountEvent` is skipped, the extra metadata defaults to `[]`, exactly as an undispatched event would report.

The LiveComponent test suite (whose bridge subscribes to these events) passes unchanged against this branch.

**3. Build the render variables in a single pass in `preRender()`**

The variables array was spread-copied twice per render: once to create the `PreRenderEvent`, then a second full `[...spread]` to append `this`, `computed`, `outerScope`, `__props` and `__context`. The second copy is replaced with in-place assignments, the variables are only re-read from the event when it was actually dispatched, and the `array_diff_key()` call is skipped when there is no outer context (the `{{ component() }}` function path).

### Benchmark

Throw-away Symfony app (`prod` env, no profiler, OPcache on, Xdebug off) rendering a page of **642 components of 15 types** modeled on a real EasyAdmin index-page profile: 225 class-backed icons, 161 nested anonymous items with `{% props %}` defaults, 100 anonymous table cells with attributes, 23 dropdowns with named blocks, 40 renders via the `component()` function, etc. Numbers are the median of 5 processes × 30 measured iterations each (5 warmup), `hrtime()` around a full page render, on PHP 8.5 / Apple Silicon.

| Step (cumulative)            | p50 / page | Δ vs 3.x
| ---------------------------- | ---------- | --------
| 3.x baseline                 | 6.16 ms    | —
| + metadata cache             | 6.02 ms    | −2.3 %
| + event guards               | 5.54 ms    | −10.1 %
| + single-pass variables      | 5.36 ms    | **−13.1 %**

The rendered HTML is byte-identical (same SHA-1) for every step, and both the TwigComponent (299 tests) and LiveComponent (387 tests) suites pass.

### DIFF
--- a/src/TwigComponent/src/ComponentFactory.php
+++ b/src/TwigComponent/src/ComponentFactory.php
@@ -13,6 +13,7 @@ namespace Symfony\UX\TwigComponent;
 
 use Psr\Container\ContainerInterface;
 use Psr\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface as IntrospectableEventDispatcherInterface;
 use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
 use Symfony\Contracts\Service\ResetInterface;
 use Symfony\UX\TwigComponent\Event\PostMountEvent;
@@ -34,6 +35,12 @@ final class ComponentFactory implements ResetInterface
     private array $metadata = [];
 
     /**
+     * Set when the dispatcher can be asked whether an event has listeners,
+     * allowing to skip creating and dispatching events nobody listens to.
+     */
+    private readonly ?IntrospectableEventDispatcherInterface $introspectableDispatcher;
+
+    /**
      * @param array<string, array>        $config
      * @param array<class-string, string> $classMap
      */
@@ -46,6 +53,7 @@ final class ComponentFactory implements ResetInterface
         private readonly array $classMap,
         private readonly Environment $twig,
     ) {
+        $this->introspectableDispatcher = $eventDispatcher instanceof IntrospectableEventDispatcherInterface ? $eventDispatcher : null;
     }
 
     public function metadataFor(string $name): ComponentMetadata
@@ -98,8 +106,7 @@ final class ComponentFactory implements ResetInterface
     public function mountFromObject(object $component, array $data, ComponentMetadata $componentMetadata): MountedComponent
     {
         $originalData = $data;
-        $event = $this->preMount($component, $data, $componentMetadata);
-        $data = $event->getData();
+        $data = $this->preMount($component, $data, $componentMetadata);
 
         $this->mount($component, $data, $componentMetadata);
 
@@ -113,8 +120,7 @@ final class ComponentFactory implements ResetInterface
             }
         }
 
-        $postMount = $this->postMount($component, $data, $componentMetadata);
-        $data = $postMount->getData();
+        [$data, $extraMetadata] = $this->postMount($component, $data, $componentMetadata);
 
         // create attributes from "attributes" key if exists
         $attributesVar = $componentMetadata->getAttributesVar();
@@ -132,7 +138,7 @@ final class ComponentFactory implements ResetInterface
             $component,
             new ComponentAttributes([...$attributes, ...$data], $this->twig->getRuntime(EscaperRuntime::class)),
             $originalData,
-            $postMount->getExtraMetadata(),
+            $extraMetadata,
         );
     }
 
@@ -182,34 +188,46 @@ final class ComponentFactory implements ResetInterface
         $mount->invoke($component, ...$parameters);
     }
 
-    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): PreMountEvent
+    /**
+     * @return array The (possibly modified) mount data
+     */
+    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): array
     {
-        $event = new PreMountEvent($component, $data, $componentMetadata);
-        $this->eventDispatcher->dispatch($event);
+        if (null === $this->introspectableDispatcher || $this->introspectableDispatcher->hasListeners(PreMountEvent::class)) {
+            $event = new PreMountEvent($component, $data, $componentMetadata);
+            $this->eventDispatcher->dispatch($event);
+            $data = $event->getData();
+        }
 
-        $data = $event->getData();
         foreach ($componentMetadata->getPreMounts() as $preMount) {
             if (null !== $newData = $component->$preMount($data)) {
-                $event->setData($data = $newData);
+                $data = $newData;
             }
         }
 
-        return $event;
+        return $data;
     }
 
-    private function postMount(object $component, array $data, ComponentMetadata $componentMetadata): PostMountEvent
+    /**
+     * @return array{0: array, 1: array} The (possibly modified) mount data and the extra metadata
+     */
+    private function postMount(object $component, array $data, ComponentMetadata $componentMetadata): array
     {
-        $event = new PostMountEvent($component, $data, $componentMetadata);
-        $this->eventDispatcher->dispatch($event);
+        $extraMetadata = [];
+        if (null === $this->introspectableDispatcher || $this->introspectableDispatcher->hasListeners(PostMountEvent::class)) {
+            $event = new PostMountEvent($component, $data, $componentMetadata);
+            $this->eventDispatcher->dispatch($event);
+            $data = $event->getData();
+            $extraMetadata = $event->getExtraMetadata();
+        }
 
-        $data = $event->getData();
         foreach ($componentMetadata->getPostMounts() as $postMount) {
             if (null !== $newData = $component->$postMount($data)) {
-                $event->setData($data = $newData);
+                $data = $newData;
             }
         }
 
-        return $event;
+        return [$data, $extraMetadata];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/TwigComponent/src/ComponentFactory.php
```diff
@@ -175,34 +188,46 @@ private function mount(object $component, array &$data, ComponentMetadata $compo
...
 
-    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): PreMountEvent
+    /**
+     * @return array The (possibly modified) mount data
+     */
+    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): array
     {
-        $event = new PreMountEvent($component, $data, $componentMetadata);
-        $this->eventDispatcher->dispatch($event);
+        if (null === $this->introspectableDispatcher || $this->introspectableDispatcher->hasListeners(PreMountEvent::class)) {
```
Kocal, line 196: Do you think we can check if we have listeners for events for a specific component class?

I mean, if we have 100 anonymous components (that don't have listener AFAIK), and one component with listener, we will still dispatch the events right?

Note that we have two specials component `ux:icon` and `ux:map`, and I think they listen to some events... will it still work? 🤔

javiereguiluz: I added a test related to events and static Twig components. `hasListeners()` is checked per event class, not per component, so once any listener is registered, events are dispatched for every component render, inlcuding anonymous. Also, `hasListeners()` is evaluated at dispatch time (not cached at construction), since the listeners are added at runtime.

I don't use `ux:map`, but I use `ux:icon` and I tested the changes in a page using 628 components (225 of them are icons) and it worked as expected.

kbond (approved):

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -57,12 +72,12 @@ public function metadataFor(string $name): ComponentMetadata
...
             ];
 
-            return new ComponentMetadata($this->config[$name]);
+            return $this->metadata[$name] = new ComponentMetadata($this->config[$name]);
         }
 
         if ($mappedName = $this->classMap[$name] ?? null) {
             if ($config = $this->config[$mappedName] ?? null) {
-                return new ComponentMetadata($config);
+                return $this->metadata[$name] = new ComponentMetadata($config);
```
smnandre, line 80:
```suggestion
                return $this->metadata[$mappedName] = new ComponentMetadata($config);
```

Not 100% sure but this may deserve some thinking

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -57,12 +72,12 @@ public function metadataFor(string $name): ComponentMetadata
                 'template' => $template,
             ];
 
-            return new ComponentMetadata($this->config[$name]);
+            return $this->metadata[$name] = new ComponentMetadata($this->config[$name]);
```
smnandre, line 75: We are keeping data twice here.. as array in $this->config[$name] and as instance in $metadata. 

Should we replace one by the other ?

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -175,34 +188,46 @@ private function mount(object $component, array &$data, ComponentMetadata $compo
         $mount->invoke($component, ...$parameters);
     }
 
-    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): PreMountEvent
+    /**
+     * @return array The (possibly modified) mount data
+     */
+    private function preMount(object $component, array $data, ComponentMetadata $componentMetadata): array
```
smnandre, line 194: what about using a void + reference, as in mount() method ?

src/TwigComponent/src/ComponentFactory.php
smnandre, line 196: Should we calculate this once and store result ? So we could avoid the entire method call ? 

```php
if (true === $this->skipPostMount) {
    $this->postMount(...) 
}
```

And in postMount we can always toggle it on first call ?

src/TwigComponent/src/ComponentFactory.php
```diff
@@ -175,34 +188,46 @@ private function mount(object $component, array &$data, ComponentMetadata $compo
...
+            $this->eventDispatcher->dispatch($event);
+            $data = $event->getData();
+            $extraMetadata = $event->getExtraMetadata();
+        }
 
-        $data = $event->getData();
         foreach ($componentMetadata->getPostMounts() as $postMount) {
             if (null !== $newData = $component->$postMount($data)) {
-                $event->setData($data = $newData);
+                $data = $newData;
```
smnandre, line 226: This may create problems for the profiler .. but this must not be a reason to reject your great optimizations here.

Kocal:
Any updates here?

src/TwigComponent/src/ComponentFactory.php
Kocal: Good question, but I'd rather keep both. `$this->config` is the raw array injected by the DI container and holds every registered component, and `throwUnknownComponentException()` relies on `array_keys($this->config)` to build the "did you mean" suggestions, so dropping it would mean instantiating a `ComponentMetadata` for every component just to list their names. `$this->metadata` is also a lazy per-request cache that `reset()` clears, while `$this->config` lives for the whole request, so they're not really duplicating the same thing. And since PHP arrays are copy-on-write, the config array is shared rather than duplicated in memory anyway.

src/TwigComponent/src/ComponentFactory.php
Kocal: That's a fair catch. The early return at the top of `metadataFor()` looks up the cache using the `$name` the caller passed, so if we keyed the cache on `$mappedName` instead, a class-name lookup would never hit it again and we'd rebuild the metadata every time. Keying on both `$name` and `$mappedName` would fix that and also avoid building a second instance when the same component gets resolved once by class name and once by its key. It's a small win though, happy to do it as a follow-up rather than block this PR on it.

src/TwigComponent/src/ComponentFactory.php
Kocal: It would work for `preMount()`, sure. But `postMount()` returns two values (the data and the extra metadata), and a tuple reads better there than two by-reference parameters, which is probably why both were left as returns. Happy to switch `preMount()` to void+reference if you feel strongly about it.

src/TwigComponent/src/ComponentFactory.php
Kocal: Listeners can be registered at any point during the request, not just at boot, so we need to check `hasListeners()` at dispatch time rather than cache the result on the first call. @javiereguiluz already explained the same thing further up in this PR. And `hasListeners()` is just an array lookup under the hood, so caching it wouldn't really save us much anyway.

src/TwigComponent/src/ComponentFactory.php
Kocal: Thanks for flagging this, I went and checked it properly: I instrumented the real `TwigComponentDataCollector` on both branches, rendered four components (one nested, one using `#[PreMount]`/`#[PostMount]`, one anonymous), and diffed the collected data. `component_count`, `render_count`, `components`, `renders`, the `children` tree, `depth`, `input_props` and `attributes` all come out identical, because `TwigComponentLoggerListener` only subscribes to `PreRenderEvent`/`PostRenderEvent`, so `hasListeners()` is true whenever the profiler is on and nothing gets skipped. There's one real behavior change: a listener that keeps the event object and calls `getData()` after the dispatch now sees the data from before the `#[PreMount]`/`#[PostMount]` methods ran, for example `PreMountEvent::getData()` returns `value => 'x'` instead of `value => 'pre-mount x'`. Nothing in symfony/ux does that though (the three LiveComponent mount subscribers all read the data during the dispatch), and the mount methods already ran after the dispatch even before this PR, so even a low-priority listener wouldn't be affected.

PR #3703, opened 2026-07-10
[StimulusBundle] Sort custom controllers to keep the generated loader deterministic

### LINKED ISSUE #3562: [StimulusBundle] ControllersMapGenerator::loadCustomControllers() returns non-deterministic controller order
**Description**

`ControllersMapGenerator::loadCustomControllers()` uses Symfony's `Finder` component without any explicit sort order. On Linux filesystems (ext4/btrfs), directory entries are stored in a hash tree, so iteration order is non-deterministic and can change when files are added or removed. This causes the generated `controllers` asset to have a different order across builds, producing different asset fingerprints even when no controller code has changed — breaking cache busting stability.

**Steps to Reproduce**

1. Have multiple custom Stimulus controllers under controllers
2. Run `bin/console asset-map:compile` twice
3. Observe that the order of controllers in the generated asset map differs between runs (visible in `public/assets/` output or by inspecting the generated JS bundle)

**Expected Behavior**

Controllers should always be emitted in a consistent, deterministic order (e.g. sorted by name).

**Actual Behavior**

Controller order varies between runs depending on the filesystem's internal directory hash state.

**Root Cause**

In [ControllersMapGenerator.php](https://github.com/symfony/ux/blob/2.x/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php), the `Finder` in `loadCustomControllers()` has no sort applied:

```php
$finder->in($this->controllerPaths)
    ->files()
    ->name(self::FILENAME_REGEX);
    // no ->sortByName() here
```

**Proposed Fix**

Add `->sortByName()` to the `Finder` chain:

```php
$finder->in($this->controllerPaths)
    ->files()
    ->name(self::FILENAME_REGEX)
    ->sortByName();
```

Note: `loadUxControllers()` is not affected since its order is determined by the static `controllers.json` file.

**Environment**

- `symfony/stimulus-bundle`: 2.x
- PHP: 8.3
- OS: Linux

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #3562
| License       | MIT

`ControllersMapGenerator::loadCustomControllers()` iterates a `Finder` without any sort:

```php
$finder->in($this->controllerPaths)
    ->files()
    ->name(self::FILENAME_REGEX);
```

Directory iteration order is filesystem-dependent (ext4/btrfs store entries in a hash tree), so the custom controllers land in the map in an arbitrary order. `StimulusLoaderJavaScriptCompiler` then emits the `import` statements in that same order, so the generated loader's contents — and therefore its digest — change between builds even though no controller changed. Cache busting is defeated.

Adding `->sortByName()` makes the order deterministic. Thanks @ywjdlq for the report, which already contained the root cause and the fix.

### Tests

`testCustomControllersAreSortedByName` asserts the order of the custom controllers in the map. It genuinely fails without the fix on my machine (ext4): the fixtures come back as `hello, hello-with-dashes, …, bye, typescript`, with `bye_controller.js` next to last.

### Note

The same code is on `2.x`. I targeted `3.x` since that is what `CONTRIBUTING.md` points at, but I'm happy to retarget if you'd rather have it on the older branch.

### DIFF
--- a/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
+++ b/src/StimulusBundle/src/AssetMapper/ControllersMapGenerator.php
@@ -66,7 +66,10 @@ class ControllersMapGenerator
         $finder = new Finder();
         $finder->in($this->controllerPaths)
             ->files()
-            ->name(self::FILENAME_REGEX);
+            ->name(self::FILENAME_REGEX)
+            // the filesystem iteration order is not stable, sort to keep the
+            // generated controllers loader (and so its digest) deterministic
+            ->sortByName();
 
         $controllersMap = [];
         foreach ($finder as $file) {
--- a/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php
+++ b/src/StimulusBundle/tests/AssetMapper/ControllersMapGeneratorTest.php
@@ -119,4 +119,49 @@ class ControllersMapGeneratorTest extends TestCase
         $minifiedController = $map['minified'];
         $this->assertTrue($minifiedController->isLazy);
     }
+
+    public function testCustomControllersAreSortedByName()
+    {
+        $mapper = $this->createMock(AssetMapperInterface::class);
+        $mapper->expects($this->any())
+            ->method('getAssetFromSourcePath')
+            ->willReturnCallback(static function ($path) {
+                // replace windows slashes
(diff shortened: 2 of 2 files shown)


Kocal (approved):

Kocal:
Thank you @Amoifr.

Issue #3711, opened 2026-07-18
[Toolkit] Make kits self-describing so ux.symfony.com renders them with zero per-kit code


Kocal:
## Problem

A kit lives in `symfony/ux` (`src/Toolkit/kits/<kit>/`), but `ux.symfony.com` **hardcodes** every kit. Adding one means 8 hand-edits, plus 1-2 per recipe:

`ToolkitKitId` enum (`case` + `color()`), `assets/icons/toolkit/<kit>.svg`, `assets/styles/toolkit-<kit>.css` (Tailwind entry + theme tokens, 238 lines for Flowbite), `assets/toolkit-<kit>.js` (import + `app.register()` each controller), `importmap.php`, `symfonycasts_tailwind.yaml` + `composer.json` tailwind list (duplicated), `composer.lock`, and one `templates/toolkit/docs/<kit>/<recipe>.md.twig` per recipe.

Kit identity is declared twice, controllers listed three times, docs written 1:1 by hand. Community kits get none of this: no preview, no rendered docs.

## Goal

Make a kit fully self-describing: metadata, theme, docs, and rendering assets all ship inside the kit (or inside Toolkit). `ux.symfony.com` becomes a generic renderer with **zero per-kit code**. Community-kit preview is out of scope here, but this is the first step toward it (a future `bin/ux-toolkit-kit-preview`).

## Key insight

`renderComponentDoc()` renders `.md.twig` with Twig into a markdown string, then parses it with CommonMark. The `.md.twig` tangles three things:

- **directives** (`::: tabs`, `> [!NOTE]`, preview) -> portable, host-agnostic
- **scaffolding** (`_base_component.md.twig`: install tabs, API table) -> logic over the `Kit`/`Recipe` model
- **host rendering** (the signed preview iframe) -> the only genuinely site-specific bit

`toolkit_code_example()` is just sugar expanding to a `::: tabs` block with a Preview and a Code tab. The fix is to separate these three.

### 1. Markdown extensions inside Toolkit (`src/Toolkit/src/Markdown/`)

For a first iteration these live in Toolkit (`Symfony\UX\Toolkit\Markdown\`), not a separate package; the site consumes them from the vendored `symfony/ux-toolkit`. `league/commonmark` is a **dev-only** Toolkit dependency (guarded by `class_exists`), so installing Toolkit to scaffold components never pulls it in; hosts that render docs provide it. Standalone League CommonMark extensions, added à la carte to your own converter (no all-in-one factory):

```php
[...]

PR #3718, opened 2026-07-19
[Toolkit] Add a documentation rendering layer so kits can describe themselves

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Part of #3711
| License        | MIT

First of two PRs implementing #3711: make kits self-describing so ux.symfony.com can render them with zero per-kit code.

## Why

Right now a UX Toolkit kit lives in symfony/ux, but ux.symfony.com hardcodes everything about it. Adding a kit means about 8 hand-edits on the site side plus per-recipe `.md.twig` docs that only the site knows how to render.

Community kits get none of this: no preview, no rendered docs, nothing.

This PR moves the rendering layer into Toolkit itself, so a kit can describe its own docs, and migrates all 63 existing recipes to the new format. The follow-up PR turns ux.symfony.com into a generic renderer on top of it. No ux.symfony.com changes are in this PR.

## What's in this PR

- **Per-recipe `README.md`, self-contained.** A README.md is a complete document: it opens with the recipe's title as a single level-1 heading (`# Alert`), followed by the description, then prose with generated pieces dropped in via directives: `::: example <Name>` embeds a live preview of `examples/<Name>.html.twig`, `::: installation` inserts the install steps, `::: api-reference` inserts the props/blocks tables generated from the component docblocks. This mirrors how Shadcn UI writes its component doc pages. It also renders as-is when browsing the recipe folder on GitHub. A recipe without a `README.md` falls back to a default layout built from its manifest.
- **All 63 existing recipes (common, shadcn, flowbite-4) get a `README.md`.** Generated with the `MdTwigConverter` below from the existing ux.symfony.com `.md.twig` docs, then reviewed by hand.
- **Recipe manifest drops `description`.** Since the title and description now live in the README.md, the field was duplicated. Removed from `RecipeManifest`, the recipe schema, the kit scaffolding, and all 63 existing recipe `manifest.json`. The kit manifest keeps its own `description` (it's the kit's catalog entry). The recipe `name` stays, it's used by the `ux:install` picker and, on the site, for navigation and listings. `Recipe::getDescription()` fills the gap: it parses the recipe's short description from its README.md's first paragraph, used for catalog blurbs and llms.txt.
- **`DocHeadingLevelChecker`** (linter), enforces the README.md shape: exactly one level-1 heading (the title), every heading after that at level 2 or deeper. Keeps a rendered page to a single title.
- **`RecipeDocRenderer`**, with two output modes: `renderAsHtml()` goes through league/commonmark plus Toolkit's own CommonMark extensions and returns a `RenderedRecipeDoc` value object carrying both the rendered HTML and the page's table of contents (extracted from the CommonMark AST during the same render, so the host doesn't have to re-parse or regex-scrape the output) plus tabs + an iframe live preview, meant for ux.symfony.com. `renderAsMarkdown()` produces portable Markdown with examples resolved to fenced code and no interactive directives, meant for LLM consumption or plain rendering.
- **The Markdown extensions behind the directives**, ported into Toolkit and rendered through Twig templates the kit ships: Alert, Tabs, Popover, CodePreview (static or iframe via a `PreviewUrlGenerator` interface the site implements), the `::: example` directive, and a `{"preview": true}` fenced-code form.
- **No-build Stimulus controllers** (tabs, popover, clipboard) and base CSS, shipped with Toolkit via the AssetMapper no-build convention, so the rendered HTML is interactive without a build step.
- **Optional `color` and `icon` fields** on the kit manifest (schema updated), so a kit carries its own identity instead of ux.symfony.com hardcoding it.
- **`LocalRegistry` gained `getAvailableKits()` and `exists()`**, so a host can enumerate the vendored kits and check one exists without hardcoding a kit list.
- **`MdTwigConverter`** (internal, one-shot): turned the existing ux.symfony.com `.md.twig` recipe docs into `README.md` for the 63-recipe migration above. Its output was a reviewed starting point, not a fully automated migration.

## What a recipe's README.md looks like

```markdown
# Alert

Displays a callout for user attention.

::: example Demo

## Installation

::: installation

## Usage

::: example Usage

## API Reference

::: api-reference
```

## Notes

`league/commonmark` (^2.8) is added as a dev-only dependency, guarded by `Assert::commonMarkAvailable()`. It's only needed for HTML rendering, and the app that renders (ux.symfony.com) already depends on it.

This PR is a draft, kept open while the companion ux.symfony.com PR (renderer) lands. CHANGELOG entry included.


src/Toolkit/src/Doc/MdTwigConverter.php
```diff
@@ -0,0 +1,94 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Toolkit\Doc;
+
+/**
+ * One-shot helper to migrate ux.symfony.com's `<kit>/<recipe>.md.twig` docs into a `README.md`: it turns
```
kbond, line 15: So this can be deleted after this migration is done?

src/Toolkit/kits/common/post-link/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Examples
+
+### Custom Method
+
+Set the `method` prop to submit the form with a spoofed HTTP method. A hidden `_method` field is added so Symfony can route the request to the matching controller.
+
+> [!WARNING]
+> Method spoofing only works when HTTP method override is enabled in your Symfony app. Set `framework.http_method_override: true` in `config/packages/framework.yaml`.
+
+::: example Custom Method {"height": "300px", "collapseClass": true}
```
kbond: What do you think about being able to inline the examples?

Something like:

````suggestion
```twig : example Custom Method {"height": "300px", "collapseClass": true}
<twig:PostLink href="/link/posts/42" method="DELETE" class="rounded-md bg-red-600 px-4 py-2 text-sm font-medium text-white hover:bg-red-500">
    Delete post
</twig:PostLink>
```
````

I think most markdown readers could render? It would look nicer to someone reading the raw readme.

**Review by kbond — COMMENTED** [review]
Love it! The Claude skill will need to be updated?

File: src/Toolkit/src/Doc/MdTwigConverter.php (hunk 1 above)
Comment (Kocal) [reply src/Toolkit/src/Doc/MdTwigConverter.php:15]: Yes, but after #3713 is merged

File: src/Toolkit/kits/common/post-link/README.md (hunk 2 above)
Comment (Kocal) [reply src/Toolkit/kits/common/post-link/README.md]: That was the format used before https://github.com/symfony/ux/pull/3299, I migrated because AFAIK the code highlighting was pretty bad (in PHPStorm or GitHub...? I don't really remember) and parsing code-blocks like that was a bit fragile for our previous needs, but maybe we don't have these needs anymore.

If we go for something like that, your suggestion would be something like that:
```markdown
### Custom Method

```twig {"preview": true, "height": "300px", "collapseClass": true}
<twig:PostLink href="/link/posts/42" method="DELETE" class="rounded-md bg-red-600 px-4 py-2 text-sm font-medium text-white hover:bg-red-500">
    Delete post
</twig:PostLink>
```.
```
(without the `.` after the triple `)

Kocal: Ah, hem, if we re-inline Twig code in the `.md` file, we will no longer be able to run Twig-CS-Fixer on them

kbond: I see... Didn't realize you moved away from that.

Even if the rendering isn't great, it would be easier to both maintain and read as a user (who is just looking at the raw readme)

Kocal: Yes totally, I will investigate

kbond: I just tested in a few places:
- GitHub: works as expected (extra json is ignored)
- PhpStorm: marked as a code block but no highlighting (extra json is ignored)

Kocal: Yeah only GitHub correctly highlight ```twig: 
image
image

The options as JSON is not standard AFAIK, so the fact it's not highlighted is normal. 

However, we can do a trick by using MDX putting code inside a fake _JSX component_ `<Example>`, where we can have some color highlighting, but it's super fragile, not adapted, and IDK if GitHub support it:
image

Kocal: IMHO, let's try to re-use inlined Twig code examples ``` in the `README.md` files, and I will see if Twig-CS-Fixer can be programmatically run on inlined code blocks

kbond: You mean try the json version?

I don't think it's that big of a deal it doesn't render in phpstorm. It's cool it does in GitHub though!

kbond: I'm just happy it doesn't break rendering elsewhere

Kocal: Yes, this one https://github.com/symfony/ux/pull/3718#discussion_r3615893461

Kocal:
Well... let's stop here 😇 
image

src/Toolkit/kits/common/post-link/README.md
Kocal: Done, examples files are now inlined in <recipe>/README.md files, snapshots were re-generated (their names changed too). 

For example, the `src/Toolkit/kits/common/logout-link/README.md` file now looks like this, and is self-sufficient to correctly renders https://127.0.0.1:9044/toolkit/kits/common/components/logout-link:
````md
# Logout Link

A link that logs the current user out through a secure POST form.

```twig {"preview":true,"height":"300px","collapseClass":true}
<twig:LogoutLink class="rounded-md bg-violet-600 px-4 py-2 text-sm font-medium text-white hover:bg-violet-500">
    Logout
</twig:LogoutLink>
```

## Installation

::: installation

## Usage

```twig
<twig:LogoutLink>
    Logout
</twig:LogoutLink>
```

> [!WARNING]
> `LogoutLink` logs the user out through a **POST** request protected by a CSRF token (token id `logout`). For it to work, your firewall's logout must **require `POST`** and have **CSRF protection enabled** — otherwise logging out will be rejected.

Restrict the logout route to `POST` so it can't be triggered by a plain link or prefetch:

```php
#[Route('/logout', name: 'app_logout', methods: ['POST'])]
public function logout(): never
{
    throw new \LogicException('This method is intercepted by the logout key on your firewall.');
}
```

Point the firewall's logout at that route (`app_logout`) and enable CSRF protection (this validates the `logout` token the component sends):

```yaml
# config/packages/security.yaml
security:
    firewalls:
        main:
            logout:
                path: app_logout
                enable_csrf: true
```

## Examples

### Specific Firewall

Set the `firewall` prop to log out from a specific firewall instead of the current one.

```twig {"preview":true,"height":"300px","collapseClass":true}
<twig:LogoutLink firewall="main" class="rounded-md bg-violet-600 px-4 py-2 text-sm font-medium text-white hover:bg-violet-500">
    Logout from the main firewall
</twig:LogoutLink>
```

## API Reference

::: api-reference

````

kbond (approved):
Lots of changes but many deletions! This will make it much easier to add recipes to kits!

I think the Claude skill needs to be updated before merging?

Kocal:
> I think the Claude skill needs to be updated before merging?

Yes! Nice catch. I also updated the command to create a kit

Kocal:
Here we go

PR #3596, opened 2026-05-25
[Toolkit] Finalize support for Blocks, add Shadcn's login-01 and login-02 blocks

seb-jean:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix https://github.com/symfony/ux/issues/3246 and continue https://github.com/symfony/ux/pull/3282
| License        | MIT

https://github.com/user-attachments/assets/69f84c90-46f3-4294-8de7-271728aa66b1

PR #3722, opened 2026-07-23
[Toolkit] Remove the unused recipe manifest `type` property

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| License       | MIT

Every recipe manifest declared `"type": "component"`, and no code path actually depended on the value. The only consumers were `Kit::getRecipes()`/`getRecipe()`, whose `RecipeType` filter was used in exactly one place — `KitContextRunner`, to select recipes when wiring Twig component resolution. Iterating all recipes there is equivalent: a recipe with no component templates contributes nothing to resolution and simply falls through. The `block` case existed only in in-memory unit-test fixtures — no real recipe ever used it. And the `create-kit` scaffolder already omitted `type` from the manifest it generates, so requiring the property was internally inconsistent.

So the field carried no information that isn't already implicit in a recipe's files. This removes it:

- Delete the `RecipeType` enum and drop the `type` property from `RecipeManifest`.
- Simplify `Kit::getRecipes()`/`getRecipe()` to drop the type filter; `KitContextRunner` now iterates all recipes.
- Remove `type` from `schema-kit-recipe-v1.json` and from all 63 kit manifests.

This clears the way for controller-only recipes (e.g. a forthcoming `closeable` recipe in the `common` kit) that ship a Stimulus controller and no Twig component — a shape the `component`/`block` enum had no honest slot for.


Kocal (changes requested):
I didn't had time to work on it, but the `type` was supposed to be `component` or `block` (https://github.com/symfony/ux/issues/3246) , like https://ui.shadcn.com/blocks, https://flowbite.com/blocks/, https://getbootstrap.com/docs/5.3/examples/#custom-components, etc..

So 👎🏻 for me

Kocal:
See open PR https://github.com/symfony/ux/pull/3596

kbond:
So a recipe that's purely a stimulus recipe, should be "component"?

Kocal:
Mmmh, let's say that big compound recipes are blocks, and others are components, yep!
It's related to [Atomic Design Methodology](https://atomicdesign.com/chapter-2), but we only have components and blocks (templates).

Note that ux.symfony.com consumes `Kit::getRecipes(RecipeType::Component)` at multiple places:

image

Everything is ready for building blocks, I just didn't have time to continue to work on them.

PR #3724, opened 2026-07-23
[Toolkit] Document Stimulus controllers in the recipe API reference

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Recipes that ship a Stimulus controller had no way to surface its public API in their documentation. This teaches `RecipeDocRenderer` to parse a controller's `@value`/`@target`/`@action` docblock tags and render them under `::: api-reference`, alongside the existing Twig component tables. The code stays the source of truth for value names/types and target names; the docblocks add descriptions, and actions are opt-in via `@action` so internal methods stay private.

Documentation is opt-in: the API reference renders only for controllers that actually carry docblock tags. A controller with no `@value`/`@target`/`@action` tag produces no API reference at all, even though its values and targets could be read from the code — nothing is documented until you choose to.

A new `StimulusControllerDocChecker` keeps those docblocks honest — descriptions capitalized and period-terminated, `@value`/`@target` tags consistent with the `static values`/`static targets` declarations in both directions, and every `@action` matching a real method.

### DIFF
--- a/src/Toolkit/src/Component/StimulusControllerAction.php
+++ b/src/Toolkit/src/Component/StimulusControllerAction.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Toolkit\Component;
+
+/**
+ * A Stimulus action method documented with an `@action` tag.
+ *
+ * Actions are opt-in: only methods flagged with `@action` are treated as public API, so internal
+ * (but still public) helper methods stay out of the reference.
+ *
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @internal
+ */
+final class StimulusControllerAction
+{
+    public function __construct(
+        public readonly string $name,
+        public readonly string $description,
+    ) {
+    }
+}
--- a/src/Toolkit/src/Doc/RecipeDocRenderer.php
+++ b/src/Toolkit/src/Doc/RecipeDocRenderer.php
@@ -24,6 +24,9 @@ use Symfony\Component\Filesystem\Path;
 use Symfony\UX\Toolkit\Assert;
 use Symfony\UX\Toolkit\Component\ComponentDoc;
 use Symfony\UX\Toolkit\Component\ComponentDocParser;
+use Symfony\UX\Toolkit\Component\StimulusController;
+use Symfony\UX\Toolkit\Component\StimulusControllerDoc;
+use Symfony\UX\Toolkit\Component\StimulusControllerDocParser;
 use Symfony\UX\Toolkit\Installer\PoolResolver;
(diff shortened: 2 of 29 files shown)


kbond:
Example of what it'll look like on ux.symfony.com

Screenshot 2026-07-23 at 4 24 16 PM

Kocal:
Since Stimulus Controllers are wrapped behind Twig components, is it really "useful" to document them? 

For example, If I can use the Accordion component from Shadcn kit, I will use the component itself and not the Stimulus controller behind it

kbond:
That's a fair point. And we can close #3726 but it is useful in https://github.com/symfony/ux/pull/3725

Kocal:
Ah I didn't think about controllers-only recipes, fair point too!

src/Toolkit/src/Component/StimulusControllerAction.php
Kocal, line 1: WDYT about prefixing all classes `Controller*` by `Stimulus`?

kbond:
> Ah I didn't think about controllers-only recipes, fair point!

But maybe they should be opt-in somehow. So we don't show the api for the ones backed by a Twig component.

1. An `@api` tag on the controller opts-in?
2. Just having the docs opts-in?

src/Toolkit/src/Doc/RecipeDocRenderer.php
```diff
@@ -199,4 +205,45 @@ private function extractApiReference(Recipe $recipe): array
...
+    /**
+     * Mirrors Stimulus' filename-to-identifier convention: drop the `_controller.js` suffix, turn
+     * nested directories into `--` and remaining underscores into `-`.
+     */
+    private static function stimulusIdentifier(string $source): string
+    {
+        $path = substr($source, strpos($source, 'assets/controllers/') + \strlen('assets/controllers/'), -\strlen('_controller.js'));
+
+        return str_replace(['/', '_'], ['--', '-'], $path);
+    }
```
Kocal: I think we already have this method somewhere, let's re-use it?

Kocal (approved):
That's really cool, just minor comments

kbond:
Ok, I made the following updates:

- Made the API reference opt-in — renders only when the controller carries docblocks.
- Updated the `symfony-ux-toolkit-kit` skill
- Applied review (Kocal + Copilot)

kbond:
@Kocal will Copilot just continue to find issues? I fixed the first round but now there's more 🤷🏻

Kocal:
> @Kocal will Copilot just continue to find issues? I fixed the first round but now there's more 🤷🏻

Ahah omg Copilot is on fire 😂 
Did you enable it automatically to run at each push? 

IMHO, https://github.com/symfony/ux/pull/3724#discussion_r3642210414 and https://github.com/symfony/ux/pull/3724#discussion_r3642210392 are worth fixing, and then we can merge

kbond:
It ran on each push automatically. It's a repo setting?

These two remaining things are similar to fixes to the same code earlier. I've been using Claude to review copilot's review and fix. Maybe I should just accept the copilot suggestions? Maybe Claude was making the fix in a way copilot didn't like

PR #3726, opened 2026-07-23
[Toolkit] Document the Stimulus API of existing kit controllers

kbond:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Applies the new controller docblocks to every controller that ships one across the `flowbite-4` and `shadcn` kits, so each recipe's documentation gains a Stimulus API reference. Values and targets are documented in full (the linter enforces both directions); actions are limited to the public methods actually wired via `data-action` in each recipe's templates.

Depends on #3724.


kbond:
I'm closing per https://github.com/symfony/ux/pull/3724#issuecomment-5063236957

PR #3727, opened 2026-07-24
[LiveComponent] Fix URL-bound array props ignored when documented with a phpdoc shape or generic

### LINKED ISSUE #2900: [LiveComponent] `LiveProp(url: true)` does not accept urls
From https://github.com/symfony/ux/pull/2607#issuecomment-3018968718

I have this property on component:
```php
    /**
     * @var array{r: array<string, true>, c: array<string, true>}
     */
    #[LiveProp(url: new UrlMapping('e'))]
    public array $expanded = ['r' => [], 'c' => []];
```

My url is `/tree?expanded[r][C-D]=1` and now it does not properly hydrate because `TypeHelper::accepts($type, $value)` in `QueryStringPropsExtractor::isValueTypeConsistent` return false.

 The problem I see is that both `r` and `c` in `$expanded` are set as `optional: false`:
```php
Symfony\Component\TypeInfo\Type\ArrayShapeType {#5071 ▼
  -type: Symfony\Component\TypeInfo\Type\GenericType {#5157 ▼
    -variableTypes: array:2 [▼
      0 => Symfony\Component\TypeInfo\Type\BuiltinType {#5159 ▼
        -typeIdentifier: Symfony\Component\TypeInfo\TypeIdentifier {#5052 ▶}
      }
      1 => Symfony\Component\TypeInfo\Type\CollectionType {#5068 ▶}
    ]
    -type: Symfony\Component\TypeInfo\Type\BuiltinType {#5158 ▶}
  }
  -isList: false
  -shape: array:2 [▼
    "c" => array:2 [▼
      "type" => Symfony\Component\TypeInfo\Type\CollectionType {#5070 ▶}
      "optional" => false
    ]
    "r" => array:2 [▼
      "type" => Symfony\Component\TypeInfo\Type\CollectionType {#5068 ▶}
      "optional" => false
    ]
  ]
  -extraKeyType: null
  -extraValueType: null
}
```

So when it tries to set only `r` part this shape is not accepted anymore. This was previously working fine.
The functionality is simple:
 - I render a tree;
 - Clicking on first level entries saves state to `$expanded = ['r' => ...]`;
 - Clicking on second level entries save state to `$expanded = ['c' => ...]`;
 - Clicking is handling via two live actions which updates only `r` or `c`.

--

Tried to split into multiple properties:

```php
    /**
     * @var array<string, true>
     */
    #[LiveProp(url: new UrlMapping('r'))]
    public array $expandedRoots = [];

    /**
     * @var array<string, true>
     */
    #[LiveProp(url: new UrlMapping('c'))]
    public array $expandedCategories = [];
```

still `TypeHelper::accepts` returns false on url `/tree?r[A-B]=1&c[01713ef6-eced-7e20-ab12-008458caf697]=1`

_Originally posted by @norkunas in https://github.com/symfony/ux/issues/2607#issuecomment-3018968718_

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2900
| License       | MIT

Since the TypeInfo migration (#2607), `RequestPropsExtractor::isValueTypeConsistent()` checks the value coming from the URL against the phpdoc-derived type with `TypeHelper::accepts()`. For an array prop documented with a shape or a generic, e.g.

```php
/** @var array{r: array<string, true>, c: array<string, true>} */
#[LiveProp(url: true)]
public array $expanded = ['r' => [], 'c' => []];
```

this check can never pass: shape keys are marked required while URL values are partial by nature, and query string values are always strings (`'1'` vs `true`). The prop is then silently ignored, while it hydrated fine before the migration (the legacy check only verified the native type, `is_array($value)` for an array prop).

This PR restores that leniency for arrays only: an array value is accepted as soon as the prop type is array-identified, everything else keeps the strict `accepts()` check. Invalid top-level values (`?arrayProp=foo`) are still rejected, as covered by the existing `invalid array value` test case.

Test cases added for the exact scenario reported in #2900 (partial shaped array), for a generic (`int[]`) and for the invalid counterpart (`?shapedArrayProp=foo`).

### DIFF
--- a/src/LiveComponent/src/Util/RequestPropsExtractor.php
+++ b/src/LiveComponent/src/Util/RequestPropsExtractor.php
@@ -98,6 +98,18 @@ final class RequestPropsExtractor
 
         $type = $livePropMetadata->getType();
 
-        return null === $type || TypeHelper::accepts($type, $value);
+        if (null === $type) {
+            return true;
+        }
+
+        // URL values for array props are stringly-typed and possibly partial:
+        // checking them against a phpdoc-derived shape or generic (e.g.
+        // array{...} or array<string, int>) would always fail, so only ensure
+        // the native type matches.
+        if (\is_array($value) && $type->isIdentifiedBy(TypeIdentifier::ARRAY)) {
+            return true;
+        }
+
+        return TypeHelper::accepts($type, $value);
     }
 }
--- a/src/LiveComponent/tests/Fixtures/Component/ComponentWithUrlBoundProps.php
+++ b/src/LiveComponent/tests/Fixtures/Component/ComponentWithUrlBoundProps.php
@@ -36,6 +36,18 @@ class ComponentWithUrlBoundProps
     #[LiveProp(url: new UrlMapping(as: 'arr_alias'))]
     public array $arrayPropAlias = [];
 
+    /**
+     * @var array{r: array<string, true>, c: array<string, true>}
+     */
+    #[LiveProp(url: true)]
+    public array $shapedArrayProp = ['r' => [], 'c' => []];
+
+    /**
+     * @var int[]
+     */
+    #[LiveProp(url: true)]
+    public array $intListProp = [];
+
     #[LiveProp(writable: true, fieldName: 'getArrayFieldName()', url: true)]
     public array $arrayPropFieldName = [];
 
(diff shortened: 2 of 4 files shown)


Kocal:
Branch 2.x only receive security fix

PR #3731, opened 2026-07-25
 Fix broken assets/tsconfig.json reference breaking webpack 5.109+ builds

### LINKED ISSUE #3728: Title: assets/tsconfig.json in UX packages references non-existent ../../../tsconfig.package.json outside the package (breaks build with webpack 5.109+)
### Environment
- Symfony 7.4, PHP 8.2
- Node v24.18.0
- webpack 5.109.0 (regression not present on 5.108.x)
- @symfony/webpack-encore 2.4.1
- symfony/ux-live-component 2.35.0 (same behavior with any symfony/ux-* package)

### Description
Every symfony/ux-* package ships `assets/tsconfig.json` with:

    { "extends": "../../../tsconfig.package.json" }

This relative path only resolves inside the symfony/ux monorepo (packages/*/assets -> repo root tsconfig.package.json). When the package is installed via Composer, the file ends up at e.g. `vendor/symfony/ux-live-component/assets/tsconfig.json`, and `../../../` resolves to `vendor/symfony/tsconfig.package.json`, which does not exist in the published package.

`.gitattributes` in the split repos export-ignores assets/src, assets/test, playwright.config.ts and vitest.config.mjs, but NOT assets/tsconfig.json — so this dangling reference ships to every consumer.

### Trigger
Starting with webpack 5.109.0, `experiments.typescript` defaults to `"auto"` and (per the 5.109 release notes) wires up automatic tsconfig.json resolution in the resolver whenever no explicit .ts/.mts/.cts loader rule is configured. This causes webpack to actually try to resolve the extends chain of assets/tsconfig.json in every installed UX package, which fails with:

    Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
    ModuleNotFoundError: Module not found: Error: ENOENT: no such file or directory, open 'vendor\symfony\tsconfig.package.json'

With webpack 5.108.x this did not happen (tsconfig auto-resolution wasn't wired in by default).

### Steps to reproduce
1. Fresh Symfony 7.4 project with webpack-encore 2.4.1, any symfony/ux-* package (e.g. ux-live-component) installed and enabled in webpack.config.js.
2. Upgrade webpack to 5.109.0.
3. Run `npm run build`.
4. Build fails with the ENOENT above for every installed UX package.

### Expected
`assets/tsconfig.json` should either not be shipped in the published/vendor package (add it to `.gitattributes` export-ignore, same as assets/src, assets/test, playwright.config.ts, vitest.config.mjs), or it should not reference a path outside the package boundary.

### Suggested fix
Add to each UX package's `.gitattributes`:

    /assets/tsconfig.json export-ignore

(or make the file self-contained instead of extending a monorepo-root-only config).

### Where this is documented

The regression trigger is documented in the webpack blog itself:

1. https://webpack.js.org/blog/2026-07-24-webpack-5-109/#built-in-support-goes-auto (Webpack 5.109, section "Built-in Support Goes `auto`"):
   "experiments.typescript enables the built-in type-stripping when Node.js supports it (>= 22.6) and no rule matches .ts / .mts / .cts (so ts-loader and swc-loader win)."
   -> This changed the default from `false` to `"auto"`, so it now enables itself on any Node >= 22.6 setup without an explicit TS loader rule.

2. https://webpack.js.org/blog/2026-05-19-webpack-5-107/#typescript-support-experimental (Webpack 5.107, section "TypeScript Support (Experimental)"):
   "Enabling the flag also wires up sensible defaults: rules for .ts / .cts / .mts, .ts added to extension resolution before .js, extensionAlias so import "./foo.js" also tries ./foo.ts, tsconfig.json resolution, and the "typescript" conditional-exports key for monorepos that ship .ts sources."
   -> This is where automatic tsconfig.json resolution gets wired in whenever the flag is on, which is what now walks into every symfony/ux package's broken assets/tsconfig.json.

The webpack 5.108 release notes (https://webpack.js.org/blog/2026-06-25-webpack-5-108/) contain no TypeScript-related changes, which is why the issue does not reproduce on 5.108.x.

zalesak:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | Fix #3728 
| License        | MIT

### Description

Backport of the `assets/tsconfig.json` export-ignore fix to the `2.x` branch. 

This prevents Webpack 5.109.0+ from throwing `ModuleNotFoundError (ENOENT)` due to the dangling `tsconfig.package.json` reference when UX packages are installed via Composer in older Symfony/UX setups.

### DIFF
--- a/src/Autocomplete/.gitattributes
+++ b/src/Autocomplete/.gitattributes
@@ -3,6 +3,7 @@
 /assets/src export-ignore
 /assets/test export-ignore
 /assets/playwright.config.ts export-ignore
+/assets/tsconfig.json export-ignore
 /assets/vitest.config.mjs export-ignore
 /doc export-ignore
 /phpunit.xml.dist export-ignore
--- a/src/Chartjs/.gitattributes
+++ b/src/Chartjs/.gitattributes
@@ -3,6 +3,7 @@
 /assets/src export-ignore
 /assets/test export-ignore
 /assets/playwright.config.ts export-ignore
+/assets/tsconfig.json export-ignore
 /assets/vitest.config.mjs export-ignore
 /doc export-ignore
 /phpunit.xml.dist export-ignore
--- a/src/Cropperjs/.gitattributes
+++ b/src/Cropperjs/.gitattributes
@@ -3,6 +3,7 @@
 /assets/src export-ignore
 /assets/test export-ignore
 /assets/playwright.config.ts export-ignore
+/assets/tsconfig.json export-ignore
 /assets/vitest.config.mjs export-ignore
 /doc export-ignore
 /phpunit.xml.dist export-ignore
--- a/src/Dropzone/.gitattributes
+++ b/src/Dropzone/.gitattributes
@@ -3,6 +3,7 @@
 /assets/src export-ignore
 /assets/test export-ignore
 /assets/playwright.config.ts export-ignore
+/assets/tsconfig.json export-ignore
 /assets/vitest.config.mjs export-ignore
 /doc export-ignore
 /phpunit.xml.dist export-ignore
--- a/src/LazyImage/.gitattributes
+++ b/src/LazyImage/.gitattributes
@@ -2,6 +2,7 @@
 /.symfony.bundle.yaml export-ignore
 /assets/src export-ignore
 /assets/test export-ignore
+/assets/tsconfig.json export-ignore
 /assets/vitest.config.mjs export-ignore
 /doc export-ignore
 /phpunit.xml.dist export-ignore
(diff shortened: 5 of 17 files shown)


Kocal:
Branch 2.x is not maintained anymore, except for security fixes.

zalesak:
I completely understand that `2.x` is under security-only maintenance. 

However, the reason I am proposing this backport is that this regression directly impacts security updates for consumer applications. Because Webpack 5.109.0+ completely breaks builds due to this dangling reference, projects locked on Symfony UX `2.x` are currently blocked from upgrading Webpack. 

This forces developers to stick with older Webpack versions, preventing them from pulling critical security patches and updates for their build tools in the future. 

Given that this fix is zero-risk (it only modifies `.gitattributes` to ignore a development file during export), would you consider merging it to unblock the Webpack upgrade path for `2.x` users?

PR #3733, opened 2026-07-26
[Toolkit] Document Stimulus controller CSS classes and outlets in the API reference

kbond:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Issues         | -
| License        | MIT

Documents a Stimulus controller's `static classes` and `static outlets` in the recipe API reference from `@css-class` and `@outlet` docblock tags — rendered as Class and Outlet tables and linted both ways like `@value`/`@target`/`@action`. This rounds out the full controller surface: values, targets, classes, outlets and actions.

The class tag is spelled `@css-class` rather than `@class` to avoid clashing with JSDoc's `@class`.

### DIFF
--- a/src/Toolkit/src/Component/StimulusControllerClass.php
+++ b/src/Toolkit/src/Component/StimulusControllerClass.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\UX\Toolkit\Component;
+
+/**
+ * A Stimulus CSS class declared by a controller's `static classes` and documented with a `@css-class` tag.
+ *
+ * The name is read from the code; the description comes from the docblock. The `attribute` is the
+ * `data-*-class` attribute the class is bound to, precomputed from the controller identifier and the name.
+ *
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @internal
+ */
+final class StimulusControllerClass
+{
+    public function __construct(
+        public readonly string $name,
+        public readonly string $attribute,
+        public readonly string $description,
+    ) {
+    }
+}
--- a/src/Toolkit/src/Component/StimulusControllerDoc.php
+++ b/src/Toolkit/src/Component/StimulusControllerDoc.php
@@ -12,7 +12,7 @@
 namespace Symfony\UX\Toolkit\Component;
 
 /**
- * The API reference extracted from a Stimulus controller: its values, targets and actions.
+ * The API reference extracted from a Stimulus controller: its values, targets, classes, outlets and actions.
  *
  * @author Kevin Bond <kevinbond@gmail.com>
  *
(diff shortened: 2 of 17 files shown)


src/Toolkit/tests/Fixtures/kits/render-stimulus-doc/widget/assets/controllers/widget_controller.js
```diff
@@ -4,6 +4,7 @@ import { Controller } from '@hotwired/stimulus';
  * @value  autoClose  Delay in milliseconds before the widget closes.
  * @value  open       Whether the widget is open on initial render.
  * @target panel      The panel shown or hidden by the widget.
+ * @class  open       Applied to the widget while it is open.
```
Kocal: `@class` is a valid JSdoc tag https://jsdoc.app/tags-class

I'm not sure if it's a good idea to use it 😕

smnandre: `@css-class` ? 

I imagine they had the same problem, and this is why they always use css-class in the docs 😅 
https://stimulus.hotwired.dev/reference/css-classes

There is another option, somehow describing better the controller instance reality

```
* @property {string} openClass
```

Kocal: Kevin also suggested me `@css-class`!

Using `@property` annotation for class, target, values etc... could be interesting, especially since the Stimulus plugin for PHPStorm has been broken for months...

kbond: Renamed to `@css-class`

src/Toolkit/src/Component/StimulusControllerClass.php
```diff
@@ -0,0 +1,32 @@
...
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @internal
+ */
+final class StimulusControllerClass
+{
+    public function __construct(
+        public readonly string $name,
+        public readonly string $attribute,
+        public readonly string $description,
```
smnandre, line 29: maybe optional ?

src/Toolkit/src/Component/StimulusControllerDoc.php
```diff
@@ -23,17 +23,19 @@ final class StimulusControllerDoc
...
         public readonly array $targets = [],
+        public readonly array $classes = [],
         public readonly array $actions = [],
     ) {
     }
 
     public function isEmpty(): bool
     {
-        return [] === $this->values && [] === $this->targets && [] === $this->actions;
+        return [] === $this->values && [] === $this->targets && [] === $this->classes && [] === $this->actions;
```
smnandre:
```suggestion
        return [] === $this->actions && [] === $this->values && [] === $this->targets && [] === $this->classes;
```

Nitpicking but according to my inner voice, this should relate more to the probabilities

smnandre (approved):
Genuine question: no outlets for now ?

kbond:
> outlets

~What do you mean? If you mean usages, not yet but I use it in #3740.~ Oh, you meant stimulus outlets 🤦🏻‍♂️. No, I thought I'd wait until needed. But maybe it should be added here...?

src/Toolkit/src/Component/StimulusControllerClass.php
kbond: If we do this, it should be in a follow-up PR that makes the change for all.

Kocal (approved):

kbond:
I just pushed support for `@outlet`.

Kocal:
Thank you @kbond.

PR #3738, opened 2026-07-27
[TwigComponent] Fix props evaluated twice by the component tag

### LINKED ISSUE #3704: Nullsafe attribute in twig component
Hi,

I've just migrated from v2.32.0 to v2.36.0.
Since then, I have some issues with my twig component and more precisely the nullsafe operator :

Before migration : 

```
<twig:Page
    title="Commandes - Affaire : {{ affaire.numero }}"
    breadcrumb="Accueil > <a href='{{ pathIndex }}'>Affaires</a> > {{ lienParent }} {{ lienAffaire }} > Commandes"
    pageTitle="Affaire : {{ affaire.numero ~ ' - ' ~ affaire.commune?.organisation?.raisonSociale ~ ' - ' ~ affaire.libelle }}"
    :menuTabs="menu_tabs"
    :menu="menu"
>
```

In Page.html.twig :
```
<h2 class="page-title">
   {{ this.pageTitle }}
</h2>
```

Was working fine.

But now, if organisation is null, I get : "Impossible to access an attribute ("raisonSociale") on a null variable".

I have to set the variable before :
`{% set pageTitle = "Affaire : " ~ affaire.numero ~ ' - ' ~ affaire.commune?.organisation?.raisonSociale ~ ' - ' ~ affaire.libelle %}
`And then : 
`:pageTitle="pageTitle"`

I've migrated to v3.2.0 to see if there was a little regression that was corrected but no.

Am I missing something ?

Thank you !

Ps : I'm using twig/twig version v3.28.0
and `<twig:Page :pageTitle="affaire.commune?.organisation?.raisonSociale" />` not working neither

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #3704
| License       | MIT

`ComponentNode` compiles the props hash twice: once for `preRender()` and once for `startEmbedComponent()`. Since Twig 3.24, compiling a null-safe chain (`a?.b`) is stateful (twigphp/Twig#4748): the node marks itself as short-circuited and caches its temp variable name. On the second subcompile, the null guard is skipped and the generated code reads a temp variable that is never assigned, so `<twig:Foo :prop="a.b?.c?.d" >...</twig:Foo>` (the tag form, compiled to `{% component %}`) fails with "Impossible to access an attribute on a null variable". The self-closing form (compiled to `{{ component() }}`) only compiles the expression once and is not affected.

This also explains why a bisect on ux would not find anything: `ComponentNode` is identical between v2.32.0 and v2.36.2. The trigger is the Twig upgrade that comes along (`?.` was added in Twig 3.23, the stateful short-circuiting in 3.24).

The fix evaluates the props once into a local variable and passes it to both calls. As a side benefit, props expressions with side effects are no longer evaluated twice at runtime.

The new test errors without the fix and passes with it (it is skipped below Twig 3.23, since the component requires `twig/twig ^3.10.3`).

### DIFF
--- a/src/TwigComponent/src/Twig/ComponentNode.php
+++ b/src/TwigComponent/src/Twig/ComponentNode.php
@@ -59,11 +59,22 @@ final class ComponentNode extends Node implements NodeOutputInterface
         }
 
         $componentRuntime = $compiler->getVarName();
+        $props = $compiler->getVarName();
 
         $compiler
                ->write(\sprintf('$%s = $this->env->getRuntime(', $componentRuntime))
                ->string(ComponentRuntime::class)
                ->raw(");\n");
+
+        // Evaluate the props once: some expressions cannot be compiled twice
+        // (e.g. the null-safe operator, whose compilation is stateful since Twig 3.24)
+        // and evaluating them twice would also duplicate any side effect.
+        $compiler
+            ->write(\sprintf('$%s = ', $props))
+            ->raw($twig_to_array)
+            ->raw('(');
+        $this->writeProps($compiler)
+            ->raw(");\n");
 
         /*
          * Block 1) PreCreateForRender handling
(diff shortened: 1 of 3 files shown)


Kocal:
Branch 2.x only receive security fix

PR #3740, opened 2026-07-29
[Toolkit][Common] Add `clipboard` and `tooltip` recipes

kbond:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

Two design-system-agnostic recipes for the `common` kit.

`clipboard` copies to the clipboard from a `source` value or element (a form control's value, an element's text, or a code block). Copied feedback is opt-in: a `success`/`idle` target swap, or a CSS `success` class the controller toggles for `successDuration`. Every copy also dispatches a `clipboard:copied` event, so you can drive your own feedback.

Screenshot 2026-07-29 at 9 17 21 AM

---

`tooltip` shows a tooltip from a `content` value or an element's native `title` (which it adopts and removes, suppressing the browser's own). It opens on hover and keyboard focus by default (`trigger` also supports `click` and `manual`), and uses [Floating UI](https://floating-ui.com/) to flip/shift into view and escape `overflow` clipping (the tooltip is rendered in `<body>`). Placement and appearance come from a single global `.tooltip` rule. It adds `@floating-ui/dom` as a dependency — the `common` kit's first.

Screenshot 2026-07-29 at 9 18 29 AM

### DIFF
--- a/src/Toolkit/CHANGELOG.md
+++ b/src/Toolkit/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 3.4.0
 
 - [Bootstrap] Add the Bootstrap kit
-- [Common] Add the `common` kit, with design-system agnostic `closeable`, `logout-link` and `post-link` recipes
+- [Common] Add the `common` kit, with design-system agnostic `clipboard`, `closeable`, `logout-link`, `post-link` and `tooltip` recipes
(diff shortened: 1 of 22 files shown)


Kocal (approved):

Kocal:
Thank you @kbond.

PR #3741, opened 2026-07-29
[Icons] Add `auto_lock` to persist on-demand icons

### LINKED ISSUE #3170: Option to skip Iconify icon cache warmup in production builds
Referencing [#2493](https://github.com/symfony/ux/issues/2493)

I can no longer disable Iconify because I need it enabled in production.
This causes `asset-map:compile` to fail during builds due to the icon cache warmup step.

It would be very useful to have an option to skip the warmup, allowing production builds to succeed without internet access.

```yaml
#18 [php_apache_prod 7/7] RUN php bin/console asset-map:compile
#18 0.297 Warming the icon cache...
#18 0.359 
#18 0.360 In ErrorChunk.php line 73:
#18 0.360                                                                                
#18 0.360   Received HTTP code 403 from proxy after CONNECT for "https://api.iconify.de/  
#18 0.360   sign/collections".                                                           
#18 0.360                                                                                
#18 0.360 
#18 0.361 In CurlResponse.php line 336:
#18 0.361                                                                                
#18 0.361   Received HTTP code 403 from proxy after CONNECT for "https://api.iconify.de/  
#18 0.361   sign/collections".                                                           
#18 0.361                                                                                
#18 0.361 
#18 0.361 asset-map:compile
#18 0.361 
#18 ERROR: process "/bin/sh -c php bin/console asset-map:compile" did not complete successfully: exit code: 1
```

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3170 
| License        | MIT

I can't remember how many projects, or how many times it happened, where an issue happened during a deployment because Iconify was unreachable, because some icons were not locked.

I suggest a new configuration `iconify.auto_lock`, `false` by default.

When `iconify.auto_lock` is enabled, each on-demand icon is written to the local icon directory the first time it is rendered, so it can be committed and served from disk in production without hitting the Iconify API.

### DIFF
--- a/src/Icons/CHANGELOG.md
+++ b/src/Icons/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 3.4.0
+
+- Add `iconify.auto_lock` option to persist "on demand" icons to the local icon directory as they are rendered
+
 ## 3.2.0
 
 - Sanitize rendered SVG icons (Iconify bodies and local files) to prevent XSS. See section 2.36.1 below for details
--- a/src/Icons/config/services.php
+++ b/src/Icons/config/services.php
@@ -20,6 +20,7 @@ use Symfony\UX\Icons\IconFactory;
 use Symfony\UX\Icons\Iconify;
 use Symfony\UX\Icons\IconRenderer;
 use Symfony\UX\Icons\IconRendererInterface;
+use Symfony\UX\Icons\Registry\AutoLockIconRegistry;
 use Symfony\UX\Icons\Registry\CacheIconRegistry;
 use Symfony\UX\Icons\Registry\ChainIconRegistry;
 use Symfony\UX\Icons\Registry\IconifyOnDemandRegistry;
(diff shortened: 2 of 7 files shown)


kbond (approved):
I like this!

Thinking we should update the recipe to enable in dev-only. WDYT?

> where an issue happened during a deployment because Iconify was unreachable, because some icons were not locked.

What's the scenario where this happens? Iconify disabled in prod?

Kocal:
> Thinking we should update the recipe to enable in dev-only. WDYT?

Totally! 

> > where an issue happened during a deployment because Iconify was unreachable, because some icons were not locked.
> 
> What's the scenario where this happens? Iconify disabled in prod?

Nope, but sometimes their API is unreachable because their system is down, or smth	 like that, and so it fails. Typically like https://github.com/symfony/ux/issues/3170 or https://github.com/symfony/ux/issues/3170#issuecomment-4222218760

kbond:
Got it! Should we recommend (via recipe) iconify also be enabled only in dev?

smnandre:
I'm not convinced there ... document how to / help chose yes ... but changing recipe also means changing defaults ... and many, many users would have their app not working on deploy/build  :|

kbond:
> but changing recipe also means changing defaults

I guess we can discuss in a recipe PR but I'd be fine with this ([ref](https://github.com/symfony/recipes#updating-recipes)) - especially because there isn't a 3.x+ recipe version out yet.

Kocal:
> users would have their app not working on deploy/build :|

I'm not sure to understand. Why? 
The PR only adds a way to automatically lock icons when rendering the page, disabled by default. 
The recipe will only been altered in dev environment

smnandre:
Hmm sorry I was only reacting to @kbond suggestion :)

> Should we recommend (via recipe) iconify also be enabled only in dev

auto_lock looks like a very interesting feature to me :)

Kocal:
Ah, my bad!

PR #3753, opened 2026-08-05
[Pagination] Add new UX Pagination component

smnandre:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no 
| Documentation? | yes
| Issues         | Fix #... 
| License        | MIT

This PR adds `symfony/ux-pagination`, an experimental bundle that provides
cursor and numbered pagination behind one request-aware `PaginatorInterface`
service. 

PHP owns the query. Twig renders accessible navigation. The browser
follows ordinary links: no JavaScript, no Stimulus, no Turbo required.

```php
$pagination = $paginator
    ->cursor($events->createQueryBuilder('event'))
    ->orderBy(['createdAt', 'id'], 'DESC')
    ->perPage(20)
    ->paginate();
```

```twig
{% for event in pagination %}
    <article>{{ event.name }}</article>
{% endfor %}

{{ ux_pagination(pagination) }}
```

### What is included

- One `PaginatorInterface` entry point: `paginate()`, `query()`, `cursor()`,
  and `fromCallbacks()`, plus named paginator profiles injectable by
  argument name, `#[Target]`, or service id.
- Three strategies: numbered, lookahead (no COUNT query), and bidirectional
  cursor.
- Adapters for arrays, Doctrine ORM 3, Doctrine DBAL 4.4+, and callbacks.
  Custom adapters register through one interface and a service tag.
- Immutable builders for page size, ordering, and URL composition: route,
  path, query-string policy, fragment, and lazy application-provided totals.
- The `ux_pagination()` Twig function and the `<twig:ux:pagination>`
  component, with default, Bootstrap, and Tailwind themes, structural
  blocks, validated attribute hooks, and accessible markup: `nav` landmark,
  `aria-current`, `rel` prev/next, translated labels.
- Translations in ten locales. Message ids are natural English sentences,
  like core validators: without symfony/translation, the identity
  translator renders readable English, `%count%` pluralization included.
- JSON serialization with stable `prev`/`next` links.
- Optional LiveComponent integration through `ComponentWithPaginationTrait`:
  one `createPagination()` method to implement, real-link fallback, URL
  synchronization that follows the configured page parameter (path mapping
  included), and link URLs that stay on the page route during re-renders,
  never on the internal component endpoint.
- `Test\PaginatorFactory` for deterministic application tests, including
  signed-cursor round trips.

### Design decisions reviewers may want to check

- Experimental: the bundle ships under the usual UX experimental notice.
- `count($pagination)` returns the items on the current page, not the
  total. This matches iteration; `getTotalItems()` returns the total. The
  method documents it.
- The cursor signing secret defaults to `%kernel.secret%`, like
  LiveComponent. Offset pagination works without any secret: the codec is a
  lazy proxy that fails only when signing is needed.
- The result interfaces stay small. Serialization shapes (`getMetadata()`,
  `getLinks()`) and URL conveniences live on the concrete classes only.
- Cursor tokens are signed, not encrypted: boundary values are readable.


src/Pagination/assets/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@symfony/ux-pagination",
+    "description": "Pagination controls for Symfony applications",
+    "license": "MIT",
+    "version": "3.2.0",
```
Kocal: 3.4.0

src/Pagination/assets/src/style.css
Kocal, line 1: Sshipping default styles is indeed super useful, but I'm a little worried about what will happen when users want to make more significant changes to styles. What if users use CSS layers, etc..

This is also a subject of BC.

I'm not sure if we should ship default styles like that or not, what about moving it in a dedicated Flex recipe?

src/Pagination/config/services.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+return static function (ContainerConfigurator $container): void {
+    $services = $container->services()
+        ->defaults()
+            ->private()
+    ;
+
+    // -- Adapters ------------------------------------------------------------
+
```
Kocal: Usually we don't do that

src/Pagination/config/services.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\UX\Pagination\Cursor\CursorCodec;
+use Symfony\UX\Pagination\Cursor\CursorCodecInterface;
+use Symfony\UX\Pagination\PaginationInfoFormatter;
+use Symfony\UX\Pagination\Paginator;
+use Symfony\UX\Pagination\PaginatorInterface;
+
+/*
+ * @author Simon André <smn.andre@gmail.com>
+ */
+
```
Kocal:
```suggestion
```

🤓

src/Pagination/doc/index.rst
```diff
@@ -0,0 +1,129 @@
...
+    Does the UI need page numbers or an exact total?
+        |
+        + -> yes -> offset
+        |
+        + -> no -> Can rows be inserted or removed while users browse?
+                       |
+                       + -> yes -> cursor
+                       |
+                       + -> no -> lookahead
+
```
Kocal, line 99: Does symfony.com documentation supports diagram? Like mermaid syntax?

src/Pagination/src/Cursor/CursorOrder.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * @return list<string>|null
+     */
+    public function getFields(): ?array
+    {
+        return $this->fields;
+    }
+
+    public function getDirection(): ?string
```
Kocal, line 91: Let's add PHPDoc for `'ASC'|'DESC'|null`

src/Pagination/src/Cursor/CursorOrder.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        return $this->direction;
+    }
+
+    public function getFingerprint(): string
+    {
+        return hash('sha256', $this->identity);
+    }
```
Kocal, line 99: Also, since UX 3.x requires PHP 8.4, should we use property hook here? private write, public read?

src/Pagination/src/Cursor/CursorOrder.php
```diff
@@ -0,0 +1,96 @@
...
+ * @author Simon André <smn.andre@gmail.com>
+ */
+final readonly class CursorOrder
+{
+    /**
+     * @param non-empty-string  $identity
+     * @param list<string>|null $fields
+     */
+    private function __construct(
+        private string $identity,
```
Kocal: I'm not sure about this property, AFAIK it's only used by `getFingerPrint()`. Do we really use `getFingerPrint()`? If yes, can't we use `return spl_object_id($this);` instead?

src/Pagination/CHANGELOG.md
```diff
@@ -0,0 +1,27 @@
+# CHANGELOG
+
+## Unreleased
```
Kocal:
```suggestion
## 3.5.0
```

src/Pagination/CHANGELOG.md
```diff
@@ -0,0 +1,27 @@
...
+- Add JSON serialization with consistent `prev`/`next` links and public result
+  contracts for common, numbered and cursor pagination state.
+- Add optional LiveComponent integration through `ComponentWithPaginationTrait`, including
+  real-link fallbacks and route-path page synchronization.
+- Add `Test\PaginatorFactory` for deterministic application tests without a
+  PHPUnit dependency.
+- Preserve request filters by default, reject malformed page and cursor input,
+  guard excessive offsets and keep canonical/redirect policy in the
+  application.
+- Keep pagination functional without a Stimulus or Turbo runtime.
```
Kocal:
```suggestion
- Add the component.
```

splitsh.json
```diff
@@ -15,6 +15,7 @@
         "ux-leaflet-map": "src/Map/src/Bridge/Leaflet",
         "ux-native": "src/Native",
         "ux-notify": "src/Notify",
+        "ux-pagination": "src/Pagination",
```
Kocal, line 18: Remember to ping Fabien for creating the repo on GitHub/Packagist

src/Pagination/src/NumberedPaginationInterface.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @template T
+ *
+ * @extends PaginationInterface<T>
+ *
+ * @author Simon André <smn.andre@gmail.com>
+ */
+interface NumberedPaginationInterface extends PaginationInterface
+{
+    public function getCurrentPage(): int;
```
Kocal, line 30: `0|positive-int` ?

src/Pagination/src/NumberedPaginationInterface.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function getPageParameterName(): string;
+
+    /**
+     * Return the exact total, or null for lookahead pagination.
+     */
+    public function getTotalItems(): ?int;
```
Kocal, line 39: `null|0|positive-int` ?

Kocal (commented):
Except minor comments and types-hardening suggestions, it looks fine to me. Symfony will finally have _built-in_ Pagination :D

src/Pagination/doc/index.rst
smnandre: You really should watch the repository I suggested you to watch 😆

seb-jean:
In the three templates under `src/Pagination/templates/theme/*.html.twig`, some variables are written in camelCase (`linkClass`, `userClass`, `rootAttributes`, `userNavigationClass` and `navigationAttributes`) These should use snake_case instead (`link_class`, `user_class`, `root_attributes`, `user_navigation_class`, `navigation_attributes`), to stay consistent with the usual Twig naming convention : https://twig.symfony.com/doc/3.x/coding_standards.html.

Example:

```twig
{% set linkClass = 'px-3 py-2 text-sm font-medium rounded-lg text-gray-700 dark:text-gray-300 bg-white dark:bg-gray-800 border border-gray-300 dark:border-gray-600 hover:bg-gray-50 dark:hover:bg-gray-700 focus:ring-2 focus:ring-blue-500 transition-colors duration-150' %}
{% set userClass = attributes.class|default('') %}
{% set rootAttributes = {
    'aria-label': 'Pagination'|trans({}, 'UXPaginationBundle'),
}|merge(attributes)|merge({
    class: ('ux-pagination-tailwind ' ~ userClass)|trim,
}) %}
{% set userNavigationClass = navigationAttributes.class|default('') %}
{% set navigationAttributes = navigationAttributes|merge({
    class: ('flex flex-wrap items-center justify-center gap-1 sm:gap-2 ' ~ userNavigationClass)|trim,
}) %}
```

src/Pagination/src/Cursor/CursorOrder.php
smnandre: The fingerprint goes into the signed cursor payload and is re-checked on decode. So it _has_ to be deterministic and stable across requests, for the whole journey of a user walking through a cursor nav.

You are right about the naming though: the property holds the canonical form of the order, not the `byIdentity()` argument. 

--> renamed to `$fingerprintInput`.

ker0x:
Doctrine 3.7 will be release with a new [CursorPaginator](https://github.com/doctrine/orm/blob/3.7.x/src/Tools/Pagination/CursorPaginator.php)

seb-jean:
> Doctrine 3.7 will be release with a new [CursorPaginator](https://github.com/doctrine/orm/blob/3.7.x/src/Tools/Pagination/CursorPaginator.php)

Yes, but there's currently a [blocking issue](https://github.com/doctrine/orm/issues/12480) for the release of cursor-based pagination in Doctrine ORM 3.7.

smnandre:
> Doctrine 3.7 will be release with a new [CursorPaginator](https://github.com/doctrine/orm/blob/3.7.x/src/Tools/Pagination/CursorPaginator.php)

Great, and as I see it some of this PR API inspired some changes on Doctrine PR.. and this is why I love Open Source ! 🤝

src/Pagination/assets/src/style.css
smnandre: Having one is like having a default theme, it allows to work out of the box, but this is clearly documented to be something user would need/want to override or probably even rewrite  themselves... 

Here it's really about "plug and play and basic working behaviour". 

Problem with a recipe is also it cannot differenciate "first install" and  update, so it would bring the same problems (if not worst, as we cannot control we does or does not its recipe:updates out there)...

Maybe though this is something we can solve later on, providing URL / commands to get theme from ux toolkit or something ?

src/Pagination/src/Cursor/CursorOrder.php
smnandre: Do we already use them ? I'm not sure I would drop one here and not in other places ... but I'd love to have some "global" decision on the whole UX level to see if and how we use them .. 

Do you have some feedback / info / opinion here ? @kbond same ?

Kocal (approved):
LGTM, waiting for your PR on ux.symfony.com :D

smnandre:
Thank you very much for the review @Kocal !

src/Pagination/assets/src/style.css
smnandre: I think you're right in the end .. still thinking about it.

src/Pagination/src/Adapter/DoctrineOrmAdapter.php
```diff
@@ -0,0 +1,553 @@
...
+ *
+ * @author Simon André <smn.andre@gmail.com>
+ */
+final class DoctrineOrmAdapter implements OffsetAdapterInterface, LookaheadAdapterInterface, CursorAdapterInterface
+{
+    use CursorValuesTrait;
+
+    public function supports(mixed $source): bool
+    {
+        return $source instanceof QueryBuilder;
```
stof, line 36: The ORM pagination tools also support Query object (which are the only way to get support for changing the fetch mode of relations)

src/Pagination/src/Adapter/DoctrineOrmAdapter.php
```diff
@@ -0,0 +1,553 @@
...
+        if (!$source instanceof QueryBuilder) {
+            throw new InvalidArgumentException('Source must be a Doctrine ORM QueryBuilder.');
+        }
+
+        $this->assertCountableQuery($source);
+
+        $qb = clone $source;
+        $alias = $this->getMainAlias($qb);
+
+        // Reset pagination limits and ordering (ORDER BY is unnecessary for COUNT)
```
stof, line 76: Why re-implementing counting yourselves (only for query builders) instead of relying on the counting provided by the Doctrine ORM paginator ?

src/Pagination/src/Adapter/DoctrineOrmAdapter.php
```diff
@@ -0,0 +1,553 @@
...
+        $cursorFields = $order->getFields();
+        $direction = $order->getDirection();
+        if (null === $cursorFields || null === $direction) {
+            throw new InvalidArgumentException('Doctrine ORM cursor pagination requires a field-based cursor order.');
+        }
+
+        $qb = clone $source;
+        $alias = $this->getMainAlias($qb);
+
+        $fields = array_map(static fn ($field) => $alias.'.'.$field, $cursorFields);
```
stof, line 195: the ORM has a CursorPaginator since version 3.7.0, which implements all the logic for cursor-based pagination.

Kocal: Yeah it's planned https://github.com/symfony/ux/issues/3873

Issue #27, opened 2020-12-07
Incomplete documentation?


AxelBriche:
Hello, I have installed Symfony UX Dropzone with this doc:
<img width="914" alt="Capture d’écran 2020-12-07 à 16 24 12" src="https://user-images.githubusercontent.com/1295251/101369466-a642f800-38a8-11eb-97d6-7b143626cde4.png">

The DropzoneType field is displayed inside the view, but dropzone is not loaded and I have 0 error in my chrome console, I only see that:
<img width="390" alt="Capture d’écran 2020-12-07 à 16 26 40" src="https://user-images.githubusercontent.com/1295251/101369788-ff129080-38a8-11eb-876a-901d8ef60b4c.png">

My controllers.json is ok, the cache is clear, `yarn install --force` and `yarn encore dev` is run.
What I have forget please?

PS: I am on  Symfony 5.1.9.

tgalopin:
You probably need to sync your Webpack Encore recipe to use the bridge and install Stimulus in your project: see the blog post on Chart.js for an example https://symfony.com/blog/new-in-symfony-the-ux-initiative-a-new-javascript-ecosystem-for-symfony#example-displaying-charts-in-php-using-symfony-ux-chart-js

celtic34fr:
Hello,

I recognize that for ux-chartjs and ux-cropperjs, we find directly or not examples of use, for ux-dropzone, it is limited to the implementation on the server side and its addition in the rendering of the form.
No server side code to retrieve the loaded image (s). Would it be possible to have a basic example to test easily like ux-chartjs? Could we have a correspondence between the directy setting of dropzonejs and the one we can do with ux-dropzone?
Thank you in advance for everything that can be done to improve the possibility of using this library.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

PR #3756, opened 2026-08-08
[TwigComponent] Support `MergeableInterface` in `ComponentAttributes`

### LINKED ISSUE #3754: [TwigComponent] Make `ComponentAttributes` trully support `MergeableInterface` from Twig
Following conversation https://github.com/symfony/ux/pull/3653#issuecomment-4883598775 (cc @seb-jean)

When using `ComponentAttributes`, merging classes can be delicate when your component accepts to override/merge its classes.
For example, in UX Toolkit, in our Tailwind-based kits, we have this pattern:
```twig
<span
    {# ... #}
    class="{{ ('a bunch of tailwind classes' ~ attributes.render('class')) | tailwind_merge }}"
    {{ attributes }}
>
    {# ... #}
</span>
```

It works, but it could be easier. 

Instead, I would like to drop this pattern by implementing it natively accross UX TwigComponent and https://github.com/tales-from-a-dev/tailwind-merge-php (cc @ker0x) 

Writting this would be equivalent, but much easier:
```twig
<span
    {# ... #}
    {{ attributes.defaults({
        class: 'a bunch of tailwind classes' | tailwind_classes
    }) }}
>
    {# ... #}
</span>
```

IMHO, things would behave like that:
- The new filter `tailwind_classes` (or `tailwind_class`) would live in https://github.com/tales-from-a-dev/tailwind-merge-php (or the bundle), it would implement a [`MergeableInterface` from Twig HTML extra](https://github.com/twigphp/Twig/blob/3.x/extra/html-extra/HtmlAttr/MergeableInterface.php)
- `ComponentAttributes` from UX TwigComponent, maybe, would need to be updated to fully be compatible with `MergeableInterface`

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3754 
| License        | MIT

The main interest is customizing how CSS classes are merged, especially for Tailwind. More broadly, any value passed to `attributes.defaults()` that implements twig/html-extra's `MergeableInterface` is now merged through its own merge protocol, for every attribute key, not just the special `class` / `data-controller` / `data-action` ones. Plain (non-mergeable) values keep the existing behavior (space-concatenation for those three keys, override otherwise), so this is fully backward compatible.

This is a big win for our UX Toolkit kits based on Tailwind, meaning that we can now write:
```twig
<span
    {# ... #}
    {{ attributes.defaults({
        class: 'a bunch of tailwind classes' | tailwind_classes
    }) }}
>
    {# ... #}
</span>
```

instead of:

```twig
<span
    {# ... #}
    class="{{ ('a bunch of tailwind classes' ~ attributes.render('class')) | tailwind_merge }}"
    {{ attributes }}
>
    {# ... #}
</span>
```

## Notes

The `tailwind_classes` filter used in the example comes from [tales-from-a-dev/twig-tailwind-extra](https://github.com/tales-from-a-dev/twig-tailwind-extra), added in [PR #27](https://github.com/tales-from-a-dev/twig-tailwind-extra/pull/27). Until that lands in a tagged release, this PR wires it via a `repositories` VCS entry and a `dev-feature/tailwind-classes-filter` constraint in `composer.json` (require-dev only); needs to be swapped for the real version before merge.

`twig/html-extra` is a soft dependency: the `MergeableInterface` routing is guarded with `interface_exists` checks and graceful `instanceof` guards, so nothing breaks when the package isn't installed.

### DIFF
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 3.5.0
+
+- Add support for `MergeableInterface` from `twig/html-extra` in `ComponentAttributes#defaults()`
+
 ## 3.4.0
 
 - Add support for dynamic component names in the `{% component %}` tag using a
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -14,6 +14,7 @@ namespace Symfony\UX\TwigComponent;
 use Symfony\UX\StimulusBundle\Dto\StimulusAttributes;
 use Symfony\UX\TwigComponent\Exception\RuntimeException;
 use Twig\Extra\Html\HtmlAttr\AttributeValueInterface;
+use Twig\Extra\Html\HtmlAttr\MergeableInterface;
 use Twig\Runtime\EscaperRuntime;
 
 /**
@@ -116,6 +117,10 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
             return null;
         }
 
+        if ($value instanceof AttributeValueInterface && null === $value = $value->getValue()) {
+            return null;
+        }
+
         if ($value instanceof \Stringable) {
             $value = (string) $value;
         }
@@ -145,8 +150,11 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
      * Set default attributes. These are used if they are not already
      * defined.
      *
-     * "class" and "data-controller" are special, these defaults are prepended to
-     * the existing attribute (if available).
+     * "class", "data-controller" and "data-action" are special, these defaults are
+     * prepended to the existing attribute (if available).
+     *
+     * Values implementing Twig HTML extra's MergeableInterface (e.g. from the
+     * "tailwind_classes" filter) are merged through the merge protocol, for any key.
      */
     public function defaults(iterable $attributes): self
     {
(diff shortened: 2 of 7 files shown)


src/TwigComponent/composer.json
```diff
@@ -42,9 +42,16 @@
         "symfony/framework-bundle": "^7.4|^8.0",
         "symfony/stimulus-bundle": "^2.9.1|^3.0",
         "symfony/twig-bundle": "^7.4|^8.0",
+        "tales-from-a-dev/twig-tailwind-extra": "dev-feature/tailwind-classes-filter",
```
Kocal: To change when https://github.com/tales-from-a-dev/twig-tailwind-extra/pull/27 is merged

src/TwigComponent/composer.json
```diff
@@ -42,9 +42,16 @@
...
         "twig/extra-bundle": "^3.10.3",
         "twig/html-extra": "^3.10.3"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/Kocal/twig-tailwind-extra"
+        }
+    ],
```
Kocal: To remove when https://github.com/tales-from-a-dev/twig-tailwind-extra/pull/27 is merged

src/TwigComponent/composer.json
kbond: What about not depending on this and just stubbing the behaviour?

Kocal: It's a dev dependency, but we can make it a stub yes (like we did for the unit test)

src/TwigComponent/composer.json
Kocal: Done

kbond (approved):
I don't feel strongly about removing the dev dep. Just thought it might make the pr a bit simpler.

src/TwigComponent/composer.json
Kocal: Done

PR #3757, opened 2026-08-09
[LiveComponent] Document how to validate components on their initial render

### LINKED ISSUE #2974: [LiveComponent] Validation doesn't seem to work
Hi team!

I'm trying to validate a property of some LiveComponent that doesn't use a Symfony Form. I'm following this: https://symfony.com/bundles/ux-live-component/current/index.html#validation-without-a-form

My code is like this:

```php
#[AsLiveComponent]
class MyComponent
{
    use DefaultActionTrait;

    #[LiveProp(writable: true)]
    #[Assert\Choice(choices: ['foo', 'bar'])]
    public string $type;

    // ...
}
```

If I put this in my template, I expect to see an error, but I don't see anything. The LiveComponent renders normally, without validating the value of the `$type` property:

```twig
<twig:MyComponent type="aaaaa"/>
```

-----

Also, on the Symfony Docs, I don't understand the followin:

```
Be sure to add the IsValid attribute/annotation to any property
where you want the object on that property to also be validated.
```

The docs doesn't show any example using that `IsValid` attribute and I'm not sure where to find it. Thanks!

javiereguiluz:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | Fix #2974
| License        | MIT

When using `ValidatableComponentTrait` without a Symfony Form, validation is triggered by a `PostHydrate` hook, so it only runs when the component state comes back from the frontend (e.g. after updating a writable `LiveProp` or calling a `LiveAction`).

The initial render is never validated, which is expected behavior but the docs didn't mention it (that's what caused the  confusion in #2974).

### DIFF
--- a/src/LiveComponent/doc/index.rst
+++ b/src/LiveComponent/doc/index.rst
@@ -2267,8 +2267,8 @@ need::
         #[Assert\Valid]
         public User $user;
 
-         #[LiveProp]
-         #[Assert\IsTrue]
+        #[LiveProp]
+        #[Assert\IsTrue]
         public bool $agreeToTerms = false;
     }
 
@@ -2276,10 +2276,17 @@ Be sure to add the ``Valid`` attribute/annotation to any property
 where you want the object on that property to also be validated.
 
 Thanks to this setup, the component will now be automatically validated
-on each render, but in a smart way: a property will only be validated
+on each re-render, but in a smart way: a property will only be validated
 once its "model" has been updated on the frontend. The system keeps
 track of which models have been updated and only stores the errors for
 those fields on re-render.
+
+.. note::
+
+    Automatic validation *only* happens *after* the component state is sent back
+    from the frontend (e.g. when a writable ``LiveProp`` is updated or a
+    ``LiveAction`` is called). If you need to validate the *initial* render of a
+    component, see :ref:`validating-on-initial-render`.
 
 You can also trigger validation of your *entire* object manually in an
 action::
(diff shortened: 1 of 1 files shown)


src/LiveComponent/doc/index.rst
```diff
@@ -2267,20 +2267,28 @@ need::
...
+        #[Assert\IsTrue]
         public bool $agreeToTerms = false;
     }
 
 Be sure to add the ``Valid`` attribute/annotation to any property
 where you want the object on that property to also be validated.
 
 Thanks to this setup, the component will now be automatically validated
-on each render, but in a smart way: a property will only be validated
+on each re-render, but in a smart way: a property will only be validated
```
smnandre, line 2279: 👍

src/LiveComponent/doc/index.rst
```diff
@@ -2267,20 +2267,28 @@ need::
...
 those fields on re-render.
 
+.. note::
+
+    Automatic validation happens *after* the component state is sent back
+    from the frontend (e.g. when a writable ``LiveProp`` is updated or a
+    ``LiveAction`` is called). The *initial* render of the component is
+    **not** validated. See :ref:`validating-on-initial-render` if you need
+    that.
+
```
smnandre:
```suggestion
    Automatic validation *only* happens *after* the component state is sent 
    back from the frontend (e.g. when a writable ``LiveProp`` is updated or a ``LiveAction`` is called). 
    If you need to validate the *initial* render of a component: see :ref:`validating-on-initial-render`.

```

wdyt ?

src/LiveComponent/doc/index.rst
```diff
@@ -2333,6 +2341,37 @@ Once a component has been validated, the component will "remember" that
 it has been validated. This means that, if you edit a field and the
 component re-renders, it will be validated again.
 
+.. _validating-on-initial-render:
+
+Validating on Initial Render
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Automatic validation is triggered when the component state is hydrated
```
smnandre: i'd add some introduction like: 

"Passing data to a Twig component generally means that it has already been validated by an upper layer of your application, most often in your controller or form handling. A component's primary responsibility is therefore to render a valid state, not to validate application data again.

Components can still enforce their own invariants when needed, especially when they are reusable across different contexts or when some combinations of properties would result in an invalid component state."

wdyt ?

smnandre (approved):
Nice addition, thank you @javiereguiluz!

src/LiveComponent/doc/index.rst
javiereguiluz: Yes, I think this was missing. Added!

src/LiveComponent/doc/index.rst
javiereguiluz: I like it, so I used your proposed content.

javiereguiluz:
Thanks for the review! I agree on the proposed changes, so I made all those changes.

Kocal:
Thanks Javier :)

PR #3760, opened 2026-08-09
[Toolkit] Migrate Tailwind kits to the `tailwind_classes` mergeable idiom

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Part of #3233, builds on #3756
| License        | MIT

Migrates 171 Toolkit templates (Shadcn and Flowbite kits) from the `tailwind_merge` class-concatenation idiom to the `tailwind_classes` filter (tales-from-a-dev/twig-tailwind-extra 1.3) combined with the native `MergeableInterface` support added in #3756. Follows up #3754.

### Before / after

```twig
{# Before #}
class="{{ ('px-4 py-2 rounded ' ~ attributes.render('class'))|tailwind_merge }}"

{# After #}
{{ attributes.defaults({ class: 'px-4 py-2 rounded'|tailwind_classes }) }}
```

With `html_cva` it is the same: `class: style.apply({variant, size})|tailwind_classes`.

### Fixes it brings

These are direct consequences of the mergeable protocol, not separable from the migration.

- Consumer-provided classes now correctly override a component's `html_cva` variant classes, matching shadcn's `cn(variants, className)` semantics. The old `apply(ctx, extra)|tailwind_merge` let variant classes win. This surfaces in one snapshot: an input-group button renders `text-sm` instead of `text-xs`.
- When a component forwards `class` to a mergeable child (Field/Label, InputGroup, Pagination), the old idiom silently dropped the forwarding component's own base classes. They now compose correctly as base < wrapper < caller (the same layering React/Vue use), so state styles finally apply — e.g. a disabled RadioGroup field now dims its label. This restores parity with shadcn upstream. **Before and after:**
Capture d’écran 2026-08-10 à 08 13 38

### Other changes

- Recipe manifests bumped to require `twig-tailwind-extra ^1.3.0`, `twig/html-extra ^3.24.0`, `symfony/ux-twig-component ^3.5`.
- The kit scaffolder (`CreateKitCommand`) now emits the new pattern.
- `ClassMergeSpacingChecker` linter removed (the concatenation idiom it guarded is gone); `AttributesDefaultsChecker` and `ComposerSymbolChecker` now understand `tailwind_classes`.
- Snapshots regenerated; `bin/ux-toolkit-kit-lint --fail-on-warning`, php-cs-fixer, and twig-cs-fixer all pass.

### Limitations

7 templates deliberately keep `tailwind_merge`. The mergeable `defaults()` form only works when the class merge and the attributes bag land on the same real HTML element.

**Base class and attributes bag live on different elements.** The Switch is the clearest example:

```twig
<div data-slot="switch" class="{{ ('group relative inline-flex ...' ~ attributes.render('class'))|tailwind_merge }}">
    <span data-slot="switch-thumb" class="..."></span>
    <input type="checkbox" {{ attributes.without('class') }} />
</div>
```

The base classes and any consumer-supplied `class` merge onto the root `<div>` (the visual track), while the consumer's other attributes go on the inner `<input>` (the form control). No single element carries both, so `defaults({class})` can't express it. Same shape: RadioGroup/Item, Accordion/Content, Modal/Content.

**Class spread onto a non-mergeable external component.** `<twig:ux:icon>` (used by the Shadcn Spinner) validates `class` as a scalar string and rejects the `TailwindClassList` object that `tailwind_classes` returns, so it keeps a pre-merged string. Note that spreading onto a mergeable Toolkit child (`<twig:Button>`, `<twig:Label>`, etc.) is not a limitation — those components were migrated and compose correctly (see Fixes above). The pre-existing sonner and combobox exceptions are unchanged.


smnandre (approved):
Approving here, but my voice has not much value here as I'm not user of it.

Naive question: could/should this go into Twig directly, now there are filters/functions to handle utilities, and cva is here too? (genuine question)

Kocal:
> Naive question: could/should this go into Twig directly, now there are filters/functions to handle utilities, and cva is here too? (genuine question)

I'm not sure to understand what you're speaking about, `ComponentAttributes` maybe? Or more precisely about `ComponentAttributes::__toString()` that ideally, should directly use [`HtmlExtension::htmlAttr()`](https://github.com/twigphp/html-extra/blob/3.x/HtmlExtension.php#L195)?

Kocal:
I just pushed a new commit that fixes the usage of the pattern on a `<twig:Component>` tag:
Capture d’écran 2026-08-10 à 08 15 57
Capture d’écran 2026-08-10 à 08 16 34

`{{ ...attributes }}` erased the default `class` attribute when being used on a `<twig:Component>`, and these default classes were not applied. 

Since this PR contains both a feature and a fix, I also added a CHANGELOG entry for the fix (so it's not lost)

PR #3762, opened 2026-08-10
[TwigComponent] Fix twig:blockquote is not a block

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

A component whose name starts with `block` breaks the block it sits in.

## Reproducer

```twig
<twig:Card>
    <twig:block name="footer">
        <twig:blockquote cite="Matt">He ho!</twig:blockquote>
    </twig:block>
</twig:Card>
```

```
Twig\Error\SyntaxError: Expected closing tag "</twig:Card>" not found at line 5.
```

The error points at `Card`, which is correctly closed, and at the last line of the
template. Nothing mentions `blockquote`, so there is little to go on.

Same failure with `<twig:blockquote />`, `<twig:block-title />`, `<twig:blocked />`, and
inside a traditional `{% block %} … {% endblock %}`.

Not affected: `<twig:Blockquote />` (the comparison is case-sensitive), and any of these
outside a block.

## Cause

`TwigPreLexer::consumeUntilEndBlock()` tracks block nesting with a fixed-length prefix
comparison:

```php
if (!$inComment && '<twig:block' === substr($this->input, $this->position, 11)) {
    ++$depth;
}
```

`'<twig:block'` is also the first 11 characters of `<twig:blockquote`, so the depth is
incremented for a component that is not a block. It is never decremented back: the
closing comparison is `'</twig:block>'`, which ends with `>` and therefore does not
match `</twig:blockquote>`. The enclosing `<twig:block>` never terminates, scanning runs
to the end of the template, and the outer component is reported as unclosed.

## Fix

Anchor the match and require that `block` is not followed by another component-name
character, using the same character class as `consumeComponentName()`:

```php
if (!$inComment && preg_match('/\G<twig:block(?![A-Za-z0-9_:@\-.])/', $this->input, $matches, 0, $this->position)) {
    ++$depth;
}
```

The closing comparison needs no change.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -455,6 +455,11 @@ class TwigPreLexer
             && 0 === substr_compare($this->input, $chars, $this->position, \strlen($chars));
     }
 
+    private function isNameChar(string $char): bool
+    {
+        return '' !== $char && (ctype_alnum($char) || str_contains('_:@-.', $char));
+    }
+
     private function consumeBlock(string $componentName): string
     {
         $attributes = $this->consumeAttributes($componentName);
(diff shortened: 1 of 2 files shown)


kbond (approved):

Kocal (approved):

Kocal:
I just rebased and fixed conflicts (due to https://github.com/symfony/ux/pull/3776), I changed the `preg_match` call for a more performant solution.

Kocal:
Thank you @smnandre.

PR #3763, opened 2026-08-10
[Autocomplete] Translate the optgroup labels returned by the AJAX endpoint

kira0269:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Issues        | -
| License       | MIT

The `group_by` option of an autocomplete field produces the `<optgroup>` labels. When the `<select>` is rendered by the form theme, those labels go through `|trans`:

```twig
<optgroup label="{{ choice_translation_domain is same as(false) ? group_label : group_label|trans({}, choice_translation_domain) }}">
```

But the labels sent as JSON by the autocomplete endpoint are emitted verbatim by `AutocompleteResultsExecutor`:

```php
$optgroups = array_map(static fn (string $label) => ['value' => $label, 'label' => $label], array_unique($optgroupLabels));
```

For an AJAX autocompleter this is the path that matters: the server-rendered `<select>` only holds the preselected option(s), so every group the user sees while typing comes from that JSON. Using a translation key in `group_by` therefore shows the raw key. It also makes the labels disagree with the server-rendered ones, which the Stimulus controller looks up by label:

```ts
const optgroupElement = parentElement.querySelector(`optgroup[label="${optgroupData.label}"]`);
```

### What this PR does

* `AutocompleteResultsExecutor` takes an optional `TranslatorInterface` and translates the optgroup labels. Values returned by `group_by` that implement `TranslatableInterface` are no longer cast to string either (casting a `TranslatableMessage` is a fatal error).
* The domain comes from a new optional `getTranslationDomain(): string|false|null` method on the autocompleter — `null` uses the default domain, `false` disables the translation. It is declared as a `@method` on `EntityAutocompleterInterface` and commented out in the interface body, like `getGroupBy()`, and detected with `method_exists()`, so existing autocompleters keep working untouched.
* `WrappedEntityTypeAutocompleter` implements it by reading the field's `choice_translation_domain` option, which is exactly what the form theme uses. So the AJAX results and the rendered `<select>` now agree.
* The optgroup `value` keeps the untranslated label, so it still ties a result to its group whatever the locale is. Only the `label` is translated.

### BC

The translator is an optional constructor argument wired with `NULL_ON_INVALID_REFERENCE`, and without it the previous behavior is kept. Autocompleters that do not implement `getTranslationDomain()` translate in the default domain, which matches what the form theme already does for the same labels.

### DIFF
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## Unreleased
+
+- Translate the `optgroup` labels returned by the AJAX endpoint, so that the `group_by`
+  option can use translation keys, like the ones rendered by the form theme already do.
+  The `getTranslationDomain()` method can be implemented by custom autocompleters to
+  choose the domain; entity autocomplete fields use their `choice_translation_domain` option.
+
 ## 2.36.2
 
 - Fix the autocomplete search query throwing an exception on PostgreSQL because of
--- a/src/Autocomplete/CHANGELOG.md
+++ b/src/Autocomplete/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 3.5
+
+- Translate the `optgroup` labels returned by the AJAX endpoint, so `group_by` can use translation keys
+- Add `EntityAutocompleterInterface::getTranslationDomain()` to choose the translation domain used for the `optgroup` labels
+
 ## 3.2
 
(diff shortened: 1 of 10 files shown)


Kocal:
Hi, please target the branch 3.x, branch 2.x is only maintained for security fixes

Kocal:
I switched the base branch to 3.x since 2.x only receives security fixes now. Could you rebase your branch on 3.x so the diff and CI stay clean? That should also get rid of the PHP 8.0 CI failures, since 3.x requires PHP 8.4.

Kocal (approved):

Kocal:
Thank you @kira0269.

PR #3764, opened 2026-08-10
[TwigComponent] Fix {% verbatim %} pre-lexing

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

`{% verbatim %}` is supposed to stop the pre-lexer from rewriting what it wraps. It does,
in one exact spelling, and only when it is not the first thing inside a component.

## Reproducer 1: any other spelling of the tag

```twig
{%- verbatim -%}<twig:Alert/>{%- endverbatim -%}
```

The content is rewritten to `{{ component('Alert') }}` instead of being left alone. Same
with `{%~ verbatim ~%}`, `{%verbatim%}`, `{%   verbatim   %}`, a newline inside the tag,
and mixed delimiters such as `{%- verbatim %} … {% endverbatim -%}`.

Twig accepts all of these. The pre-lexer recognises only the literal string
`{% verbatim %}`, one space on each side.

## Reproducer 2: verbatim as the whole content of a component

```twig
<twig:BasicComponent>{% verbatim %}hello{% endverbatim %}</twig:BasicComponent>
```

```
Twig\Error\SyntaxError: A template that extends another one cannot include content
outside Twig blocks. Did you forget to put the content inside a {% block %} tag?
```

Adding any character before the tag makes it work again, which is the tell:

```twig
{# renders "ahello" #}
<twig:BasicComponent>a{% verbatim %}hello{% endverbatim %}</twig:BasicComponent>
```

## Cause

Both come from the same branch at the top of the scanning loop.

```php
if ($this->consume('{% verbatim %}')) {
```

`consume()` compares a fixed string, hence reproducer 1.

That branch also runs before the rule that opens the implicit default block when
non-whitespace content shows up inside a component. So verbatim content is emitted
outside any `{% block %}`, and Twig rejects it. A single character before the tag trips
the default-block rule first, which is why it then works.

## Fix

Match the tag with a regex covering the whitespace control tokens Twig declares
(`whitespace_trim` is `-`, `whitespace_line_trim` is `~`):

```php
private const VERBATIM_REGEX = '/\G\{%[-~]?\s*verbatim\s*[-~]?%\}/';
private const ENDVERBATIM_REGEX = '/\{%[-~]?\s*endverbatim\s*[-~]?%\}/';
```

`consumeVerbatim()` finds the closing tag with the same pattern and copies the block
through unchanged, counting the newlines of the tags themselves, which the previous code
did not do.

The branch now opens the default block first, like every other kind of content.

Delimiter consistency is not enforced: `{%- verbatim %} … {% endverbatim -%}` is accepted,
as Twig accepts it.

## Tests

Eight cases added to the existing `TwigPreLexerTest` provider. Seven fail on `3.x`:
the five tag spellings above, mixed delimiters, and verbatim as the whole content of a
component. The eighth is a control that already passes and must keep passing, verbatim
placed after some text.

Checked in a real render too, not only through the pre-lexer: reproducer 2 now outputs
`hello`, and `{%- verbatim -%}<twig:Nope/>{%- endverbatim -%}` keeps `<twig:Nope/>` in the
output.

### DIFF
--- a/src/TwigComponent/src/Twig/TwigPreLexer.php
+++ b/src/TwigComponent/src/Twig/TwigPreLexer.php
@@ -19,6 +19,9 @@ use Twig\Lexer;
  */
 class TwigPreLexer
 {
+    private const VERBATIM_REGEX = '/\G\{%[-~]?\s*verbatim\s*[-~]?%\}/';
+    private const ENDVERBATIM_REGEX = '/\{%[-~]?\s*endverbatim\s*[-~]?%\}/';
+
     private string $input;
     private int $length;
     private int $position = 0;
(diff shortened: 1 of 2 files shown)


Kocal (approved):

Kocal:
Thank you @smnandre.

PR #3765, opened 2026-08-11
[Image] Add Image component

seb-jean:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix https://github.com/symfony/ux/pull/3188 / https://github.com/symfony/ux/pull/3549 
| License        | MIT

This PR adds a new `symfony/ux-image` package providing Twig components for rendering ``, `<picture>` and `<source>` HTML elements.

**Components**

- `<twig:ux:img>` :  renders an `` element (runtime renderer, void element)
- `<twig:ux:picture>` : renders a `<picture>` wrapper (template-based, accepts children)
- `<twig:ux:source>` : renders a `<source>` element (runtime renderer, void element)

**Twig functions**

- `ux_img()` : programmatic alternative to the component
- `ux_picture()` : programmatic alternative with sources array

**Supported attributes**

All standard HTML attributes: `src`, `alt`, `srcset`, `sizes`, `width`, `height`, `loading`, `decoding`, `fetchpriority`, `crossorigin`, `referrerpolicy`, `ismap`, `usemap`. Extra attributes (`class`, `id`, `data-*`,  …) are passed through.

`srcset` and `sizes` accept both strings and PHP arrays for readability.

**Validation**

The components validate attributes against the HTML specification at render time:

- `alt` is always required (empty string allowed for decorative images)
- `src` or `srcset` must be provided
- `srcset` must not mix width (`w`) and density (`x`) descriptors
- `srcset` must not contain duplicate descriptors
- `sizes` requires all `srcset `candidates to use `w` descriptors
- `sizes="auto"` (or `"auto, <fallback>"`) requires `loading="lazy"`
- Enum attributes (`loading`, `decoding`, `fetchpriority`, `crossorigin`, `referrerpolicy`) are validated
- `<source>` does not accept `src`

**Example**

```twig
{# Basic image #}
<twig:ux:img src="/photo.jpg" alt="A photo" width="800" height="600" loading="lazy" />
```

```twig
{# Responsive with srcset #}
<twig:ux:img
    src="/photo-800.jpg"
    alt="Hero"
    :srcset="['/photo-480.jpg 480w', '/photo-800.jpg 800w', '/photo-1200.jpg 1200w']"
    :sizes="['(max-width: 600px) 480px', '800px']"
/>
```

```twig
{# Format selection: AVIF → WebP → JPEG #}
<twig:ux:picture>
[...]

PR #3768, opened 2026-08-13
[Image] Add the UX Image package

smnandre:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no 
| Documentation? | yes
| Issues         | Fix #... 
| License        | MIT

So I had the intention to open the Upload PR first ... but #3765 kinda changed my plan :) 

I don't think it's a question of "one or the other one", and let's see what feature, DX, etc both PR bring to the table, and what we can build from there.

I'll write a longer description tomorrow, but I didn't want to let too much time pass before opening this.

/!\ Review warning: this is a big one... 

---

## TL;DR;

UX Image turns one uploaded raster image into durable responsive-image metadata, then renders native `<picture>` and `` markup without storage I/O or JavaScript.
Applications keep ownership of uploads and entities while the bundle provides one explicit processing, storage and rendering contract.

## UX Image

Process the image once:

```php
$asset = $processor->process(
    $uploadedFile,
    profile: 'product',
    storage: 'default_public',
);

$product->setImage($asset);
```

Render the persisted value anywhere:

```twig
{{ ux_picture(product.image, {
    alt: product.name,
    lazy: false,
    fetchpriority: 'high',
}) }}
```

The browser receives native responsive markup and selects the appropriate format, width and density. Rendering uses the persisted `ImageAsset` metadata only: it does not reopen the source image or contact storage.

### Objectives

UX Image provides a Symfony-native image pipeline with a deliberately small application boundary:

- process uploaded images through named, reviewable profiles instead of scattering transformation options through application code;
- persist one immutable, versioned `ImageAsset` value that remains independent from entities and storage implementations;
- render correct responsive HTML from metadata alone, with no runtime image processing and no frontend dependency;
- make file validation, processing budgets, storage publication and regeneration explicit enough for production use.

### Pipeline

```mermaid
flowchart LR
    A["UploadedFile"] --> B["Inspect real bytes"]
    B --> C["Named profile"]
    C --> D["Generate variants"]
    D --> E["Local or Flysystem storage"]
    E --> F["Immutable ImageAsset"]
    F --> G["Application persistence"]
    G --> H["ux_picture() or ux_image()"]
    H --> I["Native responsive HTML"]
```

Processing and rendering are separate operations. The expensive work happens when the application accepts the image; ordinary page rendering reads only persisted metadata.

### Features

- **Named profiles:** formats, dimensions, resize mode, quality, focal point, art direction, processing mode and revision are configured centrally;
- **Bounded processing:** binary inspection, EXIF orientation, codec capability checks, input limits, output budgets and no silent format fallback;
- **Portable storage:** local and Flysystem implementations, immutable generation paths, rollback of partial writes and explicit public URL adapters;
- **Persistable assets:** immutable schema-versioned `ImageAsset`, JSON serialization contract and optional Doctrine DBAL type;
- **Responsive rendering:** `ux_picture()`, `ux_image()` and an optional Twig Component produce `srcset`, `sizes`, intrinsic dimensions, media conditions and loading hints;
- **Operations:** configuration validation, deterministic test fixtures and bounded regeneration through application-owned providers and persisters.

### Out of Scope

- file pickers, browser upload transport and form ownership;
- application entities, authorization and database lifecycle;
- a media library, image editor or content-management workflow;
- provider-native transformation APIs or browser-to-cloud ingestion;
- automatic cleanup of image generations that are still referenced by application data.

These boundaries are intentional. UX Upload, Symfony Forms or application code can provide the input; UX Image starts when the application has an authorized `UploadedFile` to process.

### Required Dependencies

- PHP 8.4 or later;
- Symfony 7.4 or 8.x;
- TwigBundle for the rendering integration;
- one processing backend: the PHP GD extension for the default driver, or a configured custom processor.

The package also uses Symfony Config, Console, DependencyInjection, Filesystem, HttpFoundation and HttpKernel, plus PSR-6 cache contracts.

### Optional Dependencies

| Dependency                         | Enables                                                        |
| ---------------------------------- | -------------------------------------------------------------- |
| `league/flysystem ^3.0`            | Remote or application-defined storage backends                 |
| `doctrine/dbal ^4.0`               | The `image_asset` JSON persistence type                         |
| `intervention/image ^3.0`          | Imagick, VIPS or a custom Intervention driver                   |
| `intervention/image-driver-vips`   | VIPS processing with its required system extensions            |
| `symfony/ux-twig-component ^3.0`   | The optional `<twig:ux:image>` component                        |

Optional integrations are registered only when their concrete dependency is available. The core value objects, renderer contracts and local storage do not require them.

## Usage

Define the outputs required by the layout:

```yaml
# config/packages/ux_image.yaml
ux_image:
    profiles:
        product:
            directory: products
            formats: [webp, jpeg]
            sizes: '(min-width: 64rem) 50vw, 100vw'
            variants:
                small:  { width: 480, mode: fit }
                medium: { width: 960, mode: fit }
                large:  { width: 1440, mode: fit, quality: 88 }
```

Process the real uploaded file after application authorization:

```php
use Symfony\UX\Image\Processor\ImageProcessorInterface;

final class ProductImageUpdater
{
    public function __construct(
        private ImageProcessorInterface $processor,
    ) {
    }

    public function update(Product $product, UploadedFile $file): void
    {
        $product->setImage($this->processor->process(
            $file,
            profile: 'product',
            storage: 'default_public',
        ));
    }
}
```

The application persists the returned `ImageAsset` with its owning model. It can use its own mapping or the optional Doctrine DBAL type.

## Configuration

The default configuration works with local storage and GD. Applications add named profiles and storages as their layouts and deployment require:

```yaml
ux_image:
    driver: gd
    storage_root: '%kernel.project_dir%/var/ux-image'
    preferred_formats: [avif, webp, jpeg]

    limits:
        max_input_bytes: 20M
        max_width: 12000
        max_height: 12000
        max_megapixels: 40
        max_variants: 20
        max_output_megapixels: 100

    storages:
        default_public:
            public_url_prefix: /uploads/images
```

Profiles select transformation behavior. Storage is selected at processing time and recorded in the resulting asset, so the same profile can publish to different storage backends.

Processing modes are explicit:

- `immediate` writes the original and variants before returning;
- `deferred` stores an asset that can be completed later;
- `async` delegates dispatch to an application implementation of `ImageProcessingDispatcherInterface`.

The bundle does not invent an application message, owner identifier or persistence transaction for asynchronous work.

## Rendering

`ux_picture()` performs format negotiation through `<source>` elements and keeps a JPEG, PNG or original fallback in ``. `ux_image()` renders a single native `` when format negotiation is unnecessary.

Both functions support:

- intrinsic dimensions and aspect-ratio stability;
- `srcset`, profile-level `sizes` and density descriptors;
- art-directed sources with media conditions;
- lazy loading, decoding and fetch priority;
- safe application attributes without allowing overrides of renderer-owned attributes.

The optional `<twig:ux:image>` component delegates to the same renderer. There is one rendering contract, not a second component-specific implementation.

## Storage and Persistence

`ImageAsset` stores paths, dimensions, MIME information, variants, profile name and profile revision. It stores no resolved public URL. This keeps persisted data portable when a CDN hostname or URL strategy changes.

Storage publication uses immutable generation keys. A failed processing run removes only the new objects it created; an existing published generation is never overwritten. The application makes the database update durable before deleting an older generation.

Rendering never checks storage existence. The persisted asset and its stored objects therefore form an application consistency boundary.

## Security

Images are untrusted binary input. Security is applied before processing, while writing outputs and when generating public markup.

### Input Files

- the binary signature is inspected instead of trusting the filename or browser MIME type;
- non-images are rejected before a storage directory or object is created;
- SVG is rejected by default and can only enter through an explicit application sanitizer or rasterizer policy;
- EXIF orientation is normalized before transformation;
- input bytes, dimensions and decoded megapixels are bounded.

### Processing and Storage

- profile variant count and total output pixels are validated before work begins;
- requested codecs must be supported by the effective driver and never silently fall back to another format;
- storage paths reject absolute paths, backslashes, NUL bytes and traversal segments;
- partial writes are rolled back in reverse order and streams are closed on failure;
- random immutable generation names prevent application filenames from becoming storage paths.

### Authorization and Delivery

The application authorizes the owner and tenant before calling `process()`. It must never accept an `ImageAsset` JSON document or storage path directly from a client.

Public URL prefixes and CDN builders provide addressing, not access control. Private originals require private storage and an application-owned signed URL adapter or controller.

## Regeneration

Regeneration starts from application persistence, never from a storage scan. Applications implement:

- `ImageAssetProviderInterface` to expose bounded, stably ordered batches;
- `ImageAssetPersisterInterface` to publish each replacement with application-level compare-and-swap semantics.

```bash
php bin/console ux:image:regenerate product \
    --storage=default_public \
    --batch-size=100
```

The command supports dry runs, current-revision skipping, forced regeneration and resumable opaque cursors. The provider and persister keep entity knowledge and transactions where they belong: in the application.

## Tests

- **PHP:** 504 tests and 1,425 assertions;
- **Bundle integration:** container configuration, optional dependency isolation, Twig registration and initialization failures;
- **Processing:** GD and Intervention drivers, binary inspection, geometry, focal points, capabilities, limits and rollback behavior;
- **Rendering:** responsive sources, art direction, dimensions, safe attributes and deterministic HTML contracts;
- **Storage:** shared local and Flysystem behavior, path confinement, stream ownership, publication and cleanup;
- **Operations:** validation and regeneration commands, provider contracts, persistence conflicts and failure recovery.

## Documentation

The documentation follows the image lifecycle from profile design to production operation:

| Page | Focus |
| ---- | ----- |
| `overview.md` | Mental model, first decisions and package boundaries |
| `installation.md` | Installation, bundle registration and first configuration |
| `processing.md` | Profiles, formats, geometry, focal points and processing modes |
| `image-asset.md` | Persisted metadata, schema and Doctrine integration |
| `rendering.md` | Twig functions, responsive HTML and performance options |
| `storage.md` | Local storage, Flysystem, URL adapters and publication |
| `regeneration.md` | Bounded providers, persisters, cursors and replacement |
| `integrations.md` | Symfony Forms, UX Upload, Twig Component and application services |
| `security.md` | Untrusted inputs, budgets, authorization and delivery |
| `configuration.md` | Complete configuration reference |
| `testing.md` | Fixtures and application-level testing strategies |
| `debugging.md` | Diagnostics and common deployment failures |
| `architecture.md` | Internal boundaries and extension contracts |

The Markdown pages will be converted to reStructuredText before the documentation is submitted for publication.

## Coming Next

Separate pull requests can provide:

- the official Symfony Flex recipe;
- the Symfony UX website page and live examples;
- focused examples combining UX Upload and UX Image;
- additional provider-specific URL or storage adapters when their contracts are broadly reusable.


Kocal (changes requested):
Reviewed by checking the branch out locally: ran the full suite (504 tests / 1425 assertions pass on PHP 8.4 with gd + exif + imagick), ran php-cs-fixer and twig-cs-fixer, and probed the processing pipeline with throwaway scripts to confirm the behavioural findings rather than guess at them.

The design holds up well. The metadata-only rendering boundary, the immutable generation keys, `StoragePath` confinement and the bounded-provider regeneration contract are all genuinely well thought out, and the scope section says no to the right things. Most of what follows is about the GD encoding path and render-time attribute handling.

Detailed notes are inline on the diff. The three I would consider blocking:

- **Transparent PNG produces JPEG variants with a black background.** Verified: the corner pixel of a generated variant is `#000000`. Nothing in the suite covers alpha, so it stays invisible until production.
- **`class`, `alt` and `sizes` are not in the reserved-attribute list**, so passing them through `attributes` emits duplicate attributes and the caller's value is silently dropped.
- **`htmlspecialchars($v, \ENT_QUOTES)` drops the default `ENT_SUBSTITUTE`**, so any invalid UTF-8 blanks the whole attribute. Verified: `alt: "Caf\xE9 crème"` renders `alt=""`.

The lossy encoding intermediate, the Twig runtime wiring, the dead DBAL 3 methods, the unreachable non-stream storage path and the dropped TwigComponent attributes bag are API and behaviour decisions that get expensive to change once the package ships, so they are worth settling now rather than after. Everything else is polish.

Really nice work overall. The architecture survives a close read, the security posture is substantive rather than decorative, and the test suite is serious. Its one blind spot is alpha and transparency, which is exactly where the top issue lives.

Kocal (commented):
Line-by-line notes, following up on the summary review above.

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    private function encodeImage(\GdImage $image, string $outputPath, string $format, int $quality): void
+    {
+        new Filesystem()->mkdir(\dirname($outputPath));
+        $encoded = match ($format) {
+            'webp' => imagewebp($image, $outputPath, $quality),
+            'avif' => \function_exists('imageavif') ? imageavif($image, $outputPath, $quality) : throw ImageProcessingException::unsupportedFormat('avif'),
+            'png' => imagepng($image, $outputPath, (int) (9 - ($quality * 9 / 100))),
+            'jpeg', 'jpg' => imagejpeg($image, $outputPath, $quality),
```
Kocal: **Blocking: transparent PNG produces JPEG variants with a black background.**

`resize()` fills the canvas with `imagecolorallocatealpha($dst, 0, 0, 0, 127)` (L312-317), then this line hands that image to `imagejpeg()`. JPEG has no alpha channel, so GD flattens the transparency onto black rather than white.

I ran a transparent PNG through a `jpeg`-only profile and read back the generated variant: the corner pixel is `#000000`.

A transparent logo comes out with a black background in every JPEG variant, and `fill` mode letterboxes in black too. Nothing in the suite covers alpha, so this stays invisible until someone looks at a rendered page. Most libraries flatten onto white; whichever colour you pick it should be a deliberate one, and ideally configurable.

src/Image/src/Processor/Gd/GdImageProcessor.php
```diff
@@ -0,0 +1,410 @@
...
+
+        return $image instanceof \GdImage ? $image : null;
+    }
+
+    private function saveImage(\GdImage $image, string $path, int $type): void
+    {
+        match ($type) {
+            \IMAGETYPE_PNG => imagepng($image, $path),
+            \IMAGETYPE_WEBP => imagewebp($image, $path),
+            default => imagejpeg($image, $path, 90),
```
Kocal, line 408: `saveImage()` picks the encoder from the **source** image type, which makes the shared resized file a lossy intermediate whenever the upload is a JPEG: every variant goes through an extra JPEG q90 generation before the real encode runs.

Measured on a synthetic high-frequency source: mean absolute deviation of 2.64/255 per channel versus encoding straight from the resampled pixels. A WebP or AVIF variant of a JPEG upload can never be better than a recompressed q90 JPEG.

A lossless intermediate (PNG) fixes the quality side. Better still, see the note on L195: the disk round trip is avoidable entirely.

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,410 @@
...
+        $baseName = pathinfo($imageAsset->path, \PATHINFO_FILENAME);
+        $generation = bin2hex(random_bytes(12));
+        $writeSession = null !== $streamStorage ? new ImageWriteSession($streamStorage, $imageAsset->storageName) : null;
+        $localPublications = [];
+        $index = 0;
+        $outputPixels = 0;
+
+        try {
+            foreach ($plan->variants as $plannedVariant) {
+                $resizedPath = $workspace->path(\sprintf('resized-%d.%s', ++$index, pathinfo($originalPath, \PATHINFO_EXTENSION) ?: 'jpeg'));
```
Kocal: The intermediate inherits the source extension, and `resize()` then encodes it with the source's codec (see L382-389).

The round trip through disk also buys nothing here: `$encodingImage` on L208 immediately decodes the file back into memory to loop over `$plan->formats`. Having `resize()` expose a variant that returns the `GdImage` would remove one encode and one decode per variant, and would make the lossy-intermediate problem disappear along the way.

src/Image/src/Renderer/ImageRenderOptions.php
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+        $this->srcset = $normalizedSrcset;
+
+        foreach ($attributes as $name => $value) {
+            if (1 !== preg_match('/^[a-zA-Z_:][a-zA-Z0-9:._-]*$/', $name) || str_starts_with(strtolower($name), 'on')) {
+                throw new \Symfony\UX\Image\Exception\InvalidArgumentException(\sprintf('Unsafe image attribute name "%s".', $name));
+            }
+            if (\in_array(strtolower($name), ['src', 'srcset', 'width', 'height', 'loading', 'fetchpriority', 'decoding'], true)) {
+                throw new \Symfony\UX\Image\Exception\InvalidArgumentException(\sprintf('Image attribute "%s" is managed by ImageRenderOptions.', $name));
```
Kocal: **Blocking: `class`, `alt` and `sizes` are renderer-owned but not reserved here.**

The list covers `src`, `srcset`, `width`, `height`, `loading`, `fetchpriority` and `decoding`, but the renderer also emits `class`, `alt` and `sizes` itself. Passing any of those through `attributes` produces duplicate attributes:

```html
<img src="/up/a_sm.jpg" alt="x" loading="lazy" ... class="rounded" class="shadow" id="hero" />
<img src="/up/a_sm.jpg" alt="x" ... sizes="100vw" ... alt="other" sizes="50vw" />
```

That is invalid HTML, and the browser keeps the first occurrence, so the caller's value is silently dropped. The PR description promises "safe application attributes without allowing overrides of renderer-owned attributes", which currently holds for 7 attributes out of 10.

src/Image/src/Renderer/RenderedImage.php
```diff
@@ -0,0 +1,211 @@
...
+
+    public function __toString(): string
+    {
+        return $this->toHtml();
+    }
+
+    public function toHtml(): string
+    {
+        $sizes = null !== $this->options->sizes && '' !== $this->options->sizes
+            ? \sprintf(' sizes="%s"', htmlspecialchars($this->options->sizes, \ENT_QUOTES))
```
Kocal: **Blocking: passing `\ENT_QUOTES` explicitly drops `ENT_SUBSTITUTE`.**

Since PHP 8.1 the default flags are `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`. Passing `ENT_QUOTES` alone removes the substitution, and `htmlspecialchars()` then returns an **empty string** on invalid UTF-8.

Verified: rendering with `alt: "Caf\xE9 crème"` emits `alt=""`. Not an injection, but the accessibility text disappears with no error and no log, and mis-encoded strings are common when the value comes from a legacy import or a latin1 column.

`\ENT_QUOTES | \ENT_SUBSTITUTE` restores the intended behaviour. This applies to every `htmlspecialchars()` call in the file.

src/Image/src/Twig/ImageExtension.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    public function __construct(
+        private readonly ImageRuntime $imageRuntime,
+    ) {
+    }
+
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('ux_image', [$this, 'renderImage'], ['is_safe' => ['html']]),
```
Kocal: `[$this, 'renderImage']` with `ImageRuntime` injected into the constructor (L27) defeats the point of the `twig.runtime` tag the service carries.

As written, booting Twig eagerly constructs `ImageRuntime` -> `DefaultImageRenderer` -> `UrlGenerator` -> the tagged iterators, even on a page that renders no image. `[ImageRuntime::class, 'renderImage']` defers all of that to first use.

It is also the convention everywhere else in the monorepo: `MapExtension` and `UXIconsTwigExtension` both use that form, take no constructor arguments, and are marked `@internal`.

src/Image/src/Doctrine/ImageAssetType.php
```diff
@@ -0,0 +1,83 @@
...
+
+        if (!$value instanceof ImageAsset) {
+            throw new \Symfony\UX\Image\Exception\InvalidArgumentException('Expected ImageAsset instance, got '.get_debug_type($value));
+        }
+
+        // The ImageAsset object is readonly, so we can safely convert to array
+        return parent::convertToDatabaseValue($value->toArray(), $platform);
+    }
+
+    public function getName(): string
```
Kocal: `getName()` and `requiresSQLCommentHint()` were both removed from `Doctrine\DBAL\Types\Type` in DBAL 4. I checked against the 4.5 that composer resolves here: neither exists on `Type` or on `JsonType`.

Since `composer.json` requires `doctrine/dbal: ^4.0`, these are two public methods that override nothing and are never called. Safe to drop.

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,410 @@
...
+                    }
+                } finally {
+                    unset($encodingImage);
+                }
+            }
+            if (null !== $writeSession) {
+                $writeSession->commit();
+            } else {
+                foreach ($localPublications as [$stagedPath, $absolutePath]) {
+                    new Filesystem()->rename($stagedPath, $absolutePath, false);
```
Kocal: This publication loop has no rollback: the `catch` on L260 only calls `$writeSession?->rollback()`, which is null on exactly this branch. If the third `rename()` fails, the first two stay published.

It is also a broken assumption rather than just a missing guard. `$writeSession` is null only when the storage does not implement `StreamStorageInterface`, which in practice means a custom `adapter_service`, and `$targetDir` (L184) is `dirname($storage->getFilePath($asset))`, so the code treats a possibly remote adapter's return value as a local filesystem path.

Meanwhile `ConfigurationValidator::validateStorages()` already rejects any storage that is not stream-capable, so the branch is unreachable in a validated setup and incorrect if reached. It also contradicts the "partial writes are rolled back in reverse order" claim in the description. Worth either dropping it and failing fast, or supporting it properly. Same code in `InterventionImageProcessor`.

src/Image/templates/components/Image.html.twig
```diff
@@ -0,0 +1,3 @@
+{% if rendered is not null %}
+    {{ rendered.toHtml()|raw }}
```
Kocal: The component silently drops the TwigComponent attributes bag: `<twig:ux:image src="..." id="hero" data-x="1">` renders neither `id` nor `data-x`, because this template only emits `rendered.toHtml()` and `Twig/Components/Image.php` has no attributes handling.

Propagating the attributes bag onto the root element is the single most expected behaviour of a TwigComponent. The prop table in `doc/rendering.md` lists the named props but does not say the rest is discarded; the doc only mentions the PHP-only `attributes` restriction for the Twig *functions*.

src/Image/src/ProcessingLimits.php
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    public function assertOutputAllocation(int $width, int $height): void
+    {
+        if ($width > $this->maxWidth || $height > $this->maxHeight) {
+            throw ImageLimitExceededException::dimensions($width, $height, $this->maxWidth, $this->maxHeight);
+        }
+        if ($width > intdiv($this->maxOutputPixels, $height)) {
+            throw ImageLimitExceededException::outputPixels($this->maxOutputPixels + 1, $this->maxOutputPixels);
```
Kocal: This reports a fabricated pixel count: the message will always read "N+1 exceeds N", so it can never say by how much the budget was blown or which variant caused it.

`$width * $height` is bounded by `maxWidth * maxHeight` (144M with the defaults), so there is no overflow reason to avoid the real value. `GdImageProcessor` L227 passes the actual accumulated count, which makes this read as an oversight rather than a choice. Same pattern in `VariantProcessingPlanner` L79.

src/Image/src/Processor/VariantProcessingPlanner.php
```diff
@@ -0,0 +1,102 @@
...
+                );
+            } catch (\ValueError|\InvalidArgumentException $e) {
+                throw ImageProcessingException::processingFailed('plan variants', $e->getMessage());
+            }
+            $this->limits->assertOutputAllocation($geometry->canvasWidth, $geometry->canvasHeight);
+
+            $artifactPixels = $geometry->canvasWidth * $geometry->canvasHeight;
+            $remaining = $this->limits->maxOutputPixels - $outputPixels;
+            if ([] !== $formats && $artifactPixels > intdiv($remaining, \count($formats))) {
+                throw ImageLimitExceededException::outputPixels($this->limits->maxOutputPixels + 1, $this->limits->maxOutputPixels);
```
Kocal: Same as `ProcessingLimits` L52: `$this->limits->maxOutputPixels + 1` is not the actual count. Here the real value is `$outputPixels + $artifactPixels * \count($formats)`, and the variant name is in scope too, which would make the message actionable.

src/Image/templates/data_collector/image_configuration.html.twig
```diff
@@ -0,0 +1,123 @@
...
+            <b>Profiles</b>
+            <span>{{ collector.profiles|length }}</span>
+        </div>
+        <div class="sf-toolbar-info-piece">
+            <b>Warnings</b>
+            <span>{{ (collector.storageWarnings|length) + (collector.profileWarnings|length) }}</span>
+        </div>
+    {% endset %}
+
+    {{ include('@WebProfiler/Profiler/toolbar_item.html.twig', { link: profiler_url, status: collector.hasWarnings ? 'yellow' : '' }) }}
```
Kocal: `twig-cs-fixer` fails on this line: `Expecting 0 whitespace after "{"` and `Expecting 0 whitespace before "}"`.

Not enforced in CI, but it is step 4 of the pre-commit checklist.

src/Image/config/services.php
```diff
@@ -0,0 +1,234 @@
...
+use Symfony\UX\Image\Command\ValidateConfigurationCommand;
+use Symfony\UX\Image\ConfigurationReporter;
+use Symfony\UX\Image\ConfigurationValidator;
+use Symfony\UX\Image\DataCollector\ImageConfigurationCollector;
+use Symfony\UX\Image\ProcessingLimits;
+use Symfony\UX\Image\Processor\ChainImageProcessor;
+use Symfony\UX\Image\Processor\GdImageProcessor;
+use Symfony\UX\Image\Processor\ImageInspector;
+use Symfony\UX\Image\Processor\ImageInspectorInterface;
+use Symfony\UX\Image\Processor\ImageDriverInterface;
```
Kocal: `php-cs-fixer` wants `ImageDriverInterface` before `ImageInspector` / `ImageInspectorInterface`. It is the only violation across the 122 files of the package, but fabbot will pick it up.

src/Image/src/Processor/ImageInspector.php
```diff
@@ -0,0 +1,82 @@
...
+        if (!$mime) {
+            return null;
+        }
+
+        return match ($mime) {
+            'image/jpeg' => 'jpeg',
+            'image/png' => 'png',
+            'image/webp' => 'webp',
+            'image/avif' => 'avif',
+            'image/gif' => 'gif',
```
Kocal: GIF is handled two different ways by the same class. `resolveFormat()` maps `image/gif` to `gif`, but `InspectedImage::fromPath()` and `ImageDriverCapabilities::gd()` do not know about it.

So `inspect()` reports a GIF as a supported image and `inspectImage()` rejects it with `unsupportedFormat`. If GIF is deliberately out of scope (reasonable, animation makes it awkward), dropping it here would make that explicit.

src/Image/src/Storage/LocalStorage.php
```diff
@@ -0,0 +1,187 @@
...
+    ) {
+        $this->filesystem = new Filesystem();
+    }
+
+    public function store(UploadedFile $file, string $storageName, ?string $directory = null): string
+    {
+        $storageName = $this->validateStorageName($storageName);
+        $source = $file->getRealPath() ?: $file->getPathname();
+        $inspected = InspectedImage::fromPath($source, $this->limits);
+        if ('svg' === $inspected->format) {
```
Kocal: This branch is never reached. `getimagesize()` cannot identify SVG, so `InspectedImage::fromPath()` on the line above has already thrown `unsupportedFormat('unknown')` by the time we get here.

Same guard in `FlysystemStorage` L38. The actual SVG rejection happens in `RejectSvgPolicy`, which does work.

src/Image/README.md
```diff
@@ -0,0 +1,107 @@
...
+keeps ownership of forms, uploads, entities and cleanup; UX Image owns the
+processing profile, stored outputs and rendering contract.
+
+## Install
+
+```bash
+composer require symfony/ux-image
+```
+
+Symfony Flex configures local public storage. The default driver requires
```
Kocal: The README documents a Flex recipe that does not exist yet: this paragraph and the next one ("Flex also registers `TwigBundle` and `UXImageBundle`") describe behaviour the PR description itself lists under "Coming Next".

Anyone following the README today ends up with an unregistered bundle. Worth either softening the wording or landing the recipe first.

Separately, "Last reviewed: July 2026." at the bottom does not appear in any other package README.

src/Image/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# CHANGELOG
+
+## 3.5.0
+
+- Add the component.
```
Kocal: "Add the component." looks like a leftover from the package template, for something that is not a component.

src/Image/src/Processor/VariantProcessingPlanner.php
```diff
@@ -0,0 +1,102 @@
...
+                    $input->height,
+                    $width,
+                    $height,
+                    ResizeMode::from($mode),
+                    FocalPoint::fromString($position),
+                );
+            } catch (\ValueError|\InvalidArgumentException $e) {
+                throw ImageProcessingException::processingFailed('plan variants', $e->getMessage());
+            }
+            $this->limits->assertOutputAllocation($geometry->canvasWidth, $geometry->canvasHeight);
```
Kocal, line 76: Suggestion: consider skipping planned variants whose target exceeds the source.

Upscaling is disabled, so a 400px original under the injected `responsive_default` profile (640 / 1024 / 1920) writes **three identical 400px files**. Their srcset entries then collapse to a single `400w` candidate in `buildSrcset()`, so it is three encodes and three stored objects for one usable entry.

src/Image/src/ImageAsset.php
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    public function getProfileRevision(): ?string
+    {
+        return $this->profileRevision;
+    }
+
+    public function getImageSourceSet(): ImageSourceSet
+    {
+        return ImageSourceSet::fromArray($this->variants);
```
Kocal: Suggestion: this reparses the whole variants array on every call, and it sits on the page render path.

`getVariant()` is the worst case: `getAvailableFormats()` builds one set, then `getForFormat()` inside the loop builds another per format, so a single lookup costs F+1 full parses.

The constructor already builds an `ImageSourceSet` for validation (L48) and throws it away. Assigning it to a private readonly property would fix both, the class stays immutable.

src/Image/src/Processor/Intervention/InterventionImageProcessor.php
```diff
@@ -0,0 +1,347 @@
...
+    {
+        return $this->generateVariantsFromSource($imageAsset, $variantConfigs);
+    }
+
+    /**
+     * @param array<string, mixed> $variantConfigs
+     *
+     * @return array<string, list<array{name: string, path: string, format: string, mimeType: string, width: int, height: int, mode: string, quality: int, position: string, media: string|null, density: string|null}>>
+     */
+    private function generateVariantsFromSource(ImageAsset $imageAsset, array $variantConfigs, ?string $sourcePath = null): array
```
Kocal, line 163: Suggestion: this method is around 110 lines duplicated near-verbatim from `GdImageProcessor::generateVariantsFromSource()`, with only the decode and encode calls differing.

Concretely: the black-background issue, the lossy intermediate and the non-stream publication branch each need fixing twice. A small encoder abstraction (or a shared abstract base holding the plan / workspace / write-session orchestration) would leave only the driver-specific calls in each processor.

src/Image/src/UrlGenerator/CachedUrlGenerator.php
```diff
@@ -0,0 +1,96 @@
...
+            return $url;
+        } catch (\Throwable) {
+            // Graceful degradation: if cache fails, fall back to decorated generator
+            return $this->decorated->generateAssetUrl($asset);
+        }
+    }
+
+    public function generateVariantUrl(ImageAsset $asset, array $variant): string
+    {
+        $variantKey = hash('sha256', serialize([
```
Kocal: Suggestion: for the default `generic` adapter this decorator is a net loss.

It adds a SHA-256 over a `serialize()` plus a PSR-6 get/save round trip in order to replace `rtrim($prefix, '/').'/'.ltrim($path, '/')`. With a Redis pool that is a network round trip per `<source>` per image on the page.

It only pays off for CDN builders that do real work such as signing. Worth saying so in the docs, otherwise the option reads as an optimisation while being a regression for most setups.

Minor, in the same file: the `catch (\Throwable)` fallback re-invokes the decorated generator, so a genuine configuration error (unknown CDN builder) runs twice and surfaces from the catch block.

src/Image/src/Storage/ImageWriteSession.php
```diff
@@ -0,0 +1,70 @@
...
+                }
+                $this->committed[] = $storagePath;
+            }
+        } catch (\Throwable $e) {
+            $this->rollback();
+            throw $e;
+        }
+    }
+
+    public function rollback(): void
```
Kocal, line 68: Suggestion: `rollback()` stops at the first `deletePath()` failure and leaks every remaining committed path.

Wrapping the delete per path (collecting failures rather than propagating the first) would make the rollback best-effort as intended. It also matters that this is called from `commit()`'s catch block, where a throw here replaces the original exception with a less useful one.

src/Image/src/Twig/ImageRuntime.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * @param array<string, mixed> $options
+     */
+    public function renderConfigured(ImageAsset $asset, array $options = []): RenderedImage
+    {
+        $sizes = $options['sizes'] ?? null;
+        $alt = $options['alt'] ?? '';
+        $lazy = $options['lazy'] ?? true;
```
Kocal, line 86: Suggestion: the coercion here is inconsistent. Wrong types are silently normalised (`lazy: 'no'` becomes `true`, `fetchpriority: 123` becomes `'auto'`) while `fetchpriority: 'HIGH'` throws from `ImageRenderOptions`.

Unknown keys are also silently ignored, so a typo in an option name does nothing at all and gives no signal.

src/Image/src/Twig/Components/Image.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        if (!$this->src) {
+            return null;
+        }
+
+        return $this->rendered ??= $this->runtime->render($this->src, new ImageRenderOptions(
+            sizes: $this->sizes,
+            alt: $this->alt ?? '',
+            lazy: $this->lazy,
+            fetchPriority: $this->fetchpriority ?? ($this->lazy ? 'auto' : 'high'),
```
Kocal: Suggestion: two entry points, two defaults. The component sets `fetchpriority: high` when `lazy: false`, while `ux_image()` / `ux_picture()` stay on `auto` in the same situation (`ImageRuntime::renderConfigured()`).

The implicit behaviour is defensible, it just should not depend on which API the user picked.

src/Image/src/ConfigurationValidator.php
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public function validateStorages(): array
+    {
+        if (!$this->storage instanceof StreamStorageInterface) {
+            return [\sprintf('The configured storage must implement %s so image variants can be read and published.', StreamStorageInterface::class)];
+        }
+
+        $errors = [];
+        foreach (array_unique(['default_public', ...array_keys($this->storages)]) as $storageName) {
+            $path = new StoragePath('.ux-image-validation/'.bin2hex(random_bytes(12)).'.probe');
```
Kocal, line 108: Suggestion: worth documenting that `ux:image:validate` writes real objects into every configured storage, including a production bucket, then deletes them.

The `.ux-image-validation/` directory is also left behind after the probe file is removed.

src/Image/doc/index.rst
```diff
@@ -0,0 +1,24 @@
...
+================
+
+Symfony UX Image processes uploaded raster images into versioned ``ImageAsset``
+values, stores originals and responsive variants, and renders optimized
+``<picture>`` markup.
+
+The documentation is organized as a navigable, task-oriented corpus instead of
+one monolithic page:
+
+* `Installation and quick start <installation.md>`_
```
Kocal: These use RST external-link syntax pointing at `.md` files, and `.symfony.bundle.yaml` declares `doc_dir: doc`, so symfony.com would build an index of dead links.

You already flagged the Markdown to RST conversion as pending, just worth gating the docs publication on it.

src/Image/src/ImageAsset.php
```diff
@@ -0,0 +1,286 @@
...
+
+    public function getDefaultFormat(): ?string
+    {
+        $formats = $this->getAvailableFormats();
+
+        if (!$formats) {
+            return null;
+        }
+
+        foreach (['avif', 'webp', 'jpeg', 'jpg', 'png'] as $preferred) {
```
Kocal, line 169: Suggestion: this hardcodes `[avif, webp, jpeg, jpg, png]` and ignores the configurable `ux_image.preferred_formats` (and the per-profile override). An application that changes the preference order does not affect this method, which `DefaultImageRenderer::resolveDimensions()` relies on.

src/Image/src/Renderer/RenderedImage.php
```diff
@@ -0,0 +1,211 @@
...
+
+        if ('' !== $sizes) {
+            $attributes[] = ltrim($sizes);
+        }
+
+        $attributes = [
+            ...$attributes,
+            \sprintf('alt="%s"', htmlspecialchars($this->options->alt, \ENT_QUOTES)),
+            \sprintf('loading="%s"', $this->loadingAttribute()),
+            \sprintf('fetchpriority="%s"', $this->fetchPriorityAttribute()),
```
Kocal: Suggestion: `fetchpriority="auto"` and `decoding="async"` are emitted on every image even at their default values. Both are the browser default, so it is bytes on every page for no behavioural change.

src/Image/src/ImageSource.php
```diff
@@ -0,0 +1,264 @@
...
+        if (null !== $mode && !\in_array($mode, ['crop', 'fit', 'fill'], true)) {
+            throw new Exception\InvalidArgumentException(\sprintf('Unsupported image variant resize mode "%s".', $mode));
+        }
+        if (null !== $quality && ($quality < 1 || $quality > 100)) {
+            throw new Exception\InvalidArgumentException('Image variant quality must be between 1 and 100.');
+        }
+        if (null !== $position && '' === trim($position)) {
+            throw new Exception\InvalidArgumentException('Image variant position must not be empty.');
+        }
+        if (null !== $density && 1 !== preg_match('/^(?:[1-9]\d*(?:\.\d+)?x)$/', $density)) {
```
seb-jean: The density regex `[1-9]\d*...` rejects values below 1 like `0.5x`, which are valid per the [HTML spec (§ 4.8.4.2.1: "a valid floating-point number giving a number greater than zero").](https://html.spec.whatwg.org/multipage/images.html#srcset-attributes)

src/Image/src/Twig/ImageRuntime.php
```diff
@@ -0,0 +1,94 @@
...
+
+        return $this->render($asset, new ImageRenderOptions(
+            sizes: \is_string($sizes) ? $sizes : null,
+            alt: \is_string($alt) ? $alt : '',
+            lazy: \is_bool($lazy) ? $lazy : true,
+            fetchPriority: \is_string($fetchPriority) ? $fetchPriority : 'auto',
+            class: \is_string($class) ? $class : '',
+            decoding: \is_string($decoding) ? $decoding : 'async',
+            variant: \is_string($variant) ? $variant : null,
+        ));
```
seb-jean: `renderConfigured()` extracts 7 of the 9 `ImageRenderOptions` parameters but doesn't forward `srcset` or `attributes`. The Twig functions `ux_image()` / `ux_picture()` silently ignore those options with no error, while the Twig component `<twig:ux:image>` does support `srcset`. Either forward them or document the limitation explicitly.

src/Image/src/Doctrine/ImageAssetType.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function convertToDatabaseValue($value, AbstractPlatform $platform): ?string
+    {
+        if (null === $value) {
+            return null;
+        }
+
+        if (!$value instanceof ImageAsset) {
+            throw new \Symfony\UX\Image\Exception\InvalidArgumentException('Expected ImageAsset instance, got '.get_debug_type($value));
```
Kocal: Let's import the class? Same for all the other files.

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,434 @@
...
+        private readonly ?ImageProcessingDispatcherInterface $asyncDispatcher = null,
+        private readonly ?ProcessingLimits $limits = null,
+    ) {
+    }
+
+    public function process(UploadedFile $file, ?string $profile = null, string $storage = 'default_public'): ImageAsset
+    {
+        $metadata = $this->extractMetadata($file);
+        if ('image/svg+xml' === ($metadata['mime'] ?? null)) {
+            $file = ($this->svgPolicy ?? new RejectSvgPolicy())->process($file);
```
Kocal: no?
```suggestion
            $file = ($this->svgPolicy ??= new RejectSvgPolicy())->process($file);
```

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,434 @@
...
+        $profileConfig = null !== $profile ? $this->profiles[$profile] : null;
+        $directory = $profileConfig['directory'] ?? null;
+        $processing = ProcessingMode::fromProfile($profileConfig);
+        $sourceWorkspace = null;
+        $sourcePath = null;
+        if (ProcessingMode::Immediate === $processing && $this->storageManager instanceof StreamStorageInterface) {
+            $sourceWorkspace = new ProcessingWorkspace();
+            $sourcePath = $sourceWorkspace->materializeLocal(
+                $file->getRealPath() ?: $file->getPathname(),
+                $this->limits ?? new ProcessingLimits(),
```
Kocal:
```suggestion
                $this->limits ??= new ProcessingLimits(),
```

no?

Kocal: Or better, since `$this->limits ?? new ProcessingLimits()` appears multiple times in this file, add a private method, or use `$this->limits ??= new ProcessingLimits()` in the constructor?

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,434 @@
...
+        $configuredVariants = $variantConfigs['variants'];
+        /** @var array<array-key, mixed> $configuredFormats */
+        $configuredFormats = $variantConfigs['formats'];
+        try {
+            $originalPath = null !== $sourcePath
+                ? $sourcePath
+                : (null !== $streamStorage
+                    ? $workspace->materialize($streamStorage, $imageAsset, $limits)
+                    : $this->storageManager->getFilePath($imageAsset));
+            $input = \Symfony\UX\Image\InspectedImage::fromPath($originalPath, $limits);
```
Kocal: Let's import `\Symfony\UX\Image\InspectedImage`

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,434 @@
...
+                    }
+                } finally {
+                    unset($encodingImage);
+                }
+            }
+            if (null !== $writeSession) {
+                $writeSession->commit();
+            } else {
+                foreach ($localPublications as [$stagedPath, $absolutePath]) {
+                    new Filesystem()->rename($stagedPath, $absolutePath, false);
```
Kocal: Also, don't use `new FileSystem()` but import it from the container instead

src/Image/src/Processor/GdImageProcessor.php
```diff
@@ -0,0 +1,434 @@
...
+        try {
+            $originalPath = null !== $sourcePath
+                ? $sourcePath
+                : (null !== $streamStorage
+                    ? $workspace->materialize($streamStorage, $imageAsset, $limits)
+                    : $this->storageManager->getFilePath($imageAsset));
+            $input = \Symfony\UX\Image\InspectedImage::fromPath($originalPath, $limits);
+            $plan = new VariantProcessingPlanner(
+                $limits,
+                $this->geometryCalculator ?? new ResizeGeometryCalculator(),
```
Kocal: Same applies for `$this->geometryCalculator ?? new ResizeGeometryCalculator()` which is repeated two times in this file

src/Image/src/Processor/Gd/GdImageProcessor.php
```diff
@@ -0,0 +1,434 @@
...
+
+        $type = $info[2];
+
+        $src = $this->createImageFromType($inputPath, $type);
+        if (!$src) {
+            throw ImageProcessingException::processingFailed('resize', \sprintf('Unsupported image type %d for "%s".', $type, $inputPath));
+        }
+        if (\IMAGETYPE_JPEG === $type) {
+            $src = ExifOrientation::fromJpeg($inputPath)->applyTo($src);
+        }
```
Kocal, line 307: What if we upload a PNG, a HEIC or a TIFF file?

src/Image/src/Processor/ExifOrientation.php
Kocal, line 1: I would suggest to move this class under `Gd` namespace, same for other processors based on gd

src/Image/src/Processor/InterventionImageProcessor.php
```diff
@@ -0,0 +1,350 @@
...
+        private readonly ?ImageProcessingDispatcherInterface $asyncDispatcher = null,
+        private readonly ?ProcessingLimits $limits = null,
+    ) {
+    }
+
+    public function process(UploadedFile $file, ?string $profile = null, string $storage = 'default_public'): ImageAsset
+    {
+        $metadata = $this->extractMetadata($file);
+        if ('image/svg+xml' === ($metadata['mime'] ?? null)) {
+            $file = ($this->svgPolicy ?? new RejectSvgPolicy())->process($file);
```
Kocal: Same for this one

src/Image/src/Processor/InterventionImageProcessor.php
```diff
@@ -0,0 +1,350 @@
...
+
+    public function supports(string $driver): bool
+    {
+        return 'intervention' === $driver || 'imagick' === $driver || 'vips' === $driver;
+    }
+
+    private function filesystem(): Filesystem
+    {
+        return new Filesystem();
+    }
```
Kocal: Use dependency injection instead

src/Image/src/Profile/ImageProfile.php
```diff
@@ -0,0 +1,34 @@
...
+        public readonly array $formats,
+        public readonly array $variants,
+        public readonly ProcessingMode $processing,
+        public readonly array $configuration,
+    ) {
+    }
+
+    public function revision(): string
+    {
+        return hash('sha256', json_encode($this->configuration, \JSON_THROW_ON_ERROR));
```
Kocal: Using sha256 is done on purpose? Or can we use a faster hashing algorithm xxh?

src/Image/src/Renderer/RenderedImage.php
```diff
@@ -0,0 +1,211 @@
...
+    {
+        return $this->options;
+    }
+
+    public function __toString(): string
+    {
+        return $this->toHtml();
+    }
+
+    public function toHtml(): string
```
Kocal, line 80: I believe we can remove 70% of this file be leveraging `ComponentAttributes::__toString()` from TwigComponent 3.5

src/Image/src/Storage/FlysystemStorage.php
```diff
@@ -0,0 +1,122 @@
...
+    public function __construct(
+        private readonly FilesystemOperator $filesystem,
+        private readonly string $publicUrlPrefix = '',
+        private readonly ?ProcessingLimits $limits = null,
+    ) {
+    }
+
+    public function store(UploadedFile $file, string $storageName, ?string $directory = null): string
+    {
+        new StorageName($storageName);
```
Kocal: You create it but don't use it? Same comment for all other occurences

Kocal (commented):
A new bunch of reviews

Kocal (changes requested):
I had some serious doubts about several points in your PR, so here's a summary posted by Claude that will explain things better than me:

## Why hand-roll the image manipulation?

`src/Processor/` + `src/Transformation/` + `InspectedImage` + `ProcessingLimits` add up to 1886 LOC (27% of src/) plus ~2400 LOC of tests, reimplementing what Intervention Image 3 already ships: GD/Imagick/VIPS drivers, `->orient()` for EXIF, cover/contain/pad resize, focal-point crop, per-format encoders with quality, background flattening on JPEG. intervention/image is already in require-dev, and `InterventionImageProcessor` duplicates `GdImageProcessor` almost line for line. 

The three encoding bugs from the last round (transparent PNG flattened onto black, the q90 lossy intermediate, GIF advertised by the inspector but not decodable) are all edge cases a maintained library already fixed. 

Can we make intervention/image a hard requirement and drop the hand-written GD driver?

## Two delivery models that do not compose

The CDN builders only decorate paths of variants that were already generated and written to storage: `UrlGenerator::generateVariantUrl()` is only ever called with an entry from `ImageAsset::variants`. 
So with Cloudinary you still run the full GD pipeline and write N formats x M widths, then ask Cloudinary to re-transform a derivative. The reverse is not available either: processing: deferred leaves `variants = []`, so `DefaultImageRenderer::buildSources()` returns `[]` and `getFallbackSrcset()` returns `null` -> a bare  with no srcset. 

There is no URL-only mode; I think URL-driven transformation should be the primary model, so I'm opening an RFC instead of settling it here.


## The HTML assembly belongs to html_attr()

Correction to my earlier comment: `ComponentAttributes` lives in symfony/ux-twig-component, which is optional here, but twig/html-extra ^3.29 is already a hard requirement and `HtmlExtension::htmlAttr()` does the same job (merging, escaping, skipping null/false, aria-*, data-*, style, BackedEnum, iterables). 

ImageRuntime already calls `HtmlExtension::htmlAttrValue()`, then RenderedImage redoes the other half by hand with fourteen `htmlspecialchars()` calls. 

Move the `<picture>`/`` skeleton to a Twig template and `RenderedImage` goes from 211 LOC to roughly 60, while the reserved-attribute blacklist in ImageRenderOptions disappears.

Kocal:
I'm sorry but I would be favor of closing this PR for the reasons evoked earlier, but I also think it would take too much work to reorient your proposal toward something that would be truly "practical" for users, without reinventing the wheel—while, and ensuring that the package remains easy to use, maintain, and update.

I opened a RFC https://github.com/symfony/ux/issues/3834 and its associated PR https://github.com/symfony/ux/pull/3838, and I believe it's the pattern Symfony UX should deliver to the users.

PR #3769, opened 2026-08-13
[Upload] Add the UX Upload package

smnandre:
| Q              | A   |
| -------------- | --- |
| Bug fix?       | no  |
| New feature?   | yes |
| Deprecations?  | no  |
| Documentation? | yes |
| Issues         | -   |
| License        | MIT |

## TL;DR;

From a small attachment to a large interrupted transfer, UX Upload keeps one Symfony Form contract: files upload before submit and survive validation errors without being selected or sent again.
Small files use one request, large files can resume, and Symfony receives a signed, lazy `CompletedUpload` while the application owns permanent storage.

## UX Upload

Add one field and render it like any other Symfony Form field:

```php
$builder->add('attachment', FileUploadType::class);
```

```twig
{{ form_row(form.attachment) }}
```

The simple path stays simple. The same field automatically adds progress, validation persistence and resumable transfers when they are needed, without application JavaScript or a custom upload controller.

### Objectives

UX Upload provides a modern upload foundation that feels native to Symfony and Symfony UX:

- integrate with Forms, validation, Twig, translations and Stimulus instead of creating a parallel application workflow;
- make the difficult cases reliable: large files, interrupted transfers, repeated submissions, concurrency and cleanup;
- ship accessible native interaction with bundle-owned, application-overridable Twig markup;
- own temporary transfer concerns without imposing permanent storage, media models or business transactions on applications.

### Features

Transport follows the effective uploader policy automatically:

```mermaid
flowchart LR
    A["Select a file"] ==> B{"size <= chunk_size?"}
    B ==> C["Yes: one multipart POST"]
    B ==> D["No: resumable chunks"]
    C ==> E["Temporary file + signed form value"]
    D ==> E
    E ==> F["Invalid form: re-render without re-upload"]
    E ==> G["Valid form: CompletedUpload"]
    G ==> H["Application storeOnce()"]
    E ==> I["Scheduled TTL cleanup"]
```

- **Symfony-native field:** single or multiple `CompletedUpload` values, named uploader policies and values preserved across invalid form submissions;
- **adaptive transport:** one XHR request with upload progress for small files, resumable parallel chunks for larger files, with pause, retry and cancellation;
- **complete browser UX:** drag-and-drop, keyboard and clipboard selection, progress, speed, estimated time and optional image previews;
- **defence in depth:** signed policies and tokens, context binding, server-side MIME detection, bounded compression, per-part SHA-256 checks and optional whole-file checksums;
- **operational storage:** local or Flysystem temporary storage, locking, lifecycle events and scheduled TTL cleanup;
- **Symfony UX integration:** standard Form rendering, overridable Twig blocks, optional compact and dropzone CSS, a standalone JavaScript API and an optional LiveComponent bridge.

### Out of Scope

- permanent storage, media entities and application database transactions;
- public delivery, download authorization and CDN integration;
- antivirus scanning, transcoding and asynchronous media processing;
- provider-native browser-to-cloud multipart protocols;
- multi-gigabyte ingestion, which should use provider-native direct or multipart upload APIs;
- a no-JavaScript fallback or batching several files into one request.
- the Symfony Flex recipe: this PR ships the Bundle itself, and a second companion PR will add automatic bundle, route and asset registration.

## Requirements

### Required Dependencies

- PHP 8.4 or later;
- Symfony 7.4 or 8.x;
- Symfony UX 3.x and `symfony/stimulus-bundle` 3.0 or later.

Symfony Lock and MIME are required. Lock coordinates concurrent writes, retries and cleanup; MIME supports server-side content validation.

### Optional Dependencies

The uploader, controller, Form type and local temporary storage work without any of the optional packages below. These dependencies are only needed for their corresponding integration:

| Dependency                              | Enables                                                 |
| --------------------------------------- | ------------------------------------------------------- |
| `symfony/twig-bundle ^7.4\|^8.0`        | Built-in Twig form themes and template rendering        |
| `symfony/console ^7.4\|^8.0`            | The `ux:upload:cleanup` command                         |
| `symfony/security-csrf ^7.4\|^8.0`      | CSRF protection on upload mutation endpoints            |
| `symfony/security-bundle ^7.4\|^8.0`    | Automatic binding to the authenticated user             |
| `symfony/rate-limiter ^7.4\|^8.0`       | Upload initialization throttling                        |
| `league/flysystem ^3.0`                 | Flysystem-backed temporary storage                      |
| `symfony/ux-live-component ^2.20\|^3.0` | `ComponentWithUploadTrait` and the LiveComponent bridge |

## Usage

Named uploaders keep transport policy on the server:

```yaml
# config/packages/ux_upload.yaml
ux_upload:
  uploaders:
    documents:
      max_size: 50M
      allowed_types: [application/pdf]
```

```php
$builder->add('document', FileUploadType::class, [
    'uploader' => 'documents',
]);
```

On a valid submission, the field contains a lazy `CompletedUpload`. Its metadata is available without reading storage; `openStream()` is the explicit I/O boundary:

```php
if ($form->isSubmitted() && $form->isValid()) {
    /** @var CompletedUpload|null $upload */
    $upload = $form->get('document')->getData();

    if (null !== $upload) {
        $stream = $upload->openStream();

        try {
            $documentStorage->storeOnce(
                uploadId: $upload->getId(),
                stream: $stream,
                originalName: $upload->getOriginalName(),
                mimeType: $upload->getMimeType(),
            );
        } finally {
            fclose($stream);
        }
    }
}
```

`$documentStorage` and `storeOnce()` are application code. The upload ID is an idempotency key: receiving the same valid request twice must not create two permanent files.

## Configuration

Global defaults apply to every uploader; named uploaders may narrow or override them:

```yaml
ux_upload:
  storage: local
  chunk_size: 5M
  parallel_chunks: 3
  max_size: 100M
  compression: false
  integrity_algorithm: sha256
  completed_ttl: 86400

  local_storage:
    directory: "%kernel.project_dir%/var/uploads"

  uploaders:
    images:
      max_size: 20M
      allowed_types: [image/jpeg, image/png, image/webp]
```

`chunk_size` is both the direct-upload threshold and the part size for resumable uploads. Distributed deployments use Flysystem and must configure a shared lock store.

`integrity_algorithm` selects the optional whole-file checksum attempted by the browser for files up to 64 MiB. The bundled transport independently sends a SHA-256 digest for every direct body or chunk.

## Integration

### Symfony Forms and Twig

`FileUploadType` uses standard Symfony form rendering, validation errors and re-rendering. Its `layout` option selects `compact` or `dropzone`, while `show_preview` independently enables image previews in either layout. Collection entry indices are normalized in the signed field path, so removing or reordering an entry does not invalidate its upload or weaken isolation from other fields.

The form theme owns the complete markup and exposes one composable contract made of ten blocks for the standard Form row, widget, picker, item visual, progress, actions, summary and client errors. Applications import it once with Twig `use`, override one or several related blocks in the same application form theme, and can register that theme globally through `twig.form_themes`. Standard Symfony options render the label, help, row, errors and native file input attributes; `widget_attr` customizes the UX Upload container. Stimulus clones Twig `<template>` elements and updates state, native `hidden` and `disabled` properties, and accessibility attributes without constructing presentation markup. File-specific failures stay in their item; Symfony errors and general client feedback share the field-level presentation.

Two standalone Baseline 2026 stylesheets provide optional compact and dropzone treatments. Both AssetMapper `autoimport` entries are disabled by default, so the bundle remains compatible with an application's existing form theme and CSS.

This PR deliberately introduces a Form type, not a Twig Component: the form is the integration, validation and submission boundary.

### Live Components

`ComponentWithUploadTrait` can assign a completed upload to an explicitly authorized nullable `#[LiveProp]`. `#[UploadTarget]` is the allow-list, while `getUpload()` lazily resolves its signed token.

```php
#[AsLiveComponent]
final class ProfilePhoto
{
    use ComponentWithUploadTrait;

    #[LiveProp]
    #[UploadTarget(uploader: 'avatar')]
    public ?string $photo = null;

    public function getPhotoUpload(): ?CompletedUpload
    {
        return $this->getUpload('photo');
    }
}
```

The property stores the signed token rather than bytes or a path. The bridge uses LiveComponent's public JavaScript API and remains inert when the optional package is absent.

## Implementation

Direct and chunked transfers converge on the same uploader pipeline, storage adapters, validation, events and final token format. The boundary is intentional: UX Upload solves browser transfer and temporary-file reliability while applications retain their own storage and transaction model.

The default direct transport uses XHR because browsers expose upload progress there, unlike `fetch()`. Explicitly injected `fetch` transports remain supported for custom frontends, but direct progress then depends on an injected XHR factory. A direct network failure is never replayed automatically because the server may already have completed it; only an explicit `413` falls back to chunks.

Metadata access does not contact storage. Only `openStream()` may fetch bytes, which matters for remote backends and repeated form requests. Temporary files remain cleanup candidates, so applications copy the bytes they need and make that operation idempotent. Cleanup stays outside the form request because an immediate delete would make a repeated request fail.

The whole-file checksum is optional and may be absent from `CompletedUpload`. Applications that require a checksum for every file calculate it while copying into permanent storage or use a checksum supplied by their storage provider.

The built-in Flysystem adapter accepts any configured `FilesystemOperator`, including S3, Azure or GCS adapters. `MultipartUploadStorageInterface` lets a custom server-side storage complete provider-native parts, but UX Upload deliberately does not implement presigned browser-to-cloud ingestion.

The cleanup command intentionally exposes an age threshold rather than a generic batch limit. Very large remote inventories need provider pagination, lifecycle policies or an application-owned cleanup worker. Applications that increase file count, chunk size or parallelism also own the resulting browser, proxy and storage capacity.

## Security

Upload endpoints accept untrusted browser input and write potentially large objects. The Bundle therefore applies security at every transition instead of relying on the form submission alone.

### Requests and Authorization

- signed upload policies prevent the browser from widening uploader or field limits;
- signed, expiring chunk and resume URLs constrain follow-up mutations, while direct uploads require the signed field policy and, when available, CSRF protection;
- completed tokens are bound to their owner, tenant and canonical form-field path to prevent cross-context replay, while dynamic collection indices are normalized to survive reordering;
- CSRF protection is applied when the component is installed, while an optional rate limiter can throttle initialization.

### File Validation

- declared metadata, transmitted bytes and assembled size are checked independently;
- MIME type is detected from assembled content, never trusted from the browser;
- the bundled browser transport sends a SHA-256 digest for every part, while files up to 64 MiB may also carry an optional SHA-256, SHA-384 or SHA-512 whole-file checksum;
- compressed parts are accepted only when enabled and decompressed within the configured size bound.

### Storage and Concurrency

- generated paths remain inside the configured temporary prefixes;
- Symfony Lock serializes writes, completion, quota checks and cleanup;
- distributed mode requires Flysystem and an explicitly acknowledged shared lock;
- local storage is rejected for distributed deployments.

## Tests

- **PHP:** 456 tests, 4,077 assertions and 100% class, method and line coverage;
- **JavaScript:** 157 Vitest tests across 10 files;
- **Browser:** 15 Chromium scenarios with 97 assertions run in a dedicated CI job, covering real Forms/Stimulus integration, labels, help, validation errors, Bootstrap and Tailwind row delegation, direct and multiple uploads, local and global form themes, partial templates, attribute escaping, unstyled action states, inherited color schemes, accessibility and duplicate-submission prevention;
- **Storage and concurrency:** shared LocalStorage, InMemoryStorage and Flysystem contracts plus multi-process chunk tests;
- **Quality gates:** PHP CS Fixer, TypeScript, Oxlint, formatting, reproducible asset build with committed `dist` verification and Composer security audit.

## Documentation

The documentation follows the user journey from the first field to production operation:

The detailed guide is authored as 19 publishable reStructuredText pages. It covers the first field, transport and storage concepts, customization, security, integrations and production operation.

| Section              | Focus                                      | Pages                                                                                                                                                                                                                                      |
| -------------------- | ------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Start                | Product overview and installation          | [`index.rst`](src/Upload/doc/index.rst), [`installation.rst`](src/Upload/doc/installation.rst)                                                                                                           |
| Build a form         | Fields, values and named policies          | [`form.rst`](src/Upload/doc/form.rst), [`named-uploaders.rst`](src/Upload/doc/named-uploaders.rst)                                                                                                      |
| Understand the model | Ownership, lifecycle and temporary storage | [`architecture.rst`](src/Upload/doc/architecture.rst), [`upload-lifecycle.rst`](src/Upload/doc/upload-lifecycle.rst), [`storage.rst`](src/Upload/doc/storage.rst), [`persisting-uploaded-files.rst`](src/Upload/doc/persisting-uploaded-files.rst) |
| Customize            | JavaScript, Twig, CSS and server events    | [`javascript.rst`](src/Upload/doc/javascript.rst), [`customizing-upload-field.rst`](src/Upload/doc/customizing-upload-field.rst), [`events.rst`](src/Upload/doc/events.rst)                              |
| Build robust flows   | Validation, security, retry and resume     | [`validation.rst`](src/Upload/doc/validation.rst), [`security.rst`](src/Upload/doc/security.rst), [`retry-and-resume.rst`](src/Upload/doc/retry-and-resume.rst)                                         |
| Integrate            | LiveComponent bridge                       | [`live-component.rst`](src/Upload/doc/live-component.rst)                                                                                                                                               |
| Operate              | Deployment, testing and debugging          | [`production.rst`](src/Upload/doc/production.rst), [`testing.rst`](src/Upload/doc/testing.rst), [`debugging.rst`](src/Upload/doc/debugging.rst)                                                         |
| Reference            | Complete configuration                     | [`configuration.rst`](src/Upload/doc/configuration.rst)                                                                                                                                                 |

## Coming Next

Companion pull requests will address:

- a second PR with the Symfony Flex recipe for automatic bundle, route and asset registration;
- the ux.symfony.com package pages and focused interactive demos;
- cross-bundle integration examples, starting with Autocomplete-backed metadata.


.github/workflows/browser-tests.yml
```diff
@@ -27,6 +27,60 @@ concurrency:
...
+              working-directory: src/Upload
+
+            - name: Install Playwright PHP browsers
+              run: vendor/bin/playwright-install --with-deps
+              working-directory: src/Upload
+
+            - name: Run UX Upload PHP E2E tests
+              run: vendor/bin/phpunit --group integration
+              working-directory: src/Upload
+
```
Kocal, line 83: 👎🏻  for duplicating the whole job and introducing a 3rd way for E2E tests, please either use the existing E2E app with Playwright (JS) tests, or in PHP with Zenstruck Browser

src/Upload/composer.json
```diff
@@ -0,0 +1,78 @@
...
+        "symfony/ux-live-component": "^2.20|^3.0"
+    },
+    "suggest": {
+        "league/flysystem": "For advanced storage backend support (S3, GCS, Azure)",
+        "symfony/console": "For the ux:upload:cleanup command",
+        "symfony/rate-limiter": "For throttling upload initialization with a named Framework rate limiter (opt-in via ux_upload.rate_limiter)",
+        "symfony/security-bundle": "For automatically binding uploads and tokens to the authenticated Symfony user",
+        "symfony/ux-live-component": "For the ComponentWithUploadTrait bridge that pushes a completed upload onto a Live Component prop without a form submission",
+        "symfony/security-csrf": "For CSRF protection on all upload mutation endpoints"
+    },
```
Kocal, line 57: We don't use `suggest` in Symfony

Kocal (changes requested):
I wanted to play with the package while reviewing, but I got the following error when installing it:
```
Script cache:clear returned with error code 255
!!
!!  Fatal error: Constant expression contains invalid operations in /Users/kocal/workspace/symfony/ux/src/Upload/src/Form/FileUploadType.php on line 46
!!  Symfony\Component\ErrorHandler\Error\FatalError {#4465
!!    #message: "Compile Error: Constant expression contains invalid operations"
!!    #code: 0
!!    #file: "/Users/kocal/workspace/symfony/ux/src/Upload/src/Form/FileUploadType.php"
!!    #line: 46
!!    -error: array:4 [
!!      "type" => 64
!!      "message" => "Constant expression contains invalid operations"
!!      "file" => "/Users/kocal/workspace/symfony/ux/src/Upload/src/Form/FileUploadType.php"
!!      "line" => 46
!!    ]
!!  }
!!
```

Please ensure your code is compatible with PHP 8.4 🙏🏻

src/Upload/doc/installation.rst
```diff
@@ -0,0 +1,107 @@
...
+============
+
+Requirements
+------------
+
+- PHP 8.4 or later;
+- Symfony 7.4 or 8.0;
+- ``symfony/stimulus-bundle`` 3.0 or later.
+
+TwigBundle enables the built-in form themes. Console enables ``ux:upload:cleanup``. Both integrations are optional package capabilities.
```
Kocal, line 11: Do we really do that in Symfony packages?

src/Upload/doc/installation.rst
```diff
@@ -0,0 +1,107 @@
...
+    Before you start, make sure you have `StimulusBundle configured in your app`_.
+
+.. code-block:: terminal
+
+    $ composer require symfony/ux-upload
+
+This initial release does not include a Symfony Flex recipe. Register the Bundle
+and its routes explicitly:
+
+::
```
Kocal, line 27: To be applied everywhere:
```suggestion
and its routes explicitly::
```

src/Upload/doc/installation.rst
```diff
@@ -0,0 +1,107 @@
...
+
+    # config/routes/ux_upload.yaml
+    ux_upload:
+        resource: '@UXUploadBundle/config/routes.php'
+        prefix: /_ux/upload
+        trailing_slash_on_root: false
+
+This route import uses ``/_ux/upload`` for the internal endpoints. Change
+``prefix`` when the application needs another path. Keep
+``trailing_slash_on_root: false`` so the direct endpoint has no trailing slash.
```
Kocal, line 45: Could be reduced to IMHO:
```suggestion
This route import uses ``/_ux/upload`` for the internal endpoints.
```

src/Upload/doc/installation.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code-block:: terminal
+
+    $ php bin/console importmap:install
+
+UX Upload does not load CSS automatically. Enable either
+``@symfony/ux-upload/dist/compact.css`` or
+``@symfony/ux-upload/dist/dropzone.css`` in ``assets/controllers.json``. See
+:doc:`Customizing the Upload Field <customizing-upload-field>`.
+
+With Webpack Encore, install and build the dependencies managed by
```
Kocal, line 81:
```suggestion
With Reprise, install and build the dependencies managed by
```

Kocal: Anyway I'm not sure about `the dependencies managed by StimulusBundle`, we are talking about npm dependencies right? 🤔

src/Upload/doc/form.rst
```diff
@@ -0,0 +1,162 @@
...
+                'uploader' => 'documents',
+                'required' => true,
+                'show_preview' => false,
+            ]);
+        }
+    }
+
+The submitted value is ``CompletedUpload|null``:
+
+::
```
Kocal, line 35:
```suggestion
The submitted value is ``CompletedUpload|null``::
```

etc

Kocal (changes requested):

src/Upload/assets/test/unit/preview-cache.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    describe('retrieve', () => {
+        it('returns null for uncached token', async () => {
+            const result = await cache.retrieve('nonexistent');
+            expect(result).toBeNull();
+        });
+
+        it('returns blob URL for cached token', async () => {
+            // Manually put something in the mock cache
+            const blob = new Blob(['fake-image'], { type: 'image/jpeg' });
+            const response = new Response(blob, {
```
Kocal: These two specs fail on the environment CI actually runs.

```
PreviewCache retrieve returns blob URL for cached token   preview-cache.test.ts:111
PreviewCache remove deletes a cached entry                preview-cache.test.ts:132
TypeError: object.stream is not a function
    at extractBody (node:internal/deps/undici/undici:5485:25)
    at new Response (node:internal/deps/undici/undici:9135:41)
```

jsdom 26.1.0, which is what the committed lockfile resolves, has no `Blob.prototype.stream`, and Node's undici `Response` constructor calls it. I reproduced on Node v22.18.0, which matches `.nvmrc` (`22.18`), and `unit-tests.yaml` runs `pnpm run test:unit` from the root across the whole workspace, so the `js` job will hit this too. Result here was 155 passed / 2 failed, against the 157 claimed in the description.

The fix belongs in the test rather than in `preview-cache.ts`: build the `Response` from a string or an `ArrayBuffer`, or polyfill `Blob.prototype.stream` in `test/setup.ts`.

src/Upload/src/Uploader.php
```diff
@@ -0,0 +1,546 @@
...
+            // Storage duplicate detection is exists-then-write on most backends:
+            // without serialization, two concurrent requests for the same chunk
+            // could both pass the existence check and both write. The lock is keyed
+            // per (uploadId, chunkIndex) -- the finest granularity that closes the
+            // race -- so distinct chunks of the same upload still store in parallel
+            // (the whole point of parallel_chunks). It blocks rather than fails fast
+            // so the loser waits for the winner, then observes the chunk already
+            // exists and is rejected cleanly instead of corrupting the stored bytes.
+            //
+            $lockKey = \sprintf('ux_upload_chunk_%s_%d', $uploadId, $chunkIndex);
```
Kocal: The per-`(uploadId, chunkIndex)` lock here can never contend, because line 260 already takes a blocking per-upload lock (`ux_upload_lifecycle_{uploadId}`) around the whole method body. No two processes are ever inside `storeChunk()` for the same upload at the same time, so every chunk write for an upload is serialized end to end, digest computation and storage write included. `parallel_chunks` buys nothing on the server side; the client sends in parallel and the server processes one at a time.

Two places currently document the opposite:

- the comment on lines 270-278, "distinct chunks of the same upload still store in parallel — the whole point of parallel_chunks";
- the `ConcurrentDistinctChunkWriteTest` docblock (lines 19-21 and 27), "A per-upload lock would have needlessly serialized them". That test still passes because it only asserts that every write lands, which serialization satisfies just as well.

Two ways out, and I think it is worth choosing deliberately rather than patching: drop the inner lock as dead weight and correct both comments, or make the lifecycle lock shared for chunk writes and exclusive for complete/cancel so the advertised parallelism is real.

src/Upload/src/UXUploadBundle.php
```diff
@@ -0,0 +1,309 @@
...
+                ->integerNode('completed_ttl')->defaultValue(Uploader::DEFAULT_COMPLETED_TTL)->min(60)->info('Time in seconds before an assembled temporary upload expires')->end()
+                ->integerNode('max_pending_per_owner')->defaultValue(1000)->min(1)->end()
+                ->booleanNode('compression')->defaultFalse()->info('Enable gzip compression for chunks')->end()
+                ->enumNode('integrity_algorithm')->values(Uploader::INTEGRITY_ALGORITHMS)->defaultValue('sha256')->info('Hash algorithm used for optional end-to-end file integrity verification')->end()
+                ->scalarNode('max_size')->defaultValue(Uploader::DEFAULT_MAX_SIZE)->info('Maximum file size enforced server-side (0 = no limit, e.g. "100M")')->end()
+                ->arrayNode('allowed_types')->scalarPrototype()->end()->defaultValue([])->info('Allowed MIME types enforced server-side (empty = all types)')->end()
+                ->scalarNode('rate_limiter')->defaultNull()->info('Service ID of a named framework.rate_limiter factory used to throttle upload initialization (for example "limiter.ux_upload_init")')->end()
```
Kocal, line 290: The default install ends up as an unauthenticated, unthrottled write endpoint. `UploadRouteLoader` registers `/init`, `/`, `/resume` and `/remove` with no access control, `rate_limiter` defaults to `null`, and `isCsrfValid()` returns `true` when the token manager is null (which is the default, since `symfony/security-csrf` sits in `require-dev`). On top of that `AnonymousUploadContextResolver` returns a null owner, so every anonymous visitor shares one `UploadContext`.

With `max_pending_per_owner: 1000` and `max_size: 100M` that lets one actor park roughly 100 GB under `var/uploads`, and because the quota is shared rather than per-visitor, the same actor also locks every other anonymous user out of uploading at all.

For a bundle heading into the Symfony org I would argue for the opposite default, or failing that a prominent warning in `installation.rst` plus a Flex recipe that wires a limiter out of the box.

src/Upload/src/Storage/LocalStorage.php
```diff
@@ -0,0 +1,374 @@
...
+                } elseif (!isset($metadata['completedPath']) && $createdAt < $cutoffTime) {
+                    $this->abort($uploadId);
+                }
+            } finally {
+                $lock->release();
+            }
+        }
+    }
+
+    public function countPendingByContext(UploadContext $context): int
```
Kocal: This globs `tmp/*/metadata.json` and `json_decode`s every file on every single `init` call, and `Uploader::acquireQuotaLock()` serializes those calls on `ux_upload_quota_<hash>`. For anonymous uploads `UploadContext::fingerprint()` is a constant, so that is one global mutex for the whole application. At the default quota of 1000 that is up to 1000 file reads per init, one request at a time.

Compounds the throttling point above: the cost of an init grows with the number of pending sessions an attacker has already created.

src/Upload/src/Storage/LocalStorage.php
```diff
@@ -0,0 +1,374 @@
...
+        if (false === $handle) {
+            $existing = @file_get_contents($path);
+            if (false !== $existing && hash_equals($digest, hash('sha256', $existing))) {
+                return ChunkWriteResult::AlreadyPresent;
+            }
+
+            throw new InvalidArgumentException(\sprintf('Chunk %d has already been uploaded with different content.', $index));
+        }
+        try {
+            if (false === fwrite($handle, $data)) {
```
Kocal: `fwrite` returns a byte count on a short write, not `false`, so a truncated chunk is silently accepted here. Line 310 in the same file gets it right with `\strlen($chunk) !== fwrite($finalFile, $chunk)`.

The mismatch does surface eventually, but as a confusing "assembled file size does not match declared file size" at completion rather than at the point of failure.

```php
if (\strlen($data) !== fwrite($handle, $data)) {
    throw new StorageException(\sprintf('Unable to write chunk %d.', $index));
}
```

src/Upload/config/services.php
```diff
@@ -0,0 +1,150 @@
...
+            ->private()
+    ;
+
+    // -- Token ---------------------------------------------------------------
+
+    $services->set(CompletedUploadAccess::class)
+        ->arg('$storage', service(StorageInterface::class))
+    ;
+    $services->set('ux_upload.token_handler', UploadTokenHandler::class)
+        ->arg('$signer', service('uri_signer'))
```
Kocal, line 43: `UploadPolicySigner`, `ResumeTokenHandler` and `UploadTokenHandler` all sign `'?'.http_build_query($payload)` with the same `uri_signer` service and no type tag. I walked all six cross-type directions and the differing required-key sets make a practical replay very unlikely, so this is not a live hole. Still, a one-character discriminator in each payload would make it structurally impossible rather than impossible by coincidence, cheap insurance in a bundle where the token format may well grow new fields later.

src/Upload/src/Uploader.php
```diff
@@ -0,0 +1,546 @@
...
+        $output = '';
+        $length = \strlen($compressed);
+        $blockSize = 64 * 1024;
+        for ($offset = 0; $offset < $length; $offset += $blockSize) {
+            $flush = $offset + $blockSize >= $length ? \ZLIB_FINISH : \ZLIB_SYNC_FLUSH;
+            $decoded = @inflate_add($context, substr($compressed, $offset, $blockSize), $flush);
+            if (false === $decoded) {
+                throw new ValidationException(\sprintf('Failed to decompress gzip chunk %d. The data appears to be corrupted.', $chunkIndex));
+            }
+            if (\strlen($output) + \strlen($decoded) > $this->chunkSize) {
```
Kocal, line 529: The size bound is checked after `inflate_add()` has already returned the decoded block. At zlib's roughly 1032:1 ceiling a single 64 KiB input block can materialize about 66 MiB before this check fires, regardless of how small `chunk_size` is configured.

Still bounded, so not a memory-exhaustion hole, but the peak is much larger than `chunk_size` suggests. Shrinking the block size, or bounding against the remaining allowance before feeding the next block, would keep the peak closer to the configured limit.

src/Upload/src/Uploader.php
```diff
@@ -0,0 +1,546 @@
...
+            throw new ValidationException('This distributed upload storage requires a configured shared lock.');
+        }
+    }
+
+    private function assertAssemblyIntegrity(int $actualSize, ?string $actualHash, int $expectedSize, ?string $expectedHash): void
+    {
+        if ($actualSize !== $expectedSize) {
+            throw new ValidationException(\sprintf('Assembled file size %d bytes does not match declared file size of %d bytes.', $actualSize, $expectedSize));
+        }
+        if (null !== $expectedHash && '' !== $expectedHash && (null === $actualHash || !hash_equals($expectedHash, $actualHash))) {
```
Kocal: `hash_equals($expectedHash, $actualHash)` compares the client-supplied whole-file hash case-sensitively, while `storeChunk()` normalizes with `strtolower($digest)` for the per-chunk digest. A custom client sending uppercase hex therefore passes every chunk and then fails at assembly, which is a confusing place to find out.

src/Upload/src/Url/SymfonyUploadUrlGenerator.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $url = $this->urlGenerator->generate(
+            'ux_upload_handle',
+            ['uploadId' => $uploadId],
+            UrlGeneratorInterface::ABSOLUTE_URL,
+        );
+
+        // Append expiration timestamp
+        $expires = time() + $this->signatureExpiry;
+        $url .= (str_contains($url, '?') ? '&' : '?').'_expires='.$expires;
```
Kocal: `UriSigner::sign($uri, $expiration)` has handled expiration natively since Symfony 7.1, so the hand-rolled `_expires` parameter and its matching check on line 61 could both go.

src/Upload/assets/src/uploader.ts
```diff
@@ -0,0 +1,980 @@
...
+                    // Response body is not JSON -- use statusText
+                }
+                throw new Error(`Chunk upload failed: ${message}`);
+            }
+        } catch (error) {
+            if (error instanceof DOMException && error.name === 'AbortError') {
+                throw error;
+            }
+
+            // Retry with exponential backoff
```
Kocal, line 737: The retry treats every failure the same and burns three attempts with exponential backoff, so a 403 from a stale CSRF token or a 400 for an out-of-range chunk index costs about seven seconds before the user sees anything. Worth short-circuiting on 4xx other than 408 and 429.

src/Upload/assets/src/uploader.ts
```diff
@@ -0,0 +1,980 @@
...
+
+        // Ensure chunks is an array
+        if (!Array.isArray(chunks)) {
+            chunks = [];
+        }
+
+        let completedCount = existingCount;
+
+        // Process chunks in batches
+        for (let i = 0; i < chunks.length; i += parallel) {
```
Kocal: This is a strict batch barrier rather than a sliding window: each batch of `parallel` chunks waits on its slowest member before the next batch starts. On a connection with variable per-chunk latency that leaves the configured parallelism partly idle. A small worker pool pulling from the queue would use it fully.

src/Upload/composer.json
```diff
@@ -0,0 +1,78 @@
...
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=8.4",
+        "symfony/config": "^7.4|^8.0",
+        "symfony/dependency-injection": "^7.4|^8.0",
+        "symfony/event-dispatcher": "^7.4|^8.0",
+        "symfony/form": "^7.4|^8.0",
+        "symfony/http-kernel": "^7.4|^8.0",
```
Kocal, line 30: `symfony/http-foundation` is missing from `require` even though the bundle uses `UriSigner`, `Request`, `Response`, `JsonResponse` and `UploadedFile` directly. It resolves transitively through `symfony/http-kernel`, so nothing breaks, but the convention is to declare what you import.

src/Upload/src/Form/FileUploadType.php
```diff
@@ -0,0 +1,532 @@
...
+        $message = self::TRANSLATION_FALLBACKS[$id] ?? $id;
+
+        if (null === $this->translator) {
+            return strtr($message, array_map(strval(...), $parameters));
+        }
+
+        return $this->translator->trans($id, $parameters, self::TRANSLATION_DOMAIN);
+    }
+
+    private function formatBytes(int $bytes): string
```
Kocal, line 440: The thresholds are 1024-based but the labels say KB, MB and GB. Either divide by 1000 or label them KiB, MiB and GiB.

Kocal (changes requested):
Nice work, the design holds together and the security thinking shows through: signed policies, owner/tenant binding, content-based MIME detection, bounded decompression, lock-serialized writes. I checked a few things rather than assuming, and these came back clean:

- I rebuilt Dropzone, LiveComponent and Upload with the modified `bin/build_package.ts`. All three reproduce their committed `dist/` byte-for-byte, so the `css_source` array support and the conditional `onlyBundle: ['idiomorph']` are backward compatible.
- No `innerHTML` / `insertAdjacentHTML` / `createContextualFragment` anywhere in `assets/src/*.ts`. Filenames go through `textContent` and `<template>` cloning, and the Twig theme has no `|raw`.
- `Uploader.request()` injects `X-CSRF-Token` on every non-GET/HEAD call, so chunk PUT, cancel DELETE and `/resume` POST all line up with what `UploadController::handle()` requires.
- `UploadTokenHandler::isTemporaryPath()` rejects absolute paths, NUL bytes, `..` segments and anything outside `completed_prefix` before a path can reach `Path::join()`.
- No unused RST link definitions, so DOCtor-RST should stay quiet.

I left inline comments on the rest. The two that block are the failing Vitest specs and the chunk-locking question; the rest are ordinary review notes.

One process note: 30k lines in a single commit is hard to review well. Splitting core/storage, form, Stimulus, LiveComponent bridge and docs into separate commits inside the same PR would help a lot.

bin/build_package.ts
Kocal, line 1: Do we really need `css_min_suffix` option?

smnandre:
Test failures unrelated (autocomplete usual problems)

@Kocal do you want me to split PR in smaller commits, now i've closed all feedback It would be less vital to keep the discussion I suppose ?

Kocal:
Sorry didn't see you asked me something. A small squash + rebase, and feel free to merge and continue to iterate

PR #3771, opened 2026-08-13
[Icons] Migrate to modern `\Dom\XMLDocument` API

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

While working on Toolkit tests performance improvements, my friend Claude and me noticed that these tests were slow because all the icons defined in kits examples are fetched from Iconify when running the tests.

When trying to optimize things when configuring `auto_lock`, etc... the tests passed the first time (when it downloads icons), but they didn't pass when running them a 2nd time (when icons are already downloaded), because the `xmlns` attribute disappears between `fromFile()` and `fromBody()`.

When using the API `\DOMDocument::loadXML()`, the attribute `xmlns` was not preserved when using `array_map(..., [...$svgElement->attributes])` to rebuild attributes, so `xmlns` was not outputed.

Moving to the modern `\Dom\XmlElement` API from PHP 8.4 is the way to go IMHO, and it fixes the issue.

Note: this is also something I wanted to do after fixing the security issue https://github.com/symfony/ux-icons/commit/567f33ddffc25504788abc61977565381e424ada, but didn't find enough time.


src/Icons/tests/Integration/RenderIconsInTwigTest.php
```diff
@@ -26,13 +26,13 @@ public function testRenderIcons()
...
-                    <li id="sixth"><svg viewBox="0 0 24 24" fill="currentColor" class="w-6 h-6" aria-labelledby="foo"><path fill-rule="evenodd" d="M19.916 4.626a.75.75 0 0 1 .208 1.04l-9 13.5a.75.75 0 0 1-1.154.114l-6-6a.75.75 0 0 1 1.06-1.06l5.353 5.353 8.493-12.74a.75.75 0 0 1 1.04-.207Z" clip-rule="evenodd"></path></svg></li>
-                    <li id="seventh"><svg viewBox="0 0 24 24" fill="currentColor" class="w-6 h-6" aria-hidden="true"><path fill-rule="evenodd" d="M19.916 4.626a.75.75 0 0 1 .208 1.04l-9 13.5a.75.75 0 0 1-1.154.114l-6-6a.75.75 0 0 1 1.06-1.06l5.353 5.353 8.493-12.74a.75.75 0 0 1 1.04-.207Z" clip-rule="evenodd"></path></svg></li>
+                    <li id="first"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor" class="h-8 w-8" aria-hidden="true"><path fill-rule="evenodd" d="M7.5 6a4.5 4.5 0 1 1 9 0 4.5 4.5 0 0 1-9 0ZM3.751 20.105a8.25 8.25 0 0 1 16.498 0 .75.75 0 0 1-.437.695A18.683 18.683 0 0 1 12 22.5c-2.786 0-5.433-.608-7.812-1.7a.75.75 0 0 1-.437-.695Z" clip-rule="evenodd"></path></svg></li>
```
Kocal, line 29: With the fix, attribute `xmlns` is nicely rendered.

Of course, folks testing the icon rendering bytes-by-bytes in their projects will have to update their tests, but it can't be considered as a BC since the icon is still correctly rendered on the page. We already had the debate months/years ago

kbond (approved):
Makes sense!

PR #3773, opened 2026-08-13
[LiveComponent] Add LiveResponse::remove() to trigger component deletion

smnandre:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  |  no
| Documentation? | yes
| Issues         | Fix #... 
| License        | MIT

> [!WARNING]
> Based on #3761, which introduces `LiveResponse`. The diff shown here includes that
> PR's commit until it is merged. Only the second commit belongs to this one.

--

A `LiveAction` can already ask the browser for a download alongside the re-render. It
cannot ask for the opposite: that the component leave the page.

```php
#[AsLiveComponent]
class NotificationBanner
{
  use DefaultActionTrait;

  #[LiveProp]
  public Notification $notification;

  #[LiveAction]
  public function dismiss(NotificationRepository $repository): LiveResponse
  {
      $repository->markAsRead($this->notification);

      return LiveResponse::remove();
  }
}
```

The server answers `204` with `X-Live-Remove: 1` and skips the render altogether,
`PreReRender` hooks included, since nothing will be shown.

Nothing is deleted server-side. This ends the component on the page, and says nothing
about your data.

### Tearing down before removing

On the client the component is torn down **first**, before the element goes: polling
stops, it leaves the registry, its props are stripped from the element, and it refuses
any further request.

That ordering is what makes the removal safe rather than merely tidy. The element keeps
its event listeners until it is actually dropped, so without the teardown a click during
that window would send an action for a component that is already gone.

### Animating it out

The element carries a `data-live-removing` attribute on its way out, and is only dropped
once whatever the page animates on it has finished:

```css
.notification {
  transition: opacity 300ms, translate 300ms;
}

.notification[data-live-removing] {
  opacity: 0;
  translate: 2rem 0;
}
```

Nothing to declare beyond the CSS. With no animation on `[data-live-removing]` there is
nothing to wait for and the element goes on the next frame, so the default costs a frame
and no configuration. An endless animation is ignored, as it would keep the element on
the page forever.

### Notes

`LiveResponse::remove()` replaces the render rather than riding alongside it, unlike the
download responses, so a property the action changed is never displayed. Like them, it
can only be returned from a `LiveAction` or a `LiveListener`, over POST.


src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
```diff
@@ -360,6 +361,12 @@ public static function getSubscribedEvents(): array
 
     private function createResponse(MountedComponent $mounted, ?LiveResponse $liveResponse = null): Response
     {
+        if ($liveResponse?->isRemove()) {
+            // the component leaves the page, so it is never rendered: the pre-re-render hooks
+            // are skipped too, as they exist to prepare a render that will not happen
+            return new Response(null, 204, [self::REMOVE_HEADER => '1']);
+        }
```
Kocal: Since we don't trigger the pre-re-render hooks on purpose, maybe we can trigger new pre-remove hooks? 
Would it makes sense?

src/LiveComponent/src/Controller/BatchActionController.php
```diff
@@ -51,11 +56,15 @@ public function __invoke(Request $request, MountedComponent $_mounted_component,
 
             $response = $this->kernel->handle($subRequest, HttpKernelInterface::SUB_REQUEST, false);
 
+            if ($subRequest->attributes->has('_live_response')) {
+                $liveResponse = $subRequest->attributes->get('_live_response');
+            }
+
             if ($response->isRedirection()) {
                 return $response;
             }
```
Kocal, line 65:
```suggestion
            if ($response->isRedirection() || $liveResponse->isRemove()) {
                return $response;
            }
```

src/LiveComponent/assets/src/Component/index.ts
```diff
@@ -308,18 +357,20 @@ export default class Component {
...
 
+            const html = await backendResponse.getBody();
+
             // if the response does not contain a component, render as an error
-            const headers = backendResponse.response.headers;
             if (
                 !headers.get('Content-Type')?.includes('application/vnd.live-component+html') &&
-                !headers.get('X-Live-Redirect')
+                !headers.get('X-Live-Redirect') &&
+                !headers.get('X-Live-Remove')
```
Kocal:
```suggestion
                !headers.has('X-Live-Remove')
```

Since you already use `this.response.headers.has('X-Live-Remove')` before

Kocal:
Let's document that `emit()` and `dispatchBrowserEvent()` from a removing action are silently ignored (`LiveComponentSubscriber.php:364-368`)?

Also, my super-friend Claud found an issue with `LiveResponder` state that could leaks across worker requests, since `LiveResponder::reset()` can be never called, see https://claude.ai/code/artifact/0dc4c535-0770-4edc-9ddb-5009bddf9332?via=auto_preview

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
smnandre: Do you have a feature in mind ? A usage that would require it?

src/LiveComponent/src/Controller/BatchActionController.php
smnandre: Why ?

src/LiveComponent/assets/src/Component/index.ts
smnandre: Ho yes, especially as I stated it earlier 😮‍💨   thx

```ts
it('only looks at the presence of the header, not its value', () => {
    expect(makeResponse({ 'X-Live-Remove': '0' }).isRemoved()).toBe(true);
});
```

src/LiveComponent/src/EventListener/LiveComponentSubscriber.php
Kocal: For now, nothing, it's just that here we don't trigger the pre-re-render hooks anymore

src/LiveComponent/src/Controller/BatchActionController.php
Kocal: Ah, no nvm

smnandre:
>  super-friend Claud found an issue with LiveResponder state that could leaks across worker requests, since LiveResponder::reset() can be never called, see https://claude.ai/code/artifact/0dc4c535-0770-4edc-9ddb-5009bddf9332?via=auto_preview 

As its an existing problem, do you think you could ask them to fix it ? 😅 


I'll do it on a following PR elsewhen don't worry

smnandre:
> Let's document that `emit()` and `dispatchBrowserEvent()` from a removing action are silently ignored (`LiveComponentSubscriber.php:364-368`)?
> 
> Also, my super-friend Claud found an issue with `LiveResponder` state that could leaks across worker requests, since `LiveResponder::reset()` can be never called, see https://claude.ai/code/artifact/0dc4c535-0770-4edc-9ddb-5009bddf9332?via=auto_preview

I've found a way to dispatch them before the removal (that cannot be stopped from front-end) !

Kocal (approved):
To merge after #3761 and a small rebase + rebuild assets  🙏🏻

smnandre:
> To merge after #3761 and a small rebase + rebuild assets 🙏🏻

PR code & docs updated with the new method (dispatch events _then_ remove element)

I you want to have a quick check @Kocal

Kocal (approved):

PR #3781, opened 2026-08-15
[StimulusBundle] Speed up Stimulus name normalization

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

`normalizeKeyName()` ran three `preg_*` calls for every value, class and param name, and `normalizeControllerName()` ran a `preg_replace()` for every controller name. Both run on every `stimulus_controller()` / `stimulus_action()` call, on names that repeat across the whole page.

Memoize the key names (a fixed set coming from templates and controllers) and drop the regex from the controller name, which only strips a leading `@`.

Rendering 50k controllers with values, classes, an action and a target goes from ~414 ms to ~356 ms. On a more realistic page though, an admin CRUD page with 100 Stimulus controllers, that's 0.45 ms -> 0.42 ms per page: a 7% cut on a budget that's already under a millisecond. smnandre asked in review whether the project should have a minimum gain worth accepting, given the maintenance and history cost of a change like this. That's a fair question: this is a small, low-risk win, not a big one, and reviewers should weigh it as such rather than as more than it is.

```php
<?php
require __DIR__.'/src/StimulusBundle/vendor/autoload.php';

use Symfony\UX\StimulusBundle\Dto\StimulusAttributes;
use Twig\Environment;
use Twig\Loader\ArrayLoader;

$env = new Environment(new ArrayLoader([]));

for ($i = 0; $i < 5000; ++$i) {
    $attributes = new StimulusAttributes($env);
    $attributes->addController('my-controller', ['name' => 'ryan', 'isEnabled' => true, 'count' => 42], ['loading' => 'spinner']);
    $attributes->addAction('my-controller', 'onClick', 'click');
    $attributes->addTarget('my-controller', 'element');
    (string) $attributes;
}
```

Blackfire:

- before (833ms wall / 816ms CPU): https://app.blackfire.io/envs/5f4f9a62-eaa0-45ee-b7b3-a1b879f550e9/profiles/4d6d06bc-f658-4786-952a-302cc8c852ec/graph
- after (632ms wall / 628ms CPU): https://app.blackfire.io/envs/5f4f9a62-eaa0-45ee-b7b3-a1b879f550e9/profiles/c56f1954-74f2-40ad-9514-eab58bcb8912/graph
- diff: https://app.blackfire.io/envs/5f4f9a62-eaa0-45ee-b7b3-a1b879f550e9/profiles/compare/4d6d06bc-f658-4786-952a-302cc8c852ec...c56f1954-74f2-40ad-9514-eab58bcb8912/graph

### DIFF
--- a/src/StimulusBundle/src/Dto/StimulusAttributes.php
+++ b/src/StimulusBundle/src/Dto/StimulusAttributes.php
@@ -24,6 +24,14 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
 {
     private array $attributes = [];
 
+    /**
+     * Value/class/param names come from a fixed set defined in templates and
+     * controllers, so normalizing each one once is enough.
+     *
+     * @var array<string, string>
+     */
+    private static array $normalizedKeyNames = [];
+
     private array $controllers = [];
     private array $actions = [];
     private array $targets = [];
@@ -215,7 +223,9 @@ class StimulusAttributes implements \Stringable, \IteratorAggregate
      */
     private function normalizeControllerName(string $controllerName): string
     {
-        return preg_replace('/^@/', '', str_replace('_', '-', str_replace('/', '--', $controllerName)));
+        $normalized = str_replace(['/', '_'], ['--', '-'], $controllerName);
+
+        return str_starts_with($normalized, '@') ? substr($normalized, 1) : $normalized;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


smnandre (approved):
Should we define a minimum gap to accept improvements ? 

There is a hidden cost in maintenance, history, etc. 

Genuine question, I have no opinion here :)

Kocal:
> Should we define a minimum gap to accept improvements ?
> 
> There is a hidden cost in maintenance, history, etc.
> 
> Genuine question, I have no opinion here :)

Before AI, I'd say we had to be careful with performance patches when it came to code readability, but now I think it's fine... The code is still easy to read IMHO

PR #3786, opened 2026-08-15
[LiveComponent] Skip the bracket pipeline for plain model names

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

`normalizeModelName()` ran a `replace()`, a `split()`, a `map()` with a second `replace()` per part and a `join()` on every call, even for a plain name like `query` where the whole pipeline is a no-op.

It runs on every input event through `getValueFromElement()`, and twice per `ValueStore.set()` since `set()` normalizes and then calls `get()`, which normalizes again.

Return the name untouched when it contains neither `[` nor `]`. Names that do use the bracket syntax take the same path as before.

Measured over 400k calls, alternating both implementations to cancel any warm-up ordering effect:

    live form, every name bracketed   71.0 ms -> 72.1 ms   (-2%)
    half plain, half bracketed        51.0 ms -> 37.8 ms   (+35%)
    plain props only                  36.2 ms ->  6.0 ms   (+507%)

So it costs about 2% on a page that's nothing but live form fields, and pays off as soon as any plain model name is in the mix. In absolute terms this is nanoseconds per keystroke either way; the change is three lines and output is identical.

Browser-side JavaScript, so there is no Blackfire profile for this one. Benchmarked on node 22 with `node bench.mjs`, comparing both implementations side by side:

```js
function current(model) {
    return model.replace(/\[]$/, '').split('[').map((s) => s.replace(']', '')).join('.');
}

function candidate(model) {
    if (!model.includes('[') && !model.includes(']')) {
        return model;
    }

    return model.replace(/\[]$/, '').split('[').map((s) => s.replace(']', '')).join('.');
}

const plain = ['firstName', 'email', 'query', 'isEnabled', 'selectedId'];
const bracketed = ['user[firstName]', 'user[mailing][]', 'form[items][0][label]'];

for (const [label, names] of [
    ['plain only', plain],
    ['bracketed only', bracketed],
    ['mixed 80/20', [...plain, ...plain, ...plain, ...plain, ...bracketed]],
]) {
    for (const [impl, fn] of [['current', current], ['candidate', candidate]]) {
        for (const n of names) fn(n); // warm up

        const start = process.hrtime.bigint();
        for (let i = 0; i < 400000; i++) {
            fn(names[i % names.length]);
        }
        console.log(`${label} ${impl} -> ${(Number(process.hrtime.bigint() - start) / 1e6).toFixed(2)} ms`);
    }
}
```

A stray `]` with no `[` was the one input where the fast path could have diverged, so it is now covered by a unit test.

### DIFF
--- a/src/LiveComponent/assets/src/string_utils.ts
+++ b/src/LiveComponent/assets/src/string_utils.ts
@@ -31,6 +31,12 @@ export function trimAll(str: string) {
  * For example: "user[firstName]" becomes "user.firstName"
  */
 export function normalizeModelName(model: string): string {
+    // Some model names are plain (e.g. "query"), and the pipeline below is a
+    // no-op for them. ValueStore also normalizes already-normalized names.
+    if (!model.includes('[') && !model.includes(']')) {
+        return model;
+    }
+
     return (
         model
             // Names ending in "[]" represent arrays in HTML.
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -262,6 +262,7 @@ function trimAll(str) {
 	return str.replace(/[\s]+/g, " ").trim();
 }
 function normalizeModelName(model) {
+	if (!model.includes("[") && !model.includes("]")) return model;
 	return model.replace(/\[]$/, "").split("[").map((s) => s.replace("]", "")).join(".");
 }
 function getValueFromElement(element, valueStore) {
--- a/src/LiveComponent/assets/test/unit/string_utils.test.ts
+++ b/src/LiveComponent/assets/test/unit/string_utils.test.ts
@@ -30,4 +30,8 @@ describe('normalizeModelName', () => {
     it('can normalize a string ending in []', () => {
         expect(normalizeModelName('user[mailing][]')).toEqual('user.mailing');
     });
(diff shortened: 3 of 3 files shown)


src/LiveComponent/assets/src/string_utils.ts
```diff
@@ -31,6 +31,12 @@ export function trimAll(str: string) {
  * For example: "user[firstName]" becomes "user.firstName"
  */
 export function normalizeModelName(model: string): string {
+    // Most model names are plain (e.g. "query"), and the pipeline below is a
+    // no-op for them. ValueStore also normalizes already-normalized names.
```
smnandre:
```suggestion
    // Some model names are plain (e.g. "query")
```

Because with LiveComponent and live form, most model names _are_ prefixed by their form name prefix, like that: `myform[foo]`.

smnandre (approved):
Let's not add misleading comments

src/LiveComponent/assets/src/string_utils.ts
Kocal: Yep you're right, fixed!

PR #3791, opened 2026-08-18
[LiveComponent] Fix proxy turning "toJSON" and "then" protocol probes into server actions

### LINKED ISSUE #3752: [LiveComponent] Component proxy turns "toJSON"/"then" property probes into real server actions
### Bug description

`proxifyComponent()`'s `get` trap treats **any unknown string property read** as an action lookup and returns a callable that queues a real server action (`src/LiveComponent/assets/src/Component/index.ts`):

```ts
// try to call an action
return (args: string[]) => {
    return component.action.apply(component, [prop, args]);
};
```

Two well-known protocol probes make this fire *implicitly*, without anyone ever calling an action:

- `JSON.stringify(proxy)` — the JSON protocol looks up `toJSON`; the trap returns a function, so `stringify` **invokes it**, queuing a server action literally named `toJSON`.
- `await proxy` / `Promise.resolve(proxy)` — promise assimilation looks up `then`; same story with an action named `then`.

The queued action POSTs to `/_components/{name}/_batch` (serializing one live event detail reaches the proxy twice — `detail.component` and `detail.controller.proxiedComponent` — so two `toJSON` actions batch together), and the backend 500s:

```
InvalidArgumentException: The controller for URI "/_components/BookmarkWithCollection/_batch" is not callable:
Expected method "toJSON" on class "App\Twig\Components\BookmarkWithCollection".
```

Because the proxy is handed to arbitrary page JS via every `live:connect` / `live:disconnect` event (`detail.component`) and via `getComponent()`, any code that deep-serializes objects reaching those details — error trackers, analytics bridges, browser extensions doing main-world page serialization — triggers checksum-valid backend requests and server 500s that the app owner cannot prevent. We observed bursts of one 500 per live component per page load, each request body containing `"actions":[{"name":"toJSON"},{"name":"toJSON"}]`.

### How to reproduce

On any page with a live component:

```js
document.addEventListener('live:connect', (event) => {
    JSON.stringify(event.detail.component);
    // → POST /_components/<name>/_batch with action "toJSON" → 500
});
```

Likewise, `await getComponent(el)` queues an action named `then`.

Verified on symfony/ux-live-component v3.3.0 and v3.4.0 (dist identical); the trap in the current default branch is unchanged.

### Suggested fix

Don't fabricate actions for well-known protocol probe properties in the `get` trap, before the action fallback (symbols are already excluded):

```ts
if (prop === 'toJSON' || prop === 'then') {
    return undefined;
}
```

`JSON.stringify` then serializes the component normally and promise assimilation treats the proxy as a non-thenable.

App-side workaround until then:

```js
import { Component } from '@symfony/ux-live-component';

Object.defineProperty(Component.prototype, 'toJSON', {
    configurable: true,
    value() { return '[LiveComponent]'; },
});
Object.defineProperty(Component.prototype, 'then', {
    configurable: true,
    value: undefined,
});
```

Once the props are `in` the component, the proxy's first branch serves them locally and nothing hits the network.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #3752
| License       | MIT

The component proxy treats any unknown property read as a server action. `JSON.stringify()` probes `toJSON`, and promise assimilation (e.g. `await component`, or returning the proxy from an async function) probes `then`: both received a callable and invoked it, queueing a phantom server action named `toJSON` or `then`.

This PR returns `undefined` for these two protocol probes. The guard sits after the model lookup, so a model actually named `toJSON` or `then` keeps priority; only the "unknown name = action" fallback is skipped. Same family of fix as #2537, which stopped `JSON.stringify()` from serializing `__component`.

Tests cover both probes (no action queued, `await` resolves to the proxy itself) and the model-priority case.

### DIFF
--- a/src/LiveComponent/assets/dist/live_controller.js
+++ b/src/LiveComponent/assets/dist/live_controller.js
@@ -1663,6 +1663,7 @@ function proxifyComponent(component) {
 				return Reflect.get(component, prop);
 			}
 			if (component.valueStore.has(prop)) return component.getData(prop);
+			if ("toJSON" === prop || "then" === prop) return;
 			return (args) => {
 				return component.action.apply(component, [prop, args]);
 			};
--- a/src/LiveComponent/assets/src/Component/index.ts
+++ b/src/LiveComponent/assets/src/Component/index.ts
@@ -588,6 +588,14 @@ export function proxifyComponent(component: Component): Component {
                 return component.getData(prop);
             }
 
+            // protocol probes performed implicitly by JSON.stringify() ("toJSON") and
+            // by promise assimilation ("then", e.g. `await component`) must not be
+            // mistaken for actions: returning a callable would let those protocols
+            // invoke it, queueing a real server action of that name
+            if ('toJSON' === prop || 'then' === prop) {
+                return undefined;
+            }
+
             // try to call an action
             return (args: string[]) => {
                 return component.action.apply(component, [prop, args]);
--- a/src/LiveComponent/assets/test/unit/Component/index.test.ts
+++ b/src/LiveComponent/assets/test/unit/Component/index.test.ts
@@ -130,5 +130,54 @@ describe('Component class', () => {
             expect(backend.actions[0].name).toBe('save');
             expect(backend.actions[0].args).toEqual({ foo: 'bar', secondArg: 'secondValue' });
         });
+
+        it('does not turn the toJSON protocol probe of JSON.stringify() into an action', async () => {
+            const { proxy, backend } = makeDummyComponent();
+
+            // @ts-expect-error
+            expect(proxy.toJSON).toBeUndefined();
+
+            try {
+                JSON.stringify(proxy);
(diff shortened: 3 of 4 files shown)


Kocal:
Please target 3.x branch

Amoifr:
Done! Rebased onto 3.x, dist rebuilt from the 3.x sources, JS suite green (297 tests), base switched.

Kocal:
Dist files must be rebuild

Amoifr:
Rebuilt the dist with the current toolchain (Node 22 + pnpm, same as the `dist-files-unbuilt` job) and pushed. Sorry for the noise!

Kocal:
I just added a small note in the documentation, merging when it's green

Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3792, opened 2026-08-18
[LiveComponent] Fix disabled choices being submitted by live form updates

### LINKED ISSUE #2934: [LiveComponent] Disabled option in a multi-value checkbox handled incorrectly
Hi, I have a form with a `ChoiceType` field like this:

```php
->add('my_choices', ChoiceType::class, [
    'multiple' => true,
    'expanded' => true,
    'choices' => [
        'foo' => 'foo',
        'bar' => 'bar',
        'baz' => 'baz',
    ],
    'choice_attr' => fn($choice) => 'foo' === $choice ? ['disabled' => true] : [],
    'constraints' => [
        new Choice(['bar', 'baz'], multiple: true),
    ],
])
```
as you can see, the `foo` choice is disabled, so it can not be selected and this is validated on the backend.

I use this form in a live component something like this:

```php
#[AsLiveComponent('my_form')]
final class MyForm
{
    use DefaultActionTrait;
    use ComponentWithFormTrait;

    {
        return $this->createForm(MyType::class, ['my_choices' => ['foo']]);
    }
}

```
the `foo` option is selected by default (but it will be disabled when the form is rendered). If I update the form and select an another choice from the list, a live update is triggered and the form is validated, but it gives back that the foo option is invalid, but then I post the form regularly (click the submit button) the validation passes because the browser does not send disabled checkbox values to the backend, but live components kind of does. We should somehow know that the `foo` choice is disabled and ignore when updating, but keep it's original selected state.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2934
| License       | MIT

Browsers never submit disabled controls, but `extractFormValues()` (whose contract is precisely "the raw POST data that would be sent if the form were submitted") included them. For an expanded `ChoiceType` with a choice disabled through `choice_attr`, the aggregated parent value kept the disabled choice, so every live update re-submitted it: validation failed on live updates while a native submit of the very same form passed (the exact asymmetry reported).

The fix makes `extractFormValues()` apply browser parity: it drops the values of disabled choices (expanded multiple, expanded single) and treats a checked but disabled standalone checkbox as unsubmitted, whether the field is disabled at the form level (the `disabled` option) or only in HTML (a `disabled` attribute, e.g. set through `choice_attr`).

No JS change: `SetValueOntoModelFieldsPlugin` only syncs elements carrying `data-model` themselves (Symfony forms only put it on `<form>`), and the two `morphdom` syncs only run for modified or focused elements, which a disabled control cannot be.

Covered by a unit test on the extracted values, an end-to-end functional test (initially checked disabled choice, then a live update: the disabled value never reaches the submitted data), and a JS test pinning the server-side sync of a checkbox the server disables and clears (unchecked + disabled).

Note: disabled `<option>` elements of a non-expanded `<select>` have the same theoretical issue through a different code path; left out to keep this focused, happy to follow up if wanted.

### DIFF
--- a/src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
+++ b/src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
@@ -192,10 +192,13 @@ class ComponentWithFormTest extends KernelTestCase
             'choice_required_with_empty_preferred_choices' => 'ok',
             'choice_expanded' => '',
             'choice_multiple' => ['2'],
+            'choice_multiple_disabled' => [],
+            'choice_expanded_disabled' => '',
             'select_multiple' => ['2'],
             'entity' => (string) $id,
             'checkbox' => null,
             'checkbox_checked' => '1',
+            'checkbox_checked_disabled' => null,
             'file' => '',
             'hidden' => '',
             'complexType' => [
@@ -292,6 +295,48 @@ class ComponentWithFormTest extends KernelTestCase
             ->assertContains('<option value="2" selected="selected">')
             ->assertContains('<option value="1" selected="selected">')
         ;
+    }
+
+    public function testDisabledChoicesAreNeverSubmitted()
+    {
+        CategoryFixtureEntityFactory::createMany(5);
+
+        $mounted = $this->mountComponent(
+            'form_with_many_different_fields_type',
+            [
+                'initialData' => [
+                    // "foo" (value 1) is disabled through "choice_attr"
+                    'choice_multiple_disabled' => [1, 2],
+                ],
+            ]
+        );
+
+        $dehydratedProps = $this->dehydrateComponent($mounted)->getProps();
+
+        // like a browser, the checked but disabled choice is not part of the
+        // values that would be submitted
+        $this->assertSame(['2'], $dehydratedProps['form']['choice_multiple_disabled']);
+
+        // a model update must not resurrect the disabled value either
+        $crawler = $this->browser()
+            ->throwExceptions()
+            ->post('/_components/form_with_many_different_fields_type', [
+                'body' => [
+                    'data' => json_encode([
+                        'props' => $dehydratedProps,
+                        'updated' => ['form' => ['choice_multiple_disabled' => []]],
+                    ]),
+                ],
+            ])
+            ->assertSuccessful()
+            ->crawler()
+        ;
+
+        $dehydratedProps = json_decode(
+            $crawler->filter('div')->first()->attr('data-live-props-value'),
+            true
+        );
+        $this->assertSame([], $dehydratedProps['form']['choice_multiple_disabled']);
     }
 
     public function testLiveCollectionTypeAddButtonsByDefault()
--- a/src/LiveComponent/assets/test/unit/controller/model.test.ts
+++ b/src/LiveComponent/assets/test/unit/controller/model.test.ts
@@ -346,6 +346,35 @@ describe('LiveController data-model Tests', () => {
         expect(test.component.valueStore.getOriginalProps()).toEqual({ form: { check1: null, check2: '1' } });
     });
 
+    it('syncs a checkbox the server disabled and cleared: unchecked and disabled', async () => {
+        const test = await createTest(
+            { check1: '1', isDisabled: false },
+            (data: any) => `
+            <div ${initComponent(data)}>
+                <label>
+                    Checkbox 1: <input type="checkbox" data-model="check1" value="1" ${data.check1 ? 'checked' : ''} ${data.isDisabled ? 'disabled' : ''} />
+                </label>
+            </div>
+        `
+        );
+
+        const check1Element = getByLabelText(test.element, 'Checkbox 1:') as HTMLInputElement;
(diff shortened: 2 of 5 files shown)


Kocal:
Please target 3.x branch

Amoifr:
Done! Rebased onto 3.x, dist rebuilt from the 3.x sources, PHP suite green (391 tests) and JS tests too, base switched.

smnandre (commented):
Guard Breaks Backend State Sync ?

When the backend disables a previously-checked checkbox and sets its data to null/empty, `extractFormValues()` correctly returns null (line 272).

But `setValueOnElement()` guard prevents applying this null value because the element is disabled. 

Result: DOM shows checked + disabled (contradictory/broken UI), when it should show unchecked + disabled. 

Maybe the guard should NOT apply to `SetValueOntoModelFieldsPlugin` syncs, but only during morphdom to preserve user interactions during DOM updates. 

wdyt ?

src/LiveComponent/tests/Functional/Form/ComponentWithFormTest.php
```diff
@@ -456,7 +501,7 @@ public function testDataModelAttributeAutomaticallyAdded()
         ;
     }
 
-    public function testFormWithLivePropContainingAnEntityImplementingAnInterface()
+    public function testFormWithLivePropContainingAnEntityImplementingAnInterface(): void
```
smnandre: Revert (and the other ones) please :)

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -624,9 +626,12 @@ var Idiomorph = (function() {
 			} else if (ctx.head.shouldRemove(currentHeadElt) !== false) removed.push(currentHeadElt);
 		}
 		nodesToAppend.push(...srcToNewHeadNodes.values());
+		log("to append: ", nodesToAppend);
 		let promises = [];
 		for (const newNode of nodesToAppend) {
+			log("adding: ", newNode);
```
smnandre:
```suggestion
```

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -624,9 +626,12 @@ var Idiomorph = (function() {
 			} else if (ctx.head.shouldRemove(currentHeadElt) !== false) removed.push(currentHeadElt);
 		}
 		nodesToAppend.push(...srcToNewHeadNodes.values());
+		log("to append: ", nodesToAppend);
```
smnandre:
```suggestion
```

src/LiveComponent/assets/dist/live_controller.js
```diff
@@ -624,9 +626,12 @@ var Idiomorph = (function() {
...
 		let promises = [];
 		for (const newNode of nodesToAppend) {
+			log("adding: ", newNode);
 			let newElt = document.createRange().createContextualFragment(newNode.outerHTML).firstChild;
+			log(newElt);
```
smnandre:
```suggestion
```

Amoifr:
Thanks for the careful review! Applied: return types dropped from the test methods (Fabbot agrees with you), dist rebuilt on top of `3.x` so it only carries the one-line change, and the guard is now a single check for radios and checkboxes.

On the backend sync question, I think the scenario works as expected, because the guard never competes with the server: when the backend disables a previously checked checkbox and clears its data, the re-rendered HTML no longer carries `checked`, and idiomorph syncs the `checked` property during the morph, so the DOM shows unchecked and disabled before `SetValueOntoModelFieldsPlugin` runs. That plugin is the only place the guard actually kicks in, and it is exactly where the bug was: Symfony keeps the submitted data of a disabled field, so the server renders the disabled choice checked, while the value store (built from what a browser would submit, which never includes disabled controls) does not contain its value and would uncheck it. The two `setValueOnElement()` calls in `morphdom.ts` only run for elements the user modified or focused, which a disabled control cannot be. So the server-rendered state stays the source of truth for disabled controls in every path, which is the intent of the guard. Happy to add a test for your scenario if you want it pinned.

smnandre:
@Amoifr to be honest I'd be happy with one :)

Amoifr:
Here it is, and writing it was instructive: to make the test meaningful the checkbox has to carry `data-model` itself, because `SetValueOntoModelFieldsPlugin` only syncs `[data-model]` elements and Symfony forms only put `data-model` on the `<form>`. So for the #2934 case the plugin never touches the inputs, and the two `setValueOnElement()` calls in `morphdom.ts` only run for modified or focused elements, which a disabled control cannot be. In other words the guard was dead code for this bug and only covered a contrived markup, so I removed it altogether: the PR is now PHP-only (dist identical to `3.x`), plus the test for your scenario (server disables and clears: unchecked + disabled, synced by idiomorph). Rebased on current `3.x` as well.

smnandre (approved):
A great PR, that will remove/handle a lot of small bugs / weird situations we had no idea about (and that were happening, not a single doubt about that) 

Thank you very much  @Amoifr

PR #3793, opened 2026-08-18
[LiveComponent] Add conditional expressions to #[LiveListener] event names

casskir:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| Issues         | none — no existing issue, explained below
| License        | MIT

Today, a **#[LiveListener]** matches purely by event name: as soon as any component on the page emits an event, every component listening for that name fires an Ajax call and re-renders — even when the event has nothing to do with that particular component instance.                                                                                                                                                  
                                                                                                                                                                                                              
  A common example: a page renders a list of "product card" components, each listening to productUpdated. When one product changes, the event is emitted once, but every card on the page reacts, each making its own Ajax round-trip and re-render, even though only one of them is actually concerned.         

Filtering this out today means writing the check by hand inside the listener method and returning early — but by then the Ajax call has already been made and the server has already done the (wasted) work of building a response.                                                                                                                                                                                     


**Solution**        

The event name passed to #[LiveListener] can now be followed by a condition in parentheses, written as a small expression:

```php                                                                                                                                                                                              
use Symfony\UX\LiveComponent\Attribute\LiveListener;

#[LiveProp]
public int $product;

#[LiveListener('productUpdated(event.id == props.product)')]
public function refreshProduct() 
{
    // only called when the emitted event's "id" matches this component's "product" prop                                                                                                                    
}
```

The condition is evaluated entirely on the client, before any Ajax call is made: 

- event — the data emitted along with the event ($this->emit('productUpdated', ['id' => $product->getId()])).
- props — the current props of this component (including local changes not yet confirmed by the server).

If the condition doesn't pass, the listener is skipped and no request is sent to the server at all — the server-side render cycle is completely untouched by this feature. If the condition itself is invalid (e.g. a typo), it's treated as false (not "always true") and an error is logged to the browser console, so a bad expression can't accidentally make a listener fire on everything.

Conditions are written using Jexl syntax (comparisons, &&/||, nested property access, function calls...). The library is bundled directly into dist/live_controller.js, exactly like idiomorph already is - no extra install step or importmap entry is required from users of the package.

Plain #[LiveListener('eventName')] usages are completely unaffected — the condition is optional and fully backward compatible.

**Changes:**
  - Symfony\UX\LiveComponent\Attribute\LiveListener: parses eventName(condition), exposes getCondition(): ?string. 
  - AsLiveComponent::liveListeners(): now also returns the condition for each listener. 
  - debug:live-component: displays the condition (event if (condition) => action) when one is set.
  - Client (assets/src): Component/ValueStore evaluate the condition against { event, props } via jexl before queuing the action; ValueStore::getCurrentProps() was added to expose the effective (server + local dirty) props for that evaluation.                                                                                                                                                                     
  - bin/build_package.ts (repo root, shared build tooling): added jexl and its transitive @babel/runtime helpers to deps.onlyBundle, next to idiomorph, so the library gets inlined into the single-file bundle instead of left as an unresolved import. 

**Tests:**
- tests/Unit/Attribute/LiveListenerTest.php — parsing of the event-name/condition syntax (plain name, condition, nested parentheses, whitespace, empty parentheses, invalid syntax). 
- tests/Unit/Attribute/AsLiveComponentTest.php — updated/extended for the new condition key.
- assets/test/unit/controller/listener-condition.test.ts — condition passing/failing, evaluated against locally-changed props, malformed condition never firing.
- assets/test/unit/ValueStore.test.ts — new getCurrentProps() coverage. 
- dist/ was rebuilt and smoke-tested (imported directly under jsdom, confirmed the compiled bundle correctly skips/fires the action based on the condition).


bin/build_package.ts
```diff
@@ -87,7 +87,7 @@ async function main() {
         unbundle: isStimulusBundle || isReactOrVue,
         deps: {
             neverBundle: Array.from(external),
-            onlyBundle: ['idiomorph'],
+            onlyBundle: ['idiomorph', 'jexl', '@babel/runtime'],
```
Kocal: Polluting our dist files with Babel runtime helpers is a **no-go** in 2026 where all browsers supports ESM modules, ES6 features, etc... 

Please use a more modern and ESM-first library instead.

Thanks 🙏

Kocal (changes requested):

smnandre:
Thanks for the PR @casskir!

I think I understand the problem you're trying to solve: we can currently target a component name, a parent, ourselves, etc., but not a particular instance, or subset of instances, based on its current state/props.

https://symfony.com/bundles/ux-live-component/current/index.html#emitting-only-to-components-with-a-specific-name

I can definitely see the limitation here. If a page contains many instances of the same component listening to the same event, we currently don't have a way to prevent unrelated instances from reacting before the request is made.

I'm less convinced about solving this by introducing conditional expressions directly into `#[LiveListener]`, though.

I'm also not sure we want LiveComponent events to become a way to broadcast application/domain events and then let each component decide, based on its current props, whether it should react.

To me, LiveComponent events are primarily about UI/component communication. If a specific product changed on the server, I would rather have the response explicitly update or refresh the relevant UI, for example through a Turbo Stream, than broadcast a productUpdated event to every mounted component.

This is also why I'm a bit uncomfortable with `event.id == props.product`: the condition is declared in PHP, but evaluated against client-side state. It starts blurring the boundary between server state and UI state in a way I'm not sure we want to encourage.

So I can see the performance issue described by the PR, but I'm not yet convinced that conditional listeners are the right abstraction to solve it.

smnandre (commented):
cf ma previous message

bin/build_package.ts
casskir: Swapped jexl for @hypatiatech/jexl ? an actively maintained, TypeScript-first, ESM-first.

casskir:
> Thanks for the PR @casskir!
> 
> I think I understand the problem you're trying to solve: we can currently target a component name, a parent, ourselves, etc., but not a particular instance, or subset of instances, based on its current state/props.
> 
> https://symfony.com/bundles/ux-live-component/current/index.html#emitting-only-to-components-with-a-specific-name
> 
> I can definitely see the limitation here. If a page contains many instances of the same component listening to the same event, we currently don't have a way to prevent unrelated instances from reacting before the request is made.
> 
> I'm less convinced about solving this by introducing conditional expressions directly into `#[LiveListener]`, though.
> 
> I'm also not sure we want LiveComponent events to become a way to broadcast application/domain events and then let each component decide, based on its current props, whether it should react.
> 
> To me, LiveComponent events are primarily about UI/component communication. If a specific product changed on the server, I would rather have the response explicitly update or refresh the relevant UI, for example through a Turbo Stream, than broadcast a productUpdated event to every mounted component.
> 
> This is also why I'm a bit uncomfortable with `event.id == props.product`: the condition is declared in PHP, but evaluated against client-side state. It starts blurring the boundary between server state and UI state in a way I'm not sure we want to encourage.
> 
> So I can see the performance issue described by the PR, but I'm not yet convinced that conditional listeners are the right abstraction to solve it.

Thanks for the detailed feedback!

I don't think this solution moves away from the idea of UI/component communication — it just lets each component decide for itself how to react to an event, regardless of whether that event is tied to a specific instance or to particular data. That decision (react or ignore) is arguably still a UI-level concern, even if the condition happens to be based on props coming from the server.

A concrete example: imagine a table listing requests/tickets with their statuses. In an outer component we change the status of a single request, but as a result we're forced to re-render every row individually (or the whole table at once). This can easily produce 10+ requests to the server whose response we don't actually care about, since only one row's status changed. And even if the row component itself needs to react somehow (e.g. highlight the updated record), the server-side code we'd write to filter out irrelevant instances is exactly the same logic — we're just moving that same condition to the client instead of duplicating it in PHP.

I agree a Turbo Stream-based solution is possible, but it requires extra code for every such pattern: either re-rendering and swapping the whole component in the DOM, or replacing specific inner elements, or some other bespoke wiring. In practice this means writing custom glue code for what is a fairly simple problem, which increases the codebase, the chance of bugs, and the maintenance cost. With conditional listeners, this works "out of the box," since the component's standard re-render mechanism already solves it.

So while I understand the concern about blurring server/client boundaries, I'd frame it differently: the condition isn't really evaluating server state on the client — it's the component deciding, based on its own current props, whether an already-broadcast UI event is relevant to it. That's a pattern that's already implicit in x-target/name-based listeners, just applied at a finer granularity.

smnandre:
> an already-broadcast UI event 

But as soon you're involving Symfony or Twig it's not an "broadcast UI event" anymore, but a server-side transmitted one. 

---

> a table listing requests/tickets with their statuses. In an outer component we change the status of a single reques

Asking again because I need an example to figure things out: 

Is this like some sass-type UI we're talking about, and you want to do something on the list part because you just changed it on the right part (details/form) ? Or is there a scenario i'm still not getting ? 

If this is the case, you have a live component per row that right ? 
And they do change (outside this side form) live ? 
They have a state in the page ? Actions ? Data ?

Just asking because any new type of behaviour must be very explicit on the why / when / how, as we will need to document, support it, maintain it, etc etc.. and I want to be sure it's aligned with the overall scope of Live Components.

casskir:
> > an already-broadcast UI event
> 
> But as soon you're involving Symfony or Twig it's not an "broadcast UI event" anymore, but a server-side transmitted one.
> 
> > a table listing requests/tickets with their statuses. In an outer component we change the status of a single reques
> 
> Asking again because I need an example to figure things out:
> 
> Is this like some sass-type UI we're talking about, and you want to do something on the list part because you just changed it on the right part (details/form) ? Or is there a scenario i'm still not getting ?
> 
> If this is the case, you have a live component per row that right ? And they do change (outside this side form) live ? They have a state in the page ? Actions ? Data ?
> 
> Just asking because any new type of behaviour must be very explicit on the why / when / how, as we will need to document, support it, maintain it, etc etc.. and I want to be sure it's aligned with the overall scope of Live Components.

Yeah, pretty much a SaaS-type UI, that's a good way to put it. Think of a dashboard with a table of requests/tickets, and each row is its own LiveComponent instance (so yes, one component per row). On the side (or in a modal), there's a details/edit form for whichever request you're currently working on.

When you change the status in that form and submit, the outer form component emits an event like requestUpdated with the request id and new status in the payload. Right now, every row listening for requestUpdated fires, because they all match on event name alone — even though only one row actually corresponds to that id. So you get one relevant re-render and N-1 wasted ones.

To answer the specifics:

State: each row keeps its own id and status as props, nothing shared beyond that.
Actions: the row's listener just re-renders itself (e.g. to reflect the new status/badge), nothing more exotic.
Data: the event carries { id, status } — that's it, no extra domain logic riding along.
They don't change "live" outside the form on their own — the form is the only thing driving the update. The row components are just reacting to something that already happened elsewhere on the page.

So it's less "let's turn LiveComponent events into a pub/sub for domain events" and more "let the row decide for itself whether this UI event that already happened is relevant to it, instead of blindly reacting and letting the server sort it out." Hope that's a clearer picture of what I had in mind — happy to sketch a minimal reproducible example if that'd help move the discussion forward.

smnandre:
> "let the row decide for itself whether this UI event that already happened is relevant to it, instead of blindly reacting and letting the server sort it out."

Ok i can jump on board with that. This also could be seen (implementation wise) as a filter on that event ? 

https://stimulus.hotwired.dev/reference/actions#options

casskir:
> > "let the row decide for itself whether this UI event that already happened is relevant to it, instead of blindly reacting and letting the server sort it out."
> 
> Ok i can jump on board with that. This also could be seen (implementation wise) as a filter on that event ?
> 
> https://stimulus.hotwired.dev/reference/actions#options

Yeah, exactly — that's a good way to frame it, and honestly a cleaner one than "conditional listener." It's the same mental model as Stimulus's action filters: the event still fires and reaches the component, but the component decides upfront whether it actually cares, before doing anything about it.

The one difference vs. the Stimulus options you linked is that those filters mostly operate on the native event itself (key codes, modifiers, once/passive/self...), while here the filter needs to compare data carried by the event against the component's own props — so it's a bit closer to a predicate than a fixed set of named options. But conceptually it's the same idea: filter first, react only if it passes, no extra plumbing needed for the common case.

I'm totally open to leaning into that framing more explicitly — both in the docs/wording (calling it a "filter" rather than a "condition") and possibly in how constrained the syntax is.

smnandre:
Is there any world we could implement this "only" on the front end side?

PR #3469, opened 2026-04-14
[Toolkit][Shadcn] Add sidebar recipe

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | Part of #3233
| License        | MIT

Adds the `sidebar` recipe to the Shadcn kit (part of #3233).

- Multi-component recipe on the current Toolkit format (README-based examples, `tailwind_classes`, `provide()` / `inject()`): `Sidebar`, `Provider`, `Trigger`, `Rail`, `Inset`, `Header`, `Content`, `Footer`, `Separator`, `Input`, `Group`, `GroupLabel`, `GroupAction`, `GroupContent`, `Menu`, `MenuItem`, `MenuButton`, `MenuAction`, `MenuBadge`, `MenuSub`, `MenuSubItem`, `MenuSubButton`, `MenuSkeleton`.
- Collapse / expand (`offcanvas`, `icon`, `none`) driven by a Stimulus controller, with a cookie to persist the state and a keyboard shortcut, like the shadcn reference.
- `Sidebar:MenuButton` has an `as` prop so menu items can render as links (thanks @Lacoi for the ask).
- The README ships a "Using the sidebar as a page layout" section, since blocks cannot be declared inside `<twig:Sidebar:Inset>` (thanks @mlwymore for testing and reporting).
- The examples use the `dropdown-menu` recipe merged in #3232.

Originally this PR also shipped a `dropdown-menu` recipe, dropped in favor of #3232.

PR #3798, opened 2026-08-21
[Toolkit][Shadcn] Fix the hover-card usage example crashing on a plain HTML trigger

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | n/a
| License        | MIT

The "Usage" snippet of the Shadcn `hover-card` recipe puts the trigger attributes on a plain `<a>` with the `{{ ...hover_card_trigger_attrs }}` spread. That spread is a Twig Components feature and only works on `<twig:...>` tags; on a plain HTML element it reaches Twig as-is and fails at compile time with `Cannot use positional argument after argument unpacking` (a PHP fatal, not even a catchable Twig error).

The snippet is not caught by the rendering snapshots because only the `## Examples` blocks are rendered, not `## Usage`.

This switches it to `{{ html_attr(hover_card_trigger_attrs) }}`, which is the right way to render the attributes on a plain element. I verified the updated snippet renders `<a href="#" class="underline" data-slot="hover-card-trigger" tabindex="0" data-action="focus->hover-card#show blur->hover-card#hide">` through the kit context runner. Found while answering a user report on #3469, where the same mistake is an easy one to make.

### DIFF
--- a/src/Toolkit/kits/shadcn/hover-card/README.md
+++ b/src/Toolkit/kits/shadcn/hover-card/README.md
@@ -26,7 +26,7 @@ For sighted users to preview content available behind a link.
 ```twig
 <twig:HoverCard>
     <twig:HoverCard:Trigger>
-        <a href="#" class="underline" {{ ...hover_card_trigger_attrs }}>@symfony</a>
+        <a href="#" class="underline" {{ html_attr(hover_card_trigger_attrs) }}>@symfony</a>
     </twig:HoverCard:Trigger>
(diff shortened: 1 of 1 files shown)


Kocal:
Oh, nice catch

Kocal (approved):

Kocal:
Thank you @Amoifr.

PR #3799, opened 2026-08-21
[TwigComponent] Resolve typed attribute values for custom component renderers

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | no
| Issues         | reported in #3469
| License        | MIT

Since Twig Components 3.5, attribute merging relies on the typed values from `twig/html-extra` (`html_attr_type()`, `tailwind_classes`, ...). The Toolkit uses them for the `data-action` entries in its `*_trigger_attrs` hashes, so consumers can spread them on a component and still get their own `data-action` merged in.

Components rendered by a custom renderer (`<twig:ux:icon>`, `<twig:ux:map>`) don't go through a template: `ComponentRuntime` hands the raw props straight to the package runtime, which only deals with scalars (`Icon::withAttributes()`, `StimulusAttributes::addAttribute(string, string)`). Any typed value crashes with `Invalid value type for attribute "data-action". Boolean, string, int or float allowed, "Twig\Extra\Html\HtmlAttr\SeparatedTokenList" provided.`

A user hit this in #3469 with `<twig:ux:icon name="..." {{ ...dropdown_menu_trigger_attrs }} />`.

### What's in this PR

- `ComponentRuntime::render()` now resolves the props once, right before handing them to a custom renderer: an `AttributeValueInterface` value becomes its `getValue()`, a `null` value becomes `false` (the attribute gets dropped, same as `ComponentAttributes` does when rendering), and `\Stringable` values get cast to string. This resolution lives in a new `ComponentAttributes::resolved()` method, kept off the hot path (`all()`, `__toString()` and `render()` are untouched); the `component()` Twig function is declared with `needs_environment` to reach the escaper, no service wiring change involved.

- Icons, Map and any future custom renderer benefit from this without touching their own contracts: `IconRendererInterface`, `MapRendererInterface` and `ux_icon()` keep receiving plain scalars as before.

Covered by unit tests on `ComponentRuntime` and `ComponentAttributes`.

### DIFF
--- a/src/Icons/src/Twig/UXIconRuntime.php
+++ b/src/Icons/src/Twig/UXIconRuntime.php
@@ -31,17 +31,10 @@ final class UXIconRuntime implements RuntimeExtensionInterface
     }
 
     /**
-     * @param array<string, bool|string|\Stringable> $attributes
+     * @param array<string, bool|string> $attributes
      */
     public function renderIcon(string $name, array $attributes = []): string
     {
-        foreach ($attributes as $attribute => $value) {
-            // the renderer only deals with scalars, so that its decorators (e.g. caches) can rely on it
-            if ($value instanceof \Stringable) {
-                $attributes[$attribute] = (string) $value;
-            }
-        }
-
         try {
             return $this->iconRenderer->renderIcon($name, $attributes);
         } catch (IconNotFoundException $e) {
--- a/src/Icons/src/Twig/UXIconRuntime.php
+++ b/src/Icons/src/Twig/UXIconRuntime.php
@@ -15,6 +15,7 @@ use Psr\Log\LoggerInterface;
 use Symfony\UX\Icons\Exception\IconNotFoundException;
 use Symfony\UX\Icons\IconRendererInterface;
 use Twig\Extension\RuntimeExtensionInterface;
+use Twig\Extra\Html\HtmlAttr\AttributeValueInterface;
 
 /**
  * @author Simon André <smn.andre@gmail.com>
@@ -31,10 +32,23 @@ final class UXIconRuntime implements RuntimeExtensionInterface
     }
 
     /**
-     * @param array<string, bool|string> $attributes
+     * @param array<string, bool|string|AttributeValueInterface> $attributes
      */
     public function renderIcon(string $name, array $attributes = []): string
     {
+        foreach ($attributes as $attribute => $value) {
+            // Typed values produced by twig/html-extra (e.g. "html_attr_type('sst')"), as used by
+            // Twig Components attribute merging, are rendered to their string representation
+            if ($value instanceof AttributeValueInterface) {
+                if (null === $value = $value->getValue()) {
+                    unset($attributes[$attribute]);
+                    continue;
+                }
+
+                $attributes[$attribute] = $value;
+            }
+        }
+
         try {
             return $this->iconRenderer->renderIcon($name, $attributes);
         } catch (IconNotFoundException $e) {
--- a/src/Icons/src/Twig/UXIconRuntime.php
+++ b/src/Icons/src/Twig/UXIconRuntime.php
@@ -15,7 +15,6 @@ use Psr\Log\LoggerInterface;
 use Symfony\UX\Icons\Exception\IconNotFoundException;
 use Symfony\UX\Icons\IconRendererInterface;
 use Twig\Extension\RuntimeExtensionInterface;
-use Twig\Extra\Html\HtmlAttr\AttributeValueInterface;
 
 /**
  * @author Simon André <smn.andre@gmail.com>
@@ -32,20 +31,14 @@ final class UXIconRuntime implements RuntimeExtensionInterface
     }
 
     /**
-     * @param array<string, bool|string|AttributeValueInterface> $attributes
+     * @param array<string, bool|string|\Stringable> $attributes
      */
     public function renderIcon(string $name, array $attributes = []): string
     {
         foreach ($attributes as $attribute => $value) {
-            // Typed values produced by twig/html-extra (e.g. "html_attr_type('sst')"), as used by
-            // Twig Components attribute merging, are rendered to their string representation
-            if ($value instanceof AttributeValueInterface) {
-                if (null === $value = $value->getValue()) {
-                    unset($attributes[$attribute]);
-                    continue;
-                }
-
-                $attributes[$attribute] = $value;
+            // the renderer only deals with scalars, so that its decorators (e.g. caches) can rely on it
+            if ($value instanceof \Stringable) {
+                $attributes[$attribute] = (string) $value;
             }
         }
 
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 3.5.0
 
 - Add support for `MergeableInterface` from `twig/html-extra` in `ComponentAttributes#defaults()`
-- Resolve the typed attribute values of `twig/html-extra` and cast `Stringable` values before passing the props to a custom component renderer (`ux:icon`, `ux:map`, ...)
+- Make `ComponentAttributes::all()` return plain values (typed values of `twig/html-extra` resolved, `Stringable` values cast) and pass the props of custom component renderers (`ux:icon`, `ux:map`, ...) through it
 - Add support for dynamic component names in the HTML syntax using `<twig:component is="myComponent" ...`
 
 ## 3.4.0
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 3.5.0
 
 - Add support for `MergeableInterface` from `twig/html-extra` in `ComponentAttributes#defaults()`
-- Make `ComponentAttributes::all()` return plain values (typed values of `twig/html-extra` resolved, `Stringable` values cast) and pass the props of custom component renderers (`ux:icon`, `ux:map`, ...) through it
+- Add `ComponentAttributes::resolved()` to get the attributes as plain values (typed values of `twig/html-extra` resolved, `Stringable` values cast), used to pass the props of custom component renderers (`ux:icon`, `ux:map`, ...)
 - Add support for dynamic component names in the HTML syntax using `<twig:component is="myComponent" ...`
 
 ## 3.4.0
--- a/src/TwigComponent/CHANGELOG.md
+++ b/src/TwigComponent/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 3.5.0
 
 - Add support for `MergeableInterface` from `twig/html-extra` in `ComponentAttributes#defaults()`
+- Resolve the typed attribute values of `twig/html-extra` and cast `Stringable` values before passing the props to a custom component renderer (`ux:icon`, `ux:map`, ...)
 - Add support for dynamic component names in the HTML syntax using `<twig:component is="myComponent" ...`
 
 ## 3.4.0
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -139,11 +139,26 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
     }
 
     /**
+     * Returns the attributes as plain values: the typed values of twig/html-extra are
+     * resolved (a null value becomes false, i.e. the attribute is omitted) and
+     * Stringable values are cast.
+     *
      * @return array<string, string|bool>
      */
     public function all(): array
     {
-        return $this->attributes;
+        $attributes = [];
+        foreach ($this->attributes as $key => $value) {
+            if ($value instanceof AttributeValueInterface) {
+                $value = $value->getValue() ?? false;
+            } elseif ($value instanceof \Stringable) {
+                $value = (string) $value;
+            }
+
+            $attributes[$key] = $value;
+        }
+
+        return $attributes;
     }
 
     /**
--- a/src/TwigComponent/src/ComponentAttributes.php
+++ b/src/TwigComponent/src/ComponentAttributes.php
@@ -32,7 +32,7 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
     private array $rendered = [];
 
     /**
-     * @param array<string, string|bool> $attributes
+     * @param array<string, string|bool|AttributeValueInterface|\Stringable> $attributes
      */
     public function __construct(
         private array $attributes,
@@ -139,13 +139,21 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
     }
 
     /**
-     * Returns the attributes as plain values: the typed values of twig/html-extra are
-     * resolved (a null value becomes false, i.e. the attribute is omitted) and
-     * Stringable values are cast.
+     * @return array<string, string|bool|AttributeValueInterface|\Stringable>
+     */
+    public function all(): array
+    {
+        return $this->attributes;
+    }
+
+    /**
+     * Returns the attributes as plain values, i.e. the way they are rendered:
+     * the typed values of twig/html-extra are resolved (a null value becomes
+     * false, the attribute is omitted) and Stringable values are cast to string.
      *
      * @return array<string, string|bool>
      */
-    public function all(): array
+    public function resolved(): array
     {
         $attributes = [];
         foreach ($this->attributes as $key => $value) {
@@ -277,4 +285,5 @@ final class ComponentAttributes implements \Stringable, \IteratorAggregate, \Cou
     {
         return \count($this->attributes);
     }
+
 }
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -121,6 +121,7 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
                 new Reference('ux.twig_component.component_renderer'),
                 new ServiceLocatorArgument(new TaggedIteratorArgument('ux.twig_component.twig_renderer', indexAttribute: 'key', needsIndexes: true)),
                 new Reference('ux.twig_component.component_stack'),
+                new Reference('twig'),
             ])
             ->addTag('twig.runtime')
         ;
--- a/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
+++ b/src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
@@ -121,7 +121,6 @@ final class TwigComponentExtension extends Extension implements ConfigurationInt
                 new Reference('ux.twig_component.component_renderer'),
                 new ServiceLocatorArgument(new TaggedIteratorArgument('ux.twig_component.twig_renderer', indexAttribute: 'key', needsIndexes: true)),
                 new Reference('ux.twig_component.component_stack'),
-                new Reference('twig'),
             ])
             ->addTag('twig.runtime')
         ;
--- a/src/TwigComponent/src/Twig/ComponentRuntime.php
+++ b/src/TwigComponent/src/Twig/ComponentRuntime.php
@@ -15,6 +15,7 @@ use Psr\Container\ContainerInterface;
 use Symfony\UX\TwigComponent\ComponentRendererInterface;
 use Symfony\UX\TwigComponent\ComponentStack;
 use Symfony\UX\TwigComponent\Event\PreRenderEvent;
+use Twig\Extra\Html\HtmlAttr\AttributeValueInterface;
 
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
@@ -47,7 +48,7 @@ final class ComponentRuntime
     public function render(string $name, array $props = []): string
     {
         if ($this->renderers->has($normalized = strtolower($name))) {
-            return $this->renderers->get($normalized)->render($props);
+            return $this->renderers->get($normalized)->render(self::normalizeRendererProps($props));
         }
 
         return $this->renderer->createAndRender($name, $props);
@@ -55,8 +56,29 @@ final class ComponentRuntime
 
     /**
      * @param array<string, mixed> $props
-     * @param array<string, mixed> $context
      */
+    /**
+     * Custom renderers (e.g. "ux:icon" or "ux:map") receive the props as plain values: the typed attribute
+     * values of twig/html-extra are resolved the way ComponentAttributes renders them, a null value
+     * suppressing the attribute (including the defaults of the renderer), and Stringable values are cast.
+     *
+     * @param array<string, mixed> $props
+     *
+     * @return array<string, mixed>
+     */
+    private static function normalizeRendererProps(array $props): array
+    {
+        foreach ($props as $key => $value) {
+            if ($value instanceof AttributeValueInterface) {
+                $props[$key] = $value->getValue() ?? false;
+            } elseif ($value instanceof \Stringable) {
+                $props[$key] = (string) $value;
+            }
+        }
+
+        return $props;
+    }
+
     public function startEmbedComponent(string $name, array $props, array $context, string $hostTemplateName, int $index): PreRenderEvent
     {
         return $this->renderer->startEmbeddedComponentRender($name, $props, $context, $hostTemplateName, $index);
--- a/src/TwigComponent/src/Twig/ComponentRuntime.php
+++ b/src/TwigComponent/src/Twig/ComponentRuntime.php
@@ -31,7 +31,6 @@ final class ComponentRuntime
         private readonly ComponentRendererInterface $renderer,
         private readonly ContainerInterface $renderers,
         private readonly ComponentStack $componentStack,
-        private readonly ?Environment $twig = null,
     ) {
     }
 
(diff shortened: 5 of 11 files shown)


Kocal:
Oh cool, this is something I wanted to do when some days ago, when I saw that UX Icons rendering was not using our rendering logic from `ComponentAttributes`.

Instead of your change, can you try to use `ComponentAttributes` instead? 🙏
I would say it's fine even if a BC break happens, since the icon will still be rendered in the browser, and the attributes rendering behavior will match be identical to the one everyone know.

Thanks !

Kocal:
And whatever we do here, this is a feature and not a bug fix.

Can you also check for other packages that use component-like rendering, like Map and Chartjs?

Thanks ! 🙏

smnandre:
Hmm.. can we open to Stringable instead ?

Amoifr:
Thanks both! Two findings before I rework it:

1. Map and Chartjs are affected the same way: their `render(array $args)` runtimes hand the values to `StimulusAttributes::addAttribute(string $name, string $value)`, so a typed value like `data-action="...|html_attr_type('sst')"` is a `TypeError` there too.
2. `SeparatedTokenList` and `InlineStyle` from twig/html-extra are not `Stringable`, they only expose `getValue(): ?string`. So accepting `\Stringable` (which I agree we should do anyway, Icons, Map and Chartjs alike) only covers these typed values if html-extra's `AttributeValueInterface` also gains `__toString()`. @smnandre, is that something you would consider on the Twig side? With it, "open to Stringable" becomes the whole fix for the three packages.

For the `ComponentAttributes` route, the constraint is that `ux_icon()`/`IconRenderer` also run without TwigComponent, so I would do it in the component path only (`UXIconRuntime::render()`, and the Map/Chartjs equivalents): build a `ComponentAttributes` from the props, apply the package defaults through `->defaults()`, and render the tag from its string, with the normalization happening in `ComponentAttributes::__toString()` rather than in each package. That is a bigger change (new rendering entry point per package, BC on the generated markup), so I would rather get your go before writing it. Which direction do you prefer: (a) Stringable + html-extra `__toString()` in the three packages, (b) `ComponentAttributes` in the component path of the three packages, or (a) now and (b) as a follow-up?

smnandre:
Thanks for investigating Map and Chartjs as well. It confirms that the question is broader than Icons, but I think we should be careful before generalizing the current approach.

`IconRendererInterface` was deliberately exposed as a Twig-independent API. Applications may implement or decorate it, notably to cache rendered icons.

Changing its PHP signature would be a BC break for existing implementations. Keeping the `array` signature while allowing arbitrary objects inside the attributes would still be a semantic BC break: cache decorators may currently assume that attribute values are stable, serializable primitives when building their cache keys.

Both `ux_icon()` and `<twig:ux:icon>` already pass through `UXIconRuntime`. This is indeed the "appropriate" integration boundary. Twig-specific values can be normalized there before calling the renderer, without changing the renderer interface or exposing Twig types to decorators.

I still think `Stringable` would be the most open and standard contract for values that can be converted to strings. The runtime could eagerly convert them to raw strings, while the final renderer remains responsible for escaping. This keeps `IconRendererInterface` independent from both Twig and `twig/html-extra`.

Of course, this does not currently cover `SeparatedTokenList` or `InlineStyle`, since they are not `Stringable`.

There is also an expected-behavior question we should answer first: when `AttributeValueInterface::getValue()` returns `null`, should that always suppress the attribute, including a configured default, or should it mean that no invocation value was provided and therefore allow the default to apply?

The current implementation removes the key before calling `IconRenderer`, so a configured default can reappear. We should define that behavior explicitly before choosing the normalization strategy.

More generally, I am not sure Symfony UX should start adopting `twig/html-extra`'s `AttributeValueInterface` independently across Icons, Map, Chartjs, and potentially other packages. Twig has its own release cycle and can publish a new version at any time, independently from Symfony UX. Coupling several UX rendering APIs to a type owned by that package could become difficult to evolve consistently.

I think it may be worth pausing and considering this globally.

If Symfony UX needs a common HTML attributes abstraction, that sounds useful, but I would expect it to be:
- typed;
- injectable;
- extensible;
- independent from Twig and `twig/html-extra`;
- explicit about merging, omission, escaping ownership, and cacheability.

Twig and `twig/html-extra` values could then be adapted to that abstraction at the integration boundary, instead of becoming part of the core renderer contracts.

smnandre:
> Since Twig Components 3.5, attribute merging relies on the typed values of twig/html-extra (html_attr_type(), tailwind_classes, ...)

I'm late on this.. sorry.. but reading it like this makes me a little uneasy :/ 

I'm not sure it is wise for us to make `twig/html-extra` the foundation of this mechanism, especially if we then document it as the way attribute merging works and start adapting the internal implementation of every UX package around it.

The mechanism itself makes sense, but this would effectively make all UX packages depend, somewhat implicitly, on a package that lives outside the UX codebase and may have a different release and maintenance lifecycle. That feels a little fragile as a long-term architectural choice.

Actually, I think this is another very good example of something a dedicated UX bundle could help with: it could own the integration layer and register the appropriate intermediaries, decorators, or other extension points between Twig Components and `html-extra`.

That would let us benefit from the typed values provided by `html-extra` without making its current implementation details part of the implicit contract of every UX package.

Kocal:
I see this differently.

`twig/html-extra` isn't "outside" the UX codebase, it's part of the same first-party Twig ecosystem we already depend on for everything else. And Symfony UX exists precisely to make front-end tooling work well within Symfony/Twig, it lives inside that ecosystem by design, not outside it. So using Twig's own attribute tooling isn't scope creep, it's UX standing on the foundation it's built on.

IMHO an abstraction is overthinking things.

I'd rather we reuse the actual attribute rendering and merging logic `html-extra` already implements, so plain  Symfony/Twig users and UX users get the exact same behavior, instead of a parallel abstraction we'd have to keep in sync forever.

smnandre:
> twig/html-extra isn't "outside" the UX codebase

Yes it is :) I could almost say i's even outside the Twig codebase .. but that would definitively be some trolling :) 

More seriously, I shared my doubts with @nicolas-grekas yesterday after I posted here... and he told me pretty much the same things you do.. so that's all good for me. I trust you guys more than I trust my judgement here :)

Amoifr:
Thanks for sorting it out! Reworked accordingly (title and description updated, feature, not bug fix):

- the resolution now happens once in `ComponentRuntime::render()`, right before a custom renderer (`ux:icon`, `ux:map`, ...) gets the props, so Icons, Map and any future custom renderer benefit from it without touching their own contracts (`IconRendererInterface` and its decorators keep seeing scalars, as @smnandre asked);
- `Stringable` values are cast there as well, and `UXIconRuntime` casts them too for the `ux_icon()` function path;
- on the open question: a typed value resolving to `null` now becomes `false`, i.e. the attribute is dropped, defaults of the renderer included. That is the behavior of `ComponentAttributes::__toString()` for a `null` typed value (it is skipped, and a `defaults()` entry for that key is overridden, not restored), and `false` is already how `Icon::toHtml()` and `StimulusAttributes` express "no attribute", so the three paths agree.

The `twig/html-extra` require-dev bump is gone with it.

Kocal:
Following the last discussions, why not using `ComponentAttributes` for attributes merging/rendering instead of our custom logic?

smnandre:
We can, but it's still before the IconRendererInterface or the MapRendererInterface, as these bundle logically do not require twig component :)

Amoifr:
Done: `ComponentAttributes::all()` now returns the plain values (typed values resolved, `Stringable` cast), which is what its `array<string, string|bool>` docblock already promised, and `ComponentRuntime` builds a `ComponentAttributes` from the props and hands `all()` to the custom renderer. No custom logic left, and `IconRendererInterface` / `MapRendererInterface` stay untouched, before the Twig Component boundary as @smnandre asked.

src/TwigComponent/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 3.5.0
 
 - Add support for `MergeableInterface` from `twig/html-extra` in `ComponentAttributes#defaults()`
+- Make `ComponentAttributes::all()` return plain values (typed values of `twig/html-extra` resolved, `Stringable` values cast) and pass the props of custom component renderers (`ux:icon`, `ux:map`, ...) through it
```
smnandre: Beware the versions alignment there between UX Icon, UX Twig Component, html-extra ... :)

src/TwigComponent/src/Twig/ComponentRuntime.php
```diff
@@ -47,6 +51,12 @@ public function preRender(string $name, array $props): ?string
     public function render(string $name, array $props = []): string
     {
         if ($this->renderers->has($normalized = strtolower($name))) {
+            if (null !== $this->twig) {
```
smnandre: Not sure this can happen (and this class is stricly internal)

src/TwigComponent/src/Twig/ComponentRuntime.php
```diff
@@ -47,6 +51,12 @@ public function preRender(string $name, array $props): ?string
...
+                // custom renderers (e.g. "ux:icon" or "ux:map") deal with plain values: resolve the
+                // typed attribute values the way ComponentAttributes renders them
```
smnandre, line 55: Let's test if you have props maybe ? to avoid an empty round ?

smnandre: I may have missunderstood, but I imagined this would be done in the icon and maps runtime personnaly.

src/TwigComponent/src/Twig/ComponentRuntime.php
```diff
@@ -55,7 +65,6 @@ public function render(string $name, array $props = []): string
 
     /**
      * @param array<string, mixed> $props
-     * @param array<string, mixed> $context
```
smnandre, line 58: revert :)

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -139,11 +139,26 @@ public function render(string $attribute): ?string
...
+     * resolved (a null value becomes false, i.e. the attribute is omitted) and
+     * Stringable values are cast.
+     *
      * @return array<string, string|bool>
      */
     public function all(): array
     {
-        return $this->attributes;
+        $attributes = [];
+        foreach ($this->attributes as $key => $value) {
```
smnandre, line 159: Let's update the docblocks of the constructor then because currently 

array<string, string|bool> $attributes

is the type of $attributes 

:|

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -139,11 +139,26 @@ public function render(string $attribute): ?string
     }
 
     /**
+     * Returns the attributes as plain values: the typed values of twig/html-extra are
+     * resolved (a null value becomes false, i.e. the attribute is omitted) and
+     * Stringable values are cast.
+     *
```
smnandre: This feels like a BC break.

What aboiut a resolved() or normalized() method ?

src/Icons/src/Twig/UXIconRuntime.php
```diff
@@ -31,10 +31,17 @@ public function __construct(
...
      */
     public function renderIcon(string $name, array $attributes = []): string
     {
+        foreach ($attributes as $attribute => $value) {
+            // the renderer only deals with scalars, so that its decorators (e.g. caches) can rely on it
+            if ($value instanceof \Stringable) {
+                $attributes[$attribute] = (string) $value;
+            }
+        }
+
```
Kocal: I'm sorry if I'm missing something (I'm on smartphone), but why do we do that here, instead of leveraging the attributes merging/rendering in the IconRenderer?

src/TwigComponent/tests/Unit/ComponentAttributesTest.php
```diff
@@ -424,4 +424,43 @@ public static function valueProvider(): iterable
         yield 'sql injection' => ["' OR 1=1 --", '&#039; OR 1=1 --'];
         yield 'url encoded xss' => ['%3Cscript%3Ealert(1)%3C/script%3E', '%3Cscript%3Ealert(1)%3C/script%3E'];
     }
+
+    public function testAllReturnsPlainValues()
```
smnandre: Let's still have a test to prevent any regression and enforce the method returns the same thing whereas html-extra is installed of not ?

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -139,11 +139,26 @@ public function render(string $attribute): ?string
...
-        return $this->attributes;
+        $attributes = [];
+        foreach ($this->attributes as $key => $value) {
+            if ($value instanceof AttributeValueInterface) {
+                $value = $value->getValue() ?? false;
+            } elseif ($value instanceof \Stringable) {
+                $value = (string) $value;
+            }
+
+            $attributes[$key] = $value;
```
Kocal, line 166: Should we really _duplicate_ the logic from `__toString()` here?

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
```diff
@@ -121,6 +121,7 @@ static function (ChildDefinition $definition, AsTwigComponent $attribute) {
                 new Reference('ux.twig_component.component_renderer'),
                 new ServiceLocatorArgument(new TaggedIteratorArgument('ux.twig_component.twig_renderer', indexAttribute: 'key', needsIndexes: true)),
                 new Reference('ux.twig_component.component_stack'),
+                new Reference('twig'),
```
smnandre: This feels wrong.. either degine the Twig function with with need_environment or do this higher .. because injecting a Twig environment in a TwigExtension feels yeah.. a bit odd.  What do you think ? 

Also there is a risk of double escaping there, no ?

src/TwigComponent/CHANGELOG.md
smnandre: This really feels like a BC break here 🤔

Amoifr: Noted, the Icons entry only needs a `Stringable` value, no dependency on Twig Component or html-extra.

src/TwigComponent/src/Twig/ComponentRuntime.php
Amoifr: Right, the environment is no longer injected: `component()` is now declared with `needs_environment` and `render()` gets it as first argument (`<twig:ux:icon>` compiles to `component('ux:icon', ...)`, so both syntaxes go through it).

src/TwigComponent/src/Twig/ComponentRuntime.php
Amoifr: Done, `if ($props)`.

Amoifr: I kept it in `ComponentRuntime` so that Map and any future custom renderer get it for free, and the resolution logic lives in one place (`ComponentAttributes`), as Kocal suggested. Happy to move it per package if you both prefer.

src/TwigComponent/src/Twig/ComponentRuntime.php
Amoifr: Reverted.

src/TwigComponent/src/ComponentAttributes.php
Amoifr: Updated to `array<string, string|bool|AttributeValueInterface|\Stringable>`.

src/TwigComponent/src/ComponentAttributes.php
Amoifr: Agreed: `all()` is untouched again, the plain values come from a new `resolved()` method (your suggestion). Changelog reworded.

src/Icons/src/Twig/UXIconRuntime.php
Amoifr: Because `IconRendererInterface` is the Twig-independent contract smnandre wants to keep scalar-only for decorators (caches), the cast happens at the Twig boundary. `ux_icon()` is the only path that does not go through `ComponentRuntime`, hence this one.

src/TwigComponent/tests/Unit/ComponentAttributesTest.php
Amoifr: Added `testResolvedReturnsTheSameValuesAsAllForPlainValues`, no html-extra type involved.

src/TwigComponent/src/ComponentAttributes.php
Amoifr: Good point: extracted a private `resolveValue()` now shared by `__toString()`, `render()` and `resolved()`.

src/TwigComponent/src/DependencyInjection/TwigComponentExtension.php
Amoifr: Right on both counts: no more injection, `component()` is declared with `needs_environment` (see above). And no double escaping: `resolved()` returns raw values, escaping stays where it is (`Icon::toHtml()`, `StimulusAttributes`).

src/TwigComponent/CHANGELOG.md
Amoifr: Agreed, reworded: `all()` is back to its original behavior, the new `resolved()` method carries the feature.

src/Icons/src/Twig/UXIconRuntime.php
smnandre: > Because IconRendererInterface is the Twig-independent contract smnandre wants to keep scalar-only for decorators (caches)

Not really :) Because there is no technical reason for that. The caller can resolve whatever class/value it wants before calling the renderer.

But the main issue is BC: changing this interface in one of our most downloaded packages would need a pretty strong benefit, and I'm not convinced we have one here.

We can absolutely expose the inner SVG later, or provide a custom renderer. That's also the nice thing about keeping interfaces minimal.

I'm open to the feature, but I'd rather avoid adding complexity to everyone's hot path for a very specific need, especially for negligible gains in dev.

And on the broader consistency point, I'd prefer we discuss the general approach first. 'We've done it in 90% of packages, so we should do the last 10%' can otherwise feel a little like pressure rather than a technical argument.

Amoifr: Fair enough, and sorry if the consistency argument came across as pressure, that was not the intent. You convinced me on the substance too: the only path that needed the runtime cast was the `ux_icon()` function with a typed value, which nobody actually hit (the reported case went through the component syntax, now handled in `ComponentRuntime`). I dropped the Icons changes entirely: the PR is now TwigComponent-only (`resolved()` + the custom-renderer resolution), `IconRendererInterface` and `UXIconRuntime` untouched. The broader discussion about exposing richer values to renderers can happen on its own terms whenever it is useful.

smnandre: Regarding UX Icons specifically, it's the third most downloaded UX package every month, despite not being preinstalled with Symfony like Stimulus or Turbo.

It has already been more installed than LiveComponent.. in roughly half the time.
( well... and about 3× more than UX React, UX Vue, _and_ UX Svelte **combined** but 🤫 )

So yes, I tend to be a little careful with changes to that package 🥹 

But more seriously, it's also because it is used by major open source PHP projects, and I do think that's worth considering before adding dependencies or complexity, or changing existing behavior in a package that already does its job pretty well (not saying this about this PR but in general, I promise :) )

--

Thank you very much for all your PRs lately, your responsiveness, and especially your positive attitude throughout all these discussions. It genuinely brings a lot of energy and makes things more enjoyable. 

🤝

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -262,4 +274,20 @@ public function count(): int
     {
         return \count($this->attributes);
     }
+
+    /**
+     * Resolves the typed values of twig/html-extra and casts Stringable values.
+     */
+    private static function resolveValue(mixed $value): mixed
```
smnandre:
```suggestion
    private static function resolveValue(??? $value): ???
```

Is it really mixed there ? Don't we expect string|bool|null ? 

Similarely, should the $value be typed AttributeValueInterface|\Stringable

Then, I'll ask a last time because I really do think there is something I don't get here.

If AttributeValueInterface is used to signal something than can be null but then we "pass" or we change it in string .... should the interface be Stringable ?

src/TwigComponent/src/ComponentAttributes.php
```diff
@@ -117,14 +114,10 @@ public function render(string $attribute): ?string
             return null;
         }
 
-        if ($value instanceof AttributeValueInterface && null === $value = $value->getValue()) {
+        if (null === $value = self::resolveValue($value)) {
```
smnandre: That's an added method call for every attribute of every component during the render, I'm thinking this could be inlined as before maybe...  @Kocal  ?

Amoifr: Re-inlined: `__toString()` and `render()` are back to their exact original bodies (this diff no longer touches them at all), and `resolved()` carries its own resolution, out of the hot path. @Kocal that partially undoes the dedup you suggested — six duplicated lines in a cold method felt like the cheaper trade-off.

src/TwigComponent/src/ComponentAttributes.php
Amoifr: `resolveValue()` is gone with the re-inlining, so the `mixed` question resolves itself (it was `mixed` because inside `__toString()` the value can still be anything at that point — the descriptive `LogicException` for non-scalar props fires after resolution). On your last question: `getValue(): ?string` can express "absent" (null → the attribute is omitted, even overriding a default), which `__toString()` cannot — a `Stringable`-only contract could not model removal. That is, as I understand it, why html-extra keeps `AttributeValueInterface` distinct rather than just using `Stringable`.

src/Icons/src/Twig/UXIconRuntime.php
Amoifr: Completely fair given those numbers — and thanks a lot for the kind words, the feeling is mutual. 🤝

src/TwigComponent/src/ComponentAttributes.php
Kocal: Why not yes

Kocal:
Due to all the discussions in multiple threads, I'm a bit lost about where we are going and what we want to do/what we did.

What's the PR's state, where do we want to go?

Amoifr:
Fair, let me recap. **Problem** (found while testing #3469): passing the typed attribute values our Toolkit ships (`{{ ...dropdown_menu_trigger_attrs }}`, containing `html_attr_type()` values) to `<twig:ux:icon />` blows up with `Invalid value type for attribute "data-action" … SeparatedTokenList provided`.

**Current state of the PR** (now TwigComponent-only, 6 files):

- `ComponentRuntime::render()` resolves the props before handing them to a custom renderer (`ux:icon`, `ux:map`, any future one): typed values resolved, `Stringable` cast. The `component()` Twig function is declared with `needs_environment` to reach the escaper — no service wiring change.
- `ComponentAttributes` gains a cold `resolved(): array<string, string|bool>` doing that resolution, plus widened constructor docblocks. `__toString()`, `render()` and `all()` are byte-identical to current `3.x` — hot path untouched, per smnandre's last round.
- Icons and Map are no longer touched at all: `IconRendererInterface`, `UXIconRuntime` and `ux_icon()` keep their scalar-only contracts.

So: one new cold method + one resolution point at the custom-renderer boundary, unit-tested on both. That's the whole PR.

src/TwigComponent/tests/Unit/Twig/ComponentRuntimeTest.php
```diff
@@ -0,0 +1,88 @@
...
+            'data-action' => $this->createTypedValue('click->menu#toggle'),
+            'aria-hidden' => $this->createTypedValue(null),
+            'title' => new class implements \Stringable {
+                public function __toString(): string
+                {
+                    return 'User';
+                }
+            },
+            'disabled' => true,
+        ]));
```
Kocal, line 63: And so, what happens if you pass an attribute `['toto' => null, 'tata' => false]`?

It will diverges from the actual `ComponentAttributes::__toString()` right? Again, why don't we re-use a common logic? Sorry if it has been already explained, but there are too much threads and too much comments there.

Kocal:
Since it diverged a lot, I updated the PR title, description, and label, given the previous discussions.

Kocal:
To be honest, I don't really like where this PR is going, I feel like something is wrong but I can't explain why. 
I will open a new PR to leverage attributes rendering to Twig's `html_attr`, the rendering engine Symfony and Symfony UX use.

smnandre:
Feel free to open a RFC first and discuss it. But there is no way maker / toolkit bundle should dictate how the most installed / downloaded work. Especially as there is no added feature (fix me if I'm wrong with a real life scenario) ...

Kocal:
> But there is no way maker / toolkit bundle should dictate how the most installed / downloaded work

That won't be the case, these are absolutely not related. I just think attributes rendering must be normalized with `html_attr()` from Twig HTML Extra, which render attributes in a much better way than we actually do. 

Also... the MakerBundle?

**EDIT:** opened a better alternative https://github.com/symfony/ux/pull/3821 based on https://github.com/twigphp/Twig/pull/4895

Amoifr:
Nice, #3821 is a better home for this: funneling through `html_attr()`'s own logic is more consistent than resolving in TwigComponent, and it keeps the hot path untouched, which was smnandre's concern.

One note for the record: it covers the two renderers we ship, so a third-party custom renderer will need the same loop. That is fine now that `htmlAttrValue()` becomes public API (twigphp/Twig#4895), it just wasn't reachable before, which is why this PR went through the component layer.

Happy to close this one in favor of yours whenever you want. Thanks for taking the time to reshape it, and sorry for the detour through the hot path. 😄

Kocal:
PR https://github.com/twigphp/Twig/pull/4895 has been merged, and so https://github.com/symfony/ux/pull/3820 and https://github.com/symfony/ux/pull/3821 will be merged too very soon.

I believe we can close this PR.

Thanks for pointing out that subject which lead to a nice discussion. And thanks for your contribution!

PR #3815, opened 2026-08-25
[CalendarLink] Alias `IcsBuilder` so it can be autowired and decorated

### LINKED ISSUE #3812: [CalendarLink] IcsBuilder is registered without a class alias, so it cannot be autowired or decorated
**Component:** CalendarLink
**Version:** v3.4.0

## Problem

In `config/services.php`, the registry gets a class alias but the builder does
not:

```php
$services->set('ux_calendar_link.ics.builder', IcsBuilder::class);
// ...
$services->alias(CalendarLinkProviderRegistry::class, 'ux_calendar_link.registry');
```

So `CalendarLinkProviderRegistry` is autowirable while `IcsBuilder` is
reachable only by its string id. An application that wants to type-hint it —
or decorate it via `#[AsDecorator(IcsBuilder::class)]` — has to declare the
alias itself.

This looks like an oversight rather than a deliberate boundary: the two
services are registered side by side and only one got the alias. If the builder
*is* meant to be internal, `@internal` on the class would make that explicit;
right now it is a public, final, documented-looking class that simply is not
wired like one.

## Suggested fix

```php
$services->alias(IcsBuilder::class, 'ux_calendar_link.ics.builder');
```

...or mark `IcsBuilder` `@internal` if the omission is intentional. Either way
the current state is ambiguous. Happy to open a PR.

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3812 
| License        | MIT

The registry got a class alias but the builder did not, leaving it reachable only by its string service id. Aliasing `IcsBuilder::class` makes it type-hintable and decoratable like the registry.

### DIFF
--- a/src/CalendarLink/config/services.php
+++ b/src/CalendarLink/config/services.php
@@ -27,6 +27,8 @@ return static function (ContainerConfigurator $container): void {
 
     $services->set('ux_calendar_link.ics.builder', IcsBuilder::class);
 
+    $services->alias(IcsBuilder::class, 'ux_calendar_link.ics.builder');
+
     $services->set('ux_calendar_link.provider.google', GoogleCalendarLinkProvider::class)
         ->tag('ux_calendar_link.provider');
 
--- a/src/CalendarLink/CHANGELOG.md
+++ b/src/CalendarLink/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 3.5.0
+
+- Add a class alias for `IcsBuilder` so it can be autowired and decorated
+
 ## 3.1
 
 - Add component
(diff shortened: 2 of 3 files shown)


src/CalendarLink/config/services.php
```diff
@@ -27,6 +27,8 @@
 
     $services->set('ux_calendar_link.ics.builder', IcsBuilder::class);
 
+    $services->alias(IcsBuilder::class, 'ux_calendar_link.ics.builder');
```
smnandre, line 30: Should this be an interface then, much easier / "cleaner" to decorate ?

Kocal:
I'm closing, `IcsBuilder` should have been internal

PR #3817, opened 2026-08-25
[CalendarLink] Derive a stable ICS UID from the event content

### LINKED ISSUE #3810: [CalendarLink] Random UID per build creates duplicate events on re-add
**Component:** CalendarLink
**Version:** v3.4.0

## Problem

`IcsBuilder::build()` mints a fresh UID on every call:

```php
'UID:'.$this->uuidFactory->create()->toRfc4122(),
```

`UID` is the property calendar clients use to decide whether an incoming
`VEVENT` is a *new* event or an *update to one they already hold*
(RFC 5545 §3.8.4.7). Because it is random per build, the same
`CalendarEvent` serializes to a different identity every time.

In the exact flow this component targets — an "Add to calendar" button on a
booking/event page — that means:

- a user who clicks the `.ics` link, then reloads and clicks again, gets **two
  copies** of the event rather than one;
- the same event rendered on two pages (confirmation page + reminder page)
  yields two unrelated entries;
- an event whose details change cannot be re-issued as a correction, since the
  new file looks like an unrelated event.

## Current status

This is currently pinned as intended behaviour:

```php
// tests/Ics/IcsBuilderTest.php
public function testUidIsUniquePerBuild()
{
    // ...
    $this->assertNotSame($first[1], $second[1]);
}
```

`IcsBuilder` does accept a `UuidFactory`, but `UuidFactory::create()` takes no
arguments, so the UID cannot be derived from the event — and the service is
registered as `ux_calendar_link.ics.builder` with no class alias, so it is not
autowirable for a userland override either (see separate issue).

## Suggested fix

Either would resolve it:

1. **Optional explicit UID** — add `?string $uid = null` to `CalendarEvent`,
   letting the application pass its own stable domain key (a booking UUID,
   typically). Least surprising, and matches how RFC 5545 expects UIDs to be
   assigned.
2. **Deterministic derivation** — default the UID to a UUIDv5 over the event's
   own content (title, start, end, location). Keeps the zero-config path
   working while making identical events identical.

Option 1 alone would be enough for most applications; option 2 makes the
default behaviour correct for everyone.

Happy to open a PR if a direction is preferred.

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3810 
| License        | MIT

`IcsBuilder` minted a random UID on every build, so the same CalendarEvent serialized to a different identity each time and calendar clients created duplicates on re-add. 

Add an optional `uid` on CalendarEvent for an application-supplied key, and default to a UUIDv5 derived from the event content (title, start, end, location).

### DIFF
--- a/src/CalendarLink/CHANGELOG.md
+++ b/src/CalendarLink/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Make `IcsBuilder` internal
 - Serialize timed events in a named time zone with `;TZID=` and a matching `VTIMEZONE`, so recurring events no longer drift by an hour across DST
 - Accept an optional `ClockInterface` in `IcsBuilder` so `DTSTAMP` is deterministic and testable
+- Add an optional `uid` to `CalendarEvent`, and derive the ICS `UID` from the event content by default instead of a random value, so re-adding the same event no longer duplicates it
 
 ## 3.1
 
--- a/src/CalendarLink/src/CalendarEvent.php
+++ b/src/CalendarLink/src/CalendarEvent.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\UX\CalendarLink;
 
+use Symfony\Component\Uid\Uuid;
 use Symfony\UX\CalendarLink\Exception\InvalidArgumentException;
 
 /**
@@ -23,6 +24,7 @@ final class CalendarEvent
 
     /**
      * @param list<CalendarReminder> $reminders
+     * @param Uuid|null              $uid       A stable UID reused across builds so clients update the event instead of duplicating it; derived from the event content when null
      */
     public function __construct(
         public readonly string $title,
@@ -34,6 +36,7 @@ final class CalendarEvent
         public readonly ?string $url = null,
         public readonly ?CalendarRecurrence $recurrence = null,
         public readonly array $reminders = [],
+        public readonly ?Uuid $uid = null,
     ) {
         if ('' === trim($title)) {
             throw new InvalidArgumentException('Event title must not be empty.');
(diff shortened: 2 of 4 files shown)


smnandre:
Can we try matching the RFC here ? https://www.rfc-editor.org/info/rfc2445/#section-4.8.4.7 

UID has an expected format ..

Kocal:
The RFC you're linking is obsolete (see the very top of the page), and has been replaced by https://www.rfc-editor.org/info/rfc7986/.

But, it's already the case no? Or are you referring to the `return $event->uid` that allows to returns a value which is not a UUID string? 

I just pushed a modification where we rely on a `Uuid` instance instead, it should be fine now.

smnandre:
My (very badly expressed) point was that UID are encouraged to follow a format, that is not an UID in the Symfony way

> Though other algorithms will work, it is RECOMMENDED that the right-hand
      side contain some domain identifier (either of the host itself or
      otherwise) such that the generator of the message identifier can
      guarantee the uniqueness of the left-hand side within the scope of
      that domain.

and
      
>   Example:  The following is an example of this property:
       UID:19960401T080045Z-4000F192713-0052@example.com


 https://www.rfc-editor.org/info/rfc5545/#section-3.8.4.7
 
 I feel this would be a good idea if possible without re-coding many things there. Or it can be a following PR no problem for me I'll open it .. but this was more a "let's start on the most default-standard-recommended system, even if we allow all spec"

smnandre (approved):

PR #3820, opened 2026-08-26
[TwigComponent] Render component attributes through twig/html-extra's `html_attr()`

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | #3261
| License       | MIT

`ComponentAttributes` hand-rolls its own attribute rendering: it throws on `null`, drops `aria-*` `false` while turning `true` into `"true"`, renders a boolean `true` bare, and throws on arrays, iterables and enums. A partial, in places invalid reimplementation of what Twig now does natively.

Both `{{ attributes }}` and `attributes.render(...)` now resolve values through `HtmlExtension::htmlAttrValue()`, the escaper-free building block behind `twig/html-extra`'s `html_attr()`. Depends on twigphp/Twig#4895; requires `twig/html-extra` `^3.29` and `twig/twig` `^3.24`.

Unchanged: merging (`attributes.defaults(...)`, including the `class`/`data-controller`/`data-action` concatenation), the two pre-filters (already-rendered attributes, nested keys like `foo:bar`), and the constructor.

## Behavior changes

| Attribute value | Before | After |
| --- | --- | --- |
| `null` | throws | attribute omitted |
| `false` | attribute omitted | attribute omitted |
| `true` | `name` | `name=""` |
| `true` on `aria-*` | `aria-x="true"` | `aria-x="true"` |
| `false` on `aria-*` | attribute omitted | `aria-x="false"` |
| `true` on `data-*` | `data-x` | `data-x="true"` |
| array or iterable | throws | `class="a b"`, `style="color: red;"` |
| array on `data-*` | throws | JSON encoded |
| a `BackedEnum` | throws | its backing value |
| a `\Stringable` | cast to string | cast to string |

- `attributes.render()` returns `null` for an omitted attribute instead of throwing on a non-string value; the `aria-*` `false` change matches React and Vue and settles #3261
- LiveComponent's internal `data-live-preserve` marker now renders `data-live-preserve="true"`; harmless, its JavaScript matches on presence


Kocal:
Added you @nicolas-grekas as reviewer, since it seems you had similar opinions on this subject https://github.com/symfony/ux/pull/3799#issuecomment-5377993644

kbond (commented):
Screenshot 2026-08-26 at 8 21 43 AM

I love this on `ComponentAttributes`!

On the behaviour changes: there's some BC breaks there, no? But maybe to minor to be realistic?

Kocal:
> On the behaviour changes: there's some BC breaks there, no? But maybe to minor to be realistic?

Indeed we have few changes yes, listed in the PR description, but I don't consider them as a BC:
- We have 3 cases (null, array, BackedEnum) which are now correctly handled instead of throwing an exception ➡️ that's better, also throwing on `null` was a mistake
- And we have these 3 other cases were the rendering changed, but it can't be considered as a BC:

| Attribute value | Before | After |
| --- | --- | --- |
| `true` | `name` | `name=""` |
| `false` on `aria-*` | attribute omitted | `aria-x="false"` |
| `true` on `data-*` | `data-x` | `data-x="true"` |

To resume, the behavior is now identical to what Twig does, and we have less code to maintain 🎉 

---

Also, Fabien just merged https://github.com/twigphp/Twig/pull/4895, but GitHub Actions have some issues right now, so... our CI stays yellow for the moment 😓

Kocal:
Waiting for Twig 3.29 to be released, otherwise the CI jobs will keep failing.

**EDIT:** https://github.com/twigphp/Twig/releases/tag/v3.29.0 \o/

src/TwigComponent/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Add support for `MergeableInterface` from `twig/html-extra` in `ComponentAttributes#defaults()`
```
Kocal: This line can be removed

src/TwigComponent/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
...
 - Add support for dynamic component names in the HTML syntax using `<twig:component is="myComponent" ...`
 - Capture `## <description>` documentation comments written above props inside `{% props %}` (Twig 3.29+), exposed per prop via `PropsNode::getPropDocumentation()`
+- Render component attributes with twig/html-extra's `html_attr()` logic, through `HtmlExtension::htmlAttrValue()`: a `null` value now omits the attribute (instead of throwing), `aria-*` booleans render `"true"`/`"false"` symmetrically, a boolean `true` renders as `name=""` (`data-*` renders `"true"`), and array/iterable/`BackedEnum` values are supported. `attributes.render()` follows the same rules and returns `null` for an omitted attribute, instead of throwing on a non-string value. Requires `twig/html-extra` `^3.29` and `twig/twig` `^3.24`.
```
Kocal: We can definitely make it more human friendly, with bullet points

src/TwigComponent/CHANGELOG.md
Kocal: Done

src/TwigComponent/CHANGELOG.md
Kocal: Done

mpdude:
@Kocal I lost sight of #3261, so many thanks for picking up on `html_attr` and bringing this to an end. 🤙🏻💪🏻

Kocal:
That was the most logical choice to do there, thanks a lot for your work on Html extra ;)

PR #3821, opened 2026-08-26
[Icons][Map] Render attributes through twig/html-extra's html_attr() logic

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #3469
| License       | MIT

This builds on symfony/ux#3820: UX Icons and UX Map now share the same attribute rendering, through `HtmlExtension::htmlAttrValue()`, which fixes #3469 where the Toolkit's typed attribute values crashed `<twig:ux:icon>` and `<twig:ux:map>`. It needs twigphp/Twig#4895 and adds `twig/html-extra` `^3.29` to both packages.

- Also fixes a separate bug in the same diff: an omitted `aria-label`, `aria-labelledby` or `title` no longer suppresses UX Icons' automatic `aria-hidden="true"`, which used to leave the icon with no textual alternative and not hidden from assistive technology (predates this PR).

Ignore the first commit during the review, it's basically https://github.com/symfony/ux/pull/3820

### DIFF
--- a/src/Icons/CHANGELOG.md
+++ b/src/Icons/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Migrate SVG parsing to the PHP 8.4 `Dom\XMLDocument` API, icons rendered from local files now correctly include the `xmlns` attribute.
   It **may break your pipeline** if you assert on `ux_icon()` or `<twig:ux:icon>` output in your tests. Re-generate those assertions and clear your icon cache after upgrading.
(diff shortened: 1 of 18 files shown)


smnandre (approved):

Kocal:
Waiting for Twig 3.29 

**EDIT:** https://github.com/twigphp/Twig/releases/tag/v3.29.0 \o/

PR #3825, opened 2026-08-31
[Toolkit] Add the package documentation

### LINKED ISSUE #3824: [Toolkit] Add some documentation
The Toolkit package has no docs yet, so the documentation link on https://symfony.com/packages/ux-toolkit returns a 404 error.

There's no need to create a very detailed doc page showing all components and their options. A short doc explaining how to start with the Toolkit and how to use it in real Symfony projects would be enough.

Also, the issue that tracks UX Toolkit completeness (https://github.com/symfony/ux/issues/3233) seems outdated.

Thanks

Amoifr:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| Issues         | Fix #3824
| License        | MIT

The Toolkit is the only UX package without a `doc/index.rst`, which is why https://symfony.com/packages/ux-toolkit returns a 404. Twenty other packages have one.

@javiereguiluz asked for a short page rather than a catalogue of every component and option, so this covers getting started and using the Toolkit in a real project, and leaves the per-component reference to ux.symfony.com where it is generated from the kits.

What it documents:

- that the Toolkit is not a runtime dependency, since `ux:install` copies the source into your application and the code becomes yours. This felt like the thing to lead with: it is what makes the Toolkit different from a component library, and it explains why there is nothing to upgrade;
- installing a component, with and without arguments, and the case-insensitive matching with suggestions;
- choosing a kit, including kits hosted outside the repository and pinned to a branch;
- that dependencies are reported rather than installed;
- the three options of `ux:install`, and what `--force` protects you from;
- pointers to ux.symfony.com for browsing components, and to the Toolkit contributing guide for authoring a kit.

`ux:toolkit:debug-kit` and `ux:toolkit:create-kit` are deliberately left out: both are `hidden: true`, so they are kit-authoring tooling rather than part of the user-facing surface.

Everything here was checked against the source on `3.x` rather than written from memory, including the interactive flow of `InstallCommand`, the confirmation prompt in `Installer::copyFile()`, the kit names accepted by `--kit`, and the homepage URLs in each kit manifest. DOCtor-RST passes with the repository configuration.

The issue also notes that #3233 looks outdated. I have not touched it, that one is yours to judge.

### DIFF
--- a/src/Toolkit/doc/index.rst
+++ b/src/Toolkit/doc/index.rst
@@ -1,0 +1,117 @@
+Symfony UX Toolkit
+==================
+
+**EXPERIMENTAL** This component is currently experimental and is likely
+to change, or even change drastically.
+
+Symfony UX Toolkit gives you ready-to-use UI recipes for your Symfony
+application. It is part of `the Symfony UX initiative`_.
+
+It is not a library that you depend on at runtime. The ``ux:install`` command
+copies the source of a recipe into your project, and from there the code is
+yours: you read it, edit it and commit it like any other template of your
+application.
+
+Nothing is hidden in ``vendor/``, and nothing updates it for you either. That
+trade is what makes the Toolkit different from a component library, and it is
+described in `The Recipe Is Yours, For Worse and For Better`_.
+
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require --dev symfony/ux-toolkit
+
+The package only registers console commands, so it is needed while you install
+recipes, not to render them.
+
+Kits and Recipes
+----------------
+
+A *recipe* is a pack of files and dependencies: Twig components, Stimulus
+controllers, and the Composer or npm packages they need to run. ``button`` and
+``dialog`` are recipes.
+
+A *kit* is a collection of recipes sharing one design system, like Shadcn UI or
+Bootstrap. The kits shipped with the Toolkit are browsable in the
+`ux-toolkit repository <https://github.com/symfony/ux-toolkit/tree/3.x/kits>`_.
+
+Installing a Recipe
+-------------------
+
+Run the ``ux:install`` command with the name of the recipe you want:
+
+.. code-block:: terminal
+
+    $ php bin/console ux:install dialog
+
+The command asks which kit to take the recipe from, copies its files into
+your application and lists the dependencies the recipe needs.
+
+The files land where they are meant to be used: Twig templates under
+``templates/``, Stimulus controllers under ``assets/``, and if a file already
+exists, it is not replaced without asking you first, so a recipe you have
+already customized cannot be overwritten by accident.
+
+A recipe can depend on other recipes and on packages, and the two are handled
+differently:
+
+* Other recipes it depends on are copied in the same run: installing ``dialog``
+  from the Shadcn kit also writes ``button``;
+* Composer and npm packages are not installed for you: the command prints the
+  ``composer require``, ``npm install`` or ``php bin/console importmap:require``
+  command to run yourself.
+
+.. tip::
+
+    Run ``php bin/console ux:install`` without any argument and the command
+    guides you: it asks for the kit, then for the recipe. Recipe names are
+    matched case-insensitively, and when a name does not exist, the command
+    suggests the closest ones it knows.
+
+Choosing a Kit
+--------------
+
+Four kits are shipped with the Toolkit:
+
+* ``shadcn``, `Shadcn UI <https://ux.symfony.com/toolkit/kits/shadcn>`_
+* ``flowbite-4``, `Flowbite v4 <https://ux.symfony.com/toolkit/kits/flowbite>`_
+* ``bootstrap``, `Bootstrap 5.3 <https://ux.symfony.com/toolkit/kits/bootstrap>`_
+* ``common``, `design-system agnostic recipes <https://ux.symfony.com/toolkit/kits/common>`_
+  for common Symfony needs, such as logout links
+
+Pass ``--kit`` to skip the question:
+
+.. code-block:: terminal
+
+    $ php bin/console ux:install button --kit=shadcn
+
+Browse the recipes of each kit, with a live preview, on
+`ux.symfony.com <https://ux.symfony.com/toolkit>`_.
+
+The Recipe Is Yours, For Worse and For Better
+---------------------------------------------
+
+Installing a recipe copies its files into your application, and the Toolkit
+then steps out of the way. The code is part of your project, like any template
+you wrote yourself.
+
+The consequence is that there is no upgrade path. Fixes, new options and new
+features added to a kit never reach a recipe you have already installed, and
+installing the same recipe today and in six months gives two different
+versions of it.
+
+Your version control system is what closes that gap. Commit right after
+installing a recipe, and when you want the newer version, run ``ux:install``
+again and read the diff: it shows exactly what changed in the kit and what you
+have changed yourself, and you keep what you want from each side. The command
+asks before overwriting, so nothing is lost if you would rather not.
+
+In exchange, the recipe really is yours. Adapting one is editing a template:
+no override mechanism to learn, no configuration to bend, and no limit to the
+options its author thought of. That is the idea the Toolkit takes from
+`Shadcn UI <https://ui.shadcn.com/>`_, and the reason the files are copied
+rather than shipped in ``vendor/``.
+
+.. _`the Symfony UX initiative`: https://ux.symfony.com/


src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,128 @@
...
+``--destination`` (``-d``)
+    Where to copy the files. Defaults to the current directory, which is what
+    you want when running the command from the root of your project.
+
+``--force`` (``-f``)
+    Overwrite files that already exist. Without it, the command asks before
+    replacing anything, so a component you have already customized is never
+    silently overwritten.
+
+Making a Component Yours
```
javiereguiluz: In my opinion, the way the Toolkit works has:

* **One big advantage**: files are copied into your project. They are yours. You can make any change or tweak. You are not limited by the default options given by the Twig component. You can add new ones or change existing ones.
* **One big drawback**: files are copied into your project. Next time you install the same component, you can get totally different files. There's no way to get the new options, features and bug fixes of Toolkit components. You must copy+paste the files yourself manually and find the differences. If you install the same component in different projects, you'll get different versions of it.

This section explains the advantage well, but I think that the drawback is not properly explained.

Amoifr:
You are right, and the imbalance was worse than the section you commented on: the introduction had the same one-sided framing. It ended with "an upgrade of the Toolkit never changes a component you already installed", which reads as reassurance while describing exactly the drawback you are pointing at.

Both are rewritten. The section is now named "The Component Is Yours, For Better and For Worse" and says the two halves plainly:

- files are yours to edit, without an override mechanism and without being limited to the options the author thought of;
- there is no upgrade path. Fixes, new options and new features added to a kit never reach a component already installed. Reinstalling copies the current files, which may differ both from what you started with and from what you have since edited, so bringing an improvement in means comparing and merging by hand. The command asks before overwriting, so nothing is lost, but the merge is yours to do.
- and the consequence across projects: installing a component today and in six months gives two different versions, and they do not drift back together.

The introduction now points forward to that section instead of selling one half of the trade, since it is the thing to understand before adopting the Toolkit rather than a detail near the end.

DOCtor-RST caught a title underline one character short while I was at it. Green now.

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+
+.. code-block:: terminal
+
+    $ composer require twig/extra-bundle symfony/ux-twig-component
+
+    $ npm install <package>
+    # ...or, with AssetMapper
+    $ php bin/console importmap:require <package>
+
+Components of the official kits are styled with `Tailwind CSS`_, so your
```
smnandre: Not all :)  #bootstrap

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+Installing the Dependencies of a Component
+------------------------------------------
+
+A component may need Composer packages, front-end packages, or both. The
+Toolkit does not install them for you: it reports what is missing at the end of
+the installation, so you stay in control of your dependencies.
+
+.. code-block:: terminal
+
+    $ composer require twig/extra-bundle symfony/ux-twig-component
```
smnandre: Here we should show the initial command, because reader may think they need to run this install command

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+----------------------
+
+Run the ``ux:install`` command with the name of the component you want:
+
+.. code-block:: terminal
+
+    $ php bin/console ux:install button
+
+The command asks which kit to take the component from, copies its files into
+your application and lists the dependencies the component needs.
```
smnandre: This needs more explanation I think. Where ? how ? To bring some form of safety feelings to a user not certain of the impact this command would have on its stack

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+the command suggests the closest ones it knows.
+
+Choosing a Kit
+--------------
+
+A *kit* is a collection of components sharing one design. Two official kits are
+shipped with the Toolkit:
+
+* ``shadcn``, `Shadcn UI <https://ux.symfony.com/toolkit/kits/shadcn>`_
+* ``flowbite-4``, `Flowbite v4 <https://ux.symfony.com/toolkit/kits/flowbite>`_
```
smnandre, line 79: Bootstrap 👼

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Components of the official kits are styled with `Tailwind CSS`_, so your
+application needs it for them to look right.
+
+Options
+-------
+
+``--kit`` (``-k``)
+    The kit to install from: an official kit name, or the URL of a kit
+    repository, optionally suffixed with ``:branch``.
```
smnandre: Let's give examples here in the following lines, what do you think ?

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+
+``--destination`` (``-d``)
+    Where to copy the files. Defaults to the current directory, which is what
+    you want when running the command from the root of your project.
+
+``--force`` (``-f``)
+    Overwrite files that already exist. Without it, the command asks before
+    replacing anything, so a component you have already customized is never
```
smnandre: I'd rewrite this a second time in an admonition bloc ( tip ? ), something in that spirit (and in rst) :)  ? 

> [!TIP]
> Components previously installed are never silently overwritten.

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+    silently overwritten.
```
Kocal: This section is basically useless, letting people run the command with `--help` is enough

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+Symfony UX Toolkit gives you ready-to-use UI components for your Symfony
+application. It is part of `the Symfony UX initiative`_.
+
+It is not a component library that you depend on at runtime. The
+``ux:install`` command copies the source of a component into your project, and
+from there the code is yours: you read it, edit it and commit it like any other
+template of your application. Nothing is hidden in ``vendor/``, and nothing
+updates it for you either. That trade is the main thing to understand before
+adopting the Toolkit, and it is described in `The Component Is Yours, For
+Better and For Worse`_.
```
Kocal: Maybe let's make 2 paragraphs

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+------------
+
+.. code-block:: terminal
+
+    $ composer require --dev symfony/ux-toolkit
+
+The package only registers console commands, so it is needed while you install
+components, not to render them.
+
+Installing a Component
```
Kocal:
```suggestion
Installing a Recipe
```

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+----------------------
+
+Run the ``ux:install`` command with the name of the component you want:
```
Kocal:
```suggestion
Run the ``ux:install`` command with the name of the recipe you want:
```

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+
+.. code-block:: terminal
+
+    $ php bin/console ux:install button
+
+The command asks which kit to take the component from, copies its files into
```
Kocal:
```suggestion
The command asks which kit to take the recipe from, copies its files into
```

etc...

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+
+    $ php bin/console ux:install
+
+Component names are matched case-insensitively, and when a name does not exist
+the command suggests the closest ones it knows.
+
+Choosing a Kit
+--------------
+
+A *kit* is a collection of components sharing one design. Two official kits are
```
Kocal: No, it's a collection of recipes:

```suggestion
A *kit* is a collection of recipes sharing one design. Two official kits are
```

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+.. code-block:: terminal
+
+    # from an official kit
+    $ php bin/console ux:install button --kit=shadcn
+
+    # from a kit hosted on GitHub
+    $ php bin/console ux:install button --kit=https://github.com/user/my-kit
+
+    # from a specific branch of that kit
+    $ php bin/console ux:install button --kit=https://github.com/user/my-kit:branch
```
Kocal: I don't want these usages to be documented since we don't improve the DX for people creating their own kits.

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+.. code-block:: terminal
+
+    $ composer require twig/extra-bundle symfony/ux-twig-component
+
+    $ npm install <package>
+    # ...or, with AssetMapper
+    $ php bin/console importmap:require <package>
+
+Components of the official kits are styled with `Tailwind CSS`_, so your
+application needs it for them to look right.
```
Kocal: To me this section is also useless, the website and `ux:install` already shows you what command you need to run to install recipe dependencies

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+gives two different versions of it, and they will not drift back together on
+their own.
+
+Creating Your Own Kit
+---------------------
+
+A kit is a directory with a ``manifest.json`` and one directory per component.
+If you want to build one, for your company design system for instance, the
+`Toolkit contributing guide <https://github.com/symfony/ux/blob/3.x/src/Toolkit/CONTRIBUTING.md>`_
+describes the format and the tooling.
```
Kocal: Same than before, this section is basically useless since we don't improve the DX for people wanting to create their own kit, that's why it's in `WIP` here https://ux.symfony.com/toolkit#kits

Also, a kit can be created with `sfp vendor/bin/ux-toolkit-kit-create`

Kocal (changes requested):

src/Toolkit/doc/index.rst
Kocal: This should be deleted

src/Toolkit/doc/index.rst
Kocal: This should be deleted

src/Toolkit/doc/index.rst
Kocal: This should be deleted

Kocal:
Please for the next time let's discuss a bit on the issue https://github.com/symfony/ux/issues/3824 before jumping on a PR

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,141 @@
...
+``--destination`` (``-d``)
+    Where to copy the files. Defaults to the current directory, which is what
+    you want when running the command from the root of your project.
+
+``--force`` (``-f``)
+    Overwrite files that already exist. Without it, the command asks before
+    replacing anything, so a component you have already customized is never
+    silently overwritten.
+
+The Component Is Yours, For Better and For Worse
```
smnandre:
```suggestion
The Component Is Yours, For Worse and For Better
```

I'd inverse the logic.

- Here is how it work
- The drawbacks, justified or limited 
- The positive aspecrs

It allows to end on a positive note, and is in fact really a foundation for this package.

And we keep the node to shadcn :) 

wdyt ?

smnandre:
Thank you for starting this work @Amoifr !

This will help illustrate the various goals / messages we'd like to send and how we can help with these ux kits and features.

Let's work here (or on the issue pointed at by @Kocal), I have the feeling this will end with a great documentation page!

src/Toolkit/doc/index.rst
smnandre: I share your point of view here, @javiereguiluz and I do think potential should be aware of the pro/cons. 

But then, i'm thinking we could not "just" frighten them but here there is a pretty easy solution for the "update/difference" thing.

Something like: "use your favorite git / versionning software, run the install again and then leverage git diff" 

(etc.. but with a good english haha)

Amoifr:
Reworked, thanks @Kocal @smnandre @javiereguiluz.

Terminology first: the whole page said "component" where the code says "recipe". Fixed everywhere, headings included.

Deleted, per @Kocal: the dependencies section, the options reference, the kit authoring section, and the external/branch `--kit` examples.

@smnandre, three of your suggestions land on lines @Kocal asked to delete (the install command, the examples, the tip). I followed the deletions, but kept what you were after in the install section itself: it now says where the files land (`templates/`, `assets/`), that a recipe can bring another one along, and that an existing file is never replaced without asking.

The kit list now has all four, Bootstrap and Common included, and the Tailwind claim went away with the section it was in.

The last section is inverted as you suggested: how it works, then the drawback, then the good side, ending on it. The upgrade gap has your answer now rather than just a warning: commit after installing, re-run `ux:install`, read the diff. The Shadcn nod stayed.

Noted for the issue-first process, @Kocal. Sorry for jumping.

DOCtor-RST green.

src/Toolkit/doc/index.rst
Amoifr: The drawback now has an answer rather than only a warning, following @smnandre: commit right after installing, run `ux:install` again when you want the newer version, and read the diff. The section is also inverted so it ends on the good side.

src/Toolkit/doc/index.rst
Amoifr: Right, that was wrong for Bootstrap. The claim is gone, along with the section it was in, which @Kocal asked to delete.

src/Toolkit/doc/index.rst
Amoifr: I followed @Kocal here and deleted the section, so the initial command is not shown again.

src/Toolkit/doc/index.rst
Amoifr: Added. It now says where the files land (`templates/`, `assets/`), that installing `button` from Shadcn writes a single `templates/components/Button.html.twig`, that a recipe can bring another one along like `dialog` does with `button`, and that nothing is written outside the destination directory.

src/Toolkit/doc/index.rst
Amoifr: All four are listed now: `shadcn`, `flowbite-4`, `bootstrap` and `common`.

src/Toolkit/doc/index.rst
Amoifr: Deleted, per @Kocal.

src/Toolkit/doc/index.rst
Amoifr: The section is deleted per @Kocal, so there is no admonition. I kept what you were after in the install section itself: a file that already exists is not replaced without asking first, so a recipe you have customized cannot be overwritten by accident.

src/Toolkit/doc/index.rst
Amoifr: Removed.

src/Toolkit/doc/index.rst
Amoifr: Split in two.

src/Toolkit/doc/index.rst
Amoifr: Applied.

src/Toolkit/doc/index.rst
Amoifr: Applied.

src/Toolkit/doc/index.rst
Amoifr: Applied, and the "etc..." too: the page said "component" everywhere the code says "recipe", so it is renamed throughout, headings included.

src/Toolkit/doc/index.rst
Amoifr: Applied, with one deviation: "Four kits are shipped with the Toolkit", not two. `bootstrap` and `common` were missing, which is what @smnandre pointed at above.

src/Toolkit/doc/index.rst
Amoifr: Removed, only `--kit=shadcn` is left.

src/Toolkit/doc/index.rst
Amoifr: Removed.

src/Toolkit/doc/index.rst
Amoifr: Removed. Noted for `vendor/bin/ux-toolkit-kit-create`, it is not documented anywhere in the page now.

src/Toolkit/doc/index.rst
Amoifr: Applied, as "The Recipe Is Yours, For Worse and For Better", and the logic is inverted the way you describe: how it works, the drawback with the git answer, then the good side to end on. The Shadcn nod stayed.

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,103 @@
...
+The command asks which kit to take the recipe from, copies its files into
+your application and lists the dependencies the recipe needs.
+
+The files land where they are meant to be used: Twig templates under
+``templates/``, Stimulus controllers under ``assets/``. Installing ``button``
+from the Shadcn kit writes a single ``templates/components/Button.html.twig``,
+and a recipe that builds on another one, like ``dialog`` on ``button``, brings
+it along. Nothing is ever written outside the destination directory, and a file
+that already exists is not replaced without asking you first, so a recipe you
+have already customized cannot be overwritten by accident.
```
Kocal: I'm not sure about this paragraph... I feel like, either it's useless, or too precise, or not 100% right, or _hard_ to read... but I can't put it into words :/ 

Maybe I would go for using `php bin/console ux:install dialog` just before, and modify/re-organizing the paragraph like that:

```suggestion
The files land where they are meant to be used: Twig templates under
``templates/``, Stimulus controllers under ``assets/``, and if a file already exists, it's not replaced without asking you first, so a recipe you have already customized cannot be overwritten by accident.

A recipe can depend on other recipes and on packages, and the two are handled differently:
  * Other recipes it depends on are copied in the same run: installing ``dialog`` from the Shadcn kit also writes ``button``.
  * Composer and npm packages are not installed for you: the command prints the ``composer require``, ``npm install``, or ``php bin/console importmap:require`` command to run yourself.
```

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Both arguments are optional, so you can also let the command guide you:
+
+.. code-block:: terminal
+
+    $ php bin/console ux:install
```
Kocal: > Both arguments are optional

I'm not sure about what both arguments you're speaking about, but these lines can be removed IMHO

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Recipe names are matched case-insensitively, and when a name does not exist the
+command suggests the closest ones it knows.
```
Kocal: Maybe in a `..tip :` or `.. info:` ? I don't remember the syntax

src/Toolkit/doc/index.rst
```diff
@@ -0,0 +1,103 @@
...
+Installation
+------------
+
+.. code-block:: terminal
+
+    $ composer require --dev symfony/ux-toolkit
+
+The package only registers console commands, so it is needed while you install
+recipes, not to render them.
+
```
Kocal, line 28: It was clear to me, but it wasn't clear to others, so I suggest adding a new section to explain what a kit is and what recipes are:
- A kit is a kit, a bunch of recipes for a given "framework", even if it's not true for `common` since it's framework-agnostic... can be found at https://github.com/symfony/ux-toolkit/tree/3.x/kits
- a recipe, is just a "pack of files and dependencies", Twig components, Stimulus Controller, Composer/npm dependencies

Kocal (commented):
Sorry if I was too "harsh" in reviewing this PR, I'm still struggling a bit with "change in general," and I was conviced that a symfony.com doc for UX Toolkit was absolutely useless, but... you there all have shown me the opposite. 😄 

Here are some further comments that I think should help improve the documentation even more.

Thanks for working on this!

src/Toolkit/doc/index.rst
Amoifr: Added as a new "Kits and Recipes" section, right before the install steps: a recipe is a pack of files and dependencies (Twig components, Stimulus controllers, Composer/npm packages), a kit is a collection of recipes sharing one design system, with a link to the kits directory. The `common` bullet in "Choosing a Kit" already reads "design-system agnostic recipes", so the exception stays visible without weighing down the definition.

src/Toolkit/doc/index.rst
Amoifr: Applied your suggestion almost as is: the example above is now `ux:install dialog`, the overwrite guard joined the "where the files land" sentence, and the recipe/package split became a two-item list.

One thing I dropped from the old paragraph: "nothing is ever written outside the destination directory", which came from @smnandre's earlier ask for a bit of reassurance. Happy to bring it back if you two want it.

src/Toolkit/doc/index.rst
Amoifr: You are right, and it was factually wrong on top of that: `recipe` is an argument, `--kit` is an option. Removed, code block included. The "run it with no argument and let it guide you" part moved down into the tip, where it costs a single line.

src/Toolkit/doc/index.rst
Amoifr: Done, `.. tip::` is the one. It now also carries the no-argument run from the block just above.

Kocal (approved):
I kind of like it, thanks everyone who made this possible! :) 

Next step on https://ux.symfony.com/toolkit:
- remove the `See the Kits` button for the classical `Documentation` button
- at the page footer, replace the `Documentation` link to the actual symfony.com documentation

Kocal:
Thank you @Amoifr.

PR #3827, opened 2026-09-01
[TwigComponent] Treat a prop explicitly passed as `null` as defined

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

Found at work during a Twig template refactorisation, but having a component `EditorialPartant` like that:
```twig
{# @prop chevalCourse Cheval_Course Le partant commenté #}
{# @prop texte string|null Le contenu éditorial (HTML), tel que saisi en back-office #}
{%- props chevalCourse, texte -%}
{# ... #}
```

Calling `<twig:EditorialPartant chevalCourse="{{ chevalCourse }}" texte="{{ texte }}" />` where `texte` is nullable and so, can be `null`, throws the following exception:
```
Uncaught PHP Exception Twig\Error\RuntimeError:
"texte should be defined for component components/EditorialPartant.html.twig
in "components/EditorialPartant.html.twig" at line 3."
at EditorialPartant.html.twig line 3
```

It's because we use `isset()` in `if (!isset($context[\''.$name.'\'])) {` , `isset()` will returns `false` even if the property (with `null` value) is passed.  It's not the case of `array_key_exists`, which correctly evaluate if a prop exists or not, whatever its value.

I also modified the exception message, the file name and line number are redundant.

### DIFF
--- a/src/TwigComponent/src/Twig/PropsNode.php
+++ b/src/TwigComponent/src/Twig/PropsNode.php
@@ -58,7 +58,7 @@ class PropsNode extends Node
 
         foreach ($this->getAttribute('names') as $name) {
             $compiler
-                ->write('if (isset($context[\'__props\'][\''.$name.'\'])) {')
+                ->write('if (array_key_exists(\''.$name.'\', $context[\'__props\'] ?? [])) {')
                 ->raw("\n")
                 ->indent()
                 ->write('$componentClass = isset($context[\'this\']) ? get_debug_type($context[\'this\']) : "";')
@@ -70,13 +70,13 @@ class PropsNode extends Node
                 ->raw("\n")
             ;
 
-            $compiler->write('if (!isset($context[\''.$name.'\'])) {');
+            $compiler->write('if (!array_key_exists(\''.$name.'\', $context)) {');
 
             if (!$this->hasNode($name)) {
                 $compiler
                     ->write("\n")
                     ->indent()
-                    ->write('throw new \Twig\Error\RuntimeError("'.$name.' should be defined for component '.$this->getTemplateName().'.");')
+                    ->write('throw new \Twig\Error\RuntimeError(\'Prop "'.$name.'" should be defined.\');')
                     ->write("\n")
                     ->outdent()
                     ->write('}')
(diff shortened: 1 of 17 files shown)


Kocal:
Oh, it looks like that some UX Toolkit snapshots are impacted.
Well it's fine, it means that we actually found and fixed more issues.

The following example:
```twig
<twig:Navbar :background="null" theme="light" style="background-color: #e3f2fd;">
    <div class="container-fluid">
        <a class="navbar-brand" href="#">Navbar</a>
        <div class="navbar-nav flex-row gap-3">
            <a class="nav-link active" aria-current="page" href="#">Home</a>
            <a class="nav-link" href="#">Features</a>
            <a class="nav-link" href="#">Pricing</a>
        </div>
    </div>
</twig:Navbar>
```

with an explicit `:background="null"`, should have never been rendered like:
```html
<nav class="navbar bg-body-tertiary" data-bs-theme="light" style="background-color: #e3f2fd;">
```

but like that instead:
```html
<nav class="navbar" data-bs-theme="light" style="background-color: #e3f2fd;">
```

smnandre:
```
{%- props chevalCourse, texte = null -%}
```

does not work ?

Kocal:
It does, but using `=` should only defines the default value when the prop is not passed.
I mean, its only tell you if the prop is required or not.

src/TwigComponent/src/Twig/PropsNode.php
```diff
@@ -70,13 +70,13 @@ public function compile(Compiler $compiler): void
...
 
-            $compiler->write('if (!isset($context[\''.$name.'\'])) {');
+            $compiler->write('if (!array_key_exists(\''.$name.'\', $context)) {');
 
             if (!$this->hasNode($name)) {
                 $compiler
                     ->write("\n")
                     ->indent()
-                    ->write('throw new \Twig\Error\RuntimeError("'.$name.' should be defined for component '.$this->getTemplateName().'.");')
+                    ->write('throw new \Twig\Error\RuntimeError(\'Prop "'.$name.'" should be defined.\');')
```
smnandre, line 79: Why the removal of the template ?

smnandre (approved):
I had a tiny doubt related to partial isolation, class based and externals ... 
... but your tests look like they cover it all!

src/TwigComponent/src/Twig/PropsNode.php
Kocal: Explained in PR's description, the exception message is verbose for nothing, the template name is repeated 3 times:  
```
Uncaught PHP Exception Twig\Error\RuntimeError:
"texte should be defined for component components/EditorialPartant.html.twig
in "components/EditorialPartant.html.twig" at line 3."
at EditorialPartant.html.twig line 3
```

PR #3549, opened 2026-05-18
 Add Symfony UX Image component (Slim v1)

KindImagination:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | 
| License       | MIT

## Summary

Slim v1 of `symfony/ux-image` — responsive image components addressing all feedback from the original PR #3188.

### What's Included

**`<twig:ux:img>`** — Simple responsive images
- Viewport-based width syntax (`100vw md:80vw`)
- Automatic srcset/sizes generation
- Density support (`densities="x1 x2"`)
- Fallback format for older browsers

**`<twig:ux:picture>`** — Art direction
- Breakpoint-specific aspect ratios (`sm:1:1 md:16:9`)
- Ratio cascading (CSS-like inheritance)
- Exclusive media queries for correct aspect ratio

### Key Changes from Original PR

| Original Issue | Resolution |
|---|---|
| `liip/imagine-bundle` in `require` | Removed — no hard dependencies |
| No local file support | `PassThroughProvider` works out of the box |
| PreloadManager controversy | Removed entirely |
| Arrays instead of objects | `ImageConfig` value object |
| Component name conflicts | `ux:img` / `ux:picture` |
| Only TwigComponent syntax | Also supports `{{ ux_image() }}` |
| Too many features | Slim scope — iterate on v1 |

### Provider Architecture

- **PassThroughProvider** (default) — works with local files, no CDN required
- **UrlPatternProvider** — generic CDN via URL template (Cloudflare, imgix, etc.)
- **Custom providers** — implement `ProviderInterface` and tag with `ux_image.provider`

### Usage

```twig
{# Simple responsive image #}
<twig:ux:img src="/images/hero.jpg" alt="Hero" width="100vw md:80vw" />

{# Art direction #}
<twig:ux:picture src="/images/banner.jpg" alt="Banner" width="100vw md:80vw" ratio="sm:1:1 md:16:9" />

{# Without Twig components #}
{{ ux_image('/images/hero.jpg', 'Hero', { width: '100vw md:80vw' }) }}
```

### Configuration

```yaml
# config/packages/ux_image.yaml
ux_image:
    default_provider: passthrough
    providers:
        cloudflare:
            pattern: '/cdn-cgi/image/w={width},f={format},q={quality}/{src}'
```

This addresses feedback from @WebMamba, @smnandre, and @Kocal on the original PR. Happy to iterate further!

PR #3838, opened 2026-09-05
[Image] Add the UX Image package and its Glide, Cloudflare and KeyCDN bridges

### LINKED ISSUE #3834: [RFC] Symfony UX Image
Hi 👋🏻

I believe it's time to open a RFC to describe what we really want for our new package UX Image, since we already have 4 PRs trying to implement it:
- https://github.com/symfony/ux/pull/3188 by @razbakov
- https://github.com/symfony/ux/pull/3549 by @KindImagination
- https://github.com/symfony/ux/pull/3765 by @seb-jean
- https://github.com/symfony/ux/pull/3768 by @smnandre

## Prior art

The JS ecosystem has shipped this for years. Next.js has [`<Image>`](https://nextjs.org/docs/app/getting-started/images) on top of [Vercel's image optimization](https://vercel.com/docs/image-optimization), Nuxt has [`<NuxtImg>`](https://image.nuxt.com/get-started/providers) with one pluggable provider per image service, and [unpic](https://unpic.pics/learn/) generalizes the idea into a framework-agnostic library. They all solve the same problem: write a single component and get an image sized for the viewport that asked for it, instead of shipping a 4000px hero to a phone. Symfony has no equivalent. UX Image follows Nuxt's provider design.

## Where we are

**#3188** (+5492, targets 2.x) is the oldest attempt: `<twig:img>` and `<twig:picture>` with viewport-based widths (`100vw md:80vw`), ratio cascading, focal point, preload injection, and `liip/imagine-bundle` as a hard requirement. It is unmaintained and never moved to 3.x. But the shape was right: the component declares what it wants, something else does the transformation.

**#3549** (+1793) is a slimmer redo that answered most of the feedback on #3188. It introduced the piece I care about: a provider abstraction, with `ProviderInterface`, a `ux_image.provider` tag, a `default_provider` option, a `PassThroughProvider` for local files and a `UrlPatternProvider` for a generic CDN URL template. This is very close to what I would like. What is missing is a provider that actually transforms anything locally, since `PassThroughProvider` returns the URL untouched.

**#3765** (+1710) adds `<twig:ux:img>`, `<twig:ux:picture>` and `<twig:ux:source>`, plus `ux_img()` and `ux_picture()`. Validation is careful and HTML-spec accurate: no mixed `w`/`x` descriptors, `sizes="auto"` requires `loading="lazy"`, enum attributes checked. It renders and validates markup the caller has already built by hand. There is no transformation, so it does not answer the reason people would install an image package.

**#3768** (+20341 across 100 files) goes all the way in the other direction: inspect, resize and encode with GD or Intervention, publish variants to local or Flysystem storage, persist an immutable `ImageAsset`, render `<picture>` from that metadata. It is thorough and the test suite is serious. But 27% of `src/` reimplements what Intervention Image already ships, and the CDN builders only decorate the paths of variants that were already generated locally. There is no URL-only mode: a Cloudinary setup still runs the full GD pipeline first, then asks Cloudinary to re-transform a derivative.

## Scope

Storing the file is not our job. The application owns the upload, wherever it comes from (Symfony Forms, UX Upload, a console import, an existing bucket) and hands us a path. UX Image turns that path plus a requested size into URLs, and renders the markup around them. Nothing else.

That already removes storage backends, write sessions, Doctrine types and regeneration commands from the package.

The images worth optimizing are the ones the application doesn't control: uploads in a blog post, avatars on a forum, a product catalogue. Their size, format and weight come from whoever uploaded them.

Images sitting in `assets/images/` are the opposite: known at build time, already sized by the developer, and already versioned by AssetMapper, Encore, or Reprise ✨ . A plain `` is the better answer there. Icons belong to [UX Icons](https://ux.symfony.com/icons), not here.

## What the package should provide

The value of UX Image is the providers, not the pipeline.

Every image service already transforms by URL, and they all take the same handful of parameters:

| Provider | Example |
| --- | --- |
| [Glide](https://glide.thephpleague.com/) | `/img/hero.jpg?w=800&h=600&fit=crop&fm=webp&q=85` |
| [Cloudflare Images](https://developers.cloudflare.com/images/transform-images/transform-via-url/) | `/cdn-cgi/image/width=800,height=600,fit=cover,format=auto,quality=85/hero.jpg` |
| [KeyCDN](https://www.keycdn.com/support/image-processing) | `/hero.jpg?width=800&height=600&fit=cover&format=webp&quality=85` |

Different spelling, same tuple: `(path, width, height, fit, format, quality)`. A provider maps that tuple to a URL: roughly 60 lines each.

I would ship these three in v1: Glide for local, Cloudflare Images and KeyCDN to prove the contract holds across two different URL syntaxes. Cloudinary, imgix, Vercel, Netlify and the rest can live in userland behind `ProviderInterface`, and we add the ones people actually ask for.

Two reasons to go this way:

- **Less code.** No decoding, no encoding, no EXIF handling, no codec capability matrix. That code already exists, is maintained by someone else, and better tested than anything we would write.
- **Better performance.** Nothing to process when a file is uploaded, and changing a breakpoint changes a URL instead of triggering a regeneration batch over the whole database.

## URL-only does not mean CDN-only

This is the part I want to make explicit, because "URL-based" usually reads as "you now need a paid CDN".

[Glide](https://glide.thephpleague.com/) is a League package that is exactly a URL-driven image server: `?w=800&h=600&fit=crop&fm=webp&q=85`, with signed URLs and its own cache. It is the local provider, and it satisfies the same contract as the hosted ones.

That symmetry has a price, and the RFC should say so rather than let someone discover it in review. Glide is the only one of the three that runs inside your application instead of someone else's, so its bridge ships a controller, a route and a response adapter, and pulls in a real image library transitively (Intervention Image 4, via Glide 4). It also exposes a public resize endpoint: without a signing key the server actually validates, anyone can request `?w=1`, `?w=2`, `?w=3` and mint a fresh cache entry every time. Glide has signing for exactly that reason; a bridge that generates signatures without checking them is worse than no signing at all, because it reads as protection.

None of that changes the contract. Same `(path, width, format, quality, fit)` tuple, same `ProviderInterface`, same env var to switch. But the local option is a small server, not a URL template, and pretending otherwise would make the "no CDN account required" claim thinner than it is.

So the provider becomes a configuration choice, switchable per environment through an env var:

```yaml
# config/packages/ux_image.yaml
ux_image:
    provider: '%env(UX_IMAGE_PROVIDER)%'   # glide locally, cloudflare in prod
    providers:
        glide:
            source: '%kernel.project_dir%/public/uploads'
            cache: '%kernel.cache_dir%/ux_image'
        cloudflare:
            base_url: 'https://example.com'
```

No application code changes between dev and prod, and no CDN account required to get started.

## Rendering

The provider gives the URLs, TwigComponent gives the attributes:

```twig
<twig:ux:image src="/uploads/hero.jpg" alt="Hero" sizes="(min-width: 64rem) 50vw, 100vw" class="rounded" />
```

The component asks the provider for one URL per requested width, builds `srcset` and `sizes`, and lets `ComponentAttributes` render everything else. We should not write a single `htmlspecialchars()` call: `twig/html-extra` already handles merging, escaping, `null`/`false`, `aria-*`, `data-*`, `style` and `BackedEnum`.

`` only covers the common case when the provider negotiates the format itself. Cloudflare does, through `format=auto`. Glide does not: `fm` accepts `jpg`, `pjpg`, `png`, `gif`, `webp`, `avif` and `heic`, with no `auto` value. KeyCDN does not either, and has no AVIF at all.

So the contract has to expose whether a provider negotiates, and the renderer emits a `<picture>` with one `<source type>` per format when it does not. For Glide we own the controller, so it can resolve the format from the `Accept` header and answer with `Vary: Accept`, which puts it back on par with Cloudflare.

`<picture>` is also what art direction needs, when the crop itself changes with the viewport. That part can wait.

## Open questions

**How do we express the requested widths?** I would rather not make people hand-write a list of pixel values on every component. [unpic](https://unpic.pics/) answers this with a `layout` prop and derives the candidates from it: `fixed` (exact `width` and `height`, 1x and 2x), `constrained` (a maximum `width`, scaling down on smaller screens) and `fullWidth` (viewport width, `height` only). The author describes the role of the image, the package picks the ladder and writes `sizes`. The alternative is an explicit `widths` list, configured globally with a per-component override. I lean towards the first, but have no strong opinion yet.

**Signed URLs.** Glide and most CDNs support them, but the key handling differs enough that it probably deserves its own contract rather than a `secret` option bolted onto every provider.

Thoughts?

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | yes 
| Issues         | Fix #3834 
| License        | MIT

Implements https://github.com/symfony/ux/issues/3834, replaces #3188, #3549, #3765, #3768. The RFC has the prior art and the reasoning; this is the implementation.

UX Image renders responsive images by delegating every transformation to a URL-driven provider, so the package itself never decodes, resizes or encodes an image. 
A DSN picks the provider at runtime, the way `symfony/ux-map` picks a renderer, which turns "local in dev, CDN in prod" into an environment variable rather than a code change.

```yaml
ux_image:
    provider: '%env(resolve:UX_IMAGE_DSN)%'
```

```twig
<twig:ux:image src="/uploads/hero.jpg" alt="Hero" width="800" height="450" />
```

## How it fits together

```mermaid
flowchart TD
    A["Template<br/>#60;twig:ux:image src alt width height layout /#62;"] --> B["Image component"]
    B --> C["RenderOptions"]
    C --> D["ImageRenderer"]
    L["LayoutResolver<br/>resolutions, sizes, inline style"] --> D
    D -- "path, width, height, fit, format, quality" --> E["Provider"]
    S["ux_image.provider DSN"] --> R["ProviderResolver"]
    R --> F["Glide, Cloudflare or KeyCDN bridge"]
    F -- "builds" --> E
    E -- "one URL per breakpoint and per format" --> G["RenderedImage"]
    G --> H["Image.html.twig"]
    H -- "provider negotiates the format" --> I["#60;img#62; + srcset + sizes"]
    H -- "it does not" --> J["#60;picture#62; + one #60;source#62; per format"]
```

`<twig:ux:image>` (or the `ux_image()` Twig function) collects its props into a `RenderOptions` object and hands it to `ImageRenderer`. `LayoutResolver` turns the layout into a breakpoint ladder, a `sizes` attribute and the inline `style`. The renderer asks the provider for one URL per breakpoint and per format, and returns a `RenderedImage` the template prints as ``, or as `<picture>` when the provider can't negotiate the format on its own. `ComponentAttributes` and `twig/html-extra` write the attributes out, so the package has no `htmlspecialchars()` call of its own.

The provider itself is built once from the DSN: `ProviderResolver` walks the registered `ProviderFactoryInterface` services and the first one that recognizes the scheme wins. That factory is the whole surface a bridge has to implement.

Three bridges ship with it: `symfony/ux-glide-image` for local, no-CDN use, plus `symfony/ux-cloudflare-image` and `symfony/ux-keycdn-image` for CDN uses. 

---

Opened Recipes PR: https://github.com/symfony/recipes/pull/1571

### DIFF
--- a/splitsh.json
+++ b/splitsh.json
@@ -7,6 +7,10 @@
         "ux-chartjs": "src/Chartjs",
         "ux-cropperjs": "src/Cropperjs",
         "ux-icons": "src/Icons",
+        "ux-image": { "prefixes": [{ "from": "src/Image", "to": "", "excludes": ["src/Bridge"] }] },
+        "ux-cloudflare-image": "src/Image/src/Bridge/Cloudflare",
+        "ux-glide-image": "src/Image/src/Bridge/Glide",
+        "ux-keycdn-image": "src/Image/src/Bridge/KeyCdn",
         "ux-live-component": "src/LiveComponent",
         "ux-map": {
             "prefixes": [{ "from": "src/Map", "to": "", "excludes": ["src/Bridge"] }]
--- a/src/Image/.gitattributes
+++ b/src/Image/.gitattributes
@@ -1,0 +1,5 @@
+/.git* export-ignore
+/.symfony.bundle.yaml export-ignore
+/doc export-ignore
+/phpunit.dist.xml export-ignore
+/tests export-ignore
--- a/src/Image/.gitignore
+++ b/src/Image/.gitignore
@@ -1,0 +1,7 @@
+/composer.lock
+/phpunit.xml
+/vendor/
+/.phpunit.cache/
+/var/
+/config/reference.php
+/config/schema.json
--- a/src/Image/.symfony.bundle.yaml
+++ b/src/Image/.symfony.bundle.yaml
@@ -1,0 +1,3 @@
+branches: ['3.x']
+maintained_branches: ['3.x']
+doc_dir: 'doc'
--- a/src/Image/CHANGELOG.md
+++ b/src/Image/CHANGELOG.md
@@ -1,0 +1,5 @@
+# CHANGELOG
+
+## 3.6.0
+
+- Add the component
--- a/src/Image/LICENSE
+++ b/src/Image/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2024-present Fabien Potencier
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is furnished
+to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/src/Image/README.md
+++ b/src/Image/README.md
@@ -1,0 +1,16 @@
+# Symfony UX Image
+
+**EXPERIMENTAL** This component is currently experimental and is
+likely to change, or even change drastically.
+
+Symfony UX Image renders responsive images in Symfony applications by delegating transformations to a URL-based image provider, like Glide, KeyCDN or Cloudflare.
+
+**This repository is a READ-ONLY sub-tree split**. See
+https://github.com/symfony/ux to create issues or submit pull requests.
+
+## Resources
+
+- [Documentation](https://symfony.com/bundles/ux-image/current/index.html)
+- [Report issues](https://github.com/symfony/ux/issues) and
+  [send Pull Requests](https://github.com/symfony/ux/pulls)
+  in the [main Symfony UX repository](https://github.com/symfony/ux)
--- a/src/Image/composer.json
+++ b/src/Image/composer.json
@@ -1,0 +1,56 @@
+{
+    "name": "symfony/ux-image",
+    "type": "symfony-bundle",
+    "description": "Responsive images for Symfony, delivered through URL-based image providers",
+    "keywords": [
+        "symfony-ux",
+        "image",
+        "responsive",
+        "cdn"
+    ],
+    "homepage": "https://symfony.com",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Hugo Alliaume",
+            "email": "hugo@alliau.me"
+        },
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "autoload": {
+        "psr-4": {
+            "Symfony\\UX\\Image\\": "src/"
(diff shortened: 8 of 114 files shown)


Kocal:
(the CI fails because it needs Twig 3.29)

PR #3851, opened 2026-09-09
[Toolkit] Fix `ux:install` reporting the source path of the installed files

ker0x:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | no
| Issues         | -
| License        | MIT

The "Installed files" listing printed by `ux:install` is built from
`File::$sourceRelativePathName`, so it names the file *inside the kit* rather than the path it
was actually written to:

```php
$this->io->listing(array_map(static fn (File $file) => Path::join($destinationPath, $file->sourceRelativePathName), $installationReport->newFiles));
```

The two sides are independent by design — `File` carries a source *and* a destination, and a
recipe manifest maps them freely through `copy-files` — but every recipe shipped today maps
`"templates/": "templates/"`, so the two happen to be identical and the mistake is invisible.

A recipe that copies, say, `ui/` to `templates/components/` reports paths that do not exist on
disk:

```
Installed files
---------------

 * /path/to/app/ui/Notice.html.twig        ← nothing was written here
```

while the file actually landed in `/path/to/app/templates/components/Notice.html.twig`.

The fix reports `destinationRelativePathName`. The regression test installs a fixture kit whose
`copy-files` maps a source directory to a differently named destination, and asserts the
listing names the destination and not the source; it fails on `3.x` before the fix.

### DIFF
--- a/src/Toolkit/src/Command/InstallCommand.php
+++ b/src/Toolkit/src/Command/InstallCommand.php
@@ -197,7 +197,7 @@ class InstallCommand extends Command
         $this->io->success('The recipe has been installed.');
 
         $this->io->section('Installed files');
-        $this->io->listing(array_map(static fn (File $file) => Path::join($destinationPath, $file->sourceRelativePathName), $installationReport->newFiles));
+        $this->io->listing(array_map(static fn (File $file) => Path::join($destinationPath, $file->destinationRelativePathName), $installationReport->newFiles));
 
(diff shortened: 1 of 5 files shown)


smnandre (approved):

Kocal (approved):

Kocal:
Thank you @ker0x.

PR #3853, opened 2026-09-09
[Toolkit][Shadcn] Add carousel recipe

ker0x:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| License        | MIT

Adds the `carousel` recipe to the Shadcn kit.

```twig
<twig:Carousel orientation="horizontal | vertical" align="start | center | end" loop>
    <twig:Carousel:Content>
        <twig:Carousel:Item>...</twig:Carousel:Item>
        <twig:Carousel:Item>...</twig:Carousel:Item>
        <twig:Carousel:Item>...</twig:Carousel:Item>
    </twig:Carousel:Content>
    <twig:Carousel:Previous />
    <twig:Carousel:Next />
</twig:Carousel>
```

### Embla

Upstream shadcn builds `Carousel` on [Embla Carousel](https://www.embla-carousel.com/), an npm dependency. No Toolkit recipe ships one — every controller in the kit imports only `@hotwired/stimulus` — so the behaviour is reimplemented as a self-contained `carousel` controller: a flex track translated with a CSS transform, whose snap positions are **measured** from the rendered slides rather than configured, so any `basis-*` utility on a `Carousel:Item` works with no extra wiring, and they are re-measured on resize.

Consequences for the public API:

- Embla's `opts` become props on `Carousel`: `align` and `loop`.
- Embla's Autoplay **plugin** becomes an `autoplay` prop (delay in ms). The upstream "Plugins" section is therefore documented as **Autoplay**.
- Embla's `setApi` becomes a `carousel:select` event carrying `{ index, count }`. The recipe ships a small `carousel-display` controller that renders the current position from it, mirroring what `slider` does with `slider-display`.

### Notes

- Positions are expressed along the logical axis, so the same math drives horizontal, vertical and RTL carousels; only the sign of the applied transform changes. Directional classes are logical (`-ms-4`, `ps-4`, `-start-12`, `-end-12`), which also keeps the "Spacing" example overridable in both directions.
- `Carousel:Previous` renders `disabled` when the carousel does not loop, so the first slide has no enabled → disabled flash on load; the controller then syncs both buttons on every change.
- Arrow keys follow the carousel axis **and** the text direction (in RTL, <kbd>←</kbd> advances).
- A click that terminates a drag is swallowed, so dragging over a link or a button inside a slide does not activate it.

### Testing

Snapshots regenerated; `bin/ux-toolkit-kit-lint --fail-on-warning kits/shadcn`, `twig-cs-fixer`, `oxfmt` and `oxlint` are clean.

Behaviour driven in headless Chromium against a real Tailwind v4 build of the rendered examples, with no console errors:

| | Result |
|---|---|
| LTR / vertical / RTL stepping | correct transforms and snap counts per orientation |
| prev/next disabled sync | disabled at both ends, both enabled in between |
| arrow keys | follow the axis and the text direction |
| pointer drag | works on all three axes, with rubber-band and momentum snap |
| click after drag | swallowed |
| resize | re-measures and stays on the same slide |
| loop + autoplay | wraps, pauses on hover/focus, resets after interaction |
| disconnect/reconnect | no leaked timers or listeners |

[symfony-ux-toolkit-shadcn-carousel.webm](https://github.com/user-attachments/assets/074152a0-ac3c-422b-bf20-7c47060e95ff)

### DIFF
--- a/src/Toolkit/CHANGELOG.md
+++ b/src/Toolkit/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [Shadcn] Add `sheet` recipe
 - [Shadcn] Add `form` recipe
 - [Shadcn] Add `sidebar` recipe
+- [Shadcn] Add `carousel` recipe
 - [Shadcn][Flowbite] Merge component classes with `attributes.defaults({ class: '...'|tailwind_classes })` instead of `tailwind_merge` (consumer classes now override component variants)
 - [Shadcn] Fix Field/Label, InputGroup and Pagination silently dropping their own base classes when forwarding to a child component (their checked/disabled/nested/dark styles now apply)
 - [Shadcn] Fix `dialog` opening on initial render when `open` is `false`
--- a/src/Toolkit/kits/shadcn/carousel/README.md
+++ b/src/Toolkit/kits/shadcn/carousel/README.md
@@ -1,0 +1,261 @@
+# Carousel
+
+A carousel with motion and swipe, driven by a Stimulus controller.
+
+```twig {"preview":true,"height":"320px"}
+<twig:Carousel class="mx-auto w-full max-w-[12rem] sm:max-w-xs">
+    <twig:Carousel:Content>
+        {% for i in 1..5 %}
+            <twig:Carousel:Item>
+                <div class="p-1">
+                    <twig:Card>
+                        <twig:Card:Content class="flex aspect-square items-center justify-center p-6">
+                            <span class="text-4xl font-semibold">{{ i }}</span>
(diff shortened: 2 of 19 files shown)


Kocal:
Hi, and thanks for the contribution! 

I'm not a big fan with re-implementing the wheel for a carousel/slider component, it's hard to maintain, and Toolkit nicely support npm/importmap dependencies.

I'm re-opening a PR based on Embla.

PR #3866, opened 2026-09-12
[Inspector] Add a browser inspector for Symfony UX

smnandre:
| Q              | A   |
| -------------- | --- |
| Bug fix?       | no  |
| New feature?   | yes |
| Deprecations?  | no  |
| Documentation? | yes |
| Issues         | -   |
| License        | MIT |

Adds an experimental development tool to inspect Stimulus controllers,
LiveComponents, and Turbo elements directly on the page.

ux-inspector-1 - Grande

| . | . | . | 
| - | - | - |
| ux-inspector-2 - Grande | ux-inspector-3 - Grande | ux-inspector-4 - Grande | 

- **List** components and filter by framework or name.
- **Inspect** values, targets, actions, and relationships. Select components
  from the list or click them on the page.
- **Trace** events for the page or a component, including Live renders and
  Turbo requests. Expand events to view their details.

## Usage

```bash
composer require --dev symfony/ux-inspector
```

Register the bundle and its route in `dev`, 

Click the pull tab on the right edge of the screen to open the panel. It expands to show **UX** when the pointer
approaches or it receives keyboard focus.

You can also type `ux` anywhere (outside an editable field) to toggle the inspector (I love it on a daily usage!)

## Implementation

The bundle in `src/Inspector/src/` injects the Inspector into HTML responses and
serves its assets through a route. Both are disabled when `kernel.debug` is false.

The frontend in `src/Inspector/assets/src/` is a TypeScript custom element using
Shadow DOM to isolate the panel's DOM and styles from the application. It tracks
page DOM changes and stays mounted across Turbo visits.

Added 90 PHPUnit tests, 584 Vitest tests, and 26 Playwright scenarios across Chrome 
and Firefox.

The PR also updates the package build, unit-test runner, PHPStan CI, subtree
split configuration, and the E2E application's Stimulus bootstrap.

## Scope for v1

V1 only reads browser state and listens to events. Editing component state,
triggering actions, and inspecting or changing server-side state are out of scope.

## Documentation

Installation, usage, and configuration are covered in 
[the documentation](src/Inspector/doc/index.rst).

The Flex recipe is not available yet; registration steps are in 
[the documentation](src/Inspector/doc/index.rst#installation).

A companion website PR adds a demo and preview on ux.symfony.com: https://github.com/symfony/ux.symfony.com/pull/187


smnandre:
I opened a separated PR for the AssetMapper 8.2 problem

src/Inspector/assets/src/core/inspector-runtime.ts
```diff
@@ -0,0 +1,326 @@
...
+        this.#detector.observe();
+        this.#readyFrame = requestAnimationFrame(() => {
+            this.#readyFrame = null;
+            this.scan();
+            host.setAttribute('ready', '');
+        });
+    }
+
+    setStimulusApplication(application: StimulusApplicationLike | null): void {
+        (this.#registry.get('stimulus') as StimulusPlugin | undefined)?.setApplication(application);
```
squrious:
```suggestion
        this.#application = () => application;
```

Depending on the usage of this method, this could prevent the call to be no-op as the `scan()` method always override the plugin's application instance using `this.#application()`.

src/Inspector/assets/test/unit/inspector.test.js
```diff
@@ -0,0 +1,377 @@
...
+        expect(
+            connectStimulus({
+                getControllerForElementAndIdentifier: (element, identifier) =>
+                    element === target && identifier === 'search' ? instance : null,
+            })
+        ).toBe(true);
+        await vi.waitFor(() => expect(inspector.shadowRoot.querySelector('.component-row')).not.toBeNull());
+        inspector.shadowRoot.querySelector('.component-row').click();
+
+        expect(inspector.shadowRoot.querySelector('[data-group="stimulus-context"]')).toBeNull();
```
squrious: I don't get how these assertions test that runtime declarations are exposed, aren't we supposed to look for existing selectors instead of missing ones?

src/Inspector/assets/src/inspector-element.ts
```diff
@@ -0,0 +1,158 @@
...
+
+export function registerUXInspector(styles: StylesheetSource): void {
+    stylesheet = styles;
+    if (!customElements.get('ux-inspector')) customElements.define('ux-inspector', UXInspector);
+}
+
+export function connectStimulus(application: StimulusApplicationLike | null): boolean {
+    const valid = Boolean(application?.getControllerForElementAndIdentifier);
+    stimulusApplication = valid ? application : null;
+    (document.querySelector('ux-inspector') as UXInspector | null)?.setStimulusApplication(stimulusApplication);
```
squrious: `setStimulusApplication` delegates to runtime's `setStimulusApplication`, which: 
1. Overrides the plugin's application instance
2. Immediately `scan()`, calling the locator function to override the plugin's application instance again

If using the exposed `connectStimulus` this is harmless because of the global `stimulusApplication` module variable.

But I feel the two `setStimulusApplication` methods are not so useful, if the goal is only to update the scope of the locator function.

```ts
export function connectStimulus(application: StimulusApplicationLike | null): boolean {
    const valid = Boolean(application?.getControllerForElementAndIdentifier);
    stimulusApplication = valid ? application : null;
    (document.querySelector('ux-inspector') as UXInspector | null)?.scan();
    return valid;
}
```

Could do the job, and avoids to expose `UXInspector::setStimulusApplication` which wouldn't work properly is used outside `connectStimulus`.

src/Inspector/assets/src/ui/fields.ts
```diff
@@ -0,0 +1,258 @@
...
+const MAX_DISPLAYED_CLASSES = 2;
+
+export function makeField(key: string, value: unknown, options: FieldOptions = {}): HTMLElement {
+    value = SENSITIVE_KEY.test(key) ? '[redacted]' : value;
+    const structured = value !== null && typeof value === 'object';
+    const row = el('dd', { class: 'value' });
+    renderValue(row, value, options);
+    if (!options.multiline && !options.vertical && typeof value === 'string' && value.length > 32) expandableText(row);
+    if (!options.multiline && !options.vertical && typeof value === 'string' && value.length > 32)
+        row.dataset.long = '';
```
squrious:
```suggestion
    if (!options.multiline && !options.vertical && typeof value === 'string' && value.length > 32) {
       expandableText(row);
       row.dataset.long = '';
    }        
```

src/Inspector/assets/src/ui/fields.ts
```diff
@@ -0,0 +1,258 @@
...
+export function makeEmptyField(key: string): HTMLElement {
+    return el(
+        'div',
+        {
+            class: 'key-value',
+            dataset: { fieldKey: key, empty: '' },
+        },
+        String(key).length > 24
+            ? expandableText(el('dt', { class: 'key', text: key }))
+            : el('dt', { class: 'key', text: key }),
```
squrious: This construction is used in multiple places, maybe worth a helper in `ui/ui-helpers.ts`?

src/Inspector/assets/src/ui/detail-navigation.ts
```diff
@@ -0,0 +1,252 @@
...
+    #onDrillPop({ removed, current }: { removed: { id: string }; current: { element?: Element } }): void {
+        this.#destroyDrillDetail(removed.id);
+        this.restoreActivity();
+        this.#callbacks.change();
+        if (current.element) {
+            const dataMap = this.#state.get(current.element);
+            this.#callbacks.select({ element: current.element, framework: dataMap?.keys().next().value || 'default' });
+        } else {
+            this.#callbacks.clearSelection();
+            this.#callbacks.change();
```
squrious: Is this call necessary after the first one L.227?

src/Inspector/assets/src/ui/drill-stack.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    #syncNavigation(): void {
+        const currentIndex = this.#levels.length - 1;
+        const previousIndex = currentIndex - 1;
+        for (const [index, level] of this.#levels.entries()) {
+            const current = index === currentIndex;
+            level.header.classList.toggle('current', current);
+            level.header.classList.toggle('previous', index === previousIndex);
+            level.header.toggleAttribute('aria-current', current);
```
squrious: After `pop()` the current element gets its `aria-current` toggled on, with empty string as a value. Should it get the `page` value instead, as set in `push()`?

src/Inspector/assets/src/live-component/plugin.ts
```diff
@@ -0,0 +1,294 @@
...
+                }
+            }
+        }
+
+        return actions;
+    }
+
+    #readModelValue(element: Element, name: string): unknown {
+        const input = element as HTMLInputElement;
+        if (/password|secret|token|csrf/i.test(name) || input.type === 'password') return '[redacted]';
```
squrious: We could reuse `SENSITIVE_KEY` from `core/snapshot`? 
It is reused in `TreeViewer::#buildNode` as well

src/Inspector/assets/src/ui/component-list.ts
```diff
@@ -0,0 +1,166 @@
...
+        if (!target) return;
+        const related = (event as MouseEvent | FocusEvent).relatedTarget;
+        if (related instanceof Node && row!.contains(related)) return;
+        if (event.type === 'click') {
+            if (row!.matches('.component'))
+                this.element.dispatchEvent(new CustomEvent('drill-into', { detail: target }));
+            else {
+                target.element.scrollIntoView({ block: 'center', behavior: 'smooth' });
+                row!.setAttribute('aria-pressed', 'true');
+                this.#visual.onSelect?.(target);
```
squrious, line 141: This only applies to Turbo (page rules being implemented only in this plugin). I wonder if page rules really belong to the plugin interface?

When clicking on a Turbo page rule, the corresponding element is highlighted accordingly, but there is no way to deselect it. Once selected, users can only select other page rules, but never leave the selection mode.

`aria-pressed` attribute is never restored to `false` after the selection changes, leading to multiple buttons claiming a "pressed" state while they're not.

A possible UI enhancement: add a visual indicator on the selected page rule in the inspector panel. Currently there is no way to identify which rule is selected.

src/Inspector/assets/src/ui/timeline.ts
```diff
@@ -0,0 +1,595 @@
...
+        const diagnostic = this.#diagnostic(entry);
+        try {
+            await clipboard.writeText(JSON.stringify(diagnostic));
+            return true;
+        } catch {
+            return false;
+        }
+    }
+
+    addEntry(_entry: ActivityEntry): void {
```
squrious: Only usage is in inspector runtime, bound as a listener for new entries on event monitor, to schedule a flush on new events record if not already requested. It doesn't really "adds entries" to the timeline, because entries are fetched straight from the event monitor.

As the event monitor instance is already given in constructor, we could remove this method and set up the listener directly on construction: 
```ts
    constructor(monitor: EventMonitor, callbacks: TimelineCallbacks = {}) {
        this.#monitor = monitor;
        this.#monitor.addListener((entry) => {
            // The content of current addEntry() method
        });
    }
```

WDYT?

src/Inspector/assets/src/ui/drill-stack.ts
```diff
@@ -0,0 +1,122 @@
...
+        const previous = this.current;
+        if (previous) {
+            previous.scrollTop = this.#content.scrollTop;
+            previous.content.hidden = true;
+        }
+        const index = this.#levels.length;
+        const header = el(
+            'button',
+            {
+                class: 'stack-link current',
```
squrious: There is no CSS rule or query for `current` class

src/Inspector/assets/src/ui/drill-stack.ts
```diff
@@ -0,0 +1,122 @@
...
+        const label = root.header.querySelector('.stack-title');
+        if (label) label.textContent = title;
+    }
+
+    #syncNavigation(): void {
+        const currentIndex = this.#levels.length - 1;
+        const previousIndex = currentIndex - 1;
+        for (const [index, level] of this.#levels.entries()) {
+            const current = index === currentIndex;
+            level.header.classList.toggle('current', current);
```
squrious: Same as L.56

src/Inspector/assets/src/ui/panel.ts
```diff
@@ -0,0 +1,512 @@
...
+    }
+    setTargetModeActive(active: boolean): void {
+        const button = this.#actionButtons.target;
+        const label = active ? 'Stop inspecting' : 'Inspect page components';
+        button.classList.toggle('active', Boolean(active));
+        button.setAttribute('aria-pressed', String(Boolean(active)));
+        button.setAttribute('aria-label', label);
+        button.title = active ? 'Click to inspect. Shift-click to continue.' : label;
+        this.#monitorLabel.textContent = active ? 'Inspecting' : 'Watching';
+        this.#monitorLabel.parentElement?.classList.toggle('inspecting', Boolean(active));
```
squrious: No CSS rule/query for `inspecting` class (only in tests)

src/Inspector/assets/src/ui/panel.ts
```diff
@@ -0,0 +1,512 @@
...
+            { class: 'activity-tools toolbar--activity', hidden: true },
+            el(
+                'div',
+                {
+                    class: 'filter-list activity-filter-list',
+                    role: 'group',
+                    'aria-label': 'Filter activity by framework',
+                },
+                ...buttons
+            ),
```
squrious: `activity-filter-list` and `toolbar--activity` are not used for CSS styling, and only for querying in tests. 

We could remove them and use `activity-tools` instead in tests queries, WDYT?

src/Inspector/assets/src/core/inspector-runtime.ts
```diff
@@ -0,0 +1,326 @@
...
+            this.#timeline,
+            host,
+            this.#relationshipEngine,
+            this.#config.packages || {}
+        );
+        this.#panel.setActionCallbacks({
+            target: () => this.toggleTargetMode(),
+            overlay: () => this.toggleOverlay(),
+            pause: () => this.toggleLogPaused(),
+            clearLog: () => this.clearLog(),
```
squrious: I don't see these two last action callbacks being called anywhere in the panel component, which also explains why the `ICONS.pause` and `ICONS.clear` icons are still unused. 

Panel tests describe this exactly: `keeps pause and clear out of the visible Activity toolbar`, so I'm not sure if it's intended or not.

src/Inspector/assets/test/unit/live-component/plugin.test.js
```diff
@@ -0,0 +1,649 @@
...
+
+        it('returns "LiveComponent" when no name set', () => {
+            const el = document.createElement('div');
+            el.setAttribute('data-controller', 'live');
+            expect(plugin.getDisplayName(el)).toBe('LiveComponent');
+        });
+    });
+
+    describe('renderCard()', () => {
+        it('renders the expected data fields as a DocumentFragment', () => {
```
squrious, line 413: Not sure this test (and the same ones for Stimulus and Turbo) is relevant. 

It doesn't assert the returned result is a `DocumentFragment`, and makes no assertion about the "expected data fields". All of this is already asserted in other tests.

Also, the passed `data` is missing a few properties for `ComponentData<LiveData>`, without any impact on the result. Actually only the `.data` key is used for rendering in the three plugins.

Maybe `render<Live|Stimulus|Turbo>` should take the data to render directly?

src/Inspector/assets/src/ui/render-component.ts
```diff
@@ -0,0 +1,20 @@
...
+        case 'stimulus':
+            return renderStimulus(data as ComponentData<StimulusData>, context);
+        case 'livecomponent':
+            return renderLive(data as ComponentData<LiveData>, context);
+        case 'turbo':
+            return renderTurbo(data as ComponentData<TurboData>, context);
+        default:
+            return null;
+    }
+}
```
squrious: `name` and `data.type` look redundant. Based on my other comment, we could simplify to

```tst
export function renderComponent(data: ComponentData, context: RenderContext): Node | null {
    switch (data.type) {
        case 'stimulus': 
            return renderStimulus(data.data as StimulusData, context);
        // ...
    } 
}
```

Makes me also wonder if the render function should be exposed by the component itself. `ComponentData` only holds *values* now, but it's created by `parse()` factories exposed by plugins. Plugins own the render function (and can reuse UI factories from `ui/`). `renderComponent` resolves the "plugin => render function" mapping externally while it could be wired on component creation. WDYT?

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+            { getRelatedTo: () => [] },
+            {
+                stimulus: true,
+                livecomponent: false,
+                turbo: true,
+            }
+        );
+    });
+
+    it('has centered identity, navigation actions, and a minimal status footer', () => {
```
squrious, line 71: Some "is null" assertions on classes that never exist: 
- `.tabs`
- `.spacer`
- `.settings-panel`
- `.toolbar--components`

No CSS rule bound to them, test always pass until we decide to add new design rules/behaviors using them.

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+        expect(panel.element.querySelector('.tabs')).toBeNull();
+        expect(panel.element.textContent).not.toContain('Server');
```
squrious: Panel never renders "Server", I would remove this assertion

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+        expect(panel.element.querySelector('.settings-panel')).toBeNull();
+        expect(panel.element.querySelector('[aria-label="Hide inspector"]')).not.toBeNull();
+        expect(panel.element.querySelector('footer').textContent).not.toMatch(/components|events/);
+    });
+
+    it('puts search below the framework filters without a redundant rescan action', () => {
+        const filters = panel.element.querySelector('.filters');
+        expect(filters.firstElementChild.className).toBe('filter-list');
+        expect(filters.lastElementChild).toBe(panel.element.querySelector('[aria-label="Find a component"]'));
+        expect(filters.querySelector('[aria-label="Rescan components"]')).toBeNull();
```
squrious: `aria-label` is never set to "Rescan components"

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+        expect(panel.element.querySelector('.toolbar--components')).toBeNull();
+        expect(panel.element.querySelector('.toolbar--activity').hidden).toBe(true);
+        expect(panel.element.querySelectorAll('.activity-filter-list .filter')).toHaveLength(3);
+    });
+
+    it('mounts Activity as a single inline disclosure surface', () => {
+        expect(panel.element.querySelector('#panel-activity').firstElementChild).toBe(timeline.element);
+        expect(panel.element.querySelector('.split-view')).toBeNull();
```
squrious: `split-view` is never set, not styled, not queried

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+        await vi.waitFor(() => expect(panel.element.textContent).toContain('search'));
+        panel.drillInto(target);
+        await vi.waitFor(() => expect(panel.element.querySelector('.drawer')).not.toBeNull());
+
+        const drawer = panel.element.querySelector('.drawer');
+        expect(panel.element.querySelector('#panel-components').hidden).toBe(false);
+        expect(panel.element.querySelector('#panel-activity').hidden).toBe(true);
+        expect(timeline.element.parentElement).toBe(drawer);
+        expect(panel.element.querySelector('.activity-label').textContent).toBe('Activity');
+        expect(panel.element.querySelector('.detail__activity-link')).toBeNull();
```
squrious: `detail__activity-link` and `detail__activity-all` below are never set from code, never queried and never styled

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+        expect(panel.element.querySelector('#panel-components .stack-page').hidden).toBe(true);
+        expect(panel.element.querySelector('.filters').hidden).toBe(true);
+        expect(panel.element.querySelector('.component').classList.contains('selected')).toBe(true);
+        expect(panel.element.querySelector('.detail h2').textContent).toBe('search');
+        expect(panel.element.querySelector('.detail-selector').textContent).toBe('div#search');
+        expect(panel.drillBack()).toBe(true);
+        expect(panel.element.querySelector('.detail')).toBeNull();
+        expect(panel.element.querySelector('#panel-components .stack-page').hidden).toBe(false);
+        expect(panel.element.querySelector('.filters').hidden).toBe(false);
+        expect(panel.element.querySelector('.recent-components')).toBeNull();
```
squrious: `recent-components` class is never set from code, never styled and never queried

src/Inspector/assets/test/unit/ui/panel.test.js
```diff
@@ -0,0 +1,585 @@
...
+        state.clear();
+
+        const replacement = document.createElement('div');
+        replacement.id = 'search-after-navigation';
+        document.body.appendChild(replacement);
+        state.set(replacement, 'stimulus', { data: {} });
+        panel.drillInto(replacement);
+
+        expect(panel.element.querySelector('[data-group="stimulus-values"]').open).toBe(false);
+        expect(panel.element.querySelector('.component-events__search')).toBeNull();
```
squrious: `component-events__search` class is never set from code, never styled and never queried

src/Inspector/assets/test/unit/inspector.test.js
```diff
@@ -0,0 +1,377 @@
...
+        inspector?.remove();
+        document.body.replaceChildren();
+        vi.advanceTimersByTime(1000);
+        vi.useRealTimers();
+        vi.unstubAllGlobals();
+        connectStimulus(null);
+        vi.restoreAllMocks();
+    });
+
+    it('registers and renders the focused inspector shell', () => {
```
squrious, line 41: `slide-handle` class is never set from code, not styled and never queried.
`Server`/`Badges`/`Toasts` strings are never rendered
`aria-label` is never set to "Open settings"

src/Inspector/assets/test/unit/inspector.test.js
```diff
@@ -0,0 +1,377 @@
...
+    it('does not run ux shortcuts while typing', () => {
+        inspector.close();
+        const input = document.createElement('input');
+        document.body.appendChild(input);
+        input.dispatchEvent(new KeyboardEvent('keydown', { key: 'u', bubbles: true }));
+        input.dispatchEvent(new KeyboardEvent('keydown', { key: 'x', bubbles: true }));
+        expect(inspector.isOpen).toBe(false);
+    });
+
+    it('does not register Alt+I as an inspector shortcut', () => {
```
squrious: What's the purpose of this test? Avoid conflicts with browser's native shortcuts?

src/Inspector/assets/test/unit/core/external-dependencies.test.js
```diff
@@ -0,0 +1,127 @@
...
+        await vi.waitFor(() => expect(outlets(state, document.getElementById('owner-19'))).toEqual([target]));
+        expect(query.mock.calls.filter(([selector]) => selector === '.item')).toHaveLength(1);
+        parse.mockClear();
+        query.mockClear();
+        const updated = vi.fn();
+        state.addEventListener('page-updated', updated);
+        for (let i = 0; i < 50; i++) target.setAttribute('aria-label', String(i));
+        await vi.waitFor(() => expect(query.mock.calls.filter(([selector]) => selector === '.item')).toHaveLength(1));
+        expect(updated).not.toHaveBeenCalled();
+        expect(query.mock.calls.filter(([selector]) => selector === '.item')).toHaveLength(1);
```
squrious: Is this redundant with the assertion in `await vi.waitFor(...)` above?

squrious (commented):
Hi @smnandre! Just made a first review pass, hope this helps. 

Thanks for this amazing tool, I'm sure this will bring UX DX to the next stage. I also love how you made it dependency-free and self contained.

src/Inspector/assets/src/core/inspector-runtime.ts
smnandre: This was in the settings panel and the advanced usage panel that have been removed, sorry for the mess

src/Inspector/assets/test/unit/inspector.test.js
smnandre: The test now positively checks a rendered runtime default and a resolved target instead of asserting that unrelated selectors are absent.

src/Inspector/assets/src/inspector-element.ts
smnandre: Applied your suggestion: connectStimulus() updates the shared reference and requests a scan.

src/Inspector/assets/src/ui/fields.ts
smnandre: Applied: both operations now share the same condition.

src/Inspector/assets/src/ui/fields.ts
smnandre: Added expandableTextIfLong() in ui-helpers.ts and reused it for field labels and component headings

src/Inspector/assets/src/ui/detail-navigation.ts
smnandre: Removed the duplicate notification

src/Inspector/assets/src/ui/drill-stack.ts
smnandre: Fixed: the current header explicitly receives aria-current="page"; other headers have the attribute removed.

src/Inspector/assets/src/live-component/plugin.ts
smnandre: Reused SENSITIVE_KEY here while retaining the password-input check

src/Inspector/assets/src/ui/component-list.ts
smnandre: Fixed selection toggling, switching between rules, Escape, and cleanup on close, filtering, or removal. 

The selected rule now has a visual indicator and consistent aria-pressed state, with unit and browser coverage. 

Kept the optional plugin hook as framework-specific rule detection belongs to the adapter.

src/Inspector/assets/src/ui/timeline.ts
smnandre: Agreed. Timeline now subscribes directly to its monitor and unsubscribes in destroy().

src/Inspector/assets/src/ui/panel.ts
smnandre: Removed the unused current class

src/Inspector/assets/src/ui/drill-stack.ts
smnandre: Removed the unused current class

src/Inspector/assets/src/ui/drill-stack.ts
smnandre: Removed the unused current class

src/Inspector/assets/src/ui/panel.ts
smnandre: Removed

src/Inspector/assets/test/unit/live-component/plugin.test.js
smnandre: Simplified all three renderers to accept their framework data directly. Their tests now assert that the result is a DocumentFragment and check actual rendered field values.

src/Inspector/assets/src/ui/render-component.ts
smnandre: Removed the redundant name argument and simplified

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: Removed these obsolete assertions.

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: Removed.

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: emoved the obsolete selector assertion. The test still verifies where the Activity surface is mounted

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: Removed the assertion and updated the test name to describe the current search/filter layout

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: Removed both assertions

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: Removed both assertions

src/Inspector/assets/test/unit/ui/panel.test.js
smnandre: Removed.

src/Inspector/assets/test/unit/inspector.test.js
smnandre: Removed.Removed.

src/Inspector/assets/test/unit/inspector.test.js
smnandre: Removed.

smnandre:
Thank you @squrious !! What a review!

src/Inspector/assets/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "main": "dist/inspector.js",
+    "types": "dist/inspector.d.ts",
+    "exports": {
+        ".": "./dist/inspector.js"
+    },
+    "config": {
+        "browser_targets": [
+            "chrome145",
+            "firefox146"
+        ]
```
Kocal: Why? The current `es2021` target is not enough?

bin/build_package.ts
```diff
@@ -87,7 +90,7 @@ async function main() {
         unbundle: isStimulusBundle || isReactOrVue,
         deps: {
             neverBundle: Array.from(external),
-            onlyBundle: ['idiomorph'],
+            onlyBundle: isInspector ? [] : ['idiomorph'],
```
Kocal: Why?

bin/get_browsers.mjs
```diff
@@ -11,42 +11,44 @@ const installBrowserCommonOpts = {
...
-        ...installBrowserCommonOpts,
-        browser: Browser.CHROME,
-        // The lowest version where:
-        // - Chrome and associated Chromedriver could easily be downloaded
-        // - there is no compatibility issues like "WebDriver Bidi command \"session.subscribe\" failed with error"
-        // - there is no timeout issues when requesting Vitest webserver
-        // @see https://raw.githubusercontent.com/GoogleChromeLabs/chrome-for-testing/refs/heads/main/data/known-good-versions-with-downloads.json
-        buildId: '130.0.6669.0',
-    }),
+export async function getBrowsers(chromeLowest = '130.0.6669.0') {
```
Kocal: Use this syntax instead, so we can have named arguments:
```suggestion
export async function getBrowsers({ chromeLowest = '130.0.6669.0'}) {
```

src/Inspector/assets/src/inspector.css
Kocal, line 1: I believe your layers are too much generic and with conflict with the layers and styles present in Symfony apps (e.g.: when using Tailwind).


Can't we **really** scope them to UX Inspector **only**?

src/Inspector/assets/playwright.config.ts
```diff
@@ -0,0 +1,31 @@
+import { defineConfig } from '@playwright/test';
+import { getBrowsers } from '../../../bin/get_browsers.mjs';
+
+const browsers = await getBrowsers('145');
```
Kocal: 👎🏻 here, again you don't explain why you require Chrome 145 as lowest supported version, and I don't understand why you re-define the whole config file, instead of reusing  `import baseConfig from '../../../playwright.config.base';` 🤔

src/Inspector/assets/playwright.config.ts
```diff
@@ -0,0 +1,31 @@
...
+                browserName: 'chromium' as const,
+                channel: 'chrome',
+                launchOptions: { executablePath: browsers[`chrome@${version}`].executablePath },
+            },
+        })),
+        {
+            name: 'firefox-latest',
+            use: {
+                browserName: 'firefox',
+                launchOptions: { executablePath: process.env.INSPECTOR_FIREFOX_EXECUTABLE },
```
Kocal: `INSPECTOR_FIREFOX_EXECUTABLE`, where does it come from?

src/Inspector/assets/tsconfig.json
Kocal, line 1: Should be: 
```json
{
    "extends": "../../../tsconfig.package.json"
}
```

Again, please explain why you are doing such changes which are completely different from what we usually do, it would help us to understand.
Thanks!

src/Inspector/doc/index.rst
```diff
@@ -0,0 +1,187 @@
...
+====================
+
+Installation
+------------
+
+Requires PHP 8.4+, Symfony 7.4 or 8.x, and Chrome 145+ or Firefox 146+.
+
+.. caution::
+
+    This bundle is experimental. Register it only in ``dev``.
```
Kocal: This is not how we tell a package is experimetal, and we don't put requirements AFAIK

src/Inspector/doc/index.rst
```diff
@@ -0,0 +1,187 @@
...
+
+List Components
+---------------
+
+Click the small pull tab on the right edge of the screen to open the panel.
+It expands to show **UX** when the pointer approaches or the tab receives focus.
+On touch devices, **UX** is visible without hovering. The tab hides while the
+panel is open and returns when you close it.
+
+You can also type ``ux`` outside an editable field to open the panel.
```
Kocal:
```suggestion
You can also type the keys ``u`` and ``x`` outside an editable field to open the panel.
```

Kocal: no?

src/Inspector/doc/index.rst
```diff
@@ -0,0 +1,187 @@
...
+      - HTML attribute values and, when connected, default values and missing
+        CSS class attributes.
+    * - Targets and actions
+      - Elements and event bindings; checks for missing targets and action
+        methods when connected.
+    * - Relationships
+      - Outlets, parent and child controllers.
+    * - Activity
+      - Controller-prefixed custom events from ``data-action`` and a predefined
+        list of dispatch names, such as ``change`` in ``search:change``.
```
Kocal: @javiereguiluz told me that this syntax is not supported on symfony.com doc, and he was right. Please use the classic syntax instead.

Same comment for the rest of the file

src/Inspector/assets/test/setup.js
Kocal, line 1: Do we really need that?

src/Inspector/src/UXInspectorBundle.php
```diff
@@ -0,0 +1,111 @@
...
+        $debug = $builder->hasParameter('kernel.debug') ? (bool) $builder->getParameter('kernel.debug') : false;
+        $enabled = $config['enabled'] && $debug;
+
+        $inspectorConfig = [
+            'pull_tab' => $config['pull_tab'],
+            'ignore_selectors' => $config['ignore_selectors'],
+            'packages' => [
+                'stimulus' => class_exists('Symfony\\UX\\StimulusBundle\\StimulusBundle'),
+                'livecomponent' => class_exists('Symfony\\UX\\LiveComponent\\LiveComponentBundle'),
+                'turbo' => class_exists('Symfony\\UX\\Turbo\\TurboBundle'),
```
Kocal: Let's shorten these imports and directly use `StimulusBundle::class` etc?

src/Inspector/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# CHANGELOG
+
+## 3.5.0
+
+- Add the component.
```
Kocal:
```suggestion
- Add the component
```

☝🏻 🤓

splitsh.json
Kocal, line 1: Fine for `ux-dropzone`, but please keep bridges just after their _parent_ package. 
I mean, we should have the following order `ux-map`, `ux-google-map`, and `ux-leaflet-map` like before.

I can't imagine if something like that happened here https://github.com/symfony/symfony/blob/8.2/splitsh.json :p

Kocal (changes requested):
Good :)

Just a few comments, especially on our toolchain

src/Inspector/assets/src/inspector.css
smnandre: They are scoped into the shadow dom so they should have no impact at all

src/Inspector/assets/package.json
smnandre: es2021 increase bundle size: + 5,4 kB **gzip**  🤷‍♂️

smnandre:
Thank you @Kocal ! (and sorry for the mess )

src/Inspector/assets/src/inspector-element.ts
```diff
@@ -0,0 +1,155 @@
...
+        if (stylesheet?.text) {
+            const sheet = new CSSStyleSheet();
+            sheet.replaceSync(stylesheet.text);
+            shadow.adoptedStyleSheets = [sheet];
+        } else {
+            const link = document.createElement('link');
+            link.rel = 'stylesheet';
+            link.href = this.getAttribute('data-css-url') || stylesheet?.url || '';
+            shadow.append(link);
+        }
```
smnandre, line 45: @Kocal

src/Inspector/assets/package.json
Kocal: Thanks for the explanation! Nothing prevent us to upgrade the target version, **but** I would like this to apply to all our packages.

I tried with `es2022` and the size of file `src/Inspector/assets/dist/inspector.js` decreases, just as it did when you opened your PR.

I will open a PR

Kocal: Opened https://github.com/symfony/ux/pull/3882

src/Inspector/assets/src/inspector.css
Kocal: Aaah great!

src/Inspector/assets/src/inspector-element.ts
Kocal: Thx!

Kocal (approved):

src/Inspector/assets/src/ui/fields.ts
```diff
@@ -0,0 +1,253 @@
...
+                class: 'target-pill',
+                type: 'button',
+                'aria-label': `${labelStr}: show element on page`,
+                'aria-pressed': 'false',
+                on: {
+                    pointerenter: (event: Event) => emit(event.currentTarget, 'preview-element'),
+                    pointerleave: (event: Event) => emit(event.currentTarget, 'clear-element-preview'),
+                    focus: (event: Event) => emit(event.currentTarget, 'preview-element'),
+                    blur: (event: Event) => emit(event.currentTarget, 'clear-element-preview'),
+                    click: (event: Event) => {
```
squrious, line 136: I noticed some strange behaviors coming from here. 

Click on the button shows the frame around the target in the page, pointer icon changes, `aria-pressed` is `true` => correct. 

Then: 
- Click the button again doesn't revert `aria-pressed` to false, and doesn't deselect the element in the page
- Click on another button (another Stimulus target, data-model directive, etc.) in the inspector doesn't revert other's `aria-pressed` to `false`

Also, not sure if it comes from my very-AI-generated-playground, but I have a button at the bottom of a static navbar like this: 

Image

Each time I click on the button target in the inspector, the page scrolls down a little even if the button is already visible.

src/Inspector/assets/src/ui/panel.ts
```diff
@@ -0,0 +1,511 @@
...
+            {
+                id: 'panel-activity',
+                hidden: true,
+                role: 'region',
+                tabindex: '0',
+                'aria-label': 'Activity',
+            },
+            this.#timeline.element
+        );
+        this.#activity = new ActivityDrawer(timeline, this.#activityPanel, () => {
```
squrious, line 116: For some reason, the global activity panel is not scrollable like the nested ones are in component details. It could be nice to allow scrolling into it.

Global will likely have way more entries that could never be inspected otherwise: 

image

src/Inspector/assets/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": "../../../tsconfig.package.json"
```
squrious: Base `tsconfig.package.json` declares: 
```json
    "compilerOptions": {
        "lib": ["dom", "es2015"]
    }
```

In the code there are (at least) these features that require more: 
- `...someEl.querySelectorAll(...)`: requires `dom.iterable`
- `'someString'.replaceAll(...)`: requires `es2021` or later
- `Object.groupBy(...)`: requires `es2024` or later

src/Inspector/assets/src/ui/activity-drawer.ts
```diff
@@ -0,0 +1,75 @@
...
+                drawer.toggleAttribute('data-sized', true);
+                drawer.style.height = `${value}px`;
+                this.#sizes.set(content, value);
+                return value;
+            },
+        });
+        drawer.prepend(this.#resizeHandle.element);
+        this.#drawer = drawer;
+        this.#id = id;
+        this.#timeline.configure({ contextual: true, frameworks: null, element, query: element ? '' : query });
```
squrious: When inspecting activity for a specific component, timeline is always configured for no specific framework (`null`).

In my playground I have a Stimulus controller for theme switcher on the `html` element, so all "components" are children of my `theme` controller.

I don't really get how this works, but the `theme` Stimulus activity panel catches turbo events too:

Image

Not sure if it's expected behavior or not, my setup is what it is. But this could be generalized to more legit use cases I suppose.


Another found behavior: the component's activity panel fetches and filters events from the global monitor. Each time an event is recorded, even if it doesn't relate to the currently inspected component, it makes the panel refreshed. 

Observed consequences: 
- A kind of "blinking" effect when the timeline refreshes, even if nothing changes
- The icon of the activity panel is always colored based on the latest event recorded globally
- Depending on the amount of new events, alternate background colors can mess up, as they're set _before_ filtering
- Too many filtered out events will eventually clear the timeline entirely. I have a live component with polling, and I'm inspecting a parent Stimulus controller. The live component poll events are recorded but filtered. At some point, there is more filtered poll events than the max events allowed in the timeline.

src/Inspector/assets/src/ui/panel.ts
```diff
@@ -0,0 +1,511 @@
...
+                { class: 'monitor-status', role: 'status' },
+                el('span', { class: 'live-dot', 'aria-hidden': 'true' }),
+                this.#monitorLabel
+            )
+        );
+    }
+
+    #updateCounts(): void {
+        const byPlugin = this.#state.countByPlugin();
+        const activityByPlugin: Record<string, number> = {
```
squrious: This map is used later to compute the global records count (L.495).

From what I understand: 
- the monitor keeps at most 500 entries in memory
- the timeline allows at most 100 entries to be displayed

With a long living polling live component, the count ends frozen at 167.

So I suspect a little bug here. But let's say we reach the 500 limit, shouldn't we keep the total count of events including ones that have been dropped?

src/Inspector/assets/src/ui/panel.ts
```diff
@@ -0,0 +1,511 @@
...
+    #frameworkButtons(
+        filters: Set<string>,
+        buttons: Record<string, HTMLButtonElement>,
+        update: () => void
+    ): HTMLButtonElement[] {
+        return FRAMEWORKS.map(([name, label]) => {
+            const button = el(
+                'button',
+                {
+                    class: 'filter active',
```
squrious, line 424: UX suggestions: 

Framework buttons could use their respective color scheme (green/yellow/purple) in some way.

Also I always misuse these buttons. Most of the time I click them to display Stimulus, or Live, or Turbo. But it filters out what I wanted, so I remember I have to click the two others. The strong dark theme doesn't help too, selected is darkest than the dark deselected 😅 

Probably a matter of taste, but is there really a use case to display a specific combination of frameworks?

Image

src/Inspector/assets/src/live-component/render.ts
```diff
@@ -0,0 +1,102 @@
+import { makeField, makeElementField, makeGroup, makeKeyValueList } from '../ui/fields';
+import type { RenderContext } from '../types';
+import type { LiveData } from './plugin';
+export function renderLive(d: LiveData, context: RenderContext = {}): DocumentFragment {
+    const frag = document.createDocumentFragment();
+    const runtime = d.runtime;
+    const changes = context.changes as Map<string, { previous?: unknown }> | undefined;
+    const props: HTMLElement[] = [];
+    if (['updating', 'error', 'disconnected'].includes(runtime.status)) {
```
squrious, line 9: This is not really a prop, hope this will not conflict with a real prop named "status".

Also, maybe the state indicator could be always rendered, with a "ready" status most of the time?

Updating often take a few ms, I never had the time to really see the status. It's often more a visual flickering than a real indicator IMO. 

Screenshot for reference
Image

src/Inspector/assets/src/ui/component-detail.ts
```diff
@@ -0,0 +1,370 @@
...
+                    ? makeGroup(title, [el('div', { class: 'relations' }, ...groups[key])], {
+                          key: `${framework}-${key === 'parents' ? 'parent' : key}`,
+                          icon: key === 'outlets' ? 'overlay' : 'components',
+                      })
+                    : null
+            )
+            .filter((group): group is HTMLElement => Boolean(group));
+    }
+
+    #describeRelationship(edge: GraphEdge, target: Element): [RelationKind, string] {
```
squrious, line 271: Possible UX improvement: indicate the framework associated with the parent/child.

We may use related color scheme like in the main components list.

Image

squrious (commented):
Hi @smnandre! Here is a more functional oriented review.

Mostly UX suggestions and minor improvements, one or two possible bugs. The component works really well 😄 

For the `tsconfig.json` thing I have to check the build script a bit more.

Hope this helps!

src/Inspector/assets/src/ui/panel.ts
smnandre: Fixed. Activity now scrolls independently.

src/Inspector/assets/src/ui/fields.ts
smnandre: Fixed. Pills now toggle and stay in sync.

src/Inspector/assets/tsconfig.json
smnandre: Fixed. Inspector now uses the shared ES2022 target.

src/Inspector/assets/src/ui/activity-drawer.ts
smnandre: Fixed. Activity is filtered before applying the limit.

src/Inspector/assets/src/ui/panel.ts
smnandre: I'd keep this open until we agree on the expected count.

src/Inspector/assets/src/ui/panel.ts
smnandre: Agreed. I'd handle this in a separate UX pass during "alpha" release I think

src/Inspector/assets/src/live-component/render.ts
smnandre: Agreed. A dedicated state indicator would be clearer.

src/Inspector/assets/src/ui/component-detail.ts
smnandre: Good idea!

src/Inspector/src/UXInspectorBundle.php
Kocal, line 1: No integration with AssetMapper here?
I just want to confirm if you need to re-apply https://github.com/symfony/ux/pull/3867 for Inspector, or not

smnandre: No need in the end, as we expose the script via an end point (it was almost impossible to expose it only in debug, and it was a hard no for someone to have this script in the public Importmap for instance)

squrious (approved):
Thanks for the updates @smnandre!

For the `tsconfig.json` config, that makes my IDE happy again 😄 
I looked a bit more in depth and actually, these libs are already "shipped" with the new es2022 target, but overridden by the main `compilerOptions.lib` setting. This topic is going out of this PR scope IMO.

PR #3872, opened 2026-09-14
[Toolkit][Shadcn] Add questionnaire recipe

ker0x:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes
| Deprecations?  | no
| Documentation? | yes
| License        | MIT

Adds the `questionnaire` recipe to the Shadcn kit, ported from [`registry/bases/radix/ui/questionnaire.tsx`](https://github.com/shadcn-ui/ui/blob/main/apps/v4/registry/bases/radix/ui/questionnaire.tsx).

A multi-step questionnaire with single-choice, multiple-choice, freeform and skippable questions, plus progress, navigation and keyboard shortcuts.

### Composition

```
Questionnaire
├── Questionnaire:Progress
├── Questionnaire:Item
│   ├── Questionnaire:Title
│   ├── Questionnaire:Description
│   ├── Questionnaire:Choices
│   │   ├── Questionnaire:Choice
│   │   │   └── Questionnaire:ChoiceDescription
│   │   └── Questionnaire:Input
│   └── Questionnaire:Error
└── Questionnaire:Actions
    ├── Questionnaire:Previous
    ├── Questionnaire:Skip
    ├── Questionnaire:Next
    └── Questionnaire:Submit
```

```twig
<twig:Questionnaire defaultItem="direction" shortcuts="letters">
    <twig:Questionnaire:Progress />

    <twig:Questionnaire:Item name="direction" required>
        <twig:Questionnaire:Title>What should we prototype next?</twig:Questionnaire:Title>
        <twig:Questionnaire:Description>Choose a direction or write your own.</twig:Questionnaire:Description>
        <twig:Questionnaire:Choices>
            <twig:Questionnaire:Choice value="delegation">Delegation</twig:Questionnaire:Choice>
            <twig:Questionnaire:Choice value="questions">Question prompts</twig:Questionnaire:Choice>
            <twig:Questionnaire:Input aria-label="Another answer" placeholder="Type another answer…" />
        </twig:Questionnaire:Choices>
        <twig:Questionnaire:Error />
    </twig:Questionnaire:Item>

    <twig:Questionnaire:Actions>
        <twig:Questionnaire:Previous />
        <twig:Questionnaire:Skip />
        <twig:Questionnaire:Next />
        <twig:Questionnaire:Submit />
    </twig:Questionnaire:Actions>
</twig:Questionnaire>
```

### Implementation notes

- **Upstream styling.** The reference component carries most of its visuals in `cn-questionnaire-*` classes defined in `registry/styles/style-nova.css`, which the Toolkit does not ship. Those rules are inlined as Tailwind utilities in the templates, the same approach used by the `marker` recipe.
- **Native form semantics.** The root renders a `form`, `Item` renders a `fieldset` with `Title` as its `legend`, and choices are native radios / checkboxes. Checked styling is pure CSS (`has-[>input:checked]:`), so no JS is needed to keep it in sync. An item marked `multiple` submits an array and names its inputs `<name>[]`.
- **Freeform answers.** `Questionnaire:Input` shares the item name with the fixed choices, so answering one clears the other. While the input is empty its `name` is removed, otherwise it would submit an empty value that shadows the selected answer.
- **State surface.** `Item` exposes `data-active` and `data-status` (`unanswered` / `answered` / `skipped`) for styling, as shown in the *Navigation State* and *Animated Items* examples.

### Tests

- All 14 upstream examples are reused as live-preview blocks in `README.md`, plus a `Server-Driven Flow` and an `RTL` section — 17 rendering snapshots, all passing.
- The Stimulus controller was exercised in Chromium against the rendered markup (53 assertions): navigation, required-item validation, skip/reset, letter and number shortcuts, multiple selection, disabled items being stepped over, custom progress, custom error messages, the submitted `FormData`, the controller-free server-driven markup, and the questionnaire nested as a `div` inside an outer form.
- `bin/ux-toolkit-kit-lint --fail-on-warning kits/shadcn`, `twig-cs-fixer`, `oxlint` and `oxfmt` are clean.

### Server-driven rendering

`:interactive="false"` renders the questionnaire as plain markup — no `data-controller`, no `data-action` and no `data-questionnaire-*` anywhere in the tree — so the server can own the flow, as with a Symfony form flow. The flag is published with `provide()` and read by every descendant with `inject()`.

Because stripping the attributes alone would leave unusable markup, that mode also shows every item when `defaultItem` is unset, renders `Questionnaire:Error` without `hidden`, and makes `Previous` / `Skip` / `Next` `type="submit"`. `Questionnaire:Progress` gained `current` / `total` props, since its ARIA values were hardcoded and updated by JS and `aria-*` cannot live in `defaults()`.

`Questionnaire` also gained an `as` prop, so it can render a `div` and be composed inside a form the application already renders instead of nesting two `form` elements. The `README` documents this against Symfony 8.1's `form_flow_*()` and `field_*()` Twig functions.

[questionnaire.webm](https://github.com/user-attachments/assets/1f2652e7-6a88-4169-88df-2dc081fa244e)

### Note for reviewers

`Questionnaire:Previous`, `Skip`, `Next` and `Submit` wrap `<twig:Button>` and pass `data-slot="questionnaire-*"`, but `Button.html.twig` emits a literal `data-slot="button"` that wins, so the rendered buttons carry `data-slot="button"`. This is pre-existing behaviour shared with `AlertDialog:Action`, `AlertDialog:Cancel`, `InputGroup:Button` and `Pagination:Link`; nothing functional depends on it here, since the controller uses Stimulus targets. Happy to address it separately in `Button` if wanted.


ker0x:
I have added two new props to the main `Questionnaire` component. One allows you to disable interactivity and remove the Stimulus controller, whilst the other allows you to display the component in a format other than a `form`. This is intended to enable the component to be used with `FormFlow`, from the Symfony Form component. A short example of how to use it has been added to the README.

Kocal:
I'm on it

Kocal:
Hey @ker0x, before merging I rebased this onto 3.x now that the calendar and date-picker recipes are in, and squashed everything into a single commit. Here's what else changed:

- `Choice`, `Choices` and `ChoiceDescription` now compose the `field`, `radio-group`, `checkbox` and `kbd` recipes instead of restating their markup and classes, so those four are declared as recipe dependencies and `Choices` picks up the `role="radiogroup"` it was missing.
- Validation now matches upstream: the default message moved from the Stimulus controller into `Questionnaire:Error` so it can be translated and follows the item's `required` prop, the choices of an invalid item get `data-invalid`, and the first choice takes focus with a visible ring.
- `Questionnaire:Next`, `:Previous`, `:Skip` and `:Submit` were setting their base `class` as a literal attribute before the attribute spread, so a caller passing `class` silently dropped the grid placement, they now merge through `attributes.defaults()`.
- Prop and block docs were migrated to the Twig 3.29 comment syntax the kit now requires, added an `## Accessibility` section, and regenerated the snapshots.

Kocal:
Thank you @ker0x.

PR #3878, opened 2026-09-16
[Toolkit][Shadcn] Fix the rich colors of Sonner recipe

### LINKED ISSUE #3874: [Toolkit][shadcn] Sonner warning background color does not work
The warning color on the Sonner toasts doesn't show up (out of the box). The toast's background should be yellow, but instead it is white (for me).

Looks like the issue is that both bg-background and bg-amber-50 get added to the toast in sonner_controller.js. The rich color classes are concatenated to the base classes in #createToast(), and in the case of warning, bg-background wins (because bg-background comes after bg-amber-50 alphabetically, maybe?)

To reproduce:
- Install symfony/ux-toolkit 3.4
- Install the shadcn Sonner component as per https://ux.symfony.com/toolkit/kits/shadcn/components/sonner
- Create a Sonner component with richColors="1"
- Add a Toast with type="warning"
- Reload the page

In my sonner_controller.js, I fixed it by separating the bg-background and text-foreground classes into a separate const (SONNER_BASE_COLORS) and concatenated those in #createToast() if rich colors were not applied.

Other than that - love this component! I'm using the (undocumented?) 'sonner' browser event to get toasts from Live Components, in addition to server-rendered toasts for standard flash messages. Works great!

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3874 
| License        | MIT

https://github.com/user-attachments/assets/9bd7cb1c-137a-448b-b425-35924cb6cc95


src/Toolkit/kits/shadcn/sonner/assets/controllers/sonner_controller.js
```diff
@@ -220,6 +219,19 @@ export default class extends Controller {
...
+        const richColors = this.richColorsValue ? SONNER_RICH_COLORS[type] : null;
+
+        if (!richColors) {
+            li.classList.add(...SONNER_DEFAULT_COLORS);
+
+            return;
+        }
+
+        li.classList.remove(...SONNER_DEFAULT_COLORS);
+        li.classList.add(...richColors.split(' '));
```
Kocal, line 232: We don't have/want to introduce Tailwind merge JS side, so let's doing things like the gold old way :)

PR #3880, opened 2026-09-16
[LiveComponent] Handle paths beginning with double slash in LiveUrlSubscriber

### LINKED ISSUE #3856: [LiveComponent] `LiveUrlSubscriber` crashes / silently rewrites the URL when `X-Live-Url` has no path
`LiveUrlSubscriber::generateNewLiveUrl()` reads the `X-Live-Url` **request** header and assumes `parse_url()` always returns a `path` key:

```php
// src/LiveComponent/src/EventListener/LiveUrlSubscriber.php
private function generateNewLiveUrl(string $previousUrl, array $pathProps, array $queryProps): string
{
    $previousUrlParsed = parse_url($previousUrl);
    $newUrl = $previousUrlParsed['path'];                          // no guard
    ...
        $routeMatched = $router->match($previousUrlParsed['path']); // no guard
```

`parse_url()` returns **no `path` key** for several inputs that reach production, and `false` for others:

| `X-Live-Url` | `parse_url()` | `path`? |
|---|---|---|
| `/products?q=1` | `{path, query}` | yes |
| `//products` | `{host: "products"}` | **no** — `//` is parsed as protocol-relative |
| `?q=1` | `{query}` | **no** |
| `#x` | `{fragment}` | **no** |
| `//` | `false` | **no** |

The header is *not* attacker-only input. `live_controller` sends `window.location.pathname + window.location.search`, and `pathname` keeps a **doubled leading slash** whenever the user landed on a URL like `https://example.com//some-page`. Web servers routinely normalize that server-side (nginx `merge_slashes on` is the default), so the page renders `200` with no redirect and the browser keeps `//some-page` in the address bar. Every subsequent Live request then carries a pathless `X-Live-Url`. A single mistyped link in a newsletter or an external backlink is enough.

### Impact

1. **500 for components using `mapPath`.** With a non-empty `$pathProps`, `$router->match(null)` is called. `UrlMatcherInterface::match(string $pathinfo)` is userland, so `null` is not coerced: `TypeError: match(): Argument #1 ($pathinfo) must be of type string, null given`. The surrounding `try` only catches `ResourceNotFoundException`, so the TypeError escapes and the request 500s.

2. **Silent, user-visible URL rewrite for query-mapped props.** With `mapPath: false`, `$newUrl` is `null`, so the response header becomes `"?filter=x"`. `live_controller` then does:
   ```js
   history.replaceState(history.state, "", new URL(liveUrl + location.hash, location.origin))
   ```
   `new URL("?filter=x", origin)` resolves to `https://example.com/?filter=x` — **the path is dropped**. The user is on a category page, applies a filter, and the address bar silently becomes the homepage. Reload, bookmark or share now lands on the wrong page.

3. **Log noise.** `Warning: Undefined array key "path"` on every affected request. Apps that promote warnings to exceptions (default Monolog `ErrorHandler` setup) log it at ERROR level, which trips alerting.

### Reproduce

```php
$parsed = parse_url('//some-page');   // ['host' => 'some-page']
$parsed['path'];                      // Warning: Undefined array key "path" → null
```

End to end, with any app running nginx on its defaults:
1. Add a `#[LiveProp(url: true)]` to a component on `/some-page`.
2. Visit `https://example.com//some-page` — it renders `200`, the address bar keeps the `//`.
3. Interact with the component → warning logged, and the address bar is rewritten to `https://example.com/?...`.
4. With `mapPath: true` instead → `TypeError`, HTTP 500.

### Suggested fix

Guard both accesses, and bail out when the header is unusable:

```php
$previousUrlParsed = parse_url($previousUrl);

if (false === $previousUrlParsed || '' === ($path = $previousUrlParsed['path'] ?? '')) {
    return $previousUrl;   // or return early from onKernelResponse()
}

$newUrl = $path;
// ... and use $path in $router->match($path)
```

Returning early seems preferable to defaulting to `'/'`: a pathless `X-Live-Url` carries no usable location, and silently rewriting the client's URL to the site root is exactly the bug in (2).

Happy to open a PR if the approach looks right.

**Versions:** symfony/ux-live-component v2.32.0; also present on `2.x` HEAD at time of writing.

mbuliard:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #3856  
| License        | MIT

Nginx servers have a default configuration `merge_slashes` that rewrites silently URIs beginning with consecutive slashes, e.g. :
`https://domain.com//products` will send `/products` to `$_SERVER['REQUEST_URI']`.
Therefore the Symfony router will get the normalized URI, without duplicated slashes and match it against the `/products` route.
**BUT** the URI received by `LiveUrlSubscriber` is not affected by this nginx normalisation and then keep the duplicated slashes.
Trying to `parse_url` these URI will not work.

Two things introduced in this PR :
* `explode` the received URI instead of `parse_url`. Simpler, more robust. Same logic as in `Symfony\Component\HttpFoundation\Request`.
* Remove the consecutive slashes before trying to match the path with the router.

To discuss :
* Should it be handled at all ? Symfony is stricter than RFC 3986 and does not allow theses double slashes in its Routing. It works only by nginx configuration but is not supported per se by Symfony. The LiveUrlSubscriber could fallback on the current URI when going through an error.
* The PR adds fallbacks to handle a fragment (`#`) if present in the URI, but I'm not sure if it is expected to.
* Same, the PR adds fallbacks to handle a full URL, but not sure if expected to.


smnandre:
> Should it be handled at all ?

I'm very not sure. If `example.com//foo` is a valid URI, it also means it is different from `example.com/foo` so I am not sure why we should handle one for the other.

Now, to be honest, this should be something dealt with 301 redirection on the server side (as I see it)

So i'm not sure we want to add maintenance surface here ... wdyt ?

mbuliard:
> So i'm not sure we want to add maintenance surface here ... wdyt ?

I would keep the change from `parse_url` to `explode`, as Symfony did it and it adds no surface.
But handling the double slashes seems out of scope. While the RFC allows it, Symfony do not allow a route with it, so it should not happen.

> A pattern must start with a slash and must not have multiple slashes at the beginning because the generated path for this route would be confused with a network path, e.g. '//domain.com/path'.
> src/LiveComponent/vendor/symfony/routing/Route.php:113

mbuliard:
The PR is updated to reduce the change to "`parse_url` to `explode`" and not handle more

smnandre:
Thank you @mbuliard !  On my side, i'm 👍 

@Kocal @kbond do you agree with the orientation / decisions there?

smnandre (approved):

Kocal:
Thank you @mbuliard.

PR #3882, opened 2026-09-17
[Autocomplete][Chartjs][Cropperjs][Dropzone][LiveComponent][Map][Notify][React][StimulusBundle][Turbo][Vue] Update TypeScript's `target` to `es2022`

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | no
| New feature?   | yes 
| Deprecations?  | no 
| Documentation? | no 
| Issues         | Fix #... 
| License        | MIT

Related to https://github.com/symfony/ux/pull/3866#discussion_r4033685337. I hope it won't break our Stimulus controllers this time 🤞

No BC breaks here, we still support a veeeery large amount of browsers:

| Browser | ES2021 | ES2022 | Delta |
| --- | --- | --- | --- |
| Chrome / Edge | **85** (Jul 2020) | **94** (Sep 2021) | +9 versions |
| Firefox | **79** (Jul 2020) | **93** (Oct 2021) | +14 versions |
| Safari (macOS) | **14.1** (Apr 2021) | **16.4** (Mar 2023) | +2 years |
| Safari iOS | **14.5** (Apr 2021) | **16.4** (Mar 2023) | +2 years |
| Limiting feature | `WeakRef` / `FinalizationRegistry` | `static {}` blocks | — |
| Global coverage | **93.72%** | **93.18%** | -0.54 pt |
| Baseline | Widely available | Widely available | — |

Also added some shiny badges on the root README.md 🙈


smnandre:
First of all: 👍 

---

Then, I do think we should document this in our "own" BC policy, stating clearly: we are working with browsers. Difference with back-end php code hosted on servers developers controls, browsers are in userland and have them constraints and reality. They now all self-update, OS force updates on a year basis (worst case). 

I think we want to promote all the security, privacy, performance related features that has been introduced these last years, and will be introduced (quicker and quicker) the next ones ... 

ESXXX is a developer-oriented data. Baseline is the "feature-oriented" one now, and it does cover most of what we check/discriminate againts in JS/TS (as we don't offer node apps but transpiled/minified scripts).

So I do think we should act on a Baseline version, and I'd suggest we use "Widely available" as base, and allow ourselves "Newly available" for new components and dev tools. 

https://web.dev/baseline

> **Newly available**: The feature is supported by all of the core browsers, and is therefore interoperable.
>
> **Widely available**: 30 months have passed since the newly interoperable date. The feature can be used by most sites without worrying about support.

I do thing there are equivalence tables for what it means in term of ES versions, browser prefixes, etc etc.

Do you want me to open an RFC on this ? (to discuss it)

smnandre (approved):

Kocal:
> Do you want me to open an RFC on this ? (to discuss it)

You can, and I totally agree with what you're saying above.

Note that tsdown, the tool we use to build our assets, supports the [special target `baseline-widely-available`](https://tsdown.dev/options/target#supported-targets) :D

PR #3887, opened 2026-09-18
[Toolkit][Shadcn] Document accessibility, fix `<dialog>`-based recipies accessibility, and improve right-to-left layouts

Kocal:
| Q              | A
| -------------- | ---
| Bug fix?       | yes
| New feature?   | no
| Deprecations?  | no
| Documentation? | yes
| License        | MIT

Every Shadcn recipe now carries an `## Accessibility` section in its `README.md`, between the examples and the API reference. Upstream only ships one for `attachment`, `button-group`, `field` and `marker`, so those four are ported and the rest are written from what the templates and Stimulus controllers actually emit: roles, ARIA wiring, keyboard handling, and what the consumer still has to provide.

Writing them surfaced three bugs. `alert-dialog`, `dialog`, `drawer` and `sheet` rendered `aria-labelledby` and `aria-describedby` on a roleless wrapper instead of the `<dialog>` that `showModal()` opens, leaving the modal without an accessible name. `Button` set `type="button"` unconditionally, producing an invalid `<a type="button">` for the `as="a"` case its own README documents. `badge` and `item` declared only their default tag in `@prop as` while both READMEs use `as="a"`.

The right-to-left pass replaces physical spacing, alignment and border utilities with their logical counterparts, and drops the `ltr:<physical> rtl:<logical>` pairs that resolved to the same rule anyway. `toggle`, `toggle-group`, `sidebar`, `sonner`, `menubar` and `combobox` had no right-to-left handling at all. The `side` and `direction` props stay physical, matching Radix, where only `align` follows the text direction.


Menelion:
@Kocal I just call to say thank you. We, people with accessibility needs, exist, as developers too. For instance, I'm a blind developer myself and all of my PHP projects are powered by Symfony and now by Symfony UX. Thank you for caring, it really is hugely appreciated and it means a lot! Keep going, great job!

Kocal:
@Menelion Thank you for your kind words, that’s very kind! We're doing our best, but please do let us know if you ever spot any accessibility issues in Symfony UX! 🙌🏻

